PR #152, opened 2012-08-07
"Save to Zotero" panel

simonster:
This patch adds a Zotero button to the toolbar. This toolbar opens a panel that can translate the current page, save a webpage item, or perform library lookup.

Implementation issues:
- New web page items are always saved to the currently selected collection.
- ~~The "Lookup" button seems to misbehave on rollover.~~
- ~~When you select a collection, it's added to the top of the "Recently Used" list. This seems strange if the selection is on the main screen.~~

Bigger questions:
- Do we put this button in the toolbar by default?
- Does this need a different icon to more clearly differentiate it from the status bar button?
- Does this replace "Create New Item from Current Page"?
- Should generic webpage items be saved via a translator? If so, should this translator be "Embedded Metadata" or something else?
- When we detect a webpage item, should we hide the URL bar icon?

TODO:
- Localization
- Implement for connector
- Implement for bookmarklet


bwiernik:
This has largely been superseded by the current Connector interface, no?

PR #175, opened 2012-09-27
Add counters to notes, tags, and related tabs.

aurimasv:
Continuation of #167

One concern I have is that even though I choose a counter label width of 1.9em, which is wide enough for (99) and should expand with increased font size, I think this strategy fails when changing font styles (it looks like Mac uses different font). We may have to do a bit of a hack like the one suggested by @rmzelle

Also, regarding overflow with different locales, I added a scroller that appears when tabs overflow.


adam3smith:
are you closing this because you're working on a different/better implementation or because you've decided it's not a good idea?

aurimasv:
I think a bunch of the PRs were automatically closed, because they were against the master branch, which got deleted.

dstillman:
No longer relevant in Z7 (and there are counts)

PR #309, opened 2013-04-19
Convert all preference tabs to use tabbox

aurimasv:
I got a bit annoyed with the preference window changing size when switching tabs, so I converted all the tabs to use tabboxes. Now everything stays consistent when switching tabs.

I'm sure this will need more tweaking. I'll upload screenshots a bit later.


rmzelle:
> The main problem with the resizing is that it's always been a little broken.

I've been searching for solutions to the issue with pref pane cropping under OS X, and found http://forums-test.mozillazine.org/viewtopic.php?p=4121285&sid=13fd2850cbc36408e3fb6418396c19ec#p4121285 and http://www.tek-tips.com/viewthread.cfm?qid=1627464 . Maybe the latter link offers a solution?

dstillman:
No longer relevant in Z7

PR #313, opened 2013-04-23
add more item handler protocols to add by URI

adam3smith:
https://forums.zotero.org/discussion/8188/attachment-add-link-to-any-url-not-current-page/#Item_39

### DIFF
--- a/chrome/content/zotero/xpcom/attachments.js
+++ b/chrome/content/zotero/xpcom/attachments.js
@@ -211,20 +211,10 @@ Zotero.Attachments = new function(){
 			parentCollectionIDs = undefined;
 		}
 		
-		/* Throw error on invalid URLs
-		   We currently accept the following protocols:
-		   PersonalBrain (brain://)
-		   DevonThink (x-devonthink-item://)
-		   Notational Velocity (nv://)
-		   MyLife Organized (mlo://)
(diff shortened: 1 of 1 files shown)


andersjohansson:
Just chiming in (in this very old discussion, sorry) to say that I did find a use for zotero-links here!
I have the pdf for an edited volume attached to an item for the whole book. But I also need to refer to single chapters and thus add items for those. How can I open the relevant part of the pdf by double-clicking on the chapter item? By adding this link: `zotero://open-pdf/0_RVPGM4QC/257`. It works great but is of course messy to create. Zutilo's "Copy select item link" helped me to get the id for the relevant attachment.

andersjohansson:
The hard-coded whitelist seems to prevent me (it just silently fails) from adding org-protocol links now in 5.0 beta (and not before, which seems strange. It seemed to work last month on the most recent version 4).

Are there any plans on adding a configurable whitelist? Are there any problems with just letting the regex-string be modifiable as a config-variable (those who want to change it should know what they're doing)?

Don't know if this is the most reasonable place to discuss this ...
(I posted here as well: https://forums.zotero.org/discussion/64155/5-0-beta-add-link-to-uri-is-now-filtered-maybe-whitelist-could-be-customizable)

PR #394, opened 2013-09-25
Add hyphenateISBN to Zotero.Utilities

aurimasv:
Re https://forums.zotero.org/discussion/32110/to-show-isbn-with-hyphens-or-space/

The registrant table is quite long (~13kB, 2.4kB gzipped), so maybe it's not worth it.


aurimasv:
Superseded by #626

Issue #178, opened 2012-10-05
Keyboard shortcut for "Create New Item from Current Page"


muescha:

dstillman:
Addressed by 9a87f1595660a1715f3448172d6ce008e308adec

PR #403, opened 2013-10-27
Add "Create Web Page Item from Current Page" keyboard shortcut, followin...

willsALMANJ:
...g the example of the "Save to Zotero" shortcut.  Address zotero/zotero#178

A few comments:

* I tried to follow the style of `key_saveToZotero`, so I think this should be easy to review.
* I was not sure what to choose for a good default key.  The best choices mnemonically are taken (`C` and `N` by Zotero, `W` and `P` by Firefox).  I went with `I` for now.
* I did not add a case to `handleKeyPress()` of ZoteroPane.js because `key_saveToZotero` does not currently have one.  So, like `key_saveToZotero`, `key_newItemFromPage` can not be called when Zotero is focused.  I was not sure if this was the intended behavior for `key_saveToZotero` or an oversight.

### DIFF
--- a/chrome/content/zotero/overlay.xul
+++ b/chrome/content/zotero/overlay.xul
@@ -98,5 +98,9 @@
 			 key="S"
 			 oncommand="Zotero_Browser.scrapeThisPage();"
 			 modifiers="accel shift" />
+		 <key id="key_newItemFromPage"
+		 	 key="I"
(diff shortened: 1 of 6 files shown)


willsALMANJ:
Superseded by https://github.com/zotero/zotero/commit/9a87f1595660a1715f3448172d6ce008e308adec

Issue #356, opened 2013-08-01
Add "Save to Zotero" options to web page content menu


dstillman:
While there's [now](https://github.com/zotero/zotero/commit/dc8998c5bc1e582173420efae7647c3add82d62c) a shortcut key for the main Save to Zotero action, the context-menu options for the address bar icon aren't accessible, nor is the notification that there's something to save via a translator in the first place.

For accessibility, the Zotero submenu of the web page context menu should provide access to all the Save to Zotero actions.

dstillman:
No more Z4Fx, and this is done in the connectors

PR #404, opened 2013-10-28
Add "Save to Zotero" submenu to the Firefox content area context menu.  ...

willsALMANJ:
...Use the same menupopup as used for the status bar icon for the submenu.  This addresses zotero/zotero#356 .

A couple comments:
- I used the same context menu generating function used by the status bar icon for the content area context menu.  I think this is fine, but in the future it could be confusing development-wise that the `onStatusPopupShowing` function is also used for the content area context menu.  I added a comment to `onStatusPopupShowing` to point this out.
- I set the hidden status of the submenu of the content area context menu based on the hidden status of `statusImage.hidden`.  I think this is okay because it should always be set before the context menu can be shown, but again the name could be confusing.
- I only added the context menu.  A "Save to Zotero" menuitem equivalent to clicking on the status bar icon could also be added directly to the Zotero submenu but maybe it's simpler just to have the "Save to Zotero" options in the separate submenu of the Zotero submenu.

### DIFF
--- a/chrome/content/zotero/browser.js
+++ b/chrome/content/zotero/browser.js
@@ -411,7 +411,8 @@ var Zotero_Browser = new function() {
 	}
 	
 	/**
-	 * Called when status bar icon is right-clicked
+	 * Called when status bar icon is right-clicked or the "Save to Zotero" submenu
+	 * of the contentAreaContextMenu is shown
 	 */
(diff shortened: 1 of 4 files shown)


dstillman:
The full Zotero for Firefox is being discontinued, so this is no longer relevant. The connectors add save options to the content area context menu.

PR #409, opened 2013-11-05
Manage cookies received from other hosts.

aurimasv:
Currently we ignore any cookies set from other domains while attempting to download attachments (through connector imports) and I'm not entirely sure why.

One issue is downloading attachments when importing via Chrome from cell.com (e.g. http://www.cell.com/abstract/S0092-8674(12)01407-9) Some of their PDFs are hosted on sciencedirect.com so we pass the sciencedirect.com url as attachment url to Zotero Standalone. We don't necessarily need to pass any cookies when fetching these attachments to start with, but sciencedirect does include a redirect (or two) that does set cookies, which end up being required to download the PDF.

This patch allows cookies to be set for different domains during this transaction and it implements domain-level cookie management within the cookie sandbox. It certainly solves the cell.com issue ~~and it may also solve some of the issues mentioned here: https://forums.zotero.org/discussion/32937/pdf-saving-problem/~~

### DIFF
--- a/chrome/content/zotero/xpcom/connector/connector.js
+++ b/chrome/content/zotero/xpcom/connector/connector.js
@@ -136,7 +136,7 @@ Zotero.Connector = new function() {
 	 * @param	{Object}		data			RPC data. See documentation above.
 	 * @param	{Function}		callback		Function to be called when requests complete.
 	 */
-	this.callMethod = function(method, data, callback) {
+	this.callMethod = function(method, data, callback, tab) {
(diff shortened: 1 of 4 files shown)


aurimasv:
Another reason to get this resolved: https://forums.zotero.org/discussion/29354/3/query-limit-reached/#Item_28 (blocking Google Scholar captchas for retrieve metadata)

simonster:
Thanks!

PR #442, opened 2014-01-12
Generate valid html reports

retorquere:


dstillman:
Sorry, this got closed when I briefly deleted the master branch, and the original repo is gone. This code has since diverged quite a bit on the master branch, though, and the protocol handler is also being completely redone to mirror the server API, so any changes to reports would have to wait for that. (The same code might be able to be used against both the local client and the server — or, at least, that's the goal — so a reworking of reports might be a neat project.)

retorquere:
It can wait. I have made a neat nokogiri-builder clone that makes creating
html/xml a snap, so I'll just use that when the time comes.

PR #480, opened 2014-05-12
Enable Abbreviation Filter plugin in csledit.xul

fbennett:
This change enables access to the Abbreviation Filter via the csledit.xul editor, making it fully functional in Zotero as well as MLZ.

The plugin allows the user to edit the abbreviations used with a particular style, to import and export abbreviation lists, and it enables some transforms that are needed for legal citation styles. The Filter has been a clunky proof-of-concept thing, but some changes I put up yesterday, described in the [release announcement](http://citationstylist.org/2014/05/11/improvements-to-the-abbreviation-filter-for-multilingual-zotero/), should make it more inviting to people with work to do.

### DIFF
--- a/chrome/content/zotero/tools/csledit.js
+++ b/chrome/content/zotero/tools/csledit.js
@@ -138,6 +138,7 @@ var Zotero_CSL_Editor = new function() {
 			iframe.contentDocument.documentElement.innerHTML = '<div>' + Zotero.getString('styles.editor.warning.parseError') + '</div><div>'+e+'</div>';
 			throw e;
 		}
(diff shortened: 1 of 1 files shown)


fbennett:
Bumping this.

The Abbreviation Filter plugin can be used with Zotero clients in the word processor plugins, but it doesn't work in the CSL Edit pane. It would be nice to enable it there, since the editing, import and export of abbreviation lists is more of a Zotero-side than a word-processor-side activity.

fbennett:
Closing this to spare the clutter.

PR #505, opened 2014-06-22
Linked bibliography titles

fbennett:
citeproc-js now has a variableWrapper() function handle, used to add arbitrary decorations (links, etc.) to citation output. This pull request uses the new functionality to optionally link the title of entries in generated bibliographies to the DOI or the URL of the item (if available).

When the option is enabled from the "Create Bibliography from Items ..." popup, the DOI and URL in rendered entries is suppressed in favour of the link (the idea being that the link serves the same purpose, and omitting it makes the printed cite more readable).

Field decoration of CSL output could be used for various things in various ways; if this patch doesn't meet immediate favour, it may at least help get the creative juices flowing.


fbennett:
Closing this to spare the clutter.

PR #516, opened 2014-07-29
Misc. UI tweaks

aurimasv:
See commit messages.

(Most of these have been annoying me for a while)


zuphilip:
Any news on this? (I am more interested in the functionalities than the looking.)

rmzelle:
@aurimasv, so maybe http://fortawesome.github.io/Font-Awesome/icon/user/ and http://fortawesome.github.io/Font-Awesome/icon/university/ ?

rmzelle:
> I think we just need a better solution for the sync reset pane than bold red text

@dstillman, at least GitHub does something similar:

image

Otherwise, maybe the reset options could be made accessible via a dialog box. The first page of the dialog box could be a warning, and the second page (accessible via a "Next" or "I understand the risks" button) could contain the actual controls.

PR #517, opened 2014-08-01
Export fixes

aurimasv:
See commits, but...
- Fixes BOM on export. There was no way to export with BOM (sort of needed for zotero/translators#742)
- Always show character encoding options in the Export dialog, but collapse them under Advanced Options
- Show the character encoding option that is recommended by the translator (e.g. CSV is better exported as UTF-8 with BOM) in case user switches to another encoding and wants to got back to default.

Here's what this looks like (no, RIS is not normally with BOM):
Recommended character encoding

(having second thoughts on the italics though, maybe it's too much)

Also, why aren't we showing character encoding options for all exports, even if the translator does not specify anything explicitly?


dstillman:
(Sorry, referenced this PR by mistake in that commit.)

Issue #498, opened 2014-06-04
Opening zotero links (zotero://...) also opens an empty window


jlegewie:
Hi,

when I open a zotero link from a note in Standalone such as `<a href="zotero://select/items/0_PZJQAHCA">test</a>` Zotero first opens an empty window and then selects the Zotero item. Is there a way to prevent that?

[Here](https://forums.zotero.org/discussion/37171/opening-zotero-links-zotero-from-notes-also-opens-an-empty-window/#Item_8) is a brief discussion about the issue. fbennett uses a hack in MLZ to circumvent the problem but maybe there is a better solution from the Zotero side.

pashtetGP:
Hello, I set up everything so that now can open zotero://select links from OneNote - the only problem is with empty window. Can please anyone tell where this file chrome/content/zotero/zoteroPane.js
 is located for editing? Win 7, Zotero FF.

pashtetGP:
Hello, I set up everything so that now can open zotero:// links from OneNote - the only problem is with empty window. Can please anyone tell where this file chrome/content/zotero/zoteroPane.js is located for editing? Win 7, Zotero FF.

fbennett:
https://github.com/zotero/zotero/blob/4.0/chrome/content/zotero/zoteroPane.js

pashtetGP:
Thank you for the answer. I mean where it can be found on my local machine to edit? I am not a programmer so sorry if the question is a bit stupid :)

zuphilip:
..\zotero\zotero\chrome\content\zotero\zoteroPane.js

rameloot:
Is there an update on this? Conveniently opening links from Onenote would still be absolutely awesome!

xiaoronglv:
How to open a collection through Zotero link?

"zotero://select/collections/9" doesn't work.

et2010:
I use zotfile to extract annotations to a note and then generate a report from it. I have successfully set Zotero to open the links with [pdf-tools](https://github.com/politza/pdf-tools). I usually save the report as an html file. However, when I try to open zotero links in the saved html file, an empty browser window is opened. Anyone has the same issue? How can I open the link with pdf-tools?

bwiernik:
This issue appears to be fixed on Windows 10 and macOS High Sierra with the current Zotero 5 beta.

PR #534, opened 2014-08-26
Prevent empty window and tab appearance with zotero://select links

mtd91429:
A proposed solution for issue #498 - it prevents a blank window or tab from appearing after a zotero://select or zotero://open-pdf link is opened

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -2787,11 +2787,11 @@ var ZoteroPane = new function()
 			}
 			
 			//Special treatment for zotero links
-			if (uri.match(/(^zotero\:\/\/)((open-pdf)|(select))/)) {
-				var ioservice = Components.classes["@mozilla.org/network/io-service;1"]
+			if (uri.match(/^zotero\:\/\/(open-pdf|select)/)) {
+				let ioservice = Components.classes["@mozilla.org/network/io-service;1"]
 					  .getService(Components.interfaces.nsIIOService);
-				var zoteroLink = ioservice.newURI(uri, null, null);
-				var openZoteroLink = Components.classes["@mozilla.org/network/protocol;1?name=zotero"]
+				let zoteroLink = ioservice.newURI(uri, null, null);
+				openZoteroLink = Components.classes["@mozilla.org/network/protocol;1?name=zotero"]
 						.getService(Components.interfaces.nsIProtocolHandler).newChannel(zoteroLink);
 				return;
 			}
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -2786,6 +2786,16 @@ var ZoteroPane = new function()
 				return;
 			}
 			
+			//Special treatment for zotero links
+			if (uri.match(/(^zotero\:\/\/)((open-pdf)|(select))/)) {
+				var ioservice = Components.classes["@mozilla.org/network/io-service;1"]
+					  .getService(Components.interfaces.nsIIOService);
+				var zoteroLink = ioservice.newURI(uri, null, null);
+				var openZoteroLink = Components.classes["@mozilla.org/network/protocol;1?name=zotero"]
+						.getService(Components.interfaces.nsIProtocolHandler).newChannel(zoteroLink);
+				return;
+			}
+			
 			if (Zotero.isStandalone) {
 				if(uri.match(/^https?/)) {
 					this.launchURL(uri);


chrome/content/zotero/zoteroPane.js
```diff
@@ -2750,6 +2750,16 @@ var ZoteroPane = new function()
 				return;
 			}
 			
+			//Special treatment for zotero links
+			if (uri.match(/^zotero:/)) {
```
simonster: Any reason not to make this `if (uri.match(/(^zotero\:\/\/)((open-pdf)|(select))/)) {`? Otherwise it seems like we have to generate the page twice.

mtd91429:
I removed the redundant regex line and squashed

chrome/content/zotero/zoteroPane.js
```diff
@@ -2786,6 +2786,16 @@ var ZoteroPane = new function()
 				return;
 			}
 			
+			//Special treatment for zotero links
+			if (uri.match(/(^zotero\:\/\/)((open-pdf)|(select))/)) {
+				var ioservice = Components.classes["@mozilla.org/network/io-service;1"]
+					  .getService(Components.interfaces.nsIIOService);
+				var zoteroLink = ioservice.newURI(uri, null, null);
+				var openZoteroLink = Components.classes["@mozilla.org/network/protocol;1?name=zotero"]
```
dstillman: This last line doesn't need to be assigned to anything, and the above assignments would be better as `let`, not `var`.

chrome/content/zotero/zoteroPane.js
```diff
@@ -2786,6 +2786,16 @@ var ZoteroPane = new function()
 				return;
 			}
 			
+			//Special treatment for zotero links
+			if (uri.match(/(^zotero\:\/\/)((open-pdf)|(select))/)) {
+				var ioservice = Components.classes["@mozilla.org/network/io-service;1"]
```
aurimasv: `/^zotero://(open-pdf|select)\//.test(uri)`

mtd91429: I like the reduced parentheses  and implemented them in the latest commit - however, the addition of .test(uri) supersedes the special treatment and continues to open the blank window

chrome/content/zotero/zoteroPane.js
mtd91429: In the latest commit I made these changes.

mtd91429:
I rebased onto the latest commit from the 4.0 branch after implementing the above suggestions. If you want me to squash this into 1 commit, let me know (I left it as 2 for comparison purposes)

chrome/content/zotero/zoteroPane.js
```diff
@@ -2786,6 +2786,16 @@ var ZoteroPane = new function()
 				return;
 			}
 			
+			//Special treatment for zotero links
+			if (uri.match(/^zotero\:\/\/(open-pdf|select)/)) {
```
aurimasv: There's no way that `.test` doesn't work here. Also, you forgot the trailing `\/`

chrome/content/zotero/zoteroPane.js
```diff
@@ -2786,6 +2786,16 @@ var ZoteroPane = new function()
...
+				let ioservice = Components.classes["@mozilla.org/network/io-service;1"]
+					  .getService(Components.interfaces.nsIIOService);
+				let zoteroLink = ioservice.newURI(uri, null, null);
+				openZoteroLink = Components.classes["@mozilla.org/network/protocol;1?name=zotero"]
+						.getService(Components.interfaces.nsIProtocolHandler).newChannel(zoteroLink);
+				return;
```
aurimasv: What does creating an nsIChannel without calling `.open()` do here? This looks like a no-op to me, but I may be missing something.

Also, couldn't all of this `if` block be just
```
Components.classes["@mozilla.org/network/io-service;1"]
  .getService(Components.interfaces.nsIIOService)
  .newChannel(uri, null, null)
  .open()
```
As per https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsIIOService#newChannel()

dstillman: > What does creating an nsIChannel without calling .open() do here? This looks like a no-op to me, but I may be missing something.

The relevant code in zotero-protocol-handler.js that does the selection is in `newChannel`. But yeah, it should use `nsIOService.newChannel`, which takes a spec, instead of `nsIProtocolHandler`:

```javascript
Components.classes["@mozilla.org/network/io-service;1"]
  .getService(Components.interfaces.nsIIOService)
  .newChannel(uri, null, null);
```

aurimasv: OK. I'll have to read through that more carefully. On a quick look, `open-pdf` isn't listed as an extension, nor can I find any reference to that string in Zotero source. Any hint as to where it's handled?

aurimasv: nvm. Found the answer in the original issue:
>Just as a clarification: zotero://open-pdf are the links used by zotfile to open pdfs on certain pages. So it's not an official zotero protocol. Still, I hope they would be part of any solution. :)

dstillman: (Protocol handler stuff is being totally overhauled on the api_syncing branch.)

`open-pdf` is a ZotFile thing. I'm not crazy about its being here, but it doesn't really matter.

PR #537, opened 2014-09-02
[parseContextObject] More author cleanup/checking

aurimasv:


simonster:
Another good candidate for unit tests (now that we have them)

PR #568, opened 2014-10-26
Test target regexp when running web translator tests

### LINKED ISSUE #558: Zotero_TranslatorTester should test target regexp for web translators
Should test it against `document.location.href` of the loaded document somewhere around https://github.com/zotero/zotero/blob/4.0/chrome/content/zotero/tools/testTranslators/translatorTester.js#L403

aurimasv:
Closes #558


aurimasv:
Bump

simonster:
Did you test with the Chrome connector? I'm not sure `webRegexp` will make it here. But if it does I'll merge.

aurimasv:
You were right, `webRegexp` does not make it through to the background page. I worked around it and it seems to work fine now.

PR #576, opened 2014-11-01
[Async DB] Modularize Zotero.DataObject.save()

aurimasv:
At this point, I just mostly want to check if this is an OK direction.

What this allows us to do is, after extending, say, a `Zotero.Item` class, to override the `_saveData` method in order to write additional data before the DB transaction is committed. It also removes a bunch of boilerplate code from `DataObject` child classes (there's probably more that can be removed)


aurimasv:
I think this addresses all of the comments. I'll squash if this is ok.

dstillman:
We can do away with the use of ```this.ObjectsClass``` in item.js, collection.js, etc., and just use ```Zotero.Items``` and the like, right?

aurimasv:
`Zotero.Item`, `Zotero.Collection`, etc. are inheritable, just not the singleton "bucket" objects, since in order to make things work the object cache needs to be the same. Inheriting `Zotero.Collections`, `Zotero.Items` was creating two separate caches that were not staying in sync.

dstillman:
But did you end up actually using this inheritance?

aurimasv:
yes

dstillman:
Ah, OK — so it was just the plural ones that you didn't inherit. All right, this can be squashed (as appropriate — can leave most of the commits separate).

aurimasv:
Squashed.

Additionally, I removed `this.ObjectClass` from the plural singletons, since they're no longer inheritable and addressed your comments about `dateModified`. Please take a look at https://github.com/aurimasv/zotero/commit/6217a5c38eda78fb1e317c495cac2f117f73c23d

aurimasv:
bump

aurimasv:
OK, still re-parsing dateModified to ensure that it's valid, but got rid of `_dateModifiedSet` Changes were amended into the previous commit 84eb26b374c37679f0c43acf01d3708244f6b155

aurimasv:
Ugh, sorry. Had an unnecessary white space change in there. Removed that as well. 6a554ab96e6f35aae950b05845ecc58c7e413394

dstillman:
Looks good. Can squash.

dstillman:
Oh, sorry, or maybe it's squash as appropriate already?

aurimasv:
I squashed as I thought appropriate. Let me know if you want something else.

dstillman:
Nope, this is great. Thanks.

aurimasv:
Excellent! I'll be submitting smaller PRs from now on. It's quite difficult to review so many changes and small issues can set everything back. Hopefully we can get feed reading into the async-db branch this week or so.

PR #578, opened 2014-11-16
Better handling of proxies in translators

aurimasv:
Re https://forums.zotero.org/discussion/42163/remove-proxy-information-from-url-field-on-import/

### DIFF
--- a/chrome/content/zotero/xpcom/proxy.js
+++ b/chrome/content/zotero/xpcom/proxy.js
@@ -331,6 +331,13 @@ Zotero.Proxies = new function() {
 		return (onlyReturnIfProxied ? false : url);
 	}
 	
+	this.getKnownProxy = function(url) {
+		var uri = ioService.newURI(url, null, null);
+		if(Zotero.Proxies.hosts[uri.hostPort] && Zotero.Proxies.hosts[uri.hostPort].proxyID) {
+			return new Zotero.SimpleProxy(Zotero.Proxies.hosts[uri.hostPort]);
+		}
+	}
+	
 	/**
 	 * Determines whether a host is blacklisted, i.e., whether we should refuse to save transparent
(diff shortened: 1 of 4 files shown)


aurimasv:
https://forums.zotero.org/discussion/45778/bug-isi-web-of-knowledge-records-are-not-saved-but-refs-citing-them-are/ presents another issue with proxies (that, I think, this PR resolves). Currently, it seems that proxies can be double-proxied if doPost/doGet/etc. are performed with proxied URIs. I'll fix this in the translator, but this gives me an opportunity to bump the PR. (I'm aware that it needs to be rebased, which I'll do after getting some comments)

aurimasv:
Actually nvm the above. This patch doesn't touch proxies in connector mode (TODO). #627 addresses that issue.

aurimasv:
It seems to me that it would be best to strip proxies for file/link attachments for display in the right-hand pane in Zotero. We would obviously need to keep proxies for actual download during import. After import, proxying of URLs should be handled dynamically by Zotero client. The current model doesn't work well with sharing file/link attachments.

dstillman:
When you say strip for display, you mean strip for storage, right? We shouldn't be storing proxied URLs at all, should we?

aurimasv:
Right. We just can't do it before downloading

aurimasv:
@dstillman, are you generally OK with this PR? If so, I'll get this ready for merging and include proxy stripping for attachments as discussed above.

aurimasv:
@dstillman, bump bump bump

dstillman:
Sorry, not my department.

aurimasv:
@simonster to the rescue!

Demonax:
Any updates on getting this patch merged? I just spent an hour cleaning out proxy strings from my URLs and it'd be nice not to have to do that again. Cheers!

PR #584, opened 2014-11-22
Strip off stray \r when reading lines via Zotero.read()

aurimasv:
Ran into this problem with RIS and APA PsychNET translator, but I worked around it for now.


simonster:
Do we have to strip any `\r`, or just `\r` at the end? Not that I can think of a good reason `\r` in the middle of a line.

aurimasv:
Fairly sure this can only happen at the end of the string (otherwise, `\r` would be an end-of-line character). But I can't really find why this would be happening at all on MXR, tbh. Might be looking in the [wrong place](http://lxr.mozilla.org/mozilla-central/source/netwerk/base/nsReadLine.h#72)

PR #585, opened 2014-11-23
Normalize all user and translator input to NFC

aurimasv:
This includes storing values into DB and performing searches.

Note that export/display of existing data is not normalized. We can try to capture all access points to the database and normalize on output, but that seems like a lot of unnecessary normalization happening all the time. Would probably be best to just normalize existing data.

I'm not sure what this will do to sync if existing data gets normalized when it syncs down. I'm actually not sure what we _want_ it to do. Should it trigger an upload of the normalized data? Should data coming from server not get normalized? (I think the latter sounds like the best option for now)

### DIFF
--- a/chrome/content/zotero/bindings/tagselector.xml
+++ b/chrome/content/zotero/bindings/tagselector.xml
@@ -107,7 +107,9 @@
 				<parameter name="skipRefresh"/>
 				<body>
 				<![CDATA[
-					this._search = val ? val.toLowerCase() : false;
+					this._search = val
+						? val.toLowerCase().normalize()
+						: false;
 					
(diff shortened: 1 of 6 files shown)


drammock:
A quote from the [Unicode standard annex 15](http://unicode.org/reports/tr15/):

> Text exclusively containing ASCII characters (U+0000..U+007F) is left unaffected by all of the Normalization Forms. ... Text exclusively containing Latin-1 characters (U+0000..U+00FF) is left unaffected by NFC. This is effectively the same as saying that all Latin-1 text is already normalized to NFC.

I don't know for sure if there are other ignorable ranges.

PR #591, opened 2014-11-30
Fix the issue that attachment can not be saved when using firefox nightly

solawood:
Firefox introduces a new parameter on nsIWebBrowserPersist.saveURI function.
The attach function is broken due to that change.


dstillman:
Oh, never saw these changes. This looks good. Could you squash down to a single commit?

dstillman:
(While you're at it, there are two extra new lines in utilities.js that can be removed.)

dstillman:
Actually, it looks like you deleted your original repository, so I'll just do this.

dstillman:
Merged as 5f8b56ef86a898caf1449eeefd198ba6e0247d4b

PR #592, opened 2014-12-01
Fix MutationObserver tracking and clearing upon pageModified

aurimasv:
Pre-patch, we were losing references to previously-registered mutation observers when the translate instance was dropped along with the Page object on every pageModified event. This was causing multiple mutation observers to get registered on the same nodes triggering more and more detectWeb calls each time.
We are now careful to disconnect all mutation observers before redoing detection for the whole tab (i.e. all documents in page)
Re https://github.com/proquest/translators/pull/1

### DIFF
--- a/chrome/content/zotero/browser.js
+++ b/chrome/content/zotero/browser.js
@@ -332,7 +332,7 @@ var Zotero_Browser = new function() {
 		}
 		
 		// detect translators
-		tab.detectTranslators(rootDoc, doc);
+		tab.detectTranslators(doc);
 		
 		// register metadata updated event
 		if(isHTML) {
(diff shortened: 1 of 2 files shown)


aurimasv:
Bump. Would like to get this fixed so we don't cause massive browser slowdowns.

aurimasv:
bump

aurimasv:
bumping again, though I cannot rebase and resolve conflicts right now (since I can't currently contribute any new code), but it should not be too hard for someone else to do this. @simonster or @dstillman please review this as is and let me know if this is conceptually acceptable.

Also, a few notes for when this gets merged.
1. Obviously the relevant line in (Summon 2 translator)[https://github.com/zotero/translators/blob/master/Summon%202.js] needs to be uncommented
2. https://github.com/zotero/translators/commit/16ab9ed264dd173d7bd0631de9bfdaac0adda295 should be reverted since the url will now be properly updated as the page updates.

PR #593, opened 2014-12-07
Update cslpreview.xul

rmzelle:
@dstillman, this pull request replaces #7. The changes work in my setup. The main functional change is that the citation-format filter now works for CSL 1.0 styles, since the code hadn't yet been updated to find the "citation-format" attribute on cs:category.

I separated the changes as much as possible for easy review. Let me know if I should squash the commits.

### DIFF
--- a/chrome/content/zotero/tools/cslpreview.xul
+++ b/chrome/content/zotero/tools/cslpreview.xul
@@ -43,26 +43,29 @@
 			this.generateBibliography = generateBibliography;
 
 			function init() { 
-				//refresh(); 
+				//refresh();
+				
+				var iframe = document.getElementById('zotero-csl-preview-box');
+				iframe.contentDocument.documentElement.innerHTML = '<html><head><title></title></head><body><p>Select one or more items in Zotero and click the "Refresh" button to see how these items are rendered by the installed CSL citation styles.</p></body></html>';
 			}
 			function refresh() {
 				var iframe = document.getElementById('zotero-csl-preview-box');
 				var items = Zotero.getActiveZoteroPane().getSelectedItems();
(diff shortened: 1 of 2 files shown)


rmzelle:
> The ibid with some footnote style is not working, e.g. the style for the journal "Deutsche Sprache" (only `<text term="ibid" suffix="."/>` creates an error). Moreover, it is not possible to test footnote style which uses `citation-number`.

@zuphilip, do these things work properly with Zotero itself?

zuphilip:
I think the "error" I described here was actually a mistake in the style which Sebastian corrected in https://github.com/citation-style-language/styles/commit/271be590a6fe4c816e178d866d4aaf673c001f73 . The second sentence is still true: `citation-number` is not rendered and therefore cannot be tested in `csledit`. The Wordplugin for Zotero seems to work as expected.

rmzelle:
@zuphilip, doesn't the screenshot at https://github.com/zotero/zotero/pull/607#issuecomment-70377452 show the `citation-number` being rendered? (the "1" at the start of the bibliographic entry)

adam3smith:
I'm guessing @zuphilip means ```first-note-number```? That may be a bit trickier to implement, but it's a big hassle without any preview.

zuphilip:
Yes, that is kind of what I mean, e.g.
```
          <group delimiter=", ">
            <group delimiter=" ">
              <text macro="author-short"/>
              <group delimiter=" " prefix="(" suffix=")">
                <text value="s. Anm."/>
                <text variable="first-reference-note-number"/>
              </group>
            </group>
            <text variable="locator"/>
          </group>
```
(from https://www.zotero.org/styles/?q=zthk )

PR #594, opened 2014-12-09
Disable ZSA ping from bookmarklet in non-IE browsers

aurimasv:
This means that bookmarklet will only save to server.


simonster:
Just pushed.

PR #601, opened 2014-12-29
Allow meta redirects when running web translator tests

aurimasv:
This is necessary to have functional tests for websites that redirect where a URL-based session ID has expired. E.g. http://www.bdsl-online.de/BDSL-DB/suche/Titelaufnahme.xml?vid={D620881E-7697-40A8-BF2C-4E898FD12362}&erg=0&Anzeige=10&Sprache=de&contenttype=text/html&Skript=titelaufnahme&Publikation_ID=028514998

See zotero/translators#821


simonster:
Thanks!

Issue #509, opened 2014-07-03
Allow right pane to be collapsed


dstillman:
No real reason not to do this, I suppose.

https://forums.zotero.org/discussion/38050

zuphilip:
@hsharrison : Your suggestion is to replace <a href="https://github.com/zotero/zotero/blob/15108eea3f099a5a39a145ed55043d1ed9889e6a/chrome/content/zotero/zoteroPane.xul#L541">zoteroPane.xul, L541-543</a> with
```
<splitter id="zotero-items-splitter" resizebefore="closest" resizeafter="closest" onmousemove="ZoteroPane_Local.updateToolbarPosition()" oncommand="ZoteroPane_Local.updateToolbarPosition()" state="open" collapse="after">
   <grippy id="zotero-items-grippy"></grippy><
/splitter>
```
This works in my first quick test. @dstillman : Does anything speak aggainst that approach?

dstillman:
Yes, that's insufficient, since it doesn't persist across restarts and also causes the toolbar above the right pane to collapse as well.

zuphilip:
Okay, I see. The first problem seems to be solved by adding the attribute `zotero-persist="state"`. Can you confirm that? I will work on the second problem by rewriting the `updateToolbarPosition` function for that.

mtd91429:
The pull request #602 @zuphilip posted does seem to address some of the issues. However, I think that the disappearance of buttons from the `item-toolbar` is a larger issue not limited to the collapse of the `item-pane`.  

For example, on my machine (running windows 7) when I orient the window to 1/2 the screen (window-key + L/R-arrow), the toolbar always overflows the close-button and the spacers in the `collection-toolbar` (red box) and `item-toolbar` (blue box) are not compressed.

![half-screen zotero](https://cloud.githubusercontent.com/assets/3360192/5607486/f73ab8e2-942b-11e4-9741-de3bef98463b.png)


Because of this, I believe it is warranted to re-think how the toolbar is rendered. Perhaps instead of usurping the native rendering engine through a javascript function, we disconnect the absolute width correlation between the toolbars and panes leverage flex boxes to determine the best position of the toolbar components. This would reduce the probability that buttons would be pushed beyond the window's visible frame. 

Additionally/alternatively, we could enable horizontal scrolling when there is overflow. The drawback is that screen real-estate is reduced with a scrollbar.

zuphilip:
[...]

Issue #371, opened 2013-08-14
Fold advanced search into regular search


stakats:
Or at least put the button over there.

PR #602, opened 2015-01-04
Right pane (item-pane) can be collapsed

zuphilip:
This belongs to issue #509 and was asked several times in the forum.

 * Add attribute `collapse="after"` to the splitter and add a grippy element.
 * To remember its state after restart I copied `zotero-persist="state"` which seems to do the job.
 * Update the function `updateToolbarPosition` in zotero/zoteroPane.js:
   * The width of the items-toolbar is corrected if the left pane is collapsed (and all the icons are grouped on the left margin together).
   * If the right pane is collapsed, then the items-toolbar is made flexible (as the item-toolbar is already). As a result the icons from the two toolbars are spreaded evenly over the place.

I tested it in Firefox. Moreover, I installed the zotero-item-history plugin on top of it (this gives new icons to the toolbars). Everything works fine for me. One thing I encountered: if the width of the collections-pane is maximized and the right pane is collapsed, then the icons will fall out on the right. However, I hope that this will not occur often (if ever) in practice.

### DIFF
--- a/chrome/content/zotero-platform/mac/overlay.css
+++ b/chrome/content/zotero-platform/mac/overlay.css
@@ -255,7 +255,7 @@
 	background-color: #8b8b8b !important;
 }
 
-#zotero-collections-splitter[state=collapsed]
+#zotero-collections-splitter[state=collapsed], #zotero-items-splitter[state=collapsed]
 {
 	border-right: 1px;
 	border-color: #A5A5A5;
@@ -267,7 +267,7 @@
 	padding: 0;
 }
 
-#zotero-collections-splitter[state=collapsed] > grippy
+#zotero-collections-splitter[state=collapsed] > grippy, #zotero-items-splitter[state=collapsed] > grippy
 {
 	-moz-appearance: none;
 	background-image: url("chrome://zotero/skin/mac/vgrippy.png");
@@ -297,7 +297,7 @@
 	height: 8px;
 }
 
-#zotero-tags-splitter > grippy:hover, #zotero-collections-splitter > grippy:hover
+#zotero-tags-splitter > grippy:hover, #zotero-collections-splitter > grippy:hover, #zotero-items-splitter  > grippy:hover
 {
 	background-color:transparent;
 }
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -4100,18 +4100,51 @@ var ZoteroPane = new function()
 	 */
 	this.updateToolbarPosition = function() {
 		if(document.getElementById("zotero-pane-stack").hidden) return;
-		const PANES = ["collections", "items"];
-		for each(var paneName in PANES) {
-			var pane = document.getElementById("zotero-"+paneName+"-pane");
-			var splitter = document.getElementById("zotero-"+paneName+"-splitter");
-			var toolbar = document.getElementById("zotero-"+paneName+"-toolbar");
-			
-			var paneComputedStyle = window.getComputedStyle(pane, null);
-			var splitterComputedStyle = window.getComputedStyle(splitter, null);
-			
-			toolbar.style.width = paneComputedStyle.getPropertyValue("width");
-			toolbar.style.marginRight = splitterComputedStyle.getPropertyValue("width");
+		
+		var collectionsPane = document.getElementById("zotero-collections-pane");
+		var collectionsToolbar = document.getElementById("zotero-collections-toolbar");
+		var collectionsSplitter = document.getElementById("zotero-collections-splitter");
+		var itemsPane = document.getElementById("zotero-items-pane");
+		var itemsToolbar = document.getElementById("zotero-items-toolbar");
+		var itemsSplitter = document.getElementById("zotero-items-splitter");
+		var itemPane = document.getElementById("zotero-item-pane");
+		var itemToolbar = document.getElementById("zotero-item-toolbar");
+		
+		var collectionsPaneComputedStyle = window.getComputedStyle(collectionsPane, null);
+		var collectionsSplitterComputedStyle = window.getComputedStyle(collectionsSplitter, null);
+		var itemsPaneComputedStyle = window.getComputedStyle(itemsPane, null);
+		var itemsSplitterComputedStyle = window.getComputedStyle(itemsSplitter, null);
+		var itemPaneComputedStyle = window.getComputedStyle(itemPane, null);
+		
+		var collectionsPaneWidth = collectionsPaneComputedStyle.getPropertyValue("width");
+		var collectionsSplitterWidth = collectionsSplitterComputedStyle.getPropertyValue("width");
+		var itemsPaneWidth = itemsPaneComputedStyle.getPropertyValue("width");
+		var itemsSplitterWidth = itemsSplitterComputedStyle.getPropertyValue("width");
+		var itemPaneWidth = itemPaneComputedStyle.getPropertyValue("width");
+		
+		collectionsToolbar.style.width = collectionsPaneWidth;
+		collectionsToolbar.style.marginRight = collectionsSplitterWidth;
+		itemsToolbar.style.marginRight = itemsSplitterWidth;
+		
+		var itemsToolbarWidthNumber = parseInt(itemsPaneWidth, 10);
+
+		if (collectionsPane.collapsed) {
+			var collectionsToolbarComputedStyle = window.getComputedStyle(collectionsToolbar, null);
+			var collectionsToolbarWidth = collectionsToolbarComputedStyle.getPropertyValue("width");// real width (nonzero) after the new definition
+			itemsToolbarWidthNumber = itemsToolbarWidthNumber-parseInt(collectionsToolbarWidth, 10);
 		}
+
+		if (itemPane.collapsed) {
+		// Then the itemsToolbar and itemToolbar share the same space, and it seems best to use some flex attribute from right (because there might be other icons appearing or vanishing).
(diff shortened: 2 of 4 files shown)


dstillman:
1. The items-toolbar fix looks good.

1. At least on OS X, once the item pane is hidden, it's impossible to get it back without clearing the zotero-persist line from prefs.js. To fix this, the ```#zotero-collections-splitter``` rules in chrome/content/zotero-platform/mac/overlay.css need to apply to ```#zotero-items-splitter``` as well, to force a width on the collapsed splitter. (I can confirm that it's working on OS X if necessary.)

1. I think this behavior is problematic with the item pane min-width set so wide, since the pane disappears all of sudden while you're just adjusting the width. With the collections pane, it doesn't collapse until you drag pretty far past a reasonable point. On OS X with an English locale, a min-width of 250px for ```#zotero-pane``` works pretty well, collapsing shortly before the tabs run out of room. I haven't tried with other locales. We can address this in a separate issue if you want, since it may take a little work to find the right width (or figure it out programmatically).

1. I don't think centering the search box and Locate menu is the right approach, and with a smaller min-width they actually jump back to the left when the pane collapses. For the collections pane, the middle-pane buttons just collapse against the collections toolbar's buttons, and I think we want to do something similar here (though maybe with a little fixed space between the Locate menu and the sync button — not sure).

zuphilip:
Please have a look at the new version, which
 * Handle zotero-items-splitter the same as zotero-collections-splitter in zotero-platform/mac/overlay.css (I haven't found any other occurencies of "#zotero-collections-splitter" in the repository)
 * Change min-width to 250px for #zotero-item-pane (this works for me on Win 7, Firefox with German locale; moreover I think it is better than the current value but maybe it can be further optimized in another issue)
 * Change the behaviour of updateToolbarPosition for collapsed right pane: now the icons from the middle are rearranged at the left (please note I tried here several other possibilities, but they did not seem to be stable for all cases)

Can you make your test for Mac OS X?

aurimasv:
I don't particularly like the jumpy search bar. Collapsing the right pane should move the search bar to the right, not group it together with the other buttons above the item tree pane. The code also looks quite... inflated, but I'll comment on that a bit later.

dstillman:
> Collapsing the right pane should move the search bar to the right, not group it together with the other buttons above the item tree pane.

That was my initial thought too (and what I requested above), but after seeing this version I decided it actually did make sense to have the search bar grouped with the Advanced Search icon and the Locate button placed where it seems more contextual to the items list, which isn't the case over on the right. (You can actually recreate the post-collapse configuration currently just by widening the item pane. You can't recreate a version with the search bar and Locate flush against the sync button.) And the search bar will jump no matter what, since we can't make the min-width small enough to avoid that — it's just a question of whether it will jump to the right or back to the left.

aurimasv:
> it actually did make sense to have the search bar grouped with the Advanced Search icon

Yes, that makes sense in general (not just after collapse) and we do have a ticket for it (#371)

>the Locate button placed where it seems more contextual to the items list, which isn't the case over on the right

I don't think I see much distinction given that the items list is the only thing present. It's also not really applicable to the list per se, Only individual items. I was initially going to suggest hiding it altogether, but  it does make sense to allow using it without seeing item metadata. We should disable all options (except for Manage Lookup Engines... when multiples are selected), but that's for a different issue. In general, the button is very obscure and I don't think most people know what it does. In the long run, I'd like to revise it into a tab containing more integration with the web (altmetrics, pubmed commons, lookup engines, etc.).

If the only reason to move the search bar is to relocate the Locate button, then we can certainly just relocate the Locate button.

>You can actually recreate the post-collapse configuration currently just by widening the item pane

Not really sure what you mean. Do you mean just have it next to the advanced search icon? I'm not against actually placing the search box on the left (though I do think its current location is good), it's the unnecessary jumping from left to right that I don't like.

>And the search bar will jump no matter what, since we can't make the min-width small enough to avoid that — it's just a question of whether it will jump to the right or back to the left.

Right, that's fine. It will jump along with the divider, not in the opposite direction.

dstillman:
> It's also not really applicable to the list per se, Only individual items.

Well, that's the case with other middle-pane buttons (note and attachment) already. The Sync button, by contrast, seems rather different. I'm happy to see this with the elements moved over to the right, but I'm not sure how well it'll work. At the very least, I think it would require 1) moving the Advanced Search icon over next to the search and removing the dotted divider (at all times) and having those two elements both move to the right when collapsed and maybe 2) removing the divider to the right of Sync (in Firefox) and showing a divider to the left of Sync when collapsed so Locate and Sync are separated. Less sure about 2, but Locate and Sync just seem weird next to each other to me.

One problem with the above is that I'm not crazy about having two magnifying glasses next to each other, which would be the case at least on OS X.

The jumping to the left is a little unpleasant, but it avoids all of these problems, and I don't want to ask @zuphilip to start dealing with lots changes to the permanent layout. I care more about a logical configuration post-collapse than a little weirdness while switching.

> If the only reason to move the search bar is to relocate the Locate button, then we can certainly just relocate the Locate button.

Relocate it where? (I agree there are long-term possibilities, but not sure what you mean here.)

dstillman:
> Not really sure what you mean. Do you mean just have it next to the advanced search icon?

Yes. I mean that the layout you get now post-collapse, with the search bar and Locate moved over to the left next to Advanced Search, is a layout that can already happen. Having the search bar and locate grouped with Sync and far away from the Advanced Search button isn't something that's possible currently, and I think it makes less sense.

zuphilip:
<blockquote>removing the dotted divider (at all times)</blockquote>
This should be trivial and it sould avoid the jumping of the search box. Should I try that?

dstillman:
> This should be trivial and it sould avoid the jumping of the search box.

How would that avoid the jumping of the search box? I clarified which divider I meant above, in case it was unclear.

dstillman:
I meant this, if it wasn't clear:

toolbar

zuphilip:
Sorry, my fault. The other direction (to the left) would be trivial.

aurimasv:
>I think it would require 1) moving the Advanced Search icon over next to the search (at all times), removing the dotted divider to the Advanced Seach icon's left (at all times), and having those two elements both move to the right when collapsed and maybe 2) removing the divider to the right of Sync (in Firefox) and showing a divider to the left of Sync when collapsed so Locate and Sync are separated. Less sure about 2, but Locate and Sync just seem weird next to each other to me.

Sounds right to me.

>One problem with the above is that I'm not crazy about having two magnifying glasses next to each other, which would be the case at least on OS X.

That's not the case on Windows (I guess you know that), but yes, it doesn't look great on Mac. What if we move the advanced search button to the drop-down menu for regular search (I think that's what #371 is about) at all times? It's actually a rather intuitive place IMO.

>I care more about a logical configuration post-collapse than a little weirdness while switching.

Switching is the most annoying, yes, but the search bar changing positions (right to left) is also annoying.

>the layout you get now post-collapse, with the search bar and Locate moved over to the left next to Advanced Search, is a layout that can already happen. Having the search bar and locate grouped with Sync and far away from the Advanced Search button isn't something that's possible currently, and I think it makes less sense.

While it _can_ happen, I don't think it's that common. You'd have to get your item list pretty narrow to actually have it in that configuration.
narrow item list
OTOH, on a 23" monitor (which is pretty typical these days), the search box is way far from the other buttons and it's quite obvious when it is changed to the other side. (the collection list is cut off, but it's at a comfortable width)
23" right
23" left

I can add a patch for the above changes, we don't have to bother @zuphilip

dstillman:
> What if we move the advanced search button to the drop-down menu for regular search (I think that's what #371 is about) at all times? It's actually a rather intuitive place IMO.

I'm actually not sure what #371 is about. I thought about moving Advanced Search into the search drop-down, but I think that would drastically reduce the number of people who discovered advanced and saved searches, and the latter in particular are one of the best features in Zotero. I'd sooner drop the magnifying glass in the search bar. Haven't looked into how feasible that is.

In any case, @zuphilip, could you do a version with the elements moving to the right? If that ends up seeming less jarring, we can deal with the other necessary changes for that separately.

zuphilip:
I implemented a new version where the search box is moved to the right and still we will have a flex element at the right (which I think is essential for a stable version).

dstillman:
OK, yeah, that's a whole lot better. You can squash this down, and we'll deal with other stuff separately.

zuphilip:
Squashed

aurimasv:
Yup, that looks good. Thanks @zuphilip!

As Dan said, we'll probably need to add some separators for locate and sync icons. I'm also seeing overlap of those icons when collapsed, but we'll take care of it after this gets merged.

dstillman:
Thanks!

aurimasv:
Dan, do you have any idea when next Zotero release is coming out?

dstillman:
We'll need to do a compatibility fix for Firefox 35 by next Tuesday, but we might just backport the necessary changes onto 4.0.24.1. Otherwise no immediate plans.

zuphilip:
Yeah, my first contribution to zotero itself! Thank you very much for the fast review.

<blockquote> I'm also seeing overlap of those icons when collapsed, but we'll take care of it after this gets merged.</blockquote>

You are right. The sync icon is taking more space than other icons and it looks like this is a consequence of that. 

BTW while you are looking at spacing anyways: I wasn't changing the two margins in the toolbars, i.e. in collapsed mode there is an additional space (10px) between the group icon and preferences and an additional space (10px) between the search box and the locate icon. These spaces are coming from the margin which is needed for the width of the splitter in normal mode. These spaces are looking okay for me even in collapsed mode, but maybe you want to change the behaviour now.

PR #603, opened 2015-01-10
Use nav-bar as id of menu button container

fbennett:
Targeting addon-bar in MLZ, the icon refused to install on first run.
Most notes I found (apart from some possibly dated documents on MDN)
referred to nav-bar as the target for insertItem(). Changing the code
to this id works reliably in both environments, and may be more
stable in future iterations.

(Or I may be misunderstanding how this works.)


dstillman:
This seems right, but we should update the variable names too, since the add-on bar doesn't exist anymore.

dstillman:
(And you can just ```commit --amend```. I don't need to see a separate commit.)

fbennett:
There seems to be some code in [overlay.js](https://github.com/zotero/zotero/blob/4.0/chrome/content/zotero/overlay.js#L198) and in [zotero.css](https://github.com/zotero/zotero/blob/4.0/chrome/skin/default/zotero/zotero.css#L58) that is specifically aimed at the addon-bar (and not touched by the patch). Should that be removed?

dstillman:
Yes, and actually some other stuff too (the statusBarIcon pref, various image files), but I can take care of that stuff separately if you want.

fbennett:
I'll take a look tomorrow. You guys have a lot on your plate as it is.

dstillman:
No, that's not right. You'll have to rebase this against the current 4.0 head. Otherwise it'll wipe out all of those other things.

dstillman:
(And sorry, I didn't get a notification for your last comment for some reason.)

fbennett:
I've merged the latest to the branch. Not sure whether I should rebase (f) against the "Merge from 4.0" commit ... that won't mess things up in between? Sorry to be clueless but, well, I'm learning.

dstillman:
Yeah, you want to rebase, not merge. Assuming you have the main Zotero repo set up as a remote named ```upstream```, you can just do ```git rebase -i upstream/4.0 icon-to-nav-bar```, and then follow the instructions, which in the case of a single commit like this probably just means saving and quitting without editing anything. (You can technically run it without the ```-i```, but interactive mode lets you see what's happening and bail when things are looking dicey.)

dstillman:
And if you're in the branch already, you can leave out the ```icon-to-nav-bar``` part.

Once you've rebased, do a ```git status```, ```git log```, and/or ```git show``` to make sure everything looks OK, and then ```git push --force```. You need the ```--force``` because you're actually rewriting remote history. That's usually a bad thing to do but is good and proper in the case of pull requests.

fbennett:
Maybe something like that ... ?

fbennett:
Um ... no. Conflicts.

I'll try again later.

dstillman:
Sorry, forget to mention that you have to do a ```git fetch upsteam 4.0``` first.

fbennett:
I get the same result. Here are the steps I followed:

    git checkout origin 4.0
    git branch -D icon-to-nav-bar
    git fetch upstream 4.0
    git fetch origin icon-to-nav-bar:icon-to-nav-bar
    git checkout icon-to-nav-bar
    git rebase -i upstream/4.0

Rebase interactive shows a long line of commits, with the icon-to-nav-bar change at the bottom of the list. I save without altering the list.

    git push --force origin icon-to-nav-bar

This reports:

    Everything up-to-date

... which yields the unmergeable pull request with many individual commits and blocking conflicts.

Shall I just dump the changes as a patch and start over on a fresh branch, working the code in by hand as required?

(It also occurs to me that it might be better to try this after things have settled down a bit - I feel I'm just causing a distraction from more important work.)

dstillman:
Yeah, actually I guess the main commit itself is messed up too. Not sure how it got into that state, but rebasing wouldn't have worked — you have to actually undo those changes.

But you can still keep this pull request. In this branch, you can do ```git reset --hard 9232dd52``` to set it to the current top of 4.0. Then manually apply the changes again, using the last commit here as a guide. (That'll be easier than applying it in full with all those other erroneous changes and undoing them.) Then commit that and force-push.

fbennett:
Maybe something like this.

fbennett:
Oh. still conflicts. I'm at a loss; the patch was applied manually against the 4.0 head after a fetch.

I'm obviously just not understanding how git works at this end. I'll try afresh again another day.

dstillman:
No, that's on top of a commit from last July. Did you do ```git reset --hard 9232dd52``` in the branch and then apply the changes manually (by which I mean, by hand in a text editor, not via any commands)? That really should work.

You can do it via ```git rebase```, of course, but it's a bit more complicated.

Also, when I said there were erroneous parts in that commit, I may have just not been looking closely. You never need to or should make any locale changes outside of en-US, so all those locale changes can be left out (and in this case they'd likely cause conflicts).

fbennett:
Well, it looks like the garbage has gone away, anyway. Haven't tried running this with the latest code. Might work, might not, but it's bedtime.

fbennett:
Removed an unneeded code block that snuck through in the manual patch. Installed the client in a profile, the icon installed itself. Removed the icon with Customize, removal persisted on restart. Restored, and restore persisted on restart.

dstillman:
Looks good. Thanks!

fbennett:
Meanwhile, I think I may have figured out what I was doing wrong with the branches. I had 4.0 set as a remote tracking branch with both pull and push set to git@github.com:zotero/zotero.git. Push is apparently needed to tell the local git that local records are in sync with the remote (?), so I kept getting "Your branch is ahead of ..." warnings when I checked it out. This seems to have fixed it:

    git remote set-url --push upstream git@github.com:fbennett/zotero.git

A "git push upstream 4.0" now apparently updates some pointer in the local revision history for the branch; the warning has gone away, and the local log shows up-to-date records.

dstillman:
Convention here is to set your own fork as ```origin``` and the main Zotero repo as ```upstream```. My advice above was assuming that was the case, and that would explain why you were still on July 2014 after doing a hard reset to the tip of ```upstream/4.0``` after fetching upstream 4.0, and why various other things went so wrong.

fbennett:
That is how it's been laid out from the start. I'm probably still not quite understanding how branches relate to remotes, which seems to be where things went awry. Thanks for your patience, next time around I'll be more careful.

Issue #39, opened 2012-01-31
Text transform on authors


dstillman:
http://forums.zotero.org/discussion/835/change-case-automatically/#Item_7
http://forums.zotero.org/discussion/1991/names-variant-spellings-entities/#Item_3
http://forums.zotero.org/discussion/6418/list-of-feature-requests/#Item_6
http://forums.zotero.org/discussion/12638#Item_5
http://forums.zotero.org/discussion/15434/right-mouse-click-transform-text-enabled-for-author-fields/
http://forums.zotero.org/discussion/20842/
http://forums.zotero.org/discussion/21613/transform-text-for-author-field/

https://www.zotero.org/trac/ticket/1307

Maybe we should do this.

AbeJellinek:
https://forums.zotero.org/discussion/98444/feature-request-title-case-for-author-name

Issue #511, opened 2014-07-16
Parse commas in pasted creators if field is in two-field mode


dstillman:
Currently only tabs, which are unambiguous, work as delimiters, but no one is going to guess that. This should only look for a comma if there are no tabs in the input.

https://forums.zotero.org/discussion/38430/

zuphilip:
Let me summarize the feature again: If creator field is in two-field mode and the pasted text contains no tab but it contains a comma, then we want to to split on the first comma for the lastName, firstName. I guess one would have to work on <a href="https://github.com/zotero/zotero/blob/4.0/chrome/content/zotero/bindings/itembox.xml#L1742">itembox.xml</a> for that. Maybe I give it a try some days later...

PR #605, opened 2015-01-10
Text transform in context menue on creators

zuphilip:
A patch for #39 in a simple way:
- only for creatorMode = 0 (lastName, firstName)
- only title case
- only per author field (no batch edit)


zuphilip:
Because it is the same file and also about autors, I add another commit:

A patch for #511 in a simple way:
 * only if several creators are entered
 * only if no tabs
 * only if already in fieldMode = 0 (lastName, firstName)

zuphilip:
ping @dstillman

chrome/content/zotero/bindings/itembox.xml
```diff
@@ -1975,8 +1978,19 @@
 						default:
 							throw ("Invalid transform mode '" + mode + "' in zoteroitembox.textTransform()");
 					}
-					this._setFieldValue(label, newVal);
-					this._modifyField(label.getAttribute('fieldname'), newVal, this.saveOnEdit);
+					
+					var currentFieldName = label.getAttribute('fieldname');
+					if (!currentFieldName.contains('-')) {// single field, e.g. title
```
dstillman: ```String.prototype.contains``` isn't actually valid after Firefox 36 — it's [switching to String.prototype.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes). Have to just use ```indexOf``` for the time being.

chrome/content/zotero/bindings/itembox.xml
```diff
@@ -1975,8 +1978,19 @@
...
+						this._setFieldValue(label, newVal);
+						this._modifyField(currentFieldName, newVal, this.saveOnEdit);
+					} else {// creator field, e.g. creator-2-lastName
```
dstillman: Better yet, move this block up, use ```currentFieldName.startsWith('creator-')```, and have all other fields be ```else```.

dstillman: Also, style nit: comments should be on separate lines above the relevant lines

chrome/content/zotero/bindings/itembox.xml
```diff
@@ -2476,6 +2490,8 @@
 							return !hideTransforms;">
 						<menuitem label="&zotero.item.creatorTransform.nameSwap;"
 							oncommand="document.getBindingParent(this).swapNames();"/>
+						<menuitem label="&zotero.item.textTransform;"
+							oncommand="document.getBindingParent(this).textTransform(document.popupNode, 'title')"/>
```
dstillman: "Text Transform" isn't clear here. Let's go with a new string. (And can be title-cased. We can fix "Swap first/last names", which is sentence case for some reason.)

zuphilip:
I will look at your other comments later on. Just one question:

<blockquote>Re: the commit message, say "Closes #" plus the referenced ticket number so that it closes that ticket when merged. Also, you meant "creatorMode = 0", not "creatorMode = 1".</blockquote>
When merging several commits in the end, I can change that message again. Should I delete the "Closes #"? IMO it will close that issue...

dstillman:
Sorry, I'm not clear what you're asking here.

zuphilip:
How should I change the commit message (in the end)?

dstillman:
I explained above...

https://help.github.com/articles/closing-issues-via-commit-messages/

zuphilip:
Okay, I thought I had that statement already in the commit message, but that is not the case. That was the confusion. I will do it then.

chrome/content/zotero/bindings/itembox.xml
zuphilip: Any suggestions how to name it?
 * "Title Case" (which we already have...)
 * "Fix Cases"
 * "Fix Caps"
 * ...

dstillman: Well, we could actually have the "Transform Text" menu and "Title Case" within it, but if that's always going to be the only option maybe just "Convert to Title Case"? "Title Case" could technically be seen as an imperative verb, but it seems a little more awkward to have just that.

zuphilip:
Uff.. that was wrong rebasing aggainst `master`. Let me try again against `4.0`....

zuphilip:
ping @dstillman

dstillman:
Looks like right now the text transform only converts one name field at a time. Does it make more sense to convert both? Probably more likely that both names will be messed up, right?

Also, just an FYI, 1) GitHub doesn't send out notifications for new commits on pull requests, so be sure to someone at the time if you want someone to look at it, and 2) it's better not to squash until you get the final go-ahead, because it's easier for people to re-review without it.

zuphilip:
<blockquote>Looks like right now the text transform only converts one name field at a time. Does it make more sense to convert both? Probably more likely that both names will be messed up, right</blockquote>

I don't know if this is more likely. But there are names where you don't want to have title case in both parts, e.g. "Jean de La Fontaine", "Karl-Theodor zu Guttenberg". I think it is more consistent to just change the field where the menue appears. Moreover, if there are 10 authors present, then it will only change the cases in the author (and field) where it is active. What do you think?

dstillman:
Well, but what is this actually correcting? Is it not largely for cases where the names come in all-caps, in which case even if you want "Jean de La Fontaine" it's still quicker to transform and then correct? (Or do we fix those automatically? I know we had some old capitalization-fixing behavior, but I'm not sure of its current state.)

zuphilip:
Okay, I can change the behaviour that always last name AND first name is transformed (possibly one has to fix afterwards manually but this is negligable).

Would you suggest to add the option for fieldMode = 1 as well?

I can also think of some actions for all creators but then we should rename it accordingly or maybe even create two options each:
 * Swap First/Last Name
 * Swap First/Last Name for all creators
 * Convert to Title Case
 * Convert to Title Case for all creators

dstillman:
Well, I still want to know what this is meant to correct, though (ideally without rereading all the threads linked from https://github.com/zotero/zotero/issues/39). If this is mostly meant to fix all-caps imports, shouldn't we just detect those and convert them to title case automatically? Would all-caps titles ever be desired? What other cases is this addressing?

zuphilip:
It is IMO mainly about all-caps.

Is it possible that a name for a person has to be written in all-caps? I don't think so, but there are a lot of languages and national rules. It is possible that one only have the initials of a first name, e.g. "H. G. Wells", but then all-caps is the same as title case.

However, there are organizations with all-caps names, e.g. "US", "UNO", "NIH", "BASF". These should certainly not being transformed to title case automatically. I think I have also seen organization with a comma in their name and these are sometimes wrongly imported as persons (fieldMode=0).

chrome/content/zotero/bindings/itembox.xml
```diff
@@ -1975,8 +1976,22 @@
...
+					if (currentFieldName.startsWith('creator-')) {
+						let row = Zotero.getAncestorByTagName(document.popupNode, 'row');
+						let creatorIndex = parseInt(currentFieldName.split('-')[1]);
+						let fields = this.getCreatorFields(row);
+						fields[currentFieldName.split('-')[2]] = newVal;
+						this.modifyCreator(creatorIndex, fields);
+						this.item.save();
+					}
+					// Single field (e.g., title)
+					else {
```
zuphilip: This is AFAIK <a href="http://en.wikipedia.org/wiki/Indent_style#Variant:_Stroustrup">Variant Stroustrup</a> and not <a href="http://en.wikipedia.org/wiki/Indent_style#Variant:_1TBS">1TBS</a> as wished in https://www.zotero.org/support/dev/client_coding/coding_guidelines . Anyway, if you want it differently, feel free to change it afterwards.

zuphilip:
Okay, I reread all the discussion. It was always about call-caps. Some discussions are about in which phase to correct it: web translator, generic translator like for RIS, automatically, manually. There seems to be an improvement in the last years in the web translators, that most cases for all-caps are corrected in this step. But this leaves for example imports from RIS untouched. This might be okay or even wanted, see https://forums.zotero.org/discussion/20842/#Item_7 . To make such an transformation automatically at every import, seems to be even more aggressive. On the other side, if we can trust that most cases are already handled with the web translator, this leaves maybe just some cases left. For these cases could the menue be useful. @dstillman What do you think? I updated the code as well, please have a look at the new version.

zuphilip:
Newly reported: https://forums.zotero.org/discussion/49833/transform-text-sentense-case-for-authors/

PR #607, opened 2015-01-13
Localize Style Editor and Preview and other improvements

rmzelle:
Continues work done in https://github.com/zotero/zotero/pull/593.

@dstillman, this is the first time I localize stuff in XUL, so please check I did things correctly. If you don't have complaints I can localize the Style Editor as well.

Some questions:
* Zotero.getString() only works for pulling strings from "zotero.properties"?
* Do you still prefer to keep the JavaScript code embedded in the XUL files (cslpreview.xul and csledit.xul)? I'm not a big fan since my text editor doesn't recognize it as code, so I'd rather put it in a separate ".js" file.


chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -968,3 +968,6 @@ firstRunGuidance.quickFormat	= Type a title or author to search for a reference.
 firstRunGuidance.quickFormatMac	= Type a title or author to search for a reference.\n\nAfter you've made your selection, click the bubble or press Cmd-\u2193 to add page numbers, prefixes, or suffixes. You can also include a page number along with your search terms to add it directly.\n\nYou can edit citations directly in the word processor document.
 firstRunGuidance.toolbarButton.new = Click here to open Zotero, or use the %S keyboard shortcut.
 firstRunGuidance.toolbarButton.upgrade = The Zotero icon can now be found in the Firefox toolbar. Click the icon to open Zotero, or use the %S keyboard shortcut.
+
+cslpreview.instruction		= Select one or more items in Zotero and click the "Refresh" button to see how these items are rendered by the installed CSL citation styles.
+cslpreview.warning.no-items		= No items selected in Zotero.
```
dstillman: cslpreview.warning.noItems

chrome/locale/en-US/zotero/cslpreview.dtd
```diff
@@ -0,0 +1,9 @@
+<!ENTITY zotero.cslpreview.refresh		"Refresh">
+
+<!ENTITY zotero.cslpreview.citation-format		"Citation Format:">
+<!ENTITY zotero.cslpreview.citation-format.all		"all">
+<!ENTITY zotero.cslpreview.citation-format.author		"author">
+<!ENTITY zotero.cslpreview.citation-format.author-date		"author-date">
+<!ENTITY zotero.cslpreview.citation-format.label		"label">
+<!ENTITY zotero.cslpreview.citation-format.note		"note">
+<!ENTITY zotero.cslpreview.citation-format.numeric		"numeric">
```
dstillman: These should be camel case

dstillman:
> Zotero.getString() only works for pulling strings from "zotero.properties"?

Yes.

> Do you still prefer to keep the JavaScript code embedded in the XUL files (cslpreview.xul and csledit.xul)? […] I'd rather put it in a separate ".js" file.

Moving to separate files is fine.

rmzelle:
Could I get people's opinion about reorganizing the top bar of csledit.xul? I like the idea of using `<groupbox/>` to add explanatory labels to the different options, but it comes at the cost of vertical space, which can already be in short supply. Worthwhile trade-off?

We probably also need some CSS to make it look nice. It currently looks like:

screen shot 2015-01-14 at 23 13 54

fbennett:
The editor is often used for preview, without any intention to edit the
CSL. Maybe a hide/reveal button for the CSL block would be good, with
state-persistence.

That would buy you quite a bit of vertical space.

zuphilip:
* The first time(s) I was using this `csledit` I was afraid to destroy some of the existing styles. Maybe one could add more explanation about "Save as..." in presence with the loades style.
 * I think the "year only" and "suppress author" are two different function, where we only have the second. There migth be other metadata in an citation for example the title can be a substitute for the names-node.
 * I think the style does not have to be installed to render in csledit. Thus, I suggest to simply use a label "Load style".
 * The refresh button is the one, I am using most often. Should we highlight that button more? Maybe add an icon?
 * The order of the buttons and boxes seem for me a little random. From my experience I would put the "load style" and "save as..." together (in other software such functions are often in the file menu). The "cite position", "suppress author" and "page locator" are configuring the citation. There is no configuration for the list of reference. Finally, there is the "refresh" button (maybe align on the right which would be more consistent with firefox refresh button?).
 * Did you try out to align the code pane and csl pane vertically besides each other?
 * I would avoid another button for hiding/revealing the csl block. However, one can make the splitter (?) clickable in the same way as in zotero itself.

fbennett:
Splitter or button, I would prefer if the hidden/revealed state would
persist (i.e. be saved in a pref).

rmzelle:
> a hide/reveal button for the CSL block would be good, with
state-persistence. ... I would prefer if the hidden/revealed state would
persist (i.e. be saved in a pref).

We can just use the same setup used to hide the tag selector. Will put it on the to do list.

> The first time(s) I was using this csledit I was afraid to destroy some of the existing styles. Maybe one could add more explanation about "Save as..." in presence with the loades style.

We can use the output window to show some instructions (and e.g. include a link to the online docs), as I already did for cslpreview.xul.

> I think the "year only" and "suppress author" are two different function, where we only have the second.

citeproc-js supports "suppress-author" and "author-only" (see http://gsl-nagoya-u.net/http/pub/citeproc-doc.html). We can use a different label for the box, or none at all.

> I think the style does not have to be installed to render in csledit. 

But only installed styles show up in the dropdown menu.

> Did you try out to align the code pane and csl pane vertically besides each other?

I'm not a fan of vertically split windows.

zuphilip:
This is a picture of how I worked today with `csledit`:

csledit

It seems for me that the space is not optimally divided. Anything which improves it, is fine for me. A vertical split was my first thought, but maybe there other ideas?

rmzelle:
* the code window can now be collapsed. State does not persist right now, but there is code in zoteroPane.js we might be able to use ( https://github.com/zotero/zotero/blob/6de3fa4d47749d3a46eb46d1541a725f7ae15ae8/chrome/content/zotero/zoteroPane.js#L4054 ).
* I made the headers in the render window smaller to save some space. There is still a lot of space around the bibliography output (see below). @fbennett, do you know what's up with that? Using `white-space="pre-line"` instead of "pre-wrap" solves the problem, but would collapse spaces and tabs (see https://developer.mozilla.org/en-US/docs/Web/CSS/white-space). But citeproc-js already does some of that itself, right?

screen shot 2015-01-17 at 17 06 05

fbennett:
It looks like the extra vertical space is due to the hard returns in the HTML. If hack it with something like `str.replace(/>\s+</, "")` before adding it to elem.innerHTML, it may come right.

fbennett:
On the other hand, can't you just eliminate the pre-* and let the content render as normal HTML?

rmzelle:
That would solve it as well, which would probably be fine as long as citeproc-js keeps suppressing repeated spaces.

fbennett:
We'll keep suppressing spaces ... but HTML rendering will do that anyway, won't it?

rmzelle:
@dstillman, this might be a good point to merge this pull request. The commits are descriptive, but in short: the Style Preview and Style Editor have been renamed and localized, the JS code has been moved to their own files, and the editor window of the Style Editor can now be collapsed.

I can add in the groupboxes if you like the mockup at https://github.com/zotero/zotero/pull/607#issuecomment-70037982.

chrome/content/zotero/tools/csledit.xul
```diff
@@ -26,241 +26,44 @@
...
+				</menupopup>
+			</menulist>
+			<menulist id="zotero-csl-list" style="min-height: 1.6em; min-width: 100px" initialized="false" flex="1" oncommand="Zotero_CSL_Editor.loadCSL(this.selectedItem.value)"/>
 		</hbox>
 		<textbox id="zotero-csl-editor" type="timed" timeout="250" multiline="true"
 			flex="1"
 			onkeypress="Zotero_CSL_Editor.handleKeyPress(event)"
 			oncommand="document.getElementById('zotero-csl-list').selectedIndex = -1; Zotero_CSL_Editor.generateBibliography(this.value)"/>
-		<splitter/>
+		<splitter id="csledit-splitter" collapse="before" persist="state">
```
zuphilip: Any reason you take here the attribute `persist` and not `zotero-persist`? But maybe I don't really understand how this is really working in zotero as well...

rmzelle: I'm not sure about the difference. "persist" was easy to implement. Maybe the benefit of "zotero-persist" is that this state syncs? @dstillman would know.

aurimasv: IIRC zotero-persist is necessary for switching between Zotero embedded and tab mode in Firefox.

fbennett: I think zotero-persist preserves the state when the browser is stopped and restarted, by saving the value to [a preference](https://github.com/zotero/zotero/blob/4.0/defaults/preferences/zotero.js#L165).

rmzelle: Restarting the browser saved state for me with "persist" (see also https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/Attribute/persist).

aurimasv: See my comment above (persist is only saved when closing window, so it doesn't get saved when switching tabbed <-> embedded mode), but I don't think it's relevant to the style editor.

aurimasv: FYI, here's the relevant commit/comment from @simonster https://github.com/zotero/zotero/commit/47f57509e53216a1d2fc110f3c7c7dfb37135bec

gracile-fr:
I forgot to reply to #593 but it would be great to add line numbers to csledit. Very useful when the validator gives you the faulty line.

chrome/content/zotero/tools/csledit.xul
zuphilip: Thank you for the information, @aurimasv , very useful.

rmzelle:
@dstillman, can you please review this, so I don't have to rebase this later? I'll squash the commits and update the pull request title once you've taken a look. It's mostly small improvements, and nothing that should be controversial.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -968,3 +968,15 @@ firstRunGuidance.quickFormat	= Type a title or author to search for a reference.
 firstRunGuidance.quickFormatMac	= Type a title or author to search for a reference.\n\nAfter you've made your selection, click the bubble or press Cmd-\u2193 to add page numbers, prefixes, or suffixes. You can also include a page number along with your search terms to add it directly.\n\nYou can edit citations directly in the word processor document.
 firstRunGuidance.toolbarButton.new = Click here to open Zotero, or use the %S keyboard shortcut.
 firstRunGuidance.toolbarButton.upgrade = The Zotero icon can now be found in the Firefox toolbar. Click the icon to open Zotero, or use the %S keyboard shortcut.
+
+cslpreview.instruction		= Select one or more items in Zotero and click the "Refresh" button to see how these items are rendered by the installed CSL citation styles.
+cslpreview.warning.noItems		= No items selected in Zotero.
+
+csledit.saveAs.title		= Save Citation Style
+csledit.saveAs.CSL		= Citation Style Language
+csledit.warning.noItems		= No items selected in Zotero.
```
dstillman: We should use a common string for this across the two panes so localizers don't have to localize it twice.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -968,3 +968,15 @@ firstRunGuidance.quickFormat	= Type a title or author to search for a reference.
...
+csledit.warning.parseError		= Error parsing style:
+csledit.output.individualCitations		= Individual Citations
+csledit.output.singleCitation		= Single Citation (with position "first")
+csledit.output.bibliography		= Bibliography
```
dstillman: This should be moved to ```general.bibliography```.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -968,3 +968,15 @@ firstRunGuidance.quickFormat	= Type a title or author to search for a reference.
 firstRunGuidance.quickFormatMac	= Type a title or author to search for a reference.\n\nAfter you've made your selection, click the bubble or press Cmd-\u2193 to add page numbers, prefixes, or suffixes. You can also include a page number along with your search terms to add it directly.\n\nYou can edit citations directly in the word processor document.
 firstRunGuidance.toolbarButton.new = Click here to open Zotero, or use the %S keyboard shortcut.
 firstRunGuidance.toolbarButton.upgrade = The Zotero icon can now be found in the Firefox toolbar. Click the icon to open Zotero, or use the %S keyboard shortcut.
+
+cslpreview.instruction		= Select one or more items in Zotero and click the "Refresh" button to see how these items are rendered by the installed CSL citation styles.
+cslpreview.warning.noItems		= No items selected in Zotero.
+
+csledit.saveAs.title		= Save Citation Style
+csledit.saveAs.CSL		= Citation Style Language
```
dstillman: Let's pick some prefix for bib-related stuff and put this and "Bibliography" and other stuff there. Maybe ```citations.citationStyleLanguage```, ```citations.bibliography```, etc.?

chrome/locale/en-US/zotero/cslpreview.dtd
```diff
@@ -0,0 +1,10 @@
+<!ENTITY zotero.cslpreview		"Zotero Style Preview">
+<!ENTITY zotero.cslpreview.refresh		"Refresh">
```
dstillman: Move to ```general.refresh```

chrome/locale/en-US/zotero/csledit.dtd
```diff
@@ -0,0 +1,6 @@
+<!ENTITY zotero.csledit		"Zotero Style Editor">
+<!ENTITY zotero.csledit.refresh		"Refresh">
```
dstillman: Move to ````zotero.general.refresh```.

chrome/locale/en-US/zotero/csledit.dtd
```diff
@@ -0,0 +1,6 @@
...
+<!ENTITY zotero.csledit.saveAs		"Save As…">
```
dstillman: Move to ```zotero.general.saveAs```.

chrome/locale/en-US/zotero/zotero.properties
rmzelle: Would "csl" be acceptable as a base prefix? E.g.:

```
csl
csl.bibliography
csl.warning.noItems
csl.preview.instruction
csl.editor.instructions
csl.editor.warning.parseError
csl.editor.warning.noItems
```

dstillman: I think ```styles``` would be better, since "bibliography", for example, is more generic than CSL.

rmzelle: Okay, will update soon.

rmzelle:
@dstillman, I addressed the label naming: https://github.com/rmzelle/zotero/commit/686f6e8517e692b92c4e92a03631e54dd5b25235

rmzelle:
Squashed.

rmzelle:
(I don't need to rename my branch, right?)

dstillman:
Thanks!

rmzelle:
:sparkles:

PR #608, opened 2015-01-13
Fix for OpenURL 0.1 resolver detection

COV-Steve:
The discoverResolvers function was looking for the wrong XML tag name representing OpenURL version 0.1 in the response from the WorldCat Registry


dstillman:
Thanks!

PR #609, opened 2015-01-16
Allow MAX_PAGES (for retrieving metadata) to be user-adjustable

aurimasv:
Re https://forums.zotero.org/discussion/45641/error-getting-metadata/#Item_10

I don't really see a reason not to do this. We could also consider bumping the default to 15 right away.


dstillman:
I'm OK with changing the default, but I don't see any point in making this user-adjustable.

aurimasv:
Well, mostly because I'm not sure what a good value would be TBH. It would be easiest if we could let someone experiment with this and figure it out. 15 seems reasonable to me though, but I don't deal with reports and working papers.

dstillman:
Let's just bump to 10 or 15 and have someone try that in the dev build.

aurimasv:
done

PR #610, opened 2015-01-18
Attach link from URI fix

aurimasv:
Re https://forums.zotero.org/discussion/45709/attach-link-to-url-not-working/


aurimasv:
This is sort of an interesting bug. This works pre-patch on Windows from unpacked source, because NTFS it's not case-sensitive. Once you pack it into an XPI file and Firefox takes over managing file access (or test it on a case-sensitive file system I guess), this becomes a problem

PR #614, opened 2015-01-19
Allow file paths to be specified in attachment.url

aurimasv:
In import translators, it may not always be clear whether the given URI is a URL or a local file path. I'm a little concerned that this may open up some security hole (by, hypothetically, allowing web translators and, thus, web pages to attach files located throughout the file system) since we use `attachment.url` in a lot of translators and we would pay more attention to translators using `attachment.path`, but (1) translators can already use file: protocol via URL (2) I'm not really sure how useful such attack would be since Zotero copies files and whether there is really any vulnerability here. Maybe we should limit attachment.path use to import translators called directly (not by web translators)

Regression from 849803473a0386aacc86745466d56f0ff3463f48 (that piece of code was introduced by me)

Re https://forums.zotero.org/discussion/45710/ris-import-from-endnote-mac-yosemite-to-zotero-standalone/

### DIFF
--- a/chrome/content/zotero/xpcom/translation/translate.js
+++ b/chrome/content/zotero/xpcom/translation/translate.js
@@ -631,6 +631,13 @@ Zotero.Translate.Sandbox = {
 				
 				for(var i=0; i<item.attachments.length; i++) {
 					var attachment = item.attachments[i];
+					
+					// Web translators are not allowed to use attachment.path
+					if (attachment.path) {
+						if (!attachment.url) attachment.url = attachment.path;
+						delete attachment.path;
+					}
+					
 					if(attachment.url) {
 						// Remap attachment (but not link) URLs
 						attachment.url = translate.resolveURL(attachment.url, attachment.snapshot === false);
--- a/chrome/content/zotero/xpcom/translation/translate_item.js
+++ b/chrome/content/zotero/xpcom/translation/translate_item.js
@@ -226,32 +226,64 @@ Zotero.Translate.ItemSaver.prototype = {
 			return false;
 		}
 		
-		if(!attachment.path) {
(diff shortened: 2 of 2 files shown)


dstillman:
Yes, if possible we should limit this to direct usage. I don't think there's much of an attack vector, but it'd be quite strange for a remote RIS file to be able to cause a file on your computer to be copied into your Zotero library.

aurimasv:
Could we push this out as is? Imposing restrictions as per above requires a bit code restructuring. ItemSaver doesn't know about the source of the attachment and with the automatic promotion of attachment.url to attachment.path if the former fails to be parsed, this is a bit tricky to implement without breaking stuff (I think).

simonster:
For web and search translators, `attachment.url` is resolved as a fully qualified URL relative to the current page before the attachments are passed to ItemSaver, so I don't think this should raise any security concerns.

OTOH, I think I would prefer never to treat URLs as paths, and to make translators that may encounter ambiguous URI/path strings perform the check themselves. It should be possible to do a good enough job with a regexp, although we could add a function to `Zotero.Utilities` if this crops up often.

aurimasv:
>OTOH, I think I would prefer never to treat URLs as paths, and to make translators that may encounter ambiguous URI/path strings perform the check themselves

I was thinking that we could allow demoting path to URL and only allowing path from import translators. I agree that URL should never be promoted to path. I'd rather make this simpler for translators if it's not a huge hassle to clean this up in the client.

aurimasv:
OK, I think this makes the most sense. We'll need to adjust import translators to use attachment.path instead of attachment.url (Zotero RDF works already, for instance)

dstillman:
Thanks. We haven't tested this, but Simon gave the general go-ahead.

(Could've let you squash this, I guess — sorry.)

PR #617, opened 2015-01-21
Content handlers

aurimasv:
This addresses two bugs:
* If Zotero is set to handle BibTeX/RIS/etc. files in Firefox, pressing Cancel was dropping the whole file instead of continuing to handle it by default application. This makes it impossible to download RIS files with the preference enabled.
* Changing the said preference was requiring a Zotero restart for it to take effect

It also adds a new feature. Zotero code can register observers for arbitrary Zotero preferences via `Zotero.Prefs.registerObserver`


aurimasv:
Oh, look at that... this closes #474

PR #620, opened 2015-01-28
Removed non-standard quotes from zotero.properties

bwiernik:
Commit 849803473a0386aacc86745466d56f0ff3463f48 added a localization string that used non-standard quote marks in the en-US locale. These can't be processed correctly by plug-ins on some systems. Standard quotes are used in other locales, so this is appears to be an error in the en-US locale.


dstillman:
There's nothing "non-standard" about those quotes — those are proper typographic quotes, and if anything the other ones should be changed to those. What do you mean by "can't be processed correctly by plug-ins on some systems"?

(And to be clear, this dialog is broken in 4.0.25 due to another issue, since fixed in the beta, but that doesn't have to do with the quotes.)

bwiernik:
Specifically I ran into an issue trying to read the .properties file using python's .read() function. The typographic quotes get mapped to <undefined>, returning an error. I thought that the inclusion of the styled quotes was perhaps an error given that they are not used elsewhere in the locale.

dstillman:
Nope, not an error. You just need to use the appropriate method for reading Unicode files in your version of Python.

PR #622, opened 2015-02-01
[API syncing] Add Zotero.FeedReader class

aurimasv:
This is the parser that fetches a feed and processes into Zotero items (in the same format as Zotero items coming from translators).

It was designed with the idea that feeds are processed as they are being downloaded and feed items become available at different times (not all at the same time). Unfortunately, current implementation in Firefox does not actually implement methods that allow processing the feed as parts of it become available and, instead, must be processed after all of it has been downloaded. Maybe this will change in the future.

The constructor takes a single mandatory parameter, which is the URL of the feed. The feed is fetched and processed asynchronously, but as soon as the object is created.

FeedReader exposes two properties:
* feedProperties: a promise for an object describing the feed (title, summary, URL, etc.)
* itemIterator: an iterator returning promises for items in the feed. When iterating items in the feed, each item MUST be resolved before requesting the next item or the list may be incomplete (that's probably not going to happen since the feed is downloaded in its entirety before items are parsed out). The last item in the list will always resolve to the NULL object (this is a little odd, but I think it makes sense because we're making a promise for an item without knowing if there actually will be another item) or will be rejected with a reason for pre-mature termination of feed processing.

FeedReader also exposes `terminate(reason)` method, which allows the feed parsing to be terminated prematurely (again, not very relevant, since the whole feed is fetched before processing).

Here's some sample code (can be run in Scratchpad):
```
var fr = new Zotero.FeedReader('http://cyber.law.harvard.edu/rss/examples/rss2sample.xml');
Zotero.Promise.coroutine(function*() {
  Zotero.debug(yield fr.feedProperties);
  let it = fr.itemIterator;
  let item;
  while (item = yield it.next().value) {
     Zotero.debug(item);
  }
})()
```
which results in something like:
```
console.log: zotero(3):
    'title' => "Liftoff News"
    'subtitle' => "Liftoff to Space Exploration."
    'updated' => [object Date] {}
    'creators' => [object Array] [
        '0' => [object Object] {
            'firstName' => ""
            'lastName' => "editor@example.com"
            'creatorType' => "author"
        }
    ]
    'language' => "en-us"

console.log: zotero(3):
    'guid' => "http://liftoff.msfc.nasa.gov/2003/06/03.html#item573"
    'title' => "Star City"
    'abstractNote' => "How do Americans get ready to work with Russians aboard the International Space Station? They take a crash course in culture, language and protocol at Russia's Star City."
    'url' => "http://liftoff.msfc.nasa.gov/news/2003/news-starcity.asp"
    'dateModified' => "2003-06-03 09:39:21"
    'date' => "2003-06-03 04:39:21 -05"
    'creators' => [object Array] [
        '0' => [object Object] {
            'firstName' => ""
            'lastName' => "editor@example.com"
            'creatorType' => "author"
        }
    ]
    'language' => "en-us"
    'itemType' => "journalArticle"
...
```

### DIFF
--- a/chrome/content/zotero/xpcom/feedReader.js
+++ b/chrome/content/zotero/xpcom/feedReader.js
@@ -1,0 +1,527 @@
+/*
+    ***** BEGIN LICENSE BLOCK *****
+    
+    Copyright © 2015 Center for History and New Media
+                     George Mason University, Fairfax, Virginia, USA
+                     http://zotero.org
+    
+    This file is part of Zotero.
+    
+    Zotero is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Affero General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+    
+    Zotero is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Affero General Public License for more details.
+    
+    You should have received a copy of the GNU Affero General Public License
+    along with Zotero.  If not, see <http://www.gnu.org/licenses/>.
+    
+    ***** END LICENSE BLOCK *****
+*/
+
+
+/**
+ * Sample feeds:
+ * 
+ * http://cyber.law.harvard.edu/rss/examples/rss2sample.xml
+ * http://feeds.feedburner.com/acs/acbcct
+ * http://www.cell.com/molecular-cell/current.rss
+ * http://ieeexplore.ieee.org/search/searchresult.jsp?searchField%3DSearch_All%26queryText%3Dwater&searchOrigin=saved_searches&rssFeed=true&rssFeedName=water
+ * http://www.sciencemag.org/rss/current.xml
+ * http://rss.sciencedirect.com/publication/science/20925212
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/xpcom/feedReader.js
```diff
@@ -0,0 +1,527 @@
...
+	
+	/*
+	 * Feed item iterator
+	 * Each iteration returns a _promise_ for an item. The promise _MUST_ be
+	 * resolved before requesting the next item.
+	 * The last item will always be resolved to `null`, unless the feed processing
+	 * is terminated ahead of time, in which case it will be rejected with the reason
+	 * for termination.
+	 */
+	Zotero.defineProperty(FeedReader.prototype, 'itemIterator', {
```
aurimasv: I'm reconsidering this. I think it would make most sense to make usage `var iterator = new fr.ItemIterator()` rather than `var iterator = fr.itemIterator` since `fr.itemIterator !== fr.itemIterator` given this code. Yes, no, maybe?

chrome/content/zotero/xpcom/feedReader.js
```diff
@@ -0,0 +1,527 @@
...
+		}
+		
+		let creators = [];
+		for (let i=0; i<names.length; i++) {
+			creators.push(
+				Zotero.Utilities.cleanAuthor(
+					names[i],
+					role,
+					names[i].split(',').length == 2
+				)
```
aurimasv: note to self: set `fieldMode: 1` if single word

PR #626, opened 2015-02-09
Tweaks to ISBN handling

aurimasv:
Mostly re https://github.com/zotero/translators/pull/847#discussion_r24300450

Import all ISBNs as ISBN-13 from web/search translators and place them in a space-separated format. Pass first ISBN to citeproc.

The one thing that we're not doing yet (and I would like to) is hyphenate the ISBNs. There's #394, but it adds a fairly large table, so we probably don't want to do it in connectors/bookmarklet (since it means downloading a relatively large file). The alternative to that would be applying formatting when writing to DB, but that makes it a little more difficult, since we have to deal with random user input. I'd say that we can hyphenate only if the ISBNs being put in follow the same format as those from translators (i.e. a list of unhyphenated ISBN-13)


chrome/content/zotero/xpcom/translation/translate.js
```diff
@@ -616,6 +616,31 @@ Zotero.Translate.Sandbox = {
...
+						var validISBN = Zotero.Utilities.cleanISBN(isbn[0]);
+						if (!validISBN) {
+							// Back up and move up one character
+							isbnRe.lastIndex = isbn.index + 1;
+							continue;
+						}
+						
+						validISBNs.add(Zotero.Utilities.toISBN13(validISBN));
+					}
+					item.ISBN = Array.from(validISBNs.keys()).join(' ');
```
aurimasv: err, guess I should change this (and Set above), since we'll need to polyfill it otherwise.

aurimasv:
bump

aurimasv:
Added ISBN hyphenation for Zotero in Firefox and Zotero Standalone. This happens when writing to DB, but in a very specific format, so it should not interfere with anything. This means that ISBNs will be hyphenated when coming from translators (from Connectors too) and when entered by the user. But, the user would have to enter the ISBN in unhyphenated format (as 10 or 13 digits). Any additional junk in the field will prevent hyphenation. Also, anything that cannot be hyphenated (because certain ISBN ranges have not been assigned yet) will remain unhyphenated.

This will not actually validate the ISBN to make sure that the check digit is correct.

Automatic hyphenation can be turned off via `extensions.zotero.hyphenateISBN` hidden pref if someone is really annoyed.

I'll push the script to create the file containing ISBN ranges to zotero-xpi-build if this goes through

dstillman:
> Automatic hyphenation can be turned off via extensions.zotero.hyphenateISBN hidden pref if someone is really annoyed.

Is there really a need for a pref? Unless there's really a compelling reason, I'd skip it.

aurimasv:
No terribly good reason, except it would no longer be possible to enter an un-hyphenated ISBN. I figured there would be at least a handful of users that will be annoyed by this, so I put in a way out for them. I feel like hidden prefs have a pretty low cost anyway.

zuphilip:
@aurimasv : I tested the Primo translator with this new functionality in up-to-date 4.0 master branch and the ISBN is not any more captured. E.g. http://primo.bib.uni-mannheim.de/MAN_UB:MAN_ALEPH001367873 with `ZU.cleanISBN("ISBN 978-3-927455-48-1 : (Pr. in Vorb.)") = false`.

dstillman:
Yes, I can reproduce that. (And this one falls into the unit-tests-not-optional category.)

aurimasv:
Fixed in #660 but I honestly don't have the time right now to invest in adding unit tests. Feel free to back this out if unit tests are a must.

aurimasv:
I mean, back it out if you need to create a release soon. I'll get to the unit tests at some point.. probably over the weekend.

PR #630, opened 2015-02-11
Don't try to parse path if it looks like a web URL

aurimasv:
Also, cleaner parsing messages

Currently, importing file attachments with HTTP urls goes through a bunch of parsing trying to find the local file first and then falling back to URL. This patch allows those check to be short-circuited.


dstillman:
Is this important for 4.0.26?

aurimasv:
no, everything should still work fine. This just cleans up logging and avoids some unnecessary steps.

dstillman:
OK, thanks. Will wait until after the release, then.

PR #623, opened 2015-02-03
Add drop-down menu for bibliographyLocale

rmzelle:
Groundwork for adding a drop-down menu to the Styles tab of the Cite tab of the Zotero preferences, for selecting the bibliography locale. See also https://forums.zotero.org/discussion/45805/

Still needs some work, but I'd like some guidance on how to implement things:

 * I placed https://github.com/citation-style-language/locales/blob/master/locales.json at https://github.com/zotero/zotero/tree/4.0/resource/schema. Is that okay? How are similar files at this location being kept up to date (like "renamed-styles.json")?
 * Fresh installs of Zotero have an empty string for "extensions.zotero.export.bibliographyLocale". With the current code, an empty string value for bibliographyLocale is replaced by the value of Zotero.locale when the preference tab is loaded, or, if Zotero.locale isn't set, by "en-US". Is this behavior problematic if the pref can be set via the UI?
 * The drop-down menu options are limited to the locales for which we have CSL locale files. Should there be a way to specify a custom locale code for "extensions.zotero.export.bibliographyLocale" via the UI? If so, what should that look like? A "Custom..." entry at the bottom of the drop-down list, that, if selected, pops up a dialog box with a text box?
 * Should there be a top-five list of most-recently used locales at the top of the drop-down list?
 * For some reason the drop-down box stretches across the entire window. Not sure why. XUL beats me.
 * I removed `width` on `<label id="export-citePaperJournalArticleURL" width="45em">`, since it didn't do anything on OS X. Is this still necessary? Maybe we can check on other platforms.

PR #632, opened 2015-02-11
Reserve height in div for CSL Editor buttons

fbennett:
The new CSL Editor looks great - thanks to @rmzelle for putting it all together.

One tiny nit that I just stumbled across is that (under Linux, at least) the buttons get a bit squashed when the style/preview divider is drawn up to completely obscure the style code. This change reserves space in the hbox containing the buttons so they can retain their natural height.

(I hope I've gotten my branching weirdness sorted out this time!)

### DIFF
--- a/chrome/content/zotero/tools/csledit.xul
+++ b/chrome/content/zotero/tools/csledit.xul
@@ -41,7 +41,7 @@
 	<script src="csledit.js"/>
 	
 	<vbox flex="1">
-		<hbox align="center">
+		<hbox id="zotero-csl-button-bar" align="center">
 			<button id="preview-refresh-button" label="&zotero.general.refresh;" oncommand="Zotero_CSL_Editor.refresh()"/>
 			<button id="zotero-csl-save" label="&zotero.general.saveAs;" oncommand="Zotero_CSL_Editor.save()"/>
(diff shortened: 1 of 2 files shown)


rmzelle:
I'd still like to see labeled groupboxes around the various options, as I proposed at https://github.com/zotero/zotero/pull/607#issuecomment-70037982, which might take care of this as well. Since the output is now much more compact, we might want to sacrifice the space it takes to add them.

Also, once we address the CSL locale UI selection (https://github.com/zotero/zotero/pull/623), I'd like to expand the top bar with a locale drop-down menu.

gracile-fr:
<blockquote>I'd still like to see labeled groupboxes around the various options,</blockquote>If you decide to add such explanatory labels, could you consider to add an option to hide/disable them? I'm really concerned with vertical space… Or maybe, tooltips would be sufficient? Thanks!

rmzelle:
@gracile-fr, I doubt hiding/disabling is worth the added complexity. Tooltips might work, but I'd still like groupboxes (in that case, without captions) to clearly group the different UI elements (such as the label/dropdown combos for selecting the locator and cite position).

You know you can fullscreen Firefox, right, or use Zotero Standalone and work in separate windows?

gracile-fr:
Yes (but <a href="https://forums.zotero.org/discussion/46671/standalone-any-zotero-secondary-window-is-out-of-screen/#Item_1">with Standalone…</a>).

fbennett:
Closing this dated PR.

PR #638, opened 2015-02-18
Fix possible infinite loops

aurimasv:


dstillman:
Thanks for fixing these.

PR #659, opened 2015-03-02
Another regression from f0bd1e77ffab6dbc7fbd600dc1acc11844aa2e02

aurimasv:
If we want to unify CSL JSON generation and make sure that what we pass to citeproc is the same as we export to CSL JSON and otherwise (I think that's a good idea), then we have to expect that items being passed into ZU.itemToCSLJSON are either in Zotero.Item format or the format that translators have access to. The latter was based on Zotero.Item.toArray (which is deprecated) plus a few additional modifications.

For one, this patch makes it less awkward to get exactly (almost) the format that export translators receive by moving all the code into Zotero.Utilities.Internal.itemToExportFormat.

Additionally, this patch transitions the object passed to export translators to Zotero.Item.serialize() format + compatibility fields that mirror toArray format. The idea is that new export translators (and old translators as they are slowly transitioned) should only rely on the serialize format. Once that is done, we can remove the compatibility  mappings.

Thus, we can now do one of two things (1) convert the Zotero.Item passed to itemToCSLJSON to the export format, which would match whatever metadata is passed into the function from translators, or (2) use Zotero.Item.serialize(), which would create the same fields that are also available in export translators (minus compatibility mapping). Currently, itemToCSLJSON uses the compatibility field mappings, but I'll transition it to use the pure Zotero.Item.serialize format when I get some more time.

Addresses https://forums.zotero.org/discussion/47176/beta-item-fields-mapping-to-weird-csl-variables/


aurimasv:
Oh, I should mention. Here's the diff of a "presentation" item from Zotero with tags, attachments, notes, related items. The diff compares `Zotero.Translate.ItemGetter.prototype._itemToArray` format (old) to `Zotero.Utilities.Internal.itemToExportFormat` (new)
```
{
  "itemID": 8
  "itemType": "conferencePaper"
  "dateAdded": "2015-03-01 14:55:56"
  "dateModified": "2015-03-02 05:01:42"
  "libraryID": null
  "key": "HZSRMJXJ"
  "url": "http://maamodt.asp.radford.edu/Research%20-%20IO/SIOP%202005%20-%20Reference%20Symposium.pdf"
  "rights": ""
  "series": ""
  "volume": ""
  "place": "Los Angeles, CA"
  "publisher": ""
  "pages": ""
  "ISBN": ""
  "date": "April 2005"
  "callNumber": ""
  "archiveLocation": ""
  "extra": "{:genre: Symposium}"
  "DOI": ""
  "accessDate": "2015-02-25 04:55:31"
  "libraryCatalog": ""
  "conferenceName": "annual conference of the Society for Industrial and Organizational Psychology"
  "language": "en"
  "abstractNote": ""
  "title": "Reliability, validity, and adverse impact of references and letters of recommendation"
  "proceedingsTitle": "References and recommendation letters: Psychometric, ethical, legal and practical issues"
  "shortTitle": ""
  "archive": ""
  "creators": [
    {
      "creatorType": "author"
      "creatorID": 13
      "firstName": "Michael G."
      "lastName": "Aamodt"
      "fieldMode": ""
+     "libraryID": null
+     "key": "PHZIMB8Z"
    }
    {
      "creatorType": "author"
      "creatorID": 14
      "firstName": "Felice"
      "lastName": "Williams"
      "fieldMode": ""
+     "libraryID": null
+     "key": "WS6JQJGI"
    }
    {
      "creatorType": "editor"
      "creatorID": 13
      "firstName": "Michael G."
      "lastName": "Aamodt"
      "fieldMode": ""
+     "libraryID": null
+     "key": "PHZIMB8Z"
    }
  ]
  "notes": [
    {
      "itemID": 4
      "itemType": "note"
      "dateAdded": "2015-03-02 05:01:10"
      "dateModified": "2015-03-02 05:01:32"
      "libraryID": null
      "key": "4N33M8NB"
      "note": "<p>test</p>"
      "sourceItemKey": "HZSRMJXJ"
      "tags": []
      "related": []
+     "primary": {
+       "itemID": 4
+       "itemType": "note"
+       "dateAdded": "2015-03-02 05:01:10"
+       "dateModified": "2015-03-02 05:01:32"
+       "libraryID": null
+       "key": "4N33M8NB"
+     }
+     "virtual": {
+       "firstCreator": ""
+       "numNotes": 0
+       "numAttachments": 0
+     }
+     "fields": {}
+     "relatedReverse": []
+     "uniqueFields": {
+       "itemID": 4
+       "itemType": "note"
+       "dateAdded": "2015-03-02 05:01:10"
+       "dateModified": "2015-03-02 05:01:32"
+       "libraryID": null
+       "key": "4N33M8NB"
+     }
    }
  ]
  "attachments": [
    {
      "itemID": 9
      "itemType": "attachment"
      "dateAdded": "2015-03-01 14:55:56"
      "dateModified": "2015-03-01 14:55:56"
      "libraryID": null
      "key": "RIJ693M2"
      "url": "http://maamodt.asp.radford.edu/Research%20-%20IO/SIOP%202005%20-%20Reference%20Symposium.pdf"
      "accessDate": "2015-02-25 04:55:31"
      "title": "SIOP 2005 - Reference Symposium.pdf"
      "note": ""
      "sourceItemKey": "HZSRMJXJ"
      "tags": [
        {
          "primary": {
            "tagID": 6
            "libraryID": null
            "key": "GDE7GFBW"
            "dateAdded": "2015-03-01 14:55:41"
            "dateModified": "2015-03-01 14:55:56"
          }
          "fields": {
            "name": ".fix"
            "type": 0
          }
          "linkedItems": [
            "9"
          ]
          "tag": ".fix"
          "type": 0
        }
      ]
      "related": []
+     "primary": {
+       "itemID": 9
+       "itemType": "attachment"
+       "dateAdded": "2015-03-01 14:55:56"
+       "dateModified": "2015-03-01 14:55:56"
+       "libraryID": null
+       "key": "RIJ693M2"
+     }
+     "virtual": {
+       "firstCreator": ""
+       "numNotes": 0
+       "numAttachments": 0
+     }
+     "fields": {
+       "url": "http://maamodt.asp.radford.edu/Research%20-%20IO/SIOP%202005%20-%20Reference%20Symposium.pdf"
+       "accessDate": "2015-02-25 04:55:31"
+       "title": "SIOP 2005 - Reference Symposium.pdf"
+     }
+     "attachment": {
+       "linkMode": 3
+       "mimeType": "application/pdf"
+       "charset": ""
+       "path": ""
+     }
+     "relatedReverse": []
+     "uniqueFields": {
+       "itemID": 9
+       "itemType": "attachment"
+       "dateAdded": "2015-03-01 14:55:56"
+       "dateModified": "2015-03-01 14:55:56"
+       "libraryID": null
+       "key": "RIJ693M2"
+       "url": "http://maamodt.asp.radford.edu/Research%20-%20IO/SIOP%202005%20-%20Reference%20Symposium.pdf"
+       "accessDate": "2015-02-25 04:55:31"
+       "title": "SIOP 2005 - Reference Symposium.pdf"
+     }
    }
  ]
  "tags": [
    {
      "primary": {
        "tagID": 7
        "libraryID": null
        "key": "P4UQJJRE"
        "dateAdded": "2015-03-01 14:55:41"
        "dateModified": "2015-03-01 14:55:56"
      }
      "fields": {
        "name": "fairness"
        "type": 0
      }
      "linkedItems": [
        "8"
      ]
      "tag": "fairness"
      "type": 0
    }
    {
      "primary": {
        "tagID": 8
        "libraryID": null
        "key": "A9UKVKX6"
        "dateAdded": "2015-03-01 14:55:41"
        "dateModified": "2015-03-01 14:55:56"
      }
      "fields": {
        "name": "meta-analysis"
        "type": 0
      }
      "linkedItems": [
        "8"
      ]
      "tag": "meta-analysis"
      "type": 0
    }
    {
      "primary": {
        "tagID": 9
        "libraryID": null
        "key": "FPHV99K8"
        "dateAdded": "2015-03-01 14:55:41"
        "dateModified": "2015-03-01 14:55:56"
      }
      "fields": {
        "name": "performance"
        "type": 0
      }
      "linkedItems": [
        "8"
      ]
      "tag": "performance"
      "type": 0
    }
    {
      "primary": {
        "tagID": 10
        "libraryID": null
        "key": "27SWJ2SR"
        "dateAdded": "2015-03-01 14:55:41"
        "dateModified": "2015-03-01 14:55:56"
      }
      "fields": {
        "name": "recommendations"
        "type": 0
      }
      "linkedItems": [
        "8"
      ]
      "tag": "recommendations"
      "type": 0
    }
    {
      "primary": {
        "tagID": 11
        "libraryID": null
        "key": "2HZ5XFGJ"
        "dateAdded": "2015-03-01 14:55:41"
        "dateModified": "2015-03-01 14:55:56"
      }
      "fields": {
        "name": "references"
        "type": 0
      }
      "linkedItems": [
        "8"
      ]
      "tag": "references"
      "type": 0
    }
    {
      "primary": {
        "tagID": 12
        "libraryID": null
        "key": "48C7A3GF"
        "dateAdded": "2015-03-01 14:55:41"
        "dateModified": "2015-03-01 14:55:56"
      }
      "fields": {
        "name": "reliability"
        "type": 0
      }
      "linkedItems": [
        "8"
      ]
      "tag": "reliability"
      "type": 0
    }
  ]
  "related": [
    7
  ]
  "seeAlso": []
  "uniqueFields": {
    "title": "Reliability, validity, and adverse impact of references and letters of recommendation"
    "abstractNote": ""
    "date": "April 2005"
    "publicationTitle": "References and recommendation letters: Psychometric, ethical, legal and practical issues"
    "conferenceName": "annual conference of the Society for Industrial and Organizational Psychology"
    "place": "Los Angeles, CA"
    "publisher": ""
    "volume": ""
    "pages": ""
    "series": ""
    "language": "en"
    "DOI": ""
    "ISBN": ""
    "shortTitle": ""
    "url": "http://maamodt.asp.radford.edu/Research%20-%20IO/SIOP%202005%20-%20Reference%20Symposium.pdf"
    "accessDate": "2015-02-25 04:55:31"
    "archive": ""
    "archiveLocation": ""
    "libraryCatalog": ""
    "callNumber": ""
    "rights": ""
    "extra": "{:genre: Symposium}"
+   "itemID": 8
+   "itemType": "conferencePaper"
+   "dateAdded": "2015-03-01 14:55:56"
+   "dateModified": "2015-03-02 05:01:42"
+   "libraryID": null
+   "key": "HZSRMJXJ"
  }
  "publicationTitle": "References and recommendation letters: Psychometric, ethical, legal and practical issues"
  "uri": "http://zotero.org/users/local/mEzbrP5W/items/HZSRMJXJ"
+ "primary": {
+   "itemID": 8
+   "itemType": "conferencePaper"
+   "dateAdded": "2015-03-01 14:55:56"
+   "dateModified": "2015-03-02 05:01:42"
+   "libraryID": null
+   "key": "HZSRMJXJ"
+ }
+ "virtual": {
+   "firstCreator": "Aamodt and Williams"
+   "numNotes": 1
+   "numAttachments": 1
+ }
+ "fields": {
+   "url": "http://maamodt.asp.radford.edu/Research%20-%20IO/SIOP%202005%20-%20Reference%20Symposium.pdf"
+   "rights": ""
+   "series": ""
+   "volume": ""
+   "place": "Los Angeles, CA"
+   "publisher": ""
+   "pages": ""
+   "ISBN": ""
+   "date": "April 2005"
+   "callNumber": ""
+   "archiveLocation": ""
+   "extra": "{:genre: Symposium}"
+   "DOI": ""
+   "accessDate": "2015-02-25 04:55:31"
+   "libraryCatalog": ""
+   "conferenceName": "annual conference of the Society for Industrial and Organizational Psychology"
+   "language": "en"
+   "abstractNote": ""
+   "title": "Reliability, validity, and adverse impact of references and letters of recommendation"
+   "proceedingsTitle": "References and recommendation letters: Psychometric, ethical, legal and practical issues"
+   "shortTitle": ""
+   "archive": ""
+ }
+ "relatedReverse": []
}
```

There's still one small difference: new export format lacks a few properties/methods from attachments (e.g. saveFile) that don't really make much sense unless you're passing this object to the export translator. (I did add itemType, mimeType, charset after generating this diff).

dstillman:
I'll look at this later, and Simon may have some comments, but do we actually want to keep the ``serialize()`` format? If possible, I think we should just use API JSON for everything, since there's actually a specification for it. I haven't looked at the ``serialize`` format in a while, but the biggest difference I can recall is that the API doesn't nest children, so those would have to be included as separate items in the array.

dstillman:
(I believe we have JS code to generate API JSON on the API syncing branch, if not on 4.0.)

simonster:
I agree with @dstillman that it would be best to use API JSON for everything unless it makes things overly difficult.

aurimasv:
> API doesn't nest children, so those would have to be included as separate items in the array.

Do you mean `attachments` array? Or do you mean that the attachments should be at the same level as regular items and linked to from regular items?

If the latter, then that makes things unnecessarily complicated for export translators. If the former, then I'm not sure it makes sense to transition to API JSON if we have to morph it into something incompatible anyway. Might as well just stick to the current format (just make sure that it's consistently that).

dstillman:
I mean the latter. They're linked from child to parent, not the other way around. Child items have a ``parentItem`` property, which currently is an item key, but I've [https://github.com/zotero/translators/issues/775](also suggested) allowing ``parentItem: last`` to use the previous top-level item.

Why is that unnecessarily complicated? ``serialize()`` is an improvement on ``toArray()``, but they're both hacky, unspecified formats that would have to be maintained when we make data changes. It'd be far better to have a single, specified format that contains all relevant data and is maintained when we make changes, and that's API JSON.

dstillman:
Actually, it looks like I even deleted ``toArray`` and ``serialize`` from Zotero.Item back in August on the async_db branch, replacing them with ``toJSON``.

I never actually intended these functions as anything other than internal formats. While ``toArray`` (I think?) ended up being used in translators out of necessity, now that we have a well-documented format, I don't see why we wouldn't switch to that. I have no interest in maintaining those functions anymore, so even morphing API JSON into something else for translators would be preferable to me, but I think exposing API JSON to translator authors makes sense.

aurimasv:
Do we only have [documentation](https://www.zotero.org/support/dev/web_api/v3/write_requests#json_object_data)-by-[example](https://www.zotero.org/support/dev/web_api/v3/write_requests#updating_an_existing_item) for API JSON? There's no description of the (recommended) format for each field, like date for instance?

More specifically, are field names supposed to be the item type-specific field names directly from Zotero client? I.e. can we use http://aurimasv.github.io/z2csl/typeMap.xml as a reference for all fields?

dstillman:
Well, it's the same as the client (and ``toArray``) — there aren't really any recommended formats for most of the fields. ``dateAdded``, ``dateModified``, and ``accessDate`` have to be ISO 8601, which is documented (though the last one only [here](https://www.zotero.org/support/dev/web_api/v3/changes_from_v2)), but that's it.

But yes, like ``toArray``, it's just the type-specific fields, retrievable [via the API](https://www.zotero.org/support/dev/web_api/v3/types_and_fields#zotero_web_api_item_typefield_requests), plus the additional fields from the documentation and template requests (``key``, ``version``, ``parentItem``, ``itemType``, ``tags``, ``collections``, ``relations``).

aurimasv:
> there aren't really any recommended formats for most of the fields

You can somewhat gather this from the examples, but, for instance, the format for creators is not obvious ('name' for institutional authors, 'firstName'/'lastName' for personal). `type` for tags is obscure.

The following pertain to manual user entry as much as they do to API JSON and translator import, so we should document this in general and refer to it from the API JSON specification as well.
* Date field parsing is not perfect, so the recommendation would be to submit it in something like ISO 8601 (or some other unambiguous format that Zotero can parse). We should also serve the item with both parsed and unparsed dates via the API.
* Given recent changes to ISBN/ISSN fields, we should provide guidance on what the best format for submitting those would be.
* Zotero also supports some limited markup in titles and certain markup in abstract. This can be found elsewhere (at least the title part), but should be at least linked to from the API JSON spec.
* Some other formatting conventions that don't really have any documentation, but have been practiced for translators include:
    * we generally don't enter "1" for number of volumes if there is only one volume. We also don't enter "1" for first edition if it's the only edition.
    * There has been some consensus on entering numberOfPages for multi-volume works and for works that paginate different sections of the book differently. https://github.com/zotero/translators/pull/756

I'm sure there are more things that I missed above.

>But yes, like toArray, it's just the type-specific fields, retrievable via the API

"retrievable via the API", while helpful, is not really a good source of documentation, so we should probably list supported fields (and try to keep it up to date or generate automatically from the API)

dstillman:
> We should also serve the item with both parsed and unparsed dates via the API.

(Parsed dates are served via the ``meta.parsedDate`` field.)

> "retrievable via the API", while helpful, is not really a good source of documentation

Retrieving via the API is the recommended approach when actually using the API. I agree it's not as helpful for non-API usage. These may change more fluidly in the future, so generating documentation automatically from the API would make the most sense. (Can't do that easily from DokuWiki, though.)

aurimasv:
>>We should also serve the item with both parsed and unparsed dates via the API.
>
>(Parsed dates are served via the meta.parsedDate field.)

Is there a description for `meta`, `links`, and `library` fields?

dstillman:
A bunch of things are documented in [Changes From V2](https://www.zotero.org/support/dev/web_api/v3/changes_from_v2) that should probably be added to the main documentation, but they're all shown in [linked examples](https://api.zotero.org/users/475425/items/X42A7DEE?v=3).

But to clarify, we're only talking about using the contents of the ``data`` object here, not the full response JSON.

aurimasv:
OK, this is now based on API JSON v3 + `attachments` and `notes` arrays.

Because of `version` clash and rename in v3, as well as different values for attachment `linkMode` (strings vs integers), translators that claim minVersion before 4.0.27 are served a legacy-compatible version of the export translator. This should ensure that no current translator breaks, _except_ CSL JSON, which needs to be bumped to minVersion 4.0.27 in order to export `version` for `computerProgram`

This may need some more testing. (and yes, yes, excellent candidate for unit testing, but I just don't have time to mess with that right now)

aurimasv:
Done

aurimasv:
@simonster, you might want to look at https://travis-ci.org/zotero/zotero/builds/53934455 It only failed for Fx ESR and I don't see anything related to the last commit (build passed in previous commit)

simonster:
Well, someone needs to write unit tests to make sure that combining that using itemToCSLJSON doesn't change what we're passing to citeproc-js. The reason I didn't combine these earlier was that I was afraid of these kinds of regressions.

aurimasv:
OK, the data being output to citeproc-js is not being tested. The citeproc-js data was generated after reverting f0bd1e77ffab6dbc7fbd600dc1acc11844aa2e02 and it passes now with the new itemToCSLJSON implementation.

Still need to do the same for itemToExportFormat

test/tests/support.js
```diff
@@ -9,4 +9,91 @@ describe("Support Functions for Unit Testing", function() {
...
+		let citeURL = Zotero.Prefs.get("export.citePaperJournalArticleURL");
+		before(function () {
+			Zotero.Prefs.set("export.citePaperJournalArticleURL", true);
+		});
+		after(function() {
+			Zotero.Prefs.set("export.citePaperJournalArticleURL", citeURL);
+		});
+		
+		it("all citeproc-js export data should be up to date", function() {
+			this.timeout(45000);
```
aurimasv: This test appears to be quite fast on Travis-CI (~500ms), but it is taking ~25 seconds on my computer. Not really sure why there's such a huge discrepancy.

test/content/makeTestData.js
```diff
@@ -0,0 +1,29 @@
+if (ZoteroUnit.makeTestData) {
+	Components.utils.import("resource://gre/modules/osfile.jsm");
+	Components.utils.import("resource://gre/modules/AddonManager.jsm");
+	
+	AddonManager.getAddonByID("zotero-unit@zotero.org", function(addon) {
+		var dataPath = addon.getResourceURI('tests/data')
+			.QueryInterface(Components.interfaces.nsIFileURL).file.path;
```
aurimasv: Use https://github.com/zotero/zotero/blob/4.0/test/content/support.js#L150-L157 instead

test/tests/support.js
dstillman: I'd guess it's the lack of a transaction in ``populateDBWithSampleData()``, combined with terrible access speeds on Windows. Without a transaction, I believe SQLite flushes to disk after every statement.

dstillman:
So the idea here is that, when we made changes to sample data, we'd run ``./runtests.sh -t`` and then commit the changes, and otherwise it'd use the existing sample data? Any reason for -t? Maybe -g for "generate"?

bwiernik:
Perhaps also all the numbers should not be the same? In one of the initial bugs with the first CSL JSON commit was that the same number field (issue) was being sent to four or five other number fields.

test/content/support.js
```diff
@@ -1,3 +1,18 @@
...
+	if(!failed) {
+		OS.File.writeAtomic(OS.Path.join(OS.Constants.Path.profileDir, "success"), new Uint8Array(0));
+	}
+	if(!ZoteroUnit.noquit) {
+		Components.classes['@mozilla.org/toolkit/app-startup;1'].
+		getService(Components.interfaces.nsIAppStartup).
+		quit(Components.interfaces.nsIAppStartup.eForceQuit);
+	}
+}
+
```
simonster: This is not really support code. It should never be called from a test.

test/content/makeTestData.js
```diff
@@ -0,0 +1,23 @@
...
+		
+		OS.File.writeAtomic(OS.Path.join(dataPath, dataFiles[i] + '.js'), str);
+		
+		ZoteroUnit.dump('done.');
+	}
+	
+	Zotero.Prefs.set("export.citePaperJournalArticleURL", oldPref);
+	
+	quit(false);
+}
```
simonster: I think you could just stick this in `runtests.js` and remove `include.js`. I don't see a compelling advantage to splitting this up since `runtests.js` will still be pretty short.

adam3smith:
(could we add a "Blocker" label to this one? It's causing me a lot of grief in the beta and it'd be very bad to release without this fixed).

aurimasv:
>So the idea here is that, when we made changes to sample data, we'd run ./runtests.sh -t and then commit the changes, and otherwise it'd use the existing sample data? Any reason for -t? Maybe -g for "generate"?

Yes. It's also helpful to see why the test isn't passing if that's the case. "-t" was for "test data", but yeah, it's not very meaningful. "-g" sound good to me.

>Probably better for the creator names not to all be the same? One could envision an export bug where the names got jumbled...

More diverse data would be good, but this need to be automatically generated. I'm open to suggestions on how to make this better. e.g. we could prepend things like "ja" (for journal article) to regular text fields and creator roles to creator names (e.g. "jaAuthorFirst") to make values more unique. I suppose for these tests we don't particularly care about having dates for dates and numbers for numeric fields (in most cases), so we could use the same strategy there as well.

>Also, we're envisioning different, more-specific data for things like bibliographic output, right?

Yes, we'll need different data to test various aspects of Zotero.

aurimasv:
That should address all concerns.
* All field values within one item should now be unique. We're not testing for value jumbling across items here.
* "-g" is now the command line flag to generate tests
* moved all code back into runtests.js

Still need to add tests for translator item format and then clean up the commits.

aurimasv:
OK, added tests for translator export data. What's important here are the changes in d63f08b2f5f1585fd86a831fa63b8ca9513741dd which is basically the new item format for all current export translators and changes in https://github.com/aurimasv/zotero/commit/bae641bf388602f414299d4d86576787a9e0846d which is the export format for new translators.

d63f08b2f5f1585fd86a831fa63b8ca9513741dd looks right to me, but bae641bf388602f414299d4d86576787a9e0846d needs some fixing, I think. Please feel free to point out any changes that are undesirable and keep in mind that the new format is based on the Zotero web API JSON format.

aurimasv:
>If this needs to be sorted by fieldID (does it?), should concat() either here or when returning a cached array in Zotero.ItemFields.getItemTypeFields(). Right now I believe this would change the order of the stored cached array.

@dstillman, with stable stringify, we don't need to sort the array any more. Do you want to revert the `.slice()` changes to `Zotero.ItemFields`?

aurimasv:
@dstillman I think this is pretty much done.

chrome/content/zotero/xpcom/utilities.js
```diff
@@ -1527,14 +1484,14 @@ Zotero.Utilities = {
 	 */
 	"itemToCSLJSON":function(zoteroItem) {
 		if (zoteroItem instanceof Zotero.Item) {
-			zoteroItem = zoteroItem.toArray();
+			zoteroItem = Zotero.Utilities.Internal.itemToExportFormat(zoteroItem);
```
fbennett: Testing with the current beta, if a note or attachment item is argument, zoteroItem.creators is null, and itemToCSLJSON() explodes at [line 1539](https://github.com/aurimasv/zotero/blob/csl-json-export/chrome/content/zotero/xpcom/utilities.js#L1539). It may be intended that bare items never reach this function (?), but either wrapping the `for` block in a condition, or something like
``` javascript
if (!creators) {
    creators = [];
}
```
would avoid the error.

aurimasv: I don't think `itemToCSLJSON` is meant to handle notes and attachments (though I could very well be wrong. Can citeproc/CSL handle notes or attachments in any reasonable way), primarily because [`CSL_TYPE_MAPPINGS`](https://github.com/zotero/zotero/pull/659/files#diff-a098326f7cdaa37adf1398d92b384be5R103) does not map these "item" types. We do fall back to "article" in those cases, but I don't think that is the right thing to do. For unrecognized types, we should be throwing an exception here.

Attachments and notes may reach this function, so we should handle them gracefully. I'm thinking that we should return `null` in those cases. I don't think we ever intended on citing file attachment, but I'm not sure about notes.

fbennett: With a bare attachment and an empty creators array, citeproc-js will render the title and URL of an attachment. That looks an awful mess, mostly, but I think it should be allowed to render.

Bare items have been showing through in the WP plugins (they are screened out in quickFormat, but they show in the Classic view), so they will be present in documents, and a refresh will attempt to retrieve them from Zotero. If the function returns null, citeproc-js could either throw, or render a "missing data" cite (not sure what it would do currently), but either scenario would leave the user without a means of backtracking to the bare item, which they might want to do in order to identify the source and replace the bare item with a proper reference.

fbennett: `citeproc-js` will process items without a type. CSL doesn't have a "Document" type, and Zotero items of type "Document" come through that way.

aurimasv: In CSL, what is "entry" type? Would that be a good candidate for generic items, like note, attachment, and document? We can continue mapping these to "article", I don't particularly mind, but it should be made explicit.

fbennett: I guess `article` may be the best policy for the present. The lack of a `document` type in CSL was discussed in CSL (quite awhile back), but a formal input specification hasn't emerged yet:
* [Document type](http://xbiblio-devel.2463403.n2.nabble.com/Document-type-td5140098.html)
* [Citeproc-json data input specs](http://xbiblio-devel.2463403.n2.nabble.com/Citeproc-json-data-input-specs-td5135372.html)
The alternative to using `article` would be to deliver nothing, or an invalid type like `document`, but that could cause problems for other processors, if Zotero-generated CSLJSON is delivered to them. I may have been mistaken in what I wrote above: there is a mention in the linked discussions of using `article` as a proxy for no-type, so it may be serving that purpose already. @adamsmith or @rmzelle will have better knowledge of the assumptions made in styles.

aurimasv: For now I'll continue mapping these to `article`, because there are plenty of styles that do reference `article` explicitly. On the other hand, I doubt that any of these styles are actually attempting to format `document` items or file attachments/notes.

aurimasv:
@fbennett, this should be fixed in e93aad317c3ba54ca26ca024b13ef580c5140b21 and I added a mapping for notes from note to title the same way that is done for display in Zotero.

@adam3smith and @rmzelle if you have time, please take a look at the last few commits.

aurimasv:
@fbennett, could you give me some guidance why I'm getting `this.state.registry.registry[Item.id] is undefined`  @citeproc.js:14285:8 and @citeproc.js:6409:0 ? This patch does change `id` from Zotero Item ID to Item URI, which should be more meaningful, but even after reverting the change, the error remains. Also, the tests are showing the `id` is getting exported https://github.com/zotero/zotero/pull/659/files#diff-416fa9a7f56dfe0b3a5afd2238dfb317 Any thoughts on where it might be getting lost?

rmzelle:
Looks reasonable to me. We can always adjust based on feedback.

@adam3smith would know better, but I don't think we tend to recommend style authors to rely on the broad item types like "article", mostly because most (regular) items will be assigned one of their subtypes instead.

There is also a bit of history here, since pre-CSL 1.0, CSL included some type fallback behavior that since has been removed. E.g. an item of type "article-journal" would test true for a check against the type "article" (with the "type" conditional). See http://xbiblio-devel.2463403.n2.nabble.com/removing-fallback-behavior-tt3206299.html, https://forums.zotero.org/discussion/11960/item-type-testing-in-csl-10-and-fallbacks/, and https://bitbucket.org/bdarcus/csl-utils/commits/7bda7dee1871/#Lupdate.xslT153 . All styles received an one-time upgrade to add all subtypes of "article" to every conditional testing for "article" (and the same was done for "book" and "chapter").

adam3smith:
```article``` is also Mendeley's fallback when it doesn't know what to do with an item type, so aligning that makes sense.

**Comment by fbennett** [comment]
> @fbennett, could you give me some guidance why I'm getting this.state.registry.registry[Item.id] is undefined @citeproc.js:14285:8 and @citeproc.js:6409:0 ?

There have been occasional reports of this error from users, but I have not yet been able to reproduce it. If you have it under glass in a test, I would like to trial it here and track the fault down in the processor. What environment and source configuration will I need to run the tests? I'm on the road for the next five days, but when I get back I'll look forward to pinning this one down.

**Comment by aurimasv** [comment]
@dstillman, I think this should be ready for merge.

File: chrome/content/zotero/xpcom/data/itemFields.js [hunk 6]
```diff
@@ -172,7 +173,7 @@ Zotero.ItemFields = new function() {
 		}
 		
 		if (_itemTypeFields[itemTypeID]) {
-			return _itemTypeFields[itemTypeID];
+			return _itemTypeFields[itemTypeID].slice();
```
simonster: If the concern here is that a consumer could mutate these, maybe calling `Object.freeze()` on the objects when we construct them would be better than copying them every time we return them?

test/content/support.js
```diff
@@ -167,4 +196,308 @@ function resetDB() {
 	}).then(function() {
 		return Zotero.Schema.schemaUpdatePromise;
 	});
+}
+
+function stableStringify(obj, level, label) {
```
simonster: Should have a comment, although I'm pretty sure I understand what it does.

chrome/content/zotero/xpcom/utilities.js
```diff
@@ -1599,7 +1563,14 @@ Zotero.Utilities = {
 		
 		// get date variables
 		for(var variable in CSL_DATE_MAPPINGS) {
-			var date = zoteroItem[CSL_DATE_MAPPINGS[variable]];
+			var date = zoteroItem[CSL_DATE_MAPPINGS[variable]]
```
simonster: What happened to the semicolon?

test/content/support.js
```diff
@@ -167,4 +196,308 @@ function resetDB() {
...
+	return data;
+}
+
+/**
+ * Loads specified sample data from file
+ */
+function loadSampleData(dataName) {
+	Components.utils.import("resource://gre/modules/Services.jsm");
+	let data = {};
+	Services.scriptloader.loadSubScript('resource://zotero-unit-tests/data/' + dataName + '.js', data, 'UTF-8');
```
simonster: Is there a reason we need to use the subscript loader here rather than loading as JSON?

simonster:
:+1: @dstillman should review the `item.js` changes, but this looks good to me.

dstillman:
Same question about loadSubScript, but otherwise this looks good to me.

chrome/content/zotero/xpcom/data/itemFields.js
aurimasv: Honestly, I don't exactly recall, but that sounds like the concern I would have had. Looking through the code, I don't see anything that attempts to modify the returned value, so freezing it should be OK. (I've actually never come across `Object.freeze`, which is rather neat)

test/content/support.js
aurimasv: Again, it's been a while, but either I was planning to stringify/load some functions as well (something that didn't happen) or it was  just a simple way for me to access data via 'resource://' URI, which there are better ways to go about. Will fix.

chrome/content/zotero/xpcom/data/itemFields.js
aurimasv: Though at this point, the change is not relevant to the PR, so I'll just remove it from the commits and we can address it later if it's still relevant.

aurimasv:
Done. If everything is ok, I'll squash and remove f755587ec99c8589576b9c6f3f554a0486787e87

dstillman:
All right, looks good.

aurimasv:
Turns out not so much. Tests are failing. I'm looking into it.

aurimasv:
Fixed and squashed

simonster:
Thanks!

dstillman:
Well that was easy.

aurimasv:
Great! Then feed reader changes will be easy too :-)

dstillman:
So who wants to merge this into 5.0?

aurimasv:
I can't get to it until the weekend, sorry. Thesis deadline looming on
Friday.

simonster:
I'll take care of it.

PR #660, opened 2015-03-02
Revise ZU.cleanISBN

aurimasv:
Fixes regression from 37921b09100cd35697173be39f272837ec7c1f95 but also this should be better overall

### DIFF
--- a/test/tests/utilities.js
+++ b/test/tests/utilities.js
@@ -17,4 +17,159 @@ describe("Zotero.Utilities", function() {
 			}
 		});
 	});
+	describe("cleanISBN", function() {
+		let cleanISBN = Zotero.Utilities.cleanISBN;
+		it("should return false for non-ISBN string", function() {
+			assert.isFalse(cleanISBN(''), 'returned false for empty string');
+			assert.isFalse(cleanISBN('Random String 123'), 'returned false for non-ISBN string');
+			assert.isFalse(cleanISBN('1234X67890'), 'returned false for ISBN10-looking string with X in the middle');
+			assert.isFalse(cleanISBN('987123456789X'), 'returned false for ISBN13-looking string with X as check-digit');
+		});
+		it("should return false for invalid ISBN string", function() {
+			assert.isFalse(cleanISBN('1234567890'), 'returned false for invalid ISBN10');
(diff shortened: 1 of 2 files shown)


simonster:
This would be an excellent candidate for unit testing.

aurimasv:
Tests added for `cleanISSN`, `cleanISBN`, and `toISBN13`

aurimasv:
Note that there's a fairly exhaustive test on what characters should and should not be ignored between ISSN/ISBN digits. It goes through all the characters in the BMP, so that takes a while (~1-1.5 seconds). We probably don't have to be that exhaustive though, so I can trim it down if there's a concern.

test/tests/utilities.js
```diff
@@ -17,4 +17,159 @@ describe("Zotero.Utilities", function() {
 			}
 		});
 	});
+	describe("cleanISBN", function() {
+		let cleanISBN = Zotero.Utilities.cleanISBN;
+		it("should return false for non-ISBN string", function() {
+			assert.isFalse(cleanISBN(''), 'returned false for empty string');
+			assert.isFalse(cleanISBN('Random String 123'), 'retruned false for non-ISBN string');
```
zuphilip: Typo (retruned) which occurs in two other lines as well.

aurimasv: Thanks! I fixed this and squashed.

aurimasv:
@simonster can you take a look and let me know if this is how I should proceed with tests for other components

simonster:
@dstillman has more experience with mocha/chai than I do, but this looks good to me. I think a simpler test ignored chars would have been sufficient, although testing all of them doesn't hurt and 3 seconds isn't a big deal.

simonster:
Thanks!

dstillman:
We're getting a Travis failure due to a timeout for that test.

```
x  should strip off internal characters in ISSN string

Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test
```
We could [adjust the timeout](http://mochajs.org/#test-specific-timeouts) for that test, though I agree with Simon that a simpler test would probably be sufficient here.

dstillman:
(It's true that 3 seconds isn't a big deal with 30 tests, but we might feel differently when we have hundreds.)

PR #663, opened 2015-03-04
Call child translator "done" handler when the child translator is done.

aurimasv:
Fixes https://github.com/zotero/translators/pull/861

Done handler was never getting called when using getTranslatorObject


simonster:
It seems like the "done" handler is kind of a nebulous concept if you call `getTranslatorObject`. What if one calls `getTranslatorObject`, makes an HTTP request from the parent translator from the callback, and then calls a function that saves an object through the child translator from the HTTP request callback? When does "done" get called?

aurimasv:
Yes, those sort of situations would break this callback model. I don't have a good solution for it though (first thing that comes to mind is to re-introduce something like `Zotero.wait`) We'll just have to watch out for it (I don't think there's currently a translator that does this)

PR #665, opened 2015-03-07
Unit testing infrastructure

simonster:
Implements the beginnings of unit testing infrastructure using mocha/chai. The unit tests can be run locally using `test/runtests.sh`, although this will need tweaks to run on Windows. They should also run on commit using Travis-CI.

The unit tests themselves live in `test/tests`. The `index.js` file specifies separate test sets, which can be run individually by calling `test/runtests.sh <testsets>`. Right now there is only a single unit test, but hopefully we'll have more soon...


.travis.yml
```diff
@@ -0,0 +1,16 @@
...
+env:
+    matrix:
+        - FIREFOXVERSION="36.0.1"
+        - FIREFOXVERSION="31.5.0esr"
+notifications:
+    email: false
+before_install:
+    - export DISPLAY=:99.0
+    - sh -e /etc/init.d/xvfb start
+    - wget ftp://ftp.mozilla.org/pub/firefox/releases/${FIREFOXVERSION}/linux-x86_64/en-US/firefox-${FIREFOXVERSION}.tar.bz2
```
dstillman: Probably better to use HTTP, to at least give Travis the chance to do intermediate caching. No idea if they do.

Issue #582, opened 2014-11-20
Confirmation dialog for remove from collection


dstillman:
https://forums.zotero.org/discussion/42302/

guitorri:
I have just been bitten by this. I accidentally pressed delete a couple of times (I thought the focus was elsewhere) and items were removed from one of my collections. They were not yet tagged and I cannot find them back by memory.

rmzelle:
I call dibs on this one (hope to provide a fix in a few days).

PR #668, opened 2015-03-08
Add prompt for Remove Item(s) from Collection

rmzelle:
Takes care of #582.

I didn't see a way to avoid the prompt with a modifier key, since using Cmd on Mac already switches the action from "remove from collection" (force=false) to "delete" (force=true).


chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -204,8 +204,11 @@ pane.items.trash.multiple		= Are you sure you want to move the selected items to
 pane.items.delete.title			= Delete
 pane.items.delete				= Are you sure you want to delete the selected item?
 pane.items.delete.multiple		= Are you sure you want to delete the selected items?
-pane.items.menu.remove                    = Remove Item from Collection
-pane.items.menu.remove.multiple           = Remove Items from Collection
+pane.items.remove.title			= Remove from Collection
+pane.items.remove				= Are you sure you want to remove the selected item from the currently selected collection?
+pane.items.remove.multiple		= Are you sure you want to remove the selected items from the currently selected collection?
```
dstillman: I think "selected item from the currently selected collection" is too awkward. The "currently", at least, is totally redundant, and since "selected item from the selected collection" is awful, I'm inclined to go with just "…remove the selected item from this collection?". I'm not sure we use "this" that way anywhere else, but I don't know what the alternative is, since if we stop with just "…remove the selected item?" and have "Remove Item from Collection" as the button it sounds too much like deleting from the library.

aurimasv: Or drop the "this" as well. "...remove selected items from collection"

dstillman: No, that works in the button, but not for the sentence text.

aurimasv: "Are you sure you want to remove selected items from collection?" Sounds fine to me, but either way is ok.

dstillman: (Fortunately no one will ever read this message, so it doesn't matter too much. And we do actually use "this" in some contexts, so I think it's fine here.)

rmzelle:
@dstillman, changed per your suggestion and squashed.

dstillman:
Thanks!

rmzelle:
Thanks for accepting! (I'm trying to build karma for a review of #623 :P)

gracile-fr:
@rmzelle 
At the moment, if you select a child note of an item in a collection and press the del key, nothing happens (that's ok). But with this PR, the prompt (Remove Item(s) from Collection) appears, even if clicking "yes" has no effect on the child note. => The prompt should not appear in this case.

rmzelle:
Wouldn't it be preferable to bind "del" to "move to trash" for child notes? (and give a prompt)

Or would that be too confusing?

gracile-fr:
Useful but confusing, I don't know… @adam3smith @aurimasv ?

adam3smith:
I agree with Rintze - del for child items should always move them to trash (with the appropriate warning, of course). I think that's less confusing than both the status quo before and after Rintze's change.

rmzelle:
We should also keep in mind that it's possible to select a standalone item and a child note (of a different item). Pressing "del" should probably remove the item from the collection, and leave the child note alone, for that case.

dstillman:
> del for child items should always move them to trash

I don't think that's right. Having different things happen in a given view based on what's selected strikes me as confusing and dangerous. The keys should be view-specific, not item-specific. The modified delete in a collection is the move-to-trash action, and that should be the way to move child items to the trash from a collection.

> Pressing "del" should probably remove the item from the collection, and leave the child note alone, for that case.

That strikes me as totally bizarre, and I think that's the problem with having the same key do different things in the same view. If we show a "Child items cannot be removed from collections." alert, as @gracile-fr suggests, for an unmodified Delete on a child item — which I'm open to — it could show in this case as well.

rmzelle:
> That strikes me as totally bizarre

Well, it's not bizarre if we reserve the unmodified "del" key for removal from collection, right?

dstillman:
I still think it's a little weird to act on only some of the selected items, but at least that way it'd be applying the normal function on each selected item (which, for child items, is a no-op). One reason to do that is that it's probably common for people to select a range of items, with some expanded, and expect the top-level ones to be acted upon. I think we've discussed something similar in the context of dragging with some child items selected (where currently, I think, we cancel the drag), and I've said that if all child items of an item are selected we should do it. (Not saying all child items would have to be selected here, just that the use case is similar.)

rmzelle:
> it's probably common for people to select a range of items, with some expanded, and expect the top-level ones to be acted upon.

Right. In that case a "Child items cannot be removed from collections." alert would be annoying to the user, though.

dstillman:
Yes, the prompt was if we had it not do anything on mixed selections.

rmzelle:
So do you want a follow up pull request to just eliminate the alert when only child items are selected?

Is there any "itemGroup" function available to see if the selection only contains child items?

dstillman:
> So do you want a follow up pull request to just eliminate the alert when only child items are selected?

Yes please.

> Is there any "itemGroup" function available to see if the selection only contains child items?

No, but you can do something like ``this.itemsView.getSelectedItems().every(item => !item.isTopLevelItem())``.

dstillman:
OK, after trying it out, I return to my previous statement that having it act on only some of the selected items is indeed bizarre, even when the keyboard actions are consistent within each view. This should act like I said we had considered for drags: if all child items of a selected top-level item are selected, then that's fine (to allow for selections across expanded items), but if there's a partial selection of child items or, worse, a top-level item and a child item of another item, the action should be aborted (perhaps with an error message that says that child items can't be removed from collections).

Creating #681 to fix that.

(Note that that behavior hasn't changed, though — right now it does the same thing, just without a prompt, so this is already an improvement.)

PR #673, opened 2015-03-12
Add an option to not show the proxy redirection notification

LinuxMercedes:
I find that the notification you get when Zotero automatically redirects you through a proxy to get a little annoying, especially when I'm opening a bunch of tabs from Google Scholar or the like. 

Unfortunately, I only speak English, so locale translations are a bit lacking in variety.

### DIFF
--- a/chrome/content/zotero/xpcom/proxy.js
+++ b/chrome/content/zotero/xpcom/proxy.js
@@ -69,6 +69,8 @@ Zotero.Proxies = new function() {
 		Zotero.Proxies.lastIPCheck = 0;
 		Zotero.Proxies.lastIPs = "";
 		Zotero.Proxies.disabledByDomain = false;
+
+		Zotero.Proxies.showRedirectNotification = Zotero.Prefs.get("proxies.showRedirectNotification");
 	}
 	
 	/**
@@ -107,7 +109,7 @@ Zotero.Proxies = new function() {
 				if(!bw) return;
 				_showNotification(bw,
 					Zotero.getString('proxies.notification.associated.label', [host, channel.URI.hostPort]),
-					"settings", function() { _prefsOpenCallback(bw[1]) });
+					[{ label: "proxies.notification.settings.button", callback: function() { _prefsOpenCallback(bw[1]); } }]);
 			}
 		} else {
 			// otherwise, try to detect a proxy
@@ -132,7 +134,7 @@ Zotero.Proxies = new function() {
 					// Ask to save only if automatic proxy recognition is on
 					savedTransparent = _showNotification(bw,
 						Zotero.getString('proxies.notification.recognized.label', [proxy.hosts[0], channel.URI.hostPort]),
-						"enable", function() { _showDialog(proxy.hosts[0], channel.URI.hostPort, proxy) });
+						[{ label: "proxies.notification.enable.button", callback: function() { _showDialog(proxy.hosts[0], channel.URI.hostPort, proxy); } }]);
 				}
 				
 				proxy.save();
@@ -234,9 +236,15 @@ Zotero.Proxies = new function() {
 		
 		// Otherwise, redirect. Note that we save the URI we're redirecting from as the
 		// referrer, since we can't make a proper redirect
-		_showNotification(bw,
-			Zotero.getString('proxies.notification.redirected.label', [channel.URI.hostPort, proxiedURI.hostPort]),
-			"settings", function() { _prefsOpenCallback(bw[1]) });
+		if(Zotero.Proxies.showRedirectNotification) {
+			_showNotification(bw,
+				Zotero.getString('proxies.notification.redirected.label', [channel.URI.hostPort, proxiedURI.hostPort]),
+				[
+					{ label: "general.dontShowAgain", callback: function() { _disableRedirectNotification(); } },
+					{ label: "proxies.notification.settings.button", callback: function() { _prefsOpenCallback(bw[1]); } }
+				]);
+		}
+
 		browser.loadURIWithFlags(proxied, 0, channel.URI, null, null);
 	}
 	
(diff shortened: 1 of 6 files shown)


dstillman:
I'll let @simonster review this, but if this remains a visible pref, only the en-US string should actually be included. We have scripts to merge non-English strings, which will get run anyway to avoid accidental breakage, so it's better to keep the commits small. So you'll need to rebase without those non-en-US changes and force push.

chrome/content/zotero/xpcom/proxy.js
```diff
@@ -69,6 +69,8 @@ Zotero.Proxies = new function() {
 		Zotero.Proxies.lastIPCheck = 0;
 		Zotero.Proxies.lastIPs = "";
 		Zotero.Proxies.disabledByDomain = false;
+
+		Zotero.Proxies.showRedirectNotification = Zotero.Prefs.get("proxies.showRedirectNotification");
```
dstillman: You should either update this when the pref changes — which you can do with a handler in ``Zotero.Prefs`` in zotero.js — or just check the pref value directly below. Otherwise this won't change until Zotero is restarted.

(It's possible this is all premature optimization and the latter would be fine, but I understand that the existing code does this. Not sure if anyone has tested repeated pref lookup vs. property access speed.)

LinuxMercedes: I was under the impression that [this line](https://github.com/LinuxMercedes/zotero/blob/optional-proxy-redirect-notification/chrome/content/zotero/preferences/preferences_proxies.js#L48) ensures that the variables set in the Proxies initialization are kept up to date with the settings page. 

I can change it so that it just checks the pref value directly, or, if this should really have a handler for pref updates, I'd be glad to set that up.

LinuxMercedes:
Extra 'translations' removed.

chrome/content/zotero/xpcom/proxy.js
dstillman: Ah, sorry — you're right.

aurimasv:
`proxies.showRedirectNotification` needs to default to `true` in prefs.js. We don't want the notification to be disabled by default.

Not necessary, but would be neat to add another button to the notification bar that says something like "Don't notify again" (I think Firefox uses some sort of standard text for this), which would toggle the pref without going into settings. The latter is easy enough though.

LinuxMercedes:
@aurimasv I'm not sure where to default `proxies.showRedirectNotification`. I grepped for `prefs.js` but I didn't find any file named that, and `preferences.js` doesn't turn up anything enlightening.

aurimasv:
My mistake. [defaults/preferences/zotero.js](https://github.com/zotero/zotero/blob/4.0/defaults/preferences/zotero.js)

LinuxMercedes:
I'm not sure it's what firefox defaults to, but I added a message that seemed to make sense and isn't too long.

LinuxMercedes:
Those builds failed because of an oxford comma, if I'm reading the output right?!

dstillman:
Yeah, tests results on this branch aren't reliable at the moment — disregard.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -898,6 +898,7 @@ proxies.notification.associated.label	= Zotero automatically associated this sit
 proxies.notification.redirected.label	= Zotero automatically redirected your request to %1$S through the proxy at %2$S.
 proxies.notification.enable.button		= Enable…
 proxies.notification.settings.button	= Proxy Settings…
+proxies.notification.hide.button      = Don't Show Again
```
dstillman: Let's actually make this `general.dontShowAgain` above, since this will probably come in handy elsewhere. And this should be a proper typographical apostrophe.

chrome/locale/en-US/zotero/preferences.dtd
```diff
@@ -144,6 +144,7 @@
 <!ENTITY zotero.preferences.proxies.desc_after_link 		"for more information.">
 <!ENTITY zotero.preferences.proxies.transparent				"Enable proxy redirection">
 <!ENTITY zotero.preferences.proxies.autoRecognize			"Automatically recognize proxied resources">
+<!ENTITY zotero.preferences.proxies.showRedirectNotification			"Show notification when automatically redirecting through a proxy">
```
dstillman: We can drop the "automatically" here. That's just how the feature works, but it's not relevant to this option.

chrome/locale/en-US/zotero/zotero.properties
LinuxMercedes: Hmm, that will make https://github.com/zotero/zotero/pull/673/files#diff-72b2d297070644abfa5ac6c878f83f0dR434 less pretty, but it makes sense. Do you mind if I call it `general.dontShowAgain.button` or should we just give up all pretense of hiding the resource string values?

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -55,6 +55,7 @@ general.numMore                    = %S more…
 general.openPreferences				= Open Preferences
 general.keys.ctrlShift				= Ctrl+Shift+
 general.keys.cmdShift				= Cmd+Shift+
+general.dontShowAgain.button      = Don't Show Again
```
aurimasv: This should just be `general.dontShowAgain`

chrome/content/zotero/xpcom/proxy.js
```diff
@@ -417,26 +425,37 @@ Zotero.Proxies = new function() {
...
-	  * @param	{Function}	callback	callback to be executed if button is pressed
+	  * @param	{Array}	  buttons		dicts of button label ("settings", "enable", or "hide") and associated callback
 	  */
-	 function _showNotification(bw, label, button, callback) {
+	 function _showNotification(bw, label, buttons) {
 	 	var browser = bw[0];
 	 	var window = bw[1];
-	 	
+
+		buttons = buttons.map(function(button) {
```
aurimasv: Not sure why we need this. Just set the label correctly to begin with.

chrome/locale/en-US/zotero/zotero.properties
dstillman: Yeah, the point of putting this in `general.` is so that it's usable for, say, a checkbox label. So you'll just need to tweak the code that requests this.

chrome/content/zotero/xpcom/proxy.js
```diff
@@ -417,26 +425,37 @@ Zotero.Proxies = new function() {
...
 			browser.removeEventListener("pageshow", listener, false);
 		}
 		
 		browser.addEventListener("pageshow", listener, false);
 	 }
+
+	 /**
+		* Hides proxy redirection notification
+		*/
+	 function _hideRedirectNotification() {
```
aurimasv: "Disable" rather than "hide" would be more appropriate here.

chrome/content/zotero/xpcom/proxy.js
LinuxMercedes: This is in the spirit of how the code worked before I started messing with it; I personally think it does cut down on having a bunch of redundant code in places that are already pretty verbose.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -55,6 +55,7 @@ general.numMore                    = %S more…
 general.openPreferences				= Open Preferences
 general.keys.ctrlShift				= Ctrl+Shift+
 general.keys.cmdShift				= Cmd+Shift+
+general.dontShowAgain = Don't Show Again
```
dstillman: Apostrophe should be a typographic quote (’)

dstillman:
Other than the quote mark, this looks good, so you can fix that and squash.

LinuxMercedes:
Done.

dstillman:
Thanks!

PR #678, opened 2015-03-17
Making splitter and grippy more visible if collapsed

zuphilip:
I am trying to respond to the frequent question in the forum when the right pane is collapsed. My idea is to 
 * add tooltip on splitters (always showing up)
 * double the size of the splitter if collapsed
 * double the size of the grippy and center it if splitter is collapsed

splitter-styling
Let me know what you think about the approach. I guess the actual coding would need to be improved (different css, use localized strings for tooltips, ...).

### DIFF
--- a/chrome/content/zotero/zoteroPane.xul
+++ b/chrome/content/zotero/zoteroPane.xul
@@ -329,7 +329,7 @@
 				</vbox>
 				
 				<splitter id="zotero-collections-splitter" resizebefore="closest" resizeafter="closest" collapse="before"
-					zotero-persist="state"
+					zotero-persist="state" tooltiptext="Resize left pane"
 					onmousemove="document.getElementById('zotero-items-toolbar').setAttribute('state', this.getAttribute('state'));ZoteroPane_Local.updateToolbarPosition();"
 					oncommand="ZoteroPane_Local.updateToolbarPosition()">
 					<grippy id="zotero-collections-grippy"/>
@@ -538,7 +538,7 @@
 					</deck>
 				</vbox>
 				
-				<splitter id="zotero-items-splitter" resizebefore="closest" resizeafter="closest" collapse="after" zotero-persist="state"
+				<splitter id="zotero-items-splitter" resizebefore="closest" resizeafter="closest" collapse="after" zotero-persist="state" tooltiptext="Resize right pane"
 					onmousemove="ZoteroPane_Local.updateToolbarPosition()"
 					oncommand="ZoteroPane_Local.updateToolbarPosition()">
 					<grippy id="zotero-items-grippy"/>
--- a/chrome/skin/default/zotero/overlay.css
+++ b/chrome/skin/default/zotero/overlay.css
@@ -48,6 +48,17 @@
 	height: .25em;
 }*/
 
+#zotero-pane splitter[state="collapsed"] > grippy
+{
+	width: 14px;
+	background-position: 50% 50%;
+}
+
+#zotero-pane splitter[state="collapsed"]
+{
+	width: 20px;
+}
+
 #zotero-pane splitter
 {
 	border: 0;


gracile-fr:
There are, I think, two different issues.
1- many users are confused when the right pane is collapsed. They just don't have the right information.
2- it is somewhat difficult to expand it (at least on Windows 7) and this is very annoying. It's easier to get the double-arrow icon to resize rather than the hand icon (=expand). That should be the contrary, as when the pane is collapsed, you're more likely to open it completely rather than resize it. I don't really like having a huge splitter, but I don't have any other solution at the moment.

What is the difference between your 2nd and 3rd solution?

aurimasv:
I'll just throw this in here.
(1) Do we need to have click-to-collapse functionality? It's only useful if you need to toggle between collapsed and expanded states frequently, which may or may not be a reasonable assumption (for me, it's not). Otherwise, it's fairly trivial to just drag the pane to collapse/expand (granted that it's sometimes difficult to position the mouse correctly, but that's the same case for the grippy). Having the grippy positioned so close to the scrollbar and being prone to accidental clicks is probably the reason we're seeing so many people getting confused. Though, I should say that the grippy probably does provide some increased accessibility for people that have issues dragging stuff (we've seen some reports of people having problems dragging items to different collections).

(2) Do we need the panes to have a minimum width after which they end up getting collapsed? I'm not sure what's the problem that we're trying to solve with this functionality. If the user wants to decrease the pane to a point where some text or controls are out of view, that's the user's choice. The current minimum width is arbitrary and doesn't prevent things from moving out of view anyway (especially in some non-english locales). Moving the divider all the way to the left or right is not a difficult task, so having the auto-collapse at minimum width doesn't seem to introduce any benefit and could confuse some users when the pane disappears.

Even given above, I think better visibility for the collapsed panes is good, bit I personally would prefer to go the Citavi route and add text labels instead of just showing the expand arrows (that also looks different on Mac OS X)

dstillman:
> Having the grippy positioned so close to the scrollbar and being prone to accidental clicks is probably the reason we're seeing so many people getting confused.

I kind of doubt that. You can't even click to collapse on OS X, and we're getting lots of reports there. (I'm actually curious whether we're as many reports on Windows, since the collapsed grippy is considerably more visible on Windows than on OS X.) I think the collapsing on min-width is much more likely as a reason, and maybe some condition under which it's closing automatically (though I'm not sure what). The grippy is built-in functionality, it's handy if you know about it, it's more accessible, and I don't see any reason to disable it if we're going to have placeholders at all.

> Do we need the panes to have a minimum width after which they end up getting collapsed?

I think we need some min-width, for the same reason that most software applies a min-width for sidebars — because we have an interest in preventing Zotero from being configured in ways that are unusable and look awful. But we should see if we can make it collapse at a much narrower width than the min-width. If we can't, I'd favor just doing away with the drag collapsing and switching to a View menu in Standalone and a toolbar in Firefox for hiding/showing the panes. For an example of a toolbar button menu, here's Preview on OS X:

preview

I'd rather do this if there were other view-related things we wanted to make configurable, though, since otherwise it's rather valuable toolbar space to give up.

Otherwise, wider grippies when collapsed are probably the best option, but they don't look great and are obviously a waste of space. If we do go that route, note that it's different across platforms, so this would have to be made Windows- (and maybe Linux-) specific and we'd need a different version for OS X.

I'm highly skeptical that text labels can be made to look good. Certainly the screenshot a saw from Citavi is not something I want to replicate.

dstillman:
BTW, the splitter on OS X when collapsed is essentially invisible:

osxsplitter

(Looks real nice, though!)

dstillman:
I'm going to put out a hotfix release that either reverts right-pane collapsing altogether or at least makes the collapsed state not persist across restarts. I'm leaning toward the latter, which I hope would be enough for confused folks to at least get their panes back, but not sure. In any case, we're getting multiple posts a day about this, so we can't really wait for a better solution.

In all (or nearly all) forum posts about this where a platform was mentioned, it was OS X, but I'm going to make this change for all platforms for now. (For what it's worth, I'm having trouble getting a wider splitter on OS X to not look terrible.)

zuphilip:
The splitter on Mac is using `chrome://zotero/skin/mac/vgrippy.png` as coded here https://github.com/zotero/zotero/blob/4.0/chrome/content/zotero-platform/mac/overlay.css#L367 where the splitter in Window is using `chrome://global/skin/splitter/grip-right.gif` which is 4px x 115px large.

@gracile-fr It is an AND between the different points, i.e. you are already seeing all three points in the picture.

@aurimasv I tried (shortly) to come up with something similar to Citavi, but I didn't look appealing. I guess one would need to spend more time (and understanding) into that route.

@dstillman Every alternative solution you suggested could help the users, just the reverting sounds a little extreme.

dstillman:
We've pushed out 4.0.26.2 for Firefox that reopens the collapsed item pane on restart. A Standalone version will probably be out later tonight.

So we'll want to figure out a better solution, but this makes it a little bit less critical.

bwiernik:
For a View menu, perhaps moving the Column Display picker there, rather than in its current location? Might make discovery slightly easier and conceptually makes sense.

zuphilip:
The problems seemed to disappear with the fix that reopens the collapsed item pane on restart, at least I don't read any problems in the forum. Thus, it would be okay for me close this issue here.

dstillman:
Yeah, I don't think we've had any reports about the item pane since this change. We did have one for the collections pane the other day, but we get almost no reports of that in general, so I'm not sure it's worth doing the same there (and people could plausibly only want to deal with their own library root, I guess)...

PR #679, opened 2015-03-17
Fix error handling when saving snapshots

aurimasv:
Partly addresses https://forums.zotero.org/discussion/47206/bbc-translator-does-not-store-snapshot/ but still saves an empty document. The latter is addressed by #666


aurimasv:
Well, actually, it doesn't save an empty document because of the last commit there. It doesn't save attachment at all.

PR #680, opened 2015-03-21
No prompt for remove if no top-level items are selected

rmzelle:
As requested at
https://github.com/zotero/zotero/pull/668#issuecomment-84090670

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -1610,7 +1610,13 @@ var ZoteroPane = new function()
 			var prompt = (force && !fromMenu) ? false : toTrash;
 		}
 		else if (itemGroup.isCollection()) {
-			// In collection, only prompt if trashing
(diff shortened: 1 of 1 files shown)


chrome/content/zotero/zoteroPane.js
```diff
@@ -1610,7 +1610,13 @@ var ZoteroPane = new function()
 			var prompt = (force && !fromMenu) ? false : toTrash;
 		}
 		else if (itemGroup.isCollection()) {
-			// In collection, only prompt if trashing
+			
+			// Don't prompt on remove when no top-level items are selected
```
dstillman: This comment is a little confusing. It's not not prompting on remove — it's not removing in the first place.

How about "Ignore unmodified action if only child items are selected"?

rmzelle: Uhh, right. Amended the commit.

PR #683, opened 2015-03-24
Allow setting item field value via base field

aurimasv:

PR #684, opened 2015-03-24
Add ability to load sample item data from JSON

aurimasv:
This is in preparation for writing unit tests for citeproc-json changes. Wanted to get an idea on whether this would be acceptable.

Depends on #683


aurimasv:
I merged this into #659, since that PR actually takes advantage of the sample data.

PR #691, opened 2015-03-30
Trim down ISSN/ISBN test for characters that should not be ignored

aurimasv:


aurimasv:
Re https://github.com/zotero/zotero/pull/660#issuecomment-87828236

simonster:
Thanks!

Issue #619, opened 2015-01-21
Retrieve metadata via Google Scholar broken in Firefox 31-34


aurimasv:
Re https://forums.zotero.org/discussion/45809/cant-retrieve-metadata-for-pdf-an-unexpected-error-occurred/

I can reproduce this locally with ESR
```
(4)(+0000000): Translate: Parsing code for Google Scholar
(3)(+0000002): Translate: Beginning translation with Google Scholar
(5)(+0000002): Translate: Running handler 0 for select
(3)(+0000000): Translate: Attempting to set cookie through GS Settings page
(5)(+0000000): Translate: TypeError: doc.location is null in handler 0 for select
(1)(+0000001): TypeError: doc.location is null
    ===== Stack Trace =====
    setCookieThroughPrefs@C:\Users\Aurimas\Desktop\dev\profiles\esr\zotero\translators\Google Scholar.js:80:2
    prepareCookie@C:\Users\Aurimas\Desktop\dev\profiles\esr\zotero\translators\Google Scholar.js:117:3
    doWeb/<@C:\Users\Aurimas\Desktop\dev\profiles\esr\zotero\translators\Google Scholar.js:544:4
    doWeb@C:\Users\Aurimas\Desktop\dev\profiles\esr\zotero\translators\Google Scholar.js:512:3
    =======================
```
Sample PDF http://goo.gl/xJTlNv

zuphilip:
It worked for me on
<code>version => 4.0.24.1.SOURCE, platform => Win32, oscpu => Windows NT 6.1; WOW64, locale => en-US, appName => Firefox, appVersion => 34.0.5</code>

PR #692, opened 2015-03-31
Add Firefox Beta for Travis-CI testing

aurimasv:
Also makes it unnecessary to adjust Firefox version numbers for each release

Disables recognizePDF test for PDFs with no DOIs, because I think they start failing due to GS captcha. It probably clears up once there are no recent commits.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,15 +2,22 @@ language: cpp
 compiler:
     - gcc
 env:
-    matrix:
-        - FIREFOXVERSION="36.0.1"
-        - FIREFOXVERSION="31.5.0esr"
+    - FX_PATH_SUFFIX=""
+    - FX_PATH_SUFFIX="-esr"
+    - FX_PATH_SUFFIX="-beta"
+#    - FX_PATH_SUFFIX="-prior-esr"
+matrix:
+    fast_finish: true
+    allow_failures:
+        - env: FX_PATH_SUFFIX="-beta"
+        - env: FX_PATH_SUFFIX="-prior-esr"
 notifications:
     email: false
-before_install:
+install:
+    - wget -r -l1 -np -nd -R "index.html*" "http://ftp.mozilla.org/pub/mozilla.org/firefox/releases/latest${FX_PATH_SUFFIX}/linux-x86_64/en-US/?F=0&P=*.tar.*"
+    - tar -xf firefox-*.tar.*
+before_script:
     - export DISPLAY=:99.0
     - sh -e /etc/init.d/xvfb start
-    - wget http://ftp.mozilla.org/pub/firefox/releases/${FIREFOXVERSION}/linux-x86_64/en-US/firefox-${FIREFOXVERSION}.tar.bz2
-    - tar -xjf firefox-${FIREFOXVERSION}.tar.bz2
 script:
     - test/runtests.sh -x firefox/firefox
--- a/test/runtests.sh
+++ b/test/runtests.sh
@@ -77,9 +77,13 @@ user_pref("extensions.zotero.firstRunGuidance", false);
 user_pref("extensions.zotero.firstRun2", false);
 EOF
 
-
+# -v flag on Windows makes Firefox process hang
 if [ -z $IS_CYGWIN ]; then
 	echo "`MOZ_NO_REMOTE=1 NO_EM_RESTART=1 \"$FX_EXECUTABLE\" -v`"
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -2,15 +2,21 @@ language: cpp
...
+        - env: FX_PATH_SUFFIX="-beta"
+        - env: FX_PATH_SUFFIX="-prior-esr"
 notifications:
     email: false
 before_install:
     - export DISPLAY=:99.0
     - sh -e /etc/init.d/xvfb start
-    - wget http://ftp.mozilla.org/pub/firefox/releases/${FIREFOXVERSION}/linux-x86_64/en-US/firefox-${FIREFOXVERSION}.tar.bz2
-    - tar -xjf firefox-${FIREFOXVERSION}.tar.bz2
+    - wget ftp://ftp.mozilla.org/pub/mozilla.org/firefox/releases/latest${FX_PATH_SUFFIX}/linux-x86_64/en-US/firefox-*.tar.bz2
```
dstillman: This was HTTP on purpose, so that Travis could potentially cache it. (FTP caching seems less likely.) We could make a separate GET request to the directory index to get the version number, store that in a variable, and then request that via HTTP without the wildcard.

test/runtests.sh
```diff
@@ -77,6 +77,11 @@ user_pref("extensions.zotero.firstRunGuidance", false);
 user_pref("extensions.zotero.firstRun2", false);
 EOF
 
+
+if [ -z $IS_CYGWIN ]; then
+	echo "`MOZ_NO_REMOTE=1 NO_EM_RESTART=1 \"$FX_EXECUTABLE\" -v`"
+fi
+
```
dstillman: Is this supposed to be here?

test/tests/recognizePDF.js
```diff
@@ -37,7 +37,8 @@ describe("PDF Recognition", function() {
 		});
 	});
 
-	it("should recognize a PDF without a DOI", function() {
+	// Failing, probably due to captchas after frequent commits
+	it.skip("should recognize a PDF without a DOI", function() {
```
dstillman: We don't want to disable this altogether, though. If anything, we should set a variable when we're running within Travis and only skip in that case.

.travis.yml
dstillman: e.g., ```curl -s 'http://ftp.mozilla.org/pub/mozilla.org/firefox/releases/latest-beta/linux-x86_64/en-US/' | sed -n -E 's/.+firefox-([0-9\.b]+\.tar\.[a-z0-9]+).+/\1/p'```

test/tests/recognizePDF.js
simonster: Yeah, this is an important code path to test, e.g. #619. Maybe the best approach would be to detect if we actually hit the captcha in the test and if so to skip it but print something to the console.

.travis.yml
aurimasv: Best I can tell from reading the docs and related issues on github, at this point there is no HTTP caching on Travis (.com or .org). I've actually never seen it cache the download in all previous builds.

There is APT caching, which is currently only available for .com. Unfortunately, while I was able to get APT installation working for stable, beta, and auroras, the PPAs do not seem to be kept up to date as strictly (especially for aurora). Still seems ok for the current setup, so we could go that route. As I said, though, it's not yet enabled on .org.

test/runtests.sh
aurimasv: Yes, though I seem to have misplaced a comment. `-v` doesn't work on Windows. It just hangs the Fx process. Otherwise, it's pretty useful to know which version exactly the test is running on, since the version is not hard coded.

test/tests/recognizePDF.js
aurimasv: I agree, this was a bit of a quick and dirty solution to not have tests fail on Travis right now. Will play around with this some more.

.travis.yml
dstillman: I wasn't saying that they do any caching right now — just that HTTP might be a better protocol to use, since Travis could just set an environment variable pointing to a caching proxy and it would work.

aurimasv:
How's this?

aurimasv:
OK. Done. Let me know if it's ok to squash

dstillman:
Yup, thanks.

aurimasv:
Done

PR #706, opened 2015-04-28
Switch Retrieve Metadata via GS to use `processDocuments`

aurimasv:
This way we should be able to set cookie... but it's not working.

GS translator was switched to setting the BibTeX setting via cookie only due to breakage on Google Scholar (website) side. Previously, we relied on setting the BibTeX setting via GS preferences, which made this work via Retrieve Metadata using XMLHTTPRequest.

With this switch to processDocuments, I'm still seeing that the cookie is not being set.

@dstillman @simonster any thoughts?

I figured this would be an easier fix, but it looks like that's not going to be the case. I'll probably just rewrite GS translator to use the Cite link and fetch a magic value from there before fetching BibTeX

### DIFF
--- a/chrome/content/zotero/recognizePDF.js
+++ b/chrome/content/zotero/recognizePDF.js
@@ -672,17 +672,22 @@ var Zotero_RecognizePDF = new function() {
 			return (delay > 0 ? Q.delay(delay) : Q())
 			.then(function() {
 				Zotero.HTTP.lastGoogleScholarQueryTime = Date.now();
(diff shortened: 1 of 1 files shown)


aurimasv:
No longer relevant because of https://github.com/zotero/translators/commit/d7baf5b771806285b3fab3f8ffca68bd3543b167 but I'm still not entirely clear on why cookies cannot be manipulated when the translator is called from Retrieve Metadata vs URL bar icon on Google Scholar page.

Issue #699, opened 2015-04-14
Mock external services in tests


dstillman:
Our Travis tests are currently failing [all the time](https://travis-ci.org/zotero/zotero/builds) for reasons that have nothing to do with the code under test. Right now, for example, [we're failing](https://travis-ci.org/zotero/zotero/jobs/58499872) because of a mismatch in a title returned from an ISBN lookup. That was caused by [a translator issue](https://github.com/zotero/translators/commit/81fd1c9a99ef30b51603261db95698adc4aa5661), and it's at best distracting and at worst unnecessarily alarming to have it showing up on random commits and pull requests. I'd say we want to avoid external services almost entirely in Travis and just have them for local test runs, which the main developers should be doing all the time. In my view, Travis tests should almost always be passing on the main branch, with failures limited to unexpected side effects and untested pull requests.

We can create a utility function using [Sinon](http://sinonjs.org/) to mock ``Zotero.HTTP.request()`` for various things, but I'm not sure how we want to handle things that depend on translators, and particularly a case like this where it wasn't even that the LoC MARC response changed, but that a different site was being used entirely.

@simonster, @aurimasv, thoughts?

dstillman:
I also think we should disable translator/style updates from the repository in Travis, so that failures are tied to specific commits. That means that a bug like this wouldn't show up until we bumped the translator submodule, which would be the relevant commit. (This one actually wouldn't have shown up for that, because it was specific to repository behavior, but as long as repo updates are enabled for local test runs, a test that checked which translator was used for a particular ISBN lookup would've caught this bug locally, even if the actual external response was mocked.)

aurimasv:
I would like to keep Travis tests as close as possible to local tests (unless technically impossible, e.g. solving CAPTCHA) and as close as possible to what the actual client would do. IMO Travis tests are there to help fool-proof commits/PRs, so saying that this part of the software we'll ask the developers to test locally doesn't sound that great to me. Basically, if the tests pass, then, if Zotero were to be released from that commit, all its functions should behave as we expect them. Obviously easier said than done.

[...]

PR #707, opened 2015-04-29
Prep test framework for testing on zotero/translators

aurimasv:
For now we just test Add by Identifier and Retrieve Metadata.

https://travis-ci.org/aurimasv/translators/builds/60495194

Re #699


aurimasv:
@dstillman, do you have any concerns?

chrome/content/zotero/xpcom/schema.js
```diff
@@ -477,26 +479,37 @@ Zotero.Schema = new function(){
...
-					if (Zotero.Prefs.get('automaticScraperUpdates')) {
-						Zotero.proxyAuthComplete
-						.then(function () {
-							Zotero.Schema.updateFromRepository(2, function () deferred.resolve());
-						})
-					}
-				}
-				else {
-					Zotero.proxyAuthComplete
+			if (!runRemoteUpdateWhenComplete) return;
```
dstillman: I know this way is much cleaner, but this is all rewritten on api_syncing, so this will just cause conflicts. Anything that's not strictly necessary should stay unchanged. (4.0 should be considered close to frozen at this point.)

test/runtests.sh
```diff
@@ -29,12 +28,17 @@ Options
...
+ -t                  run only translator-dependent tests
  TESTS               set of tests to run (default: all)
 DONE
 	exit 1
 }
 
-while getopts "x:dc" opt; do
+DEBUG=false
+UPDATE_TRANSLATORS=true
+TRANSLATOR_TESTS=false
```
dstillman: We don't want to disable translator tests on zotero/zotero, though. We just don't want them to run on Travis until we've mocked external services, and we want automatic translator updating disabled in Travis and on zotero/translators.

chrome/content/zotero/xpcom/schema.js
aurimasv: Do you have a ballpark timeline for 5.0 (at least beta)? Maybe it's worth just waiting until then to get tests running on zotero/translators.

test/runtests.sh
aurimasv: Why do we want to disable translator updates on zotero/zotero? Shouldn't we be testing with what the users are running?

chrome/content/zotero/xpcom/schema.js
dstillman: We'll have some sort of release — probably not beta, but possibly "Alpha" or "Preview" — by the end of the month. So yeah, it's probably easier to just do this against that.

aurimasv: Ok. Great! We'll wait until then to fix this up. I guess I need to get feed reader merged in before end of the month.

test/runtests.sh
```diff
@@ -60,6 +68,10 @@ else
 	TESTS="$(join , "${ARGS[@]}")"
 fi
 
+if [ "$TRANSLATOR_TESTS" = true ]; then
+	TESTS="translatorDependentFeatures"
```
dstillman: I don't want to move all translator-related tests into a single file. Lookup tests should stay in lookup.js and be runnable via 'lookup'. I think we can either do something with a subdirectory or just keep a hard-coded list of translator-related files in runtests.js. (We'll just need some mechanism for excluding those.)

test/runtests.sh
dstillman: Only disabled for zotero/zotero in Travis. Explained in https://github.com/zotero/zotero/issues/699#issuecomment-94552471

PR #709, opened 2015-05-02
Don't report translator failures from tests

aurimasv:


aurimasv:
Or, I guess, we could disable this on a per-test basis.

PR #710, opened 2015-05-02
Map from baseID to item type fieldID in itemToCSLJSON

fbennett:
While testing new code for legal support in MLZ in a branch based on current 4.0, I found that base-named fields (caseName, reporterVolume, firstPage) were not coming through to the processor. On inspection, it seems that are a couple of problems in `itemToCSLJSON()`.

**`ItemFields.getName()`**

The call to `getName()` at line 1555 should be from `Zotero.ItemFields` to get a fieldID.

**`getFieldIDFromTypeAndBase()`**

* From `item.toArray()`, we get a JS object with the item-specific mapped field names as keys.
* In `CSL_TEXT_MAPPINGS`, base names are used for the Zotero-side field mappings.

To use `CSL_TEXT_MAPPINGS` against the object we get from `item.toArray()`, we need to map from the base name to the item type's field name, not the other way around. This pull request runs the mapping in that direction, and the base-named fields come through.

I haven't tested in Zotero proper, and if I've missed something please forgive the traffic; but from reading the code, this looks right.

### DIFF
--- a/chrome/content/zotero/xpcom/utilities.js
+++ b/chrome/content/zotero/xpcom/utilities.js
@@ -1549,10 +1549,10 @@ Zotero.Utilities = {
 					value = zoteroItem[field];
 				} else {
 					var fieldID = Zotero.ItemFields.getID(field),
-						baseMapping;
-					if(Zotero.ItemFields.isValidForType(fieldID, itemTypeID)
-							&& (baseMapping = Zotero.ItemFields.getBaseIDFromTypeAndField(itemTypeID, fieldID))) {
-						value = zoteroItem[Zotero.ItemTypes.getName(baseMapping)];
+						fieldMapping;
+					if((fieldMapping = Zotero.ItemFields.getFieldIDFromTypeAndBase(itemTypeID, fieldID))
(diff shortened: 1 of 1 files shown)


dstillman:
I imagine this is obviated by (or at least conflicts with) https://github.com/zotero/zotero/pull/659, but @aurimasv can confirm.

fbennett:
Ah, true. The [sample](https://github.com/zotero/zotero/pull/659#issuecomment-76664106) posted by @aurimas shows `proceedingsTitle`, though, which would map to base name `publicationTitle`. I'll keep an eye on developments around #659.

fbennett:
Nope, I see it - the sample I linked to above was early in the discussion. It looks like the new-look `itemToExportFormat()` maps to base names. Will sit tight and look forward to further developments.

aurimasv:
I think this is addressed. @Bennett, you can take a look at https://github.com/zotero/zotero/pull/659/files#diff-416fa9a7f56dfe0b3a5afd2238dfb317 to see what's currently getting exported to citeproc-js

fbennett:
Yep - looking forward to the landing. In time, I hope to make myself useful on the test-writing front, but it all looks really good from here.

Issue #487, opened 2014-05-20
add "Export Library" option for right-click context menu for group libraries


adam3smith:
Also export the currently selected library when choosing "Export Library" from Actions menu, ideally put library name at the top of the export dialog.
https://forums.zotero.org/discussion/22217/export-zotero-group-library-to-endnote/#Item_14

retorquere:
If I were to write a patch to add right-click export, against what baseline would you want that? The 4.0 branch, the 4.0.26.4 tag, the master branch?

aurimasv:
4.0 branch

retorquere:
Is there a build procedure beyond zipping things up, and is there a test suite that I should pass?

adam3smith:
you'd typically run straight from the git checkout and not build while coding, but if you do want to build, zipping is all you need to do, though make sure you do have a recursive checkout, since you'll be missing a bunch of components otherwise. 

Simon and Aurimas have been working on a [test suite](https://github.com/zotero/zotero/tree/4.0/test), but I don't think there's anything relevant in there yet, so you can submit without running this.

retorquere:
Crap, I didn't do a recursive clone. Can his still be done after the initial checkout? The patch is ready, just 50-odd lines of code.

aurimasv:
See https://www.zotero.org/support/dev/source_code#running_git_installations for running Zotero from source.

At this point, the test suite coverage is far from complete, but you should still run all currently available tests to make sure that your changes are not breaking anything else. You can run the tests by running `runtests.sh` from the "test" directory of zotero repo. It would be great if you could also write new tests for the features you are submitting, but that is not always trivial to do, so I don't think this will currenlty be enforced.

dstillman:
`git submodule init`
`git submodule update`

aurimasv:
`git submodule update --init --recursive` in one step

retorquere:
OK, done all that (needed to start afresh anyhow as I had done a checkout from the main repo, not a clone), but when I checkout 4.0 and do `zip -r /home/emile/zotero/zotero.xpi . -x *.git*`, I get a sort-of-working Zotero that has no translators loaded. This is without my mods on a different branch, and after I did a "reset translators"; the translators directory remains empty. Possibly related to an error spit out on the console: https://gist.github.com/retorquere/71bac09c0cbd468e94e6

dstillman:
You can't ZIP up a Git checkout. You need to run it directly, as explained in the linked page above.

[...]

Issue #518, opened 2014-08-07
Promisification


dstillman:
As a prelude to API syncing, in a private branch I've been working on switching the data layer to use asynchronous database access, which is necessary to avoid freezing the UI. (Zotero predates asynchronous database access in Mozilla, but synchronous access on the main thread is now strongly discouraged because of its effect on UI responsiveness.) Unfortunately, due to conflicts between synchronous and asynchronous access, it looks like we'll need to switch over all DB queries in Zotero to be async. More unfortunately, async DB access means returning [promises](https://github.com/petkaantonov/bluebird#what-are-promises-and-why-should-i-use-them) from many core data layer methods, which means rewriting essentially the entire codebase to use promises. This seems like a terrible idea that we'll regret later, but we also haven't come up with a better solution.

This will have fairly major effects on the Zotero client ecosystem, so I'm creating this ticket to share some of the issues involved. I'll be pushing my private ```async_db``` branch up to GitHub so that people can start playing with it. (It should go without saying that you shouldn't use this branch with a real database.)

1) I've switched us from the Q promise library (which we were using in various places) to [Bluebird](https://github.com/petkaantonov/bluebird), which based on published tests is much faster than Q and uses much less memory. With ES6 generators, Bluebird is supposedly nearly as fast as callbacks alone.

2) I'm using ES6 generators and yield almost exclusively, since for data layer code they're the most natural way of using promises, resulting in code very similar to the previous synchronous code. Using ES6 generators means we'll have a Fx26 dependency, which is too bad for Fx24 ESR users, but Fx31 ESR is already available.

3) I'm rewriting many existing methods instead of renaming everything as *Async. We can't keep synchronous versions around anyway, and having everything named *Async seems silly. The main benefit would be that we'd get more obvious breakage for calls that hadn't been updated, but since we're only getting run-time errors either way, I don't think it's worth the code ugliness. (There are still separate non-async and async DB methods, but I'll be making those use non-*Async names once all the synchronous methods have been replaced.)

4) All Zotero plugins are going to break. I don't think there's really any pleasant way to deal with this. (As far as I know Mozilla's add-on dependency checking can't help here.) Once we have a firm version number for this release, I think we'll just have to tell plugin authors that they should add in error messages telling people running the old version of the plugin against the new Zotero version to upgrade to a newer version of the plugin and vice versa.

5) I've experimented with a few different approaches to adding promises, with varying implications for existing code:

[...]

PR #712, opened 2015-05-11
export groups (#487)

retorquere:


dstillman:
This would need to be based on libraryID, not specific to groups. (libraryID is null for My Library in 4.0.)

retorquere:
Right, but this would only show up for groups, not for libraries. Library export is handled as it was before.

dstillman:
There will be other types of libraries in 5.0.

dstillman:
And there's also no reason the Export Library on My Library shouldn't run through the same code.

retorquere:
I just figured collections also get a "Export Collection" menu item, and the UI calls it a group, not a library. There's also that group export really does do something else -- library export is inferred by the Zotero code from the absence of limiting filters (groups/collections/items) and the collection-gathering is different for both (but only from a technical POV). Your call though.

chrome/content/zotero/xpcom/translation/translate_item.js
```diff
@@ -786,6 +786,19 @@ Zotero.Translate.ItemGetter.prototype = {
 		this.numItems = this._itemsLeft.length;
 	},
 	
+	"setGroup":function(group, getChildCollections) {
+		this._itemsLeft = Zotero.DB.columnQuery('select itemID from items where libraryID in (select libraryID from groups where groupID = ?) and not itemID in (select itemID from deletedItems)', [group.id]);
```
dstillman: This will all need to be redone to not be group-specific, but there shouldn't be any SQL code in here. It should call Zotero.Items.getAll(), as setAll() does below, but with a libraryID for the second parameter. (I'm fixing the messy argument order in 5.0.)

dstillman:
Well, the UI calls them Group Libraries. And again, there will be different library types in 5.0, so this should work on those too without any new code.

retorquere:
OK, I'll get on that. I'll need some help squashing commits when I'm done -- never did get the hang of that.

dstillman:
No problem.

Note that the existing interfaces likely predate the existence of other libraries, so they can be changed. E.g., you can rename setAll() to setLibrary() and include a libraryID in the call. Zotero.getCollections() also can take a libraryID. (The API here is messy and has generally been cleaned up (e.g., Zotero.Collections.getCollectionsInLibrary()) on the api_syncing branch, which will become 5.0, but translation is currently broken there, which is why I'll allow this on 4.0.)

retorquere:
OK, I have it working, how do I squash the commit?

aurimasv:
Don't worry about squashing for now, just push the additional commits. It makes it easier for us to review. You can squash once everything looks right.

retorquere:
done

dstillman:
This doesn't seem to work. Looks like it's exporting all items, not just those from the selected group.

chrome/content/zotero/fileInterface.js
```diff
@@ -129,7 +131,8 @@ var Zotero_File_Interface = new function() {
 	 */
 	function exportFile() {
```
aurimasv: nit: we should rename this to something like `exportLibrary`, since this name is confusing and inconsistent with functions below. As you're renaming, you can also just use the `this.exportLibrary = function() {...` convention, since that is what Zotero is slowly transitioning to (as opposed to declaring the functions at the top of the object as is done here)

chrome/content/zotero/fileInterface.js
```diff
@@ -76,6 +76,8 @@ Zotero_File_Exporter.prototype.save = function() {
 			translation.setCollection(this.collection);
 		} else if(this.items) {
 			translation.setItems(this.items);
+		} else if(this.libraryID) {
+			translation.setLibraryID(this.LibraryID);
```
aurimasv: lower case

chrome/content/zotero/fileInterface.js
dstillman: That doesn't need to happen for this patch, though. 4.0 isn't the place to make unnecessary changes, since it will just result in lots of conflicts.

retorquere:
Hey, you're right. I hadn't noticed because my test group lib has the same items as the main one. I'll investigate.

chrome/content/zotero/fileInterface.js
```diff
@@ -129,7 +131,8 @@ var Zotero_File_Interface = new function() {
 	 */
 	function exportFile() {
 		var exporter = new Zotero_File_Exporter();
-		exporter.name = Zotero.getString("pane.collections.library");
+		exporter.libraryID = ZoteroPane_Local.getSelectedLibraryID();
```
aurimasv: This could return false if something other than a library is selected (e.g. "Group Libraries" label). I don't think we should be able to get to this point in those circumstances, so you should throw if that's the case.

chrome/content/zotero/fileInterface.js
retorquere: What would you prefer to see lowercased? The "ID"?

aurimasv: `libraryID` (this is why this is not working btw)

retorquere: Ah, I see now, thanks!

retorquere:
Looks like _libraryID is in use in translate.js and gets overwritten. What varname can I use instead?

aurimasv:
I'm guessing it's happening here: https://github.com/zotero/zotero/blob/4.0/chrome/content/zotero/xpcom/translation/translate.js#L1173 The variable name seems appropriate, I just think that you need to make sure to pass the library ID into `translate` at some point. Don't have time right now to give you more guidance though.

chrome/content/zotero/xpcom/translation/translate_item.js
```diff
@@ -1034,4 +1034,4 @@ Zotero.Translate.ItemGetter.prototype = {
 		return obj;
 	}
 }
-Zotero.Translate.ItemGetter.prototype.__defineGetter__("numItemsRemaining", function() { return this._itemsLeft.length });
\ No newline at end of file
+Zotero.Translate.ItemGetter.prototype.__defineGetter__("numItemsRemaining", function() { return this._itemsLeft.length });
```
aurimasv: Please revert this to not pollute git blame

retorquere:
I've gone another route in https://github.com/zotero/zotero/commit/9a0820cca2f300c02da565d6d437460301a3f952

retorquere:
This also fixes the problem that the main library gets exported instead of the group library. It didn't happen before as I was passing around a group in a differently-named variable.

chrome/content/zotero/fileInterface.js
retorquere: done

retorquere: I've also chomped the trailing newlines

chrome/content/zotero/xpcom/translation/translate_item.js
retorquere: done

chrome/content/zotero/fileInterface.js
```diff
@@ -129,7 +131,9 @@ var Zotero_File_Interface = new function() {
...
+		if exporter.libraryID === false throw('No library selected');
```
aurimasv: Are you testing this? I'm fairly certain that's a syntax error. Can't test it right now, so in the off chance that it's not, please use parentheses for conditionals.

aurimasv: Also, `throw new Error("...")` so it generates a stack trace

chrome/content/zotero/xpcom/translation/translate.js
```diff
@@ -2182,15 +2188,21 @@ Zotero.Translate.Export.prototype._prepareTranslation = function() {
...
 		getCollections = configOptions.getCollections || false;
-	if(this._collection) {
-		this._itemGetter.setCollection(this._collection, getCollections);
-		delete this._collection;
-	} else if(this._items) {
-		this._itemGetter.setItems(this._items);
-		delete this._items;
-	} else {
-		this._itemGetter.setAll(getCollections);
+	switch ((this._export || {}).type) {
```
aurimasv: It should be an error to call this without setting `this._export` first (may have to fix that in current code). Likewise, it should be an error to call this with unknown `this._export.type`, so the `default` should throw.

chrome/content/zotero/xpcom/translation/translate.js
```diff
@@ -2182,15 +2188,21 @@ Zotero.Translate.Export.prototype._prepareTranslation = function() {
...
+			this._itemGetter.setCollection(this._export.collection, getCollections);
+			break;
+		case 'items':
+			this._itemGetter.setItems(this._export.items);
+			break;
+		case 'library':
+			this._itemGetter.setAll(this._export.id, getCollections);
+			break;
+		default:
+			this._itemGetter.setAll();
```
aurimasv: You forgot the `getCollections` argument.

chrome/content/zotero/xpcom/translation/translate_item.js
```diff
@@ -786,11 +786,11 @@ Zotero.Translate.ItemGetter.prototype = {
...
 	},
 	
-	"setAll":function(getChildCollections) {
-		this._itemsLeft = Zotero.Items.getAll(true);
+	"setAll":function(libraryID, getChildCollections) {
+		this._itemsLeft = Zotero.Items.getAll(true, libraryID);
 		
 		if(getChildCollections) {
-			this._collectionsLeft = Zotero.getCollections();
+			this._collectionsLeft = Zotero.getCollections(null, false, libraryID);
```
aurimasv: The second parameter here should be `true` (i.e. recursively) as is the case for `setCollection` above. Looks like this was a bug.

chrome/content/zotero/fileInterface.js
aurimasv: Do you mean changing the function name or the second part or both?

chrome/content/zotero/xpcom/translation/translate_item.js
retorquere: I had been wondering about that.

chrome/content/zotero/fileInterface.js
retorquere: Dammit, the extension pointed to a different repo checkout. I've nuked all but one version.

chrome/content/zotero/xpcom/translation/translate.js
retorquere: Alright, I was just trying to be cautious here.

chrome/content/zotero/xpcom/translation/translate.js
retorquere: But that could just be removed as the default is going to throw anyway

retorquere:
OK, latest comments have been addressed. Is there any way in which I can help building up the testing infrastructure? I have really gotten used to automated unit tests.

chrome/content/zotero/fileInterface.js
dstillman: Both. Any remaining patches on 4.0 need to stay as small as possible.

retorquere: So no action on my part?

dstillman: That's right.

retorquere:
Oh and if someone could walk me through the squash, that would be very much appreciated.

chrome/content/zotero/xpcom/translation/translate.js
```diff
@@ -2182,15 +2188,22 @@ Zotero.Translate.Export.prototype._prepareTranslation = function() {
...
 		getCollections = configOptions.getCollections || false;
-	if(this._collection) {
-		this._itemGetter.setCollection(this._collection, getCollections);
-		delete this._collection;
-	} else if(this._items) {
-		this._itemGetter.setItems(this._items);
-		delete this._items;
-	} else {
-		this._itemGetter.setAll(getCollections);
+	if (!this._export) { throw new Error('No export set up'); }
```
aurimasv: you don't need explicit `throw` here. It will throw implicitly from the `switch` if `this._export` is not set.

retorquere: Only if this._export is set, right? Otherwise it will throw a "this._export is null" error. Or is that the desired effect? Easily removed.

dstillman:
> Is there any way in which I can help building up the testing infrastructure? I have really gotten used to automated unit tests.

It's not worth writing tests for most things on the 4.0 branch because most of the API is changing with [promisification](https://github.com/zotero/zotero/issues/518). But the api_syncing branch, which will become 5.0, has [~100 tests](https://github.com/zotero/zotero/tree/api_syncing/test/tests) and counting. That branch isn't quite ready for public development (which is why it's not on master yet), but it will be very soon, after which we'll require tests for most things.

chrome/content/zotero/fileInterface.js
```diff
@@ -76,6 +76,8 @@ Zotero_File_Exporter.prototype.save = function() {
 			translation.setCollection(this.collection);
 		} else if(this.items) {
 			translation.setItems(this.items);
+		} else {
+			translation.setLibraryID(this.libraryID);
```
aurimasv: Let's make sure that `this.libraryID` is set and throw if none of the above are set. Currently, all code sets one of these, but it will probably make debugging a lot easier if we forget at some point. (throw something like "collection, items, or libraryID must be set before calling Zotero_File_Exporter#save()")

aurimasv:
That's all I have right now. If @dstillman has no other comments, you can squash after making these changes. http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html may be helpful. We can be more specific if you need more guidance.

chrome/content/zotero/fileInterface.js
retorquere: How would we distinguish between main library export and libraryID not being set?

chrome/content/zotero/xpcom/translation/translate.js
aurimasv: JavaScript engine throwing "cannot read property '_export' of null" is good enough here, since we get a stack anyway.

chrome/content/zotero/fileInterface.js
retorquere: done.

retorquere: What's the ETA on 5.0 BTW? Should I be scrambling to rewrite BBT?

retorquere:
What should I make of the warning at the top of that gitready page: "Only do this on commits that haven’t been pushed an external repository"?

aurimasv:
It's only relevant if someone else forks the branch in question and does further work on it. That could really mess them up. In case of GitHub topic branches where you are the only committer, that's not relevant.

retorquere:
Can someone verify the squash went OK?

aurimasv:
No, that didn't work. What command exactly did you run?

dstillman:
Before trying something else, I would recommend using git reflog, finding the commit you were at before that rebase (i.e., the last commit you made on this branch), and doing a git reset --hard to that commit.

aurimasv:
I think your local 4.0 branch is out of date. Make sure you either rebase against upstream/4.0 or you pull upstream changes into your local 4.0 branch

dstillman:
(Or just rebase against the first commit in the topic branch that isn't yours, such that you only see your own 8 commits in the interactive list (after doing a hard reset to undo the above). Rebasing against upstream again isn't usually necessary if you checked out the topic branch from the upstream branch recently, and GitHub will tell you on the pull request when it can't do a clean merge.)

retorquere:
Ugh I hate this part of git. I did `git rebase -i HEAD~8`, and squashed all but the top one. I'm sorry but I don't understand the terminology starting from "before trying anything else" on. I have a patch file that has just my changes, at this point it sounds easier to just get a fresh checkout, apply the patch, and push that.

dstillman:
I understand this can be tricky, but we really prefer keeping these things on the original pull requests.

dstillman:
Make a backup of your repo. Then run `git reflog`, and look down the list until you see the 8 commits you had before, which should be before some rebase junk. Find the hash of the most recent commit. Then do a `git reset --hard <that commit>`. Then you should be back where you were before, and you can do a push -f from there. Then let's pause and reevaluate.

dstillman:
(The commit to hard-reset to might be 0b36e99963834025bcc47b97b3adf3b11b519b87, based on the fetch I did right before you did this.)

dstillman:
[Disregard my deleted message. I think I fetched after you rebased.]

retorquere:
OK, I'm there. Should we move this to gitter?

dstillman:
OK, then force-push to get those 8 commits back on this branch.

retorquere:
OK, that should be there now.

aurimasv:
(if you want live chat, you can join zotero's IRC channel irc://irc.freenode.net/zotero-dev )

retorquere:
I don't mind either way. The pull request now says "10 files changed", that cannot be right.

dstillman:
Join zotero-dev and we'll walk you through this.

dstillman:
Thanks!

Issue #265, opened 2013-03-07
Add Edit button in proxies list


dstillman:
http://forums.zotero.org/discussion/28186/

rmzelle:
Would you like to see a similar button for the site-specific Quick Copy settings list? If I hadn't looked at the code, I'd never have figured out that double-clicking an entry allows it to be edited.

dstillman:
Yeah, I was just thinking the same.

rmzelle:
I'm taking this on.

PR #715, opened 2015-05-14
Add Edit button to proxy list

rmzelle:
PR for #265.

### DIFF
--- a/chrome/content/zotero/preferences/preferences_proxies.js
+++ b/chrome/content/zotero/preferences/preferences_proxies.js
@@ -56,6 +56,14 @@ Zotero_Preferences.Proxies = {
 	
 	
 	/**
+	 * Enables UI buttons when proxy is selected
+	 */
+	enableProxyButtons: function () {
+		document.getElementById('proxyTree-edit').disabled = false;
+		document.getElementById('proxyTree-delete').disabled = false;
+	},
+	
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/preferences/preferences_proxies.js
```diff
@@ -56,6 +56,14 @@ Zotero_Preferences.Proxies = {
 	
 	
 	/**
+	 * Enables UI buttons when proxy is selected
+	 */
+	enableProxyButtons: function () {
+		document.getElementById('proxyTree-edit').disabled = undefined;
+		document.getElementById('proxyTree-delete').disabled = undefined;
```
dstillman: I don't think there's any functional difference, but just for convention, these should be `false`, not `undefined`. (You can just commit --amend that and force-push. This looks great otherwise.)

rmzelle: Done. The original code used `undefined`, so I initially kept it. Probably saves a bit or two, but agree that `false` is clearer.

rmzelle:
You can leave the other ticket open so we/I don't forget to add a similar Edit button to the site-specific quick copy settings. I'll work on that once the locale menu PR is merged.

dstillman:
Thanks!

Issue #295, opened 2013-04-10
Right click on URL and DOI fields


mronkko:
Some fields (e.g. Title, Publication) have additional options available on right click (Transform case). The URL and DOI fields could have similar functionality to open the URL or DOI link on the browser.

bwiernik:
I think this was closed by https://github.com/zotero/zotero/pull/716 ?

dstillman:
Apparently. I didn't even remember that we did this.

One minor thing is that we're copying the URL with a percent-encoded DOI. At least in the case of slashes, that's not a problem for doi.org, so we could consider leaving those intact and just encoding other characters. (Or maybe we don't need to encode at all. `#` would break things, but hopefully that's not actually used in any DOIs. Not sure about other characters.)

bwiernik:
Technically, `#` ` ` `%` `"` and `?` are all allowed in DOI names but their use is discouraged and in practice few if any publishers use them. I think we could safely not encode the DOI as part of Copy as URL. 

See https://www.doi.org/doi_handbook/2_Numbering.html#2.5.2.3

PR #716, opened 2015-05-15
Add DOI-field context-menu

rmzelle:
I tried my hand at https://github.com/zotero/zotero/issues/295.

The "Open in Browser" function works, but my attempt to add a "Copy" function failed, and I was wondering if I could get some help. I found the incantation:

```
Components.classes["@mozilla.org/widget/clipboardhelper;1"]
                      .getService(Components.interfaces.nsIClipboardHelper)
                      .copyString("testString");
```

in fileInterface.js, but using it directly via my menuitem's `oncommand` breaks the code. Anything I'm missing?

### DIFF
--- a/chrome/locale/en-US/zotero/zotero.dtd
+++ b/chrome/locale/en-US/zotero/zotero.dtd
@@ -126,6 +126,8 @@
 <!ENTITY zotero.item.textTransform.titlecase			"Title Case">
 <!ENTITY zotero.item.textTransform.sentencecase			"Sentence case">
 <!ENTITY zotero.item.creatorTransform.nameSwap      	"Swap First/Last Names">
+<!ENTITY zotero.item.viewOnline							"View Online">
+<!ENTITY zotero.item.copyAsURL							"Copy as URL">
 
 <!ENTITY zotero.toolbar.newNote				"New Note">
 <!ENTITY zotero.toolbar.note.standalone				"New Standalone Note">
--- a/chrome/content/zotero/bindings/itembox.xml
+++ b/chrome/content/zotero/bindings/itembox.xml
@@ -413,6 +413,13 @@
 								label.setAttribute("isButton", true);
 								label.setAttribute("onclick", "ZoteroPane_Local.loadURI('" + doi + "', event)");
 								label.setAttribute("tooltiptext", Zotero.getString('locate.online.tooltip'));
+								valueElement.setAttribute('contextmenu', 'zotero-doi-menu');
+								
+								var openURLMenuItem = document.getElementById('zotero-doi-menu-view-online');
+								openURLMenuItem.setAttribute("oncommand", "ZoteroPane_Local.loadURI('" + doi + "', event)");
+								
+								var copyMenuItem = document.getElementById('zotero-doi-menu-copy');
+								copyMenuItem.setAttribute("oncommand", "Zotero.Utilities.Internal.copyTextToClipboard('" + doi + "')");
 							}
 						}
 						else if (fieldName == 'abstractNote') {
@@ -2478,6 +2485,10 @@
 							return !hideTransforms;">
 						<menuitem label="&zotero.item.creatorTransform.nameSwap;"
 							oncommand="document.getBindingParent(this).swapNames();"/>
+					</menupopup>
+					<menupopup id="zotero-doi-menu">
+						<menuitem id="zotero-doi-menu-view-online" label="&zotero.item.viewOnline;"/>
+						<menuitem id="zotero-doi-menu-copy" label="&zotero.item.copyAsURL;"/>
 					</menupopup>
 					<zoteroguidancepanel id="zotero-author-guidance" about="authorMenu" position="after_end" x="-25"/>
 				</popupset>
--- a/chrome/content/zotero/xpcom/utilities_internal.js
+++ b/chrome/content/zotero/xpcom/utilities_internal.js
@@ -30,6 +30,16 @@
  * @class Utility functions not made available to translators
  */
 Zotero.Utilities.Internal = {
+	
+	/**
+	 * Unicode normalization
+	 */
+	"copyTextToClipboard":function(str) {
+		Components.classes["@mozilla.org/widget/clipboardhelper;1"]
+			.getService(Components.interfaces.nsIClipboardHelper)
+			.copyString(str);
+	},
+	
 	 /*
 	 * Adapted from http://developer.mozilla.org/en/docs/nsICryptoHash
 	 *


rmzelle:
See also https://github.com/zotero/zotero/blob/7761335bc7f1c9d0aa2daccbfa76efdd318af6c0/chrome/content/zotero/fileInterface.js#L192 . I couldn't find any other existing functions to copy a string to the clipboard.

dstillman:
I don't see any point to a separate context-menu option for copying. That already exists, and exists for every field: you click once on the field (which selects the full text) and then right-click.

chrome/locale/en-US/zotero/zotero.dtd
```diff
@@ -126,6 +127,7 @@
 <!ENTITY zotero.item.textTransform.titlecase			"Title Case">
 <!ENTITY zotero.item.textTransform.sentencecase			"Sentence case">
 <!ENTITY zotero.item.creatorTransform.nameSwap      	"Swap First/Last Names">
+<!ENTITY zotero.item.openURL							"Open in Browser">
```
dstillman: I think "Open in Browser" is a little awkward, among other things because for most people this will already be in Firefox, and it's weird to refer to Firefox as "Browser" from within Firefox.

We could probably just reuse the existing "View Online" string, though that would be a little weirder for the URL field, and seems a bit anachronistic in general. And conversely, "Open URL" would work fine for the URL field, but would be a little weirder for this. Unless we have something obviously better, maybe stick with "View Online" for now, since we already have it?

dstillman: And actually, the tooltip for the DOI and URL labels is already "Go to this item online", so "View Online" is consistent with that.

aurimasv:
>I don't see any point to a separate context-menu option for copying.

That's true for URL, but it's not the same for DOI. I'd like a way to copy DOI as URL. If we're adding a context menu for that, I think it may make sense to make the right-click available on both the DOI label and the field for better discoverability.

dstillman:
> I'd like a way to copy DOI as URL.

Why? DOIs aren't URLs. Just because there's a resolver doesn't mean anyone should be representing them as such.

dstillman:
And if you want the real URL, you can just click through and copy the URL of the loaded page.

aurimasv:
I don't really see why not. If I want to send someone a link to an article, I'd prefer that it were a hyperlink so they can just follow it. No reason to make them type our the resolver prefix.

dstillman:
I guess a "Copy as URL" is fine. I would certainly put it below "View Online", not above. In almost all cases there'll also be a URL in the URL field as well, which seems like a better thing to give someone unless you really want a (slightly) shortened URL.

aurimasv:
>In almost all cases there'll also be a URL in the URL field as well

It's actually not that uncommon to not have a URL and just have a DOI (just looking at my library).

>which seems like a better thing to give someone unless you really want a (slightly) shortened URL

I'll have to disagree with you on the usefulness of a DOI URL, rather than a resolved one. There's a reason DOI was invented.

dstillman:
> I think it may make sense to make the right-click available on both the DOI label and the field for better discoverability.

I think just having it on the field is fine. This way is consistent with our other context-menu options (e.g., we don't have "Transform Text" on the Title label).

aurimasv:
Oh yeah, that's fine. I misread the code. Thought the context menu was on the label.

zuphilip:
Is the DOI-field context-menu also the place where we would want to have the "update metadata by calling the doi translator" function?

rmzelle:
I agree with @aurimasv that resolvable DOI URLs are useful. I can share them with colleagues without having to explain them how to use a DOI resolver, since it's just a link, and they are stable, unlike some journal article URLs.

I'll make the change to use the "locate.online.label" string from zotero.properties for "View Online". What should we use for "Copy as URL"? zotero.item.copyAsURL in zotero.dtd?

Any tips for getting the copy-function to work?

dstillman:
Oh, forgot that View Online is in zotero.properties. Can just duplicate it as `zotero.item.viewOnline`, then, unless we have a better string — I believe Transifex will automatically suggest the other for this string. (It's annoying that these are two different sets of strings, but it's not worth the code contortions to reuse.) And yes, `zotero.item.copyAsURL` is fine.

dstillman:
Re: the clipboard, you have double quotes within double quotes. You can just use single quotes instead, but it's probably better to create a `copyTextToClipboard` function in zoteroPane.js and call that.

dstillman:
Or better yet, `Zotero.Utilities.Internal.copyTextToClipboard`, since this doesn't depend on a window.

rmzelle:
Great, thanks! (those CDATA sections always trip up my IDEs)

rmzelle:
I made the changes you asked for.

chrome/locale/en-US/zotero/zotero.dtd
```diff
@@ -126,6 +126,8 @@
 <!ENTITY zotero.item.textTransform.titlecase			"Title Case">
 <!ENTITY zotero.item.textTransform.sentencecase			"Sentence case">
 <!ENTITY zotero.item.creatorTransform.nameSwap      	"Swap First/Last Names">
+<!ENTITY zotero.item.viewOnline							"View Online">
+<!ENTITY zotero.item.copyAsUrl							"Copy as URL">
```
dstillman: Style: ``copyAsURL`` (and above in a few places)

(Can just amend this.)

rmzelle:
Done. Let me know if I should squash this down.

dstillman:
All right, you can squash.

rmzelle:
Squashed.

dstillman:
Thanks!

gracile-fr:
I wonder (because I need to translate it) whether "Copy as URL" is not a bit obscure for most users.

URL is quite common in French, but not as mainstream as "link" or "address" (though I know they do not refer rigorously to the same concept), so I've been slightly hesitant in the past to keep "URL" (not to speak of "URI"…). And it's reinforced here by the fact that the DOI is actually transformed into a URL, which is then copied : "Copy *as* URL" (this "as" sounds strangely for me)
Is it absolutely not the case in the English-speaking world ? i.e. is "URL" really common and clear (I'm not talking about tech-savvy people) ? 

Ideally, this kind of options should be "discoverable" by a normal user, not only by those who are looking for them, no ? FWIW, Mozilla Firefox has a "Send Link" option (it copies & pastes a url in the e-mail client… which generally transforms it into a link).

dstillman:
I think "URL" is surprisingly well understood. I don't think I've ever seen anyone confused by it, and even the Chrome start page says "Search Google or type URL" in the search box. We also have a "URL" field ourselves a few fields down from this.

But it's true that browser menu options tend to use "Link", and I'm OK with changing this to "Copy as Link". I think "Copy Link" is kind of confusing, since it's not a link to begin with.

gracile-fr:
<blockquote>I think "URL" is surprisingly well understood. </blockquote> Ok, thanks. I'd probably say the same in French, i.e. surprisingly !
FWIW, I had translated it as "Copier comme un lien (URL)" = Copy as Link (URL)…

dstillman:
> Copier comme un lien (URL)

Well, I wouldn't do that. If you think "URL" isn't clear to French speakers, you could just use "lien", but having both is unnecessarily messy.

gracile-fr:
"URL" alone, as you said, is ok, "Copier comme un URL" is not. But, you're right, that's messy to have both, I'll change. Thanks and sorry for the (L10n) hassle :-)

PR #720, opened 2015-05-20
Add to delimiterRegexp in utilities.js

wbt:
Added handful of characters including quotes that should be ignored when figuring out what starts a word for title case.
Per bug report at: https://forums.zotero.org/discussion/49510/title-casing-should-ignore-smart-quotes/
Reference: http://www.fileformat.info/info/unicode/char/search.htm?q=201&preview=entity

I may not be doing this right, but it's an attempt to help!

### DIFF
--- a/chrome/content/zotero/xpcom/utilities.js
+++ b/chrome/content/zotero/xpcom/utilities.js
@@ -773,7 +773,7 @@ Zotero.Utilities = {
 			"down", "as"];
 		
 		// this may only match a single character
-		const delimiterRegexp = /([ \/\u002D\u00AD\u2010-\u2015\u2212\u2E3A\u2E3B])/;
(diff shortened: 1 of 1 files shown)


wbt:
I'm not sure sure this is the fix - the issue might be on line 812.

aurimasv:
This should be OK, but you should test it.

Best I can tell, all the characters on line 812 should just be moved to delimiterRegExp and 812 can pretty much go away (the characters may already be included in the ranges actually). In that case, we should definitely write tests to make sure this doesn't break in unexpected ways, which means this will get a bit more involved.

As is, I don't see how this would break anything.

wbt:
Too bad this couldn't get fixed!

dstillman:
Sorry, this got closed automatically when we deleted the 4.0 branch, but the relevant code no longer exists in this repo in any case. If this problem still exists, open a new pull request in https://github.com/zotero/utilities that includes some tests (which are much easier to run there) and we can merge it.

PR #743, opened 2015-05-28
Add basic tests for fulltext search

egh:
- also add withImportedAttachmentFromFile support function

I finally got these basics search tests to work with async. I am trying to establish a baseline of search tests so that sqlite fts has something to compare to.

### DIFF
--- a/test/content/support.js
+++ b/test/content/support.js
@@ -242,3 +242,16 @@ function resetDB() {
 		return Zotero.Schema.schemaUpdatePromise;
 	});
 }
+
+
+/**
+ * Imports an attachment from a test file.
+ * @param {string} filename - The filename to import (in data directory)
+ * @return {Promise<Zotero.Item>}
+ */
+function importFileAttachment(filename) {
+	let testfile = getTestDataDirectory();
+	filename.split('/').forEach((part) => testfile.append(part));
+	return Zotero.Attachments.importFromFile({file: testfile});
+}
+
--- a/test/tests/searchTest.js
+++ b/test/tests/searchTest.js
@@ -83,4 +83,68 @@ describe("Zotero.Search", function() {
 			assert.propertyVal(conditions[0], 'value', 'foo')
 		});
 	});
+
+	describe("#search()", function () {
+		let win;
+		let fooItem;
+		let foobarItem;
+
+		before(function* () {
+			// Hidden browser, which requires a browser window, needed for charset detection
+			// (until we figure out a better way)
+			win = yield loadBrowserWindow();
+			fooItem = yield importFileAttachment("search/foo.html");
+			foobarItem = yield importFileAttachment("search/foobar.html");
+		});
+
+		after(function* () {
+			if (win) {
+				win.close();
+			}
+			yield fooItem.erase();
+			yield foobarItem.erase();
+		});
+
+		it("should return matches with full-text conditions", function* () {
+			let s = new Zotero.Search();
+			s.addCondition('fulltextWord', 'contains', 'foo');
+			let matches = yield s.search();
+			assert.lengthOf(matches, 2);
+			assert.sameMembers(matches, [fooItem.id, foobarItem.id]);
+		});
+
+		it("should not return non-matches with full-text conditions", function* () {
+			let s = new Zotero.Search();
+			s.addCondition('fulltextWord', 'contains', 'baz');
+			let matches = yield s.search();
+			assert.lengthOf(matches, 0);
+		});
+
+		it("should return matches for full-text conditions in ALL mode", function* () {
+			let s = new Zotero.Search();
+			s.addCondition('joinMode', 'all');
+			s.addCondition('fulltextWord', 'contains', 'foo');
+			s.addCondition('fulltextWord', 'contains', 'bar');
+			let matches = yield s.search();
+			assert.deepEqual(matches, [foobarItem.id]);
+		});
+
+		it("should not return non-matches for full-text conditions in ALL mode", function* () {
+			let s = new Zotero.Search();
+			s.addCondition('joinMode', 'all');
+			s.addCondition('fulltextWord', 'contains', 'mjktkiuewf');
+			s.addCondition('fulltextWord', 'contains', 'zijajkvudk');
+			let matches = yield s.search();
+			assert.lengthOf(matches, 0);
+		});
+
+		it("should return a match that satisfies only one of two full-text condition in ANY mode", function* () {
+			let s = new Zotero.Search();
+			s.addCondition('joinMode', 'any');
+			s.addCondition('fulltextWord', 'contains', 'bar');
+			s.addCondition('fulltextWord', 'contains', 'baz');
+			let matches = yield s.search();
+			assert.deepEqual(matches, [foobarItem.id]);
+		});
+	});
 });
--- a/test/tests/data/search/foo.html
+++ b/test/tests/data/search/foo.html
@@ -1,0 +1,9 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>hello</title>
+  </head>
+  <body>
+    <p>foo</p>
+  </body>
+</html>
--- a/test/tests/data/search/foobar.html
+++ b/test/tests/data/search/foobar.html
@@ -1,0 +1,9 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>hello</title>
+  </head>
+  <body>
+    <p>foo bar</p>
+  </body>
+</html>


test/content/support.js
```diff
@@ -241,4 +241,26 @@ function resetDB() {
...
+}
+
+
+/**
+ * Imports an attachment and calls a function when it is ready.
+ * @param {string} filename - The filename to import (in data directory)
+ * @param {function} f - A function that takes on argument, the id of the item.
+ * @returns {Promise}
+ */
+function* withImportedAttachmentFromFile(filename, f) {
```
aurimasv: I think you need to wrap this in `Zotero.Promise.coroutine`, otherwise calling this function just returns the generator object, not the promise you expect. Don't see anything about Mocha/Chai being able to handle that (maybe I'm missing something). Try to make one of your tests fail, I'm fairly sure that it will still pass.

test/content/support.js
```diff
@@ -241,4 +241,26 @@ function resetDB() {
...
+	let testfile = getTestDataDirectory();
+	testfile.append(filename);
+	let insertedId = yield Zotero.Attachments.importFromFile({file: testfile});
+	return waitForItemEvent("refresh").then(function(ids) {
```
aurimasv: You shouldn't need to wait for refresh event. `Zotero.Attachments.importFromFile` returns a promise for an itemID, which is resolved only after the item is added to the DB. I think the other places in the test suite where this us used do not return such promises. I'm guessing that you might not ever even see this event fire at this point (after `Zotero.Attachments.importFromFile` has already resolved above)

egh:
Reworked this. The tests are now passing (and, more importantly, running :)

test/content/support.js
```diff
@@ -241,4 +241,17 @@ function resetDB() {
...
+
+
+/**
+ * Imports an attachment from a test file.
+ * @param {string} filename - The filename to import (in data directory)
+ * @returns {Promise} as returned by Zotero.Attachments.importFromFile
+ */
+function importFromTestFile(filename) {
+	let testfile = getTestDataDirectory();
+	testfile.append(filename);
```
dstillman: I think subdirectories will come in handy for test data files, so this should take a relative path (e.g., "search/foo.html").  (Just splitting on forward slashes would be fine.) Then we can move and rename the HTML files.

test/tests/searchTest.js
```diff
@@ -83,4 +83,58 @@ describe("Zotero.Search", function() {
...
 		});
 	});
+
+	describe("#search()", function () {
+		let win;
+	
+		before(function () {
+			// Hidden browser, which requires a browser window, needed for charset detection
+			// (until we figure out a better way)
+			if (!Zotero.isStandalone) {
```
dstillman: This was apparently my code in attachmentsTest.js, but I have zero memory of this, and I can't imagine why it'd be necessary to test for Standalone every time we call loadBrowserWindow(). If we get tests running in Standalone, loadBrowserWindow() can just return the current window. So can take this out.

test/tests/searchTest.js
```diff
@@ -83,4 +83,58 @@ describe("Zotero.Search", function() {
...
+		});
+
+		after(function () {
+			if (win) {
+				win.close();
+			}
+		});
+	
+		it('should return results with fulltext conditions', function* () {
+			let id = yield importFromTestFile("search_foo.html");
```
dstillman: I've changed the signature of importFromFile() to just return the new item directly, so these will need to be updated. `item.id` will give you the id.

test/content/support.js
```diff
@@ -241,4 +241,17 @@ function resetDB() {
...
 	});
-}
\ No newline at end of file
+}
+
+
+/**
+ * Imports an attachment from a test file.
+ * @param {string} filename - The filename to import (in data directory)
+ * @returns {Promise} as returned by Zotero.Attachments.importFromFile
```
dstillman: Should update this with what it's actually returning (which is now `{Promise<Zotero.Item>}`. Don't need the rest of the line. Also we use `@return`, not `@returns`.

test/tests/searchTest.js
```diff
@@ -83,4 +83,58 @@ describe("Zotero.Search", function() {
...
+			}
+		});
+
+		after(function () {
+			if (win) {
+				win.close();
+			}
+		});
+	
+		it('should return results with fulltext conditions', function* () {
```
dstillman: We've used double-quotes for all test titles so far. No difference, but should do the same here for consistency.

test/tests/searchTest.js
```diff
@@ -83,4 +83,58 @@ describe("Zotero.Search", function() {
...
 	});
+
+	describe("#search()", function () {
+		let win;
+	
+		before(function () {
+			// Hidden browser, which requires a browser window, needed for charset detection
+			// (until we figure out a better way)
+			if (!Zotero.isStandalone) {
+				return loadBrowserWindow().then(window => win = window);
```
dstillman: I know this wasn't your code to begin with, but better to make the `before` handle a generator and use `win = yield loadBrowserWindow();` here.

test/tests/searchTest.js
```diff
@@ -83,4 +83,58 @@ describe("Zotero.Search", function() {
...
+			let id = yield importFromTestFile("search_foobar.html");
+			let s = new Zotero.Search();
+			s.addCondition('joinMode', 'all');
+			s.addCondition('fulltextWord', 'contains', 'foo');
+			s.addCondition('fulltextWord', 'contains', 'bar');
+			let results = yield s.search();
+			assert.equal(results[0], id);
+		});
+		
+		it('should not return failed matches for ALL fulltext conditions', function* () {
```
dstillman: Despite the condition names, let's go with "full-text" (with hyphen) for these, and also say "for full-text conditions in ALL mode" for clarity.

dstillman: And maybe "non-matches" instead of "failed matches", since there's no error state here.

egh:
@dstillman Thanks for all the feedback. I'll make the changes you requested.

Generally, do you prefer me to amend the commit and force push or make changes in a new commit?

test/content/support.js
```diff
@@ -241,4 +241,17 @@ function resetDB() {
...
 		return Zotero.Schema.schemaUpdatePromise;
 	});
-}
\ No newline at end of file
+}
+
+
+/**
+ * Imports an attachment from a test file.
+ * @param {string} filename - The filename to import (in data directory)
```
aurimasv: nit: no dash after `filename`

test/content/support.js
```diff
@@ -241,4 +241,17 @@ function resetDB() {
...
+ * @returns {Promise} as returned by Zotero.Attachments.importFromFile
+ */
+function importFromTestFile(filename) {
```
aurimasv: To clarify, could we call this `importFileAttachment` (it's not clear what we're importing from test file)

test/tests/searchTest.js
```diff
@@ -83,4 +83,58 @@ describe("Zotero.Search", function() {
...
+		it('should return results with fulltext conditions', function* () {
+			let id = yield importFromTestFile("search_foo.html");
+			let s = new Zotero.Search();
+			s.addCondition('fulltextWord', 'contains', 'foo');
+			let results = yield s.search();
+			assert.equal(results[0], id);
+		});
+		
+		it('should not return failed matches with fulltext conditions', function* () {
+			let id = yield importFromTestFile("search_foo.html");
```
aurimasv: since these are not getting cleaned up after each test, we should probably just import these files in the `before` handler once. Otherwise, this may get complicated. Also, I think you should reset the DB before running all the search tests, since they may start getting affected by other items that are imported in previous tests.

test/content/support.js
dstillman: No, with the hyphen is the style I've been using. It makes it easier to read from the code itself (and is what JSDoc uses in their examples).

test/content/support.js
dstillman: (I think this is meant to mirror importFromFile, so it's clear if you're familiar with the Zotero.Attachments API, but in other contexts importFileAttachment is probably clearer.)

test/tests/searchTest.js
dstillman: (`resetDB()`)

test/content/support.js
aurimasv: My bad. Haven't noticed that change.

dstillman:
> Generally, do you prefer me to amend the commit and force push or make changes in a new commit?

You can add regular (not-amended) commits and push normally, which makes it easier for us to see the changes. Once we give a PR the final OK, you can rebase and squash and force-push.

egh:
Thanks, @dstillman & @aurimasv. I believe I have addressed all the issues your raised.

egh:
I could not get `resetDB()` to work successfully. Because the tests are running fine now, I left it as is.

test/tests/searchTest.js
```diff
@@ -83,4 +83,57 @@ describe("Zotero.Search", function() {
...
+				win.close();
+			}
+		});
+
+		it("should return results with full-text conditions", function* () {
+			let s = new Zotero.Search();
+			s.addCondition('fulltextWord', 'contains', 'foo');
+			let results = yield s.search();
+			assert.equal(results[0], fooItem.id);
+		});
```
aurimasv: Is there a guarantee for the order? This should return both file attachments, no? Could use `assert.sameMembers` here

test/tests/searchTest.js
```diff
@@ -83,4 +83,57 @@ describe("Zotero.Search", function() {
...
+			assert.lengthOf(results, 0);
+		});
+
+		it("should return results for full-text conditions in ALL mode", function* () {
+			let s = new Zotero.Search();
+			s.addCondition('joinMode', 'all');
+			s.addCondition('fulltextWord', 'contains', 'foo');
+			s.addCondition('fulltextWord', 'contains', 'bar');
+			let results = yield s.search();
+			assert.equal(results[0], foobarItem.id);
```
aurimasv: I'd also test to make sure there is only one result. So maybe use `.sameMembers` again

dstillman: sameMembers actually doesn't guarantee a count, so you can keep this, but add `lengthOf(results, 1)`.

aurimasv: But that's only if the members are duplicated, no? That's a good point though nonetheless.

aurimasv:
>I could not get resetDB() to work successfully.

Can you elaborate? This should be working, but it might be timing out.

test/tests/searchTest.js
dstillman: Right, the issue is that `assert.sameMembers([1, 1], [1])` passes.

test/tests/searchTest.js
egh: whoops, good point!

egh:
@aurimasv I could elaborate, but as @dstillman pointed out I can use `.erase()` which should have the same effect, right?

aurimasv:
No, you want to rest before the tests. @dstillman is suggesting cleaning up after

egh:
Right, ... obviously it's a little late in the day for me :)

egh:
@aurimasv 
```javascript
		before(function* () {
			// Hidden browser, which requires a browser window, needed for charset detection
			// (until we figure out a better way)
			this.timeout(60000);
			yield resetDB();
			win = yield loadBrowserWindow();
			fooItem = yield importFileAttachment("search/foo.html");
			foobarItem = yield importFileAttachment("search/foobar.html");
		});
```
```    #search()
Unhandled rejection:

Error: Database permanently closed; not re-opening
    Zotero.DBConnection.prototype._getConnectionAsync<@chrome://zotero/content/xpcom/db.js:1151:1
From previous event:
    Zotero.DBConnection.prototype.queryAsync@chrome://zotero/content/xpcom/db.js:611:9
    Zotero.DBConnection.prototype.backupDatabase<@chrome://zotero/content/xpcom/db.js:1021:11
    self.setTimeout/<.notify@resource://zotero/bluebird.js:54:5
From previous event:
    Zotero.DBConnection.prototype.observe@chrome://zotero/content/xpcom/db.js:858:4

Error: Database permanently closed; not re-opening
    Zotero.DBConnection.prototype._getConnectionAsync<@chrome://zotero/content/xpcom/db.js:1151:1
From previous event:
    Zotero.DBConnection.prototype.queryAsync@chrome://zotero/content/xpcom/db.js:611:9
    Zotero.DBConnection.prototype.backupDatabase<@chrome://zotero/content/xpcom/db.js:1021:11
    self.setTimeout/<.notify@resource://zotero/bluebird.js:54:5
From previous event:
    Zotero.DBConnection.prototype.observe@chrome://zotero/content/xpcom/db.js:858:4
```

dstillman:
For the moment, you can probably fix that with a `yield Zotero.Promise.delay(n);` — start with 100 for `n` and increase until it doesn't happen. I'll look into a more general fix.

dstillman:
Also, are you getting that running all tests, or just the search tests, or just this block with `.only`?

egh:
Just the search tests: `xvfb-run bash runtests.sh searchTest.js`

egh:
OK, n=3000 doesn't work, n=5000 seems to.

egh:
I am having trouble consistently preventing it even with n=10000.

It would probably be easier if instead of foo & bar & baz I used some random strings that are very unlikely to appear in any other document, avoiding the need to clear the db. This would also improve performance by avoiding a slow step in the tests.

dstillman:
Well, just leave it out for now. I'll try to fix that issue, and I'd also like to see if we can just store a copy of the DB after initialization that can be copied in on a `resetDB` call to avoid the schema initialization. (`resetDB` only takes about around 1 second for me, but it can be extremely slow on Travis.)

test/tests/data/search/a.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>hello</title>
+  </head>
+  <body>
+    <p>mjktkiuewf</p>
```
aurimasv: I actually did prefer something more... legible. Makes it easier to ensure that there are no typos, etc. Currently resetting DB is not working, but there are no conflicts yet either. I'm sure we'll figure out how to fix the DB resetting soon and we'll patch this up at that time. For now, I think foo bar baz is good without DB reset.

aurimasv:
Other than that, looks good to me. Keep in mind that we don't get notifications when new commits are added to PRs so you have to ping us to make sure we see changes.

test/tests/data/search/a.html
egh: It does look kind of crazy, but I also think it's nice to not need to reset the db. First, it will speed up tests. Second, more non-matching data in the db can only make the test more robust, it seems to me.

aurimasv: I wouldn't worry too much about the couple seconds that it takes to reset the DB. The additional data may make the test more robust, but we can't really count on it, since we have no idea how populated the database is at that point (e.g. some test right before this ends up resetting the DB). So it really gives us no additional confidence. Overall, I still prefer legibility and I still think that once `resetDB` is working robustly, these tests should reset the DB. Not a big deal right now though.

egh: If it's only a couple seconds, that seems fine.
A little randomness (for instance, as to what the database is actually populated) is a good thing to have for your tests, in my opinion, since it increases the possibility of a failure. :)

dstillman: Yes, sorry if I wasn't clear — the strings should remain as they were before, so that they're easily readable, and we'll deal with DB resetting, which isn't an issue now, separately. (There may be dozens more tests, so having to parse the difference between random strings wouldn't be fun.)

egh:
All concerns addressed. The build is failing but these test are passing.

dstillman:
OK, this can be squashed.

egh:
Squashed! Thanks for all your feedback.

dstillman:
Thanks!

PR #746, opened 2015-05-31
Merge #659

simonster:


chrome/content/zotero/xpcom/data/item.js
```diff
@@ -4181,6 +4181,7 @@ Zotero.Item.prototype.toJSON = Zotero.Promise.coroutine(function* (options) {
 	
 	obj.dateAdded = Zotero.Date.sqlToISO8601(this.dateAdded);
 	obj.dateModified = Zotero.Date.sqlToISO8601(this.dateModified);
+	if(obj.accessDate) obj.accessDate = Zotero.Date.dateToISO(Zotero.Date.sqlToDate(obj.accessDate));
```
simonster: So this converts to UTC, whereas `Zotero.Date.sqlToISO8601` does not. What exactly do we want here?

simonster:
@dstillman Can you take a look at the test failures here? They're in `dataObjectsTest.js`, which I don't think I changed.

test/tests/supportTest.js
```diff
@@ -11,4 +11,185 @@ describe("Support Functions for Unit Testing", function() {
...
+	// 		for (let itemName in oldData) {
+	// 			delete oldData[itemName].id;
+	// 		}
+	// 		for (let itemName in newData) {
+	// 			delete newData[itemName].id;
+	// 		}
+			
+	// 		assert.deepEqual(oldData, newData, 'citeproc-js export data has not changed');
+	// 	}));
+	// });
```
simonster: Still need to figure out what to do with citeproc-js, since it wants items to be retrieved synchronously. Probably we should generate the CSL JSON before calling citeproc-js.

chrome/content/zotero/xpcom/data/item.js
aurimasv: I think that `dateAdded` and `dateModified` are stored in UTC, so `toJSON` maintains that here. Whether we want `accessDate` exported in UTC, I'm not entirely sure. IIRC, it's stored and treated as local date within the client, but for API JSON format (being an interchange format), it would make sense to export in UTC (idk if that's what happens now for web API).

aurimasv:
Seems like tests from test/tests/translateTest.js are not running

simonster:
@aurimasv Fixed.

chrome/content/zotero/xpcom/data/item.js
dstillman: No, it's stored as UTC — we don't store anything as local time — and returned from getField as UTC as well. So I think this can just use sqlToISO8601.

dstillman:
@simonster Sorry, left out some lines. dataObjects test failures should be fixed with the latest commits.

simonster:
There are [intermittent failures](https://travis-ci.org/zotero/zotero/jobs/64859472) on Travis that appear to be triggered by this PR but produced by code elsewhere. We need to track these down, but I'm going to go ahead and merge this anyway so that it doesn't get conflicted out again.

PR #747, opened 2015-05-31
Fix Library Lookup and Locate Engines Lookup

### LINKED ISSUE #744: Library lookup saves item (twice) instead of performing lookup

zuphilip:
* This resolves #744.
 * Changed private function `_constructLookupFunction` and its calls. Add there also stopPropagation() function.

### DIFF
--- a/chrome/content/zotero/browser.js
+++ b/chrome/content/zotero/browser.js
@@ -530,21 +530,30 @@ var Zotero_Browser = new function() {
 			menuitem.setAttribute("tooltiptext", Zotero.getString("locate.libraryLookup.tooltip"));
 			menuitem.setAttribute("image", "chrome://zotero/skin/locate-library-lookup.png");
 			menuitem.setAttribute("class", "menuitem-iconic");
-			menuitem.addEventListener("command", _constructLookupFunction(tab, function(event, obj) {
-				var urls = [];
-				for (let i = 0; i < obj.newItems.length; i++) {
-					var url = Zotero.OpenURL.resolve(obj.newItems[i]);
-					if(url) urls.push(url);
-				}
-				ZoteroPane.loadURI(urls, event);
-			}), false);
-			popup.appendChild(menuitem);		
+			menuitem.addEventListener("command",
+				function (event) {
+					_constructLookupFunction(tab, function(event_param, obj) {
+						var urls = [];
+						for (let i = 0; i < obj.newItems.length; i++) {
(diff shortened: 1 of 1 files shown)


simonster:
I was just about to look into this, thanks! But couldn't we just make the function returned by `_constructLookupFunction` call `e.stopPropagation()` at the end, instead of doing this reorganization?

If we are going to reorganize things in this way, then `_constructLookupFunction` should be renamed, since it's no longer constructing a function.

zuphilip:
I think there are two errors currently: For one the lookup is not working and second the "parent" function is called (sometimes even twice). The `stopPropagation` will solve the second problem and could also be moved somewhere else. However, I couldn't exactly find out what causes the first problem (and I tried several things...).

I thought about renaming the function. Because all the lookup specific stuff are in the other part. Moreover, I think to remember that there migth be other ideas to include in this menue where maybe this function may be suitable.

zuphilip:
A suggestion for renaming: `performAfterTranslation`.

simonster:
@zuphilip This patch appears to be sufficient on my system:

```
diff --git a/chrome/content/zotero/browser.js b/chrome/content/zotero/browser.js
index 518f21b..35ee645 100644
--- a/chrome/content/zotero/browser.js
+++ b/chrome/content/zotero/browser.js
@@ -732,6 +732,7 @@ var Zotero_Browser = new function() {
                        Zotero_Browser.progress.show();
                        Zotero_Browser.progress.changeHeadline(Zotero.getString("ingester.lookup.performing"));
                        tab.page.translate.translate(false);
+                       e.stopPropagation();
                }
        }
        
diff --git a/t
```

Is there a case where that doesn't work?

zuphilip:
My test cases are http://www.amazon.de/dp/0141036141/ and http://www.worldcat.org/title/nineteen-eighty-four/oclc/842042029 . They still fail with your patch to perform the lookup on my FF 38.0.1, Win 7, with Zotero based on branch `4.0` latest commit https://github.com/zotero/zotero/commit/4b995dd467b8e51e4e05c8673a7205cbff22be30 + 3 lines from your patch.

simonster:
Both of those work for me with that patch on both OS X and Win 8.1, also with FF 38.0.1. If there is a difference in the behavior of that patch and this PR, I'd like to know what it is. It looks to me as though they should be functionally identical.

zuphilip:
Ui, I was misreading your `diff`. Sorry. You are right, this works the same.

PR #750, opened 2015-06-02
[Feed Reader] Data layer

aurimasv:
Let's try this.

I still need to write more tests (and there are quite possibly some additional issues with the DB transactions), so don't read too much into it. It would probably be best to just look at the existing tests for now.


test/tests/feedTest.js
```diff
@@ -0,0 +1,156 @@
...
+			assert.isUndefined(feed.lastUpdate, "lastUpdate is null");
+			assert.isUndefined(feed.lastCheckError, "lastCheckError is null");
+		});
+		it("should not allow setting a feed as child of another feed", function *() {
+			let feed = new Zotero.Feed();
+			
+			assert.throws(() => feed.parentID = 1, 'Cannot set parent key for feed', 'cannot set parent by ID');
+			assert.throws(() => feed.parentKey = 1, 'Cannot set parent key for feed', 'cannot set parent by key');
+		});
+		it("should not allow saving a feed with the same url", function *() {
```
aurimasv: I'm not 100% set on this one, but I also don't see any reason to allow it. Interestingly, the test actually currently fails, even though the SQLite schema specifies UNIQUE for url.

test/tests/feedTest.js
```diff
@@ -0,0 +1,156 @@
...
+	describe("#erase()", function() {
+		it("should erase a saved feed", function *() {
+			let feed = new Zotero.Feed();
+			feed.name = 'Foo';
+			feed.url = 'http://www.exampled.com';
+			let id = yield feed.saveTx({skipEditCheck: true});
+			feed = yield Zotero.Feeds.getAsync(id);
+			yield feed.eraseTx();
+			
+			yield assert.isRejected(feed.eraseTx(), "does not allow erasing feed twice");
```
aurimasv: This currently does not thorw/rejects and I guess it's not a big deal, but IMO it should be an error to try and delete an object twice.

test/tests/feedTest.js
```diff
@@ -0,0 +1,156 @@
...
+			
+			feed = yield Zotero.Feeds.getAsync(id);
+			assert.notOk(feed, "erased feed no longer exists");
+		});
+		it("should throw if trying to erase an unsaved feed", function() {
```
aurimasv: Same for this

test/tests/librariesTest.js
```diff
@@ -0,0 +1,106 @@
...
+		assert.isAbove(Zotero.Libraries.feedLibraryID, -1, "#feedLibraryID is a positive integer");
+	});
+	describe("#getAll()", function() {
+		let ids = Zotero.Libraries.getAll();
+		
+		it("should return an array of valid library IDs", function() {
+			assert.isArray(ids, "returns an array");
+			assert(ids.reduce(function(res, id) { return res && Number.isInteger(id) && id >= 0 }, true), "All IDs are positive integers");
+		})
+		it("should return all library IDs", function *() {
```
aurimasv: Interestingly, this currently fails https://travis-ci.org/zotero/zotero/jobs/65026201#L516 but I haven't investigated why yet.

test/tests/feedTest.js
dstillman: Yeah, I'm fine with restricting this.

aurimasv:
Sorry @dstillman, this PR inflated a bit, but I think it's mostly ok now, so please take a look when you get a chance.

I'll add a couple notes to these commits

aurimasv:
I guess I can make this a bit easier. Please give me a few minutes to clean up the commits.

dstillman:
So I guess this would've been a good question to ask a long time ago, but why implement feeds as collections and not as individual libraries, the same way that groups are just libraries with some extra data in another table?

aurimasv:
What would be the advantage? I think they're conceptually similar, but here are some advantages of a collection-like model:
* Initial plan (and still sort of planned) was to make it possible to group different feeds into collections for organizational purposes.
* You could, in theory, have the same items (by GUID) being pulled into different feeds. Reading one item would mark that item as read in all feeds.
* Marking all feeds as read could be easier if they're in the same library (i.e. right-click library and select mark as read)

dstillman:
> Initial plan (and still sort of planned) was to make it possible to group different feeds into collections for organizational purposes.

Ah, OK. I suppose that's sufficient justification. And you might want a recursive-collections display of the sub-feeds from a parent collection.

chrome/content/zotero/xpcom/data/dataObjects.js
```diff
@@ -74,6 +72,9 @@ Zotero.defineProperty(Zotero.DataObjects.prototype, 'primaryFields', {
 	get: function () Object.keys(this._primaryDataSQLParts)
 }, {lazy: true});
 
+Zotero.defineProperty(Zotero.DataObjects.prototype, 'primaryDataSQLFrom', {
+	get: function() " " + this._primaryDataSQLFrom + " " + this._primaryDataSQLWhere
+}, {lateInit: true});
```
aurimasv: This and the change above are actually no longer relevant. Some hacky code relied on this previously.

chrome/content/zotero/xpcom/data/feed.js
```diff
@@ -0,0 +1,255 @@
...
+	this._feedCleanupAfter = null;
+	this._feedRefreshInterval = null;
+	this._feedUnreadCount = null;
+	
+	if (url) this.url = url;
+}
+
+Zotero.Feed._super = Zotero.Collection;
+Zotero.Feed.prototype = Object.create(Zotero.Feed._super.prototype);
+Zotero.Feed.constructor = Zotero.Feed;
```
aurimasv: note to self: use `Zotero.extendClass`

chrome/content/zotero/xpcom/data/feed.js
```diff
@@ -0,0 +1,255 @@
...
+		throw new Error('Feed URL is not set');
+	} else {
+		// Validate URL
+		let uri,
+			invalidUrlError = "Invalid feed URL " + this.url;
+		try {
+			uri = Components.classes["@mozilla.org/network/io-service;1"]
+				.getService(Components.interfaces.nsIIOService)
+				.newURI(this.url, null, null);
+			this.url = uri.spec;
```
aurimasv: note to self: move this down

dstillman:
On the other hand, one could say the same thing about group libraries — and other sorts of future subscriptions, of which there'll likely be some —  and that what we actually need is the ability to group libraries of different types into folders. They could be combined in the interface or in separate sections, but either way we'd want the container to be a folder, not a collection...

dstillman:
I don't think the GUID issue matters (that could easily be applied across libraries), and the marking-as-read wouldn't be an issue as long as there was a common container. (Latter would need to go somewhere else if you could mix and match libraries of different types in folders, but I'm not sure we'd do that. In any case, just a UI issue).

aurimasv:
It's not just about guid, but items being completely different in different libraries, though that's not a show stopper. I'm honestly not really set on one way or another. Either implementation seems ok to me, but I don't really see an advantage to the library implementation. Maybe you have something specific in mind. Changing current implementation would not be difficult.

aurimasv:
I guess one advantage is that deleting a library would delete all items within it. Also items would be guaranteed to be in a library.

dstillman:
Yeah, my sense is that there'll be more that we'll have to override or special-case than if these were libraries. And I think we may want the ability to organize both these and other library types into folders, which wouldn't be possible if these were collections.

I do see the advantage of having items with the same id in different feeds be the same item, but that also seems pretty rare. And even that would require special-casing, because if you removed an item from all feeds it'd still exist in the feed library if we relied on default behavior.

So if you have a decent solution for the marking-as-read-across libraries issue (which, honestly, is optional — and I don't even know for sure if we can trust ids), I think it'd probably make more sense to go the library route.

aurimasv:
OK Dan, here goes. Feeds as Libraries. Zotero.Library is now a class and Zotero.Group and Zotero.Feed both inherit from it. Zotero.Library mostly follows Zotero.DataObject conventions, but doesn't inherit from it directly.

chrome/content/zotero/xpcom/data/collection.js
```diff
@@ -23,7 +23,7 @@
     ***** END LICENSE BLOCK *****
 */
 
-Zotero.Collection = function() {
+Zotero.Collection = function(name) {
```
aurimasv: I've been trying to make constructors take required parameters, since they always need to be set anyway. libraryID defaults to user library, so it's not technically required.

chrome/content/zotero/xpcom/data/collections.js
```diff
@@ -46,13 +46,23 @@ Zotero.Collections = function() {
 		hasChildCollections: "(SELECT COUNT(*) FROM collections WHERE "
 			+ "parentCollectionID=O.collectionID) != 0 AS hasChildCollections",
 		hasChildItems: "(SELECT COUNT(*) FROM collectionItems WHERE "
-			+ "collectionID=O.collectionID) != 0 AS hasChildItems"
+			+ "collectionID=O.collectionID) != 0 AS hasChildItems",
```
aurimasv: hmm, I looks like this may have gotten left over after a merge conflict. Will clean it up.

chrome/content/zotero/xpcom/data/collections.js
```diff
@@ -46,13 +46,23 @@ Zotero.Collections = function() {
...
 	};
 	
 		
 	this._primaryDataSQLFrom = "FROM collections O "
-			+ "LEFT JOIN collections CP ON (O.parentCollectionID=CP.collectionID)";
+			+ "LEFT JOIN collections CP ON (O.parentCollectionID=CP.collectionID)"
```
aurimasv: Undo this as well

chrome/content/zotero/xpcom/data/collections.js
```diff
@@ -46,13 +46,23 @@ Zotero.Collections = function() {
...
 	
+	this._relationsTable = "collectionRelations";
+	
+	// Choose correct item type when loading from DB row
+	this._getObjectForRow = function(row) {
+		if (row.feedUrl) {
+			return new Zotero.Feed();
+		}
```
aurimasv: Another leftover chunk of code (sorry, making notes for myself)

chrome/content/zotero/xpcom/data/dataObject.js
```diff
@@ -1171,9 +1181,13 @@ Zotero.DataObject.prototype._initErase = function (env) {
 		key: this.key
 	};
 	
+	if (!env.options.skipEditCheck) this.editCheck();
+	
 	if (env.options.skipDeleteLog) {
 		env.notifierData[this.id].skipDeleteLog = true;
 	}
+	
+	return Zotero.Promise.resolve(true);
```
aurimasv: Should wrap this in `Zotero.Promise.method` instead

chrome/content/zotero/xpcom/schema.js
```diff
@@ -1469,7 +1469,7 @@ Zotero.Schema = new function(){
 			});
 			yield _updateDBVersion('compatibility', _maxCompatibility);
 			
-			var sql = "INSERT INTO libraries (libraryID, libraryType, editable, filesEditable) "
+			var sql = "INSERT INTO libraries (libraryID, type, editable, filesEditable) "
 				+ "VALUES "
 				+ "(?, 'user', 1, 1), "
 				+ "(4, 'publications', 1, 1)"
```
aurimasv: Why are we specifying library IDs here at all? why not just let DB assign those? (granted that they would pretty much always end up 1 and 2 anyway)

chrome/content/zotero/xpcom/uri.js
```diff
@@ -95,13 +95,14 @@ Zotero.URI = new function () {
...
+			case 'feed':
 				var id = Zotero.Users.getCurrentUserID();
 				if (!id) {
 					throw new Zotero.Error("User id not available", "USER_NOT_AVAILABLE");
 				}
 				
-				if (libraryType == 'publications') {
-					return "users/" + id + "/publications";
+				if (libraryType == 'publications' || libraryType == 'feed') {
+					return "users/" + id + "/" + libraryType;
```
aurimasv: Need to update this. We no longer have a single feed library.

test/tests/groupTest.js
```diff
@@ -1,13 +1,28 @@
-"use strict";
-
```
aurimasv: should revert this

test/tests/itemsTest.js
```diff
@@ -2,6 +2,7 @@ describe("Zotero.Items", function () {
 	var win, collectionsView, zp;
 	
 	before(function* () {
+		this.timeout(10000);
 		win = yield loadZoteroPane();
```
aurimasv: This was taking too long when running items tests directly without running other tests first.

test/tests/translateTest.js
```diff
@@ -44,16 +44,18 @@ function saveItemsThroughTranslator(translatorType, items) {
...
-		"	var json = JSON.parse('"+JSON.stringify(items).replace(/'/g, "\'")+"');\n"+
+		"	var json = JSON.parse('"+JSON.stringify(items).replace(/['\\]/g, "\\$&")+"');\n"+
 		"	for (var i=0; i<json.length; i++) {"+
 		"		var item = new Zotero.Item;\n"+
 		"		for (var field in json[i]) { item[field] = json[i][field]; }\n"+
 		"		item.complete();\n"+
 		"	}\n"+
-		"}"));
+		"}"
+	));
```
aurimasv: Undo this change

aurimasv:
And I still need to fix some sync issues (see failing test), though I'm not quite yet sure what's going on there.

chrome/content/zotero/xpcom/data/collection.js
dstillman: The problem with this is that it's messier and less clear later if we add a required parameter. And setting the name before the libraryID is awkward. In the absence of named parameters, I think I'd prefer to just use properties. (We do use a required parameter for itemTypeID for Zotero.Item, but that one isn't ever going to change, it does have to be set before most other fields, and seems less weird before libraryID because it's actually defining the kind of item you're creating.)

resource/schema/userdata.sql
```diff
@@ -238,11 +257,11 @@ CREATE INDEX deletedItems_dateDeleted ON deletedItems(dateDeleted);
 
 CREATE TABLE libraries (
     libraryID INTEGER PRIMARY KEY,
-    libraryType TEXT NOT NULL,
-    editable INT NOT NULL,
-    filesEditable INT NOT NULL,
+    type TEXT NOT NULL,
+    editable INT NOT NULL DEFAULT 0,
+    filesEditable INT NOT NULL DEFAULT 0,
```
dstillman: I left defaults out for these on purpose because the code creating libraries really does need to make the decision as to editability. Creating a library isn't a common action, so there's no need for this to be implicit.

resource/schema/userdata.sql
```diff
@@ -205,6 +205,25 @@ CREATE TABLE collectionRelations (
...
+    lastCheckError TEXT,
+    cleanupAfter INT,
+    refreshInterval INT,
+    FOREIGN KEY (libraryID) REFERENCES libraries(libraryID) ON DELETE CASCADE
+);
+
+CREATE TABLE feedItems (
+    itemID INTEGER PRIMARY KEY,
+    guid TEXT NOT NULL UNIQUE,
+    readTimestamp TIMESTAMP,
```
dstillman: `readTime`? (Also, we're not actually using the time, are we? Unless there's a compelling need to show or use that anywhere, I think this can just be `read INT` with a boolean.)

aurimasv: We use the time to decide when to purge read items.

dstillman: Ah, OK.

chrome/content/zotero/xpcom/schema.js
dstillman: It's for consistency with migrateUserDataSchema, where we use 1 and 4 (known unused ids, coming from the global libraryIDs in 4.0) for upgrading users. So, actually, the '?' for 'user' should just be 1 here.

dstillman:
I wouldn't worry about the sync failure for now. We'll need to discuss our general syncing plan for feeds.

chrome/content/zotero/xpcom/data/collection.js
aurimasv: I see in support.js we pass down `params` object as argument. Is that what we want to do in general?

aurimasv:
>I wouldn't worry about the sync failure for now. We'll need to discuss our general syncing plan for feeds, and I can help with that part.

The changes in this PR affect syncing of group/user libraries though. I'm more than happy to leave that part to you though.

aurimasv:
I think that should address most comments. Still need to do:
* Add tests for Zotero.URI
* Change constructor parameters
* Make table columns unique?

chrome/content/zotero/xpcom/data/collection.js
dstillman: I'm OK with a `params` argument, as an alternative to setting properties after the constructor.

dstillman:
Yeah, I can deal with sync breakage. I have a big sync patch coming on my end anyway.

aurimasv:
Please take a look. We're getting close, I think.

aurimasv:
OK, how about now?

chrome/content/zotero/xpcom/data/feedItem.js
```diff
@@ -0,0 +1,143 @@
...
+		env.transactionOptions.onCommit = () => {
+			if (superOnCommit) superOnCommit();
+			this.ObjectsClass._setGUIDMapping(this.guid, env.id);
+		};
+	}
+	
+	return proceed;
+});
+
+Zotero.FeedItem.prototype.forceSave = function(options) {
```
dstillman: If this is going to to exist, seems like it should at least be `forceSaveTx()` (and `forceEraseTx()` below).

dstillman:
I might have more notes later, and we'll need to make some adjustments once we figure out the organization stuff, but this is looking good, so you can squash it down as appropriate.

dstillman:
(Other than the few failing (non-sync) tests.)

aurimasv:
>(Should fix the failing (non-sync) tests, though.)

>`Cannot edit feedItem in read-only Zotero library`

Is the result of reverting [this](https://github.com/aurimasv/zotero/commit/e097aba50479eede71d4138ed715fa7ca6368760#commitcomment-13321934), so I think it should currently fail because that throw needs to be fixed.

Also, could you OK the last commit before I squash this all down?

dstillman:
OK, all set.

aurimasv:
Squish squashed

chrome/content/zotero/xpcom/data/feedItem.js
dstillman: Could you adjust these?

aurimasv: Sorry about that. Fixed and squashed.

aurimasv:
:tada:

zuphilip:
Wow, :clap: , cool!

PR #751, opened 2015-06-02
Check if attempting to use a legacy generator function in Mocha tests

aurimasv:
Since we don't wrap those in `Zotero.Promise.coroutine`, which means that they never end up running. Alternatively we could wrap them, but there's no reason we should be using these (and no one uses them on purpose)

I'm thinking that there should also be a way to throw or warn if `assert` is never invoked during a test, but I haven't figured out how to do that yet.

### DIFF
--- a/test/content/runtests.js
+++ b/test/content/runtests.js
@@ -138,7 +138,7 @@ function Reporter(runner) {
 	});
 }
 
-// Monkey-patch Mocha to check instanceof Error using compartent-local
+// Monkey-patch Mocha to check instanceof Error using compartment-local
 // Error object
 Mocha.Runner.prototype.fail = function(test, err){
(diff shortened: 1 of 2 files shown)


dstillman:
Re: the latter, I used a version of <a href="https://github.com/mochajs/mocha/wiki/Assertion-counting">this</a> for the streaming server tests.

dstillman:
Example: https://github.com/zotero/stream-server/blob/master/test/support/assertion_count.js

dstillman:
Thanks for catching these. I'll fix the item test that's failing, which isn't testing the right thing anyway.

aurimasv:
I got burned by this a couple times already. Seems like I  missed a few of these cases in my tests as well.

PR #756, opened 2015-06-05
Attachment proxy fixes

aurimasv:
Seems like this should have been detected much earlier. It appears that attachment downloads from proxied pages in Firefox should have been broken since 7f7fe90ad5fc21502fb1ab2e84f793628cd9892d

The second commit doesn't really have much impact, so we can remove it for now, until #578 (probably for 5.0 release)


simonster:
Thanks! I think the reason that we didn't detect this earlier is that it only affects proxies detected by subdomain dropping and not proxies detected by Zotero.Proxies.

PR #757, opened 2015-06-05
Fix institutional author export to CiteProc

aurimasv:
Re https://forums.zotero.org/discussion/49780/beta-singlefield-authors-not-citing-properly/

CiteProc updated since @fbennett altered institutional author handling in 1.0.599 https://bitbucket.org/fbennett/citeproc-js/issue/173/fix-institutional-authors-jiggery-poker

### DIFF
--- a/chrome/content/zotero/xpcom/utilities.js
+++ b/chrome/content/zotero/xpcom/utilities.js
@@ -1553,7 +1553,12 @@ Zotero.Utilities = {
 			creatorType = CSL_NAMES_MAPPINGS[creatorType];
 			if(!creatorType) continue;
 			
-			var nameObj = {'family':creator.lastName, 'given':creator.firstName};
+			var nameObj;
+			if (creator.lastName || creator.firstName) {
+				nameObj = {'family': creator.lastName, 'given': creator.firstName};
+			} else if (creator.name) {
+				nameObj = {'literal': creator.name};
+			}
 			
 			if(cslItem[creatorType]) {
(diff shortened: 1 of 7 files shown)


fbennett:
There shouldn't be any difference in the way citeproc-js handles authors post-1.0.599, in the configuration used by Zotero.

The [linked thread](https://bitbucket.org/fbennett/citeproc-js/issue/173/fix-institutional-authors-jiggery-poker) is about code that expressed individual authors as affiliated with an institution that followed them in the input list. It was an over-engineered hack to start with, and the discussion with Chris Maloney at NIH convinced me at long last that it should not be the processor default&mdash;but that behavior was disabled in Zotero very early on.

aurimasv:
Ok, but note that 4.0 is currently at 1.0.543 and (I think) this PR changes how institutional authors are exported (via literal, not family name IIRC). If that's ok, I can pull the update out, though there shouldn't be a reason to keep it back anyway.

fbennett:
The `literal` key should be fine.

bwiernik:
Should this issue have a Blocker tag? It would probably be bad to push a release with citing of institutional authors broken.

aurimasv:
I pulled out the citeproc-js changes, so this should be a fairly trivial merge now.

aurimasv:
Looks like citeproc-js update was necessary after all. See https://forums.zotero.org/discussion/50512/institutional-singlefield-creator/

PR #761, opened 2015-06-11
Always save snapshots in UTF-8 encoding

aurimasv:
Re https://forums.zotero.org/discussion/49897/thecreate-web-page-item-from-current-pageis-sometimes-not-working-correctly/

There is something wrong with the way `nsIScriptableUnicodeConverter` converts to gbk (maybe other formats too). The HTML of http://www.cmiw.cn/forum.php is truncated in the middle of an em tag. In general, `ConvertFromUnicode` is supposed to return an `ACString` (not `AString`), which is not something that JavaScript can handle well.

### DIFF
--- a/chrome/content/zotero/webpagedump/domsaver.js
+++ b/chrome/content/zotero/webpagedump/domsaver.js
@@ -199,7 +199,7 @@ var wpdDOMSaver = {
 			// Changed by Dan for Zotero
 			"script": true, // no scripts
 
-			"encodeUTF8": false, // write the DOM Tree as UTF-8 and change the charset entry of the document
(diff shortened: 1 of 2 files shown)


aurimasv:
One tiny bug is that the placeholder in the search box on that page is not encoded properly, but I'm not going to pursue it at this point.

dstillman:
Hmm, there's actually some weird font changing too, at least on OS X. The CSS is the same, but the system fonts (i.e., fonts with weird names that are swapped in by the system to cover non-Latin scripts) used are different. I don't totally know how OS X handles these sorts of scripts, so I'm not sure what's going on here. Are you seeing a difference in the font rendering between the two charsets?

aurimasv:
I don't see any difference in fonts on Windows

PR #768, opened 2015-06-22
Replace getTargets with getStatementsMatching

zuphilip:
As suggested in the same file `getTargets` is deprecated since 2.1


zuphilip:
@dstillman : I found this old PR here. Any reaction would be appreciated.

dstillman:
Not my department, but @simonster would be able to comment.

zuphilip:
This PR does what is mentioned some lines below, i.e. that this function is deprecated since version 2.1 and should be replaced:
https://github.com/zotero/zotero/blob/81b876a5b2e3938793c708eb552fcf308d12ddc2/chrome/content/zotero/xpcom/translation/translate.js#L3159

PR #770, opened 2015-06-22
Allows the replacement of multiple spaces in the platform name

AlexisMignon:
On linux 64bit with a 32bit version of firefox:
 ```javascript
Zotero.platform.replace(" ", "-")
```

replaces only the first occurence of " ", which makes zotero unable to detect the presence of PDFtools installed manually.

With:

 ```javascript
Zotero.platform.split(' ').join('-')
``` 

all spaces are replaced and Zotero can detect the presence of PDFtools.


dstillman:
Fixed with `replace(/ /g, '-')`. Thanks.

PR #775, opened 2015-06-28
Add possibility to save as webpage also when translators are available

### LINKED ISSUE #644: Include web page save option in toolbar menu on every page
As suggested on https://github.com/zotero/zotero/commit/9a87f1595660a1715f3448172d6ce008e308adec#commitcomment-9867103

zuphilip:
This is an attempt to resolve #644 

zotero-web-page-save

### DIFF
--- a/chrome/content/zotero/browser.js
+++ b/chrome/content/zotero/browser.js
@@ -537,6 +537,16 @@ var Zotero_Browser = new function() {
 				_constructLookupFunction(tab, function(e, obj) {
 					Zotero_LocateMenu.locateItem(e, obj.newItems);
 				}), true);
+				
+			menuitem = document.createElement("menuitem");
+			menuitem.setAttribute("label", "Save to Zotero as Web Page (with snapshot)");
+			menuitem.setAttribute("class", "menuitem-iconic");
+			menuitem.addEventListener("command", function (event) {
+				ZoteroPane.addItemFromPage('temporaryPDFHack', !Zotero.Prefs.get('automaticSnapshots') );
+				event.stopPropagation();
+			}); 
+			popup.appendChild(menuitem);
+			
 		}
 		else {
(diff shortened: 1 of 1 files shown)


dstillman:
1) This option should go in the top section, below any translators.

2) The label needs to be localized.

3) It should have the web page icon.

4) This is performing the inverse of the snapshot pref, all the time, which both isn't what we want and means that the label is incorrect.

More generally, we need to decide how we want to handle snapshots and files with this menu. Showing all possible permutations (all preference states for all save options) isn't really feasible, so maybe we want to just have checkbox menuitems that allow for toggling those two prefs from the menu.

aurimasv:
>More generally, we need to decide how we want to handle snapshots and files with this menu.

What about Shift as a toggle for the option. I'm not sure if we can make the menu options respond to Shift press though to display altered behavior.

dstillman:
That doesn't really work, since there are more than two possible states. (And I think we'd need the pref checkboxes regardless, since very few people would use the modifier key.)

zuphilip:
Okay, I try to work on 1) - 4) the next days.

The options for the translators and snapshots are currently in the preferences tab:
preferences-general
Maybe, we could take these three points "out there"
 *  Automatically take snapshots when creating items from web pages
 * Automatically attach associated PDFs and other files when saving items
 * Automatically tag items with keywords and subject headings

and include it in the zotero picker menu or a dialog which is reachable from this menu. Something like the settings for the Zotero picker (similar to the "Manage User Scripts..." in <a href="http://www.publicvoid.dk/content/binary/WindowsLiveWriter/MyFirefoxPlugins_A77D/My-Firefox-Plugins-Greasemonkey-UI_thumb.png">Greasemonkey menu</a>).

dstillman:
I took care of this — thanks. (And I noticed that my existing code here was also unlocalized. Now fixed.)

PR #777, opened 2015-06-29
Allow re-using file names when saving translators

aurimasv:
This is mostly for Scaffold, where saving a translator where the label doesn't match file name (e.g. some cyrillic names) would instead create a new file.

Also fixes some translator saving logic in general and some escaping issues in Zotero.Translate tests.

If the general idea of this PR is ok, I'll add unit tests.

### DIFF
--- a/chrome/content/zotero/xpcom/translation/translators.js
+++ b/chrome/content/zotero/xpcom/translation/translators.js
@@ -368,38 +368,46 @@ Zotero.Translators = new function() {
 	
 	/**
 	 * @param	{String}		metadata
-	 * @param	{String}		metadata.translatorID		Translator GUID
-	 * @param	{Integer}		metadata.translatorType		See TRANSLATOR_TYPES in translate.js
-	 * @param	{String}		metadata.label				Translator title
-	 * @param	{String}		metadata.creator			Translator author
-	 * @param	{String|Null}	metadata.target				Target regexp
-	 * @param	{String|Null}	metadata.minVersion
-	 * @param	{String}		metadata.maxVersion
-	 * @param	{String|undefined}	metadata.configOptions
-	 * @param	{String|undefined}	metadata.displayOptions
-	 * @param	{Integer}		metadata.priority
-	 * @param	{String}		metadata.browserSupport
-	 * @param	{Boolean}		metadata.inRepository
-	 * @param	{String}		metadata.lastUpdated		SQL date
+	 *   @param	{String}		metadata.translatorID		Translator GUID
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/xpcom/translation/translators.js
```diff
@@ -368,39 +368,47 @@ Zotero.Translators = new function() {
...
+	 *   @param	{String}		metadata.creator			Translator author
+	 *   @param	{String|Null}	metadata.target				Target regexp
+	 *   @param	{String|Null}	metadata.minVersion
+	 *   @param	{String}		metadata.maxVersion
+	 *   @param	{String|undefined}	metadata.configOptions
+	 *   @param	{String|undefined}	metadata.displayOptions
+	 *   @param	{Integer}		metadata.priority
+	 *   @param	{String}		metadata.browserSupport
+	 *   @param	{Boolean}		metadata.inRepository
+	 *   @param	{String}		metadata.lastUpdated		SQL date
```
aurimasv: Idk how you will feel about this, but I think it helps a little bit with such long lists. Also, the spacing looks nice, but maybe we should use spaces instead of tabs, since it doesn't really look good with other tab width settings. (I use 2 spaces/tab, for instance)

chrome/content/zotero/xpcom/translation/translators.js
```diff
@@ -368,39 +368,47 @@ Zotero.Translators = new function() {
...
-			}
-			if (!found) {
-				throw ("Invalid translatorType '" + metadata.translatorType + "' in Zotero.Translators.save()");
+			throw ("metadata.translatorType not provided in Zotero.Translators.save()");
+		}
+		
+		// Validate translator type
+		var found = false;
+		for (let type in TRANSLATOR_TYPES) {
+			if (metadata.translatorType & TRANSLATOR_TYPES[type]) {
```
aurimasv: I guess `metadata.translatorType` is a bit redundant here

chrome/content/zotero/xpcom/translation/translators.js
```diff
@@ -418,23 +426,45 @@ Zotero.Translators = new function() {
...
 		// JSON.stringify has the benefit of indenting JSON
 		var metadataJSON = JSON.stringify(metadata, null, "\t");
-		
-		var str = metadataJSON + "\n\n" + code,
-			translator;
+		var str = metadataJSON + "\n\n" + code;
 		
 		var translator = Zotero.Translators.get(metadata.translatorID);
-		var sameFile = translator && destFile == translator.path;
-		if (sameFile) return;
```
aurimasv: I feel like I'm missing something here, since the statement is so explicit (unless it was just an oversight)

PR #778, opened 2015-06-29
Fix #774

### LINKED ISSUE #774: Combo Menue on SpringerLink wrong
The menue under the new combo zotero icon is wrong on (some/every?) SpringerLink page. E.g.

zotero-icon-menue

The icon is correctly indicating that we have here a book chapter, however the entries in the menue below are the wrong case for webpages. Example page http://link.springer.com/chapter/10.1007/978-3-642-18351-5_2 . However, every other type of page (amazon, worldcat, some library catalogues) I tested seemed to work fine.

(I tested with latest 4.0 version, i.e. based on e868c75. The error seems to come from the value `tab.page.saveEnabled=false`.)

simonster:
@dstillman Was there some rationale behind clearing `saveEnabled` here?

### DIFF
--- a/chrome/content/zotero/browser.js
+++ b/chrome/content/zotero/browser.js
@@ -807,7 +807,6 @@ Zotero_Browser.Tab.prototype.clear = function() {
 Zotero_Browser.Tab.prototype.detectTranslators = function(rootDoc, doc) {
 	if (doc instanceof HTMLDocument) {
 		if (doc.documentURI.startsWith("about:")) {
-			this.page.saveEnabled = false;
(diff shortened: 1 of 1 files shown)


dstillman:
I think I originally was disabling the icon on about:blank pages, mirroring the existing behavior of not triggering the address bar icon on about:* pages, but I decided not to disable the icon again after it's first enabled, and this way obviously doesn't work anyway.

dstillman:
(And this meant that the icon wasn't enabled if you started up on about:blank, but that's not very important.)

PR #787, opened 2015-07-04
Tweak site-specific quick copy.

aurimasv:
* Only consider HTTP/HTTPS urls
* Match against empty path
* Allow matching against hostname only (e.g. localhost)
* No need to use regexps for matching

I don't think this fixes any of the current issues being reported on the forums, but I'm neither able to reproduce them, nor do I see how this code could cause issues.


aurimasv:
Closed by 242581a270daf602697dac7b74d8eced35a408b6

PR #788, opened 2015-07-04
Resolve symbolic links when importing attachments

aurimasv:
Follow-up to dfb28ff3f675157b7ef2a6c4fe92320f2f81cc38

The trick to getting `.target` to work was setting `.followLinks` to `true`, which is somewhat silly given the purpose of `.target`

### DIFF
--- a/chrome/content/zotero/xpcom/data/item.js
+++ b/chrome/content/zotero/xpcom/data/item.js
@@ -3114,8 +3114,8 @@ Zotero.Item.prototype.relinkAttachmentFile = function(file, skipItemUpdate) {
 		throw('Cannot relink linked URL in Zotero.Items.relinkAttachmentFile()');
 	}
 	
-	if (file.leafName.endsWith(".lnk")) {
-		throw new Error("Cannot relink to Windows shortcut");
+	if (file.isSymlink()) {
+		file = Zotero.File.getTargetFile(file);
 	}
 	
 	var newName = Zotero.File.getValidFileName(file.leafName);
--- a/chrome/content/zotero/xpcom/itemTreeView.js
+++ b/chrome/content/zotero/xpcom/itemTreeView.js
@@ -3213,23 +3213,36 @@ Zotero.ItemTreeView.prototype.drop = function(row, orient, dataTransfer) {
 					// Otherwise file, so fall through
 				}
 				
+				// Resolve symbolik links
+				var moveFile = dragData.dropEffect == 'move';
+				if (file.isSymlink()) {
+					var linkPath = file.path;
+					file = Zotero.File.getTargetFile(file);
+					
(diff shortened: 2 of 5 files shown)


chrome/content/zotero/xpcom/data/item.js
```diff
@@ -3114,8 +3114,8 @@ Zotero.Item.prototype.relinkAttachmentFile = function(file, skipItemUpdate) {
 		throw('Cannot relink linked URL in Zotero.Items.relinkAttachmentFile()');
 	}
 	
-	if (file.leafName.endsWith(".lnk")) {
-		throw new Error("Cannot relink to Windows shortcut");
+	if (file.isSymlink()) {
+		file = Zotero.File.getTargetFile(file);
```
dstillman: This method works nicely, but we should probably still block *.lnk files if isSymlink() returns false, to prevent .lnk files from being added on, say, OS X. (That would be crazy, but I'm sure it's happened, and once they're in they can cause all sorts of problems. It's probably rare enough, though, that we could just throw in the various places rather than displaying the alert from the removed code below.)

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -594,6 +594,8 @@ zotero.preferences.advanced.debug.error = An error occurred sending debug output
 
 dragAndDrop.existingFiles = The following files already existed in the destination directory and were not copied:
 dragAndDrop.filesNotFound = The following files were not found and could not be copied:
+dragAndDrop.invalidLink.title = File Not Found
+dragAndDrop.invalidLink.text = The following file shortcut points to a non-existent file and cannot be imported: \n\n%S
```
dstillman: For newlines we can just add in the code rather than having them in the string (where they're prone to getting mangled).

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -594,6 +594,8 @@ zotero.preferences.advanced.debug.error = An error occurred sending debug output
 
 dragAndDrop.existingFiles = The following files already existed in the destination directory and were not copied:
 dragAndDrop.filesNotFound = The following files were not found and could not be copied:
+dragAndDrop.invalidLink.title = File Not Found
```
dstillman: We have a "File Not Found" already. Probably should just make a general.fileNotFound.

chrome/content/zotero/xpcom/itemTreeView.js
```diff
@@ -3213,23 +3213,36 @@ Zotero.ItemTreeView.prototype.drop = function(row, orient, dataTransfer) {
 					// Otherwise file, so fall through
 				}
 				
+				// Resolve symbolik links
```
dstillman: symbolic

PR #790, opened 2015-07-08
Add more debug logging for RecognizePDF

aurimasv:
Should be helpful to track down why users get locked out.

So far, I've seen that clearing all cookies presents you with a new sort of CAPTCHA, which we don't account for. That should result in "No metadata found" messages. I forgot to look at the DOM the first time I got it and I haven't been able to reproduce it since, but I'll keep playing around.


dstillman:
Available in the 4.0 beta

PR #794, opened 2015-07-09
Update citeproc-js to 1.1.25

aurimasv:
And split off Zotero additions
Re https://forums.zotero.org/discussion/50512/institutional-singlefield-creator/ and follow-up from #757

### DIFF
--- a/chrome/content/zotero/xpcom/citeproc-prereqs.js
+++ b/chrome/content/zotero/xpcom/citeproc-prereqs.js
@@ -1,0 +1,390 @@
+var CSL_HOST = {
+    debug: function (str) {
+        Zotero.debug("CSL: " + str);
+    },
+    error: function (str) {
+        Zotero.debug("CSL error: " + str);
+    }
+};
+function DOMParser() {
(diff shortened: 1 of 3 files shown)


dstillman:
I haven't reviewed or tested this, but I'm going to merge it so that it has a little time to bake in the beta before 4.0.27.5, which should be out this weekend.

fbennett:
Two further revisions to the processor are relevant to Zotero users.

1.1.37 addresses a scoping bug in RTF output. [[Forum post](https://forums.zotero.org/discussion/50304/problems-with-et-al)] [[Changes](https://bitbucket.org/fbennett/citeproc-js/commits/5a36644e44baf622d15e7332598d2f025c909a77)]

1.1.38 stops forcing capitalizing of leading name particles in the bibliography [[Forum post](https://forums.zotero.org/discussion/50199/lowercase-first-letter-of-the-surname)] [[Changes](https://bitbucket.org/fbennett/citeproc-js/commits/1d6d4032f31deb530db6c28c1b341ad324e7166a)]

Judging from the forum post associated with the latter, it looks like it has been a source of significant annoyance.

PR #795, opened 2015-07-10
Fix saving locale in Word documents

aurimasv:
Seems like we forgot to do this properly.

Also fixes some logic with remembering last-used style/locale. Should now remember all explicit user selections.

### DIFF
--- a/chrome/content/zotero/bibliography.js
+++ b/chrome/content/zotero/bibliography.js
@@ -33,8 +33,11 @@
 // Used by rtfScan.xul, integrationDocPrefs.xul, and bibliography.xul
 
 var Zotero_File_Interface_Bibliography = new function() {
-	var _io, _saveStyle;
-	var lastSelectedLocale; // Only changes when explicitly selected
+	var _io;
+	
+	// Only changes when explicitly selected
+	var lastSelectedStyle,
+		lastSelectedLocale;
 	
 	/*
 	 * Initialize some variables and prepare event listeners for when chrome is done
@@ -60,7 +63,6 @@ var Zotero_File_Interface_Bibliography = new function() {
 		// if no style is requested, get the last style used
 		if(!_io.style) {
 			_io.style = Zotero.Prefs.get("export.lastStyle");
-			_saveStyle = true;
 		}
 		
 		// add styles to list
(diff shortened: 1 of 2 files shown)


aurimasv:
There's another issue with locales reported here https://forums.zotero.org/discussion/50520/cambio-de-comillas-en-mla-espanol/

For existing documents that previously used `bibliographyLocale` hidden pref to create bibliography, that setting no longer exists and the document did not have a locale saved. So `undefined` is passed to `.getCiteProc` as `locale`, which then defaults to `Zotero.locale`. That would be different from the original `bibliographyLocale`. The obvious solution would be to make it default to the quick-copy locale, but that may still not match what was originally set. So maybe we can just expect users to notice the change in their bibliography, go into document preferences, and set the locale there.

dstillman:
Sorry, what's broken here (with the first part)? The locale seems to be saved properly in different documents for me.

aurimasv:
I think you have to restart Zotero to observe the issue. Seems like it remembers document pref state quite aggressively. We never serialized locale to document, so there's no way that we were actually saving the locale in the long run.

dstillman:
Oh, yeah. That's no good at all.

dstillman:
> The obvious solution would be to make it default to the quick-copy locale, but that may still not match what was originally set.

We move ``bibliographyLocale`` to the two new prefs, though, right? So using ``quickCopy`` would be better than nothing — at least if they didn't change the quickCopy pref their document wouldn't change?

aurimasv:
Yeah, that's true. I was also not sure if that would have some other unforseen consequences. So I guess it should be quickCopy || Zotero.locale

dstillman:
> So I guess it should be quickCopy || Zotero.locale

Do you know off-hand where it gets that value?

aurimasv:
style.js getCiteProc

dstillman:
Thanks.

I think I'll use export.lastLocale here rather than export.quickCopy.locale. It doesn't matter if the point is just to use the migrated value, since that goes to both, but if there happened to be some use of styles that didn't pass a locale (which I don't think is the case anymore within Zotero), using the locale set for a specific other purpose (Quick Copy) seems weirder than just using the last-chosen style.

PR #800, opened 2015-07-16
Check single items in selection window

rmzelle:
Green-lighted by @dstillman and @aurimasv (see https://github.com/zotero/translators/issues/919#issuecomment-121369540).

Works for me. Should merge cleanly to "master".


rmzelle:
Thanks!

rmzelle:
For some reason http://emboj.embopress.org/content/18/21/5843 does not preselect the single result for me. Zotero for Firefox 4.0.28. Can anyone reproduce?

rmzelle:
The original URL which I tested this with gives the same result (http://jmm.sgmjournals.org/content/journal/jmm/10.1099/00222615-15-2-189). It works with Zotero for Firefox 4.0.28.3, but not with Standalone.

PR #806, opened 2015-07-21
Parse author names

aurimasv:
Currently non-dropping particle parsing is being done automatically in CiteProc. This patch disables automatic parsing and moves it to `itemToCSLJSON`, which means that names are parsed when exporting to CSL JSON.

Also fixes the non-dropping particle handling reported [here](https://forums.zotero.org/discussion/30974/1/any-idea-why-an-a-author-comes-last-in-the-bibliography/) via citeproc update

### DIFF
--- a/test/tests/utilities.js
+++ b/test/tests/utilities.js
@@ -272,5 +272,89 @@ describe("Zotero.Utilities", function() {
 			
 			assert.isUndefined(cslJSON.PMID, 'field labels are case-sensitive');
 		});
+		it("should parse particles in creator names", function() {
+			let creators = [
+				{
+					// No particles
+					firstName: 'John',
+					lastName: 'Smith',
+					creatorType: 'author',
+					expect: {
+						given: 'John',
+						family: 'Smith'
+					}
+				},
+				{
+					// dropping and non-dropping
+					firstName: 'Jean de',
+					lastName: 'la Fontaine',
+					creatorType: 'author',
+					expect: {
+						given: 'Jean',
+						"dropping-particle": 'de',
+						"non-dropping-particle": 'la',
+						family: 'Fontaine'
+					}
+				},
+				{
+					// only non-dropping
+					firstName: 'Vincent',
+					lastName: 'van Gogh',
+					creatorType: 'author',
+					expect: {
+						given: 'Vincent',
+						"non-dropping-particle": 'van',
+						family: 'Gogh'
+					}
+				},
+				{
+					// only dropping
+					firstName: 'Alexander von',
+					lastName: 'Humboldt',
+					creatorType: 'author',
+					expect: {
+						given: 'Alexander',
+						"dropping-particle": 'von',
+						family: 'Humboldt'
+					}
+				},
+				{
+					// institutional author
+					lastName: 'Jean de la Fontaine',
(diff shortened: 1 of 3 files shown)


chrome/content/zotero/xpcom/citeproc.js
```diff
@@ -14388,12 +14373,78 @@ CSL.Engine.prototype.retrieveAllStyleModules = function (jurisdictionList) {
 }
 CSL.parseParticles = function(){
     var PARTICLES = [
+        ["al-", [[[0,1], null],[null,[0,1]]]],
+        ["at-", [[[0,1], null],[null,[0,1]]]],
+        ["ath-", [[[0,1], null],[null,[0,1]]]],
+        ["aá¹¯-", [[[0,1], null],[null,[0,1]]]],
```
dstillman: Encoding?

aurimasv:
Also probably fixes https://forums.zotero.org/discussion/50628/bibliography-alignment/ but no one has confirmed.

chrome/content/zotero/xpcom/citeproc.js
aurimasv: Yeah. Looks like bitbucket serves this incorrectly in the Raw view. Will fix later today

aurimasv:
Fixed encoding

aurimasv:
Pulled out citeproc update, since it got bumped independently and added tests. Should be ok to merge, I think.

test/tests/utilities.js
```diff
@@ -272,5 +272,89 @@ describe("Zotero.Utilities", function() {
...
+				},
+				{
+					// protected last name
+					firstName: 'Jean de',
+					lastName: '"la Fontaine"',
+					creatorType: 'author',
+					expect: {
+						given: 'Jean de',
+						family: 'la Fontaine'
+					}
```
aurimasv: this mirrors current citeproc behavior (AFAICT), but it's somewhat odd. As can be seen, protecting the last name prevents parsing of the first name as well. To be honest, I'm not sure when parsing the first name is actually important (is dropping particle ever used independently?), but if it is, this isn't right.

aurimasv: CC @adam3smith @fbennett @rmzelle (sorry if you already got notified)

fbennett: Sorry about the update to the name parser; I addressed a user request and fixed a bug. The name parser should probably be spun off to a separate project, but for now I'll just be careful not to touch it further without running changes past everyone affected.

About the dropping-particle no-parse, I noticed that in the code, but didn't look into it because the processor tests were passing. I think that there are currently no name formats in which the given name is not followed immediately by the dropping-particle, which would explain it. It *should* parse out, though, to be correct.

I don't have time to rework the code at the moment - I have my hands full with the Zotero 5.0 migration of Juris-M - but pull requests against the processor code are welcome.

aurimasv: That's ok. This isn't critical, since I don't think it makes a difference. Another thing to consider is that double-quote protection is not stripped off from first name field, so if the user expects to protect first name from parsing, it would actually just keep the quotes. Simple enough for the user to go back and fix.

@dstillman, this issue shouldn't hold back the PR merge.

PR #812, opened 2015-07-27
generate secure DOI links

katrinleinweber:
Because https-support for dx.doi.org was recently [upgraded](https://www.ssllabs.com/ssltest/analyze.html?d=dx.doi.org&latest), and to bumb privacy & security for researches up a bit, I suggest to stop generating plain http links where possible.

Because Zotero's server apparently enforces https, the next commit in this PR could be to change all zotero.org URLs as well. More ideas? Would doing the same to the namespaces containing a mozilla.org URL be a problem?

### DIFF
--- a/chrome/content/zotero/bindings/itembox.xml
+++ b/chrome/content/zotero/bindings/itembox.xml
@@ -409,7 +409,7 @@
 							// Pull out DOI, in case there's a prefix
 							var doi = Zotero.Utilities.cleanDOI(val);
 							if (doi) {
-								doi = "https://dx.doi.org/" + encodeURIComponent(doi);
+								doi = "https://doi.org/" + encodeURIComponent(doi);
 								label.setAttribute("isButton", true);
 								label.setAttribute("onclick", "ZoteroPane_Local.loadURI('" + doi + "', event)");
 								label.setAttribute("tooltiptext", Zotero.getString('locate.online.tooltip'));
--- a/chrome/content/zotero/bindings/itembox.xml
+++ b/chrome/content/zotero/bindings/itembox.xml
@@ -409,7 +409,7 @@
 							// Pull out DOI, in case there's a prefix
 							var doi = Zotero.Utilities.cleanDOI(val);
 							if (doi) {
-								doi = "http://dx.doi.org/" + encodeURIComponent(doi);
+								doi = "https://dx.doi.org/" + encodeURIComponent(doi);
 								label.setAttribute("isButton", true);
 								label.setAttribute("onclick", "ZoteroPane_Local.loadURI('" + doi + "', event)");
 								label.setAttribute("tooltiptext", Zotero.getString('locate.online.tooltip'));
--- a/chrome/content/zotero/locateMenu.js
+++ b/chrome/content/zotero/locateMenu.js
@@ -410,7 +410,7 @@ var Zotero_LocateMenu = new function() {
 			if(doi && typeof doi === "string") {
 				doi = Zotero.Utilities.cleanDOI(doi);
 				if(doi) {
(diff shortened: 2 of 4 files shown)


aurimasv:
citeproc.js changes should be submitted [upstream](https://bitbucket.org/fbennett/citeproc-js/)

Might as well also change dx.doi.org to the preferred doi.org

I think zotero links are all https at this point, unless something was overlooked. But even if they are not, zotero.org supports HSTS, so it redirects through https anyway (in up-to-date browsers). I'm not aware of mozilla.org links from Zotero client.

katrinleinweber:
> [...] change dx.doi.org to the preferred doi.org

The [site itself](https://www.doi.org/) `attemps to identify itself with invalid information`, according to Firefox. The cert seems to be for http://cnri.reston.va.us/ which looks legit, and resolving works for me, so I hope we are not risking this error popping up for users.

aurimasv:
Not www.doi.org. Just doi.org.

katrinleinweber:
https://doi.org/ falls back to http://www.doi.org/ for me?!? But the resolver works via https, so I don't think we need to worry about their strange serving of the site itself.

> [...] zotero.org supports HSTS, so it redirects

Yeah, but it's one server-side operation we can save, right? I still found many http://www.zotero.org/ links, in all localised `zotero.properties`-files for example. Same for the gnu.org links in the license sections, and I'll look for more, if there is no disagreement about switching them.

aurimasv:
SSL Labs says their setup is weird, but should work https://www.ssllabs.com/ssltest/analyze.html?d=doi.org (not the www part, but we don't care about that). Though given the risk of ssl certificate failures, I'm not sure that it's worth resolving DOIs over HTTPS. There is very little privacy being lost here and the final redirect can be over http or https regardless and there is nothing we can do about it.

katrinleinweber:
> [...] final redirect can be over http or https regardless

The DOI Foundation & the journals can be lobbied to adjust that. Also, it's IMHO about offering "herd-immunity" so that the privacy relevant https traffic doesn't stand out as s/much ;-)

In any case, what about the normal URLs in descriptions, localisations etc.?

aurimasv:
(didn't notice this from the mobile site, but this PR should probably be against the master branch. 4.0 is mostly frozen. Though wait for @dstillman to pitch in before rebasing)

>Yeah, but it's one server-side operation we can save, right?

The redirect happens client-side based on a pre-loaded list of domains before sending any request to the server (in Firefox and Chrome anyway, I'm not sure about other browsers). We can still adjust the URLs for consistency, though I would take this opportunity to move a lot of these URLs (or at least certain prefixes) to a global variable that we can reference (e.g. `ZOTERO_CONFIG.DOCS_URL`).

>I still found many http://www.zotero.org/ links, in all localised zotero.properties-files for example

I found 4 strings that contain http URLs in the locale files (en-US only, since others are just copies).

>Same for the gnu.org links in the license sections

Those are never exposed to the user, so we're not going to change them just for kicks.

dstillman:
Yeah, what @aurimasv says.

Their setup is definitely weird (and inconsistent), but I haven't seen any errors. Maybe they'll get things a bit more in order before 5.0 comes out.

I believe you'll need to close this PR and open a new one against 5.0. I don't think you can change the PR branch via rebasing.

dstillman:
And to clarify, if you do submit a PR against 5.0 for doc links (which is fine, though as Aurimas says zotero.org links will never go out over HTTP in up-to-date browsers), only the en-US locales should be changed.

katrinleinweber:
OK, thanks for the hints :-)

katrinleinweber:
@aurimasv Submitted upstream: https://bitbucket.org/fbennett/citeproc-js/pull-requests/8/generate-secure-doi-links/diff

PR #813, opened 2015-07-27
Don't drop access date from CSL JSON if "cite URL" pref is unchecked

aurimasv:
Regression from f0bd1e77ffab6dbc7fbd600dc1acc11844aa2e02
Re https://forums.zotero.org/discussion/50652/date-accessed-no-longer-appearing-in-bibliography/

### DIFF
--- a/test/tests/citeTest.js
+++ b/test/tests/citeTest.js
@@ -1,0 +1,48 @@
+describe("Zotero.Cite.System", function() {
+	describe("#retrieveItem()", function() {
+		let article, citeSystem;
+		before(function() {
+			// Load sample item into DB
+			let data = populateDBWithSampleData(loadSampleData('journalArticle'));
+			article = Zotero.Items.get(data.journalArticle.id);
+			
+			// instantiate Zotero.Cite.System object
+			citeSystem = new Zotero.Cite.System();
+		});
+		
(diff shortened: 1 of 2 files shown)


test/tests/citeTest.js
```diff
@@ -0,0 +1,48 @@
...
+			citeSystem = new Zotero.Cite.System();
+		});
+		
+		it('should retrieve item given a Zotero.Item', function() {
+			let cslItem = citeSystem.retrieveItem(article);
+			assert.ok(cslItem);
+			assert.equal(cslItem.id, article.id);
+		});
+		
+		it.skip('should retrieve item given an item URI', function() {
```
aurimasv: this is actually not working and I'm not entirely sure what https://github.com/zotero/zotero/blob/4.0/chrome/content/zotero/xpcom/cite.js#L502 is expecting as input. @simonster do you know off the top of your head before I dig into the word processor code?

simonster: It's expecting an integration session ID and an anonymous item ID, which is how we cite items embedded in the document that are not present in the library.

aurimasv: Any thoughts on how to test it? Doesn't sound like it would be very simple. Also, I think this test is still something we would want to work in the future (no?), so I'll leave it in.

PR #819, opened 2015-08-01
[COinS export]: Avoid exporting editors, translators, etc.

zuphilip:
Discussed at https://forums.zotero.org/discussion/50880/every-contributor-exported-as-author-in-coins-export/#Item_0

 * Contributors which are not authors, such as editors, translators etc., are not anymore exported to au, aufirst, aulast.
 * These contributors cannot be saved in COinS format at all.
 * Fixed export for corporate authors (aucorp).
 * Changed nonexisting attribute "isInstitution" with "fieldMode".

This affects the COinS export as well as the OpenUrl resolving.

### DIFF
--- a/chrome/content/zotero/xpcom/openurl.js
+++ b/chrome/content/zotero/xpcom/openurl.js
@@ -203,25 +203,33 @@ Zotero.OpenURL = new function() {
 			}
 		}
 		
-		if(item.creators && item.creators.length) {
-			// encode first author as first and last
-			var firstCreator = item.creators[0];
-			if(item.itemType == "patent") {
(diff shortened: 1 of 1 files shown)


aurimasv:
Could you create this PR against #537? Then we (or I) will also add unit tests to cover both PRs

zuphilip:
I could make a PR into your branch https://github.com/aurimasv/zotero/tree/coins which is connected to the other issue. Is that what you suggest?

aurimasv:
Yes, that's what I meant.

zuphilip:
Here it is https://github.com/aurimasv/zotero/pull/27

PR #820, opened 2015-08-04
Fix issues with computerProgram citations

aurimasv:
Re https://forums.zotero.org/discussion/50943/error-updating-document-invalid-field-false-in-itemfieldsisbasefield/

### DIFF
--- a/chrome/content/zotero/xpcom/date.js
+++ b/chrome/content/zotero/xpcom/date.js
@@ -330,6 +330,8 @@ Zotero.Date = new function(){
 				}
 				
 				if(date.month) date.month--;		// subtract one for JS style
+				else delete date.month;
+				
 				Zotero.debug("DATE: retrieved with algorithms: "+JSON.stringify(date));
 				
 				parts.push(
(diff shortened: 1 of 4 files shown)


chrome/content/zotero/xpcom/date.js
```diff
@@ -369,7 +371,7 @@ Zotero.Date = new function(){
 		}
 		
 		// MONTH
-		if(!date.month) {
+		if(date.month === undefined) {
```
dstillman: Can you explain these changes?

dstillman: (I understand JS months, etc., but what are these fixing?)

aurimasv: It fixes https://github.com/zotero/zotero/pull/820/files#diff-416fa9a7f56dfe0b3a5afd2238dfb317L1351 because of https://github.com/zotero/zotero/pull/820/files#diff-a098326f7cdaa37adf1398d92b384be5R1615 and the code above guarantees that `date.month` is `undefined`

dstillman:
K, thanks.

PR #824, opened 2015-08-09
Return citation preview as HTML in integration.js

fbennett:
Aims to address reports of raw RTF code in the Classic View citation editor.

  * [problems-with-last-update-firefox-word-2010-not-starter-edition](https://forums.zotero.org/discussion/50595/problems-with-last-update-firefox-word-2010-not-starter-edition)
  * [trouble-editing-citations-in-word-2008](https://forums.zotero.org/discussion/24306/trouble-editing-citations-in-word-2008)
  * [utilizing-show-editor-adds-a-i-in-my-footnote](https://forums.zotero.org/discussion/51037/utilizing-show-editor-adds-a-i-in-my-footnote)

I may not fully understand the code, but it looks like the editor object in styled-textbox.xml is set to deliver either RTF or HTML from HTML source. With this patch, the citeproc-js preview (used for the source) is delivered in HTML, which seems to bring things right.
    
If there are no obvious problems with this, it probably wants some torture testing before it's fully trusted. (For what it's worth, I've been running a copy of Juris-M with this change for a couple of weeks with no apparent problems.)

### DIFF
--- a/chrome/content/zotero/xpcom/integration.js
+++ b/chrome/content/zotero/xpcom/integration.js
@@ -2824,7 +2824,7 @@ Zotero.Integration.Session.prototype.previewCitation = function(citation) {
 	var citationsPre, citationsPost, citationIndices;
 	[citationsPre, citationsPost, citationIndices] = this._getPrePost(citation.properties.zoteroIndex);
 	try {
-		return this.style.previewCitationCluster(citation, citationsPre, citationsPost, "rtf");
+		return this.style.previewCitationCluster(citation, citationsPre, citationsPost, "html");
 	} catch(e) {
 		throw e;
 	}


simonster:
The harder case may be the edit bibliography dialog, where making the editor work with the citeproc-js output would almost certainly require some changes to styled-textbox.xml. I'm not sure whether it's easier to make that code capable of translating citeproc-js's present RTF output to HTML or to make it translate citeproc-js's HTML output (as well as its own HTML output) to RTF.

simonster:
We could also try to make the plugins use HTML instead of RTF, but i think there may be some cases where the Mac Word plugin actually needs to use RTF because it needs to insert a bookmark to know where the inserted data is due to bugs in Word APIs.

PR #826, opened 2015-08-11
fix abbreviation for http://www.ncbi.nlm.nih.gov/nlmcatalog/7505598 (…

adam3smith:
…following NLM)
(The change is from: J. Chem. Soc. [Perkin 1] to J. Chem. Soc. Perkin 1 )


aurimasv:
The thing is that the list is automatically generated, so we'll lose these changes if the list is updated at a later time (and we should keep it up to date). I'm not really sure where the script is to generate the list though.

adam3smith:
yeah, @simonster would have to chime in on that -- I think you had asked about the origins of the list before?

rmzelle:
Simon's addition of the abbreviations file (https://github.com/adam3smith/zotero-1/commit/4d2cee782489e416f81ec1c3a070a2eb27e0515e#diff-26006487a933b5af36ddb7bf205f94b7) seems to predate the https://github.com/citation-style-language/abbreviations repo, so I don't where the JSON came from.

adam3smith:
yes, that's right. The repo is Carles's work on the same issue a bit later.

adam3smith:
@dstillman as per https://forums.zotero.org/discussion/65491/biochemistry-incorrectly-referred-to-as-biochemistry-mosc#latest 
Not sure why Travis is failing (I think becuase this was originally created long time ago against 4.0 branch. Let me know if you need anything else.

PR #828, opened 2015-08-12
Forbid "Attachments" column to be resized

gracile-fr:
You might be reluctant to forbid the user to resize it, but in this case (a graphic column, no text), I think this is better. Each time I resize one column, this one is widened and I never want that, who does?

### DIFF
--- a/chrome/content/zotero/zoteroPane.xul
+++ b/chrome/content/zotero/zoteroPane.xul
@@ -517,7 +517,8 @@
 									class="treecol-image"
 									label="&zotero.tabs.attachments.label;"
 									src="chrome://zotero/skin/attach-small.png"
-									zotero-persist="width ordinal hidden sortActive sortDirection"/>
+									fixed="true"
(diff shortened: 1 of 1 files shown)


chrome/content/zotero/zoteroPane.xul
```diff
@@ -517,6 +517,7 @@
 									class="treecol-image"
 									label="&zotero.tabs.attachments.label;"
 									src="chrome://zotero/skin/attach-small.png"
+									fixed="true"
 									zotero-persist="width ordinal hidden sortActive sortDirection"/>
```
aurimasv: Also need to remove "width" from zotero-persist. If someone has that column at non-standard width, they would be stuck with it.

gracile-fr: Thank you. I was not sure of this one.

dstillman:
Thanks. Can squash this.

gracile-fr:
(done)

PR #831, opened 2015-08-18
make items tree column focused when select a item by running zotero://select/items/xxx.

simpzan:
It's really convenient to find an item using ZotQuery on Mac. 
But there's a minor issue is that when the original focus is not on item tree column, select item will not focus the item tree column. So I have to use my cursor to make it focused to operate on the item. 
The expected behavior is that when select item, the item tree column should always be focused. 
This is the reasoning behind this commit.


dstillman:
Thanks!

PR #834, opened 2015-08-26
Fix bug in Create Bibliography from Item w/Notes

fbennett:
When the user selects "Create Bibliography from Item(s)" with the "Notes" option, some entries come out incorrectly, when the input items require disambiguation in the selected style.

[HTML export of Presentation items with same author a](https://forums.zotero.org/discussion/51346/html-export-of-presentation-items-with-same-author-and-title-creates-duplicates/#Item_4)

In this case, Zotero is iterating over appendCitationCluster() to build the citations list. The lack of documentation on appendCitationCluster() is the root source of the issue - what the function does is a little more complex than its name implies.

Under the hood, appendCitationCluster() invokes processCitationCluster() to generate citations to be added, and just strips off the error segment from the return. What is not obvious is that processCitationCluster() function returns multiple citations when disambiguation is required. In Zotero code, the first-listed citation in the return from appendCitationCluster() is always picked. When input does not require disambiguation, this will always be correct. When disambiguation is performed, it will be wrong.

With this patch, all items returned from appendCitationCluster() are inserted in the appropriate locations of the formatted citation list, so that disambiguation operations are fully reflected in the return.

### DIFF
--- a/chrome/content/zotero/xpcom/cite.js
+++ b/chrome/content/zotero/xpcom/cite.js
@@ -81,8 +81,15 @@ Zotero.Cite = {
 		}
 		
 		var styleClass = cslEngine.opt.class;
-		var citations = [cslEngine.appendCitationCluster({"citationItems":[{"id":item.id}], "properties":{}}, true)[0][1]
-			for each(item in items)];
+		var citations=[];
+		for (var i=0, ilen=items.length; i<ilen; i++) {
+			var item = items[i];
+			var outList = cslEngine.appendCitationCluster({"citationItems":[{"id":item.id}], "properties":{}}, true);
+			for (var j=0, jlen=outList.length; j<jlen; j++) {
+				var citationPos = outList[j][0];
+				citations[citationPos] = outList[j][1];
+			}
+		}
 		
 		if(styleClass == "note") {
 			if(format == "html") {


fbennett:
@dstillman @aurimasv Possible to review?

This one is a straightforward bug fix; the processor is currently called incorrectly when a bibliography listing is generated from the cs:citation segment of a style. Any listing with disambiguated entries is likely to be affected.

fbennett:
Thanks!

dstillman:
This causes duplicate results when generating notes:

https://forums.zotero.org/discussion/52886/

dstillman:
@fbennett: Any thoughts on this?

fbennett:
Well, that doesn't sound good. Checking soon.

fbennett:
Within copyItemsToClipboard() in fileInterface.js, the processor is instantiated once, and then the function `Zotero.Cite.makeFormattedBibliographyOrCitationList()` is invoked on it twice, once for `html`, and once for `text`. The function is now non-destructive, so the second invocation registers a second set of items in the processor.

For a fix, we could:

* Adjust `makeFormattedBibliographyOrCitationList()` to make it always clear the processor registry on each invocation (which seems the right policy); or 

* Instantiate the processor afresh in `copyItemsToClipboard()` immediately before the second invocation of `makeFormattedBibliographyOrCitationList()` (a direct fix for this clipboard-copy bug, but seems hackish).

It's my bug, so I can set up a pull request to fix it. Which direction should I go in?

dstillman:
> Adjust makeFormattedBibliographyOrCitationList() to make it always clear the processor registry on each invocation (which seems the right policy)

That seems right to me.

fbennett:
It turns out that we need to instantiate cslEngine outside of `makeFormattedBibliographyOrCitationList()` to pick up the locale, which is controlled by the calling context. I tried emptying the processor registry with calls to its `updateItems()` and `updateUncitedItems()` methods with an empty array, but for some reason that doesn't take effect. Reinstantiating the processor is the only working fix of the two above.

An alternative would be to alter `makeFormattedBibliographyOrCitationList()` so that it reruns the processor with existing content if the `items` argument is empty. That would be a lot more efficient than reinstantiating, but it would introduce implicit behaviour that you may find yet one step more hackish. Which solution seems best?

(Sorry for all the queries; but asking here first seems better than cluttering up the tracker with wannabe pull requests.)

dstillman:
If reusing the engine isn't going to work, could we just change `makeFormattedBibliographyOrCitationList()` to take the locale directly and instantiate the engine within that function?

If so, we can reinstantiate in `fileInterface.js` in 4.0 and change the signature in 5.0 to take a locale directly.

fbennett:
Sounds good. I'll try to put something together.

PR #847, opened 2015-09-21
Upgrade citeproc-js to version 1.1.57 (name particles)

fbennett:
This processor version changes the name particle parsing function `citeproc.parseParticles()`, invoked in Zotero via #806. The existing version is based on an explicit particle list assembled by contributors to two extensive forum threads, [here](https://forums.zotero.org/discussion/51491/double-surnames-starting-with-te-end-up-quoted-incorrect/) and  (more recently) [here](https://forums.zotero.org/discussion/51783/parsing-problem-on-italian-names/).

To recap CSL name particle conventions, leading particles in the `lastName` field are ordinarily classified as "non-dropping," and trailing particles in the `firstName` field as "dropping." The list mentioned in the preceding paragraph specifies the acceptable classifications for the elements in a set of particles, and imposes a default classification when the input does not conform. For example, the Dutch particle "het" is specified as non-dropping *only*, and so will be treated as such, even if it appears at the end of the `firstName` field.

Everyone (including me) agrees that forcibly reclassifying particles in the processor can be confusing to users, and should be abandoned. Also, extensive discussion (in response to scepticism expressed by yours truly) turned up zero examples of uppercase name "particles" that are not actually permanently fixed, sortable elements of the surname itself.

This change follows the summary above by simply treating all lowercase leading elements in the `lastName` field as non-dropping particles, and all trailing elements in the `firstName` field as dropping particles. It also corrects a small bug, noted in the second of the threads linked above, whereby a space separating an apostrophe-terminated particle from its following sibling was quashed in output. With this revision, the space in that instance will be preserved if present in the input.

While this change will alter citation output for some users, the behaviour is much easier to explain and to understand. The following text can be added to the Knowledge Base on `zotero.org` to cover the (probably rare) event of user confusion:

> **Name Particles**
> 
> Some names include lowercase elements that require special treatment when formatting citations 
> and bibliography entries. These elements are referred to as "particles," and the Citation Style 
> Language (CSL) has a simple set of rules for classifying them:
> 
> 1. Lowercase elements at the beginning of the "last name" field are **non-dropping particles**.
> 2. Lowercase elements at the end of the "first name" field are **dropping particles**.
> 3. When the "last name" field is entirely enclosed in double quotation marks, it is treated as a
> **fixed surname**, and any leading lowercase elements it contains are treated as a fixed part of 
> the last name. This is an exception to rule (1) above.

The particle list used by the current particle parsing method is still supplied on the processor source object, since it may be useful to projects, for the construction of parsers and UI involving name particles.

A draft name particle UI extension for Zotero will be the subject of a separate pull request.


aurimasv:
You should use https://github.com/zotero/zotero-build/blob/master/update-citeproc.sh to update citeproc-js in Zotero. All of the CSL_CHROME code was moved out into a separate file.

But I think that @dstillman updates it right before release anyway, now that there is a script for it. I guess it would be good to beta test this though.

fbennett:
Okay. I'll just fix the pull request to lose that part of the code.

fbennett:
@dstillman To reduce clutter, I guess I'll withdraw this along with #848, since the change here will come through in your scripted update anyway as @aurimasv notes.

PR #848, opened 2015-09-21
Name particles menu

fbennett:
This change introduces menu-driven adjustment of name particles, as discussed in two extensive threads on the subject, [here](https://forums.zotero.org/discussion/51783/parsing-problem-on-italian-names/) and (more recently) [here](https://forums.zotero.org/discussion/51783/parsing-problem-on-italian-names/). It depends on #847, which can be merged without regard to this change.

The operation of the menus is illustrated in [a screencast](http://our.law.nagoya-u.ac.jp/howto/names-ui-demo.webm). Source for user-level documentation for `zotero.org` in the event of merge is [here](https://github.com/Juris-M/zotero/wiki). For immediate reference, the text is as follows:

> **Name Particles**
> 
> *Classification of particles*
> 
> Some names include lowercase elements that require special treatment when formatting citations and bibliography entries. These elements are referred to as "particles," and the Citation Style Language (CSL) has a simple set of rules for classifying them:
> 
> 1. Lowercase elements at the beginning of the "last name" field are **non-dropping particles**.
> 2. Lowercase elements at the end of the "first name" field are **dropping particles**.
> 3. When the "last name" field is entirely enclosed in double quotation marks, it is treated as a **fixed surname**, and any leading lowercase elements it contains are treated as a fixed part of the last name. This is an exception to rule (1) above.
> 
> These are the rules applied by CSL, but there is currently no uniform standard for the expression of name particles in bibliographic data. As a result, Zotero items acquired from other sources often contain names that are incorrectly formatted. To produce correctly formatted citations and bibliography entries, and to avoid errors in name disambiguation, name particles must be adjusted to their correct form.
> 
> The classification of particles is influenced by several factors that cannot be easily derived from sparse bibliographic data, such as the national origin of the author, and the number of syllables in the author's surname. Names are also ultimately personal, and an author's preferred way of writing their own name is a final, determinative factor. For these reasons, the classification of particles cannot be completely automated; it is up to the curator of a library to adjust individual names to their correct form.
> 
> *The "Adjust Particles" context menu*
> 
> Item creator fields in Zotero two-field mode have an **Adjust Particles** option to streamline the task of adjusting name particles. For names that contain no particles, the option is disabled. When particle candidates are recognized, the option opens a menu of possible particle layouts. Options are presented under four headings:
> 
> * **Non-dropping particle:** The "last name" field contains at least one non-dropping particle
> * **Dropping particle:** All particles are in the "first name" field, as dropping particles
> * **No particle:** The first element (or all elements) of a non-dropping particle set is capitalized. In this case, the CSL processor will not treat the element as a name particle when generating citations.
> * **Fixed surname:** The "last name" field, including leading lowercase elements, is enclosed in double quotes. As indicated above, such a surname is treated as a fixed unit, and the lowercase elements will always be included, both when sorting names and when generating citations.
> 
> *Particle recognition and the context menu*
> 
> Particles are recognized when either of two conditions is satisfied:
> 
> * Leading *lowercase* elements on the "last name" field and trailing *lowercase* elements on the "first name" field are *always* treated as particles. These are referred to as the "core particles" of an entry.
> * The (possibly empty) set of "core particles" will be extended to include further elements if, and only if, the extended set matches an internal list of particles held by Zotero.
> 
> A match to the internal list is referred to as a **specified particle set**. In this case, the non-dropping/dropping classifications offered in the menu are limited to the known possibilities for the matched set, and **No particle** combinations are included.
> 
> If even the "core particles" fail to match the Zotero list, it is treated as an **unspecified particle set**. In this case, all possible combinations of the core particles are presented in the menu, but the **No particle** combinations (i.e. uppercase forms) are omitted.

### DIFF
--- a/chrome/content/zotero/bindings/itembox.xml
+++ b/chrome/content/zotero/bindings/itembox.xml
@@ -1661,6 +1661,511 @@
 			</method>
 			
 			
+			<method name="getNameParticleVariants">
+			  <parameter name="lastNameValue"/>
+			  <parameter name="firstNameValue"/>
+			  <body>
+				<![CDATA[
+					function splitParticles(nameValue, firstNameFlag, caseOverride) {
+						// Parse particles out from name fields.
+						// * nameValue (string) is the field content to be parsed.
+						// * firstNameFlag (boolean) parse trailing particles
+						//	 (default is to parse leading particles)
+						// * caseOverride (boolean) include all but one word in particle set
+						//	 (default is to include only words with lowercase first char)
+						// Returns an array with:
+						// * (boolean) flag indicating whether a particle was found
+						// * (string) the name after removal of particles
+						// * (array) the list of particles found
+						var origNameValue = nameValue;
+						nameValue = caseOverride ? nameValue.toLowerCase() : nameValue;
+						var particleList = [];
+						var apostrophe;
+						if (firstNameFlag) {
+							apostrophe ="\u02bb";
+							nameValue = nameValue.split("").reverse().join("");
+						} else {
+							apostrophe ="-\u2019";
+						}
+						var rex = new RegExp("^([^ ]+[" + apostrophe + " \'] *)(.+)$");
+						var m = nameValue.match(rex);
+						while (m) {
+							var m1 = firstNameFlag ? m[1].split("").reverse().join("") : m[1];
+							var firstChar = m ? m1 : false;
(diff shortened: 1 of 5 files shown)


fbennett:
https://forums.zotero.org/discussion/51783/2/parsing-problem-on-italian-names/#Item_36

fbennett:
:bomb:

Issue #552, opened 2014-10-05
Use %Documents%/Zotero Data as default data directory location


dstillman:
Follow up from https://github.com/zotero/zotero/issues/551#issuecomment-57949989

bwiernik:
I can confirm that folders placed in the root of the Users folder (or the root of the OS drive) on Windows 8 and 10 are retained through a system reset/refresh if the user chooses for files to be retained (but obviously not if they select to keep nothing).

dstillman:
Closed by 79700969e15f604f3a05781287468d14d8f1936b

PR #867, opened 2015-11-04
Detect when data directory is in Dropbox

### LINKED ISSUE #437: Detect when data directory is in Dropbox
Just show an error (with "Don't show again"?) or refuse to start?

adomasven:
Closes #437


chrome/content/zotero/xpcom/file.js
```diff
@@ -1024,4 +1024,8 @@ Zotero.File = new function(){
 		
 		throw e;
 	}
+
+	this.isDropboxDirectory = function(path) {
+		return path.toLowerCase().includes('dropbox');
```
dstillman: `String.prototype.includes()` is Firefox 40, but the last Firefox ESR is 38 and the next (45) isn't [until March](https://www.mozilla.org/en-US/firefox/organizations/faq/), so unfortunately we're stuck with `indexOf()` for now.

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -381,7 +381,26 @@ Components.utils.import("resource://gre/modules/osfile.jsm");
 					return false;
 				}
 			}
-			
+			if(Zotero.File.isDropboxDirectory(dataDir.path) && Zotero.Prefs.get('warnDropboxDataDir')){
```
aurimasv: Do we need this pref?

aurimasv: (as in, how often do people change data dirs that we need to make sure not to bother them again?)

dstillman: Pref is only for the startup warning. Setting from prefs doesn't have the checkbox.

We could just always show at startup, but in theory if you really know what you're doing and don't have another computer it's OK.

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -381,7 +381,26 @@ Components.utils.import("resource://gre/modules/osfile.jsm");
...
+					Zotero.getString('general.dontShowWarningAgain'),
+					check);
+
+					// Yes - display dialog.
+					if (index == 0) {
+						Zotero.chooseZoteroDirectory(true);
+					}
+					if(check.value === true) {
+						Zotero.Prefs.set('warnDropboxDataDir', false);
+					}
```
aurimasv: nit: indents

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -381,7 +381,26 @@ Components.utils.import("resource://gre/modules/osfile.jsm");
 					return false;
 				}
 			}
-			
+			if(Zotero.File.isDropboxDirectory(dataDir.path) && Zotero.Prefs.get('warnDropboxDataDir')){
+				let check = {value: false};
+				ps = Services.prompt;
```
aurimasv: why assign this at all?

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -381,7 +381,26 @@ Components.utils.import("resource://gre/modules/osfile.jsm");
...
+				index = ps.confirmEx(null,
```
aurimasv: This could end up as global. Add `var` (or `let`)

chrome/content/zotero/xpcom/zotero.js
aurimasv: I guess the use of `ps` below, though given the brevity of `Services.prompt`, this seems unnecessary

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -111,7 +111,11 @@ attachmentBasePath.clearBasePath.button                        = Clear Base Dire
 dataDir.notFound						= The Zotero data directory could not be found.
 dataDir.previousDir					= Previous directory:
 dataDir.useProfileDir				= Use %S profile directory
+dataDir.existingDirInDropbox.title  = Data Directory In Dropbox
+dataDir.existingDirInDropbox.text   = Your current data directory appears to be in a Dropbox directory. This might lead to data corruption.\n\nChange Zotero data directory now?
 dataDir.selectDir					= Select a Zotero data directory
+dataDir.selectedDirInDropbox.title		= Directory In Dropbox
```
aurimasv: Does this need to be different from above? Also, I'd suggest "Inside" instead of "In"

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -111,7 +111,11 @@ attachmentBasePath.clearBasePath.button                        = Clear Base Dire
 dataDir.notFound						= The Zotero data directory could not be found.
 dataDir.previousDir					= Previous directory:
 dataDir.useProfileDir				= Use %S profile directory
+dataDir.existingDirInDropbox.title  = Data Directory In Dropbox
+dataDir.existingDirInDropbox.text   = Your current data directory appears to be in a Dropbox directory. This might lead to data corruption.\n\nChange Zotero data directory now?
```
aurimasv: "may" instead of "might", I think

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -1069,53 +1088,44 @@ Components.utils.import("resource://gre/modules/osfile.jsm");
...
-							}
+							dialogTitle = Zotero.getString('dataDir.selectedDirNonEmpty.title');
+							dialogText = Zotero.getString('dataDir.selectedDirNonEmpty.text');
 						}
 					}
-					else {
+					else { // Directory empty
+						dialogTitle = Zotero.getString('dataDir.selectedDirEmpty.title');
+						dialogText = Zotero.getString('dataDir.selectedDirEmpty.text', Zotero.appName) + '\n\n'
+								+ Zotero.getString('dataDir.selectedDirEmpty.useNewDir');
```
aurimasv: Maybe as part of another PR, but seems like this is very... primitive. We should simply ask the user if they wish to move their existing library to the selected directory or if they want to create a new library at the new location and then perform the move ourselves when Zotero restarts.

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -1069,53 +1088,44 @@ Components.utils.import("resource://gre/modules/osfile.jsm");
...
+					}
+					// Warning dialog to be displayed
+					if(dialogText !== ''){
```
aurimasv: nit: missing spaces

aurimasv: also a few other places above

chrome/content/zotero/xpcom/zotero.js
dstillman: Yeah, this is old code, so we can discuss that separately. (It'd be great to do, but we didn't do it in the past because we don't have a native method for moving a directory, and it's somewhat of a scary proposition — anytime we touch files we end up with user reports because of security software and the like. At the very least, we'd need a function that did a whole lot of tracking and cleanup on errors. We also need to shut down everything beforehand and be prepared to start it up again after recovering from an error, though that part, at least, we now have the ability to do.)

chrome/content/zotero/xpcom/zotero.js
aurimasv: Yes, that makes sense for startup. Didn't look at this carefully enough. I'd rename it to "checkDataDirLocationOnStartup" (for clarity and because there might be other services to check for in the future) and move it to be the first check in the `if` statement.

chrome/locale/en-US/zotero/zotero.properties
dstillman: If we went specific, I'd think we'd want something like "Unsafe Data Directory Location" (since we might apply this to other locations in the future), but I think we could just use general.warning ("Warning") here.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -111,7 +111,11 @@ attachmentBasePath.clearBasePath.button                        = Clear Base Dire
...
 dataDir.selectDir					= Select a Zotero data directory
+dataDir.selectedDirInDropbox.title		= Directory In Dropbox
+dataDir.selectedDirInDropbox.text		= The directory you selected appears to be in a Dropbox directory. This might lead to data corruption.\n\nCreate Zotero files in this directory anyway?
```
dstillman: This is accurate, but we can probably just assume that it's Dropbox. (If someone has a path with "dropbox" in it that's not Dropbox, they'll probably understand, in any case.)

How about: "Choosing a data directory within Dropbox may corrupt your Zotero database.\n\nUse this directory anyway?"

dstillman:
The pref alert looks good.

After trying the startup one, I think it can be modal before DB access in Standalone, but it probably shouldn't be modal in Firefox — there I think it'd be better to wait until there's some interaction with Zotero. (Yes, that's after some DB access, but since this one is only for existing users anyway, they've already had plenty of chance to do damage.) So the prompt logic can go in a `checkDataDirectory()` function after `chooseZoteroDirectory` in zotero.js, and it can be called either at Zotero startup for Zotero.isStandalone or from maybe `toggleDisplay()` in overlay.js. For the latter, `checkDataDirectory()` can set a flag so that it only runs once per session.

(That will also avoid having to use different text — the current message only mentions Zotero at the end, so it'd be a little confusing.)

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -1069,53 +1088,44 @@ Components.utils.import("resource://gre/modules/osfile.jsm");
...
-							Zotero.getString('dataDir.selectedDirEmpty.title'),
-							Zotero.getString('dataDir.selectedDirEmpty.text', Zotero.appName) + '\n\n'
-								+ Zotero.getString('dataDir.selectedDirEmpty.useNewDir'),
+							dialogTitle,
+							dialogText,
 							buttonFlags,
 							null,
 							null,
-							moreInfoCallback ? Zotero.getString('general.moreInformation') : null,
+							moreInfoCallback ? Zotero.getString('general.help') : null,
```
dstillman: I know it was in the old code, but general.help doesn't actually exist, so this can stay general.moreInformation.

chrome/content/zotero/xpcom/zotero.js
dstillman: Use this style to avoid mistakes like that:

```
var index = ps.confirmEx(
	null,
	dialogTitle,
	dialogText,
	buttonFlags,
	null,
	null,
	moreInfoCallback ? Zotero.getString('general.help') : null,
	null,
	{}
);
```

chrome/content/zotero/xpcom/zotero.js
dstillman: (I agree this is really bad, though, and this check will cause it to come up more for existing users. Maybe when we're done here @adomasven can investigate the feasibility of a `Zotero.File.moveDirectory()` that we can use for this and also for #552.)

dstillman:
So @adomasven, you can push changes back to this branch. When it's ready to merge, we'll have you do a rebase, squash, and force-push to clean up the extra commits.

chrome/content/zotero/xpcom/zotero.js
adomasven: It is declared above as global for the function with `var`. IDE was complaining about multiple declarations.

aurimasv: Correct, but all declarations are inside conditionals, so conceivably it may be possible for it to end up undeclared. Also, if something is declared within a block, removing that block should have no effect on that same variable somewhere else. Here it would. It's ok to re-declare variables with `var`, the IDE should only be warning you, not throwing errors. With `let`, it's an error to re-declare the same variable within the same block,  but that wouldn't be an error here either.

adomasven: Per Javascript spec, functions are actually preprocessed before running and all `var` declarations are moved to the front of the function, so even if you have variables declared with `var` inside conditionals which don't run, they are still available within the function scope. Valid concern though, I will turn off the IDE warnings for multiple declarations (as I assume there will is existing code with multiple var declarations) and go for `let` in this case, since we're moving towards ES6.

KarlHegbloom: It seems to me that if you have more than one computer, but you never edit
in more than one place, then drop box isn't really going to cause any
trouble. Perhaps Zotero should use the similar solution that SMTP servers
use, in coordinating with MUA.

dstillman: @KarlHegbloom This is a line note on a pull request. If you have a comment on the proposed implementation you can make it on the [ticket](https://github.com/zotero/zotero/issues/437). Don't know what you mean about SMTP, though. In any case, I'm deleting this.

adomasven:
Changes pushed. Also fixed a stray bug, where clicking the *Choose* button would not change the radio button properly.
image

dstillman:
Now if a Custom location is already set, and I click on Choose, I get the "Restart Required" prompt without getting a filepicker.

chrome/content/zotero/overlay.js
```diff
@@ -209,7 +210,14 @@ var ZoteroOverlay = new function()
 			
 			// Make visible
 			ZoteroPane.makeVisible();
-			
+
+			// Warn about unsafe data directory on first display
+			if(Zotero.Prefs.get('warnUnsafeDataDirOnStartup') && this.warnUnsafeDataDir) {
```
dstillman: These checks should just go in Zotero.checkDataDirectory(), with the flag set on the `Zotero` object, and this code should just call Zotero.checkDataDirectory(). (Function overhead doesn't matter.) There's only one `Zotero`, but there can be multiple `ZoteroOverlay`s, so this will reappear if a second window was opened.

chrome/content/zotero/xpcom/file.js
```diff
@@ -1024,4 +1024,9 @@ Zotero.File = new function(){
 		
 		throw e;
 	}
+
+
+	this.isDropboxDirectory = function(path) {
+		return path.toLowerCase().indexOf('dropbox') >= 0;
```
dstillman: Nit: `!= -1` is our usual pattern

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -111,7 +111,9 @@ attachmentBasePath.clearBasePath.button                        = Clear Base Dire
 dataDir.notFound						= The Zotero data directory could not be found.
 dataDir.previousDir					= Previous directory:
 dataDir.useProfileDir				= Use %S profile directory
+dataDir.existingDirInDropbox.text   = Your current data directory appears to be in a Dropbox directory. This may lead to data corruption.\n\nChange Zotero data directory now?
 dataDir.selectDir					= Select a Zotero data directory
+dataDir.selectedDirInDropbox.text		= Choosing a data directory within Dropbox may corrupt your database.\n\nUse this directory anyway?
```
dstillman: Let's actually generalize these for use with other services later:

```
dataDir.unsafeLocation.selected.dropbox = Choosing a data directory within Dropbox may corrupt your database.
dataDir.unsafeLocation.selected.useAnyway = Use this directory anyway?
dataDir.unsafeLocation.existing.dropbox = Your Zotero data directory is within Dropbox, which may lead to data corruption.
dataDir.unsafeLocation.existing.chooseDifferent = Would you like to choose a different location now?
```

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -1162,6 +1154,30 @@ Components.utils.import("resource://gre/modules/osfile.jsm");
...
+				null, null, null,
+				Zotero.getString('general.dontShowWarningAgain'),
+				check
+			);
+
+			// Yes - display dialog.
+			if (index == 0) {
+				Zotero.chooseZoteroDirectory(true);
+			}
+			if(check.value === true) {
```
dstillman: Can just be `if (check.value)` (and nit: space after `if` here and elsewhere — space after keyword, not after function name)

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -1162,6 +1154,30 @@ Components.utils.import("resource://gre/modules/osfile.jsm");
 		
 		return useProfileDir ? true : file;
 	}
+
+
+	this.checkDataDirectory = function (path){
+		if(Zotero.File.isDropboxDirectory(path)) {
```
dstillman: Nit: inconsistent spacing. Space after `if` and before `{` in the first line.

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -1069,53 +1071,44 @@ Components.utils.import("resource://gre/modules/osfile.jsm");
...
-								}, 1);
-								return false;
-							}
+							dialogTitle = Zotero.getString('dataDir.selectedDirNonEmpty.title');
+							dialogText = Zotero.getString('dataDir.selectedDirNonEmpty.text');
 						}
 					}
-					else {
-						var buttonFlags = ps.STD_YES_NO_BUTTONS;
+					else { // Directory empty
```
dstillman: Style: comment on the line above

chrome/content/zotero/overlay.js
dstillman: And let's change pref to `warnOnUnsafeDataDir`.

chrome/content/zotero/overlay.js
```diff
@@ -188,7 +189,7 @@ var ZoteroOverlay = new function()
 		Zotero.debug("zoteroPane.getAttribute('minheight'): " + zoteroPane.getAttribute('minheight'));
 		Zotero.debug("savedHeight: " + savedHeight);
 		*/
-		
+
```
aurimasv: nit: undo

chrome/content/zotero/overlay.js
aurimasv: Also, might as well flip the checks, since `this.warnOnUnsafeDataDir` is dirt cheap.

chrome/content/zotero/overlay.js
adomasven: Once again, my IDE does removes trailing whitespace automatically for the lines I touch. It is usually a good thing too: less whitespace means smaller files. I was looking into turning it off, but decided not to, since it probably does more good than bad. Opinions?

dstillman: The note is because PRs generally shouldn't change unrelated lines, even for stylistic reasons — you can use `git checkout -p` (or `git reset -p` to keep) to get rid of any lines in the diff that don't belong.

In terms of whitespace, file size isn't a concern. We're not religious about this in submitted code, but we generally leave whitespace-line indents in place. In my opinion it makes for easier editing — this way if you go to add a line between two indented blocks, you're already at the right indent level without having to hit Tab or go up to the end of the block above.

adomasven:
Updated. Also, made the radio button functionality more intuitive. Now it only prompts with filepicker if you click the *Choose...* button, otherwise it prompts for restart (and keeps the custom data directory). Also does nothing, if you click on the option which is already selected. Appears to be working every way now.

chrome/content/zotero/preferences/preferences_advanced.js
```diff
@@ -198,24 +198,11 @@ Zotero_Preferences.Advanced = {
 	
 	onDataDirUpdate: function (event) {
 		var radiogroup = document.getElementById('dataDir');
-		var path = document.getElementById('dataDirPath');
 		var useDataDir = Zotero.Prefs.get('useDataDir');
+		var newUseDataDir = radiogroup.selectedIndex === 1;
```
aurimasv: nit: unless required, prefer `==`

chrome/content/zotero/preferences/preferences_advanced.xul
```diff
@@ -181,10 +181,7 @@
 									onsyncfrompreference="return Zotero_Preferences.Advanced.getDataDirPath();"
 									readonly="true" flex="1"/>
 								<button label="&zotero.preferences.dataDir.choose;"
-									oncommand="var file = Zotero.chooseZoteroDirectory(true);
-										if (!file) {
-											event.stopPropagation();
-										}"/>
+										oncommand="return Zotero_Preferences.Advanced.chooseDataDir(event)"/>
```
aurimasv: nit: indents

dstillman:
> Also, made the radio button functionality more intuitive. Now it only prompts with filepicker if you click the Choose... button, otherwise it prompts for restart (and keeps the custom data directory).

I think we actually want to keep the existing behavior of showing the filepicker in that case too — changing the data directory is a rather serious affair, so it's pretty important to give people the chance to set the directory before restarting. The filepicker opens to the existing directory anyway, so it's easy to just click through the file picker if that's really what you want, and people generally aren't going to be toggling between the two settings such that skipping the picker would be regularly helpful.

(If we made any change, I think it would be to just clear the custom location path when the custom option was deselected. It's probably sometimes useful when people are messing around with things to remind them where they had it set previously, though.)

adomasven:
Ah, it opens up in "Recently Used" for me on Ubuntu, which I was finding rather unintuitive, hence the decision. But given our primary target is Mac people and you saying that, it is best to keep it as it is then. Although clearing the field is definitely a good idea.

dstillman:
Actually, I was wrong — it doesn't do that currently, because we don't set `fp.displayDirectory`. So we could do that, but in any case better to give people the choice.

adomasven:
Updated. Restored radio button behaviour.

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -1156,12 +1151,41 @@ Components.utils.import("resource://gre/modules/osfile.jsm");
...
+		if (forceQuitNow || index == 0) {
 			Services.startup.quit(Components.interfaces.nsIAppStartup.eAttemptQuit);
 		}
 		
 		return useProfileDir ? true : file;
 	}
+
+
+	this.warnOnUnsafeDataDir = true;
+	this.checkDataDirectory = function (path) {
```
dstillman: Let's actually call this `checkForUnsafeDataDirectory()` to be clear, since there are lots of other checks we could be doing.

dstillman:
Other than that name change, this looks great. You can squash and force-push.

adomasven:
Name changed and squash force-pushed.

dstillman:
Thanks. Can you rewrite the commit message?

adomasven:
Whoop, yes, the commit message could have been better. Fixed now.

PR #869, opened 2015-11-07
Relax condition in parseContextObject

### LINKED ISSUE #851: Relax condition in parseContextObject
It looks that sometimes the COinS data from some sources are containing the elements `rft.subject` and `rft.description` also they aren't technically valid in the choosen `rft_val_fmt`, see <a href="https://forums.zotero.org/discussion/52676/incomplete-coins-import-from-german-education-index/">discussion in the forum</a>. However, there does not seem anything aggainst not accepting this as well for importing in web translators or in general when `parseContextObject` is used.

As far as I see, we could just delete the condition <a href="https://github.com/zotero/zotero/blob/master/chrome/content/zotero/xpcom/openurl.js#L441">here</a>
```
	} else if(format == "info:ofi/fmt:kev:mtx:dc") {
```
and accept the following elements for all values of `format`. Would that be okay?

zuphilip:
Especially, allow rft.subject and rft.description in core OpenURL without Dublin Core declared as format. This fixes #851

### DIFF
--- a/chrome/content/zotero/xpcom/openurl.js
+++ b/chrome/content/zotero/xpcom/openurl.js
@@ -438,7 +438,10 @@ Zotero.OpenURL = new function() {
 				} else if(key == "rft.appldate") {
 					item.date = value;
 				}
-			} else if(format == "info:ofi/fmt:kev:mtx:dc") {
+			} else {
+				// The following keys are technically only valid in Dublin Core
+				// (i.e., format == "info:ofi/fmt:kev:mtx:dc") but in practice
+				// 'format' is not always set
 				if(key == "rft.identifier") {
 					if(value.length > 8) {	// we could check length separately for
 											// each type, but all of these identifiers


adam3smith:
FWIW, I'd be happy to have this in.

chrome/content/zotero/xpcom/openurl.js
```diff
@@ -438,7 +438,10 @@ Zotero.OpenURL = new function() {
 				} else if(key == "rft.appldate") {
 					item.date = value;
 				}
-			} else if(format == "info:ofi/fmt:kev:mtx:dc") {
+			} else {
+				// the following keys are technically only valid in Dublin Core
+				// i.e. format == "info:ofi/fmt:kev:mtx:dc"
+				// however in practice this is not always set thorough
```
dstillman: Cleanup:

> // The following keys are technically only valid in Dublin Core
> // (i.e., format == "info:ofi/fmt:kev:mtx:dc") but in practice
> // 'format' is not always set

(Can squash that and force push.)

zuphilip:
Done

dstillman:
Thanks! (Cherry-picked to 4.0 as well in add3acbcd153cbe13a1eafb59fd9dbc05e1c94d0.)

PR #870, opened 2015-11-08
Prevent duplicate items when copying citations to clipboard

fbennett:
This is a followup to #834, which turned out to be a little less than the "straightforward bug fix" I declared it to be.

This PR addresses the duplicate-cite issue reported [on the Zotero forums](https://forums.zotero.org/discussion/52886/).

As per discussion under #834, this fix will be specific to the 4.0 branch. In 5.0, the plan will be to reduce clutter by changing the signature of `makeFormattedBibliographyOrCitationList()` to receive the locale (plus the style), and instantiate the processor in that function.


dstillman:
Thanks!

PR #872, opened 2015-11-17
Improves item dragging UX

### LINKED ISSUE #36: Allow parent item drag to collection if all children are selected
Right now one test for dragging to a collection is whether any child items are selected, in which case the drag is disallowed. Selecting all child items should be equivalent to selecting just the parent.

http://forums.zotero.org/discussion/21083/fails-dragging-multiple-items-to-a-collection/#Item_11

adomasven:
This PR includes:
- Alert when draging invalid items to My Publications. The new method may be extended for other dragging cases as well.
- Read-only icons for read-only collections (to indicate that dragging is disallowed)
- Fixes #36 . Allows dragging children items to collections when it is reasonable to just assume the parent item.

Waiting for additions, modifications and guidance for better new function locations (if any).


chrome/content/zotero/xpcom/itemTreeView.js
```diff
@@ -3075,6 +3075,37 @@ Zotero.ItemTreeView.prototype.drop = Zotero.Promise.coroutine(function* (row, or
...
+		// https://bugzilla.mozilla.org/show_bug.cgi?id=100180
+		setTimeout(function() { 
+			Services.prompt.alert(null,
+			Zotero.getString('drag.invalid.title'),
+			Zotero.getString('drag.publications.onlyTopLevel')) 
+		}, 50);
+	}
+	
+}
+
```
dstillman: I doubt this belongs here — shouldn't it just go in the drop handling code in collectionTreeView.js?

Also, the prompt bug you link to was fixed in Firefox 30. You still might need to show the prompt after the drag, but ideally you would be able to omit the delay in the setTimeout call — a low timeout often points to a race condition, and usually what's actually needed is just to run the code at the next turn of the event loop.

Also, style:

```javascript
Services.prompt.alert(
    null,
    Zotero.getString('drag.invalid.title'),
    Zotero.getString('drag.publications.onlyTopLevel')
)
```

Also, better to use `Zotero.alert()`, which skips the alert during tests. (Tests can still use a Sinon spy to make sure alerts are shown instead of resorting to the slower `waitForDialog()`.)

chrome/content/zotero/xpcom/data/items.js
```diff
@@ -625,6 +625,47 @@ Zotero.Items = function() {
...
+		// This is unfortunately 90x (!) faster than any of ES6 syntactic sugar
+		let result = [];
+		for (let id in itemMap) {
+			result.push(itemMap[id])
+		}
+			
+		return result;	
+	}
+	
+
```
dstillman: I don't think this the behavior we want. Dragging a child item shouldn't work, and I don't have a problem with just disallowing the drag in those cases. What #36 specifies is allowing the drag if both the parent item and all child items are selected (as you would get if you selected an range with a parent expanded). To account for the case of a search, where some child items were in gray, we should probably allow for any number of child items of selected parents. But I don't think we don't want to allow dragging of only child items, because that's confusing.

chrome/content/zotero/xpcom/collectionTreeView.js
```diff
@@ -662,6 +662,9 @@ Zotero.CollectionTreeView.prototype.getImageSrc = function(row, col)
 	
 	switch (collectionType) {
 		case 'library':
+			if(! treeRow.editable) {
```
dstillman: Style nit: For all of these, space after `if` (keyword, not function), no space after `!`.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -293,6 +293,9 @@ pane.item.related.count.singular = %S related:
 pane.item.related.count.plural  = %S related:
 pane.item.parentItem			= Parent Item:
 
+drag.invalid.title 				= Invalid item
+drag.publications.onlyTopLevel	= Only Zotero items are allowed in My Publications
```
dstillman: "Only top-level Zotero items are allowed in My Publications."

(Child items are still Zotero items.)

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -293,6 +293,9 @@ pane.item.related.count.singular = %S related:
 pane.item.related.count.plural  = %S related:
 pane.item.parentItem			= Parent Item:
 
+drag.invalid.title 				= Invalid item
```
dstillman: Capitalized: "Invalid Item" for a title

(Not sure if this merits its own string — every string creates extra localization work — but I don't know if there's an existing `general.` string that would work. "Error" seems a little strong.)

chrome/content/zotero/xpcom/itemTreeView.js
adomasven: Well, the dropend event is called on the item that was being dragged, so it made sense for it to be there.

Yeah, it appears to either not have been properly fixed on Linux or there is another bug. I was testing it and it basically freezes, then displays like 20 of the alerts in one go. How do I go about displaying this in the next turn of the even loop, though?

chrome/locale/en-US/zotero/zotero.properties
adomasven: Maybe move "Invalid Item" to something like `general.invalidItem`? could see this cropping up in different contexts?

chrome/content/zotero/xpcom/itemTreeView.js
dstillman: But all the handling code for the drop is already in collectionTreeView, so that's where this should go.

> How do I go about displaying this in the next turn of the even loop, though?

You just leave out the timeout on the setTimeout().

adomasven: I disagree as this function is called when the item is dragged onto a collection as well as onto another item. It will be extremely confusing later:

- When `onDragEnd` doesn't get called after dragging onto other collections
- When you start dragging items onto other items and a collection `onDragEnd` is being called

It could go in `libraryTreeView`, but it is not really used for the `collectionTreeView`. I'd say it should just stay there.

adomasven:
This is done up, unless you have other comments, I shall squash and push.

dstillman:
Bug: If you drag a top-level item and hit Esc before letting go, you get the dialog erroneously.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -293,6 +293,9 @@ pane.item.related.count.singular = %S related:
 pane.item.related.count.plural  = %S related:
 pane.item.parentItem			= Parent Item:
 
+drag.invalid.title 				= Invalid Item
+drag.publications.onlyTopLevel	= Only top-level Zotero items are allowed in My Publications
```
dstillman: Oh, so top-level notes and attachments can't be dragged to My Publications either, so my changed string doesn't work either. "Notes and attachments cannot be dragged to My Publications independently."?

I don't love that, but we don't have a great term the inverse of those — we call them "regular items" internally, but that's not clear at all.

adomasven: I thought they were called Zotero items. I don't think there really is a good user-friendly name for them unless you do some sort of brandish thing with them, like Zoterios™, but that's just silly.

dstillman: They are, but the problem is just that notes and attachments are Zotero items too.

chrome/content/zotero/xpcom/itemTreeView.js
dstillman: Sorry, not sure I'm following. I wasn't suggesting that you need `onDragEnd` in `collectionTreeView` — I'm saying the code that shows the prompt shouldn't be decoupled from the code that actually blocks the drop, which is in `canDropCheck` in `collectionTreeView`. It doesn't make sense to block in one place and show a prompt to the user in another place.

The only advantage to triggering it from the item side (ignoring the blocking part) is that it could work when you have multiple Zotero panes open and drag from one items list to the other, but 1) that's only possible in Zotero for Firefox (and might not be possible in the future), 2) it doesn't actually work at the moment because we disallow cross-library cross-window drag (rather than duplicate all the `collectionTreeView` logic), and 3) it's possible you could drag a `zotero/item` from somewhere other than an `itemTreeView`, so it makes more sense to handle this from the drop code. If we do re-allow cross-window cross-library drag, we'll need to move the `canDropCheck` code to `libraryTreeView`, but the prompt should still stay with the blocking code, and right now that's in `collectionTreeView`

adomasven: Oh. Okay, so the reason for this is that `canDropCheck` is called in `onDragOver`, not on drop, because it also has to set cursor and underlying item highlation depending on whether the drop is allowed or not. Which means that you get the alert as soon as you hover over *My Publications*.

adomasven:
This one is waiting a final review and merge still with final adjustments made (like the Esc bug fixed)

PR #873, opened 2015-11-18
Remove tags on shift-drag

### LINKED ISSUE #852: Batch remove tag
Hello all
thx for your great software
You added a way by drag and drop to add tags, but nothing to batch remove tags: a feature like this would be very very usefull, especially when someone clumsy batch added an unwanted tag
And I guess it's not so hard to implement it that since the feature of batch adding by drag and drop is already there: I suggest that by pressing Shift key or minus or... and doing a drag and drop it batch deletes the selected tag
Many forum topic -unanswered for most- are begging for such feature so please dear devs, please make our dream real!
^^ thanks for reading

adomasven:
Closes #852

Currently shows a *+* icon for adding a tag and
a *link* icon for removing it.

There does not appear to be any way to have a custom mouse cursor,
which makes this interaction extremely unintuitive. It might be
possible to hack this in, but the issue does not really warrant the effort.
E.g. `window.setCursor` does not respond whilst dragging.

### DIFF
--- a/chrome/content/zotero/bindings/tagselector.xml
+++ b/chrome/content/zotero/bindings/tagselector.xml
@@ -934,13 +934,16 @@
 							}
 						}
 						*/
-						
+
+						// Store the event, because drop event does not have shiftKey attribute set
+						Zotero.DragDrop.currentEvent = event;
 						event.target.setAttribute('draggedOver', true);
 						event.preventDefault();
-						event.effectAllowed = event.dropEffect = "link";
+						event.dataTransfer.dropEffect = "copy";
 					}
 					
 					this.onDragExit = function (event) {
+						Zotero.DragDrop.currentEvent = null;
 						event.target.setAttribute('draggedOver', false);
 					}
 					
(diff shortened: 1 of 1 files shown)


dstillman:
Let's skip the link part and just rely on Shift. The link icon is probably more rather than less confusing. People will just have to read about this in the documentation or try it and get lucky.

dstillman:
(And yeah, I've never been able to get a custom mouse cursor. I don't think I've seen it in Firefox or Thunderbird, which is usually a good sign that it's not possible.)

dstillman:
We could use 'move' instead to just hide the cursor. Maybe that's better?

adomasven:
The *move* cursor has a little arrow for it. Just as confusing. A minus sign would be ideal, but does not seem to be possible, so probably is best to just stick to the same cursor. I'll change it up.

dstillman:
> The move cursor has a little arrow for it.

Not on OS X, but yeah, I guess Linux and Windows might both show a different cursor.

adomasven:
Updated and force-pushed.

PR #875, opened 2015-11-19
Recognizes and opens files even without extensions.

### LINKED ISSUE #842: Opening PDF with stored 'application/pdf' but no .pdf extension doesn't work
Instead of the PDF being displayed internally in pdf.js, a messed-up open/save dialog is shown:

dialog

adomasven:
Closes #842

Also:
Changes AsyncChannel to be a coroutine. Couldn't find any tests for it, but Dan said it was okay to do it and testing on the attachment types I have seems to be fine.

### DIFF
--- a/components/zotero-protocol-handler.js
+++ b/components/zotero-protocol-handler.js
@@ -1149,17 +1149,13 @@ function AsyncChannel(uri, gen) {
 }
 
 AsyncChannel.prototype = {
-	asyncOpen: function (streamListener, context) {
+	asyncOpen: Zotero.Promise.coroutine(function* (streamListener, context) {
 		if (this.loadGroup) this.loadGroup.addRequest(this, null);
 		
 		var channel = this;
 		
 		var resolve;
 		var reject;
-		var promise = new Zotero.Promise(function () {
-			resolve = arguments[0];
-			reject = arguments[1];
-		});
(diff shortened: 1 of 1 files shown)


components/zotero-protocol-handler.js
```diff
@@ -767,6 +767,7 @@ function ZoteroProtocolHandler() {
 							}
 						}
 					}
+					file.extension
```
dstillman: Stray line?

components/zotero-protocol-handler.js
```diff
@@ -1230,9 +1225,8 @@ AsyncChannel.prototype = {
 					Zotero.debug("AsyncChannel: Got URI from generator");
 				}
 				
-				let uri = data;
-				uri.QueryInterface(Ci.nsIURL);
-				this.contentType = Zotero.MIME.getMIMETypeFromExtension(uri.fileExtension);
+				let sample = yield Zotero.File.getSample(data);
+				this.contentType = Zotero.MIME.getMIMETypeFromData(sample);
```
dstillman: If there's an extension, try `getMIMETypeFromExtension` first, to avoid reading the file unnecessarily?

adomasven:
Adjusted. Force-pushed

PR #878, opened 2015-11-28
Recognize parent styles via http and https

rmzelle:
Untested, but this should help Zotero recognize whether a certain independent-parent style is already installed, even if the "independent-parent" link of the dependent style uses "https" while the parent style ID uses "http" (and vice versa).

PS. The “independent-parent” link provides an URL from which the parent of a dependent style can be retrieved. It’s technically not an ID, although it can be used as such for the styles in the CSL repository, since we require that “self” links and style IDs are identical.

PPS. I haven’t been able to test this because of what seems to be a regression in “master”: https://github.com/citation-style-language/styles/issues/1771#issuecomment-160223820

PPS. This was prompted by https://github.com/citation-style-language/styles/issues/1771

### DIFF
--- a/chrome/content/zotero/xpcom/style.js
+++ b/chrome/content/zotero/xpcom/style.js
@@ -403,28 +403,35 @@ Zotero.Styles = new function() {
 			}
 		});
 		
-		// User wants to install/update
-		if(source && !_styles[source]) {
-			// Need to fetch source
-			if(source.substr(0, 7) === "http://" || source.substr(0, 8) === "https://") {
-				try {
-					let xmlhttp = yield Zotero.HTTP.request("GET", source);
-					yield _install(xmlhttp.responseText, origin, true);
-				}
-				catch (e) {
-					if (typeof e === "object" && e instanceof Zotero.Exception.Alert) {
-						throw new Zotero.Exception.Alert(
-							"styles.installSourceError",
-							[origin, source],
-							"styles.install.title",
-							e
-						);
+		// For dependent style, check if independent-parent is installed,
+		// otherwise try to install it
+		if (source) {
+			var httpRegex = new RegExp("/^https?:\/\//i");
+			var httpSource = source.replace(httpRegex,"http://");
+			var httpsSource = source.replace(httpRegex,"https://");
+		
+			if (!_styles[httpSource] || !_styles[httpsSource]) {
+				// Need to fetch source
+				if (source.match(httpRegex)) {
+					try {
+						let xmlhttp = yield Zotero.HTTP.request("GET", source);
+						yield _install(xmlhttp.responseText, origin, true);
 					}
-					throw e;
+					catch (e) {
+						if (typeof e === "object" && e instanceof Zotero.Exception.Alert) {
+							throw new Zotero.Exception.Alert(
+								"styles.installSourceError",
+								[origin, source],
+								"styles.install.title",
+								e
+							);
+						}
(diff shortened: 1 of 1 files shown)


dstillman:
I haven't really studied this, but will this actually work, or will it just download the parent style but then fail when it tries to use the HTTPS parent as an id?

In any case, I'm going to hold this until we can write a test for it.

I was going to post about this issue to the CSL list, though, because it seems like this highlights some odd requirements for `independent-parent`. `independent-parent` needs to be deferenceable, but it also needs to be able to identify the local style, which may or may not have a matching id. There could even be an HTTP redirect in place, so even if the `id` was kept static and the `link` was updated to the new location, it wouldn't match the original `independent-parent`. I believe this means that clients actually need to keep track of the `independent-parent` URI used to download the style and use that for the local parent lookup, regardless of the values in the downloaded parent file.

dstillman:
> In any case, I'm going to hold this until we can write a test for it.

(To clarify, styles don't work on master yet generally.)

rmzelle:
> or will it just download the parent style but then fail when it tries to use the HTTPS parent as an id?

Uhhh, yeah, probably. I haven't tracked down the code yet where an installed parent style is actually called via one of its dependents.

> independent-parent needs to be deferenceable, but it also needs to be able to identify the local style, which may or may not have a matching id. There could even be an HTTP redirect in place, so even if the id was kept static and the link was updated to the new location, it wouldn't match the original independent-parent

With the current scheme, wouldn't it be reasonable to require that the "independent-parent" link must always resolve to the style and match the style ID of the parent? In that case redirects wouldn't be an issue, although cases where this rule isn't followed should throw an error.

rmzelle:
In general, I'm still not very happy with how we deal with IDs and dereferenceable links, by the way. Maybe we should aim to eliminate the web-part from the style ID? E.g. for independent styles we could do something like 

```
<id>apa</id>
<link rel="repository" href="https://zotero.org/styles/"/>
```

and for dependents

```
<id>apa-dependent</id>
<id rel="parent">apa</id>
<link rel="repository" href="https://zotero.org/styles/"/>
```

(we could require that a deferenceable URL can always be constructed by concatenating the repository base URL and the style ID)

This would greatly reduce the ID space, though.

zuphilip:
Actually, I think it is quite nice to use HTTP URIs for identifyer, see also the first two <a href="https://en.wikipedia.org/wiki/Linked_data#Principles">Linked Data Principles</a>. What is currently the idea behind having `<id>` and as well `<link rel="self">`?

rmzelle:
@zuphilip, yeah, maybe you're right. Anyway, the point of the pull request was to treat "http://" and "https://" as the same when it comes to comparing style IDs.

> What is currently the idea behind having <id> and as well <link rel="self">

Per the spec, only the "independent-parent" link for dependent styles is required. The "self" link is optional, but all styles need an ID. The idea was that not all styles are hosted online. If you create a local style, it doesn't need a "self" link.

rmzelle:
@dstillman, shall we close this? It makes more sense to create a CSL issue, since as you mentioned, the "independent-parent" link currently serves a dual role of link (with can be either HTTP or HTTPS) and identifier (of the parent). It might be good to see if we need to change this behavior, or just clearly document the requirements in the CSL spec, but that's all within the scope of CSL.

For now we don't allow "https" on "independent-parent" and "template" links, so Zotero should be save with CSL repo styles.

PR #879, opened 2015-12-03
Restores the functionality of 4.0 for sync settings 

adomasven:
And adds an indicator for valid sync credentials

### DIFF
--- a/chrome/content/zotero/preferences/preferences_sync.js
+++ b/chrome/content/zotero/preferences/preferences_sync.js
@@ -24,19 +24,232 @@
 */
 
 "use strict";
+Components.utils.import("resource://gre/modules/Services.jsm");
 
 Zotero_Preferences.Sync = {
-	init: function () {
+	init: Zotero.Promise.coroutine(function* () {
 		this.updateStorageSettings(null, null, true);
-		
-		document.getElementById('sync-api-key').value = Zotero.Sync.Data.Local.getAPIKey();
-		
+
+		var username = Zotero.Users.getCurrentUsername() || "";
+		var apiKey = Zotero.Sync.Data.Local.getAPIKey();
+		this.displayFields(apiKey ? username : "");
+		if (apiKey) {
+			try {
+				var keyInfo = yield Zotero.Sync.Runner.checkAccess(
+					Zotero.Sync.Runner.getAPIClient({apiKey}),
+					{timeout: 5000}
+				);
+				this.displayFields(keyInfo.username);
+			}
+			catch (e) {
+				// API key wrong/invalid
+				if (!(e instanceof Zotero.HTTP.UnexpectedStatusException) &&
+					!(e instanceof Zotero.HTTP.TimeoutException)) {
+					Zotero.alert(
+						window,
+						Zotero.getString('general.error'),
+						Zotero.getString('sync.error.apiKeyInvalid', Zotero.clientName)
+					);
+					this.unlinkAccount(false);
+				}
+				else {
+					throw e;
+				}
+			}
+		}
+
+
 		// TEMP: Disabled
 		//var pass = Zotero.Sync.Storage.WebDAV.password;
 		//if (pass) {
 		//	document.getElementById('storage-password').value = pass;
 		//}
+	}),
+
+	displayFields: function (username) {
+		document.getElementById('sync-unauthorized').hidden = !!username;
+		document.getElementById('sync-authorized').hidden = !username;
+		document.getElementById('sync-reset-tab').disabled = !username;
+		document.getElementById('sync-username').value = username;
+		document.getElementById('sync-password').value = '';
+		document.getElementById('sync-username-textbox').value = Zotero.Prefs.get('sync.server.username');
+
+		var img = document.getElementById('sync-status-indicator');
+		img.removeAttribute('verified');
+		img.removeAttribute('animated');
 	},
+
+
+	credentialsKeyPress: function (event) {
+		var username = document.getElementById('sync-username-textbox');
+		username.value = username.value.trim();
+		var password = document.getElementById('sync-password');
+
+		var syncAuthButton = document.getElementById('sync-auth-button');
+
+		syncAuthButton.setAttribute('disabled', 'true');
+
+		// When using backspace, the value is not updated until after the keypress event
+		setTimeout(function() {
+			if (username.value.length && password.value.length) {
+				syncAuthButton.setAttribute('disabled', 'false');
+			}
+		});
+
+		if (event.keyCode == 13) {
+			Zotero_Preferences.Sync.linkAccount(event);
+		}
+	},
+
+
+	linkAccount: Zotero.Promise.coroutine(function* (event) {
+		var username = document.getElementById('sync-username-textbox').value;
+		var password = document.getElementById('sync-password').value;
+
+		if (!username.length || !password.length) {
+			this.updateSyncIndicator();
+			return;
+		}
+
+		// Try to acquire API key with current credentials
+		this.updateSyncIndicator('animated');
+		var json = yield Zotero.Sync.Runner.createAPIKeyFromCredentials(username, password);
+		this.updateSyncIndicator();
+
+		// Invalid credentials
+		if (!json) {
+			Zotero.alert(window,
+				Zotero.getString('general.error'),
+				Zotero.getString('sync.error.invalidLogin')
+			);
+			return;
+		}
+
+		if (!(yield this.checkUser(json.userID, json.username))) {
+			// createAPIKeyFromCredentials will have created an API key,
+			// but user decided not to use it, so we remove it here.
+			Zotero.Sync.Runner.deleteAPIKey();
+			return;
+		}
+
+		this.displayFields(json.username);
+	}),
+
+	/**
+	 * Updates the auth indicator icon, depending on status
+	 * @param {string} status
+	 */
+	updateSyncIndicator: function (status) {
+		var img = document.getElementById('sync-status-indicator');
+		
+		img.removeAttribute('animated');
+		if (status == 'animated') {
+			img.setAttribute('animated', true);
+		}
+	},
+
+	unlinkAccount: Zotero.Promise.coroutine(function* (showAlert=true) {
+		if (showAlert) {
+			if (!Services.prompt.confirm(
+				null,
+				Zotero.getString('general.warning'),
+				Zotero.getString('sync.unlinkWarning', Zotero.clientName)
+			)) {
+				return;
+			}
+		}
+
+		this.displayFields();
+		yield Zotero.Sync.Runner.deleteAPIKey();
+	}),
+
+
+	/**
+	 * Make sure we're syncing with the same account we used last time, and prompt if not.
+	 * If user accepts, change the current user, delete existing groups, and update relation
+	 * URIs to point to the new user's library.
+	 *
+	 * @param	{Integer}	userID			New userID
+	 * @param	{Integer}	libraryID		New libraryID
+	 * @return {Boolean} - True to continue, false to cancel
+	 */
+	checkUser: Zotero.Promise.coroutine(function* (userID, username) {
+		var lastUserID = Zotero.Users.getCurrentUserID();
+		var lastUsername = Zotero.Users.getCurrentUsername();
+
+		if (lastUserID && lastUserID != userID) {
+			var groups = Zotero.Groups.getAll();
+
+			var ps = Components.classes["@mozilla.org/embedcomp/prompt-service;1"]
+				.getService(Components.interfaces.nsIPromptService);
+			var buttonFlags = (ps.BUTTON_POS_0) * (ps.BUTTON_TITLE_IS_STRING)
+				+ (ps.BUTTON_POS_1) * (ps.BUTTON_TITLE_CANCEL)
+				+ (ps.BUTTON_POS_2) * (ps.BUTTON_TITLE_IS_STRING)
+				+ ps.BUTTON_POS_1_DEFAULT
+				+ ps.BUTTON_DELAY_ENABLE;
+
+			var msg = Zotero.getString('sync.lastSyncWithDifferentAccount', [lastUsername, username]);
+			var syncButtonText = Zotero.getString('sync.sync');
+
+			msg += " " + Zotero.getString('sync.localDataWillBeCombined', username);
+			// If there are local groups belonging to the previous user,
+			// we need to remove them
+			if (groups.length) {
+				msg += " " + Zotero.getString('sync.localGroupsWillBeRemoved1');
+				var syncButtonText = Zotero.getString('sync.removeGroupsAndSync');
+			}
+			msg += "\n\n" + Zotero.getString('sync.avoidCombiningData', lastUsername);
+
+			var index = ps.confirmEx(
+				null,
+				Zotero.getString('general.warning'),
+				msg,
+				buttonFlags,
+				syncButtonText,
+				null,
+				Zotero.getString('sync.openSyncPreferences'),
+				null, {}
+			);
+
+			if (index > 0) {
+				if (index == 2) {
+					var wm = Components.classes["@mozilla.org/appshell/window-mediator;1"]
+							   .getService(Components.interfaces.nsIWindowMediator);
+					var lastWin = wm.getMostRecentWindow("navigator:browser");
+					lastWin.ZoteroPane.openPreferences('zotero-prefpane-sync');
+				}
+				return false;
+			}
+		}
+
+		yield Zotero.DB.executeTransaction(function* () {
+			if (lastUserID != userID) {
+				if (lastUserID) {
+					// Delete all local groups if changing users
+					for (let group of groups) {
+						yield group.erase();
+					}
+
+					// Update relations pointing to the old library to point to this one
+					yield Zotero.Relations.updateUser(userID);
+				}
+				// Replace local user key with libraryID, in case duplicates were
+				// merged before the first sync
+				else {
+					yield Zotero.Relations.updateUser(userID);
+				}
+
+				yield Zotero.Users.setCurrentUserID(userID);
+			}
+
+			if (lastUsername != username) {
+				yield Zotero.Users.setCurrentUsername(username);
+			}
+		})
+
+		return true;
+	}),
+
 	
 	updateStorageSettings: function (enabled, protocol, skipWarnings) {
 		if (enabled === null) {
--- a/chrome/content/zotero/preferences/preferences_sync.xul
+++ b/chrome/content/zotero/preferences/preferences_sync.xul
@@ -47,216 +47,240 @@
 		<tabbox>
 			<tabs>
 				<tab label="&zotero.preferences.settings;"/>
-				<tab label="&zotero.preferences.sync.reset;"/>
+				<tab id="sync-reset-tab" label="&zotero.preferences.sync.reset;" disabled="true"/>
 			</tabs>
 			
 			<tabpanels>
 				<tabpanel orient="vertical">
-					<groupbox>
-						<caption label="&zotero.preferences.sync.syncServer;"/>
-						
-						<hbox>
-							<grid>
+					<vbox id="sync-unauthorized">
+						<groupbox>
+							<caption label="&zotero.preferences.sync.syncServer;"/>
+
+							<hbox>
+								<grid>
+									<columns>
+										<column/>
+										<column/>
+									</columns>
+
+									<rows>
+										<row>
+											<label value="&zotero.preferences.sync.username;"/>
+											<textbox id="sync-username-textbox"
+												preference="pref-sync-username"
+												onkeypress="Zotero_Preferences.Sync.credentialsKeyPress(event);"/>
+										</row>
+										<row>
+											<label value="&zotero.preferences.sync.password;"/>
+											<textbox id="sync-password" type="password"
+												onkeypress="Zotero_Preferences.Sync.credentialsKeyPress(event);"/>
+										</row>
+										<vbox align="center">
+											<hbox align="baseline">
+												<button id="sync-auth-button"
+													label="&zotero.preferences.sync.setUpSync;"
+													oncommand="Zotero_Preferences.Sync.linkAccount(event)"
+													disabled="true"/>
+												<label id="sync-status-indicator"/>
+											</hbox>
+										</vbox>
+									</rows>
+								</grid>
+								<vbox style="width:2em"/>
+								<vbox>
+									<label class="zotero-text-link" value="&zotero.preferences.sync.createAccount;" href="http://zotero.org/user/register"/>
+									<separator class="thin"/>
+									<label class="zotero-text-link" value="&zotero.preferences.sync.lostPassword;" href="http://zotero.org/user/lostpassword"/>
+									<separator class="thin"/>
+									<label class="zotero-text-link" value="&zotero.preferences.sync.about;" href="http://www.zotero.org/support/sync"/>
+								</vbox>
+							</hbox>
+						</groupbox>
+					</vbox>
+
+					<vbox id="sync-authorized" hidden="true">
+						<groupbox>
+							<caption label="&zotero.preferences.sync.syncServer;"/>
+
+							<hbox>
+								<grid>
+									<columns>
+										<column/>
+										<column/>
+									</columns>
+
+									<rows>
+										<row>
+											<label value="&zotero.preferences.sync.username;"/>
+											<label id="sync-username" value="Username"/>
+										</row>
+										<row>
+											<box/>
+											<button label="&zotero.preferences.sync.unlinkAccount;"
+													oncommand="Zotero_Preferences.Sync.unlinkAccount()"/>
+										</row>
+										<!--
+										<row>
+											<box/>
+											<button label="Access Control" oncommand="Zotero.alert('Not implemented');"/>
+										</row>
+										-->
+										<row>
+											<box/>
+											<checkbox label="&zotero.preferences.sync.syncAutomatically;"
+													  disabled="true"/>
+										</row>
+										<row>
+											<box/>
+											<checkbox label="&zotero.preferences.sync.syncFullTextContent;"
+													  tooltiptext="&zotero.preferences.sync.syncFullTextContent.desc;"
+													  disabled="true"/>
+										</row>
+									</rows>
+								</grid>
+								<vbox>
+									<label class="zotero-text-link" value="&zotero.preferences.sync.about;" href="http://www.zotero.org/support/sync"/>
+								</vbox>
+							</hbox>
+
+						</groupbox>
+
+						<groupbox id="storage-settings">
+							<caption label="&zotero.preferences.sync.fileSyncing;"/>
+
+							<!-- My Library -->
+							<hbox>
+								<checkbox label="&zotero.preferences.sync.fileSyncing.myLibrary;"
+										preference="pref-storage-enabled"
+										oncommand="Zotero_Preferences.Sync.updateStorageSettings(this.checked, null)"/>
+								<menulist id="storage-protocol" class="storage-personal"
+										style="margin-left: .5em"
+										preference="pref-storage-protocol"
+										oncommand="Zotero_Preferences.Sync.updateStorageSettings(null, this.value)">
+									<menupopup>
+										<menuitem label="Zotero" value="zotero"/>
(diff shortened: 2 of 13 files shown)


dstillman:
OK, I'll have specific code notes, but biggest thing here is that we don't want to keep the username/password fields showing after authentication. I have an API key entered already, and the sync pane shows a blank username/password with a checkbox with the active username shown only on hover. It should instead just show me the linked account, similar to Dropbox, with the ability to unlink (and maybe to change access).

Two main options here would be to have credentials fields only show when no API key or to just have a "Set Up Sync…" button that asked for username/password in a prompt or wizard (with an Advanced button/toggle for configuring access).

For former, that would mean instead of onchange we'd want a log in button, but I'm not sure what wording to use — "Authenticate…" works but is a little technical, "Sign In…"/"Log In…" is weird for a client, "Set Up Sync…"/"Configure Sync…" could work but isn't as clearly associated with the login fields.

Whether or not the username/password are in a separate prompt, we should probably hide everything else in the dialog when unlinked, which would make a "Set Up Sync…" button less weird (because isn't setting up sync what this pane is for?). It'd be good to do that anyway to communicate that syncing doesn't work at all if you haven't entered data sync credentials.

Should probably also do a key info request (there's an APIClient method for that) using the API key when opening the pane. If it fails with a 403, it's never going to work again and the client can be considered unlinked, though the username field could be prepopulated.

chrome/content/zotero/xpcom/sync/syncAPIClient.js
```diff
@@ -430,6 +434,46 @@ Zotero.Sync.APIClient.prototype = {
 	}),
 	
 	
+	getAPIKeyFromCredentials: Zotero.Promise.coroutine(function* () {
+		var apiKey;
+		let body = JSON.stringify({
+			username: this.username,
+			password: this.password,
+			name: Zotero.isStandalone ? "Zotero Standalone" : "Zotero for Firefox",
```
dstillman: "Automatic Zotero Client Key" — user agent will be determined automatically, and we might share prefs between Firefox and Standalone at some point

chrome/content/zotero/xpcom/sync/syncAPIClient.js
```diff
@@ -430,6 +434,46 @@ Zotero.Sync.APIClient.prototype = {
...
+						library: true,
+						write: true
+					}
+				}
+			}
+		});
+		let headers = {
+			"Content-Type": "application/json"
+		};
+		var uri = this.baseURL + "keys/";
```
dstillman: No trailing "/"

chrome/content/zotero/xpcom/sync/syncAPIClient.js
```diff
@@ -430,6 +434,46 @@ Zotero.Sync.APIClient.prototype = {
...
+		var response = yield this.makeRequest("POST", uri, {body, headers, successCodes: [201, 403]});
+		if (response.status == 403) {
+			return "";
```
dstillman: Return `false` instead

chrome/content/zotero/xpcom/sync/syncAPIClient.js
```diff
@@ -430,6 +434,46 @@ Zotero.Sync.APIClient.prototype = {
...
+		}
+		
+		var json = this._parseJSON(response.responseText);
+		apiKey = json.key;
+		if (! apiKey) {
```
dstillman: Style: No space after `!`

chrome/content/zotero/xpcom/sync/syncAPIClient.js
```diff
@@ -430,6 +434,46 @@ Zotero.Sync.APIClient.prototype = {
 	}),
 	
 	
+	getAPIKeyFromCredentials: Zotero.Promise.coroutine(function* () {
+		var apiKey;
+		let body = JSON.stringify({
```
dstillman: Should stick to `var` for function scope (you have a mix throughout)

chrome/content/zotero/xpcom/sync/syncAPIClient.js
```diff
@@ -30,12 +30,16 @@ if (!Zotero.Sync) {
...
+	if (!(options.apiKey || (options.username && options.password))) {
+		throw new Error("neither apiKey nor (username, password) is set");
+	}
 	if (!options.caller) throw new Error("caller not set");
 	
 	this.baseURL = options.baseURL;
 	this.apiVersion = options.apiVersion;
 	this.apiKey = options.apiKey;
+	this.username = options.username;
+	this.password = options.password;
```
dstillman: Can't these just be parameters to `getAPIKeyFromCredentials`? They should only be used for one request, so no need to store them.

chrome/content/zotero/xpcom/sync/syncRunner.js
```diff
@@ -1244,20 +1255,23 @@ Zotero.Sync.Runner_Module = function (options = {}) {
...
+					return "";
+				}
 			}
-			throw new Error("Unimplemented");
-			// Get API key from server
-			
-			// Store API key
 			
-			// Remove old logins and username pref
+			yield apiKey = Zotero.Sync.Runner.createAPIKeyFromCredentials(username, password);
```
dstillman: `yield` in wrong place

chrome/content/zotero/xpcom/sync/syncAPIClient.js
adomasven: This is a matter of creating the `APIClient` without passing in `apiKey` (which throws an exception), so I figured I may as well store the credentials. In general was not certain what's the best way to instantiate `APIClient` without the API key. Maybe just pass in an empty string instead in `Sync.Runner` or set up a custom flag like `noApiKey`?

dstillman: Ah, right. We could drop the API key requirement in the constructor and just throw if there's no API key in `makeRequest` unless `options.noAPIKey` is passed, which would only be the case from `APIClient.getAPIKeyFromCredentials()`.

test/tests/preferences_syncTests.js
```diff
@@ -0,0 +1,73 @@
...
+				usernameElem.value = username;
+				passwordElem.value = password;
+				
+				// Triggered by `change` event for usernameElem and passwordElem;
+				yield win.Zotero_Preferences.Sync.setCredentials();
+			});
+
+			var apiKey = Zotero.Utilities.randomString(24);
+
+			before(function *() {
```
dstillman: Style: `function* () {`

test/tests/preferences_syncTests.js
```diff
@@ -0,0 +1,73 @@
+describe("Sync Preferences", function () {
+	describe("Settings", function () {
```
dstillman: Any reason for this extra wrapper?

dstillman:
(The other implication is that there should be an "Unlink This Zotero…" button (as in Dropbox) when authenticated that triggers a `DELETE` request for the key.)

test/tests/preferences_syncTests.js
adomasven: It's inline with how other preference tests are scoped for different subsections. I.e. this is the *Settings* tab, not *Reset*.

dstillman: Ah, got it.

adomasven:
Okay. I've just pushed a commit which makes the workflow a bit more sensible.

When unlinked, you're presented with an auth form. When linked, it displays the old settings with some additional buttons like *Access Control*. Unlinking also sends the `DELETE` request to the dataserver.

dstillman:
OK, some notes to start:

1) This seems to just use presence of stored username for the test. I entered a username but no password and closed the window, and when I came back it thought I was authenticated. And then "Unlink Account" didn't even work (without closing and reopening the window), because it failed with "API key not set" (presumably on the DELETE request).

2) Keep the UI elements left-justified, with the web links to the right. Centered doesn't work with the web links and checkboxes.

3) Pressing Enter from within the text fields should trigger the set up.

4) Should show the error (e.g., "Invalid username or password") in the dialog, not just on hover over the error icon. Modal alert is fine, if adding a status line looks too weird.

5) "Set-Up Sync" -> "Set Up Syncing" (I don't love this, but still trying to figure out what I do like)

6) Disable "Set Up Syncing" until there are values in both text fields.

7) I think unlinking the old key can happen asynchronously, with the UI resetting immediately. Worst-case scenario there's an extra key lying around. Better than delaying or failing.

8) "Unlink Account" -> "Unlink Account…", with a confirmation dialog similar to what Dropbox shows (e.g., clarifying that data isn't deleted). And this can happen in a separate PR, but we actually should probably give the option of either leaving data in place or resetting the local data directory (which would mean deleting zotero.sqlite and 'storage').

adomasven:
2) It looks better centered in my opinion, but fair enough.

3) I was looking into that, but couldn't find a straightforward way to do it, so I just dropped it. `keypress` event is probably the one to go for if doing it in the convoluted way? `oninput` is not triggered with non-character keys.

7) Does `oncommand` properly wait for promises or are coroutines blocking? I was expecting it to not block anyway.

dstillman:
3) Check the WebDAV code in the same pane, which triggers Verify Server on Enter.

7) It's not blocking, but you don't update the UI until after the DELETE finishes.

adomasven:
Adjustments pushed.

dstillman:
OK, getting there.

1) The text fields shouldn't blur on Enter — can just trigger the click (which doesn't blur the fields either when you do it manually).

2) If the button is disabled, pressing Enter shouldn't have any effect. (This will probably be addressed by removing the blur.)

3) When deleting a field, the button doesn't disable until after another keypress.

4) Can remove "Access Control" for now.

(There are a few other UI issues, but I'm going to rework this a little in a follow-up commit.)

chrome/content/zotero/preferences/preferences_sync.js
```diff
@@ -24,19 +24,125 @@
...
+			try {
+				var keyInfo = yield Zotero.Sync.Runner.checkAccess(Zotero.Sync.Runner.getAPIClient({apiKey}));
+				Zotero.Prefs.set('sync.server.username', keyInfo.value);
+				this.displayFields(keyInfo.username);
+			} catch (e) {
+				// API key wrong/invalid
+				Zotero.alert(
+						window,
+						Zotero.getString('general.error'),
+						Zotero.getString('sync.error.apiKeyInvalid')
```
dstillman: Nit: Extra indent

chrome/content/zotero/preferences/preferences_sync.js
```diff
@@ -24,19 +24,125 @@
...
+		var username = Zotero.Prefs.get('sync.server.username');
+		var apiKey = Zotero.Sync.Data.Local.getAPIKey();
+		if (apiKey) {
+			this.displayFields(username);
+			try {
+				var keyInfo = yield Zotero.Sync.Runner.checkAccess(Zotero.Sync.Runner.getAPIClient({apiKey}));
+				Zotero.Prefs.set('sync.server.username', keyInfo.value);
+				this.displayFields(keyInfo.username);
+			} catch (e) {
+				// API key wrong/invalid
```
dstillman: This needs to check `if (e instanceof Zotero.HTTP.UnexpectedStatusException)` and not unlink if so. Otherwise a temporary server error would unlink anyone who opened the pane. (That'd be a 5xx, but any 4xx that made it through (i.e., not 404, which is caught) should probably similarly be ignored.)

chrome/content/zotero/preferences/preferences_sync.js
```diff
@@ -24,19 +24,125 @@
...
+		}
+
+		if (event.keyCode == 13) {
+			event.target.blur();
+			Zotero_Preferences.Sync.setUpSync(event);
+		}
+	},
+
+
+	setUpSync: Zotero.Promise.coroutine(function* (event) {
```
dstillman: Still not sure what the button will be, but let's call this `linkAccount`.

chrome/content/zotero/preferences/preferences_sync.js
```diff
@@ -24,19 +24,125 @@
...
+		var apiKey = yield Zotero.Sync.Runner.getAPIKeyFromCredentials(username, password);
+		if (apiKey) {
+			Zotero.Prefs.set('sync.server.username', username);
+			this.displayFields(username);
+		}
+		else {
+			this.updateSyncIndicator(0);
+			Zotero.alert(window,
+					Zotero.getString('general.error'),
+					Zotero.getString('sync.error.invalidLogin')
```
dstillman: Extra indent

chrome/content/zotero/preferences/preferences_sync.js
```diff
@@ -24,19 +24,125 @@
...
+			Zotero_Preferences.Sync.setUpSync(event);
+		}
+	},
+
+
+	setUpSync: Zotero.Promise.coroutine(function* (event) {
+		var username = document.getElementById('sync-username-textbox').value;
+		var password = document.getElementById('sync-password').value;
+
+		if (username.length <= 0 || password.length <= 0) {
```
dstillman: `!username.length || !password.length`

chrome/content/zotero/preferences/preferences_sync.js
```diff
@@ -24,19 +24,125 @@
...
+			);
+		}
+	}),
+
+	/**
+	 * Updates the auth indicator icon, depending on status
+	 * @param {int} status 
+	 */
+	updateSyncIndicator: function (status) {
+		const animated=1;
```
dstillman: Can just take a string (e.g., 'animated') as a parameter, to make calling code clearer.

chrome/content/zotero/preferences/preferences_sync.xul
```diff
@@ -47,216 +47,237 @@
...
 			</tabs>
 			
 			<tabpanels>
 				<tabpanel orient="vertical">
-					<groupbox>
-						<caption label="&zotero.preferences.sync.syncServer;"/>
-						
-						<hbox>
-							<grid>
+					<vbox id="sync-auth-form">
```
dstillman: Let's call this `sync-unauthorized`, for consistency/clarity.

chrome/content/zotero/preferences/preferences_sync.xul
```diff
@@ -47,216 +47,237 @@
...
+											<textbox id="sync-password" type="password"
+												onkeypress="Zotero_Preferences.Sync.credentialsKeyPress(event);"/>
+										</row>
+										<vbox align="center">
+											<hbox align="baseline">
+												<button id="sync-auth-button"
+														label="&zotero.preferences.sync.setUpSync;"
+														oncommand="Zotero_Preferences.Sync.setUpSync(event)"
+														disabled="true"/>
+												<label id="sync-status-indicator" tooltiptext=""/>
```
dstillman: No `tooltiptext` attribute if it's empty

chrome/content/zotero/xpcom/sync/syncAPIClient.js
```diff
@@ -430,6 +429,54 @@ Zotero.Sync.APIClient.prototype = {
 	}),
 	
 	
+	getAPIKeyFromCredentials: Zotero.Promise.coroutine(function* (username, password) {
+		var apiKey;
+		var body = JSON.stringify({
+			username: username,
+			password: password,
```
dstillman: Omit redundant values

chrome/content/zotero/xpcom/sync/syncAPIClient.js
```diff
@@ -430,6 +429,54 @@ Zotero.Sync.APIClient.prototype = {
...
+		});
+		if (response.status == 403) {
+			return false;
+		}
+		
+		var json = this._parseJSON(response.responseText);
+		apiKey = json.key;
+		if (!apiKey) {
+			return false;
+		}
```
dstillman: If there's no `key` property, it should just throw, since that should never happen.

chrome/content/zotero/xpcom/sync/syncLocal.js
```diff
@@ -154,6 +156,30 @@ Zotero.Sync.Data.Local = {
...
 	
+	removeLegacyLogins: function () {
+		var loginManagerHost = 'chrome://zotero';
+		var loginManagerRealm = 'Zotero Sync Server';
+		
+		Zotero.debug('Removing legacy Zotero sync credentials (api key acquired)');
+		
+		var loginManager = Components.classes["@mozilla.org/login-manager;1"]
+			.getService(Components.interfaces.nsILoginManager);
+		loginManager.removeAllLogins();
```
dstillman: Yeah, don't want this — this would wipe out all the user's saved passwords in Firefox.

chrome/content/zotero/xpcom/sync/syncRunner.js
```diff
@@ -1191,6 +1188,24 @@ Zotero.Sync.Runner_Module = function (options = {}) {
...
+
+
+	this.getAPIKeyFromCredentials = Zotero.Promise.coroutine(function* (username, password) {
+		var client = this.getAPIClient();
+		var apiKey = yield client.getAPIKeyFromCredentials(username, password);
+		if (apiKey) {
+			Zotero.Sync.Data.Local.setAPIKey(apiKey);
+		}
+		return apiKey;
+	})
```
dstillman: Swap order of these two methods

chrome/content/zotero/xpcom/sync/syncRunner.js
```diff
@@ -1244,20 +1259,23 @@ Zotero.Sync.Runner_Module = function (options = {}) {
 	
 	
 	var _getAPIKeyFromLogin = Zotero.Promise.coroutine(function* () {
-		var apiKey = "";
+		var apiKey;
 		let username = Zotero.Prefs.get('sync.server.username');
 		if (username) {
-			let password = Zotero.Sync.Data.Local.getLegacyPassword(username);
+			let password = Zotero.Sync.Data.Local.password;
```
dstillman: This is left over from an earlier revision, right? There's no such property anymore?

chrome/locale/en-US/zotero/preferences.dtd
```diff
@@ -49,7 +49,10 @@
 <!ENTITY zotero.preferences.prefpane.sync							"Sync">
 <!ENTITY zotero.preferences.sync.username							"Username:">
 <!ENTITY zotero.preferences.sync.password							"Password:">
-<!ENTITY zotero.preferences.sync.syncServer						"Zotero Sync Server">
+<!ENTITY zotero.preferences.sync.syncServer						"Zotero Data Sync">
+<!ENTITY zotero.preferences.sync.accessControl						"Access Control">
```
dstillman: Can remove this for now along with the button.

chrome/content/zotero/preferences/preferences_sync.xul
```diff
@@ -47,216 +47,237 @@
...
+								<grid>
+									<columns>
+										<column/>
+										<column/>
+									</columns>
+
+									<rows>
+										<row>
+											<label value="&zotero.preferences.sync.username;"/>
+											<label id="sync-username" value="Username"/>
```
dstillman: No `value` attribute needed when unset

test/tests/syncRunnerTest.js
```diff
@@ -718,4 +749,20 @@ describe("Zotero.Sync.Runner", function () {
 			assert.isBelow(lastSyncTime, new Date().getTime());
 		})
 	})
+	
+	describe("#getAPIKeyFromCredentials()", function() {
+		it("should return API key when credentials valid", function *() {
```
dstillman: Nit: `function* (` (same below)

test/tests/syncRunnerTest.js
```diff
@@ -718,4 +749,20 @@ describe("Zotero.Sync.Runner", function () {
...
+			setResponse('createKey.validCredentials');
+			
+			var key = yield runner.getAPIKeyFromCredentials('Username', 'correctPassword');
+			assert.equal(key, apiKey);
+		});
+		
+		it("should return an empty string when credentials invalid", function *() {
```
dstillman: Update description

test/tests/preferences_syncTests.js
```diff
@@ -0,0 +1,84 @@
+describe("Sync Preferences", function () {
```
dstillman: Filename should be `preferences_syncTest.js` (singular, not plural)

test/tests/syncRunnerTest.js
```diff
@@ -718,4 +749,20 @@ describe("Zotero.Sync.Runner", function () {
 			assert.isBelow(lastSyncTime, new Date().getTime());
 		})
 	})
+	
+	describe("#getAPIKeyFromCredentials()", function() {
+		it("should return API key when credentials valid", function *() {
+			setResponse('createKey.validCredentials');
```
dstillman: Stubs are fine in the pref tests, but for `syncRunner` we should test the actual data being uploaded — `setResponse` is generally more appropriate for testing GET requests. Look for `server.respond()` in syncEngineTest.js for examples of asserting on the POSTed data.

Should add a test for `#deleteAPIKey()` as well.

adomasven:
Amended according to comments.

dstillman:
Oh, so we should actually show `authorized` first before waiting for the API request — right now it flashes the login form, and if the API is being slow that could be for a while. We could show the spinner next to the authorized username and change that to the checkmark once verified. Only issue is that the spinner might alarm people (who didn't check the tooltip) since it would look sort of like it's syncing — we could show a status line too, but hopefully in most cases the API would be fast enough that the spinner would disappear almost immediately.

Also, if the key is invalid, we should keep the old username, since there's a good chance it's still the same user, and people have trouble with usernames.

We should also [set a timeout](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Synchronous_and_Asynchronous_Requests#Example_using_a_timeout) on the auth check XHR. (Otherwise, if we didn't disable Unlink Account, a slow auth check could unlink a subsequent new reauthorization, or if we did disable Unlink Account it could stay disabled for a long time.) I don't think we use an XHR timeout anywhere right now, so that probably needs to be added as an `options` feature to `Zotero.HTTP` and passed as an option to `makeRequest` from `APIClient::getAPIKeyFromCredentials()`. (Could pass through from the pref code, but until anything else uses this, it's probably fine to hard-code in `APIClient::getAPIKeyFromCredentials()`.)

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -820,12 +820,14 @@ sync.error.sslConnectionError 		= SSL connection error
...
+sync.error.apiKeyInvalid			= We could not authenticate your account with Zotero servers. Please re-enter your account details.
 
 sync.lastSyncWithDifferentAccount	= This Zotero database was last synced with a different zotero.org account ('%1$S') from the current one ('%2$S').
 sync.localDataWillBeCombined		= If you continue, local Zotero data will be combined with data from the '%S' account stored on the server.
 sync.localGroupsWillBeRemoved1		= Local groups, including any with changed items, will also be removed from this computer.
 sync.avoidCombiningData				= To avoid combining or losing data, revert to the '%S' account or use the Reset options in the Sync pane of the Zotero preferences.
 sync.localGroupsWillBeRemoved2		= If you continue, local groups, including any with changed items, will be removed and replaced with groups linked to the '%1$S' account.\n\nTo avoid losing local changes to groups, be sure you have synced with the '%2$S' account before syncing with the '%1$S' account.
+sync.unlinkWarning					= Are you sure you want to unlink this account from Zotero?\n\nThis will stop syncing your database but no data will be lost.
```
dstillman: "Are you sure you want to unlink this account?\n\n%S will no longer sync your data, but your data will remain locally."

And pass in `Zotero.appName`. (We haven't generally done this historically, but we want to start not hard-coding "Zotero" anywhere, for the sake of third-party builds.)

(When we make this fancier and give options, let's remember to make groups read-only after unlinking if people choose to keep them.)

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -820,12 +820,14 @@ sync.error.sslConnectionError 		= SSL connection error
 sync.error.checkConnection 			= Error connecting to server. Check your Internet connection.
 sync.error.emptyResponseServer 		= Empty response from server.
 sync.error.invalidCharsFilename		= The filename '%S' contains invalid characters.\n\nRename the file and try again. If you rename the file via the OS, you will need to relink it in Zotero.
+sync.error.apiKeyInvalid			= We could not authenticate your account with Zotero servers. Please re-enter your account details.
```
dstillman: "%S could not authenticate your account. Please re-enter your account details."

With `Zotero.appName`

dstillman:
OK, other than those changes, this looks great.

adomasven:
>Oh, so we should actually show authorized first before waiting for the API request — right now it flashes the login form, and if the API is being slow that could be for a while. 

Okay, the flashing was due to a bug with how I was storing the username in the preferences. Since it first checks if the API key exists locally and then displays the layout (using the username from the prefs), which happens basically instantly.

>I don't think we use an XHR timeout anywhere right now, so that probably needs to be added as an options feature to Zotero.HTTP and passed as an option to makeRequest from APIClient::getAPIKeyFromCredentials(). (Could pass through from the pref code, but until anything else uses this, it's probably fine to hard-code in APIClient::getAPIKeyFromCredentials().)

The timeout was actually needed on `Runner::checkAccess()`, which made passing the option through preferences code necessary.

dstillman:
Oh, one more thing: instead of using `sync.server.username`, this should use `Zotero.Users.getCurrentUsername()` and `Zotero.Users.setCurrentUsername()`, with the latter called in `SyncRunner.getAPIKeyFromCredentials()`. Otherwise clearing the username pref shows the login form in the prefs, even though there's still an API key registered. `getLegacyPassword()` should still check the pref, but it can run `Zotero.Prefs.clear()` on it in `removeLegacyLogins()`.

adomasven:
Changed and pushed.

adomasven:
I have just realised that setting `Zotero.Users.setCurrentUsername()` will break the alert, which notifies about syncing with a different user than previously. `Zotero.Runner.checkUser()` calls `Zotero.Users.getCurrentUsername()` to determine the previous username and display it in a warning prompt. 

We need to store the username pre-sync for display purposes, but we need to retain the old username for the alert to keep current functionality. Or we could show the alert when first setting up the credentials, but a user might type their email in and might get a false notification (since the saved username will be their Zotero username, not email).

dstillman:
Oh, yeah, I was actually thinking we should move that warning to account-linking time.

dstillman:
(And the actual comparison is based on userID, not username.)

adomasven:
It is, but it displays a warning with [old and new usernames](https://github.com/zotero/zotero/blob/master/chrome/content/zotero/xpcom/sync/syncRunner.js#L455)

dstillman:
Yeah, but that won't make it a false notification. They won't get the warning unless it's a different account.

adomasven:
Ah yes, I was thinking it would be a problem with the current alert when done on first sync. I'll move it to linking time.

adomasven:
Pushed with linking alert for differing `userID`. This seems to have broken one of the `Zotero.Relations` tests.

dstillman:
I'm getting "undefined" in the Username field now if I open the prefs.

I have "extensions.zotero.sync.server.username" set but no user stored in the DB.

adomasven:
Amended.

dstillman:
Sorry, where I said `Zotero.appName` I should have said `Zotero.clientName`. (`appName` shows "Firefox".)

dstillman:
All right, other than that, this looks great — can squash this down.

adomasven:
Yeah, I saw it displaying "Firefox", but assumed maybe we will change that variable, because I couldn't find something like `Zotero.clientName`. Shall I change it and squash push?

adomasven:
Squashed and pushed.

dstillman:
Actually, could you pull out sinon-as-promised as a separate commit (before the other one)?

adomasven:
Done

test/tests/preferences_syncTest.js
```diff
@@ -0,0 +1,158 @@
...
+				yield setCredentials("Username", "correctPassword");
+				
+				assert.equal(Zotero.Sync.Data.Local.getAPIKey(), apiKey);
+				assert.equal(doc.getElementById('sync-unauthorized').getAttribute('hidden'), 'true');
+			});
+
+
+			it("should display dialog when credentials incorrect", function* () {
+				getAPIKeyFromCredentialsStub.resolves(false);
+				var dialog = waitForDialog();
```
dstillman: This is failing in Travis due to the `noUserInput` flag affecting `Zotero.alert()`. Should use a spy instead.

PR #888, opened 2015-12-23
Fix/sync auth form

### LINKED ISSUE #833: Warn before syncing into an empty library
We should make sure we warn before syncing into a new library. It's impressive when it just works, but it causes too much confusion when people, say, move a data directory without disabling auto-sync.

(I think right now we only warn when the userID changes.)

### LINKED ISSUE #886: Auth form shown in sync prefs until key info check finishes
Looks like this is still happening. Instead, if there's a DB username and API key it should show the status as linked until it knows for sure otherwise.

Follow-up from #879

adomasven:
Closes #886 
Closes #833 

Did not really require `Zotero.Library.hasItems` method, but I had implemented it before (my idea was to check for absence of items in current library and presence of items on the sync server), so we may as well keep it - might come in handy.

### DIFF
--- a/chrome/content/zotero/xpcom/data/library.js
+++ b/chrome/content/zotero/xpcom/data/library.js
@@ -567,6 +567,15 @@ Zotero.Library.prototype.updateSearches = Zotero.Promise.coroutine(function* () 
 	this._hasSearches = !!(yield Zotero.DB.valueQueryAsync(sql, this.libraryID));
 });
 
+Zotero.Library.prototype.hasItems = Zotero.Promise.coroutine(function* () {
+	if (!this.id) {
+		throw new Error("Library is not saved yet");
+	}
+	let sql = 'SELECT COUNT(*)>0 FROM items WHERE libraryID=?';
+
+	return Zotero.DB.valueQueryAsync(sql, this.libraryID);
+});
+
 Zotero.Library.prototype.hasItem = function (item) {
 	if (!(item instanceof Zotero.Item)) {
 		throw new Error("item must be a Zotero.Item");
--- a/chrome/content/zotero/xpcom/sync/syncRunner.js
+++ b/chrome/content/zotero/xpcom/sync/syncRunner.js
@@ -136,14 +136,20 @@ Zotero.Sync.Runner_Module = function (options = {}) {
 		
 		try {
 			let client = this.getAPIClient({ apiKey });
-			
 			let keyInfo = yield this.checkAccess(client, options);
-			if (!keyInfo) {
+			
+			let emptyLibraryContinue = yield this.checkEmptyLibrary(keyInfo);
+			if (!emptyLibraryContinue) {
 				this.end();
-				Zotero.debug("Syncing cancelled");
+				Zotero.debug("Syncing cancelled because user library is empty");
 				return false;
 			}
-			
+
+			if (!Zotero.Users.getCurrentUserID()) {
+				Zotero.Users.setCurrentUserID(keyInfo.userID);
+				Zotero.Users.setCurrentUsername(keyInfo.username);
+			}
+
 			let engineOptions = {
 				apiClient: client,
 				caller: this.caller,
@@ -230,6 +236,45 @@ Zotero.Sync.Runner_Module = function (options = {}) {
 		if (!json.access) throw new Error("'access' not found in key response");
 		
 		return json;
+	});
+
+
+	// Prompt if library empty and there is no userID stored
+	this.checkEmptyLibrary = Zotero.Promise.coroutine(function* (keyInfo) {
+		let library = Zotero.Libraries.userLibrary;
+		let userID = Zotero.Users.getCurrentUserID();
+
+		if (!userID) {
+			let hasItems = yield library.hasItems();
+			if (!hasItems) {
+				let ps = Services.prompt;
+				let index = ps.confirmEx(
+					null,
+					Zotero.getString('general.warning'),
+					Zotero.getString('sync.warning.emptyLibrary', [keyInfo.username, Zotero.clientName]) + "\n\n"
+						+ Zotero.getString('sync.warning.existingDataElsewhere', Zotero.clientName),
+					(ps.BUTTON_POS_0 * ps.BUTTON_TITLE_IS_STRING) 
+						+ (ps.BUTTON_POS_1 * ps.BUTTON_TITLE_CANCEL)
+						+ (ps.BUTTON_POS_2 * ps.BUTTON_TITLE_IS_STRING),
+					Zotero.getString('sync.sync'),
+					null, 
+					Zotero.getString('dataDir.changeDataDirectory'), 
+					null, {}
+				);
+				if (index == 1) {
+					return false;
+				}
+				else if (index == 2) {
+					var win = Services.wm.getMostRecentWindow("navigator:browser");
+					win.openDialog("chrome://zotero/content/preferences/preferences.xul", null, null, {
+						pane: 'zotero-prefpane-advanced',
+						tabIndex: 1
+					});
+					return false;
+				}
+			}
+		}
+		return true;
 	});
 	
 	
--- a/chrome/locale/en-US/zotero/zotero.properties
+++ b/chrome/locale/en-US/zotero/zotero.properties
@@ -112,6 +112,7 @@ dataDir.notFound						= The Zotero data directory could not be found.
 dataDir.previousDir					= Previous directory:
 dataDir.useProfileDir				= Use %S profile directory
 dataDir.selectDir					= Select a Zotero data directory
+dataDir.changeDataDirectory 		= Change Data Directory…
 dataDir.unsafeLocation.selected.dropbox = Choosing a data directory within Dropbox may corrupt your database.
 dataDir.unsafeLocation.selected.useAnyway = Use this directory anyway?
 dataDir.unsafeLocation.existing.dropbox = Your Zotero data directory is within Dropbox, which may lead to data corruption.
@@ -828,6 +829,8 @@ sync.localDataWillBeCombined		= If you continue, local Zotero data will be combi
 sync.localGroupsWillBeRemoved1		= Local groups, including any with changed items, will also be removed from this computer.
 sync.avoidCombiningData				= To avoid combining data, revert to the '%S' account or use the Reset options in the Sync pane of the Zotero preferences.
 sync.unlinkWarning					= Are you sure you want to unlink this account?\n\n%S will no longer sync your data, but your data will remain locally.
+sync.warning.emptyLibrary			= You are about to sync the ‘%1$S’ account to an empty %2$S database. This could happen if you removed your previous %2$S database or if the location of your data directory changed.
+sync.warning.existingDataElsewhere  = If your %S data exists elsewhere on your computer, you should move it to your current data directory or change your data directory location to point to the existing data.
 
 sync.conflict.autoChange.alert			= One or more locally deleted Zotero %S have been modified remotely since the last sync.
 sync.conflict.autoChange.log			= A Zotero %S has changed both locally and remotely since the last sync:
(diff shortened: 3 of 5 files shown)


dstillman:
Instead of running a query to update `hasItems` on every item addition/deletion, we should just run a query on-demand in `Zotero.Library::hasItems()`, which will be called at sync time when there's no userID.

(Also, this will need to be rebased on top of my WebDAV commit — sorry.)

chrome/content/zotero/xpcom/sync/syncRunner.js
```diff
@@ -230,6 +236,24 @@ Zotero.Sync.Runner_Module = function (options = {}) {
...
+
+		if(!userID && !library.hasItems()) {
+			if(! Services.prompt.confirm(
+				null,
+				Zotero.getString('general.warning'),
+				Zotero.getString('sync.warning.emptyLibrary', Zotero.clientName)
+			)) {
+				return false;
+			}
+		}
```
dstillman: We generally avoid `confirm`, because it doesn't provide control over the button titles. Should use `confirmEx` with `sync.sync` and either `sync.cancel` or just "Cancel".

adomasven:
I had worked more with different zotero items since and realized, that `updateItems`/`hasItems` could possibly include attachments and notes as well, since they can be top-level. Is this correct and should I update accordingly?

adomasven:
Changes pushed.

dstillman:
Not sure what you mean about attachments and notes, but isn't this still running an extra query on every item addition/deletion?

I don't think we need/want an `updateItems` at all, and I don't think we need this to be library-specific. I think there could just be a single function with no parameters that does a `COUNT(*) > 0` for items across the whole DB. I'm not sure where that should go — maybe `Zotero.Items.hasItems()`?

adomasven:
Whoops, didn't remove `updateItems` calls where I should have. 

I created `hasItems` in line with `hasCollections` and `hasSearches`, which then made sense to have them in `Zotero.Library`. In general, I still think it might make more sense to have a per-library `hasItems` function as it could come in handy in the future. Something which does `COUNT(*) > 0` for all items would probably be less versatile.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -828,6 +828,7 @@ sync.localDataWillBeCombined		= If you continue, local Zotero data will be combi
 sync.localGroupsWillBeRemoved1		= Local groups, including any with changed items, will also be removed from this computer.
 sync.avoidCombiningData				= To avoid combining data, revert to the '%S' account or use the Reset options in the Sync pane of the Zotero preferences.
 sync.unlinkWarning					= Are you sure you want to unlink this account?\n\n%S will no longer sync your data, but your data will remain locally.
+sync.warning.emptyLibrary			= Your are about to sync to an empty %1$S database. This might be due to a new installation %1$S or the location of your data directory changing.\n\nDo you want to continue?
```
dstillman: This actually shouldn't happen for a new Zotero installation, because a new installation will need to set the userID in the prefs before syncing, right? So this would only be a new database, either from a deleted previous database or a new data directory, with an existing Firefox or Zotero profile with an API key stored.

I also think we should provide a third button to open the data dir prefs.

sync.warning.emptyDatabase = You are about to sync the ‘%1$S’ account to an empty %2$S database. This could happen if you removed your previous %2$S database or if the location of your data directory changed.
sync.warning.existingDataElsewhere = If your %1$S data exists elsewhere on your computer, you should move it to your current data directory or change your data directory location to point to the existing data.

dataDir.changeDataDirectory = Change Data Directory…

chrome/content/zotero/xpcom/sync/syncRunner.js
```diff
@@ -230,6 +236,24 @@ Zotero.Sync.Runner_Module = function (options = {}) {
...
 	});
+
+
+	// Prompt if library empty and there is no userID stored
+	this.checkEmptyLibrary = Zotero.Promise.coroutine(function* (client) {
+		let library = Zotero.Libraries.userLibrary;
+		let userID = Zotero.Users.getCurrentUserID();
+
+		if(!userID && !library.hasItems()) {
+			if(! Services.prompt.confirm(
```
dstillman: Style: no space after `!`

dstillman: Oops, outdated diff, but same note for a bunch of different places

chrome/content/zotero/xpcom/data/library.js
```diff
@@ -567,6 +570,22 @@ Zotero.Library.prototype.updateSearches = Zotero.Promise.coroutine(function* ()
 	this._hasSearches = !!(yield Zotero.DB.valueQueryAsync(sql, this.libraryID));
 });
 
+Zotero.Library.prototype.hasItems = Zotero.Promise.coroutine(function* () {
+	if (! this.id) {
+		throw new Error("Library is not saved yet");
+	}
+	if (this._hasItems === null) {
+		yield this.updateItems();
+	}
```
dstillman: This will never be run, because `hasItems` is populated during library load. We can have a per-library hasItems — that's not technically the test we want for this, but I think it wouldn't matter because you can't have groups without a userID, and not having a userID is a prereq for this to be called — but it should still be run on-demand and not cached, since we're not updating it like the other two.

Right now, if you delete the DB, sync, get the warning, cancel, create a new item (or import thousands of items), and sync again, you get the warning again.

chrome/locale/en-US/zotero/zotero.properties
adomasven: Ah yes, I wrote this string when I thought I'd check with the remote repository whether there were any items there and haven't updated for logic changes.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -828,6 +829,8 @@ sync.localDataWillBeCombined		= If you continue, local Zotero data will be combi
 sync.localGroupsWillBeRemoved1		= Local groups, including any with changed items, will also be removed from this computer.
 sync.avoidCombiningData				= To avoid combining data, revert to the '%S' account or use the Reset options in the Sync pane of the Zotero preferences.
 sync.unlinkWarning					= Are you sure you want to unlink this account?\n\n%S will no longer sync your data, but your data will remain locally.
+sync.warning.emptyLibrary			= Your are about to sync your account to an empty %1$S database. This could happen if you removed your previous %1$S database or if the location of your data directory changed.
```
adomasven: Says *your account* instead of *the '%1$S' account*, because with empty database there is no reliable way to get the account. We could use prefs, but it does not have to be populated.

dstillman: Sorry, meant to clarify — we can just move the check after `checkAccess`, such that we'll already have `keyInfo.username`, right?

adomasven:
Amended.

chrome/content/zotero/xpcom/data/library.js
```diff
@@ -305,6 +307,7 @@ Zotero.Library.prototype._loadDataFromRow = function(row) {
 	
 	this._hasCollections = !!row.hasCollections;
 	this._hasSearches = !!row.hasSearches;
+	this._hasItems = !!row.hasItems;
```
dstillman: Remove

chrome/content/zotero/xpcom/data/library.js
```diff
@@ -79,6 +80,7 @@ Zotero.defineProperty(Zotero.Library, '_rowSQLSelect', {
 	value: "L.libraryID, " + Zotero.Library._dbColumns.map(function(c) "L." + c + " AS " + Zotero.Library._colToProp(c)).join(", ")
 		+ ", (SELECT COUNT(*)>0 FROM collections C WHERE C.libraryID=L.libraryID) AS hasCollections"
 		+ ", (SELECT COUNT(*)>0 FROM savedSearches S WHERE S.libraryID=L.libraryID) AS hasSearches"
+		+ ", (SELECT COUNT(*)>0 FROM items I WHERE I.libraryID=L.libraryID) as hasItems"
```
dstillman: Remove

chrome/content/zotero/xpcom/data/library.js
```diff
@@ -33,6 +33,7 @@ Zotero.Library = function(params = {}) {
 	
 	this._hasCollections = null;
 	this._hasSearches = null;
+	this._hasItems = null;
```
dstillman: Remove

chrome/content/zotero/xpcom/data/library.js
```diff
@@ -567,6 +570,16 @@ Zotero.Library.prototype.updateSearches = Zotero.Promise.coroutine(function* ()
 	this._hasSearches = !!(yield Zotero.DB.valueQueryAsync(sql, this.libraryID));
 });
 
+Zotero.Library.prototype.hasItems = Zotero.Promise.coroutine(function* () {
+	if (!this.id) {
+		throw new Error("Library is not saved yet");
+	}
+	let sql = 'SELECT COUNT(*)>0 FROM items WHERE libraryID=?';
+	let hasItems = !!(yield Zotero.DB.valueQueryAsync(sql, this.libraryID));
```
dstillman: return directly

chrome/content/zotero/xpcom/sync/syncRunner.js
```diff
@@ -231,6 +237,39 @@ Zotero.Sync.Runner_Module = function (options = {}) {
 		
 		return json;
 	});
+
+
+	// Prompt if library empty and there is no userID stored
+	this.checkEmptyLibrary = Zotero.Promise.coroutine(function* (keyInfo) {
+		let library = Zotero.Libraries.userLibrary;
+		let userID = Zotero.Users.getCurrentUserID();
+		let hasItems = yield library.hasItems();
```
dstillman: This should be a very quick query, but still no reason to run it unless `!userID`.

chrome/content/zotero/xpcom/data/item.js
```diff
@@ -3710,6 +3710,12 @@ Zotero.Item.prototype._eraseData = Zotero.Promise.coroutine(function* (env) {
 });
 
 
+Zotero.Item.prototype._finalizeErase = Zotero.Promise.coroutine(function* (env) {
+	yield Zotero.Item._super.prototype._finalizeErase.call(this, env);
+
+});
+
+
```
dstillman: Remove

adomasven:
Ah, sorry about those leftovers! Should be doing these things on my own before committing!

dstillman:
I was actually thinking it was better to just cancel the sync and send people to the prefpane, since if we use `chooseZoteroDirectory()` directly you can't see what the current directory is and you can't toggle between the default dir and a custom one. When people encounter this and don't want to go ahead with the sync, there's a decent chance there's some confusion or uncertainty where the extra context that the prefpane provides might be helpful.

dstillman:
(Oh, and also can't use Show Data Directory, which might be necessary for moving stuff to the right place.)

adomasven:
Pushed

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -828,6 +829,8 @@ sync.localDataWillBeCombined		= If you continue, local Zotero data will be combi
 sync.localGroupsWillBeRemoved1		= Local groups, including any with changed items, will also be removed from this computer.
 sync.avoidCombiningData				= To avoid combining data, revert to the '%S' account or use the Reset options in the Sync pane of the Zotero preferences.
 sync.unlinkWarning					= Are you sure you want to unlink this account?\n\n%S will no longer sync your data, but your data will remain locally.
+sync.warning.emptyLibrary			= Your are about to sync the ‘%1$S‘ account to an empty %2$S database. This could happen if you removed your previous %2$S database or if the location of your data directory changed.
```
dstillman: `‘%1$S’`

dstillman: and "You are"

dstillman:
All right, can squash. (For the individual commits, when referencing a ticket, be sure it also says what's being fixed. Usually that's as simple as "Closes #nnn, <title of #nnn>", if it's short enough. That way a git log doesn't require going to GitHub to see what was fixed.)

adomasven:
Squished and pushed.

PR #889, opened 2015-12-27
fixes regression in note export from https://github.com/zotero/zotero…

adam3smith:
…/commit/47bf9c38e96e41d066a1c267365a6010be42b038

### DIFF
--- a/chrome/content/zotero/xpcom/utilities_internal.js
+++ b/chrome/content/zotero/xpcom/utilities_internal.js
@@ -497,7 +497,11 @@ Zotero.Utilities.Internal = {
 				item.linkMode = zoteroItem.attachmentLinkMode;
 				item.mimeType = item.contentType;
 			}
-			
+			// preserve notes
+			if(item.note) { 
+				item.uniqueFields.note = item.note;
+			}
+
 			return item;
 		}
 	},


chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -497,7 +497,11 @@ Zotero.Utilities.Internal = {
 				item.linkMode = zoteroItem.attachmentLinkMode;
 				item.mimeType = item.contentType;
 			}
-			
+			// preserve notes
+			if(item.note) { 
+				item.uniqueFields.note = item.note;
+			}
+
```
aurimasv: This is for attachment note, right? Then it should be in the `if` statement just above. You can drop this `if` statement. Also, you can drop the "Fix linkMode" comment above too, since it is quite self-explanatory and no longer makes sense.

adam3smith: no, this fixes standalone notes, which otherwise have empty uniqueFields

adam3smith:
Relevant Zotero forum thread: https://forums.zotero.org/discussion/55489/problems-with-zotero-rdf/

adam3smith:
any chance we could get this in?  @aurimasv  -- any objections on this beyond removing the comment?

chrome/content/zotero/xpcom/utilities_internal.js
aurimasv: OK, could you add a space after `if` then? I'd also prefer to move this next to other unique fields (L456)

aurimasv:
Please fix the tests for this: https://github.com/zotero/zotero/blob/4.0/test/tests/translateTest.js#L223

adam3smith:
OK on moving and space but don't follow on the test? Do you just mean add a test for this--because it's passing right now and I don't see how setting uniqueFields should affect any of those tests?

aurimasv:
Right, the test doesn't currently cover this. You should add a check to make sure that field is set in legacy mode.

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -497,7 +497,11 @@ Zotero.Utilities.Internal = {
 				item.linkMode = zoteroItem.attachmentLinkMode;
 				item.mimeType = item.contentType;
```
aurimasv: let's also assign this to `item.uniqueFields.mimeType`

aurimasv:
Could you also move L457-459 next to L419-421 (one of the lines is actually duplicated)

dstillman:
Can't say I actually understand the uniqueFields stuff, but that seems to fix this and attachment linkMode/contentType. Tests will have to happen on 5.0.

Issue #849, opened 2015-09-24
Make API_URL and SYNC_URL available as config options


ralight:
It would be very helpful if the API_URL and SYNC_URL options could be exposed (as advanced options/through about:config only if necessary). This will make it easier to use a dataserver hosted in a different place once extension signing hits.

Thanks for your consideration.

dstillman:
Closed by https://github.com/zotero/zotero/commit/084f5a72df24912820724e8fe46b39b837deb554

PR #901, opened 2016-01-24
Expose sync server in preferences

eseifert:
Expose SYNC_URL by creating inputs for the sync server URL to make it easier to use a custom dataserver. At the moment, changing the server URL requires a patched Zotero Firefox add-on, which is particularly inconvenient considering forced extensions signing (see issue #849).

### DIFF
--- a/chrome/content/zotero/preferences/preferences_sync.xul
+++ b/chrome/content/zotero/preferences/preferences_sync.xul
@@ -31,6 +31,7 @@
 			helpTopic="sync">
 		<preferences>
 			<preference id="pref-sync-autosync" name="extensions.zotero.sync.autoSync" type="bool"/>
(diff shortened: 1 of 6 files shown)


adam3smith:
see Dan's comment here: https://groups.google.com/d/msg/zotero-dev/tF1M3FZ-GLk/s5r2w6zyqPwJ
in particular:
> All of this is to say that, as things stand now, we're still not comfortable supporting third-party installations or the users who use them, and we feel those users should get easily identifiable patched versions from the people running the server installations. 

@dstillman may have changed his mind or be willing to rethink, but that was the last time this came up.
Also, given the compromise Zotero helped forge with the mozilla folks on extension signing, getting patched Zotero versions signed is now pretty viable.

PR #902, opened 2016-01-25
Feed Reader UI

adomasven:
Okay, here goes.

Based of Aurimas's initial work from a year ago, rebased on current master, imposed upon new data-layer changes, with a bunch of tests added (and bugs hunted down in the process).

~~Translation is not supported yet and will be submitted as a separate PR.~~
Translation added on the same branch.

~~Note:
Expiration logic is now as follows: the feed items are removed from the feed after X days (specified in feed options) from marking them as _read_. I am not sure if this is the most intuitive behaviour.~~
Expired items are only removed if they are no longer provided on the RSS feed too. This prevents having empty feeds for those that update infrequently


chrome/content/zotero/feedSettings.js
```diff
@@ -0,0 +1,188 @@
...
+					cleanUrl = ios.newURI('http://' + url, null, null).spec;
+				} catch (e) {}
+			}
+			throw e;
+		}
+		
+		if (!cleanUrl) return;
+		
+		if (/^https?:\/\/[^\/\s]+\/\S/.test(cleanUrl)) return cleanUrl;
+	};
```
aurimasv: https://github.com/zotero/zotero/blob/master/chrome/content/zotero/xpcom/attachments.js#L693 has the same (from a quick glance) logic. We can probably move it to a more common place and reuse.

chrome/content/zotero/feedSettings.js
```diff
@@ -0,0 +1,188 @@
...
+			document.getElementById('feed-title').value = data.title;
+		}
+		
+		let ttl;
+		if (data.ttl !== undefined) {
+			ttl = Math.floor(data.ttl / 60);
+		} else {
+			ttl = 1;
+		}
+		document.getElementById('feed-ttl').value = ttl;
```
aurimasv: Not sure why I originally thought this was a good idea, but I think that if users can only set the TTL in hour increments, we should store it as such (rather than minutes). Maybe there was a good reason though.

chrome/content/zotero/feedSettings.js
```diff
@@ -0,0 +1,188 @@
...
+		
+		let url = cleanURL(document.getElementById('feed-url').value);
+		urlTainted = false;
+		if (!url) return;
+		
+		try {
+			let fr = feedReader = new Zotero.FeedReader(url);
+			yield fr.process();
+			let feed = fr.feedProperties;
+			if (feedReader !== fr || urlTainted) return;
```
aurimasv: Maybe make a comment about the logic here, since it's not very obvious. The idea is that if `validateUrl`/`invalidateUrl` were called some time during `fr.process()`, then we should abort further processing, since this is an outdated "session"

chrome/content/zotero/feedSettings.js
aurimasv: Oh yeah, the RSS spec defines TTL in minutes. I guess nvm.

chrome/content/zotero/xpcom/collectionTreeView.js
```diff
@@ -1263,8 +1331,8 @@ Zotero.CollectionTreeView.prototype._rememberOpenStates = Zotero.Promise.corouti
 		
 		var open = this.isContainerOpen(i);
 		
-		// Collections default to closed
-		if (!open && treeRow.isCollection()) {
+		// Collections and feeds default to closed
+		if (!open && treeRow.isCollection() && treeRow.isFeed()) {
```
aurimasv: That should be an `||` no?

chrome/content/zotero/xpcom/data/feed.js
```diff
@@ -267,7 +295,150 @@ Zotero.Feed.prototype._finalizeSave = Zotero.Promise.coroutine(function* (env) {
...
+			// Delete invalid data
+			delete item.guid;
+			delete item.dateAdded;
+			
+			feedItem.fromJSON(item);
+			toAdd.push(feedItem);
+		}
+		
+		// Save in reverse order
+		let savePromises = new Array(toAdd.length);
```
aurimasv: This can be removed, since we save sequentially.

chrome/content/zotero/xpcom/data/feed.js
```diff
@@ -89,10 +112,22 @@ Zotero.defineProperty(Zotero.Feed.prototype, 'isFeed', {
...
+		this._updating = !!v;
+		Zotero.Notifier.trigger('statusChanged', 'feed', this.id);
+	}
+});
 
 (function() {
 // Create accessors
-let accessors = ['name', 'url', 'refreshInterval', 'cleanupAfter'];
+let accessors = ['name', 'url', 'refreshInterval', 'cleanupAfter', 
+	'lastCheck', 'lastUpdate', 'lastCheckError'];
```
aurimasv: I think it made sense to make these getters only, since these values should not be getting modified outside of this class.

chrome/content/zotero/xpcom/data/feedItem.js
```diff
@@ -135,6 +140,18 @@ Zotero.FeedItem.prototype._saveData = Zotero.Promise.coroutine(function* (env) {
 	}
 });
 
+Zotero.FeedItem.prototype.toggleRead = Zotero.Promise.coroutine(function* (state) {
+	state = state !== undefined ? !!state : !this.isRead;
+	let changed = this.isRead != state;
+	this.isRead = state;
```
aurimasv: I suppose it doesn't make much sense to update this if it hasn't changed. I think it would be appropriate to move this into the `if` statement.

chrome/content/zotero/xpcom/data/feedItems.js
```diff
@@ -94,6 +94,35 @@ Zotero.FeedItems = new Proxy(function() {
...
+			for (let item of items) {
+				if (!item.isRead) {
+					state = true;
+					break;
+				}
+			}
+		}
+		
+		for (let i=0; i<items.length; i++) {
+			items[i].toggleRead(state);
```
aurimasv: @dstillman would it help performance to wrap this into a single DB transaction? Could also avoid updating unread count for each item.

dstillman: Yes, writes are much faster when wrapped in a transaction.

dstillman: Also this is missing a `yield`.

chrome/content/zotero/xpcom/data/feeds.js
```diff
@@ -98,16 +98,75 @@ Zotero.Feeds = new function() {
 		return this._cache.libraryIDByURL[url] !== undefined;
 	}
 	
-	this.getAll = function() {
+	this.getAll = function(ids) {
```
aurimasv: Isn't that what `.get` method is for?

chrome/content/zotero/xpcom/data/feedItems.js
```diff
@@ -94,6 +94,35 @@ Zotero.FeedItems = new Proxy(function() {
...
 	
+	this.toggleReadByID = Zotero.Promise.coroutine(function* (ids, state) {
+		if (!Array.isArray(ids)) {
+			if (typeof ids != 'string') throw new Error('ids must be a string or array in Zotero.FeedItems.toggleReadByID');
+			
+			ids = [ids];
+		}
+		let items = yield this.getAsync(ids);
+		
+		if (state == undefined) {
```
dstillman: If this is really supposed to test for `undefined` as opposed to false, it should be `===`.

chrome/content/zotero/xpcom/data/feedItems.js
aurimasv: Omission of `yield`, I think is intentional. We don't need to do this sequentially (at least in the current setup).

chrome/content/zotero/xpcom/data/feedItems.js
dstillman: Er, scratch that: `false` != `undefined`. (It's `null` == `undefined`, and `!state` would be true for both `false` and `undefined`. But this will be false for `false`.)

chrome/content/zotero/xpcom/data/item.js
```diff
@@ -3751,7 +3754,14 @@ Zotero.Item.prototype.fromJSON = function (json) {
...
 			if (val) {
-				let d = Zotero.Date.isoToDate(val);
+				let d = new Date(val);
+				if (isNaN(d.getTime())) {
+					d = Zotero.Date.sqlToDate(val, true);
+				}
+				if (isNaN(d.getTime())) {
+					d = Zotero.Date.strToDate(val);
+					d = new Date(d.year, d.month, d.day);
+				}
```
aurimasv: The data being parsed is supposed to follow a fairly strict standard. When do we have non-ISO dates being passed to this function?

chrome/content/zotero/xpcom/data/feedItem.js
```diff
@@ -135,6 +140,18 @@ Zotero.FeedItem.prototype._saveData = Zotero.Promise.coroutine(function* (env) {
...
+	if (changed) {
+		yield this.saveTx({skipEditCheck: true, skipDateModifiedUpdate: true});
+		
+		let feed = Zotero.Feeds.get(this.libraryID);
+		feed.updateUnreadCount();
```
dstillman: Missing `yield`

chrome/content/zotero/xpcom/feedReader.js
```diff
@@ -0,0 +1,538 @@
...
+		let language = Zotero.FeedReader._getFeedField(feed, 'language', 'dc')
+			|| Zotero.FeedReader._getFeedField(feed, 'language');
+		if (language) info.language = language;
+		
+		let ttl = Zotero.FeedReader._getFeedField(feed, 'ttl');
+		if (ttl) info.ttl = ttl;
+		
+		this._feedProperties = info;
+		this._feed = feed;
+		return info;
```
aurimasv: seems like nothing is being done with the returned value, so no point in returning it.

chrome/content/zotero/xpcom/collectionTreeView.js
```diff
@@ -200,7 +202,34 @@ Zotero.CollectionTreeView.prototype.refresh = Zotero.Promise.coroutine(function*
 		}),
 		added++
 	);
-	
+
+	// TODO: Unify feed and group adding code
+	// Add feeds
+	var feeds = Zotero.Feeds.getAll();
```
dstillman: Missing `yield`

dstillman: Sorry, not async. (Was thinking of `Zotero.Items.getAll()`.)

chrome/content/zotero/xpcom/feedReader.js
```diff
@@ -0,0 +1,538 @@
...
+				let feedItem = Zotero.FeedReader._getFeedItem(item, this._feedProperties);
+				if (!feedItem) continue;
+				
+				let lastItem = this._feedItems[this._feedItems.length - 1];
+				this._feedItems.push(Zotero.Promise.defer()); // Push a new deferred promise so an iterator has something to return
+				lastItem.resolve(feedItem);
+			}
+
+		this._feedProcessed.resolve();
+		}
```
aurimasv: Something doesn't look right here. I think `this._feedProcessed.resolve();` and `}` are flipped here (that would account for the indenting issue too). Also, we need a `.catch` before the `.finally` that would `.reject()` the `this._feedProcessed`, otherwise things might hang if there's an error.

chrome/content/zotero/xpcom/data/item.js
dstillman: Yeah, this wouldn't be the right place for this — `fromJSON()` is parsing our own data format.

chrome/content/zotero/xpcom/feedReader.js
```diff
@@ -0,0 +1,538 @@
...
+			// so something could have ended up in the hour _or_ minute field
+			date = getFeedField(feedEntry, 'pubDate')
+				/* In case it was magically pulled from some other field */
+				|| ( date.getUTCFullYear() + '-'
+					+ (date.getUTCMonth() + 1) + '-'
+					+  date.getUTCDate() );
+		} 
+		else {
+			date = Zotero.FeedReader._formatDate(date);
+			// Add time zone
```
aurimasv: leftover comment?

chrome/content/zotero/xpcom/data/feedItems.js
dstillman: That's only for very top-level code to decide, though (e.g., event handlers). Data layer functions should finish their work before returning. (If they don't, we end up with race conditions in tests, etc.)

chrome/content/zotero/xpcom/feedReader.js
```diff
@@ -0,0 +1,538 @@
...
+			items.dateModified = date;
+		}
+	}
+	
+	if (!item.dateModified) {
+		// When there's no reliable modification date, we can assume that item doesn't get updated
+		Zotero.debug("FeedReader: Feed item missing a modification date (" + item.guid + ")");
+	}
+	
+	if (item.dateModified) {
```
aurimasv: just an else

chrome/content/zotero/xpcom/feedReader.js
```diff
@@ -0,0 +1,538 @@
...
+	let domFragment = feedText.createDocumentFragment(domDiv);
+	return Zotero.Utilities.dom2text(domFragment, field);
+};
+
+/*
+ * Format JS date as SQL date
+ */
+Zotero.FeedReader._formatDate = function(date) {
+	return Zotero.Date.dateToSQL(date, true);
+}
```
aurimasv: Unnecessary wrapper.

chrome/content/zotero/xpcom/data/feedItems.js
dstillman: (Yielding on `Zotero.Promise.all()` would be an option, but write transactions are serialized anyway, so we don't gain anything from the parallelism like we would if this were, say, a bunch of network requests. So better to just do it sequentially, and in a transaction (which requires it to use `yield` anyway).)

chrome/content/zotero/xpcom/itemTreeView.js
```diff
@@ -260,6 +261,44 @@ Zotero.ItemTreeView.prototype.setTree = Zotero.Promise.coroutine(function* (tree
...
+	var colId, col, dir, currentCol, currentDir;
+	
+	for (let i=0, len=this._treebox.columns.count; i<len; i++) {
+		let column = this._treebox.columns.getColumnAt(i);
+		if (column.element.getAttribute('sortActive')) {
+			currentCol = column;
+			currentDir = column.element.getAttribute('sortDirection');
+		}
+		column.element.removeAttribute('sortActive');
+		column.element.removeAttribute('sortDirection');
```
aurimasv: move this inside the if and break

chrome/content/zotero/xpcom/itemTreeView.js
```diff
@@ -260,6 +261,44 @@ Zotero.ItemTreeView.prototype.setTree = Zotero.Promise.coroutine(function* (tree
 });
 
 
+Zotero.ItemTreeView.prototype.setSortColumn = function() {
+	var colId, col, dir, currentCol, currentDir;
```
aurimasv: use let and declare variables as they are used.

chrome/content/zotero/xpcom/itemTreeView.js
```diff
@@ -260,6 +261,44 @@ Zotero.ItemTreeView.prototype.setTree = Zotero.Promise.coroutine(function* (tree
...
+		column.element.removeAttribute('sortActive');
+		column.element.removeAttribute('sortDirection');
+	}
+	
+	colId = Zotero.Prefs.get('itemTree.sortColumnId');
+	dir = Zotero.Prefs.get('itemTree.sortDirection');
+	// Restore previous sort setting
+	if (! this.collectionTreeRow.isFeed() && colId) {
+		col = this._treebox.columns.getNamedColumn(colId);
+		dir = Zotero.Prefs.get('itemTree.sortDirection');
```
aurimasv: This is redundant with L278

chrome/content/zotero/xpcom/itemTreeView.js
```diff
@@ -260,6 +261,44 @@ Zotero.ItemTreeView.prototype.setTree = Zotero.Promise.coroutine(function* (tree
...
+		Zotero.Prefs.set('itemTree.sortColumnId', null);
+		Zotero.Prefs.set('itemTree.sortDirection', null);
```
aurimasv: Why are we setting these to `null`?

chrome/content/zotero/xpcom/uri.js
```diff
@@ -172,7 +188,7 @@ Zotero.URI = new function () {
 			return path;
 		}
 		
-		if (obj instanceof Zotero.Item) {
+		if (obj instanceof Zotero.Item || obj instanceof Zotero.Feed) {
```
aurimasv: Is that supposed to be `Zotero.FeedItem`? In which case, it is an instance of `Zotero.Item` anyway, so this is not necessary.

chrome/content/zotero/zoteroPane.js
```diff
@@ -515,6 +515,18 @@ var ZoteroPane = new function()
 					this.highlightTimer = null;
 				}
 				ZoteroPane_Local.collectionsView.setHighlightedRows();
+				return;
+			} else if (event.keyCode == event.DOM_VK_BACK_QUOTE) {
```
aurimasv: This was a quick hack I put in and I think I got feedback the backtick might not be the best choice for the toggle (although I quite like it TBH). I believe this needs to be configurable.

chrome/content/zotero/zoteroPane.js
```diff
@@ -843,6 +855,20 @@ var ZoteroPane = new function()
...
+	this.newFeed = Zotero.Promise.coroutine(function* () {
+		let data = {};
+		window.openDialog('chrome://zotero/content/feedSettings.xul', 
+			null, 'centerscreen, modal', data);
+		if (!data.cancelled) {
+			let feed = new Zotero.Feed();
+			feed.url = data.url;
+			feed.name = data.title;
+			feed.refreshInterval = data.ttl;
+			feed.cleanupAfter = data.cleanAfter;
```
aurimasv: Might as well make these match up.

chrome/content/zotero/xpcom/data/feed.js
```diff
@@ -89,10 +112,22 @@ Zotero.defineProperty(Zotero.Feed.prototype, 'isFeed', {
...
+Zotero.defineProperty(Zotero.Feed.prototype, 'unreadCount', {
+	get: function() this._feedUnreadCount
+});
+Zotero.defineProperty(Zotero.Feed.prototype, 'updating', {
+	get: function() this._updating,
+	set: function(v) {
+		if (!v == !this._updating) return; // Unchanged
+		this._updating = !!v;
+		Zotero.Notifier.trigger('statusChanged', 'feed', this.id);
+	}
```
aurimasv: We should make this a getter instead of getter/setter. The getter should return `!!this._updating`. Then use `this._updating` to store the promise for the update. See `._updateFeed` below for more on this

chrome/content/zotero/xpcom/data/feed.js
```diff
@@ -267,7 +295,150 @@ Zotero.Feed.prototype._finalizeSave = Zotero.Promise.coroutine(function* (env) {
...
+			}
+		}
+	} catch(e) {
+		Zotero.debug("Error clearing expired feed items.");
+		Zotero.debug(e);
+	}
+});
+
+Zotero.Feed.prototype._updateFeed = Zotero.Promise.coroutine(function* () {
+	this.updating = true;
```
aurimasv: We should actually make sure that this doesn't get re-triggered until updating is done. Save a deferred promise in `this._updating` and resolve it at the end, then clear `this._updating`. Then, before anything else, return `this._updating` here if we're in the process of updating.

chrome/content/zotero/zoteroPane.js
dstillman: Can make it configurable, but I don't have a particular objection to backtick. (For what it's worth, Thunderbird uses 'M' (**M**ark?), and we do crib their number-keys approach to tagging.)

chrome/content/zotero/zoteroPane.js
```diff
@@ -1901,6 +1923,35 @@ var ZoteroPane = new function()
...
+		};
+		
+		window.openDialog('chrome://zotero/content/feedSettings.xul', 
+			null, 'centerscreen, modal', data);
+		if (data.cancelled) return;
+		
+		feed.name = data.title;
+		feed.refreshInterval = data.ttl;
+		feed.cleanupAfter = data.cleanAfter;
+		yield feed.save({skipEditCheck: true});
```
aurimasv: `.saveTx` and I don't think we need the `skipEditCheck` any more (this dates back to when feeds extended collections, rather than libraries)

chrome/locale/en-US/zotero/zotero.dtd
```diff
@@ -254,6 +258,17 @@
...
 
+<!ENTITY zotero.feedSettings.title                      "Feed Settings">
+<!ENTITY zotero.feedSettings.saveButton.label           "Save">
+<!ENTITY zotero.feedSettings.url.label                  "URL">
+<!ENTITY zotero.feedSettings.title.label                "Title">
+<!ENTITY zotero.feedSettings.refresh.label1             "Refresh Interval:">
+<!ENTITY zotero.feedSettings.refresh.label2             "hour(s)">
+<!ENTITY zotero.feedSettings.title.label                "Title">
+<!ENTITY zotero.feedSettings.cleanAfter.label1         "Remove read articles after ">
+<!ENTITY zotero.feedSettings.cleanAfter.label2         "day(s)">
```
aurimasv: We might have to avoid formatting the settings this way for localization purposes. The "day(s)" and "hour(s)" may not make sense placed after the number in other locales. Let's turn this into a single string. Something like "Remove read articles after (days):" (Or something better)

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -173,23 +175,31 @@ pane.collections.library		= My Library
...
+pane.collections.feedLibraries      = Feeds
 pane.collections.trash			= Trash
 pane.collections.untitled		= Untitled
 pane.collections.unfiled		= Unfiled Items
 pane.collections.duplicate		= Duplicate Items
 
 pane.collections.menu.rename.collection			= Rename Collection…
-pane.collections.menu.edit.savedSearch			= Edit Saved Search
+pane.collections.menu.edit.savedSearch			= Edit Saved Search…
+pane.collections.menu.edit.savedSearch			= Edit Feed…
```
aurimasv: should be `.feed`

chrome/locale/en-US/zotero/zotero.dtd
dstillman: We do this currently in the General pane: "Automatically remove items in the trash deleted more than" [x] "days ago"

I think it's OK in theory, since localizers can put whatever they want in either side to make it make sense. If they translate literally on Transifex it might not work, but it can be fixed by someone who tries in the software itself.

Mozilla (who's generally pretty careful about its localization — and has whole pluralization systems that are far more detailed and correct than what we do) does this in a number of places in the Thunderbird prefs:

"After displaying for" [x] "seconds"
"Let my contacts know what I am idle after" [x] "minutes of activity"
"Auto Save every" [x] "minutes"

So at least to the extent that TB still represents Mozilla's practices, it seems to be OK by their standards.

aurimasv:
Thanks, Adomas, for taking over. I know that many of the comments I made are probably directed at my own code. Sorry about that! There might be some deeper issues that I have not touched on yet, but that's all I have for now.

IIRC, there is an issue with the current data model where the same item in two different feeds could cause problems, since we store feed-specific information based on guid alone. Because feeds are now libraries, each feed gets a separate copy of item data even though the guid is the same. We probably want to start storing feed data by guid + libraryID.

chrome/content/zotero/xpcom/data/feed.js
adomasven: Yes, however `_set()` has some handling code for lastCheckError and generally there has been code like `this.lastCheck = Zotero.Date.dateToSQL(new Date(), true);` within the class, which looked valid but did nothing. `_set()` could be used within the class, but I went over other classes within Zotero and did not find any variables without setters, so decided this is the more consistent (if less encapsulating) solution.

dstillman: There are definitely some properties without setters in Zotero. E.g., `firstCreator` and `sortCreator` in `Zotero.Item`.

adomasven: So should I go for something like 
```
this._set('_feedItemLastCheck', Zotero.Date.dateToSQL(new Date(), true);
``` 
within the class like this: https://github.com/adomasven/zotero/blob/feature/feed-reader-UI/chrome/content/zotero/xpcom/data/feed.js#L404 ?

chrome/content/zotero/xpcom/data/feedItems.js
adomasven: Which is why always sticking to strict comparators is a good idea in general.

chrome/content/zotero/xpcom/data/feed.js
dstillman: There's no special handling for that in `_set()`, so why not just assign it directly like other private properties, such as `this._feedUnreadCount` in `updateUnreadCount`?

dstillman: But yes, if it needs special handling, that'd be fine. Otherwise good to use direct access for performance reasons (though I doubt that matters here).

adomasven: `this._changed` is managed down the line by `Zotero.Item.prototype._set`, required for saving. `this.unreadCount` is a calculated property, not stored in the database. Which I now realise, has problems, because it never gets initialized when a feed is created (restored from DB).

adomasven: Scratch that, it is being initialized with initial load.

chrome/content/zotero/xpcom/itemTreeView.js
adomasven: Most of these are declared in if statements, which would make them fall out of block scope. Overall, common consensus between using `var` or `let` is not reached within the codebase, I feel. @dstillman prefers var for function scope, I myself would prefer to either sticking with one or the other, without mixing.

chrome/content/zotero/xpcom/itemTreeView.js
adomasven: To be explicit about unsetting this value. Is there an preferred way of doing this?

chrome/content/zotero/zoteroPane.js
adomasven: Consensus? I like backtick myself too, easy to touch-locate and useful for multi-hand scenarios, if you have the right hand on the mouse/touchpad/arrow keys.

chrome/content/zotero/xpcom/itemTreeView.js
dstillman: We don't do this in any case — these should just use `var` on first assignment, or `let` if limited to block scope.

My argument for `var` has been that being able to (re)initialize variables at function scope close to where they're used without worrying about whether they're declared above or below is a good feature.

adomasven: By the end of the day, sticking to what makes sense should be the end goal. And I feel like this makes the most sense, knowing these variables can get initialized within if statements or be left undefined completely. It's sensible to have them declared there right at the top, as a reminder that they may or may not be initialized.

dstillman: Doesn't really matter with a small function like this, but as a general style rule, I'd say this is fine for `col`, `currentCol`, and `currentDir`, since those might not be set, but `colId` and `dir`, which are always assigned at top level, should have `var` on their respective lines.

adomasven: Agreed

chrome/content/zotero/xpcom/itemTreeView.js
dstillman: Zotero.Prefs.clear() to unset a pref

chrome/content/zotero/xpcom/data/feed.js
adomasven: This loses the magic of notifier being triggered. Any reason to prefer the no-magic way?

chrome/content/zotero/zoteroPane.js
adomasven: `skipEditCheck` still relevant. Performed for any `dataObject`, checks `isEditable()`

dstillman: A `Library` isn't a `DataObject`. Just modeled similarly.

chrome/content/zotero/xpcom/itemTreeView.js
aurimasv: >My argument for `var` has been that being able to (re)initialize variables at function scope close to where they're used without worrying about whether they're declared above or below is a good feature.

I don't think I've ever seen a case where that is a good idea. If you're declaring a variable inside a block that is intended to be visible outside of block, you should be making it visible explicitly to help readability. If you are redeclaring a variable in the same block, you are clearly not aware of the other declaration, or you wouldn't be doing so, so being informed of that is a great idea.

The issue with declaring everything up top for me is that half way down the function, when a new variable starts being used, I have to remember if it was declared at top. Most of the time, I just assume it was.

chrome/content/zotero/xpcom/itemTreeView.js
aurimasv: I guess my question is more along the lines of "why are we unsetting this?" Seems like that should only be done if we're displaying items specifically in an unsorted state.

chrome/content/zotero/zoteroPane.js
aurimasv: I think it's a good default. Can make it properly configurable later, if needed.

aurimasv: Oh, before I forget, one of the requests was to add read/unread toggles to context menus for item and collection trees.

adomasven: Noted https://github.com/adomasven/zotero/issues/2

chrome/content/zotero/xpcom/itemTreeView.js
dstillman: I'm talking about something like this:

```javascript
function myFunction(objs) {
    var obj;
    while (obj = objs.shift()) {
        [...]
    }
    
    // a bunch of code

    var obj;
    while (obj = otherObjs.shift()) {
        [...]
    }
}
```
With `let`, the code at the bottom would throw an error because `obj` is reused, even though they're both temporary variables closely tied to their sections of code. You have to either 1) use a different variable name below, even if you're purposely following a similar code pattern, or 2) keep track of the usage above, which means leaving out the `let` below (which seems undesirable for the same reasons that declaring variables at the top of the function is undesirable) but being sure to add it back in if the above block is removed (though "use strict;" would at least help with that).

chrome/content/zotero/xpcom/itemTreeView.js
adomasven: Overall this functions has some complicated logic, because there are 4 scenarios: 
- feed -> feed
- non-feed -> non-feed
- feed -> non-feed
- non-feed -> feed
We only want to store the sort order, if we switched from a non-feed view to a feed view. When we switch to from one feed to another, we don't want to store the previous sort-order, because it's been set to `columnAdded`, hence the check on L290. And when we restore the sort order, after switching back to a non-feed, we clear the pref, so the L290 check has `!colId == true`.

chrome/content/zotero/xpcom/itemTreeView.js
adomasven: Which is why I think just sticking to either only `let` or only `var` for new code would be the best way to go about things. Really, there isn't one best way to do this (different approaches have different drawbacks and benefits). Mixing the 2 leads to fairly fruitless discussions about style, when in reality we are just trying to write code that has best readability and is consistent with itself.

adomasven: Also, in a dynamic language like JS, I think having variable declarations in general makes no sense. E.g. Ruby and Python do not have variable declarations and everything is function scope, which mostly makes sense. Then you can reuse the same variable somewhere else, without worrying about declarations. I don't get it why they didn't use the chance with ES6 to remove declarators completely and instead added 2 more, which make their usage even more ambiguous. I think just sticking to `var` everywhere should be preferable and cause least headaches (however, double declarations cause problems in strict mode, so using `let` which has the hard enforcement also makes sense).

dstillman: This isn't really a question of inconsistency. `let` should definitely be used for block scope, as we've done for years. But that doesn't mean that it's inconsistent to use `var for function scope if we agree that there are benefits.

aurimasv: I'd argue that (1) is the right thing to do, but very well.

adomasven: I think it should be a question of consistency, purely to avoid debates about which declarator should have been used, because either *can* makes sense, and at least in this case are readable (although I did move the declaration of `colId` further down in the commit I'm working on right now).

chrome/content/zotero/xpcom/itemTreeView.js
aurimasv: OK, though I think I would add comments to make the logic clearer. Also, IIRC, there was some discussion of making treeview columns different for feed vs non-feed view, but this looks ok for now.

chrome/content/zotero/xpcom/itemTreeView.js
dstillman: > I think it should be a question of consistency, purely to avoid debates about which declarator should have been used

That makes it sound like using one or the other is ambiguous, though. There's not any ambiguity if the rule is just "`var` for function scope, `let` for block scope". I'm not 100% wedded to that, but it's what we've done for years, and I'm just explaining that there's been a justification for it. (Places where `let` isn't used within blocks is generally old code.)

>  I'd argue that (1) is the right thing to do, but very well.

That could get annoying, though:

```javascript
var sql = "SELECT...";
Zotero.DB.query(sql);

sql = "INSERT..."
Zotero.DB.query(sql);

//
// more code
//

// Did I use 'sql' above? Do I need to go check? Or run without and wait for a strict mode warning?
var sql = "SELECT...";
Zotero.DB.query(sql);
```

And using different variables each time would be distracting.

dstillman: (`let` everywhere also means that things break if you throw a `try` around something while debugging. I really think they both have their purposes.)

aurimasv: >I don't get it why they didn't use the chance with ES6 to remove declarators completely

Because that's not backwards compatible.

>// Did I use 'sql' above? Do I need to go check?

(1) Huge functions like that probably need some refactoring. If you can't keep track of what the whole function is doing, it's doing too much. (2) Yes, you should check. You should know what your code is doing in the context of the whole function. (3) A thrown exception for reusing `let foo;` is a whole lot easier to catch than unintentionally overwriting a variable. Strict mode does the same, but I don't think we have that declared in all scripts either.

adomasven: This discussion will come up next time someone else starts working with this codebase (and even Aurimas comments on it, because the rule "var for function scope, let for block scope" is kind of ambiguous).
I am fine with sticking to the rule, but we should at least be more lenient about PR comments for variable declarations where both could reasonably make sense.

aurimasv: >(let everywhere also means that things break if you throw a try around something while debugging. I really think they both have their purposes.)

In my experience, if something is throwing, you already know where it's throwing from, so you wrap a single line in try-catch and that's trivial to change as needed.

In conclusion, I'll take safety over convenience.

adomasven:
Changes pushed.

> Thanks, Adomas, for taking over. I know that many of the comments I made are probably directed at my own code. Sorry about that!
No worries.

> IIRC, there is an issue with the current data model where the same item in two different feeds could cause problems, since we store feed-specific information based on guid alone. Because feeds are now libraries, each feed gets a separate copy of item data even though the guid is the same. We probably want to start storing feed data by guid + libraryID.
The issue is indeed there. Given we are delaying the translation of feed items for better performance, sharing feed items between feeds may be favourable. This would mean that an item would have to be translated once for multiple feeds. In which case some data-layer restructuring may be in order to support many-to-many relationships between feed items and feeds.
This may be a good time to also consider whether we may want to introduce many-to-many relationships between all items and libraries, given the plans to introduce multiple user libraries and subscriptions, which would presumably also extend `Zotero.Library`, potentially causing a lot of item duplication.

dstillman:
Outside of feeds, we link items between libraries with behind-the-scenes relations links. We'd like to start doing more with merging updated data across those links, but having separate items in separate libraries avoids a lot of problems, so we avoid use the same data in different libraries directly.

For this, if two entries in different feeds have the same URL, we might as well use the translated data for both, but I don't think we need to rework the whole data layer for this case, which will be rare and I think fairly unique.

adomasven:
Developments since last code review:
- Test coverage is good on relevant features (+ some bugs for feed parsing weeded out)
- Item translation added on select and on drag to library. Shows progress window during translation on drag.
- Feed items sorted by natural order + an option in preferences to change the sort order added
- Context menu options added to toggle read state for feeds and items. There's a preference to change the read state key in Shortcut Preferences.

Things worth considering:
- What to show on the right side item pane for feed items and feeds.
- Preference placement for feed sort order.

chrome/content/zotero/feedSettings.js
```diff
@@ -0,0 +1,172 @@
...
+			title.disabled = false;
+			ttl.disabled = false;
+			document.getElementById('feed-cleanupAfter').disabled = false;
+			document.documentElement.getButton('accept').disabled = false;
+		}
+		catch (e) {
+			Zotero.debug(e);
+		}
+		finally {
+			if (feedReader === fr) feedReader = null;
```
dstillman: This throws `fr is not defined` every time a feed URL is entered. (This would be a case where I personally would just use `var`.)

More generally, there are all sorts of errors being thrown from that dialog, for various inputs:

```
uri is not defined feedSettings.js:45:0

NS_ERROR_NOT_AVAILABLE: Component returned failure code: 0x80040111 (NS_ERROR_NOT_AVAILABLE) [nsIChannel.contentType] FeedProcessor.js:1292:0

NS_ERROR_UNEXPECTED: Component returned failure code: 0x8000ffff (NS_ERROR_UNEXPECTED) [nsISAXXMLReader.onStopRequest] FeedProcessor.js:1298:0

e is undefined bluebird.js:99:0

fr is not defined feedSettings.js:143:0
```

dstillman:
feedsettings

- The "Advanced Options" should be vertically centered with the dropmarker.
- The "URL" and "Title" labels should have colons, and they should be right-aligned. The text fields should be left aligned. (See General pane of prefs for an example.)
- The URL field shouldn't be a search field for now. (We'll hopefully have a way to search for feeds eventually, though.)

dstillman:
`itemTreeView::cycleHeader()` doesn't seem to be properly disabled in feed view. I can still on a column and change the sort indicator, if not the actual sort.

dstillman:
Trying to create an item manually in a feed throws an error, as does editing. It looks like the view isn't properly being marked as non-editable for some reason — the UI should be disabled automatically for non-editable libraries.

dstillman:
You can keep the classes/code for the feed-refreshing icon, but let's remove that image and stick to the regular feed icon for now — we'll do something better than those arrows.

dstillman:
If we're only going to have an Info tab in the right-hand pane, we should just remove the whole tab bar in this view.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -160,6 +160,8 @@ pane.collections.delete			= Are you sure you want to delete the selected collect
 pane.collections.delete.keepItems = Items within this collection will not be deleted.
 pane.collections.deleteWithItems.title = Delete Collection and Items
 pane.collections.deleteWithItems = Are you sure you want to delete the selected collection and move all items within it to the Trash?
+pane.feed.deleteWithItems.title = Delete Feed and Items
+pane.feed.deleteWithItems = Are you sure you want to delete the selected feed and all items within it?
```
dstillman: I think the items can be implicit here, and we can just refer to deleting the feed itself. (Talking about deleting items might even be a little scary if people have dragged items from the feed.)

chrome/skin/default/zotero/zotero.css
```diff
@@ -430,6 +430,14 @@ label.zotero-text-link {
 	max-width: 29.5em;
 }
 
+.zotero-advanced-options>.zotero-advanced-options-label>dropmarker {
+	list-style-image: url("chrome://browser/skin/toolbarbutton-dropdown-arrow.png");
```
dstillman: Oh, part of the issue with the blurriness may be that this — you generally can't actually rely on browser images, because they're not the same cross-platform. And the dropmarker seems to have a default image anyway. If I disable this rule, the same icon remains but the blurriness goes away. (It still shifts on rotate, though.)

test/content/support.js
```diff
@@ -409,6 +427,10 @@ function getTestDataDirectory() {
 	       QueryInterface(Components.interfaces.nsIFileURL).file;
 }
 
+function getTestDataItemUrl(path) {
```
dstillman: `getTestDataURL`

test/content/support.js
```diff
@@ -409,6 +427,10 @@ function getTestDataDirectory() {
...
+	return OS.Path.join("resource://zotero-unit-tests/data", path);
```
dstillman: `OS.Path.join()` shouldn't be used for URLs — on Windows, that uses '\'. Can just make sure there's a '/ at the beginning of `path` and concatenate.

chrome/content/zotero/bindings/itembox.xml
```diff
@@ -341,7 +341,9 @@
 							fieldNames.push(Zotero.ItemFields.getName(fields[i]));
 						}
 						
-						fieldNames.push("dateAdded", "dateModified");
+						if (! (this.item instanceof Zotero.FeedItem)) {
```
dstillman: Nit: No space after `!` (here and elsewhere)

chrome/content/zotero/feedSettings.js
```diff
@@ -0,0 +1,172 @@
...
+		urlTainted = false;
+	
+	let cleanURL = function(url) {
+		let cleanUrl = Zotero.Utilities.cleanURL(url, true);
+
+		if (cleanUrl) {
+			if (/^https?:\/\/[^\/\s]+\/\S/.test(cleanUrl)) {
+				return cleanUrl;
+			} else {
+				Zotero.debug(uri.scheme + " is not a supported protocol for feeds.");
```
dstillman: No period at the end of debug statements (here and elsewhere)

chrome/content/zotero/feedSettings.js
```diff
@@ -0,0 +1,172 @@
...
+		if (data.url) {
+			document.getElementById('feed-url').value = data.url;
+		}
+		
+		if (!data.url) {
+			this.invalidateUrl();
+		} else {
+			// Do not allow to change URL for existing feed
+			document.getElementById('feed-url').readOnly = true;
+		}
```
dstillman: Combine with above block

chrome/content/zotero/feedSettings.js
```diff
@@ -0,0 +1,172 @@
...
+			ttl = 1;
+		}
+		document.getElementById('feed-ttl').value = ttl;
+		
+		let cleanupAfter = data.cleanupAfter;
+		if (cleanupAfter === undefined) cleanupAfter = 2;
+		document.getElementById('feed-cleanupAfter').value = cleanupAfter;
+		
+		if (data.url && !data.urlIsValid) {
+			this.validateUrl();
```
dstillman: Usually better not to leave async functions non`yield`ing — something could be added later below this call that expects the URL to be validated already. Better to make use `coroutine` on `init` and add a `yield` here.

chrome/content/zotero/feedSettings.js
```diff
@@ -0,0 +1,172 @@
...
+		}
+	};
+	
+	this.invalidateUrl = function() {
```
dstillman: Our style is to capitalize acronyms when camel-casing, so `invalidateURL()`.

chrome/content/zotero/feedSettings.js
```diff
@@ -0,0 +1,172 @@
...
+		if (!urlIsValid) return;
+		
+		urlIsValid = false;
+		document.getElementById('feed-title').disabled = true;
+		document.getElementById('feed-ttl').disabled = true;
+		document.getElementById('feed-cleanupAfter').disabled = true;
+		document.documentElement.getButton('accept').disabled = true;
+	};
+	
+	this.validateUrl = Zotero.Promise.coroutine(function* () {
```
dstillman: `validateURL`

chrome/content/zotero/feedSettings.xul
```diff
@@ -0,0 +1,50 @@
...
+	<script src="include.js"/>
+	<script src="feedSettings.js"/>
+	
+	<vbox>
+		<hbox align="center">
+			<label value="&zotero.feedSettings.url.label;" control="feed-url"/>
+			<textbox id="feed-url" flex="1" type="search" size="2"
+				oninput="Zotero_Feed_Settings.invalidateUrl()"
+				oncommand="Zotero_Feed_Settings.validateUrl()"
+				focused="true" newlines="replacewithspaces"
```
dstillman: As long as this is a URL field, might as well do `newlines="stripsurroundingwhitespace"`, which trims internal and external whitespace.

dstillman:
Refresh interval should be in minutes, not hours, even if it defaults to 60.

dstillman:
We can't make the General pane of the prefs any longer than it is — it already doesn't fit on many screens. For now, at least, we can create a Feeds tab in the Advanced section of the prefs, and we'll either convert that into something more general or find another place for it later.

Can do this later, but we also probably want to have default global refresh/TTL settings, with the Advanced Options section giving the option between the defaults and overrides.

chrome/content/zotero/preferences/preferences_advanced.xul
```diff
@@ -217,37 +218,37 @@
 						<rows id="zotero-keys-rows">
 							<row id="zotero-keys-new-item">
 								<label value="&zotero.preferences.keys.newItem;" control="textbox-newItem"/>
-								<label/>
+								<label class="cmd-shift"/>
```
dstillman: Should probably call this `modifier` for clarity.

chrome/content/zotero/preferences/preferences_general.xul
```diff
@@ -43,6 +43,8 @@
 			<preference id="pref-groups-copyChildFileAttachments" name="extensions.zotero.groups.copyChildFileAttachments" type="bool"/>
 			<preference id="pref-groups-copyChildLinks" name="extensions.zotero.groups.copyChildLinks" type="bool"/>
 			<preference id="pref-groups-copyTags" name="extensions.zotero.groups.copyTags" type="bool"/>
+			
+			<preference id="pref-feeds-sortAsc" name="extensions.zotero.feeds.sortAsc" type="bool"/>
```
dstillman: `sortAscending`

chrome/content/zotero/preferences/preferences_general.xul
```diff
@@ -130,6 +132,18 @@
...
+			<caption label="&zotero.preferences.feeds;"/>
+			
+			<label value="&zotero.preferences.feeds.sorting;"/>
+			<vbox style="margin-left: 2em">
+				<radiogroup id="feed-sort" orient="horizontal" align="center" preference="pref-feeds-sortAsc">
+					<radio label="&zotero.preferences.feeds.sorting.newestFirst;" value="false"/>
+					<radio label="&zotero.preferences.feeds.sorting.oldestFirst;" value="true"/>
+				</radiogroup>
+			</vbox>
+		</groupbox>
```
dstillman: I think a menulist would be better:

"Sorting: [oldest | newest] items first"

chrome/content/zotero/xpcom/attachments.js
```diff
@@ -688,32 +688,13 @@ Zotero.Attachments = new function(){
...
-						return ios.newURI('http://' + uri, null, null).spec;
-					} catch (e) {}
-				}
-				
-				Zotero.debug('cleanAttachmentURI: Invalid URI: ' + uri, 2);
-				return false;
-			}
-			throw e;
-		}
+		return Zotero.Utilities.cleanURL(uri, tryHttp);
```
dstillman: This should have a debug line:

"Zotero.Attachments.cleanAttachmentURI() is deprecated -- update your code"

(We can replace the existing calls in short order, but we'll need a reminder.)

chrome/content/zotero/xpcom/collectionTreeView.js
```diff
@@ -1263,8 +1331,8 @@ Zotero.CollectionTreeView.prototype._rememberOpenStates = Zotero.Promise.corouti
 		
 		var open = this.isContainerOpen(i);
 		
-		// Collections default to closed
-		if (!open && treeRow.isCollection()) {
+		// Collections and feeds default to closed
+		if (!open && treeRow.isCollection() || treeRow.isFeed()) {
```
dstillman: Add explicit parens around `!open && treeRow.isCollection()` for clarity.

chrome/content/zotero/xpcom/collectionTreeView.js
```diff
@@ -1850,8 +1927,25 @@ Zotero.CollectionTreeView.prototype.drop = Zotero.Promise.coroutine(function* (r
 			return;
 		}
 		
+		var items = yield Zotero.Items.getAsync(ids);
+		if (!items) {
```
dstillman: `items` is only possibly false if an integer is passed to `getAsync()`. If an array is passed, it would be an empty array.

chrome/locale/en-US/zotero/zotero.properties
adomasven: Hmm. Maybe something like "Unsubscribe" could be more fitting?

chrome/content/zotero/xpcom/data/feed.js
```diff
@@ -267,7 +298,160 @@ Zotero.Feed.prototype._finalizeSave = Zotero.Promise.coroutine(function* (env) {
...
 
-Zotero.Feed.prototype._finalizeErase = Zotero.Promise.method(function(env) {
+Zotero.Feed.prototype.getExpiredFeedItemIDs = Zotero.Promise.coroutine(function* () {
+	let sql = "SELECT itemID AS id FROM feedItems "
+		+ "LEFT JOIN items I USING (itemID) "
+		+ "WHERE I.libraryID=? "
+		+ "AND readTime IS NOT NULL "
+		+ "AND julianday('now', 'utc') - (julianday(readTime, 'utc') + ?) > 0";
+	let expiredIDs = yield Zotero.DB.queryAsync(sql, [this.id, {int: this.cleanupAfter}]);
+	return expiredIDs.map(row => row.id);
```
dstillman: You can just use `columnQueryAsync()`.

chrome/locale/en-US/zotero/zotero.properties
dstillman: Yeah, let's try that.

chrome/content/zotero/bindings/itembox.xml
adomasven: I am having serious problems fighting this one style urge.

chrome/content/zotero/xpcom/data/feed.js
```diff
@@ -267,7 +298,160 @@ Zotero.Feed.prototype._finalizeSave = Zotero.Promise.coroutine(function* (env) {
...
+			Zotero.debug(e);
+		}
+		this._set('_feedLastCheckError', e.message || 'Error processing feed');
+	}
+	if (toAdd.length) {
+		yield Zotero.DB.executeTransaction(function* () {
+			// Save in reverse order
+			for (let i=toAdd.length-1; i>=0; i--) {
+				// Saving currently has to happen sequentially so as not to violate the
+				// unique constraints in itemDataValues (FIXME)
```
dstillman: Can take out this FIXME — DB operations aren't ever going to be parallelized. That'd really only be for network operations.

dstillman: (By the way, I'm just going through the full diff, so I'm sure I'm commenting on some things from the original code.)

chrome/content/zotero/xpcom/data/feedItem.js
```diff
@@ -128,16 +171,116 @@ Zotero.FeedItem.prototype._saveData = Zotero.Promise.coroutine(function* (env) {
...
+			collection = yield Zotero.Collections.getAsync(collectionID);
+		}
+		win.Zotero_Browser.progress.show();
+		win.Zotero_Browser.progress.Translation.scrapingTo(libraryID, collection);
+	}
+	
+	// Load document
+	let hiddenBrowser = Zotero.HTTP.processDocuments(
+		this.getField('url'), 
+		(item) => deferred.resolve(item),
```
dstillman: Style: No parens around single arrow-function arguments

chrome/content/zotero/xpcom/data/feedItem.js
```diff
@@ -128,16 +171,116 @@ Zotero.FeedItem.prototype._saveData = Zotero.Promise.coroutine(function* (env) {
...
+	translate.setDocument(doc);
+	
+	// Load translators
+	deferred = Zotero.Promise.defer();
+	translate.setHandler('translators', (me, translators) => deferred.resolve(translators));
+	translate.getTranslators();
+	let translators = yield deferred.promise;
+	if (! translators || !translators.length) {
+		Zotero.debug("No translators detected for FeedItem " + this.id + " with url " + this.getField('url'), 2);
+		throw new Zotero.Error("No translators detected for FeedItem " + this.id + " with url " + this.getField('url'))
```
dstillman: "feed item", "URL"

chrome/content/zotero/xpcom/data/feedItem.js
```diff
@@ -128,16 +171,116 @@ Zotero.FeedItem.prototype._saveData = Zotero.Promise.coroutine(function* (env) {
...
+	
+	// clean itemData
+	const deleteFields = ['attachments', 'notes', 'id', 'itemID', 'path', 'seeAlso', 'version', 'dateAdded', 'dateModified'];
+	for (let field of deleteFields) {
+		delete itemData[field];
+	}	
+	
+	this.fromJSON(itemData);
+	this.isTranslated = true;
+	this.forceSaveTx();
```
dstillman: Missing `yield`

chrome/content/zotero/xpcom/data/feedItems.js
```diff
@@ -40,6 +40,7 @@ Zotero.FeedItems = new Proxy(function() {
 			let obj = zi_primaryDataSQLParts.call(this);
 			obj.feedItemGUID = "FeI.guid AS feedItemGUID";
 			obj.feedItemReadTime = "FeI.readTime AS feedItemReadTime";
+			obj.feedItemTranslatedTime = "FeI.translatedTime AS feedItemTranslatedTime";
```
dstillman: Unless `FeI` is actually necessary for disambiguation, these should just be `FI`, as we do for other two-word tables.

adomasven:
> If we're only going to have an Info tab in the right-hand pane, we should just remove the whole tab bar in this view.
All good, but leaves with a big padding bit at the top 
image
In HTML I would handle this with jQuery or similar, by setting a class with overridden CSS. What's the standard way to do this with XML? 

> Refresh interval should be in minutes, not hours, even if it defaults to 60.
Why? There's going to be very few feeds that update more often than every few hours, most will be daily. For some feeds people may want to set something like 12 hours, which isn't nice in minutes at all.

On a related note, there still isn't a call to Zotero.Feeds.scheduleNextFeedCheck anywhere on initialization, because I couldn't figure out where the best place for it would be. Advice?

dstillman:
>> If we're only going to have an Info tab in the right-hand pane, we should just remove the whole tab bar in this view.

> All good, but leaves with a big padding bit at the top. In HTML I would handle this with jQuery or similar, by setting a class with overridden CSS. What's the standard way to do this with XUL?  

`display: none` on the `tabbox` and `margin-top: 0` on the `tabpanels` should do it.

>> Refresh interval should be in minutes, not hours, even if it defaults to 60.

> Why? There's going to be very few feeds that update more often than every few hours, most will be daily. For some feeds people may want to set something like 12 hours, which isn't nice in minutes at all.

For scholarly feeds, that's true. But non-scholarly feeds could be updated much more frequently. (And most people won't change the default. The ones that do can probably do a little math if necessary. We could use a menulist for the units but that seems like overkill.)

> On a related note, there still isn't a call to Zotero.Feeds.scheduleNextFeedCheck anywhere on initialization, because I couldn't figure out where the best place for it would be. Advice?

Generally for this sort of thing, kick it off in an `init()` in `Zotero.Feeds` and call that in zotero.js with all the other `init()` calls.

adomasven:
> display: none on the tabbox and margin-top: 0 on the tabpanels should do it.

So something like `element.addAttribute('style', 'padding-top: 0')`? Ewwish, but probably the most reasonable approach.

dstillman:
>> display: none on the tabbox and margin-top: 0 on the tabpanels should do it.

>So something like element.addAttribute('style', 'padding-top: 0')? Ewwish, but probably the most reasonable approach.

Oh, no, sorry if that wasn't clear — you can still just do it via a class on `#zotero-view-tabbox`. You're just changing the top margin that's set for `#zotero-view-tabbox > tabpanels` in overlay.css. (And I meant `display: none` on `tabs`, not `tabbox`.)

dstillman:
I'm not sure what, but we probably want some sort of progress indicator for on-click translation. Have to think about the best approach.

adomasven:
> Oh, no, sorry if that wasn't clear — you can still just do it via a class on #zotero-view-tabbox. You're just changing the top margin that's set for #zotero-view-tabbox > tabpanels in overlay.css. (And I meant display: none on tabs, not tabbox.)

There's some miscommunication happening here. Let's try again. 

Ok, the problem is that the `tabpanels` element has `padding-top: 1.5em`. So the way I would handle this situation is hiding the `tabs` element (seen in the screenshot, with the awful top padding) and then adding a class on the `tabpanels` that overrides the top padding to be `padding-top: 0.25em`. However, there is not easy way to add or remove classes with default JS. Stuff like jQuery provides `addClass()` and `removeClass()` methods in HTML world. The easy way to handle this in our case is to use `tabpanels.addAttribute('style', 'padding-top: 0')`, but that's uglyish. Hence the question, whether there's a better way. Manually editing `tabpanels.className` is even uglier.

dstillman:
You can use `elem.classList.add()/remove()`, or you can just set another attribute and use an attribute selector (e.g., `#zotero-view-tabbox[hidetabs=true]`).

(Also, it's `margin: 12px 0 0 0;` that's the issue here. `padding: 1.5em .25em .25em;` is overridden to 0 within `#zotero-view-tabbox`.)

dstillman:
(Oh, looks like that top margin is OS X–specific, actually (from chrome/content/zotero-platform/mac/overlay.css). So might need to override a few different rules.)

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -173,23 +175,33 @@ pane.collections.library		= My Library
...
 pane.collections.menu.createBib.savedSearch		= Create Bibliography From Saved Search…
+pane.collections.menu.createBib.feed		= Create Bibliography From Feed…
 
 pane.collections.menu.generateReport.collection	= Generate Report from Collection…
 pane.collections.menu.generateReport.savedSearch	= Generate Report from Saved Search…
+pane.collections.menu.generateReport.feed	= Generate Report from Feed…
+
+pane.collections.menu.refresh.feed			= Refresh Feed
+pane.collections.menu.toggleRead.markRead	= Mark Read
+pane.collections.menu.toggleRead.markUnread	= Mark Unread
```
dstillman: "Mark as Read" and "Mark as Unread" are a bit more standard. (Can change the labels to those too.)

Can also drop the `toggleRead`, but these don't really belong here anyway, since they're used in the items list too (and apply more to the items). I'd go with `pane.items.markAsRead`.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -493,6 +505,7 @@ save.link.error					= An error occurred while saving this link.
 save.error.cannotMakeChangesToCollection = You cannot make changes to the currently selected collection.
 save.error.cannotAddFilesToCollection = You cannot add files to the currently selected collection.
 save.error.cannotAddToMyPublications = You cannot save items directly to My Publications. To add items to My Publications, drag them from another library.
+save.error.cannotAddToFeed = You cannot save items directly to feeds.
```
dstillman: "You cannot save items to feeds." (no "directly" here because you can't add items indirectly either, whereas with My Publications you can actually drag items in)

dstillman:
All right, I didn't look at the tests, but that's it for the rest. A bunch of stuff we can do in follow-up patches, but this is great to start.

I have some general concerns about the on-click translation — that it's not clear it's happening; that it's not fast enough to be useful (such that, if we did add link attachments, they would be frustratingly slow to appear when based on translation rather than rel="enclosure"); that people will want to arrow-key quickly through results, unwittingly getting themselves blocked by site firewalls for DOSing... Have to think about this some more. (It's also hard to get a sense of it while translation is broken for a lot of sites.)

adomasven:
> I have some general concerns about the on-click translation — that it's not clear it's happening; that it's not fast enough to be useful (such that, if we did add link attachments, they would be frustratingly slow to appear when based on translation rather than rel="enclosure"); that people will want to arrow-key quickly through results, unwittingly getting themselves blocked by site firewalls for DOSing... Have to think about this some more. (It's also hard to get a sense of it while translation is broken for a lot of sites.)

Agreed. It could be sped up, by keeping a hidden browser open in the background, but only just. Translating in the background is still a possibility. Another related point is whether we should refresh feeds in the background?

> (such that, if we did add link attachments, they would be frustratingly slow to appear when based on translation rather than rel="enclosure")
Regarding enclosed items, I think it is naive to expect that a lot of feeds are going to have a comprehensive list of enclosed items and they probably won't correspond very well to what we would have after translation. Could potentially lead to confusion, when the list of items in the feed do not correspond to items after translation.

chrome/content/zotero/xpcom/storage/storageLocal.js
```diff
@@ -21,6 +21,7 @@ Zotero.Sync.Storage.Local = {
 		switch (libraryType) {
 		case 'user':
 		case 'publications':
+		case 'feed':
```
dstillman: I don't think this should be here — if this is getting called on feed libraries during sync, that's a bug elsewhere that should be fixed.

chrome/content/zotero/zoteroPane.js
```diff
@@ -2469,7 +2455,6 @@ var ZoteroPane = new function()
 			show.push(m.moveToTrash);
 		}
 		
-		show.push(m.sep3, m.exportItems, m.createBib, m.loadReport);
```
dstillman: You removed these from all items, not just feeds.

chrome/content/zotero/xpcom/data/feeds.js
```diff
@@ -68,6 +68,51 @@ Zotero.Feeds = new function() {
...
+				Zotero.debug("Feed " + feed.url + " does not exist in remote JSON. Deleting");
+				yield feed.eraseTx();
+			}
+		}
+		// Because existing json[feed.url] got deleted, `json` now only contains new feeds
+		for (let url in json) {
+			Zotero.debug("Feed " + url + " is being created from remote JSON");
+			let feed = new Zotero.Feed(json[url]);
+			yield feed.saveTx();
+			yield feed.updateFromJSON(json[url]);
```
dstillman: Is this correct? Passing the same object to the constructor and updateFromJSON?

dstillman:
OK, so the idea with ``Zotero.Feeds.restoreFromJSON()`` is that, when an updated 'feeds' synced setting comes in, the sync code (or some observer) will pass it to that, with merge=true if the local setting has changed as well and merge=false if only the remote setting has changed?

A couple things on synced settings:

1) I'm a little concerned with doing everything in one big JSON object. It's certainly easier, but it means that marking an item as read requires sending the full set of read items in all feeds. It's possible that for it to really be a problem someone would have to 1) have tons of feeds and 2) override the default retention settings, but it doesn't seem like too much of a stretch: if someone has 20 feeds, and they retain 200 read items in each one, and each item in the read state object takes up about 100 chars, that's a 400+ KB change each time, both to the disk and the API. For the API we could in theory support PATCH with a JSON diff libary for synced settings, and it'd gzip pretty well, but this still seems like potentially a lot of data to change for every mark-as-read operation.

An alternative would be to 1) store one setting with subscription data for all feeds and 2) store one setting per feed with read state. We'd update these in a transaction such that there were never any orphaned read-state settings. This would scale to any number of feeds, and the size of a change would only be affected by the number of items in a given feed.

If someone tried to retain feed items permanently they'd end up back in a similar situation. For that, we could maybe shard by guid prefix, with the list of shards stored with the feed setting, but to get a decent distribution we'd need to use different guids. Which brings us to…

2) Maybe we want to use a hash of the feed item guid instead of the guid itself? That would limit each read-state entry to 32 bytes + 8 or so bytes, instead of potentially 100 or more, and allow for better sharding.

(There's also a minor privacy element — while syncing hashes of a fixed set of known entries isn't that privacy-enhancing, particularly when some of those read entries may then be synced up as regular Zotero data, it seems a little better to do that than to sync a direct list of all the entries people have looked at recently.)

chrome/content/zotero/xpcom/itemTreeView.js
```diff
@@ -259,6 +260,43 @@ Zotero.ItemTreeView.prototype.setTree = Zotero.Promise.coroutine(function* (tree
...
+			currentDir = column.element.getAttribute('sortDirection');
+			column.element.removeAttribute('sortActive');
+			column.element.removeAttribute('sortDirection');
+			break;
+		}
+	}
+	
+	let colId = Zotero.Prefs.get('itemTree.sortColumnId');
+	// Restore previous sort setting (feed -> non-feed)
+	if (! this.collectionTreeRow.isFeed() && colId) {
```
dstillman: House style: `sortColumnID`, `colID`

test/content/support.js
```diff
@@ -453,8 +453,12 @@ function getTestDataDirectory() {
 	       QueryInterface(Components.interfaces.nsIFileURL).file;
 }
 
-function getTestDataItemUrl(path) {
-	return OS.Path.join("resource://zotero-unit-tests/data", path);
+function getTestDataUrl(path) {
```
dstillman: Style: `getTestDataURL`

chrome/content/zotero/xpcom/data/items.js
```diff
@@ -133,9 +133,10 @@ Zotero.Items = function() {
 	 * @param  {Integer}  libraryID
 	 * @param  {Boolean}  [onlyTopLevel=false]   If true, don't include child items
 	 * @param  {Boolean}  [includeDeleted=false] If true, include deleted items
-	 * @return {Promise<Array<Zotero.Item>>}
+	 * @param  {Boolean}  [onlyIDs=false] 		 If true, resolves only with IDs
+	 * @return {Promise<Array<Zotero.Item|Integer>>}
 	 */
-	this.getAll = Zotero.Promise.coroutine(function* (libraryID, onlyTopLevel, includeDeleted) {
+	this.getAll = Zotero.Promise.coroutine(function* (libraryID, onlyTopLevel, includeDeleted, onlyIDs=false) {
```
dstillman: We call this `asIDs` in other functions.

chrome/content/zotero/zoteroPane.js
```diff
@@ -595,25 +601,36 @@ var ZoteroPane = new function()
 				//event.stopPropagation();
 				return;
 			}
+			else if (command == 'toggleRead') {
+				// Toggle read/unread
+				let row = this.collectionsView.getRow(this.collectionsView.selection.currentIndex);
```
dstillman: `let row = this.collectionsView.selectedTreeRow;`

chrome/content/zotero/zoteroPane.js
```diff
@@ -1886,6 +1917,48 @@ var ZoteroPane = new function()
...
+		let feed = this.collectionsView.getRow(this.collectionsView.selection.currentIndex).ref;
+		let feedItemIDs = yield Zotero.FeedItems.getAll(feed.libraryID, true, false, true);
+		yield Zotero.FeedItems.toggleReadByID(feedItemIDs, true);
+	});
+
+	
+	this.editSelectedFeed = Zotero.Promise.coroutine(function* () {
+		if (!this.collectionsView.selection.count) return;
+		
+		let feed = this.collectionsView.getRow(this.collectionsView.selection.currentIndex).ref;
```
dstillman: `let feed = this.collectionsView.selectedTreeRow.ref;`

adomasven:
> OK, so the idea with Zotero.Feeds.restoreFromJSON() is that, when an updated 'feeds' synced setting comes in, the sync code (or some observer) will pass it to that, with merge=true if the local setting has changed as well and merge=false if only the remote setting has changed?

That's right.

> I'm a little concerned with doing everything in one big JSON object. It's certainly easier, but it means that marking an item as read requires sending the full set of read items in all feeds. It's possible that for it to really be a problem someone would have to 1) have tons of feeds and 2) override the default retention settings, but it doesn't seem like too much of a stretch: if someone has 20 feeds, and they retain 200 read items in each one, and each item in the read state object takes up about 100 chars, that's a 400+ KB change each time, both to the disk and the API. For the API we could in theory support PATCH with a JSON diff libary for synced settings, and it'd gzip pretty well, but this still seems like potentially a lot of data to change for every mark-as-read operation.

> An alternative would be to 1) store one setting with subscription data for all feeds and 2) store one setting per feed with read state. We'd update these in a transaction such that there were never any orphaned read-state settings. This would scale to any number of feeds, and the size of a change would only be affected by the number of items in a given feed.

I was thinking it may be problematic the way it is, but by the time we start splitting feed data up into multiple entries within SyncedSettings of the user library it seems like we're using the wrong tool for the job. When you initially talked about SyncedSettings, I assumed we would treat feeds as libraries regarding sync and each of them would have a list of read items within the SyncedSettings of the library. We can do what you proposed if we just want to bash out sync for this quickly, but if feeds are going to become a serious Zotero feature then I'd say a more tailored solution is in order.

> If someone tried to retain feed items permanently they'd end up back in a similar situation. For that, we could maybe shard by guid prefix, with the list of shards stored with the feed setting, but to get a decent distribution we'd need to use different guids.

Yeah, so I mentioned feed item expiration a couple of times here and there, but it didn't catch anyone's attention. The way it's implemented now is that when you mark an item read it is removed from the feed after the number of days specified to retain items (i.e. the "Remove expired items after X days" setting) IF the item also does not exist in the current XML of the RSS feed itself. Which means that people shouldn't be able to retain a huge number of read feed items. Whether the current expiration strategy is the best solution is a different question, but the above shouldn't be an issue.

> Maybe we want to use a hash of the feed item guid instead of the guid itself? That would limit each read-state entry to 32 bytes + 8 or so bytes, instead of potentially 100 or more, and allow for better sharding.

I'm not sold for the privacy, but for the sake of compression we could do this. Makes restoring from remote sync more complicated, since you now need to hash the GUIDs of every feed item to mark them read, but we could do this.

dstillman:
> The way it's implemented now is that when you mark an item read it is removed from the feed after the number of days specified to retain items (i.e. the "Remove expired items after X days" setting) IF the item also does not exist in the current XML of the RSS feed itself. Which means that people shouldn't be able to retain a huge number of read feed items.

Why not? Can't they say "Remove expired items after 365 days"?

adomasven:
>> The way it's implemented now is that when you mark an item read it is removed from the feed after the number of days specified to retain items (i.e. the "Remove expired items after X days" setting) IF the item also does not exist in the current XML of the RSS feed itself. Which means that people shouldn't be able to retain a huge number of read feed items.

> Why not? Can't they say "Remove expired items after 365 days"?

Fair point.

chrome/content/zotero/xpcom/data/feeds.js
adomasven: It is. From [these fields](https://github.com/zotero/zotero/pull/902/files#diff-d953027eb50324dcf90b1cd22cfe7eaaR340) `feed.updateFromJSON()` uses `markedAsRead` and the rest of the fields are used by the constructor. I anticipated `feed.updateFromJSON()` potentially needing other fields later, so I passed the full JSON, but we could only pass `markedAsRead` to reduce confusion.

chrome/content/zotero/xpcom/data/feeds.js
```diff
@@ -23,6 +23,7 @@
     ***** END LICENSE BLOCK *****
 */
 
+<<<<<<< HEAD
```
dstillman: Looks like this commit is invalid.

dstillman: Actually, don't worry about it — I can just take care of it locally.

dstillman:
:tada: Merged in 6c43e75d269f2a5a165ced63ab1aeaf28f0bcf04 — thanks @adomasven and @aurimasv for all your work on this! (Now let's make it even better.)

PR #927, opened 2016-03-25
Add feed menu buttons. Close adomasven/zotero#9.

adomasven:
Not mentioned in the issue, but also adds a button to mark selected feed items read/unread which is context aware.

### DIFF
--- a/chrome/content/zotero/itemPane.js
+++ b/chrome/content/zotero/itemPane.js
@@ -25,6 +25,7 @@
 
 var ZoteroItemPane = new function() {
 	var _lastItem, _itemBox, _notesLabel, _notesButton, _notesList, _tagsBox, _relatedBox;
+	var _translationTarget;
 	
 	this.onLoad = function () {
 		if (!Zotero) {
@@ -46,6 +47,22 @@ var ZoteroItemPane = new function() {
 		_relatedBox = document.getElementById('zotero-editpane-related');
 	}
 	
+	
+	this.init = function() {
+		Zotero.debug("Initializing ZoteroItemPane");
+		let lastTranslationTarget = Zotero.Prefs.get('feeds.lastTranslationTarget');
+		if (lastTranslationTarget) {
+			if (lastTranslationTarget[0] == "C") {
+				_translationTarget = Zotero.Collections.get(parseInt(lastTranslationTarget.substr(1)));
+			} else if (lastTranslationTarget[0] == "L") {
+				_translationTarget = Zotero.Libraries.get(parseInt(lastTranslationTarget.substr(1)));
+			}
+		}
+		if (!_translationTarget) {
+			_translationTarget = Zotero.Libraries.userLibrary;
+		}	
+		this.setTranslateButton();
+	}
 	
 	/*
 	 * Load a top-level item
@@ -185,6 +202,92 @@ var ZoteroItemPane = new function() {
 	}
 	
 	
+	this.translateSelectedItems = Zotero.Promise.coroutine(function* () {
+		var collectionID = _translationTarget.objectType == 'collection' ? _translationTarget.id : undefined;
+		var items = ZoteroPane_Local.itemsView.getSelectedItems();
+		for (let item of items) {
+			yield item.translate(_translationTarget.libraryID, collectionID);
+		}
+	});
+	
+	
+	this.buildTranslateSelectContextMenu = function (event) {
+		var menu = document.getElementById('zotero-item-addTo-menu');
+		// Don't trigger rebuilding on nested popupmenu open/close
+		if (event.target != menu) {
+			return;
+		}
+		// Clear previous items
+		while (menu.firstChild) {
+			menu.removeChild(menu.firstChild);
+		}
+		
+		var libraries = Zotero.Libraries.getAll();
+		for (let library of libraries) {
+			if (!library.editable || library.libraryType == 'publications') {
+				continue;
+			}
+			Zotero.Utilities.Internal.createMenuForTarget(library, menu, function(event, libraryOrCollection) {
+				ZoteroItemPane.setTranslationTarget(libraryOrCollection);
+				event.stopPropagation();
+			});
+		}
+	};
+	
+	
+	this.setTranslateButton = function() {
+		var label = Zotero.getString('general.addTo', _translationTarget.name);
+		var elem = document.getElementById('zotero-feed-item-addTo-button');
+		elem.setAttribute('label', label);
+
+		var key = Zotero.Keys.getKeyForCommand('saveToZotero');
+		
+		var tooltip = label 
+			+ (Zotero.rtl ? ' \u202B' : ' ') + '(' 
+			+ (Zotero.isMac ? '⇧⌘' : Zotero.getString('general.keys.ctrlShift'))
+			+ key + ')';
+		elem.setAttribute('tooltiptext', tooltip);
+
+		var objectType = _translationTarget._objectType;
+		var imageSrc = Zotero.Utilities.Internal.getCollectionImageSrc(objectType);
+		elem.setAttribute('image', imageSrc);
+	};
+	
+
+	this.setTranslationTarget = function(translationTarget) {
+		_translationTarget = translationTarget;
+		if (translationTarget.objectType == 'collection') {
+			Zotero.Prefs.set('feeds.translationTarget', "C" + translationTarget.id);
+		} else {
+			Zotero.Prefs.set('feeds.translationTarget', "L" + translationTarget.libraryID);
+		}
+		ZoteroItemPane.setTranslateButton();
+	};
+	
+		
+	this.setToggleReadLabel = function() {
+		var markRead = false;
+		var items = ZoteroPane_Local.itemsView.getSelectedItems();
+		for (let item of items) {
+			if (!item.isRead) {
+				markRead = true;
+				break;
+			}
+		}
+		var elem = document.getElementById('zotero-feed-item-toggleRead-button');
+		if (markRead) {
+			var label = Zotero.getString('pane.item.markAsRead');
+		} else {
+			label = Zotero.getString('pane.item.markAsUnread');
+		}
+		elem.setAttribute('label', label);
+
+		var key = Zotero.Keys.getKeyForCommand('toggleRead');
+		var tooltip = label + (Zotero.rtl ? ' \u202B' : ' ') + '(' + key + ')'
+		elem.setAttribute('tooltiptext', tooltip);
+	};
+
+
 	function _updateNoteCount() {
 		var c = _notesList.childNodes.length;
 		
--- a/chrome/content/zotero/itemPane.xul
+++ b/chrome/content/zotero/itemPane.xul
@@ -24,6 +24,7 @@
     ***** END LICENSE BLOCK *****
 -->
 <?xml-stylesheet href="chrome://zotero/skin/itemPane.css" type="text/css"?>
+<?xml-stylesheet href="chrome://zotero-platform/content/itemPane.css" type="text/css"?>
 
 <!DOCTYPE window SYSTEM "chrome://zotero/locale/zotero.dtd">
 
@@ -33,11 +34,21 @@
 	
 	<vbox id="zotero-item-pane" zotero-persist="width">
 		<!-- Trash -->
-		<hbox id="zotero-item-top-buttons-holder" hidden="true">
+		<hbox id="zotero-item-pane-top-buttons-trash" class="zotero-item-pane-top-buttons" hidden="true">
 			<button id="zotero-item-restore-button" label="&zotero.items.menu.restoreToLibrary;"
 				oncommand="ZoteroPane_Local.restoreSelectedItems()"/>
 			<button id="zotero-item-delete-button" label="&zotero.item.deletePermanently;"
 				oncommand="ZoteroPane_Local.deleteSelectedItems()"/>
+		</hbox>
+		
+		<!-- Feed -->
+		<hbox id="zotero-item-pane-top-buttons-feed" class="zotero-item-pane-top-buttons" hidden="true">
+			<button id="zotero-feed-item-toggleRead-button" flex="1"
+				oncommand="ZoteroPane_Local.toggleSelectedItemsRead();"/>
+			<button id="zotero-feed-item-addTo-button" type="menu-button" flex="1"
+				oncommand="ZoteroItemPane.translateSelectedItems()">
+				<menupopup id="zotero-item-addTo-menu" onpopupshowing="ZoteroItemPane.buildTranslateSelectContextMenu(event);"/>
+			</button>
 		</hbox>
 		
 		<!-- Commons -->
--- a/chrome/content/zotero/xpcom/itemTreeView.js
+++ b/chrome/content/zotero/xpcom/itemTreeView.js
@@ -456,14 +456,6 @@ Zotero.ItemTreeView.prototype.notify = Zotero.Promise.coroutine(function* (actio
 		return;
 	}
 	
-	// FeedItem may have changed read/unread state
-	if (type == 'feedItem' && action == 'modify') {
-		for (let i=0; i<ids.length; i++) {
-			this._treebox.invalidateRow(this._itemRowMap[ids[i]]);
-		}
-		return;
-	}
-	
 	if (type == 'search' && action == 'modify') {
 		// TODO: Only refresh on condition change (not currently available in extraData)
 		yield this.refresh();
@@ -482,6 +474,12 @@ Zotero.ItemTreeView.prototype.notify = Zotero.Promise.coroutine(function* (actio
 	}
 	
 	var collectionTreeRow = this.collectionTreeRow;
+
+	if (collectionTreeRow.isFeed() && action == 'modify') {
+		for (let i=0; i<ids.length; i++) {
+			this._treebox.invalidateRow(this._rowMap[ids[i]]);
+		}
+	}
 	
 	var madeChanges = false;
 	var refreshed = false;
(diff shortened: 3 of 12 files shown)


chrome/content/zotero/xpcom/data/item.js
```diff
@@ -1274,7 +1274,7 @@ Zotero.Item.prototype._saveData = Zotero.Promise.coroutine(function* (env) {
 		yield Zotero.DB.queryAsync(sql, env.sqlValues);
 		
 		if (!env.options.skipNotifier) {
-			Zotero.Notifier.queue('modify', 'item', itemID, env.notifierData);
+			Zotero.Notifier.queue('modify', this._objectType, itemID, env.notifierData);
```
dstillman: Hmm, is this right? I would expect various things (e.g., in itemTreeView) not to work properly if a 'feedItem' event is passed instead of 'item'. But maybe it also avoids things we don't want?

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -271,6 +271,7 @@ pane.item.duplicates.writeAccessRequired	= Library write access is required to m
 pane.item.duplicates.onlyTopLevel			= Only top-level full items can be merged.
 pane.item.duplicates.onlySameItemType		= Merged items must all be of the same item type.
 
+pane.item.addToLibrary			= Add to %S
```
dstillman: Can probably just be `general.addTo`

chrome/content/zotero/xpcom/data/item.js
adomasven: ItemTreeView actually has code to handle feedItem correctly, but this wasn't set, so it was doing a lot more work than necessary. I came across this accidentally, while thinking of using notifier (didn't in the end), but may as well fix this.

dstillman: OK. Various other things also expect 'item', but maybe not anything we actually want to run for feed items. Can just keep an eye out for glitches.

adomasven:
Updated with feed button icons and deasyncified collection functions. Needs attention on OSX, because the button with a side menu looks like this:
OSX Button

chrome/skin/default/zotero/itemPane.css
```diff
@@ -67,3 +67,29 @@
...
+/* Set to hidden in user-agent css for some reason. */
+#zotero-feed-item-addTo-button .button-icon {
+	display: block;
+	margin-right: 5px
+}
+
+/* Set to hidden in user-agent css for some reason. */
+#zotero-feed-item-addTo-button .menu-iconic-left {
+	visibility: visible;
+}
```
dstillman: Remove these — set class `menuitem-iconic`/`menu-iconic` on the menuitem/menu instead.

https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/menuitem#Style_classes
https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/menu#Style_classes

chrome/content/zotero/itemPane.xul
```diff
@@ -33,13 +33,23 @@
 	
 	<vbox id="zotero-item-pane" zotero-persist="width">
 		<!-- Trash -->
-		<hbox id="zotero-item-top-buttons-holder" hidden="true">
+		<hbox id="zotero-item-pane-top-buttons-trash" hidden="true">
```
dstillman: Need to update this id in itemPane.css. Also, the style of the new buttons should match the styles used for the trash buttons there (e.g., `-moz-appearance: toolbar;`, and `-moz-box-flex` on the buttons), so should use a class for both.

chrome/content/zotero/itemPane.js
```diff
@@ -185,6 +189,132 @@ var ZoteroItemPane = new function() {
...
+		
+		var libraries = Zotero.Libraries.getAll();
+		for (let library of libraries) {
+			if (!library.editable || library.libraryType == 'publications') {
+				continue;
+			}
+			let libraryMenu = _createTranslationMenuForTarget(library);
+			menu.appendChild(libraryMenu);
+		}
+	};
```
dstillman: We'll want to do the same thing in zoteroPane.js for the items list context menu, and possibly other places, so some of this should just be a function in utilities_internal.js (along with `createTranslationMenuForTarget`) that takes a menu element to populate and an array of libraries.

chrome/skin/default/zotero/itemPane.css
```diff
@@ -67,3 +67,29 @@
...
+
+#zotero-feed-item-addTo-button button {
+	overflow: hidden;
+	text-overflow: ellipsis;
+}
+
+/* Set to hidden in user-agent css for some reason. */
+#zotero-feed-item-addTo-button .button-icon {
+	display: block;
+	margin-right: 5px
```
dstillman: These can be removed as per below, but do need to set `height: 16px;` here to prevent double-size icons on HiDPI OS X.

dstillman:
The "Mark as Read"/"Mark as Unread" label doesn't update when the read state changes, except by clicking away and back.

Proper way to do this is probably to add a `registerObserver()` for 'item' events to `itemPane.js::onLoad` and unregister in an `onUnload` (which would be added as an event listener at the bottom of itemPane.js). I'm also fine with a hack in `itemTreeView::notify` for now (commented as such) that just checks `isFeed()` and calls `ZoteroItemPane.setToggleReadLabel()` from there when there's a 'feedItem' 'modify'.

chrome/content/zotero/itemPane.js
```diff
@@ -185,6 +189,132 @@ var ZoteroItemPane = new function() {
...
+			menu.label = label;
+			menu.type;
+			let menupopup = document.createElementNS(ns, 'menupopup');
+			menu.appendChild(menupopup);
+			return menu;
+		}
+		
+		var imageSrc = _getCollectionImageSrc(libraryOrCollection._objectType);
+		var menuitem = _createMenuitem(libraryOrCollection.name, libraryOrCollection.id, imageSrc,
+				ZoteroItemPane.setTranslationTarget.bind(this, libraryOrCollection));
```
dstillman: Instead of the `_blockTranslate` stuff above, you can just do this:

```
		var menuitem = _createMenuitem(
			libraryOrCollection.name,
			libraryOrCollection.id,
			imageSrc,
			function (event) {
				ZoteroItemPane.setTranslationTarget(libraryOrCollection);
				event.stopPropagation();
			}
		);
```

`setTranslationTarget` will trigger but not the command on the button.

chrome/content/zotero/itemPane.js
```diff
@@ -185,6 +189,132 @@ var ZoteroItemPane = new function() {
...
+
+	/**
+	 * Create a libraryOrCollection DOM tree to place in <menupopup> element.
+	 * 
+	 * @return {Node<menuitem>/Node<menu>}
+	 */
+	function _createTranslationMenuForTarget(libraryOrCollection) {
+		var ns = "http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul";
+		function _createMenuitem(label, value, icon, command) {
+			let menuitem = document.createElementNS(ns, 'menuitem');
```
dstillman: `createElementNS` shouldn't be necessary — `document` should be a `XULDocument` already (and will be if you use `menu.ownerDocument` after moving this function).

dstillman:
We should add the shortcut key in parentheses to the Mark as Read/Unread tooltip, but " (`)" is kind of impossible to decipher, so I'm somewhat inclined to switch to "M" (which is what Thunderbird uses), which would be much easier to see. It also has the advantage of there being a mnemonic in Latin languages (and we could even make it localizable if we wanted).

We should also have a shortcut key — probably with a modifier — for adding to the currently selected target. Not sure what, though — don't want to use anything particularly valuable if it's just for this. One possibility is to reuse Cmd-Shift-S ("Save to Zotero") for this. Right now in Firefox that only works when the Zotero pane doesn't have focus — otherwise it actually apparently throws an error (and it doesn't exist in Standalone). Since this is a very similar operation, it might make sense to allow that to work when a feed item is selected, unless we think that's too confusing. (In any case, see `getCaptureTooltip` in browser.js for an example of showing platform-specific modifier keys in the tooltip.)

adomasven:
> We should add the shortcut key in parentheses to the Mark as Read/Unread tooltip, but " (`)" is kind of impossible to decipher, so I'm somewhat inclined to switch to "M" (which is what Thunderbird uses), which would be much easier to see. It also has the advantage of there being a mnemonic in Latin languages (and we could even make it localizable if we wanted).

Setting the shortcut to 'M' breaks being able to type-to-search in the item tree view. We could add a modifier to it although "(`)" does work really well, even if hard to decipher.

dstillman:
Need icons on the menus as well — right now showing only for menuitems.

adomasven:
> Need icons on the menus as well — right now showing only for menuitems.

I think it's good UX cue to indicate with the icon what's clickable and what's a menu. A bit arbitrary, but makes sense, I've seen it done in other software.

dstillman:
No, it's just ugly. There's already a arrow to indicate what's a menu. See, e.g., bookmarks menu in Firefox — it would look (equivalently) awful if the folders didn't have icons.

test/tests/itemPaneTest.js
```diff
@@ -69,4 +69,20 @@ describe("Item pane", function () {
...
+			it("Updates label when state of an item changes", function* () {
+				let feed = yield createFeed();
+				yield selectLibrary(win, feed.libraryID);
+				let item = yield createDataObject('feedItem', {libraryID: feed.libraryID});
+				yield itemsView.selectItem(item.id);
+				let button = doc.getElementById('zotero-feed-item-toggleRead-button');
+				
+				assert.equal(button.getAttribute('label'), 'Mark As Read');
+				yield item.toggleRead(true);
+				assert.equal(button.getAttribute('label'), 'Mark As Unread');
```
dstillman: Need to use the localized strings.

chrome/skin/default/zotero/itemPane.css
```diff
@@ -42,13 +42,13 @@
...
+#zotero-item-pane-top-buttons-trash {
 	-moz-appearance: toolbar;
 	-moz-box-pack: center;
 	min-height: 3em;
 }
 
-#zotero-item-top-buttons-holder > button {
+#zotero-item-pane-top-buttons-trash > button {
 	-moz-box-flex: 1
 }
```
dstillman: The feed buttons need these styles too — should use a class that applies to both.

adomasven: If we're going to add a class that applies to both of them, we may as well just use the `flex='1'` attribute (which is what I did for feed buttons).

dstillman: Doesn't really matter — this way it gets set in one place.

adomasven:
Updates:

- Shortcut tooltips for buttons. Kept (`) for marking as read and it looks okay on Linux, not sure about OSX. Added CMD+SHIFT+S/CTRL+SHIFT+S for saving feed items.
- The menu building function is now in `utilities_internal.js`.
- Icons on all menu items.

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -806,6 +806,80 @@ Zotero.Utilities.Internal = {
...
+	 * If has no children, returns a <menuitem> element, otherwise <menu>.
+	 * 
+	 * @param {Library/Collection} libraryOrCollection
+	 * @param {function} clickAction function to execute on clicking the menuitem.
+	 * 		Receives the event and libraryOrCollection for given item.
+	 * 
+	 * @return {Node<menuitem>/Node<menu>}
+	 */
+	createMenuForTarget: function(libraryOrCollection, clickAction) {
+		var win = Services.wm.getMostRecentWindow("navigator:browser");
```
dstillman: Should just pass the document here — don't want to use windows. (And seems like the createElementNS came back below, which shouldn't be necessary.)

dstillman: (Pass the document or just pass the element to append to and use `elem.ownerDocument`)

chrome/content/zotero/itemPane.js
```diff
@@ -44,6 +45,23 @@ var ZoteroItemPane = new function() {
...
+		var addToLibraryButton = document.getElementById('zotero-feed-item-addTo-button');
+		key = Zotero.Keys.getKeyForCommand('saveToZotero');
+		if (key) {
+			// Add RLE mark in RTL mode to make shortcut render the right way
+			var tooltip = (Zotero.rtl ? ' \u202B' : '') + '('
+			+ (Zotero.isMac ? '⇧⌘' : Zotero.getString('general.keys.ctrlShift'))
+			+ key
+			+ ')';
+		}	
+		addToLibraryButton.setAttribute('tooltiptext', tooltip);
```
dstillman: The parentheses go after the regular tooltip (as in the existing "Zotero"/"Save to Zotero" buttons and the Firefox toolbar buttons).

dstillman:
So for mail I feel extremely strongly that there should be a mark-as-read timeout of a few seconds (and hate clients that don't allow that), but for feeds I'm much more open to immediate marking, since, unlike with mail (ideally), you're much less likely to care about the majority of entries and can quickly navigate through, marking as read as you go, without worrying about the time delay. I think with the delay we're back to a situation like we had with translation, where you're compelled to wait for something to finish (though at least here you can trigger it sooner).

The other advantage to immediate marking is that it's kind of distracting to have the "Mark as Read" button change each time a few seconds into your trying to read the right-hand pane, and if we mark immediately we can display it immediately as "Mark as Unread". (Probably need to special-case that, since otherwise there would be a slight delay for the async DB write.)

adomasven:
Updated with nits and marking read on select. I agree it is nicer this way.

dstillman:
Looks like a test might need to be updated:

>     Feed buttons
      Mark as Read/Unread
      ✖ [FAIL] Updates label when state of an item changes
        AssertionError: expected 'Mark As Unread' to equal 'Mark As Read' at
          @resource://zotero-unit-tests/itemPaneTest.js:82:1

chrome/content/zotero/itemPane.js
```diff
@@ -185,6 +188,90 @@ var ZoteroItemPane = new function() {
...
+	};
+	
+
+	this.setTranslationTarget = function(translationTarget) {
+		_translationTarget = translationTarget;
+		ZoteroItemPane.setTranslateButton()
+	};
+	
+		
+	this.setToggleReadLabel = function() {
```
dstillman: setToggleReadLabel -> toggleReadLabel

chrome/locale/en-US/zotero/preferences.dtd
```diff
@@ -140,7 +140,8 @@
 <!ENTITY zotero.preferences.keys.quicksearch				"Quick Search">
 <!ENTITY zotero.preferences.keys.newItem					"Create a New Item">
 <!ENTITY zotero.preferences.keys.newNote					"Create a New Note">
-<!ENTITY zotero.preferences.keys.toggleRead					"Mark Item Read/Unread">
+<!ENTITY zotero.preferences.keys.toggleAllRead				"Mark All Items As Read/Unread">
+<!ENTITY zotero.preferences.keys.toggleRead					"Mark An Item As Read/Unread">
```
dstillman: This is going to be confusing, since it's not clear that it's about feed items (and some people want read/unread functionality elsewhere, but we're not really interested in offering that). So probably should be "Mark All Feed Items as Read/Unread" and "Mark Feed Item as Read/Unread".

chrome/content/zotero/xpcom/itemTreeView.js
```diff
@@ -664,6 +656,10 @@ Zotero.ItemTreeView.prototype.notify = Zotero.Promise.coroutine(function* (actio
 			sort = true;
 		}
 		
+		if (collectionTreeRow.isFeed()) {
+			this._ownerDocument.defaultView.ZoteroItemPane.setToggleReadLabel();
+		}
+
```
dstillman: This actually isn't quite right — as long as the full Zotero for Firefox exists, it's possible to have multiple windows open (and we could conceivably allow that in Standalone at some point). Also possible it would fail if in feed pane with no items selected (again for multiple windows). Should check the currently selected row (if there's only one) against the modified ids to know whether to toggle.

(Also the `invalidateRow()` thing, which it seems like you had before above but removed.)

dstillman:
We should store the state of the target button in a pref, in the same format as used in the `lastViewedFolder` pref and used by `collectionTreeView::selectByID()` (though we only need to worry about "L" and "C"). If target not found when loading the pane, revert to My Library.

adomasven:
> We should store the state of the target button in a pref, in the same format as used by collectionTreeView::selectByID() (though we only need to worry about "L" and "C" — e.g., "L1" or "C23"). If target not found when loading the pane, revert to My Library.

Done.

dstillman:
I'm getting two different errors now:

Selecting items in a feed:

```
TypeError: this._itemRowMap is undefined

    Zotero.ItemTreeView.prototype.notify<@chrome://zotero/content/xpcom/itemTreeView.js:480:4
```

Selecting views in the left pane:

```
TypeError: Zotero.Prefs.get(...) is undefined

    Zotero.ItemTreeView.prototype.getSortFields@chrome://zotero/content/xpcom/itemTreeView.js:2192:24
    Zotero.ItemTreeView.prototype.sort@chrome://zotero/content/xpcom/itemTreeView.js:1374:19
    Zotero.ItemTreeView.prototype.setTree<@chrome://zotero/content/xpcom/itemTreeView.js:233:3
```

adomasven:
Amended

defaults/preferences/zotero.js
```diff
@@ -55,6 +55,7 @@ pref("extensions.zotero.groups.copyTags", true);
 pref("extensions.zotero.feeds.sortAscending", false);
 pref("extensions.zotero.feeds.defaultTTL", 1);
 pref("extensions.zotero.feeds.defaultCleanupAfter", 2);
+pref("extensions.zotero.feeds.translationTarget", false);
```
dstillman: Let's add a "last" in there for consistency: `extensions.zotero.feeds.lastTranslationTarget`

Issue #935, opened 2016-03-31
RTF markup appears in Edit Bibliography preview pane


dstillman:
RTF markup is shown in the Edit Bibliography preview pane for entries with rich-text formatting.

According to @simonster, we try to convert the RTF to HTML (in styled-textbox.xml), but the RTF generated by citeproc-js changed at some point, so you can end up with, e.g., `{\i{}Title}. n.d.` in the preview window.

Possible fixes:

1. Fix the replacement rules
2. Insert HTML in the plugins instead
3. Insert the bibliography as separate fields, which would allow us to remove the preview pane altogether, since edits could then be made in the document directly (as they already can be for citations). This would also make hyperlinks from citations to the bibliography work.

dstillman:
Replacement rules are [here](https://github.com/zotero/zotero/blob/be4901078396f43683973412f870bc0617c33584/chrome/content/zotero/bindings/styled-textbox.xml#L52) — not sure exactly what changed in citeproc-js to know how hard it would be to update these.

fbennett:
IIRC the reason for that change to the RTF markup was to get formatting to revert correctly in nested markup. I'll take a look. If we can get the existing function to round-trip the markup for now, that will open up time for you to work toward (3) at a deliberate pace.

PR #937, opened 2016-04-01
HTML/RTF and RTF/HTML conversion for styled textbox

fbennett:
This addresses #935. I haven't tested it extensively, but it round-trips quotes and high-bit characters okay. If balanced tags come up that are missed in the parse they can be added to the spec object.


dstillman:
That's great — thanks so much!

fbennett:
My pleasure!

PR #938, opened 2016-04-01
Closes #adomasven/zotero/10. Adds a menu option to add feeds found on the current page.

adomasven:
Abuses the same API firefox uses to build its feed menus.

### DIFF
--- a/chrome/content/zotero/zoteroPane.xul
+++ b/chrome/content/zotero/zoteroPane.xul
@@ -36,8 +36,7 @@
 
 <overlay id="zotero"
 	xmlns="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul">
-
-	<script src="feedSettings.js"/>
+	
 	<script src="include.js"/>
 	<script src="zoteroPane.js" type="application/javascript;version=1.8"/>
 	<script src="fileInterface.js"/>
@@ -107,13 +106,13 @@
 					<toolbarbutton id="zotero-tb-library-add-menu" class="zotero-tb-button" tooltiptext="&zotero.toolbar.newLibrary.label;" type="menu">
 						<menupopup id="zotero-tb-library-add-popup">
 							<menuitem id="zotero-tb-group-add" label="&zotero.toolbar.newGroup;" oncommand="ZoteroPane_Local.newGroup()"/>
-							<menu label="&zotero.toolbar.feeds.new;">
-								<menupopup id="zotero-tb-feed-add-more">
+							<menu id="zotero-tb-feed-add-menu" label="&zotero.toolbar.feeds.new;">
+								<menupopup>
 									<menuitem id="zotero-tb-feed-add-fromURL" label="&zotero.toolbar.feeds.new.fromURL;" 
 										command="cmd_zotero_newFeed_fromURL"/>
-									<menuitem id="zotero-tb-feed-add-fromPage" label="&zotero.toolbar.feeds.new.fromPage.menu;"
+									<menuitem id="zotero-tb-feed-add-fromPage" label="&zotero.toolbar.feeds.new.fromPage;"
 										oncommand="ZoteroPane_Local.newFeedFromPage()" observes="singleFeedMenuitemState"/>
-									<menu id="zotero-tb-feed-add-fromPage-menu" label="&zotero.toolbar.feeds.new.fromPage.menu;"
+									<menu id="zotero-tb-feed-add-fromPage-menu" label="&zotero.toolbar.feeds.new.fromPage;"
 										observes="multipleFeedsMenuState">
 											<menupopup oncommand="ZoteroPane_Local.newFeedFromPage(event)" 
 												onpopupshowing="FeedHandler.buildFeedList(event.target)"/>
@@ -258,7 +257,6 @@
 					<menuitem class="menuitem-iconic zotero-menuitem-new-collection" label="&zotero.toolbar.newCollection.label;" command="cmd_zotero_newCollection"/>
 					<menuitem class="menuitem-iconic zotero-menuitem-new-saved-search" label="&zotero.toolbar.newSavedSearch.label;" command="cmd_zotero_newSavedSearch"/>
 					<menuitem class="menuitem-iconic zotero-menuitem-new-collection" label="&zotero.toolbar.newSubcollection.label;" oncommand="ZoteroPane_Local.newCollection(ZoteroPane_Local.getSelectedCollection().key)"/>
-					<menuitem class="menuitem-iconic zotero-menuitem-new-feed" label="&zotero.toolbar.feeds.new;" command="cmd_zotero_newFeed_fromURL"/>
 					<menuitem class="menuitem-iconic zotero-menuitem-refresh-feed" label="&zotero.toolbar.feeds.refresh;" oncommand="ZoteroPane_Local.refreshFeed();"/>
 					<menuseparator/>
 					<menuitem class="menuitem-iconic zotero-menuitem-show-duplicates" label="&zotero.toolbar.duplicate.label;" oncommand="ZoteroPane_Local.setVirtual(ZoteroPane_Local.getSelectedLibraryID(), 'duplicates', true)"/>
--- a/chrome/content/zotero/zoteroPane.xul
+++ b/chrome/content/zotero/zoteroPane.xul
@@ -36,7 +36,8 @@
 
 <overlay id="zotero"
 	xmlns="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul">
-	
+
+	<script src="feedSettings.js"/>
 	<script src="include.js"/>
 	<script src="zoteroPane.js" type="application/javascript;version=1.8"/>
 	<script src="fileInterface.js"/>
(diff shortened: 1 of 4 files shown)


chrome/content/zotero/zoteroPane.xul
```diff
@@ -36,7 +36,8 @@
 
 <overlay id="zotero"
 	xmlns="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul">
-	
+
+	<script src="feedSettings.js"/>
```
dstillman: This should be at the bottom of the script list, if anything, but why does this need to be in this file? Isn't it only used in feedSettings.xul?

adomasven: Whoops, forgot to remove it.

chrome/content/zotero/zoteroPane.xul
```diff
@@ -245,14 +258,14 @@
 					<menuitem class="menuitem-iconic zotero-menuitem-new-collection" label="&zotero.toolbar.newCollection.label;" command="cmd_zotero_newCollection"/>
 					<menuitem class="menuitem-iconic zotero-menuitem-new-saved-search" label="&zotero.toolbar.newSavedSearch.label;" command="cmd_zotero_newSavedSearch"/>
 					<menuitem class="menuitem-iconic zotero-menuitem-new-collection" label="&zotero.toolbar.newSubcollection.label;" oncommand="ZoteroPane_Local.newCollection(ZoteroPane_Local.getSelectedCollection().key)"/>
-					<menuitem class="menuitem-iconic zotero-menuitem-new-feed" label="&zotero.toolbar.newFeed.label;" command="cmd_zotero_newFeed"/>
-					<menuitem class="menuitem-iconic zotero-menuitem-refresh-feed" label="&zotero.toolbar.refreshFeed.label;" oncommand="ZoteroPane_Local.refreshFeed();"/>
+					<menuitem class="menuitem-iconic zotero-menuitem-new-feed" label="&zotero.toolbar.feeds.new;" command="cmd_zotero_newFeed_fromURL"/>
```
dstillman: This menuitem isn't actually used, is it? I don't think it appears anywhere (because the Feeds header isn't selectable), and the `newFeed` position isn't referenced in buildCollectionContextMenu.

adomasven: Correct. This has carried over from the initial branch by Aurimas. Should it stay or should it go? I say go.

dstillman: Go.

chrome/locale/en-US/zotero/zotero.dtd
```diff
@@ -128,6 +125,13 @@
 <!ENTITY zotero.toolbar.openURL.label					"Locate">
 <!ENTITY zotero.toolbar.openURL.tooltip				"Find through your local library">
 
+<!ENTITY zotero.toolbar.feeds.new					"New Feed…">
+<!ENTITY zotero.toolbar.feeds.new.fromURL			"From URL…">
+<!ENTITY zotero.toolbar.feeds.new.fromPage			"From Page">
+<!ENTITY zotero.toolbar.feeds.new.fromPage.menu		"From Page…">
```
dstillman: `zotero.toolbar.feeds.new.fromPage` is unused, I think, so can just change `fromPage.menu` to `fromPage`.

chrome/content/zotero/zoteroPane.xul
```diff
@@ -106,7 +107,19 @@
 					<toolbarbutton id="zotero-tb-library-add-menu" class="zotero-tb-button" tooltiptext="&zotero.toolbar.newLibrary.label;" type="menu">
 						<menupopup id="zotero-tb-library-add-popup">
 							<menuitem id="zotero-tb-group-add" label="&zotero.toolbar.newGroup;" oncommand="ZoteroPane_Local.newGroup()"/>
-							<menuitem id="zotero-tb-feed-add" label="&zotero.toolbar.newFeed.label;" command="cmd_zotero_newFeed"/>
+							<menu label="&zotero.toolbar.feeds.new;">
+								<menupopup id="zotero-tb-feed-add-more">
```
dstillman: Should put an id — `zotero-tb-feed-add-menu` — on the menu instead. (Will become clear why shortly.)

dstillman:
So this works…really quite well.

And now the bad news. The "From Page…" stuff won't work at all in Standalone, so we probably need both the original menuitem and the menu, with their visibility set in `ZoteroPane.init()` depending on `Zotero.isStandalone`.

Not sure what happens to the `observes` lines if no corresponding broadcaster. Probably nothing, and we can just leave them hidden?

adomasven:
> So this works…really quite well.

I know, right.

> Not sure what happens to the observes lines if no corresponding broadcaster. Probably nothing, and we can just leave them hidden?

Yup, should be fine.

adomasven:
Amended

chrome/content/zotero/zoteroPane.xul
```diff
@@ -245,14 +257,13 @@
 					<menuitem class="menuitem-iconic zotero-menuitem-new-collection" label="&zotero.toolbar.newCollection.label;" command="cmd_zotero_newCollection"/>
 					<menuitem class="menuitem-iconic zotero-menuitem-new-saved-search" label="&zotero.toolbar.newSavedSearch.label;" command="cmd_zotero_newSavedSearch"/>
 					<menuitem class="menuitem-iconic zotero-menuitem-new-collection" label="&zotero.toolbar.newSubcollection.label;" oncommand="ZoteroPane_Local.newCollection(ZoteroPane_Local.getSelectedCollection().key)"/>
-					<menuitem class="menuitem-iconic zotero-menuitem-new-feed" label="&zotero.toolbar.newFeed.label;" command="cmd_zotero_newFeed"/>
```
dstillman: Have to remove the entry in `buildCollectionContextMenu` or this breaks the menu.

chrome/content/zotero/zoteroPane.xul
```diff
@@ -106,7 +106,19 @@
 					<toolbarbutton id="zotero-tb-library-add-menu" class="zotero-tb-button" tooltiptext="&zotero.toolbar.newLibrary.label;" type="menu">
 						<menupopup id="zotero-tb-library-add-popup">
 							<menuitem id="zotero-tb-group-add" label="&zotero.toolbar.newGroup;" oncommand="ZoteroPane_Local.newGroup()"/>
-							<menuitem id="zotero-tb-feed-add" label="&zotero.toolbar.newFeed.label;" command="cmd_zotero_newFeed"/>
+							<menu id="zotero-tb-feed-add-menu" label="&zotero.toolbar.feeds.new;">
```
dstillman: We want both the menuitem and the menu here. In Standalone, there should only be the "New Feed…" menuitem, with ("From Page…") implied since that's the only option. (That's what I meant about having the id on the menu, so it can be used for hiding in Standalone.)

adomasven: Ah. Well I'm working on a feed-discovery branch which adds another menuitem there and kind of have it at the back of my head, so I didn't even think about it this way. I could hide it for this branch, but probably not much point?

dstillman: What's the entry in the feed-discovery branch? I'd think we'd want to have a single interface for adding, with an option to enter a URL manually from there.

dstillman: (In other words, optimize for the probably majority use case of people who don't know what a feed URL is or where to find one, but might find this feature useful if given an easy way to add feeds.)

adomasven: Just says "discover" and opens up a dialog which has the discovery interface.

dstillman: Yeah, particularly since the "From Page" option will be going away in Standalone (even if you might be able to add from the connector side), I think we want a single New Feed… option, with an advanced way to add a feed URL from there.

adomasven: I went that route because combining the two requires some kind of decent UI solution and I just wanted to focus on the other stuff

dstillman: Yeah, don't worry too much about the UI. We can deal with that separately.

dstillman: For now, can just stick a button at the bottom of the window in the discovery branch that opens the other URL window. We'll make it look nice after.

dstillman: Oh, but then the question is what we'll do in Firefox on the discovery branch, since "From URL…" won't work.

dstillman: I think the options are 1) no submenu, just "New Feed…" and "New Feed from Page…" top-level items, with the latter not appearing in Standalone, or 2) separate "Discover…", "From URL…", and "From Page…" options in the submenu, with the latter not appearing in Standalone.

I guess, in the interest of expediency, we can go with option 2.

adomasven: So no changes here then, right?

dstillman: Yeah

stakats:
Agree that we should think about adding this to the connector side of things. Could fit under the dropdown translator icon, which already provides multiple "save" options.

dstillman:
I fixed the `buildCollectionContextMenu` thing and merged this in 0a63c7665a46d984bfa979990d2c9025f864010c.

PR #946, opened 2016-04-12
Bug fix progressWindow.js: nArcs undefined

jlegewie:
`nArcs` is undefined in `this.ItemProgress.prototype.setProgress()`. This has been the case for a long time but never came up because the progress indicator isn't used in Zotero, I believe. I think it's a nice UI element though and could be used for non-disruptive progress indicators.

### DIFF
--- a/chrome/content/zotero/xpcom/progressWindow.js
+++ b/chrome/content/zotero/xpcom/progressWindow.js
@@ -332,6 +332,7 @@ Zotero.ProgressWindow = function(_window = null) {
 	this.ItemProgress.prototype.setProgress = _deferUntilWindowLoad(function(percent) {
 		if(percent != 0 && percent != 100) {
 			// Indication of partial progress, so we will use the circular indicator
+			var nArcs = 20;			
 			this._image.style.backgroundImage = "url('chrome://zotero/skin/progress_arcs.png')";
 			this._image.style.backgroundPosition = "-"+(Math.round(percent/100*nArcs)*16)+"px 0";
 			this._hbox.style.opacity = percent/200+.5;


jlegewie:
Not sure why this fails. The commit shouldn't change anything...

dstillman:
It does seem like this never would've worked, even in https://github.com/zotero/zotero/commit/a750203f4fa3a9ade79cc57dde9d96e7f147c899. Do you have code handy to demo this?

dstillman:
Oh, nm, I see it in the other ticket.

dstillman:
It does seem like this is at least supposed to be hooked up via `attachmentProgressHandler`. Not sure why that's not working.

jlegewie:
Here is some code (all ran from Scratchpad in browser environment).

Without my fix, the `item_progress.setProgress()` call leads to an error because `nArcs is not defined`:
```
var progress_win = new Zotero.ProgressWindow();
progress_win.changeHeadline('Headline!');
var icon = 'chrome://zotero/skin/treeitem-journalArticle.png';
var item_progress = new progress_win.ItemProgress(icon, "Exporting to file 'blabla.bib'");
progress_win.show();
item_progress.setProgress(10);
```

Here is code how it is supposed to work (using `Zotero.ZotFile.ProgressWindow`, which duplicated `Zotero.ProgressWindow` with the fix and some unrelated changes).
```
var progress_win = new Zotero.ZotFile.ProgressWindow();
progress_win.changeHeadline('Headline!');
var icon = 'chrome://zotero/skin/treeitem-journalArticle.png';
var item_progress = new progress_win.ItemProgress(icon, "Exporting to file 'blabla.bib'");
progress_win.show();
var update_progress = function(i) {
  item_progress.setProgress(i);
  if (i < 100) setTimeout(() => update_progress(i + 1), 10);
}
setTimeout(() => update_progress(1), 100);
```

jlegewie:
`nArcs` refers to the number of steps for the progress indicator as provided by `chrome://zotero/skin/progress_arcs.png`. The variable is used two lines after I define it.

Exporting items is a good example where this could be used as a non-disruptive progress indicator.

dstillman:
Right, I know — thanks. I was just saying that `nArcs` was never defined, even in the original code where it was first used.

Thanks for the example code.

PR #948, opened 2016-04-14
Add functionality to retrieve pdfs from urls using the translator framework

jlegewie:
This pull requests adds new functionality based on the translator framework to retrieve PDFs from urls. I think this is useful considering that people often add items to their library without access to their library (e.g. from home). This is a convenient way to try adding the item again. I also think it can be extended to try and find pdfs in multiple sources (URI, revolved DOI, google scholar).
#### Summary of changes
- Added API method `Zotero.Translators.getPdfFromURL`. The function takes an url and a parent zotero item and then loads the url in a hidden browser, looks for fitting translators, uses the translator to retrieve pdf attachments (not item metadata), and finally adds the item to the parent
- Added API method `Zotero.Browser.createHiddenBrowserWithURL` to create a hidden browser, start loading a url and return a promise.
- UI change: added right click menu to the URL field in the right-pane item view with the option to "Retrieve associated PDFs" (see screenshot).
#### Remaining work
- You might not like the naming etc of the new API methods.
- The right-click menu for the URL field is not the only way in which this could be used. Here are some other UI ideas: (a) add a menu item "Retrieve PDFs" when users right-click on items in the middle pane. That would make it much more accessible and it could be extended to try different sources (URI, revolved DOI, google scholar). (b) Add "Retrieve associated PDFs"  to the DOI field in the right pane as well for the `http://doi.org/...` url.
- Currently "Retrieve associated PDFs" always appears when their is a valid URL. This could be more restrictive and first check for valid translators with `Zotero.Translators.getWebTranslatorsForLocation(url)`
- No visual feedback for the user. 
- Error handling is probably not the best.

screen shot 2016-04-14 at 11 11 23 am


adam3smith:
(@jlegewie -- haven't tested this yes, but this is amazing. It's been on my wishlist forever; thanks for working on it.)

dstillman:
This is awesome — thanks! I'll look at this further after we get a 5.0 beta out. Just for some quick feedback:

1) We'd definitely want this in the item context menu rather than via right-click anywhere in the item pane.

2) `getPdfFromURL` -> `getPDFFromURL`, and I think more appropriate in `Zotero.Utilities.Internal`.

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -2752,6 +2752,7 @@ Zotero.Browser = new function() {
 	
 	this.createHiddenBrowser = createHiddenBrowser;
 	this.deleteHiddenBrowser = deleteHiddenBrowser;
+	this.createHiddenBrowserWithURL = createHiddenBrowserWithURL;
```
dstillman: We don't do this anymore for methods. `this.createHiddenBrowserWithURL = function () {` below.

chrome/content/zotero/xpcom/translation/translators.js
```diff
@@ -468,4 +468,36 @@ Zotero.Translators = new function() {
 			[fileName, JSON.stringify(metadataJSON), lastModifiedTime]
 		);
 	}
+
+	/**
+	 * [* description]
+	 * @param {string}   url Url used to retrieve PDF
+	 * @param {zitem} parent Zotero item to which the pdf is attached
```
dstillman: `{zitem}` -> `{Zotero.Item}`

chrome/content/zotero/xpcom/translation/translators.js
```diff
@@ -468,4 +468,36 @@ Zotero.Translators = new function() {
...
+		var atts = zitem[0].attachments.filter(att => att.mimeType == "application/pdf");
+		if (atts.length == 0) throw new Error("No attachments on page");
+		var itemSaver = new Zotero.Translate.ItemSaver({
+			"libraryID": parent.libraryID,
+			"attachmentMode": Zotero.Translate.ItemSaver.ATTACHMENT_MODE_DOWNLOAD,
+			"forceTagType": 1,
+			"cookieSandbox": new Zotero.CookieSandbox(browser, url),
+			"baseURI": url
+		});
+		return itemSaver._saveAttachment(atts[0], parent.getID(), (translator, progress) => {return;});
```
dstillman: Just `() => {}` if you're just passing a no-op function.

dstillman: And `parent.id` — `Zotero.Item.prototype.getID()` is deprecated.

jlegewie:
Okay, I made all the suggested changes except for the UI. Note that the current implementation **is not** for "right-click anywhere in the item pane". The menu only appears when you right click on the URL or the DOI and not when you click anywhere else (similar  to "view online" and "copy as URL" for DOI). The screenshot is misleading in that sense because the short title is highlighted. But I agree that it would fit nicely in the item context menu and maybe try both the DOI url and the url field. I didn't do it now because I wasn't sure where exactly to place it etc so we can get back to that when you get a chance to take a closer look.

**Changes**

- `Zotero.Translators.getPdfFromURL()` -> `Zotero.Utilities.Internal.getPDFFromURL()`
- Added "Retrieve associated PDFs" to the DOI right click menu (below "View Online" and "Copy as URL")
- `this.createHiddenBrowserWithURL` (note that the other methods in `zotero.js` defined as `this.createHiddenBrowser = createHiddenBrowser()`)
- Small adjustments to (`parent.id`, `() => {}`, documentation etc)

dstillman:
> Note that the current implementation is not for "right-click anywhere in the item pane".

Right, I just meant I don't think it makes sense to put this anywhere in the item pane, because from the user's perspective this doesn't need to have anything to do with the URL or DOI — all they want is to get a PDF from the item.

>  I didn't do it now because I wasn't sure where exactly to place it etc so we can get back to that when you get a chance to take a closer look.

Sounds good. (On a technical level, in case you want to take a look, the relevant code is in `buildItemContextMenu()` in zoteroPane.js, with a corresponding menuitem going in `zotero-itemmenu` in zoteroPane.xul.)

KarlHegbloom:
I've noticed that Google Chrome on Android produces nicer PDFs when I print
to PDF then either Firefox or Chrome on Linux. When I use the desktop
browsers to produce a PDF from Google Scholar case, and then I highlight
and make notes in it, and then import highlights using Zotfile, it gives
really weird results with spaces inside words and all kinds of trouble. But
if I print to file from Google Chrome on Android Kama it produces a perfect
PDF and when I import the highlights they come out just the way they're
supposed to.

It occurs to me that it would be nice to have Zotero itself able to make a
PDF from a web page because of the ability of PDF to allow  makeing
annotations and things. But it's important that it make a reasonably
nice-looking PDFs the way that Google Chrome on Android does.

dstillman:
> I've noticed that Google Chrome on Android produces nicer PDFs when I print to PDF then either Firefox or Chrome on Linux. When I use the desktop browsers to produce a PDF from Google Scholar case, and then I highlight and make notes in it, and then import highlights using Zotfile, it gives really weird results with spaces inside words and all kinds of trouble. But if I print to file from Google Chrome on Android Kama it produces a perfect PDF and when I import the highlights they come out just the way they're supposed to. It occurs to me that it would be nice to have Zotero itself able to make a PDF from a web page because of the ability of PDF to allow makeing annotations and things. But it's important that it make a reasonably nice-looking PDFs the way that Google Chrome on Android does.

@KarlHegbloom, please post feature requests to the forums. (Your request also doesn't really have anything to do with this thread, which is about downloading actual PDFs for existing items.)

bwiernik:
When adding other PDF sources, it might be worth considering adding the open access DOI resolver (http://doai.io/) to the list.

jlegewie:
Okay, I have added a menu item `Retrieve associated PDF` and added a progress window for user feedback. Currently, the option is grey if there is no DOI and no URL and it only appear if the user selects a single regular item. The menu items calls `ZoteroPane_Local.retrievePDFSelectedItem()`, which first first the DOI url and if that doesn't work the URL. Would be possible add google scholar for a third attempt.

- Confirm location of menu item. Currently it's under "Add note" and "add attachment".
- There is no timeout connected with `Zotero.Browser.createHiddenBrowserWithURL`. I had a super slow loading page and the process took forever (the page also didn't load properly in Chrome). So maybe it would be a good idea to time out `loadURI` but I don't think there is direct way to do it.
- The option `Retrieve associated PDF` still appears when the user clicks on "URL" or "DOI". Just wanted confirm before I remove it. I personally think that it could stay but happy to remove.
- You might want to look at the error handling

jlegewie:
The 5.0 beta has been out for a while and I just wanted to check in about this. Still interest? If there is, I would update the pull request and address any comments from you guys.

KarlHegbloom:
I often make print to PDF of web pages,  especially things like Google
Scholar cases,  then attach them and use zotfile to put them on the tablet
for annotation... Chrome on Android makes perfect PDF's but neither firefox
nor Chrome on the Linux desktop make good PDF's. One of them make PDF's
with extraneous spaces in the pasted text or extracted annotations, for
example. I have seen a tool that uses the Qt webkit to render a PDF but it
requires an X server or xvfb... Anyway, it would be cool to have a web page
print to PDF then attach feature but it must make good PDF or it's
practically useless.

adam3smith:
@KarlHegbloom -- you misread what this pull request does. It's not at all related to converting webpages to PDF, but rather will fetch PDFs via Zotero translators where the PDF already exists, as it does for most academic articles. There are several threads on a print to PDF feature on the Zotero forums where you can follow up.

kevflanagan:
Hey, my laptop was stolen with my research library on it. Luckily Zotero had the metadata online. I'm looking for a way to automatically pull pdfs from urls, doi etc listed in the metadata rather than go through them manually. Would this feature do the trick? How do I install it? Using zotero in the browser currently. Many Thanks

adomasven:
@kevflanagan Please create a new thread on [zotero forums](https://forums.zotero.org/) and you will certainly receive help. The github tracker is for development use only.

bwiernik:
Based on responses to forum questions, it seems that there is still definite interest in a 5.0-compatible version of this patch.

eparimbelli:
I think this is a great feature. Any idea on the timing/version this would be included for release?
I see the CI integration is failing but that could be fixed by someone with write access right?

dstillman:
@jlegewie: Sorry for the delay on this. I'd be happy to move forward on this. Are you still interested in working on this, or do you want us to take it from here?

jlegewie:
Feel free to take it from here. It could only get back to this around April, which I can do as well if nothing happens until then.

mehdifanai:
This is a very useful feature. It's implemented on Endnote for a long time. I wish Zotero had it as well. Thanks

swandle06:
Any idea when this will be available?  I'm transitioning from Endnote to Zotero, and this is at the top of the list for features I miss the most.

krishnakumarg1984:
I am waiting for this feature too. Jabref has had it for a long time. When will this be merged?

dstillman:
[An expanded version](https://github.com/zotero/zotero/commit/679a6d5cc7e6c87931ea3b484332dda168ad5b5d) of this functionality is now available in the latest <a href="https://www.zotero.org/support/dev_builds">Zotero beta</a>. (We'll also be applying some of it to saving from the connector, but that's not done yet.) Thanks again to @jlegewie for his work on this.

PR #978, opened 2016-04-27
Fix some for each ... in cases

zuphilip:
The `for-each-in` loops are deprecated and will cause some warnings in the browser console (and are only compatible by Gecko):
```
JavaScript 1.6's for-each-in loops are deprecated; consider using ES6 for-of instead
```

@dstillman I can continue here to look at more of these loop cases, if you agree.

### DIFF
--- a/chrome/content/zotero/preferences/preferences_sync.js
+++ b/chrome/content/zotero/preferences/preferences_sync.js
@@ -213,9 +213,9 @@ Zotero_Preferences.Sync = {
 	updateStorageSettingsGroups: function (enabled) {
 		var storageSettings = document.getElementById('storage-settings');
 		var menulists = storageSettings.getElementsByTagName('menulist');
-		for (let i=0; i<menulists.length; i++) {
-			if (menulists[i].className == 'storage-groups') {
-				menulists[i].disabled = !enabled;
+		for (let menulist of menulists) {
+			if (menulist.className == 'storage-groups') {
+				menulist.disabled = !enabled;
 			}
 		}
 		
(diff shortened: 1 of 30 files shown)


dstillman:
Thanks. While we're changing these, we should use `let` instead of `var`.

zuphilip:
I changed to use `let` and add more cases, which I could easily verify as loops in arrays.

Should I been worried because Travis failed now? I read in the Travis report "Error: timeout of 10000ms exceeded", and the tests on the other two branches did pass.

zuphilip:
Okay, Travis is happy this time. @dstillman Are the changes okay so far?

chrome/content/zotero/preferences/preferences_sync.js
```diff
@@ -213,9 +213,9 @@ Zotero_Preferences.Sync = {
 	updateStorageSettingsGroups: function (enabled) {
 		var storageSettings = document.getElementById('storage-settings');
 		var menulists = storageSettings.getElementsByTagName('menulist');
-		for each(var menulist in menulists) {
-			if (menulist.className == 'storage-groups') {
-				menulist.disabled = !enabled;
+		for (let i=0; i<menulists.length; i++) {
```
dstillman: NodeList is iterable in Firefox, so you should be able to use `for...of` here too.

zuphilip: Do you prefer `for...of` here? I thought that the loop here is more consistent to some other loops in the `zotero` code: https://github.com/zotero/zotero/search?utf8=%E2%9C%93&q=getElementsByTagName

dstillman: Yes, those mostly just predate `for...of` and iterable `NodeList`. `for...of` is better anywhere except possibly for some highly performance-sensitive operations.

PR #1006, opened 2016-05-23
Add a second layout and a button to switch layouts

LinuxMercedes:
My laptop has a tiny screen resolution (1024x768) and sometimes it gets hard to use Zotero since everything gets split into collumns smaller than IEEE's.
I've added a second layout that puts the Items Pane above the Item Pane so that I can read the information fields and write notes easier.
You can switch between the standard layout and this layout by clicking the 'S' button in the top right, next to the tab view toggle button.

Screenshot of layout

(As a bonus, you also get part of a clustering literature review!)

So far I haven't found any weird UI glitches but I would greatly appreciate some more testing.

I think the only thing that is left to do is to make a proper icon for it instead of just a cryptic 'S'.

### DIFF
--- a/install.rdf
+++ b/install.rdf
@@ -6,7 +6,7 @@
   
     <em:id>zotero@chnm.gmu.edu</em:id>
     <em:name>Zotero</em:name>
-    <em:version>4.0.29.9.SOURCE</em:version>
+    <em:version>4.0.29.10.SOURCE</em:version>
     <em:creator>Center for History and New Media<br/>George Mason University</em:creator>
     <em:contributor>Dan Cohen</em:contributor>
     <em:contributor>Sean Takats</em:contributor>
--- a/resource/config.js
+++ b/resource/config.js
@@ -15,7 +15,7 @@ var ZOTERO_CONFIG = {
 	BOOKMARKLET_ORIGIN: 'https://www.zotero.org',
 	HTTP_BOOKMARKLET_ORIGIN: 'http://www.zotero.org',
 	BOOKMARKLET_URL: 'https://www.zotero.org/bookmarklet/',
-	VERSION: '4.0.29.9.SOURCE'
+	VERSION: '4.0.29.10.SOURCE'
 };
 
 EXPORTED_SYMBOLS = ["ZOTERO_CONFIG"];
--- a/update.rdf
+++ b/update.rdf
@@ -7,7 +7,7 @@
     <RDF:Seq>
       <RDF:li>
         <RDF:Description>
-          <version>4.0.29.9.SOURCE</version>
+          <version>4.0.29.10.SOURCE</version>
           <targetApplication>
             <RDF:Description>
               <id>{ec8030f7-c20a-464f-9b0e-13a3a9e97384}</id>


dstillman:
This is pretty nice. We've actually been considering this sort of layout for narrower browser windows in the upcoming web library redesign, and supporting this in the client seems like a good idea.

Instead of a toolbar button, I think it'd make more sense for the switching to just be done via the General pane of the preferences (and, more importantly, from a menu in Standalone, but that can be done separately).

Beyond that, my only concern is any buttons in the item pane, which become somewhat comically stretched in this layout (e.g., "Edit in separate window" in the above screenshot). But we can try to figure something else out for that separately.

LinuxMercedes:
Thanks!

My reason for using a toolbar button to switch layout is that when using Zotero as a pop-up instead of on its own tab, the new layout can be pretty hard to use, so being able to switch back and forth quickly is nice. I suppose that it would be best to include a layout setting for both tab and non-tab views? 

I can throw in a standalone menu in this PR. I usually don't use standalone, so that didn't cross my mind, but I suppose I'll need to switch at some point soon. 

Those buttons can already get pretty wide as it is. Maybe it would make sense to have them pop onto the same row as the Related and Tags fields after some point?

dstillman:
5.0 will likely be Standalone-only, so pane and tab modes (the the corresponding pref and button for the latter) are going away anyway. So I don't think we'd put a switcher for this in the toolbar — we'd either actually use responsive layout like on a webpage (which would be a little weird in an app but we can talk about it) or put it in the menu and pref. Having this in the View menu and the prefs is pretty standard (e.g., email clients).

> I can throw in a standalone menu in this PR.

That'd be great, but if you're not set up to do that, feel free to leave that out. I don't know all the details, but [standalone.xul](https://github.com/zotero/zotero/blob/master/chrome/content/zotero/standalone/standalone.xul) would be the place to start. We don't have a View menu currently, but I'm sure we'll be able to find other uses for it.

LinuxMercedes:
Ah, okay, that makes sense. I don't think I'd want a responsive design, plus coding it to work well would be tricky. Since this feature is mostly here because I wanted it, I've got no problem adding a menu. 

What do you want to call the two layouts? I mentally refer to them as 'horizontal' and 'vertical', but my mental definitions of 'horizontal' and 'vertical' are backwards of what the `orient` attribute uses. 

Thunderbird calls this new layout 'classic' and the traditional Zotero layout 'vertical'.

rmzelle:
It'll only save a few pixels, but I also noticed that in the screenshot above, the spacing of nested collections is wider than on my Windows copy:

Screenshot above, crop:
image

Windows 7, crop:
image

LinuxMercedes:
@rmzelle I think that's just how Firefox on Linux renders treeviews. 

@dstillman I can't get zotero-standalone to build right now, so let's merge this PR whenever it is done and I'll do the View menu in standalone in a separate PR when I get it working.

LinuxMercedes:
That build failure looks spurious.

dstillman:
Oh, we need this against master, not 4.0. (Not sure, but I think you'll need to open a new PR.)

LinuxMercedes:
https://github.com/zotero/zotero/pull/1010

PR #1010, opened 2016-05-24
Add a second horizontal layout to Zotero for small screens

LinuxMercedes:
See https://github.com/zotero/zotero/pull/1006 for more discussion and a screenshot!

### DIFF
--- a/chrome/content/zotero/preferences/preferences_general.xul
+++ b/chrome/content/zotero/preferences/preferences_general.xul
@@ -32,6 +32,7 @@
 		
 		<preferences id="zotero-prefpane-general-preferences">
 			<preference id="pref-fontSize" name="extensions.zotero.fontSize" type="string"/>
+			<preference id="pref-layout" name="extensions.zotero.layout" type="string" onchange="Zotero.getActiveZoteroPane().updateLayout();"/>
 			<preference id="pref-automaticScraperUpdates" name="extensions.zotero.automaticScraperUpdates" type="bool"/>
 			<preference id="pref-reportTranslationFailure" name="extensions.zotero.reportTranslationFailure" type="bool"/>
 			<preference id="pref-automaticSnapshots" name="extensions.zotero.automaticSnapshots" type="bool"/>
@@ -56,6 +57,16 @@
 				</columns>
 			
 				<rows id="zotero-prefpane-general-rows">
+					<row>
+						<hbox align="center">
+							<label value="&zotero.preferences.layout;" control="layout"/>
+						</hbox>
+						<radiogroup id="layout" orient="horizontal" align="center" preference="pref-layout">
+							<radio id="horizontal" label="&zotero.preferences.layout.horizontal;" value="horizontal"/>
+							<radio id="vertical" label="&zotero.preferences.layout.vertical;" value="vertical"/>
+						</radiogroup>
+					</row>
+
 					<row>
 						<hbox align="center">
 							<label value="&zotero.preferences.fontSize;" control="fontSize"/>
--- a/chrome/content/zotero/preferences/preferences_general.xul
+++ b/chrome/content/zotero/preferences/preferences_general.xul
@@ -62,8 +62,8 @@
 							<label value="&zotero.preferences.layout;" control="layout"/>
 						</hbox>
 						<radiogroup id="layout" orient="horizontal" align="center" preference="pref-layout">
-							<radio id="horizontal" label="&zotero.preferences.layout.standard;" value="Standard"/>
-							<radio id="vertical" label="&zotero.preferences.layout.stacked;" value="Stacked"/>
+							<radio id="horizontal" label="&zotero.preferences.layout.standard;" value="standard"/>
+							<radio id="vertical" label="&zotero.preferences.layout.stacked;" value="stacked"/>
 						</radiogroup>
 					</row>
 
--- a/chrome/content/zotero/preferences/preferences_general.xul
+++ b/chrome/content/zotero/preferences/preferences_general.xul
@@ -32,7 +32,7 @@
 		
 		<preferences id="zotero-prefpane-general-preferences">
 			<preference id="pref-fontSize" name="extensions.zotero.fontSize" type="string"/>
-			<preference id="pref-layout" name="extensions.zotero.layout" type="string" onchange="Zotero.getActiveZoteroPane().updateLayout();"/>
+			<preference id="pref-layout" name="extensions.zotero.layout" type="string"/>
 			<preference id="pref-automaticScraperUpdates" name="extensions.zotero.automaticScraperUpdates" type="bool"/>
 			<preference id="pref-reportTranslationFailure" name="extensions.zotero.reportTranslationFailure" type="bool"/>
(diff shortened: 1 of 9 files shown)


chrome/content/zotero/preferences/preferences_general.xul
```diff
@@ -32,6 +32,7 @@
 		
 		<preferences id="zotero-prefpane-general-preferences">
 			<preference id="pref-fontSize" name="extensions.zotero.fontSize" type="string"/>
+			<preference id="pref-layout" name="extensions.zotero.layout" type="string" instantApply="true" onchange="Zotero.getActiveZoteroPane().updateLayout();"/>
```
dstillman: Why `instantApply` here? This should be platform-dependent.

chrome/content/zotero/preferences/preferences_general.xul
```diff
@@ -58,6 +59,17 @@
 				<rows id="zotero-prefpane-general-rows">
 					<row>
 						<hbox align="center">
+							<label value="&zotero.preferences.layout;" control="layout"/>
+						</hbox>
+						<radiogroup id="layout" orient="horizontal" align="center" preference="pref-layout"
+							onblur="Zotero.Prefs.set('layout', this.value);">
```
dstillman: The `onblur` shouldn't be necessary — that's the point of the `preference`.

chrome/content/zotero/preferences/preferences_general.xul
dstillman: (I like instantApply on OS X, but none of the other prefs on Windows and Linux do it.)

chrome/content/zotero/preferences/preferences_general.xul
LinuxMercedes: Y'know, I could have sworn I had trouble with getting this to work right otherwise, but now that I take it out again it works. Guess I was too hasty the first time through.

chrome/content/zotero/zoteroPane.xul
```diff
@@ -357,218 +358,220 @@
...
+									ondragenter="return ZoteroPane_Local.itemsView.onDragEnter(event)"
+									ondragover="return ZoteroPane_Local.itemsView.onDragOver(event)"
+									ondrop="return ZoteroPane_Local.itemsView.onDrop(event)"/>
+							</tree>
+
+							<!-- Label for displaying messages when items pane is hidden
 							 (e.g. "Advanced search mode — press Enter to search.")-->
 						<vbox id="zotero-items-pane-message-box" pack="center" align="center"/>
 					</deck>
 				</vbox>
```
dstillman: Indentation here isn't right

LinuxMercedes: Fix'd

chrome/content/zotero/preferences/preferences_general.xul
LinuxMercedes: Shall I add it just for OS X then?

dstillman:
(No need to rebase each time — better to just add new commits, and we'll squash at the end.)

dstillman:
This is what I get when I toggle to Horizontal on OS X:

layout

dstillman:
I don't think "Horizontal" and "Vertical" make intuitive sense — I think of the classic view as horizontal, even if Thunderbird calls it vertical. "Standard"/"Stacked"? "Wide"/"Narrow"? Doesn't technically affect the width, but the new layout is better for narrow windows, so maybe that's the clearest way of explaining it. (For that matter, does it affect the min-width?)

chrome/content/zotero/preferences/preferences_general.xul
dstillman: No, it's automatic.

LinuxMercedes:
Haha, I think of the classic view as "vertical"! Standard/Stacked makes sense to me; Wide/Tall would also make sense. 

I set a min-width on the `box` of 560px since anything narrower than that started to scrunch the toolbars.

LinuxMercedes:
I tried to repro your issue by checking out `master`, opening about:config and clearing `zotero.pane.persist` and `zotero.layout`, quitting (which should save old-style `persist` values), then checking out `layout` and re-opening Firefox, but it worked for me. 

I wonder if there is some weird interaction between an existing `zotero.pane.persist` setting or some other saved setting and the new layout. Or, it could be that OS X Firefox defaults those widths to 0 or min-width for some reason?

dstillman:
New profile:

new_profile

LinuxMercedes:
Well, I pushed something that fixes one bug I ran into while testing in
a new profile. I'm not optimistic it'll fix yours as well, but try it out?

dstillman:
layout

dstillman:
If I set `display: none` on `zotero-items-splitter`, I get this:

layout

LinuxMercedes:
Perhaps this is some interaction with the mac-specific stuff? If you open Zotero in a tab, you can use the Inspector on it. That may shed some light on things. 
2016-05-24-194100_670x139_scrot

dstillman:
(It doesn't require tab mode — you can use Browser Toolbox (a.k.a. remote debugging) to debug any chrome UI.)

Anyway, I can't debug this further at the moment — if you have access to a Mac, you can take a look. Otherwise this may have to wait a while. Has to be tested on Windows also.

LinuxMercedes:
I got my hands on a Mac and I think I got everything working. I'll try out Windows sometime, but I looked at `win/overlay.css` and didn't see anything that would obviously break this.

dstillman:
Awesome, thanks — that works.

One slight glitch with the stacked layout on OS X is that the splitter above the item pane is different from the bottom border of the item pane tabs, which is now right below it. We can worry about that later, though.

Let's go with "Standard"/"Stacked" for now. And maybe we'll replace those in the prefs with graphical representations at some point.

Looks fine on Windows, but I'm seeing some weird behavior changing the setting. Clicking between the two radios changes the layout — but only on the first click after opening the prefs, and even though about:config doesn't show the pref changing until after OK is clicked. It's also changing the layout to the reverse of the setting. Are you seeing any of this? (Works fine with instantApply on OS X.) Might be related to the `onchange` on the `preference`? Actually better to put that handler in Zotero.Prefs._handlers in zotero.js anyway, so that it applies no matter how the pref is changed.

LinuxMercedes:
Hmm, I'm not seeing any of the Windows weirdness on Linux; switching views does switch automatically ad nauseam. Let me see if I can reproduce it on a box here.

chrome/content/zotero/preferences/preferences_general.xul
```diff
@@ -58,6 +59,16 @@
 				<rows id="zotero-prefpane-general-rows">
 					<row>
 						<hbox align="center">
+							<label value="&zotero.preferences.layout;" control="layout"/>
+						</hbox>
+						<radiogroup id="layout" orient="horizontal" align="center" preference="pref-layout">
+							<radio id="horizontal" label="&zotero.preferences.layout.standard;" value="Standard"/>
+							<radio id="vertical" label="&zotero.preferences.layout.stacked;" value="Stacked"/>
```
dstillman: Pref values should be lowercase.

dstillman:
Thanks!

PR #1011, opened 2016-05-25
Don't yield blurOpenField

LinuxMercedes:
I think this got lost somewhere along the way based on `git blame`.

This also fixes a bug where sometimes clicking on new items doesn't display them until you switch tabs in the Item Pane.

### DIFF
--- a/chrome/content/zotero/itemPane.js
+++ b/chrome/content/zotero/itemPane.js
@@ -78,7 +78,7 @@ var ZoteroItemPane = new function() {
 			switch (index) {
 				case 0:
 				case 2:
-					yield box.blurOpenField();
+					box.blurOpenField();
 					// DEBUG: Currently broken
 					//box.scrollToTop();
 					break;


dstillman:
Not sure I follow. itembox.xml::blurOpenField() returns a promise, so it should be yielded.

LinuxMercedes:
Hmm, this may not be the right fix, then, but I have an issue then where clicking on items in the Items pane does not update the Item pane because `viewItem` never gets past line 81 (verified by putting debug statements before and after line 81).

dstillman:
Can you reproduce that reliably? Shouldn't be necessary, but are you able to log an error if you wrap that line in a try/catch with `Zotero.debug(e)`?

LinuxMercedes:
I can reproduce it reliably on my box. Putting that try-catch in makes it work; here's the caught exception:

```

(3)(+0000004): Error: Field not specified

    Error: Field not specified
        Zotero.Item.prototype.setField@chrome://zotero/content/xpcom/data/item.js:661:9
        _modifyField/<@chrome://zotero/content/bindings/itembox.xml:1980:7
    From previous event:
        this.spawn@chrome://zotero/content/xpcom/zotero.js:1753:11
        blurOpenField@chrome://zotero/content/bindings/itembox.xml:2263:13
        ZoteroItemPane</this.viewItem<@chrome://zotero/content/itemPane.js:82:12
    From previous event:
        ZoteroItemPane.viewItem@chrome://zotero-better-bibtex/content/itemPane.js:47:5
        ZoteroPane</this.itemSelected/<@chrome://zotero/content/zoteroPane.js:1448:13
    From previous event:
        this.spawn@chrome://zotero/content/xpcom/zotero.js:1753:11
        ZoteroPane</this.itemSelected@chrome://zotero/content/zoteroPane.js:1326:10
        onselect@chrome://browser/content/browser.xul:1:1
        onxblmousedown@chrome://global/content/bindings/tree.xml:1103:14
```

LinuxMercedes:
Debugged this some more: turns out it was a side effect of Better BibTex. That's odd since I thought I checked this against my zotero testing profile as well...

PR #1012, opened 2016-05-25
Closes #1008. Omit certain item list columns in feed views

### LINKED ISSUE #1008: Omit certain items list columns in Feed views
We shouldn't show Date Added, Date Modified, Attachments, or Notes at least, and they should be greyed out in the column picker. Switching back and forth shouldn't affect those columns' presence in other views.

adomasven:
Omitted Date Added, Date Modified, Attachments and Notes. Implemented using custom treecol attributes with `zotero-persist` to retain the configuration between sessions.

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -1242,11 +1242,11 @@ var ZoteroPane = new function()
 			
 			document.getElementById('zotero-items-tree').view = this.itemsView;
 			
-			// Add events to treecolpicker to update menu before showing/hiding
 			try {
 				let treecols = document.getElementById('zotero-items-columns-header');
 				let treecolpicker = treecols.boxObject.firstChild.nextSibling;
 				let menupopup = treecolpicker.boxObject.firstChild.nextSibling;
+				// Add events to treecolpicker to update menu before showing/hiding
 				let attr = menupopup.getAttribute('onpopupshowing');
 				if (attr.indexOf('Zotero') == -1) {
 					menupopup.setAttribute('onpopupshowing', 'ZoteroPane.itemsView.onColumnPickerShowing(event);')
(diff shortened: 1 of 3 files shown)


dstillman:
I think it'd be better to generalize the `treecol` attribute, since this may apply to other views — maybe `hiddenviews`? (I don't love that name, but not sure what would be better.) Can be a comma-separate list of `type` values from `Zotero.CollectionTreeRow.prototype.type` (e.g., 'feed').

adomasven:
Yeah, could do. Probably `hidden-in="feeds publications"` could work. Mimic classes being space separated.

dstillman:
Yeah, that works.

adomasven:
Updated

chrome/content/zotero/zoteroPane.js
```diff
@@ -1256,6 +1256,28 @@ var ZoteroPane = new function()
 						// Keep whatever else is there
 						+ ' ' + menupopup.getAttribute('onpopuphidden');
 				}
+				
+				// Hide certain columns for feeds
+				let elems = treecols.getElementsByAttribute('hidden-in', '*');
+				for (let i = 0; i < elems.length; i++) {
+					let shouldHide = elems[i].getAttribute('hidden-in').split(' ')
+						.indexOf(collectionTreeRow.type) != -1;
+					let usuallyHidden = elems[i].hasAttribute('usually-hidden');
```
dstillman: So `usually-hidden` means "manually hidden by the user outside of views that force it hidden"?

If so, I think `user-hidden` might be clearer.

chrome/content/zotero/zoteroPane.js
```diff
@@ -1256,6 +1256,28 @@ var ZoteroPane = new function()
...
+					// The slightly convoluted logic here is such:
+					// If the current tree being displayed is in the list of types 'hidden-in'
+					// specifies and it's not already been hidden by another 'hidden-in' rule
+					// (i.e. it's normal state of being displayed has not been suppressed yet),
```
dstillman: its

adomasven:
Changed `usually-hidden` to `user-hidden`. Fixed grammar.

PR #1013, opened 2016-05-26
Add items to a collection inside a DB transaction

LinuxMercedes:
Simply requiring a transaction here causes RecognizePDF to fail with an unknown error.


dstillman:
This is by design. If this is called somewhere not within a transaction, the transaction should be added to the calling code.

Issue #1014, opened 2016-05-26
5.0: Retrieve Metadata from PDF fails in a collection


dstillman:
Follow-up from https://github.com/zotero/zotero/pull/1013

PR #1015, opened 2016-05-26
Add items from RecognizePDF in a DB transaction

LinuxMercedes:
This fixes an 'An unexpected error occurred.' issue when attempting to get metadata from a PDF.

### DIFF
--- a/chrome/content/zotero/xpcom/data/collection.js
+++ b/chrome/content/zotero/xpcom/data/collection.js
@@ -358,8 +358,8 @@ Zotero.Collection.prototype._finalizeSave = Zotero.Promise.coroutine(function* (
  * @param {Number} itemID
  * @return {Promise}
  */
-Zotero.Collection.prototype.addItem = function (itemID) {
-	return this.addItems([itemID]);
+Zotero.Collection.prototype.addItem = function (itemID, options) {
+	return this.addItems([itemID], options);
 }
 
 
@@ -371,7 +371,9 @@ Zotero.Collection.prototype.addItem = function (itemID) {
  * @param {Number[]} itemIDs
  * @return {Promise}
  */
-Zotero.Collection.prototype.addItems = Zotero.Promise.coroutine(function* (itemIDs) {
+Zotero.Collection.prototype.addItems = Zotero.Promise.coroutine(function* (itemIDs, options = {}) {
+	options.skipDateModifiedUpdate = true;
+
 	if (!itemIDs || !itemIDs.length) {
 		return;
 	}
@@ -389,9 +391,7 @@ Zotero.Collection.prototype.addItems = Zotero.Promise.coroutine(function* (itemI
 		
 		let item = this.ChildObjects.get(itemID);
 		item.addToCollection(this.id);
-		yield item.save({
-			skipDateModifiedUpdate: true
-		});
+		yield item.save(options);
 	}
 	
 	yield this.loadDataType('childItems');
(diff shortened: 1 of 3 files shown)


dstillman:
Thanks, but we need a test for this, and the transaction should go around the `item.save` below as well, and the collection get can be made synchronous. If you want to do all that, that's great — if not, no worries, and I'll deal with it later (#1014).

LinuxMercedes:
I may do that if I get some time. 

Somewhat off-topic: where is the right place to report bugs in `master`? I try to make a policy of 'provide a fix with a bug report', but I've got one that I have no idea where to start debugging it. 

Also, thanks for being patient with my out-of-the-blue pull requests!

LinuxMercedes:
Well, I took a stab at it, but I'm not quite sure what is going on with these unit tests.

chrome/content/zotero/xpcom/data/collection.js
```diff
@@ -371,7 +371,10 @@ Zotero.Collection.prototype.addItem = function (itemID) {
  * @param {Number[]} itemIDs
  * @return {Promise}
  */
-Zotero.Collection.prototype.addItems = Zotero.Promise.coroutine(function* (itemIDs) {
+Zotero.Collection.prototype.addItems = Zotero.Promise.coroutine(function* (itemIDs, options) {
+	options = options || {}
```
dstillman: `function* (itemIDs, options = {})`

chrome/content/zotero/xpcom/data/item.js
```diff
@@ -1396,9 +1396,11 @@ Zotero.Item.prototype._saveData = Zotero.Promise.coroutine(function* (env) {
 			
 			let newParentItemNotifierData = {};
 			//newParentItemNotifierData[newParentItem.id] = {};
-			Zotero.Notifier.queue(
-				'modify', 'item', parentItemID, newParentItemNotifierData, env.options.notifierQueue
-			);
+			if (!env.options.skipNotifier) {
+				Zotero.Notifier.queue(
+						'modify', 'item', parentItemID, newParentItemNotifierData, env.options.notifierQueue
+						);
```
dstillman: Here and elsewhere, keep original style (one tab, closing paren aligned with opening line)

LinuxMercedes: Whoops, vim autoformat got too excited.

test/tests/recognizePDFTest.js
```diff
@@ -1,4 +1,4 @@
-describe.skip("PDF Recognition", function() {
+describe("PDF Recognition", function() {
```
dstillman: Oh, sorry, didn't realize these were currently disabled. These should actually stay disabled until they mock the HTTP services.

Asking you to do this was actually a bit of a tall order, among other things because these tests predate some advances in Zotero testing technology. Let's leave this commit out, and I'll add a test separately that shows the best way to do this.

dstillman: Yeah, it would still fail an assertion for the item existing in the collection, but the recognizePDF code should obviously send errors back up the chain. (That's specific to that code, not a general problem.) I'll deal with that separately as well. Thanks.

LinuxMercedes: Sounds good to me; I'm definitely in over my head here.

I'll rebase this in a second.

dstillman:
> Somewhat off-topic: where is the right place to report bugs in master? I try to make a policy of 'provide a fix with a bug report', but I've got one that I have no idea where to start debugging it. 

We generally point feature requests and user-level bug reports to the forums, but if it's code-related or you're asking with an eye towards fixing it, opening an issue here is fine for obvious bugs.

Thanks!

PR #1019, opened 2016-05-31
Don't keep collections collapsed

### LINKED ISSUE #825: Reopen collapsed collections pane on restart
https://forums.zotero.org/discussion/51039/

Should probably just do this too, until we have a better solution for both. (Follow-up from #678.)

LinuxMercedes:
Closes #825


dstillman:
Closed by https://github.com/zotero/zotero/commit/787e57f6a29b73ee9fac2b4a4d7da7b78bbe1fe8

PR #1029, opened 2016-06-11
Fix more for-each-in

zuphilip:


zuphilip:
Yeah! That was the :tada: 8000th commit :tada: for `zotero/zotero` (only counted in github). :smile:

PR #1032, opened 2016-06-13
Changes date parsing and handling for feeds. Closes #1024, #1025

### LINKED ISSUE #1024: Update feed item date parsing for Firefox 47
Firefox 46:

```
new Date("2015-06-07 20:56:00") + ""
/*
Invalid Date
*/
```

Firefox 47:
```
new Date("2015-06-07 20:56:00") + ""
/*
Sun Jun 07 2015 20:56:00 GMT-0400 (EDT)
*/
```

This means that [this line](https://github.com/zotero/zotero/blob/fdd90b04d305b117ab5235ce3229b311e8775306/chrome/content/zotero/xpcom/data/feedItem.js#L115), which previously didn't work, now does, but it parses the date in the local time zone, causing [this test](https://github.com/zotero/zotero/blob/fdd90b04d305b117ab5235ce3229b311e8775306/test/tests/feedItemTest.js#L103) to fail in non-UTC-equivalent time zones.

1) Would `accessDate` ever actually exist in feed items? I don't think so? So we should probably test an SQL date in `dateAdded` or `dateModified` (which come from the feed entry dates, right?) instead.

2) Such dates (which hopefully won't actually appear) should be interpreted as UTC.

### LINKED ISSUE #1025: Update Date field handling for feeds
1) The Date field should be shown by default in the feeds view.

2) Right now, it seems the Date field is being stored and displayed as an HTTP date, and I'm seeing GMT dates. Instead of storing that as an HTTP date, we should parse that, store it as a UTC SQL date, and then display it localized in the middle and right panes. (The former will require a change to `itemTreeView.js::getCellText()`, the latter to `itembox.xml::createValueElement()`. We don't normally localize the Date field for non-feed items, but in this case it makes sense.)

adomasven:
Closes #1024 
Closes #1025 

Also see [this line](https://github.com/zotero/zotero/pull/1032/files#diff-3a5e2c983420f6c7932beb3925d76976R1046) where after a decision that '.' separator for date in dmy standard may be a bit too crazy it was changed to '/'.

### DIFF
--- a/chrome/content/zotero/bindings/itembox.xml
+++ b/chrome/content/zotero/bindings/itembox.xml
@@ -1289,11 +1289,15 @@
 						case 'dateAdded':
 						case 'dateModified':
 						case 'accessDate':
+						case 'date':
 						
 						// TEMP - NSF
 						case 'dateSent':
 						case 'dateDue':
 						case 'accepted':
+							if (fieldName == 'date' && this.item._objectType != 'feedItem') {
+								break;
+							}
 							if (valueText) {
 								var date = Zotero.Date.sqlToDate(valueText, true);
 								if (date) {
--- a/chrome/content/zotero/xpcom/date.js
+++ b/chrome/content/zotero/xpcom/date.js
@@ -477,6 +477,49 @@ Zotero.Date = new function(){
 		return date;
 	}
 	
+	this.isHTTPDate = function(str) {
+		var dayNames = ['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun'];
+		var monthNames = ["Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep",
+					"Oct", "Nov", "Dec"];
+		str = str.trim();
+		var temp = str.split(',');
+		if (temp.length > 1) {
+			var dayOfWeek = temp[0];
+			if(dayNames.indexOf(dayOfWeek) == -1) {
+				return false;
+			}
+
+			str = temp[1].trim();
+		}
+		temp = str.split(' ');
+		temp = temp.filter((t) => ! t.match(/^\s*$/));
+		if (temp.length < 5) {
+			return false;
+		}
+		if (!temp[0].trim().match(/[0-3]\d/)) {
+			return false;
+		}
+		if (monthNames.indexOf(temp[1].trim()) == -1) {
+			return false;
+		}
+		if (!temp[2].trim().match(/\d\d\d\d/)) {
+			return false;
+		}
+		temp.splice(0, 3);
+		var time = temp[0].trim().split(':');
+		if (time.length < 2) {
+			return false;
+		}
+		for (let t of time) {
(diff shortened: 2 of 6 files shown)


chrome/content/zotero/bindings/itembox.xml
```diff
@@ -1289,11 +1289,15 @@
...
 						case 'dateModified':
 						case 'accessDate':
+						case 'date':
 						
 						// TEMP - NSF
 						case 'dateSent':
 						case 'dateDue':
 						case 'accepted':
+							if (fieldName == 'date' && this.item._objectType != 'feedItem') {
+								break
```
dstillman: Missing colon

test/tests/feedItemTest.js
```diff
@@ -92,18 +92,35 @@ describe("Zotero.FeedItem", function () {
...
+				},
+				{
+					itemType: "journalArticle",
+					date: "07/06/2015" // american format
+				}
+			];
+			var expectedDates = [
+				'2015-06-07 20:56:00',
+				'2016-06-13 10:25:57',
+				'2015-06-18 20:00:00',
```
dstillman: This is wrong, though — the 20 is being picked up as a time. This shouldn't actually be parsed by new Date — we should add a Zotero.Date.isHTTP() and limit the parsing to that. Otherwise it should use strToDate(), which (I think) should get June 18, 2015.

test/tests/feedItemTest.js
```diff
@@ -92,18 +92,35 @@ describe("Zotero.FeedItem", function () {
...
+				Zotero.Date.dateToSQL(new Date(2015, 6, 6), true)
```
dstillman: Can we just store without the time when it comes from strToDate? This is a little hacky.

Also, if I recall, the parsing of "07/06/2015" will actually depend on the current locale, so this is a little tricky. Probably need to overwrite Zotero.locale to get a reliable test here.

adomasven:
Updated. The code is now a bit spaghetti unfortunately.

chrome/content/zotero/xpcom/date.js
```diff
@@ -477,6 +477,48 @@ Zotero.Date = new function(){
 		return date;
 	}
 	
+	this.isHTTPDate = function(str) {
+		var dayNames = ['Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat', 'Sun'];
+		var monthNames = this.getMonths().short;
```
dstillman: This gets localized dates. We need the English abbreviations from the spec.

(But not sure why you didn't just find a regexp for this whole function? E.g., http://stackoverflow.com/q/9352003)

adomasven: >(But not sure why you didn't just find a regexp for this whole function? E.g., http://stackoverflow.com/q/9352003)

My google-fu deceived me.

adomasven: Updated. I trust this code I wrote and I don't trust a regex without enough approval/disapproval. So this code it is.

PR #1033, opened 2016-06-14
Add a tab to change synced libraries in sync preferences

adomasven:
Adds a preference dialog to change which libraries (groups) are synced by default. Only GUI.

Saves the pref as a JSON array of group IDs in `sync.syncedLibraries`.


chrome/content/zotero/preferences/preferences_sync.xul
```diff
@@ -267,6 +262,23 @@
...
+					<groupbox>
+						<caption label="&zotero.preferences.sync.syncedLibraries;"/>
+						
+						<tree id="synced-libraries-tree" flex="1" hidecolumnpicker="true" rows="10" seltype="single" editable="true"
+							ondblclick="Zotero_Preferences.Sync.dblClickSyncedLibrary(this.currentIndex);"
+							onclick="Zotero_Preferences.Sync.clickSyncedLibrary(event)"
+							onkeyup="if (event.keyCode == event.DOM_VK_SPACE) { Zotero_Preferences.Sync.dblClickSyncedLibrary(this.currentIndex); }">
+							<treecols>
+								<treecol editable="true" type="checkbox" id="synced-libraries-checked" label="&zotero.preferences.sync.syncedLibraries.sync;"/>
+								<treecol editable="false" id="synced-libraries-name" flex="1" label="&zotero.preferences.sync.syncedLibraries.sync;"/>
```
dstillman: Wrong label on this one

dstillman:
I'm seeing some weird behavior with clicking/double-clicking, at least with instantApply. (You can probably toggle instantApply on temporarily on the pref to test.) With about:config open and filtered to the setting, it doesn't always change in line with the checkbox showing and hiding.

dstillman:
Actual libraryIDs in the pref are all sorts of messed up for me after various clicking/double-clicking patterns — so I close the dialog with no groups selected and have some in the pref, clicking on group rows toggles libraryID 1 on/off, etc.

chrome/content/zotero/preferences/preferences_sync.js
```diff
@@ -177,6 +177,129 @@ Zotero_Preferences.Sync = {
...
+		 
+		var cell = row.firstChild.firstChild;
+		cell.setAttribute('value', indexOfId != -1);
+	},
+	
+	
+	initSyncedLibraries: Zotero.Promise.coroutine(function* () {
+		var apiKey = Zotero.Sync.Data.Local.getAPIKey();
+		var client = Zotero.Sync.Runner.getAPIClient({apiKey});
+		var keyInfo = yield Zotero.Sync.Runner.checkAccess(client, {timeout: 5000});
```
dstillman: What happens if this (or another API request) fails? Dialog should at least close.

chrome/content/zotero/preferences/preferences_sync.js
```diff
@@ -177,6 +177,129 @@ Zotero_Preferences.Sync = {
...
+
+		// Add default rows
+		addRow(Zotero.getString("pane.collections.library"), null, true, false);
+		addRow(Zotero.getString("pane.collections.publications"), null, true, false);
+		addRow(Zotero.getString("pane.collections.feeds"), null, true, false);
+
+		var syncedLibraries = JSON.parse(Zotero.Prefs.get('sync.syncedLibraries') || '[]');
+		
+		// Load up remote groups
+		var groupIDs = yield client.getGroupVersions(keyInfo.userID);
```
dstillman: We don't want to use `getGroupVersions` here. Instead of the `.../groups?format=versions` request that that makes, we want a `getGroups` that just makes a `.../groups` request, which includes the group names in the response. (And we should maybe rename `getGroupInfo` to `getGroup` for consistency, since it's just doing the same thing but with a `groupID`.)

chrome/content/zotero/preferences/preferences_sync.js
```diff
@@ -177,6 +177,129 @@ Zotero_Preferences.Sync = {
...
+		for (let groupID in groupIDs) {
+			let group = yield client.getGroupInfo(groupID);
+			addRow(group.data.name, group.id, syncedLibraries.indexOf(group.id) != -1);
+		}
+		
+		// Remove the animated row
+		clearTimeout(loadTimeout);
+		treechildren.removeChild(treechildren.firstChild);
```
dstillman: This temporary line doesn't really work — it doesn't make sense (because there are already libraries displayed) and makes everything jump around. If we were going to do a temp line in the list this way, it should say "Loading groups…" and be in the place where the groups will go.

But since we can remove the individual group calls, I'd say we should either not show a temporary line at all (since most of the time the loading will be very quick) or not show any library lines at first and show "Loading libraries…". I'd be inclined to do the former.

dstillman:
Dialog should be much narrower — maybe 415px.

dstillman:
Checkboxes are misaligned:

align

dstillman:
Button should be disabled when "Sync automatically" is unchecked

chrome/content/zotero/preferences/syncedLibraries.xul
```diff
@@ -0,0 +1,67 @@
...
+<!DOCTYPE window [
+  <!ENTITY % prefWindow SYSTEM "chrome://zotero/locale/preferences.dtd">
+  %prefWindow;
+  <!ENTITY % common SYSTEM "chrome://zotero/locale/zotero.dtd">
+  %common;
+]>
+
+<dialog xmlns="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul"
+	title="" buttons="accept"
+	id="zotero-quickCopySiteEditor"
```
dstillman: Old id

dstillman:
"Synced Libraries" is a little misleading, since this only applies to auto-sync (or will once we add a manual option, but that's easy). "Libraries to Sync" is a little better?

And autoSyncLibraries instead of syncedLibraries for the internal names?

adomasven:
> Button should be disabled when "Sync automatically" is unchecked

So will pressing the sync button sync all libraries or just My Library, My Publications and Feeds? Surely you'd want these settings apply there too?

dstillman:
> So will pressing the sync button sync all libraries or just My Library, My Publications and Feeds? Surely you'd want these settings apply there too?

Oh, good point. Not sure about my button placement then, either — probably the button should go first, followed by "Sync automatically" on the next row.

So, let's back up — do we actually want to allow context-menu syncing of libraries disabled here? Or should you need to enable syncing for a library here to have a context-menu sync option at all? Not allowing would also be a bit safer, since you couldn't accidentally trigger a sync when you were trying to create a new collection or something. It would also be easier to communicate that way, since then we _could_ call it "Libraries to Sync". Otherwise we'd need something like:

"Choose the libraries to sync when you press the Sync button and during automatic syncing. If a library is unchecked, you can still sync the library from its context menu."

Need to think about this a little more, including what we'll want beyond the beta. (Ability to set a library as read-only (on the key) for a given client, regardless of the member's permissions, in order to prevent accidental modifications? Ability to remove a library from the client (and key) altogether? Is there a use case for wanting a library present but only manually synced, and synced regularly enough that toggling it on in the prefs would be annoying?)

adomasven:
> So, let's back up — do we actually want to allow context-menu syncing of libraries disabled here? Or should you need to enable syncing for a library here to have a context-menu sync option at all? Not allowing would also be a bit safer, since you couldn't accidentally trigger a sync when you were trying to create a new collection or something.

So I was working under the assumption you disable batch group syncing here, to prevent data loss/downloading excess data. And then you still have a context menu option to sync libraries manually (possibly with a warning message?)

And then I guess we'd need to create empty groups for those that aren't auto-synced, so the users can right click to sync.

> Ability to set a library as read-only (on the key) for a given client, regardless of the member's permissions, in order to prevent accidental modifications? Ability to remove a library from the client (and key) altogether? Is there a use case for wanting a library present but only manually synced, and synced regularly enough that toggling it on in the prefs would be annoying?

Limited data plans on home/travel computer? I think this could be included after the beta under the above plan. And then additional sync controls can be built on top of that fairly straighforwardly, I'd say, if there's need on the user end.

dstillman:
> And then you still have a context menu option to sync libraries manually (possibly with a warning message?)

I guess no real need for a warning, particularly given that you'll be able to stop syncs (once I get around to implementing it), and if we're saying that after the beta this is largely for limited data plans (which was my thinking too).

> And then I guess we'd need to create empty groups for those that aren't auto-synced, so the users can right click to sync.

I think it'd be OK not to create groups that haven't been selected here at least once.

I think "Libraries to Sync" is still a little clearer than "Synced Libraries" for what this will actually control.

So given that this won't apply just to auto-sync, I think we want this for the layout:

prefs

dstillman:
After clicking "Choose Libraries":

> group.data is undefined
> preferences_sync.js:319

chrome/content/zotero/preferences/preferences_sync.js
```diff
@@ -177,6 +177,148 @@ Zotero_Preferences.Sync = {
...
+		
+		// Add default rows
+		addRow(Zotero.getString("pane.collections.library"), null, true, false);
+		addRow(Zotero.getString("pane.collections.publications"), null, true, false);
+		addRow(Zotero.getString("pane.collections.feeds"), null, true, false);
+		
+		// Add group rows
+		var librariesToSync = JSON.parse(Zotero.Prefs.get('sync.librariesToSync') || '[]');
+		for (let group in groups) {
```
dstillman: `in` -> `of`

chrome/content/zotero/preferences/preferences_sync.js
```diff
@@ -177,6 +177,148 @@ Zotero_Preferences.Sync = {
...
+		addRow(loadingLabel, "loading", false, false);
+		var cell = treechildren.firstChild.getElementsByAttribute("value", "loading")[0];
+		function animateLoadingLabel() {
+			if (cell.getAttribute('label').length < loadingLabel.length + 3) {
+				cell.setAttribute('label', cell.getAttribute('label') + '.');
+			} else {
+				cell.setAttribute('label', loadingLabel);
+			}
+			loadTimeout = setTimeout(animateLoadingLabel, 1000.0/3.0);
+		}
```
dstillman: Clever as this is, I don't think it works — I suspect ellipses are different in different locales, and even in English three periods isn't technically typographically correct. (If we did do it, we'd also want it to start with the three instead of none, since starting with none looks like a mistake.) I think we can just add a real ellipsis to the localized term. Sorry to ruin the fun.

dstillman:
Other than figuring out how to represent "Feeds", I don't think there's any reason not to make the user libraries editable too. We'd just initialize the pref with "My Library, "My Publications", and something for feeds when linking the account.

Also we should clear the pref when unlinking.

adomasven:
> Other than figuring out how to represent "Feeds", I don't think there's any reason not to make the user libraries editable too. We'd just initialize the pref with "My Library, "My Publications", and something for feeds when linking the account.

Well, the way things are set up now, Feeds wouldn't sync unless you sync My Library regardless. But not having them there is confusing too? Thoughts?

And it's a bit weird to not want to sync your own library, but something else. But if you're willing to implement, then we can go for that.

dstillman:
> Well, the way things are set up now, Feeds wouldn't sync unless you sync My Library regardless. But not having them there is confusing too? Thoughts?

Just "My Library & Feeds" for now?

> And it's a bit weird to not want to sync your own library, but something else.

You might use Zotero groups at work but not want your personal library on your work computer.

dstillman:
> JSON.parse: unexpected character at line 1 column 1 of the JSON data  preferences_sync.js:302

With empty pref

dstillman:
If no pref, should default to the same as when linking, though. Otherwise anyone upgrading the beta won't have anything checked in the dialog.

Also need to decide what to do for syncing if no pref, which will also be the default case when upgrading the beta. Sync everything, I think? Logic being: anyone who's already using the beta and has set up syncing is already syncing everything, so would reasonably expect it to continue, whereas new beta users who link their accounts will see the "Choose Libraries…" button.

dstillman:
Actually, by that logic, if no pref, everything should be checked in the dialog.

chrome/content/zotero/preferences/preferences_sync.js
```diff
@@ -308,7 +311,7 @@ Zotero_Preferences.Sync = {
 		
 		// Add group rows
 		for (let group of groups) {
-			addRow(group.data.name, group.id, librariesToSync.indexOf(group.id) != -1);
+			addRow(group.data.name, group.id, librariesToSync === true || librariesToSync.indexOf(group.id) != -1);
```
dstillman: When would this be `true`?

dstillman:
Just started reviewing :) <br><br>:octocat: *Sent from [GH](http://nodegh.io).*

dstillman:
For some reason double-click isn't toggling for me on the first three rows. Works on the last three.

adomasven:
Updated.

dstillman:
I'm getting all sorts of bugginess with the checkboxes now.

PR #1042, opened 2016-06-20
Adds custom view groups for different library types.

### LINKED ISSUE #1034: Store separate column settings for feeds
So after https://github.com/zotero/zotero/issues/1030, I'm realizing we have some other awkwardness with columns when switching between feeds and elsewhere:

1) If you add a column in another view, it's present in feeds. Sometimes this makes sense, but since feed items aren't translated, for a lot of fields this will be an empty column.

2) It's weird not to have Date in Feeds, but the Date column isn't present by default.

Seems like we should maybe have a different default set of columns for feeds and remember the visibility settings for those separately from other library types?

Not sure any future library types (e.g., subscriptions) would share columns with feeds. Feeds seem sort of distinct here, since they have pretty fixed metadata.

adomasven:
I originally attempted this with zotero-persists and column attributes,
but there is no good way to make it succint paramswise and the code was
painful to look at too. Thus different group settings are stored in
preferences.

Currently there are two view groups: "feed" and "default". Items view
columns have two new attributes:
`default-in` - a space separated list of views in which a column is
visible by default
`disabled-in` - a space separated list of views in which a column is
disabled by default (invisible + not possible to enable)

Currently the settings are stored in a pref named `itemsView.columnVisibility`, but it could probably use a better name.

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -1243,39 +1243,41 @@ var ZoteroPane = new function()
 			document.getElementById('zotero-items-tree').view = this.itemsView;
 			
 			try {
+				let tree = document.getElementById('zotero-items-tree');
 				let treecols = document.getElementById('zotero-items-columns-header');
 				let treecolpicker = treecols.boxObject.firstChild.nextSibling;
 				let menupopup = treecolpicker.boxObject.firstChild.nextSibling;
 				// Add events to treecolpicker to update menu before showing/hiding
 				let attr = menupopup.getAttribute('onpopupshowing');
 				if (attr.indexOf('Zotero') == -1) {
-					menupopup.setAttribute('onpopupshowing', 'ZoteroPane.itemsView.onColumnPickerShowing(event);')
+					menupopup.setAttribute('onpopupshowing', 'ZoteroPane.itemsView.onColumnPickerShowing(event); '
 						// Keep whatever else is there
-						+ ' ' + attr;
-					menupopup.setAttribute('onpopuphidden', 'ZoteroPane.itemsView.onColumnPickerHidden(event);')
+						+ attr);
+					menupopup.setAttribute('onpopuphidden', 'ZoteroPane.itemsView.onColumnPickerHidden(event); '
 						// Keep whatever else is there
-						+ ' ' + menupopup.getAttribute('onpopuphidden');
+						+ menupopup.getAttribute('onpopuphidden'));
 				}
 				
-				// Hide certain columns for feeds
-				let elems = treecols.getElementsByAttribute('hidden-in', '*');
-				for (let i = 0; i < elems.length; i++) {
-					let shouldHide = elems[i].getAttribute('hidden-in').split(' ')
-						.indexOf(collectionTreeRow.type) != -1;
-					let userHidden = elems[i].hasAttribute('user-hidden');
-					// The slightly convoluted logic here is such:
-					// If the current tree being displayed is in the list of types 'hidden-in'
-					// specifies and it's not already been hidden by another 'hidden-in' rule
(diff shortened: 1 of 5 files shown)


chrome/content/zotero/zoteroPane.js
```diff
@@ -1243,39 +1243,41 @@ var ZoteroPane = new function()
...
-					} else if (!shouldHide && userHidden) {
-						// Restore old value on exit from feed view
-						elems[i].setAttribute('hidden', elems[i].getAttribute('user-hidden'));
-						elems[i].removeAttribute('user-hidden');
+				// Items view column visibility for different groups
+				let prevViewGroup = tree.getAttribute('current-view-group');
+				let curViewGroup = collectionTreeRow.visibilityGroup;
+				tree.setAttribute('current-view-group', curViewGroup);
+				if (curViewGroup != prevViewGroup) {
+					let cols = Array.prototype.slice.call(treecols.querySelectorAll('treecol'));
```
dstillman: `Array.from()`, and though it almost certainly doesn't matter, might as well use the (generally faster) `treecols.getElementsByTagName()`

chrome/content/zotero/zoteroPane.js
```diff
@@ -1243,39 +1243,41 @@ var ZoteroPane = new function()
...
+					let settings = JSON.parse(Zotero.Prefs.get('itemsView.columnVisibility') || '{}');
+					// Store previous view settings
+					settings[prevViewGroup] = cols.map((col) => col.getAttribute('hidden') == 'true' ? 0 : 1);
```
dstillman: What if we add a column in the middle? Won't all the column states after be off?

Also, when I first open Zotero on this branch, I'm getting an array with an empty property in the pref:

`{"":[1,1,0,1,1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,1,0]}`

chrome/content/zotero/zoteroPane.xul
```diff
@@ -370,186 +370,187 @@
...
 								<treecols id="zotero-items-columns-header">
 									<treecol
-										id="zotero-items-column-title" primary="true"
+										id="zotero-items-column-title" primary="true" default-in="default feed"
 										label="&zotero.items.title_column;"
 										flex="4" zotero-persist="width ordinal hidden sortActive sortDirection"/>
 									<splitter class="tree-splitter"/>
 									<treecol
-										id="zotero-items-column-firstCreator"
+										id="zotero-items-column-firstCreator" default-in="default"
```
dstillman: You've been looking at more feeds, but I'd guess that enough have creators that it would make sense to show this by default in feeds.

adomasven:
Maybe disable all columns not parsed by feed reader completely?

dstillman:
>  Maybe disable all columns not parsed by feed reader completely?

Yeah, that makes sense.

dstillman:
This seems to reset column settings after a restart. (I haven't switched views away from My Library yet, if that makes a difference, but I did add a column and then restart Firefox and it went back to Title/Creator/Attachment.)

dstillman:
Nope, not working at all for me, even switching to a collection and back.

dstillman:
Not unrelated: this needs tests.

test/tests/zoteroPaneTest.js
```diff
@@ -310,4 +310,50 @@ describe("ZoteroPane", function() {
 			assert.lengthOf(Object.keys(conditions), 2);
 		});
 	});
+	
+	describe("#onCollectionSelected()", function() {
+		var cv;
+		
+		before(function* () {
+			cv = zp.collectionsView;
```
dstillman: The `loadZoteroPane` in the last test will break `cv` for any tests that get added after, so this should be moved to `beforeEach`.

PR #1046, opened 2016-06-22
Change const to var in utilities

zuphilip:
In order to make these mappings also available when the
script is used with [`loadSubScript`](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/mozIJSSubScriptLoader) from outside, we
have to use `var` instead of `const`, `let`.

See https://github.com/aurimasv/z2csl/issues/6


zuphilip:
Thank you!

PR #1047, opened 2016-06-23
Show a hard warning dialog when linking to a different account

adomasven:
When linking to a different account than previously synced with, warns that all data will be lost and requires user to input previous username, to confirm the action. Alternatively allows to select a new data directory location.

This is a change from previous functionality, where a sync attempt with a different account used to result in a merge of the two libraries, but it was decided that the merging behaviour should be independent of sync.


chrome/content/zotero/hardConfirmationDialog.xul
```diff
@@ -0,0 +1,50 @@
...
+
+
+<?xml-stylesheet href="chrome://global/skin/" type="text/css"?>
+
+<!DOCTYPE overlay [ <!ENTITY % zoteroDTD SYSTEM "chrome://zotero/locale/zotero.dtd"> %zoteroDTD; ]>
+
+<dialog xmlns="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul"
+	title="" buttons="cancel,accept"
+	id="zotero-hardConfirmationDialog"
+	width="700"
```
dstillman: 700 is way too wide here.

It would also be nice to make the text in the dialog the size of text in an confirmEx prompt. Right now, at least on OS X, it's much smaller (though the same as other custom pop-ups from the prefs).

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -1052,6 +1052,15 @@ Zotero.Utilities.Internal = {
 		}
 		elem.appendChild(menu);
 		return menu;
+	},
+
+	/**
+	 * Quits zotero, optionally restarting. Mostly useful for stubbing in tests.
```
dstillman: Capitalize "Zotero". Remove "Mostly useful…" — perfectly appropriate to have a central function for quitting/restarting without this being about tests

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -1052,6 +1052,15 @@ Zotero.Utilities.Internal = {
...
+	 * @param restart {Boolean}
```
dstillman: Wrong order, and `restart` is optional: `{Boolean} [restart=false]`

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -507,6 +509,35 @@ Components.utils.import("resource://gre/modules/osfile.jsm");
 			}
 		}
 		
+		// Check for DB purge
+		var newAccountFile = dataDir.clone();
+		newAccountFile.append('new-account');
```
dstillman: Since we'll use this for unlinking too (though that's not implemented yet, it looks like), this isn't really just for `new-account`. Can be `reset-data-directory` or something.

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -507,6 +509,35 @@ Components.utils.import("resource://gre/modules/osfile.jsm");
...
+				
+				newAccountFile.remove(false);
+
+				Zotero.newAccount = true;
+			}
+			catch (e) {
+				// Restore from backup?
+				alert(e);
+			}
+		}
```
dstillman: We should just combine this with the restore-from-server stuff above, which can also delete 'storage'.

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -507,6 +509,35 @@ Components.utils.import("resource://gre/modules/osfile.jsm");
...
+				}
+				
+				var dbfile = Zotero.getZoteroDatabase();
+				if (dbfile.exists()) {
+					dbfile.remove(false);
+				}
+				
+				newAccountFile.remove(false);
+
+				Zotero.newAccount = true;
```
dstillman: This is just about trigger a sync regardless of the auto-sync setting, right? Should make it more explicit than `newAccount`.

test/tests/syncLocalTest.js
```diff
@@ -22,62 +22,108 @@ describe("Zotero.Sync.Data.Local", function() {
...
+			sinon.stub(Zotero, 'forceNewDataDirectory').returns(true);
+			yield Zotero.Users.setCurrentUserID(1);
+			yield Zotero.Users.setCurrentUsername("A");
 			
-			var cont = yield Zotero.Sync.Data.Local.checkUser(null, 1, "A");
+			waitForDialog(null, 'extra1', 'chrome://zotero/content/hardConfirmationDialog.xul');
+			waitForDialog();
+			var cont = yield Zotero.Sync.Data.Local.checkUser(window, 2, "B");
+			var newAccountFile = Zotero.getZoteroDirectory().clone();
+			newAccountFile.append('new-account');
```
dstillman: Here and elsewhere, it's better to use OS.File for anything new instead of nsIFile, since we're more likely to mock OS.File in Electron and, regardless, something based on string paths will probably be easier to adapt. So, e.g., `OS.Path.join(Zotero.getZoteroDirectory().path, 'new-account');` (but not actually `new-account`, per the note above).

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -1188,6 +1219,40 @@ Components.utils.import("resource://gre/modules/osfile.jsm");
 		
 		return useProfileDir ? true : file;
 	}
+	
+	
+	this.forceNewDataDirectory = function() {
+		var win = Services.wm.getMostRecentWindow('navigator:browser');
```
dstillman: Should probably take an optional `win` parameter and use that instead of this if provided, since this is triggered from the pref window, not the main browser window.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -113,13 +113,17 @@ dataDir.notFound						= The Zotero data directory could not be found.
...
 dataDir.changeDataDirectory 		= Change Data Directory…
+dataDir.pickNewDataDirectory		= Pick New Data Directory…
 dataDir.unsafeLocation.selected.dropbox = Choosing a data directory within Dropbox may corrupt your database.
 dataDir.unsafeLocation.selected.useAnyway = Use this directory anyway?
 dataDir.unsafeLocation.existing.dropbox = Your Zotero data directory is within Dropbox, which may lead to data corruption.
 dataDir.unsafeLocation.existing.chooseDifferent = Would you like to choose a different location now?
 dataDir.selectedDirNonEmpty.title		= Directory Not Empty
 dataDir.selectedDirNonEmpty.text		= The directory you selected is not empty and does not appear to be a Zotero data directory.\n\nCreate Zotero files in this directory anyway?
+dataDir.mustSelectEmpty.title			= Directory Not Empty
+dataDir.mustSelectEmpty.text			= The directory you selected is not empty. You need to select an empty directory to continue.
```
dstillman: "The directory you selected is not empty. You must select an empty directory to continue."

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -113,13 +113,17 @@ dataDir.notFound						= The Zotero data directory could not be found.
 dataDir.previousDir					= Previous directory:
 dataDir.useProfileDir				= Use %S profile directory
 dataDir.selectDir					= Select a Zotero data directory
+dataDir.selectNewDir				= Select a new Zotero data directory
 dataDir.changeDataDirectory 		= Change Data Directory…
+dataDir.pickNewDataDirectory		= Pick New Data Directory…
```
dstillman: "Choose New Data Directory…"

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -113,13 +113,17 @@ dataDir.notFound						= The Zotero data directory could not be found.
 dataDir.previousDir					= Previous directory:
 dataDir.useProfileDir				= Use %S profile directory
 dataDir.selectDir					= Select a Zotero data directory
+dataDir.selectNewDir				= Select a new Zotero data directory
```
dstillman: "Zotero" should be a parameter

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -846,13 +850,13 @@ sync.error.emptyResponseServer 		= Empty response from server.
...
-sync.warning.existingDataElsewhere  = If your %S data exists elsewhere on your computer, you should move it to your current data directory or change your data directory to point to the existing data.
+account.unlinkWarning					= Are you sure you want to unlink this account?\n\n%S will no longer sync your data, but your data will remain locally.
+account.warning.emptyLibrary			= You are about to sync the ‘%1$S’ account to an empty %2$S database. This could happen if you removed your previous database or if the location of your %2$S data directory changed.
+account.warning.existingDataElsewhere	= If your %S data exists elsewhere on your computer, you should move it to your current data directory or change your data directory to point to the existing data.
+account.lastSyncWithDifferentAccount	= This Zotero database was last synced with a different %1$S account (‘%2$S’) from the current one (‘%3$S’). If you continue, the data and files belonging to ‘%2$S’ will be removed from your computer and Zotero will restart.
+account.alternativePickNewDataDir		= Alternatively you can click ‘%S’ to choose a new data directory for ‘%S’.
```
dstillman: `account.alternativelyChooseNewDataDir = Alternatively, you can choose a new data directory for ‘%S’.`

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -846,13 +850,13 @@ sync.error.emptyResponseServer 		= Empty response from server.
...
-sync.warning.existingDataElsewhere  = If your %S data exists elsewhere on your computer, you should move it to your current data directory or change your data directory to point to the existing data.
+account.unlinkWarning					= Are you sure you want to unlink this account?\n\n%S will no longer sync your data, but your data will remain locally.
+account.warning.emptyLibrary			= You are about to sync the ‘%1$S’ account to an empty %2$S database. This could happen if you removed your previous database or if the location of your %2$S data directory changed.
+account.warning.existingDataElsewhere	= If your %S data exists elsewhere on your computer, you should move it to your current data directory or change your data directory to point to the existing data.
+account.lastSyncWithDifferentAccount	= This Zotero database was last synced with a different %1$S account (‘%2$S’) from the current one (‘%3$S’). If you continue, the data and files belonging to ‘%2$S’ will be removed from your computer and Zotero will restart.
```
dstillman: `This %1$S database was last synced with a different account (‘%2$S’) from the current one (‘%3$S’). If you continue, data associated with the ‘%2$S’ account will be removed from this computer.`

(I don't think we particularly need to mention the restarting.)

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -846,13 +850,13 @@ sync.error.emptyResponseServer 		= Empty response from server.
...
+account.alternativePickNewDataDir		= Alternatively you can click ‘%S’ to choose a new data directory for ‘%S’.
+account.confirmLastUsername				= To confirm that you want to remove all data and files associated with ‘%1$S’ enter ‘%1$S’ in the textbox below.
```
dstillman: It might make sense to use a phrase rather than the username. E.g., GitHub uses "delete my account".

Maybe: `To confirm that you want to remove all data associated with the ‘%1$S’ account from this computer, enter ‘delete’ in the textbox below.`

Could do something like "delete %1$S data", but that might be too complicated.

The other thing is that there's not really any need for this confirmation if no data has changed locally and all data has synced. We now have the `synced` flag for data. If we add something similar for files — to flag when they've all been uploaded, which might be useful somewhere in the UI in general — we could eliminate this confirmation in those cases.

dstillman:
We should clear WebDAV settings when changing usernames, and probably revert back to Zotero File Storage for My Library (since the new account may very well have ZFS files).

dstillman:
We should use a better button title than "OK". "Switch Accounts" or something.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -846,13 +850,13 @@ sync.error.emptyResponseServer 		= Empty response from server.
...
+account.badConfirmation					= You have to enter precisely ‘%S’ to confirm that you want to continue.
```
dstillman: Instead of a separate dialog, the accept button should just be disabled until the text is entered correctly.

chrome/locale/en-US/zotero/zotero.properties
adomasven: We tell the users every other time when restarting. It could be done in a separate prompt, but that's just excessive.

chrome/locale/en-US/zotero/zotero.properties
adomasven: GitHub uses repository name for repo deletion, hence the decision, but we can change that to 'delete' I guess. No strong opinion.

If they have a 5GB repo locally it still makes sense to prompt them with a hard confirmation. If the user ever wants to switch back to the previous account they wouldn't want to redownload all that data.

dstillman: > GitHub uses repository name for repo deletion, hence the decision, but we can change that to 'delete' I guess.

Yeah, but I think that GitHub example is a bit different — you can have multiple repos, and that's confirming that you're deleting the right one. In this case you're pretty much just confirming that you want to go ahead with this action, and I think entering your username for that is a little weird. I don't love "delete", but I don't have a great alternative other than "delete %1$S data", which seems too error-prone.

chrome/locale/en-US/zotero/zotero.properties
dstillman: It doesn't really affect the user, though, and I don't think it would be particularly surprising or alarming that "Change Account" would cause the window to briefly disappear and reappear.

adomasven: I guess we won't be running in FX anymore, so yeah. Fair point.

chrome/content/zotero/hardConfirmationDialog.xul
adomasven: At least for me on Linux prompts go as far as 819px (as reported by *xwininfo*).
image

If the text doesn't wrap the prompts are the width of maximum line width. So 700 on my end is kind of conservative. Any suggestions on how to resolve this disparity?

chrome/content/zotero/xpcom/zotero.js
adomasven: How do you want to use this for unlinking?

dstillman: When you unlink, it asks if you want to delete your local data (defaulting to no, probably), and if you do it kicks off the same process (but with no username, so forced auto-sync should be a no-op).

adomasven: I don't think we should bother users with this sort of thing. If they want a new library, they'll find a way (probably by switching the data directory). And introduces code complexity, since you'll want to show prompt on manual unlink, but not when the API key expires or similar. I do agree with `new-account` to `reset-data-directory` change though.

dstillman: This isn't for a new library — it's to "log out" of Zotero. We've wanted a "lab mode" like this for years, and it's a pretty big omission right now. People shouldn't need to know anything about the data directory to remove their data from a computer (e.g., a shared computer).

dstillman: (And showing a prompt on a manual click of the Unlink button is hardly code complexity.)

dstillman: I actually suspect deleting local data is the more common case (from people at libraries and labs) than wanting to unlink but keep local data. I'm saying it should be the default only for safety.

chrome/content/zotero/hardConfirmationDialog.js
```diff
@@ -41,17 +41,22 @@ Zotero.HardConfirmationDialog = {
 		if (this.io.extra1Label) {
 			document.documentElement.buttons = document.documentElement.buttons + ',extra1';
 			document.documentElement.getButton('extra1').label = this.io.extra1Label
+		} if (this.io.acceptLabel) {
+			document.documentElement.getButton('accept').label = this.io.acceptLabel
 		}
 		
+		this.onKeyup();
```
dstillman: Let's call it `onKeyUp`

chrome/content/zotero/hardConfirmationDialog.xul
dstillman: Yeah, prompts on OS X are 475px:

modal

Can look at the Mozilla source to see how they handle similar dialogs, but regardless of what we do elsewhere, I don't think super-wide is correct.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -846,13 +850,14 @@ sync.error.emptyResponseServer 		= Empty response from server.
...
-sync.warning.existingDataElsewhere  = If your %S data exists elsewhere on your computer, you should move it to your current data directory or change your data directory to point to the existing data.
+account.unlinkWarning					= Are you sure you want to unlink this account?\n\n%S will no longer sync your data, but your data will remain locally.
+account.warning.emptyLibrary			= You are about to sync the ‘%1$S’ account to an empty %2$S database. This could happen if you removed your previous database or if the location of your %2$S data directory changed.
+account.warning.existingDataElsewhere	= If your %S data exists elsewhere on your computer, you should move it to your current data directory or change your data directory to point to the existing data.
+account.lastSyncWithDifferentAccount	= This %1$S database was last synced with a different account (‘%2$S’) from the current one (‘%3$S’). If you continue, data associated with the ‘%2$S’ account will be removed from this computer.
+account.alternativeChooseNewDataDir		= Alternatively you can click choose a new data directory for ‘%S’.
```
dstillman: Again:

> account.alternativelyChooseNewDataDir = Alternatively, you can choose a new data directory for ‘%S’.

(You should just copy and paste these.)

dstillman:
> TypeError: this.onKeyup is not a function
> hardConfirmationDialog.js:48:3

dstillman:
Text is still too small, at least on OS X:

dialog

If you're not seeing this, I can fix in a follow-up.

adomasven:
> Text is still too small, at least on OS X

Can't see this. Probably some custom css file included in other OS X alert dialogs.

dstillman:
A little hard to tell with the small text, but it looks like there may only be one newline above "Alternatively" in the above? should be two.

dstillman:
Or, maybe better, that line should just be part of the previous string. (I don't think we want to hard-code spaces between separate strings for localization reasons.)

adomasven:
> A little hard to tell with the small text, but it looks like there may only be one newline above "Alternatively" in the above? should be two.

So the way I put the hard confirmation dialog together is that it has a `textboxLabel` and a general `text` attribute. `text` newlines are just new `<label/>` elements, then a `<sep>` and then the label for the textbox. Do you reckon just keep `text` and add `<sep>` every new line?

dstillman:
Much better:

dialog2

chrome/content/zotero/preferences/preferences_sync.js
```diff
@@ -166,11 +167,21 @@ Zotero_Preferences.Sync = {
...
 				null,
 				Zotero.getString('general.warning'),
-				Zotero.getString('account.unlinkWarning', Zotero.clientName)
+				Zotero.getString('account.unlinkWarning', Zotero.clientName),
+				Zotero.getString('account.unlinkWarning.removeData', Zotero.Users.getCurrentUsername()),
+				check
 			)) {
+				if (check.value) {
+					var resetDataDirFile = OS.Path.join(Zotero.getZoteroDirectory().path, 'reset-data-directory');
+					yield OS.File.writeAtomic(resetDataDirFile, '');
```
dstillman: `Zotero.File.putContentsAsync()`?

dstillman:
dialog

Bunch of problems with this:

1) Main text says data will remain, but that wouldn't be true if checkbox was selected. This needs to be rewritten.

2) First-person text is weird. If we keep the checkbox, I think it could just be "Remove local data" with an explanation in the text above. (Alternatively could get rid of the checkbox and have "Unlink" and non-default "Unlink and Remove Local Data", but that might be a bit more awkward.)

3) Button text needs to be specific (e.g., "Unlink"). (We almost never want "OK" except in an alert.)

(The size of the checkbox text is also wrong, obviously, but I can try to fix that.)

adomasven:
> 1) Main text says data will remain, but that wouldn't be true if checkbox was selected. This needs to be rewritten.
> 
> 2) First-person text is weird. If we keep the checkbox, I think it could just be "Remove local data" with an explanation in the text above. (Alternatively could get rid of the checkbox and have "Unlink" and non-default "Unlink and Remove Local Data", but that might be a bit more awkward.)

What about

> Unlinking your account will prevent Zotero from syncing your data. To remove the data from this computer check the checkbox below.
> 
> [ ] Remove local data

And a button that says "Unlink".

> (The size of the checkbox text is also wrong, obviously, but I can try to fix that.)

Using `Services.prompt.confirmCheck()`, so this is FX or OSX idea of how such checkbox should look like (probably intended for "don't ask me again prompts"). Will need custom dialog otherwise.

dstillman:
> Unlinking your account will prevent Zotero from syncing your data. To remove the data from this computer check the checkbox below.
>
>  [ ] Remove local data

That's OK, though I don't love referring to other UI elements in the same dialog.

Maybe this?

> Unlinking your account will prevent Zotero from syncing your data.
>
> [ ] Remove my Zotero data from this computer

(I know I said the previous first-person text was weird, but this [mirrors Firefox's uninstaller](http://s104.photobucket.com/user/ffextensionguru/media/ff-uninstall.png.html).
 
> Using Services.prompt.confirmCheck(), so this is FX or OSX idea of how such checkbox should look like (probably intended for "don't ask me again prompts").

Ah, so it is. Same in another Fx dialog. Never mind.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -850,8 +850,9 @@ sync.error.emptyResponseServer 		= Empty response from server.
 sync.error.invalidCharsFilename		= The filename '%S' contains invalid characters.\n\nRename the file and try again. If you rename the file via the OS, you will need to relink it in Zotero.
 sync.error.apiKeyInvalid			= %S could not authenticate your account. Please re-enter your account details.
 
-account.unlinkWarning					= Are you sure you want to unlink this account?\n\n%S will no longer sync your data, but your data will remain locally.\n
-account.unlinkWarning.removeData		= I want to remove data associated with '%S' account from this computer.
+account.unlinkWarning					= Unlinking your account will prevent %S from syncing your data.
+account.unlinkWarning.removeData		= Remove my %S data from this computer.
```
dstillman: No period.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -856,7 +856,7 @@ account.unlinkWarning.button			= Unlink Account
 account.warning.emptyLibrary			= You are about to sync the ‘%1$S’ account to an empty %2$S database. This could happen if you removed your previous database or if the location of your %2$S data directory changed.
 account.warning.existingDataElsewhere	= If your %S data exists elsewhere on your computer, you should move it to your current data directory or change your data directory to point to the existing data.
 account.lastSyncWithDifferentAccount	= This %1$S database was last synced with a different account (‘%2$S’) from the current one (‘%3$S’). If you continue, data associated with the ‘%2$S’ account will be removed from this computer.
-account.confirmDelete					= I understand that data associated with ‘%S’ will be removed from this computer
+account.confirmDelete					= Remove data associated with ‘%S’ account
```
dstillman: `Remove data associated with the ‘%S’ account`

dstillman:
This still unlinks on Cancel.

dstillman:
"API key not set" error after unlinking

PR #1048, opened 2016-06-23
Fixes freezing occuring on linux while running tests

adomasven:
server_connectorTest - alternating port prevents
exceptions not catchable in JS for httpd server when the socket
and the port remains open after httpd.stop() callback

support - changes the window on which `setTimeout()` is ran in dialogs.
If the timeout is ran on the main window the `dialog` object appears to
lose certain properties and not respond to interactions completely.

### DIFF
--- a/test/tests/server_connectorTest.js
+++ b/test/tests/server_connectorTest.js
@@ -14,10 +14,12 @@ describe("Connector Server", function () {
 		
 		win = yield loadZoteroPane();
 		connectorServerPath = 'http://127.0.0.1:' + Zotero.Prefs.get('httpServer.port');
-		testServerPath = 'http://127.0.0.1:' + testServerPort;
 	});
 	
 	beforeEach(function () {
+		// Alternate ports to prevent exceptions not catchable in JS
+		testServerPort += (testServerPort & 1) ? 1 : -1;
+		testServerPath = 'http://127.0.0.1:' + testServerPort;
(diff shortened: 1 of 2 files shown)


test/tests/server_connectorTest.js
```diff
@@ -14,10 +14,12 @@ describe("Connector Server", function () {
 		
 		win = yield loadZoteroPane();
 		connectorServerPath = 'http://127.0.0.1:' + Zotero.Prefs.get('httpServer.port');
-		testServerPath = 'http://127.0.0.1:' + testServerPort;
 	});
 	
 	beforeEach(function () {
+		// Alternate ports to prevent exceptions not catchable in JS
+		testServerPort += (testServerPort & 1) ? 1 : -1;
```
dstillman: Is this actually reliable? That is, when the port is left open, is it eventually closed, such that reusing it later might work, or does it just remain open, such that this would only work if it happens once? (And wouldn't this be subject to a race condition if the port remained opened long enough for this to be run twice?)

adomasven: I've successfully ran Server Connector tests every time after this change. Before that it was 50/50. And 50% success rate is also good argument for that the ports do get closed eventually - may be a matter of miliseconds. I have not investigated this further and just went for the easy solution that works. And if it doesn't - we can always investigate.

PR #1062, opened 2016-07-12
Allow ZU.doGet to have requestHeaders

zuphilip:
This is mainly for translators which have to work with content negotiation,
e.g. ORCID API or API calls of ISBN agencies.


zuphilip:
I haven't tested this, but the reason for this PR is to have something to discuss. The code is copied from the `doHead` function below.

zuphilip:
Should I be worried about the failed checks by Travis?

zuphilip:
Okay, I updated this PR, tested it and Travis is now also happy. Moreover, I wrote a translator for ORCID using this functionality. @dstillman What do you think?

PR #1074, opened 2016-08-10
Fixes feed sync bugs after conflicts.

adomasven:
Reported https://forums.zotero.org/discussion/61573?page=1#Item_4

SyncedSettings.set() caches values. If an object passed to set() is
modified after the call then get() returns that modified object.

### DIFF
--- a/test/tests/syncedSettingsTest.js
+++ b/test/tests/syncedSettingsTest.js
@@ -1,5 +1,5 @@
 describe('Zotero.SyncedSettings', function() {
-	it('modifying object after setting does not affect cached value', function* () {
+	it('should not affect cached value when modifying the setting after #set() call', function* () {
 		let setting = {athing: 1};
 		yield Zotero.SyncedSettings.set(Zotero.Libraries.userLibraryID, 'setting', setting);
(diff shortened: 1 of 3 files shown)


test/tests/syncedSettingsTest.js
```diff
@@ -0,0 +1,10 @@
+describe('Zotero.SyncedSettings', function() {
+	it('modifying object after setting does not affect cached value', function* () {
+		let setting = {athing: 1};
+		yield Zotero.SyncedSettings.set(Zotero.Libraries.userLibraryID, 'setting', setting);
+		
+		setting.athing = 2;
+		let storedSetting = Zotero.SyncedSettings.get(Zotero.Libraries.userLibraryID, 'setting');
+		assert.notDeepEqual(setting, storedSetting);
+	});
+});
```
dstillman: Test should begin with "should".

```javascript
describe("#set()", function () {
    it("should store a unique copy of a passed object", function* () {
        […]
    }
});
```

dstillman:
Whole bunch of tests are failing with this patch.

PR #1082, opened 2016-08-24
Adds conditions for WebExtensions in shared connector code

adomasven:
These changes are essential in getting the firefox webextension as well as zotero-connector built from 5.0 to work. May as well push them out in the public beta.

All tests pass for me locally. Saving to client from existing connector works fine too.

### DIFF
--- a/chrome/content/zotero/xpcom/connector/cachedTypes.js
+++ b/chrome/content/zotero/xpcom/connector/cachedTypes.js
@@ -84,7 +84,7 @@ Zotero.Connector_Types = new function() {
 					return ZOTERO_CONFIG.BOOKMARKLET_URL+"images/"+icon;
 				} else if(Zotero.isFx) {
 					return "chrome://zotero/skin/"+icon;
-				} else if(Zotero.isWebExtension || Zotero.isChrome) {
+				} else if(Zotero.isBrowserExt || Zotero.isChrome) {
 					return chrome.extension.getURL("images/"+icon);
 				} else if(Zotero.isSafari) {
 					return safari.extension.baseURI+"images/"+icon;
--- a/chrome/content/zotero/xpcom/connector/cachedTypes.js
+++ b/chrome/content/zotero/xpcom/connector/cachedTypes.js
@@ -84,7 +84,7 @@ Zotero.Connector_Types = new function() {
 					return ZOTERO_CONFIG.BOOKMARKLET_URL+"images/"+icon;
 				} else if(Zotero.isFx) {
 					return "chrome://zotero/skin/"+icon;
-				} else if(Zotero.isBrowserExt || Zotero.isChrome) {
+				} else if(Zotero.isBrowserExt) {
 					return chrome.extension.getURL("images/"+icon);
 				} else if(Zotero.isSafari) {
 					return safari.extension.baseURI+"images/"+icon;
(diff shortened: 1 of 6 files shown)


chrome/content/zotero/xpcom/translation/translate.js
```diff
@@ -374,7 +374,10 @@ Zotero.Translate.Sandbox = {
 				
 				var translator = translation.translator[0];
 				translator = typeof translator === "object" ? translator : Zotero.Translators.get(translator);
-				translation._loadTranslator(translator)
+				Zotero.Promise.resolve(translator)
```
dstillman: The idea here is that `Zotero.Translators.get()` in the connector is async? Can you document that?

chrome/content/zotero/xpcom/connector/translator.js
```diff
@@ -303,7 +303,7 @@ Zotero.Translators = new function() {
 		}
 		
 		// Store
-		if(Zotero.isChrome || Zotero.isSafari) {
+		if((Zotero.isWebExtension || Zotero.isChrome) && !Zotero.isFx || Zotero.isSafari) {
```
dstillman: 1) As noted elsewhere, I'm not sure why we need `isWebExtension` at all, and 2) I don't see how the `!Zotero.isFx` would ever be necessary, since `isFx` isn't true in the connector (at least currently — we can reconsider once we stop sharing code).

(Some fragile order of operations stuff here, in any case.)

chrome/content/zotero/xpcom/connector/cachedTypes.js
```diff
@@ -84,7 +84,7 @@ Zotero.Connector_Types = new function() {
 					return ZOTERO_CONFIG.BOOKMARKLET_URL+"images/"+icon;
 				} else if(Zotero.isFx) {
 					return "chrome://zotero/skin/"+icon;
-				} else if(Zotero.isChrome) {
+				} else if(Zotero.isBrowserExt || Zotero.isChrome) {
```
dstillman: Just `isBrowserExt`, which should be true in Chrome.

adomasven:
Right. This should be ready to merge.

chrome/content/zotero/xpcom/translation/translator.js
```diff
@@ -29,10 +29,10 @@ var TRANSLATOR_REQUIRED_PROPERTIES = ["translatorID", "translatorType", "label",
 // Properties that are preserved if present
 var TRANSLATOR_OPTIONAL_PROPERTIES = ["browserSupport", "minVersion", "maxVersion",
                                       "inRepository", "configOptions", "displayOptions",
-                                      "hiddenPrefs"];
+                                      "hiddenPrefs", "itemType"];
```
dstillman: What is this for?

adomasven: I'm not entirely sure what happened to internal representation of translators going from 4.0 to 5.0, but it seems that they used to be represented and passed around as objects and now there's a `Zotero.Translator` class, with a constructor, which limits the types of fields available on construction.

`itemType` is a translator property, which wasn't included and is set [here](https://github.com/adomasven/zotero/blob/905d94c681ceff92c071d7e41a117282bf2042ce/chrome/content/zotero/xpcom/translation/translate.js#L1437-L1437). Translators get passed around between background and injected scripts and they get serialised and deserialised [here](https://github.com/adomasven/zotero-connectors/blob/feature/zotero5-translator-code-support/src/common/messages.js#L78-L103), so this property needs to be supported by `Zotero.Translators.prototype.init()`.

Issue #1021, opened 2016-05-31
Don't trigger translator detection on iframes


dstillman:
It seems like we're doing a huge amount of extra work on advertising iframes. Are there legit cases where we actually need to check iframes? Can we just not? Alternatively, can we allow translators to specify when they need to be run on iframes, only when the target regexp matches (so that we'd do regexp tests on all iframes but not run `detectWeb()` for all generic translators)?

/cc @simonster @zotero/translator-reviewers

zuphilip:
There are legit cases where we have to check iframes during translation, e.g. [LexisNexis](https://github.com/zotero/translators/blob/master/LexisNexis.js). Some more can be found with [this search in the repo](https://github.com/zotero/translators/search?utf8=%E2%9C%93&q=iframe). However, I am not sure, that we need the iframes for detection already or just for during the scraping part of the translators.

dstillman:
OK, yeah, so Lexis is definitely a case where we need to run detection on the iframe. But that's an example where it seems like the translator could indicate that it needs the regexp tested against all iframes, which it could even do with a `rootTarget` property that matched the root URL. We never really worried about regexp performance, but we now have almost 500 translators, and if there can now be dozens of iframes per page, that's a lot of unnecessary tests that might actually affect performance.

But the low-hanging fruit here would be to avoid running detection for the generic translators when in a non-root frame. That would mean we wouldn't detect DOI/COinS/etc. in frames, but that seems like a fair trade-off for avoiding a lot of work on ad-heavy pages.

I think we could do this by setting a flag on Zotero.Translate.Web in browser.js::detectTranslators() when `doc != rootDoc`, and Zotero.Translate.Web.prototype._getTranslatorsGetPotentialTranslators() would pass a `requireTarget` argument to Zotero.Translators.getWebTranslatorsForLocation(), causing it to skip the generic translators.

This isn't totally representative of the gains we'd see, but adding

```javascript
if (rootDoc != doc) {
    Zotero.debug("Ignoring non-root doc");
    return;
}
```

after the first block in `browser.js::contentLoad()` avoids 300ms in translate_firefox.js on some sites on my system (I think from wrapping the document object and such — not sure how performance in the connector would compare).

dstillman:
[...]

PR #1090, opened 2016-09-15
Changes for #1021 to support targetAll translator property.

adomasven:


test/tests/server_connectorTest.js
```diff
@@ -285,4 +285,29 @@ describe("Connector Server", function () {
 			);
 		});
 	});
+	
+	describe('/connector/getTranslatorCode', function() {
+		it('should respond with translator code', function* () {
+			var code = 'function detectWeb() {}\nfunction doImport() {}';
+			var translator = buildDummyTranslator(4, code);
+			sinon.stub(Zotero.Translators, 'get').returns(translator);
```
dstillman: We should test that `Zotero.Translators.get` is getting a request for the right translator.

test/tests/server_connectorTest.js
```diff
@@ -285,4 +285,29 @@ describe("Connector Server", function () {
...
+						translatorID: "dummy-translator",
+					})
+				}
+			);
+			
+			assert.equal(response.response, code);
+			
+			Zotero.Translators.get.restore();
+		})
+	});
```
dstillman: Might as well move this above the other two `describe` blocks to be both alphabetically and chronologically correct.

chrome/content/zotero/xpcom/translation/translators.js
```diff
@@ -46,7 +46,7 @@ Zotero.Translators = new function() {
 		Zotero.debug("Initializing translators");
 		var start = new Date;
 		
-		_cache = {"import":[], "export":[], "web":[], "search":[]};
+		_cache = {"import":[], "export":[], "web":[], "webWithTargetAll":[], "search":[]};
```
adomasven: Suggestions for a better way to do this are welcome.

chrome/content/zotero/xpcom/translation/translators.js
```diff
@@ -267,76 +270,92 @@ Zotero.Translators = new function() {
...
-						
-						if(j === 0) {
-							if(knownProxy) {
-								converterFunctions.push(Zotero.Proxies.properToProxy);
-							} else {
-								converterFunctions.push(null);
+			for (let translator of allTranslators) {
+				translatorLoop:
+				for (let rootSearchURI in rootSearchURIs) {
+					let isGeneric = (!translator.webRegexp.root && translator.runMode === Zotero.Translator.RUN_MODE_IN_BROWSER);
```
dstillman: I don't think `RUN_MODE_IN_BROWSER` is related to generic, so I don't think it should be combined (which also makes the below condition rather confusing). Can use a different variable for the other checks.

PR #1096, opened 2016-09-19
Fix nested translators that rely on child translator tags

adomasven:
Accidentally discovered a bug while testing connector translation functionality

### DIFF
--- a/chrome/content/zotero/xpcom/translation/translate.js
+++ b/chrome/content/zotero/xpcom/translation/translate.js
@@ -127,8 +127,10 @@ Zotero.Translate.Sandbox = {
 				}
 			}
 
-			// Canonicalize tags
-			if(item.tags) item.tags = translate._cleanTags(item.tags);
+			// If we're not in a child translator, canonicalize tags
+			if (!translate._parentTranslator) {
+				if(item.tags) item.tags = translate._cleanTags(item.tags);
+			}
 			
 			// if we're not supposed to save the item or we're in a child translator,
 			// just return the item array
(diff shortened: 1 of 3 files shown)


test/tests/translateTest.js
```diff
@@ -571,6 +572,49 @@ describe("Zotero.Translate", function() {
 			assert.equal(pdf.attachmentLinkMode, Zotero.Attachments.LINK_MODE_IMPORTED_URL);
 			checkTestTags(pdf, true);
 		});
+		
+		it('child translators should not convert tags to canonical form', function* () {
```
dstillman: I know it's not done everywhere above, but let's try to keep these as sentences like they're intended:

`it('should not convert tags to canonical form in child translators')`

chrome/content/zotero/xpcom/translation/translate.js
```diff
@@ -169,13 +173,14 @@ Zotero.Translate.Sandbox = {
...
 						notes.splice(j--, 1);
-					} else if(typeof(note) == "object") {
-						// Canonicalize tags
-						if(note.tags !== undefined) note.tags = translate._cleanTags(note.tags);
-					} else {
+					} else if(typeof(note) != "object") {
 						// Convert to object
 						notes[j] = {"note":note.toString()}
 					}
+					// If we're not in a child translator canonicalize tags
```
dstillman: Here and above, add comma: "If we're not in a child translator, canonicalize tags"

chrome/content/zotero/tools/testTranslators/testTranslators.js
```diff
@@ -549,18 +549,27 @@ function haveTranslators(translators, type) {
...
+			if(translatorTestView.canRun) {
+				translatorTestViewsToRun[type].push(translatorTestView);
+			}
+		}
+		
+		translatorTestStats[type].update();
+		var ev = document.createEvent('HTMLEvents');
+		ev.initEvent('ZoteroHaveTranslators-'+type, true, true);
+		document.dispatchEvent(ev);	
+	});
```
dstillman: This is separate, right?

adomasven: Yeah, I'll put this on a different branch, actually

PR #1111, opened 2016-10-16
Remove occurences of legacy/non-standard/non-strict js syntax

tnajdek:
- Replace legacy syntax `for each...in` with `for...of`
- Replaced non-strict syntax for octal number literals (strict version requires octals to be prefixed with `0o`, non-strict allows just `0`)
- Remove use of non-standard array comprehension syntax (it has [never made it](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Array_comprehensions) to a standard and [is no longer supported by babel](https://babeljs.io/docs/learn-es2015/#comprehensions))
- Remove use of syntax where curly brackets are omitted in function definition
- Remove non-standard syntax for conditional exception handling
- Few other syntax fixes (e.g. yield inside a non-generator function)

### DIFF
--- a/chrome/content/zotero/bindings/itembox.xml
+++ b/chrome/content/zotero/bindings/itembox.xml
@@ -475,7 +475,7 @@
 							
 							var popup = button.appendChild(document.createElement("menupopup"));
 							
-							for each(var v in this._fieldAlternatives[fieldName]) {
+							for (let v of this._fieldAlternatives[fieldName]) {
 								var menuitem = document.createElement("menuitem");
 								var sv = Zotero.Utilities.ellipsize(v, 60);
 								menuitem.setAttribute('label', sv);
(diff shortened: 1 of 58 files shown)


chrome/content/zotero/bindings/itembox.xml
```diff
@@ -475,7 +475,7 @@
 							
 							var popup = button.appendChild(document.createElement("menupopup"));
 							
-							for each(var v in this._fieldAlternatives[fieldName]) {
+							for(let v of this._fieldAlternatives[fieldName]) {
```
dstillman: Let's change all these to have a space after the `for`

dstillman:
Can you rebase to remove that merge commit? (Not sure what happens if the GitHub rebase option is used when that exists.)

chrome/content/zotero/xpcom/data/search.js
```diff
@@ -1148,7 +1148,7 @@ Zotero.Search.prototype._buildQuery = Zotero.Promise.coroutine(function* () {
 						// for the collection/search
 						if (objLibraryID === undefined) {
 							let foundLibraryID = false;
-							for each (let c in this._conditions) {
+							for(let c of Object.values(this._conditions)) {
```
dstillman: Space after `for` for all these too

tnajdek:
@dstillman: Updates to the PR as suggested.

PR #1120, opened 2016-11-09
Add server_connector endpoint to import styles and import translateable resources

adomasven:
Adds an endpoint to import styles and importable resources (BibTeX, RDF, RIS). Part of a huge upcoming update to connectors.

### DIFF
--- a/chrome/content/zotero/xpcom/http.js
+++ b/chrome/content/zotero/xpcom/http.js
@@ -192,6 +192,10 @@ Zotero.HTTP = new function() {
 			if (!headers["Content-Type"]) {
 				headers["Content-Type"] = "application/x-www-form-urlencoded";
 			}
+			if (headers["Content-Type"] == 'multipart/form-data') {
+				// Allow XHR to set Content-Type with boundary for multipart/form-data
+				delete headers["Content-Type"];
+			}
 			
 			if (options.compressBody && this.isWriteMethod(method)) {
 				headers['Content-Encoding'] = 'gzip';
--- a/chrome/content/zotero/xpcom/http.js
+++ b/chrome/content/zotero/xpcom/http.js
@@ -192,7 +192,7 @@ Zotero.HTTP = new function() {
 			if (!headers["Content-Type"]) {
 				headers["Content-Type"] = "application/x-www-form-urlencoded";
 			}
-			if (headers["Content-Type"] == 'multipart/form-data') {
+			else if (headers["Content-Type"] == 'multipart/form-data') {
 				// Allow XHR to set Content-Type with boundary for multipart/form-data
 				delete headers["Content-Type"];
 			}
--- a/chrome/content/zotero/xpcom/server.js
+++ b/chrome/content/zotero/xpcom/server.js
@@ -384,7 +384,9 @@ Zotero.Server.DataListener.prototype._processEndpoint = function(method, postDat
 		if(postData && this.contentType) {
 			// check that endpoint supports contentType
 			var supportedDataTypes = endpoint.supportedDataTypes;
-			if(supportedDataTypes && supportedDataTypes.indexOf(this.contentType) === -1) {
+			if(supportedDataTypes && 
+				(supportedDataTypes == '*' || supportedDataTypes.indexOf(this.contentType) === -1)) {
+				
 				this._requestFinished(this._generateResponse(400, "text/plain", "Endpoint does not support content-type\n"));
 				return;
 			}
--- a/chrome/content/zotero/xpcom/server.js
+++ b/chrome/content/zotero/xpcom/server.js
@@ -389,16 +389,27 @@ Zotero.Server.DataListener.prototype._processEndpoint = function(method, postDat
 				return;
 			}
 			
-			// decode JSON or urlencoded post data, and pass through anything else
-			if(supportedDataTypes && this.contentType === "application/json") {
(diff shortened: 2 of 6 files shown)


chrome/content/zotero/xpcom/server_connector.js
```diff
@@ -594,6 +594,44 @@ Zotero.Server.Connector.Progress.prototype = {
 };
 
 /**
+ * Install a csl style or import a file using import translation
+ * 
+ * Accepts:
+ * 	[resourceURL, resourceType, resourceText]
```
dstillman: I think it'd be better to treat this (and other new endpoints) as somewhat more formal APIs rather than as internal hacks — that would make our own code clearer and also make it easier for others to build tools that use the same endpoints. We'd have to change things to make the `Content-Type` request header available to `init` (instead of using a `resourceType` string), and then we could use an object with clearer named properties for the other two params.

adomasven:
Changed to use `Content-Type: multipart/form-data`.

chrome/content/zotero/xpcom/server_connector.js
```diff
@@ -597,55 +597,55 @@ Zotero.Server.Connector.Progress.prototype = {
...
+ * 	- {name: styleName}
  */
  
 Zotero.Server.Connector.InstallStyle = function() {};
-Zotero.Server.Endpoints["/connector/installStyles"] = Zotero.Server.Connector.InstallStyle;
+Zotero.Server.Endpoints["/connector/importStyle"] = Zotero.Server.Connector.InstallStyle;
 Zotero.Server.Connector.InstallStyle.prototype = {
 	supportedMethods: ["POST"],
-	supportedDataTypes: ["multipart/form-data"],
+	supportedDataTypes: ["text/x-csl", "application/vnd.citationstyles.style+xml"],
```
dstillman: Don't need `text/x-csl`.

Really, we're extremely unlikely to ever support other kinds of style files, so there's not a whole lot of reason to enforce this at all. It made more sense in a shared endpoint, if the type was passed through to the function. But don't have strong feelings here.

Without the second parameter to `Zotero.Styles.install`, what happens in the error messages?

adomasven: >Without the second parameter to Zotero.Styles.install, what happens in the error messages?

We get "null" for style name in those. Not the end of the world, but I guess that gives a legit reason to bother with GET params after all.

dstillman: Yeah, that won't do, since this will become the main way people will install styles.

chrome/content/zotero/xpcom/http.js
```diff
@@ -192,6 +192,10 @@ Zotero.HTTP = new function() {
 			if (!headers["Content-Type"]) {
 				headers["Content-Type"] = "application/x-www-form-urlencoded";
 			}
+			if (headers["Content-Type"] == 'multipart/form-data') {
```
dstillman: `else if`

chrome/content/zotero/xpcom/server.js
```diff
@@ -384,21 +384,34 @@ Zotero.Server.DataListener.prototype._processEndpoint = function(method, postDat
 		if(postData && this.contentType) {
 			// check that endpoint supports contentType
 			var supportedDataTypes = endpoint.supportedDataTypes;
-			if(supportedDataTypes && supportedDataTypes.indexOf(this.contentType) === -1) {
+			if(supportedDataTypes && 
+				(supportedDataTypes == '*' || supportedDataTypes.indexOf(this.contentType) === -1)) {
```
dstillman: Wouldn't this make it so that including a `Content-Type` header for any `supportedDataTypes: '*'` endpoint would result in a 400?

chrome/content/zotero/xpcom/server_connector.js
```diff
@@ -594,6 +594,53 @@ Zotero.Server.Connector.Progress.prototype = {
...
 
 /**
+ * Translates resources using import translators
+ * 	
+ * Returns:
+ * 	- Object[Item] an array of imported items
+ */
+ 
+Zotero.Server.Connector.ImportTranslate = function() {};
+Zotero.Server.Endpoints["/connector/importFile"] = Zotero.Server.Connector.ImportTranslate;
```
dstillman: Name should match the endpoint. Maybe just `/connector/import`/`Connector.Import`, since the GUI function is just "Import…".

chrome/content/zotero/xpcom/server.js
```diff
@@ -460,6 +473,40 @@ Zotero.Server.DataListener.prototype._requestFinished = function(response) {
...
+		let unixHeaderBoundary = field.indexOf("\n\n");
+		let windowsHeaderBoundary = field.indexOf("\r\n\r\n");
+		if (unixHeaderBoundary < windowsHeaderBoundary && unixHeaderBoundary != -1) {
+			fieldData.header = field.slice(0, unixHeaderBoundary);
+			fieldData.body = field.slice(unixHeaderBoundary+2);
+		} else if (windowsHeaderBoundary != -1) {
+			fieldData.header = field.slice(0, windowsHeaderBoundary);
+			fieldData.body = field.slice(windowsHeaderBoundary+4);
+		} else {
+			throw ('Malformed multipart/form-data body');
```
dstillman: `throw new Error`

Issue #1138, opened 2016-12-26
Edit Citation ignores rich-text formatting


dstillman:
Insert a citation, edit it, show editor, make a word bold, click OK. The bold text is not applied to the original citation.

@fbennett, it seems like https://github.com/zotero/zotero/pull/939 broke this — moving `unescapeHTML()` up in that function means the `<strong>`/etc. tags are stripped before they're converted to RTF. I'm not totally clear on what that patch and the one it was following up were doing, so maybe you could take a look at this?

fbennett:
Will do.

KarlHegbloom:
Can you make it so that html, such as <abbr> is stripped prior to sorting,
but left in the returned result? I Wan to put it into abbrev filter
outputs. Also,  some Wikipedia pages get given <i> around parts of the
title, and it affects sort order but should not.

PR #1144, opened 2017-01-04
Preserve format tags and entities in bibliography editor

fbennett:
Addresses #1138, fixing errors in previous commits 55bfe54a (#939) and 9431e0de (#937), which were meant to address #935.
- Apply `unescapeHTML` after RTF tag and entity conversions
- Remove RTF high-bit character escape from entity converter, and apply it separately after remaining HTML entities are decoded by `unescapeHTML`.

@dstillman  I hope this fixes everything around HTML/RTF conversion of formatted text delivered by `citeproc-js`, but give me a shout if anything is still messed up. 

(The moral of the story is that I should read more carefully—I missed that `unescapeHTML` strips HTML tags.)

### DIFF
--- a/chrome/content/zotero/bindings/styled-textbox.xml
+++ b/chrome/content/zotero/bindings/styled-textbox.xml
@@ -84,7 +84,7 @@
 					var _rexData = [
 						[
 							[
-								["<span style=\"font-variant:small-caps;\">"],
+								["<span +style=\"font-variant: *small-caps;\">"],
 								["{\\scaps ", "{\\scaps{}"]
 							],
 							[
@@ -94,7 +94,7 @@
 						],
 						[
 							[
-								["<span style=\"text-decoration:underline;\">"],
+								["<span +style=\"text-decoration: *underline;\">"],
 								["{\\ul{}", "{\\ul "]
 							],
 							[
@@ -164,7 +164,7 @@
 						],
 						[
 							[
-								["<span style=\"font-variant:normal;\">"],
+								["<span +style=\"font-variant: *normal;\">"],
 								["{\\scaps0{}", "{\\scaps0 "]
 							],
 							[
@@ -174,7 +174,7 @@
 						],
 						[
 							[
-								["<span style=\"font-style:normal;\">"],
+								["<span +style=\"font-style: *normal;\">"],
 								["{\\i0{}", "{\\i0 "]
 							],
 							[
@@ -184,7 +184,7 @@
 						],
 						[
 							[
-								["<span style=\"font-weight:normal;\">"],
+								["<span +style=\"font-weight: *normal;\">"],
 								["{\\b0{}", "{\\b0 "]
 							],
 							[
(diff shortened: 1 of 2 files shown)


dstillman:
Bold and italic now work again, but underlining still doesn't appear to be making it back to Word.

Also, not sure if this is related to these patches, but if I click "Hide Editor…" and then "Show Editor…", all formatting is lost. (I can look at that part if it's unrelated.)

fbennett:
Underlining should be working now (yay).

Closing the popup with Hide Editor state is how I've always removed custom formatting. It does seem a little abrupt to revert to the auto-generated cite on Hide/Show in the open popup, but I just checked that Juris-M 4.0 does that too, and I think Zotero is probably the same (?).

fbennett:
A tweak to the popup caches custom formatting when the editor view is toggled. Saving with the editor in hidden state reverts  to the generated citation form.

fbennett:
In the bibliography editor, Revert is not enabled when an entry is edited, unless the popup is closed and reopened. It looks like a listener may be missing.

fbennett:
@dstillman Got those items fixed up, should be ready for review again.

chrome/content/zotero/bindings/styled-textbox.xml
```diff
@@ -339,14 +346,13 @@
...
-					txt = txt.replace(/<span class=\"tab\">&nbsp;<\/span>/g, "\\tab{}");
-					txt = Zotero.Utilities.unescapeHTML(txt);
+					txt = this.convert("htmlRTF", txt);
 					for (var i=0,ilen=this._htmlRTFmap.length; i < ilen; i++) {
 						var entry = this._htmlRTFmap[i];
 						txt = txt.replace(entry[0], entry[1]);
 					}
-					txt = this.convert("htmlRTF", txt);
+					txt = Zotero.Utilities.unescapeHTML(txt);
+					txt = txt.replace(/[\x7F-\uFFFF]/g, function(aChar) { return "\\uc0\\u"+aChar.charCodeAt(0).toString()+"{}"});
```
dstillman: Why does this have to be here, after the `unescapeHTML`? How would `unescapeHTML` affect those escape sequences?

chrome/content/zotero/bindings/styled-textbox.xml
```diff
@@ -204,6 +204,16 @@
 						}
 					}
 					
+					function normalizeRegexpString(str) {
```
dstillman: `normalizeRegExpString` (to be consistent with `new RegExp`)

chrome/content/zotero/bindings/styled-textbox.xml
fbennett: The editor sets high-bit characters (like endash) as HTML named entities, which can't be converted by a generic function. A side effect of `unescapeHTML` is to convert them to character strings, so that they can be RTF-escaped.

dstillman: Ah, OK, got it.

dstillman:
Thanks!

PR #1147, opened 2017-01-06
Fix non-standard syntax

tnajdek:


chrome/content/zotero/charsetMenu.js
```diff
@@ -68,7 +68,7 @@ var Zotero_Charset_Menu = new function() {
 			Components.utils.import("resource://gre/modules/CharsetMenu.jsm");
 			var cmData = CharsetMenu.getData();
 			for (let charsetList of [cmData.pinnedCharsets, cmData.otherCharsets]) {
-				for each(var charsetInfo in charsetList) {
+				for(let charsetInfo in charsetList) {
```
dstillman: Should be `of`, no?

(Also, space after `for`.)

tnajdek:
good catch! I've updated the commit to fix this.

chrome/content/zotero/preferences/preferences_sync.js
```diff
@@ -639,7 +639,7 @@ Zotero_Preferences.Sync = {
 						// TODO: better error handling
 						
 						// Verify username and password
-						var callback = function () {
+						var callback = function* () {
```
adomasven: `Zotero.Promise.coroutine(function()* {` here

tnajdek:
Tweaked as per @adomasven comment above

PR #1148, opened 2017-01-06
Remove tmp-* directories when attachment download fails. Closes #1128

### LINKED ISSUE #1128: Remove 'tmp-*' storage directories on attachment add failure
E.g., when a PDF isn't a PDF and is removed

adomasven:
I can add code to remove all _tmp-*_ dirs to this branch too, but I am not sure whether that's good at all. We shouldn't be orphaning any files/directories to begin with. I'd say we should log a startup error instead if there are any orphaned directories and we'll hopefully catch that eventually, if it does crop up. 

On the other hand, I am not sure whether this is likely at all now.

### DIFF
--- a/chrome/content/zotero/xpcom/attachments.js
+++ b/chrome/content/zotero/xpcom/attachments.js
@@ -326,54 +326,52 @@ Zotero.Attachments = new function(){
 
 			yield deferred.promise;
 			let sample = yield Zotero.File.getSample(tmpFile);
-			if (contentType == 'application/pdf' &&
+			try {
+				if (contentType == 'application/pdf' &&
 					Zotero.MIME.sniffForMIMEType(sample) != 'application/pdf') {
-				let errString = "Downloaded PDF did not have MIME type "
-					+ "'application/pdf' in Attachments.importFromURL()";
-				Zotero.debug(errString, 2);
-				Zotero.debug(sample, 3);
-				throw(new Error(errString));
-			}
-			
-			// Create DB item
-			var attachmentItem;
-			var destDir;
-			yield Zotero.DB.executeTransaction(function* () {
-				// Create a new attachment
-				attachmentItem = new Zotero.Item('attachment');
-				if (libraryID) {
-					attachmentItem.libraryID = libraryID;
+					let errString = "Downloaded PDF did not have MIME type "
+						+ "'application/pdf' in Attachments.importFromURL()";
+					Zotero.debug(errString, 2);
+					Zotero.debug(sample, 3);
+					throw(new Error(errString));
 				}
-				else if (parentItemID) {
-					let {libraryID: parentLibraryID, key: parentKey} =
-						Zotero.Items.getLibraryAndKeyFromID(parentItemID);
-					attachmentItem.libraryID = parentLibraryID;
-				}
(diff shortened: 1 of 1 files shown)


chrome/content/zotero/xpcom/attachments.js
```diff
@@ -369,25 +380,7 @@ Zotero.Attachments = new function(){
...
-					if (destDir && destDir.exists()) {
-						destDir.remove(true);
-					}
-				}
-				catch (e) {
-					Zotero.debug(e, 1);
-				}
-				
-				throw e;
-			});
```
dstillman: I don't think this is right — this way, if something in the transaction fails, stuff won't get deleted.

chrome/content/zotero/xpcom/attachments.js
```diff
@@ -332,6 +332,17 @@ Zotero.Attachments = new function(){
 					+ "'application/pdf' in Attachments.importFromURL()";
 				Zotero.debug(errString, 2);
 				Zotero.debug(sample, 3);
+				try {
+					if (tmpDir && tmpDir.exists()) {
+						tmpDir.remove(true);
+					}
+					if (destDir && destDir.exists()) {
+						destDir.remove(true);
+					}
```
dstillman: `destDir` hasn't yet been set to anything here.

chrome/content/zotero/xpcom/attachments.js
```diff
@@ -332,6 +332,17 @@ Zotero.Attachments = new function(){
...
+				}
+				catch (e) {
+					Zotero.debug(e, 1);
+				}
```
dstillman: Since this only relates to `tmpDir`, and `tmpDir` is passed to `Zotero.Utilities.Internal.saveURI`, I think it might make more sense to do the `tmpDir` clean up in that function.

adomasven: Not entirely sure what you mean here. We can only check the content type of the content after it is saved, so cleanup in `Zotero.Utilities.Internal.saveURI` doesn't really work, unless we do the content type check there (which I don't think we should).

adomasven:
Does the updated version make more sense then?

chrome/content/zotero/xpcom/attachments.js
```diff
@@ -326,54 +326,52 @@ Zotero.Attachments = new function(){
...
-				attachmentItem.parentID = parentItemID;
-				attachmentItem.attachmentLinkMode = Zotero.Attachments.LINK_MODE_IMPORTED_URL;
-				attachmentItem.attachmentContentType = contentType;
-				if (collections) {
-					attachmentItem.setCollections(collections);
+					let errString = "Downloaded PDF did not have MIME type "
+						+ "'application/pdf' in Attachments.importFromURL()";
+					Zotero.debug(errString, 2);
+					Zotero.debug(sample, 3);
+					throw(new Error(errString));
```
dstillman: No parens after `throw`, here or below. (Statement, not function.)

chrome/content/zotero/xpcom/attachments.js
dstillman: Yeah, disregard.

Issue #64, opened 2012-02-21
Non-English month parsing doesn't work in Standalone


dstillman:
http://forums.zotero.org/discussion/16783/#Item_12

gracile-fr:
I just want to point out that Zotero for Firefox being discontinued, this issue has no workaround. And, by the way, I want to say I'm impressed by the amount of work done for 5.0.

sjvudp:
> I just want to point out that Zotero for Firefox being discontinued, this issue has no workaround.

I'm very much confused about the meaning of that: The current "standalone" is based on Firefox, so will there be a completely different standalone program, or will the Firefox plugin be discontinued? And what are the reasons?

gracile-fr:
The _full_ Zotero for Firefox is going to be discontinued.
Standalone is indeed based on Firefox (xul) but that will change in the future, due to <a href="https://blog.mozilla.org/addons/2015/08/21/the-future-of-developing-firefox-add-ons/">Mozilla decision</a> to deprecate the use of XUL to create add-ons.
https://groups.google.com/forum/#!msg/zotero-dev/yy4-q_ZUA4M/UBOP9OHMEAAJ
https://groups.google.com/forum/#!msg/zotero-dev/20KD_Sy6FkM/W19IYhlSFwAJ

gracile-fr:
Thank you very much Dan !

PR #1156, opened 2017-01-16
Amend short form of French months (continuing #64)

gracile-fr:
The Firefox French language pack contains some mistakes regarding the short form of months. I don't know where Mozilla found this list… Anyway French month parsing didn't work (for juin, juillet, décembre).

Actually these values aren't even the correct abbreviations but only the three or four first letters of the correct abbreviations. See the <a href="https://github.com/citation-style-language/locales/blob/master/locales-fr-FR.xml#L298">French CSL locale</a> for the correct abbreviations.

There's still a small glitch: when you put your mouse over the date field, if it contains "juill." "déc.", the tooltip shows 00 for the month, even if the parsing is correct (and the parsing indicator says d m y).

### DIFF
--- a/resource/schema/dateFormats.json
+++ b/resource/schema/dateFormats.json
@@ -306,13 +306,13 @@
 			"mar",
 			"avr",
 			"mai",
-			"jun",
-			"jul",
+			"juin",
+			"juil",
 			"aoû",
 			"sep",
 			"oct",
 			"nov",
-			"dec"
+			"déc"
 		],
 		"long": [
 			"janvier",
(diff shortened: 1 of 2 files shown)


dstillman:
> There's still a small glitch: when you put your mouse over the date field, if it contains "juill." "déc.", the tooltip shows 00 for the month, even if the parsing is correct (and the parsing indicator says d m y).

I'm not seeing that. If I put "14 déc. 2014" for a date, I see "d m y" and the tooltip shows "2014-12-14".

PR #1158, opened 2017-01-18
Limits the size of maximum snapshot file to 8MB

adomasven:
Note that this does not limit the maximum size the snapshot will use on
disk and only cancels individual files. This also technically orphans the files
but getting the correct filenames to delete is non-trivial (WBP seems to remove
'.min' from minified filenames for example) and snapshot directories are a general
mess not to be looked at anyway.

See https://forums.zotero.org/discussion/comment/266997 for the original report


adomasven:
This got closed by me somehow by accident. Reopening.

adomasven:
Irrelevant since SingleFile https://github.com/zotero/zotero/pull/1846

PR #1166, opened 2017-02-03
Fix localization of "Preferences" in menu

gracile-fr:
Regression 5dbe677f251bc6114ab926698e844aaeeaec4293
Fix accesskey (wasn't done in 23fc663378a0adc770b91adc135ac57ddcbddb93 )

https://forums.zotero.org/discussion/64243/5-0-beta-menu-localization-kb-shortcuts


gracile-fr:
(and could you explain to me why there's this second commit (merge remote-tracking branch) and how can I avoid that - indeed after my commit and PR, I did git fetch upstream / git merge upstream master / git push origin master - not good? Thanks!)

dstillman:
This is already localized — the word is just "Options" on Windows. It's only "Préférences" on Mac and Linux.

https://github.com/zotero/zotero/blob/55e9a0ee3508b1c0099c92f15f8b540d2768d05e/chrome/locale/fr-FR/zotero/mozilla/browser.dtd

Should be the same in Firefox, since we're using their locale files for this.

(For the merge commit, you want to rebase against master, not merge.)

gracile-fr:
Ok, it's not a localization issue indeed :-)
In Zotero 4.0.29.17 on Windows, it's "Tools->Preferences". I know it used to be "Options" on Windows some time ago, but I think that it should be "Preferences" for consistency: that's the way we've always called this pane in the documentation or the forums. Why sticking to Firefox conventions about naming while Zotero will not rely on XUL in the near future?

dstillman:
I'm open to this, and I agree that it would make documentation and the forums easier. I assume Firefox has its reasons for calling it Options, though, presumably for consistency with other Windows programs. I haven't looked around, but is there precedent for "Preferences" in other Windows programs?

gracile-fr:
I've checked in French, I assume the translations were literal…
_Options_: MS-Word (and MS-Office Suite), Omnipage.
_Preferences_: Adobe Acrobat Pro, Notepad++, Gimp, VLC media player.
_Both_: SumatraPDF (Preferences->Options)

(and for the merge commit, I remember it's not appropriate to rebase or revert once something is shared with others, but as this PR is unsuccessful, can I revert those commits? I should have created a new branch to avoid that…)

dstillman:
OK, so presumably Firefox is following Office here, but it seems "Preferences" wouldn't be too out of place here, so I'm OK switching to that. I'll take care of that.

> and for the merge commit, I remember it's not appropriate to rebase or revert once something is shared with others, but as this PR is unsuccessful, can I revert those commits? I should have created a new branch to avoid that…

Yeah, that doesn't matter. You can do a `git reset --hard upstream/master` to revert to the latest official code.

gracile-fr:
> Yeah, that doesn't matter. You can do a git reset --hard upstream/master to revert to the latest official code.

Thank you!

Issue #1167, opened 2017-02-06
Use "Preferences" instead of "Options" on Windows


dstillman:
https://github.com/zotero/zotero/pull/1166#issuecomment-277752927

dstillman:
And [move to Edit menu](https://forums.zotero.org/discussion/66458/preferences-under-edit-linux-tools-windows), probably

PR #1168, opened 2017-02-06
Add shortcut key for "Open Preferences" (Ctrl/Cmd-Shift-P)

gracile-fr:
Cf. https://forums.zotero.org/discussion/64243/5-0-beta-menu-localization-kb-shortcuts

### DIFF
--- a/chrome/content/zotero/preferences/preferences_advanced.xul
+++ b/chrome/content/zotero/preferences/preferences_advanced.xul
@@ -54,6 +54,7 @@
 			<preference id="pref-keys-toggleAllRead" name="extensions.zotero.keys.toggleAllRead" type="string"/>
 			<preference id="pref-keys-copySelectedItemCitationsToClipboard" name="extensions.zotero.keys.copySelectedItemCitationsToClipboard" type="string"/>
 			<preference id="pref-keys-copySelectedItemsToClipboard" name="extensions.zotero.keys.copySelectedItemsToClipboard" type="string"/>
+			<preference id="pref-keys-openPreferences" name="extensions.zotero.keys.openPreferences" type="string"/>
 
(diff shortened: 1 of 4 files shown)


gracile-fr:
(zotero.preferences.keys.openPreferences is a duplicate of general.openPreferences from<a href="https://github.com/zotero/zotero/blob/master/chrome/locale/en-US/zotero/zotero.properties#L55">zotero.properties</a> but I'm not sure properties files can be used in a XUL file.)

dstillman:
This shouldn't appear in or affect macOS. `Cmd-;` is the universal shortcut for opening prefs on macOS and already works.

Not sure if there's an equivalent on Windows or Linux. I'm disinclined to use Ctrl-Shift-P because it's the Command Palette in at least Sublime and Atom, but Ctrl-P might be OK if we're dropping Shift anyway as discussed in the other ticket. But we should see what other apps on those platforms use.

(And we can probably stop making these configurable, at least via the prefs window, so probably don't need to add this to preferences_advanced.xul, which will make the Mac part easier.)

gracile-fr:
Ok, thanks for your comment.

gracile-fr:
Ok, thanks for your comment.
<blockquote>what other apps on those platforms use.</blockquote>"Platforms" refers to "Linux and Windows" or to "Sublime and Atom" (i.e. Electron I'd assume)?
If the former, on Windows:
<br>MS-Word: <code>Alt-F, O</code> (but that's not really a shortcut)
<br>Adobe Acrobat Pro: <code>Ctrl-K</code>
<br>VLC media player: <code>Ctrl-P</code>
<br>iTunes: <code>Ctrl-,</code>

Cf. https://github.com/zotero/zotero/issues/1167

PR #1172, opened 2017-02-11
Get a better last name if current guess starts with weird symbols

adomasven:
Addresses https://forums.zotero.org/discussion/64352/5-0-beta-patent-rss-feeds

Not sure if this is the best way to handle it, but looking at `Utilities.cleanAuthors()`, I would question most of that code.

### DIFF
--- a/chrome/content/zotero/xpcom/utilities.js
+++ b/chrome/content/zotero/xpcom/utilities.js
@@ -179,9 +179,13 @@ Zotero.Utilities = {
 				var lastName = author;
 			}
 		} else {
-			var spaceIndex = author.lastIndexOf(" ");
-			var lastName = author.substring(spaceIndex+1);
-			var firstName = author.substring(0, spaceIndex);
+			// Don't parse "Firstname Lastname [Country]" as "[Country], Firstname Lastname"
+			var spaceIndex = author.length;
+			do {
+				spaceIndex = author.lastIndexOf(" ", spaceIndex-1);
+				var lastName = author.substring(spaceIndex + 1);
+				var firstName = author.substring(0, spaceIndex);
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/xpcom/utilities.js
```diff
@@ -179,9 +179,13 @@ Zotero.Utilities = {
...
-			var spaceIndex = author.lastIndexOf(" ");
-			var lastName = author.substring(spaceIndex+1);
-			var firstName = author.substring(0, spaceIndex);
+			// Don't parse "Firstname Lastname [Country]" as "[Country], Firstname Lastname"
+			var spaceIndex = author.length;
+			do {
+				spaceIndex = author.lastIndexOf(" ", spaceIndex-1);
+				var lastName = author.substring(spaceIndex + 1);
+				var firstName = author.substring(0, spaceIndex);
+			} while (!/\w/.test(lastName[0]) && spaceIndex > 0)
```
dstillman: 1) Any changes here would need supporting tests.

2) `\w` and the like aren't really usable for user input (as surely you would agree). This would fail on any family name that started with an extended character. XRegExp and its much more extensive character classes, including one for punctuation, are available as `this.XRegExp` in `Zotero.Utilities`, though.

adomasven: > `\w` and the like aren't really usable for user input (as surely you would agree)

I do agree. I was looking for a better solution that includes extended characters, but couldn't find anything obvious online and there is also code just above, which has a regexp

```
var allCaps = 'A-Z' + 
 	'\u0400-\u042f';		//cyrilic
```

that made go for a less-than-ideal solution.

adomasven:
Updated

PR #1180, opened 2017-02-21
Default to first translator for /connector/savePage if not provided

adomasven:
Addresses https://groups.google.com/forum/#!topic/zotero-dev/QCCcoSTHE70


dstillman:
I changed this to use a generic webpage with a COinS tag. We can extend to a larger page as necessary.

PR #1187, opened 2017-02-27
Probable typo in dataObject.js

fbennett:
This line just crashed on me during a sync operation in Juris-M 5.0. It looks like a typo.


dstillman:
Thanks!

PR #1198, opened 2017-03-16
Replaced the 9 tag colours with 15 new ones based on the http://clrs.cc

janbaykara:
The nine colours on offer to users are dull and restrictive. I've replaced them with fifteen colours, based on the http://clrs.cc web library.

### Before
screen shot 2017-03-16 at 01 16 30

### After
screen shot 2017-03-16 at 01 16 47

screen shot 2017-03-16 at 00 33 05


bwiernik:
These mostly look nice, but I'm colorblind (protanomaly) and colors 6-8 are really hard to tell apart.

janbaykara:
Should just have a colourpicker tbh, lets people personalise their tags without stepping on others' specific needs.

dstillman:
New colors were added in 081f6bc77d775d97fbbb2f456628c99072fcb27d

PR #1199, opened 2017-03-16
[WIP] Set displayed date format to uniform format

holocronweaver:
This is a WIP pull request related to [this feature request](https://forums.zotero.org/discussion/14572/a-ui-proposal-for-a-better-date-field/p2).

The item date shown in the item tree view and item box now use a uniform format. 

Currently the format is always the ISO date because this is the format I prefer, but a final solution would probably create a user setting to toggle between ISO or locale dates with locale being default. Alternatively users could choose a custom date format. I am fine with any solution that allows ISO dates.

### DIFF
--- a/chrome/content/zotero/bindings/itembox.xml
+++ b/chrome/content/zotero/bindings/itembox.xml
@@ -1296,6 +1296,13 @@
 						case 'dateDue':
 						case 'accepted':
 							if (fieldName == 'date' && this.item._objectType != 'feedItem') {
+								if (valueText) {
+									var date = Zotero.Date.strToDate(valueText);
+									if (date.part == undefined) {
+										// valueText = Zotero.Date.isoToDate(date).toLocaleDateString();
+										valueText = Zotero.Date.dateToISO(date);
+									}
+								}
 								break;
 							}
 							if (valueText) {
--- a/chrome/content/zotero/xpcom/date.js
+++ b/chrome/content/zotero/xpcom/date.js
@@ -623,14 +623,12 @@ Zotero.Date = new function(){
 		return string;
 	}
 	
-	this.strToISO = function (str) {
-		var date = this.strToDate(str);
-		
-		if(date.year) {
+	this.dateToISO = function (date) {
+		if (date.year) {
 			var dateString = Zotero.Utilities.lpad(date.year, "0", 4);
 			if (parseInt(date.month) == date.month) {
 				dateString += "-"+Zotero.Utilities.lpad(date.month+1, "0", 2);
-				if(date.day) {
+				if (date.day) {
 					dateString += "-"+Zotero.Utilities.lpad(date.day, "0", 2);
 				}
 			}
@@ -639,6 +637,11 @@ Zotero.Date = new function(){
 		return false;
 	}
 	
+	this.strToISO = function (str) {
+		var date = this.strToDate(str);
+		var dateString = this.dateToISO(date);
+		return dateString;
+	}
 	
 	this.sqlToISO8601 = function (sqlDate) {
 		var date = sqlDate.substr(0, 10);
--- a/chrome/content/zotero/xpcom/itemTreeView.js
+++ b/chrome/content/zotero/xpcom/itemTreeView.js
@@ -1096,6 +1096,13 @@ Zotero.ItemTreeView.prototype.getCellText = function (row, column)
 		case 'zotero-items-column-accessDate':
 		case 'zotero-items-column-date':
 			if (column.id == 'zotero-items-column-date' && !this.collectionTreeRow.isFeed()) {
+				if (val) {
+					var date = Zotero.Date.strToDate(val);
+					if (date.part == undefined) {
+						// val = Zotero.Date.isoToDate(date).toLocaleDateString();
+						val = Zotero.Date.dateToISO(date);
+					}
+				}
 				break;
 			}
 			if (val) {


dstillman:
Setting aside the format issue, if I'm reading this correctly, it would strip any unparsed components from displayed dates. As noted a few times in the linked thread, that's not what we'd want — the primary reason this issue exists in the first place is because we don't want to throw out data we don't understand.

If you look at Mark's proposal in that thread, he suggests that fully parseable dates be displayed uniformly but dates with additional elements be left alone (at least for now). I think that's a good compromise, and it'll then give us a clearer picture of what additional parsing issues need to be addressed.

`strToDate` includes a `parts` property that I believe should provide a mechanism for determining whether dates are fully parsed.

dstillman:
(And thanks for working on this! To be clear, I do hate that we show non-uniform dates for unambiguously equivalent dates, so it'll be great to fix this.)

holocronweaver:
Okydoke, I updated my pull request so that if `date.part` from `strToDate` is defined, then the date will be displayed as is, otherwise the ISO date will be used. This leaves any date that contains data beyond year, month, and day in a non-uniform state. I suppose either `strToDate` could be expanded or a new function could be added which starts with `strToDate` then further parses the `date.part` to take care of these edge cases.

I tested it on my DB of 10k+ items and it seems to work as intended. Not sure why those couple CI tests failed.

holocronweaver:
@dstillman When you get a chance, could you take a look at why those two CI tests are failing? I looked at their code and I don't see how my changes could affect them.

chrome/content/zotero/bindings/itembox.xml
```diff
@@ -1296,6 +1296,13 @@
 						case 'dateDue':
 						case 'accepted':
 							if (fieldName == 'date' && this.item._objectType != 'feedItem') {
+								if (valueText) {
+									var date = Zotero.Date.strToDate(valueText);
```
dstillman: `var` -> `let`

chrome/content/zotero/xpcom/itemTreeView.js
```diff
@@ -1096,6 +1096,13 @@ Zotero.ItemTreeView.prototype.getCellText = function (row, column)
 		case 'zotero-items-column-accessDate':
 		case 'zotero-items-column-date':
 			if (column.id == 'zotero-items-column-date' && !this.collectionTreeRow.isFeed()) {
+				if (val) {
+					var date = Zotero.Date.strToDate(val);
```
dstillman: `var` -> `let`

chrome/content/zotero/xpcom/date.js
```diff
@@ -623,14 +623,12 @@ Zotero.Date = new function(){
 		return string;
 	}
 	
-	this.strToISO = function (str) {
-		var date = this.strToDate(str);
-		
-		if(date.year) {
+	this.dateToISO = function (date) {
```
dstillman: It would be good to avoid adding a `dateTo` function that's not referring to JS `Date` — we have `strToDate` and `formatDate` that refer to our weird custom object, but the other `dateTo*` functions take `Date` objects. Maybe `datePartsToISO`?

chrome/content/zotero/bindings/itembox.xml
```diff
@@ -1296,6 +1296,13 @@
...
+									if (date.part == undefined) {
+										// valueText = Zotero.Date.isoToDate(date).toLocaleDateString();
+										valueText = Zotero.Date.dateToISO(date);
```
dstillman: OK, so the biggest issue here is that we can't merge this with ISO dates, at least as the default. I still believe what I said in the thread, which is that 8601 is a computer format that would be foreign to most people, and it'd also be inconsistent with the existing localized Date Added/Modified fields in the itembox and items list:

date-columns

(Going from, say, `Jan. 2008` to `2008-01` in the itembox would also be a significant decrease in readability for most people, which is arguably worse than the inconsistency with other items.)

So for this patch, we should output a locale date string, and we can deal with how we'd allow using ISO dates in a separate issue (and if we did that, we'd want it to affect Date Added/Modified too). Using `toLocaleDateString()` I think also means we need to additionally limit this to when values exist for `year`, `month`, and `day`.

I know your priority is ISO dates, so if you don't have time to finish this at the moment to show locale dates, I can take it from here.

dstillman (changes requested):

chrome/content/zotero/xpcom/date.js
dstillman: Oh, and the tests are failing because `Zotero.Date.dateToISO` actually already exists, expecting a JS Date, and you overrode it, which is causing some code to get `false` instead.

holocronweaver:
Thanks Dan. I will try to take care of this over the weekend. 

I think what I will do is make your suggested changes, including switching to localized dates, and create a separate branch to add an option for ISO dates. I would indeed be pretty bummed if I went through all the bother only to end up with localized dates again. X-D

BTW, the reason I prefer ISO dates is they are in order of importance: year usually is all I care about, and in certain cases month might be relevant, and day rarely matters. I can scan numbered months faster than their names. To me that is the most human-friendly approach, but I understand that not everyone agrees which is why making it an option seems like the natural approach.

retorquere:
BBT's dateparser is available for inclusion/relicensing/whatnot should that be helpful. It handles quite a few edge cases in date parsing but only parses when it has a positive date match on the full field.

holocronweaver:
Ah, completely forgot about this pull request. Will try to finish it up over Thanksgiving.

dstillman:
https://forums.zotero.org/discussion/75455/feature-suggestion-iso-8601-date-format-in-column-view#latest

dstillman:
https://forums.zotero.org/discussion/91699/inconsisten-way-of-displaying-dates-when-they-are-written-exactly-the-same

PR #1202, opened 2017-03-21
Fix document preferences dialog failing when styles unloaded.

### LINKED ISSUE #685: Display warning and (if possible) offer to install style when using document with missing style
https://forums.zotero.org/discussion/47635/

### LINKED ISSUE #1084: 5.0: Wait for style load when showing Create Bibliography window
Progress indicator in dialog?

Regardless, don't fail with a "Styles not yet loaded" exception

### LINKED ISSUE #1162: On doc prefs window error, don't set document properties
Most noticeable now because the window fails with styles not loaded (which we should also fix, along with #1084).

Possible this happens on a regular cancel as well — I didn't check.

adomasven:
Closes #1084

~I'm not really sure what #1162 is about, but I'd like to address it on this same branch. I could not reproduce document prefs being stored on failure.~

Edit:
Closes #1162 
Closes #685

### DIFF
--- a/chrome/content/zotero/xpcom/integration.js
+++ b/chrome/content/zotero/xpcom/integration.js
@@ -40,6 +40,19 @@ const FORCE_CITATIONS_RESET_TEXT = 2;
 // this is used only for update checking
 const INTEGRATION_PLUGINS = ["zoteroMacWordIntegration@zotero.org",
 	"zoteroOpenOfficeIntegration@zotero.org", "zoteroWinWordIntegration@zotero.org"];
+	
+// These must match the constants defined in zoteroIntegration.idl 
+const DIALOG_ICON_STOP = 0;
+const DIALOG_ICON_WARNING = 1;
+const DIALOG_ICON_CAUTION = 2;
+
+const DIALOG_BUTTONS_OK = 0;
+const DIALOG_BUTTONS_OK_CANCEL = 1;
+const DIALOG_BUTTONS_YES_NO = 2;
+const DIALOG_BUTTONS_YES_NO_CANCEL = 3;
+
+const NOTE_FOOTNOTE = 1;
+const NOTE_ENDNOTE = 2;
 
 Zotero.Integration = new function() {
 	Components.utils.import("resource://gre/modules/Services.jsm");
@@ -210,6 +223,19 @@ Zotero.Integration = new function() {
 		};
 	}
 	
+	this.getApplication = function(agent, command, docId) {
+		// Try to load the appropriate Zotero component; otherwise display an error
+		try {
+			var componentClass = "@zotero.org/Zotero/integration/application?agent="+agent+";1";
+			Zotero.debug("Integration: Instantiating "+componentClass+" for command "+command+(docId ? " with doc "+docId : ""));
+			return Components.classes[componentClass]
+				.getService(Components.interfaces.zoteroIntegrationApplication);
+		} catch(e) {
+			throw new Zotero.Exception.Alert("integration.error.notInstalled",
+				[], "integration.error.title");
+		}	
+	},
+	
 	/**
 	 * Executes an integration command, first checking to make sure that versions are compatible
 	 */
@@ -230,17 +256,8 @@ Zotero.Integration = new function() {
 			inProgress = true;
 			
 			// Check integration component versions
-			_checkPluginVersions().then(function() {		
-				// Try to load the appropriate Zotero component; otherwise display an error
-				try {
-					var componentClass = "@zotero.org/Zotero/integration/application?agent="+agent+";1";
-					Zotero.debug("Integration: Instantiating "+componentClass+" for command "+command+(docId ? " with doc "+docId : ""));
-					var application = Components.classes[componentClass]
-						.getService(Components.interfaces.zoteroIntegrationApplication);
-				} catch(e) {
-					throw new Zotero.Exception.Alert("integration.error.notInstalled",
-						[], "integration.error.title");
-				}
+			return _checkPluginVersions().then(function() {		
+				var application = Zotero.Integration.getApplication(agent, command, docId);
 				
 				// Try to execute the command; otherwise display an error in alert service or word processor
 				// (depending on what is possible)
@@ -267,9 +284,7 @@ Zotero.Integration = new function() {
 							if(document) {
 								try {
 									document.activate();
-									document.displayAlert(displayError,
-											Components.interfaces.zoteroIntegrationDocument.DIALOG_ICON_STOP,
-											Components.interfaces.zoteroIntegrationDocument.DIALOG_BUTTONS_OK);
+									document.displayAlert(displayError, DIALOG_ICON_STOP, DIALOG_BUTTONS_OK);
 								} catch(e) {
 									showErrorInFirefox = true;
 								}
@@ -821,10 +836,7 @@ Zotero.Integration.CorruptFieldException.prototype = {
 			field = this.fieldGetter._fields[this.fieldIndex];
 		field.select();
 		this.fieldGetter._doc.activate();
-		var result = this.fieldGetter._doc.displayAlert(msg,
-			Components.interfaces.zoteroIntegrationDocument.DIALOG_ICON_CAUTION, 
-			Components.interfaces.zoteroIntegrationDocument.DIALOG_BUTTONS_YES_NO_CANCEL);
-		
+		var result = this.fieldGetter._doc.displayAlert(msg, DIALOG_ICON_CAUTION, DIALOG_BUTTONS_YES_NO_CANCEL); 
 		if(result == 0) {
 			return Zotero.Promise.reject(new Zotero.Exception.UserCancelled("document update"));
 		} else if(result == 1) {		// No
@@ -878,9 +890,7 @@ Zotero.Integration.CorruptBibliographyException.prototype = {
 		
 		var msg = Zotero.getString("integration.corruptBibliography")+'\n\n'+
 				  Zotero.getString('integration.corruptBibliography.description');
-		var result = this.fieldGetter._doc.displayAlert(msg, 
-					Components.interfaces.zoteroIntegrationDocument.DIALOG_ICON_CAUTION, 
-					Components.interfaces.zoteroIntegrationDocument.DIALOG_BUTTONS_OK_CANCEL);
+		var result = this.fieldGetter._doc.displayAlert(msg, DIALOG_ICON_CAUTION, DIALOG_BUTTONS_OK_CANCEL);
 		if(result == 0) {
 			return Zotero.Promise.reject(new Zotero.Exception.UserCancelled("clearing corrupted bibliography"));
 		} else {
@@ -991,8 +1001,7 @@ Zotero.Integration.Document.prototype._getSession = Zotero.Promise.coroutine(fun
 			}
 			
 			var warning = this._doc.displayAlert(Zotero.getString("integration.upgradeWarning"),
-				Components.interfaces.zoteroIntegrationDocument.DIALOG_ICON_WARNING,
-				Components.interfaces.zoteroIntegrationDocument.DIALOG_BUTTONS_OK_CANCEL);
+				DIALOG_ICON_WARNING, DIALOG_BUTTONS_OK_CANCEL);
 			if(!warning) {
 				return Zotero.Promise.reject(new Zotero.Exception.UserCancelled("document upgrade"));
 			}
@@ -1007,9 +1016,11 @@ Zotero.Integration.Document.prototype._getSession = Zotero.Promise.coroutine(fun
 					[], "integration.error.title"));
 		}
 		
-		if(Zotero.Integration.sessions[data.sessionID]) {
+		if (Zotero.Integration.sessions[data.sessionID]) {
+			// If communication occured with this document since restart
 			this._session = Zotero.Integration.sessions[data.sessionID];
 		} else {
+			// Document has zotero data, but has not communicated since Zotero restart
 			this._session = this._createNewSession(data);
 			try {
 				yield this._session.setData(data);
@@ -1027,7 +1038,6 @@ Zotero.Integration.Document.prototype._getSession = Zotero.Promise.coroutine(fun
 				}
 			}
 			
-			this._doc.setDocumentData(this._session.data.serializeXML());
 			this._session.reload = true;
 		}
 		return Zotero.Promise.resolve(this._session);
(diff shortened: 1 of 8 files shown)


adomasven:
I've also pushed some code which will fetch the style from repo automatically. I can submit as separate PR as there might be security implications? Embed a rogue url address in the doc style id and get zotero to fetch it? Since we don't execute any JS that seems harmless enough to me. And the improvement in functionality is huge.

chrome/content/zotero/xpcom/style.js
```diff
@@ -190,7 +190,26 @@ Zotero.Styles = new function() {
...
+				let styleText = yield Zotero.File.getContentsFromURLAsync(id, {timeout: 5000});
+				yield Zotero.Styles.install(styleText, id, true);
+				style = this.get(id, skipMappings);
+			} catch (e) {
+				Zotero.debug(`Failed to retrieve style ${id}`);
+				Zotero.debug(e);
+			}	
+		}
+		return style;
+	});
```
dstillman: I don't think this should be part of a `getAsync()`. There are lots of things that are `*Async` in Zotero now, and they don't generally kick off HTTP requests. So I think we should make this more explicit, with a separate function (`fetch` or `download` or something) that explicitly makes a request if a `get()` fails. (I also don't think this is likely to be generally necessary functionality — it's pretty specific to the case of an embedded style in a stored document.)

chrome/content/zotero/xpcom/style.js
```diff
@@ -190,7 +190,26 @@ Zotero.Styles = new function() {
...
+		if (!style && download) {
+			try {
+				// Don't wait too long
+				Zotero.debug(`Fetching required style ${id} asynchronously`);
+				let styleText = yield Zotero.File.getContentsFromURLAsync(id, {timeout: 5000});
+				yield Zotero.Styles.install(styleText, id, true);
+				style = this.get(id, skipMappings);
+			} catch (e) {
+				Zotero.debug(`Failed to retrieve style ${id}`);
+				Zotero.debug(e);
```
dstillman: Can be `Zotero.logError(e)`

chrome/content/zotero/xpcom/style.js
```diff
@@ -190,7 +190,26 @@ Zotero.Styles = new function() {
...
+	
+	this.getAsync = Zotero.Promise.coroutine(function* (id, skipMappings, download=true) {
+		yield this.init();
+		var style = this.get(id, skipMappings);
+		
+		if (!style && download) {
+			try {
+				// Don't wait too long
+				Zotero.debug(`Fetching required style ${id} asynchronously`);
+				let styleText = yield Zotero.File.getContentsFromURLAsync(id, {timeout: 5000});
```
dstillman: We <a href="http://docs.citationstyles.org/en/stable/specification.html#info">should be</a> checking for a `rel="self"` first, falling back to the `id`. (The whole id situation in CSL is non-ideal — the `id` would ideally be non-changing, with a (changeable) `rel="self"` for download (in which case renamed-styles.json might still be used for creating server-side `rel="link"` redirects but wouldn't be necessary for client-side resolution. But that's a discussion for elsewhere.)

/cc @rmzelle

adomasven: This is not entirely on-point either, because we're getting the id from the document, where that's the only thing we store. Although this fact is a decent case to actually store rel="self" in the document prefs as well for future compatibility.

adomasven:
> This is not entirely on-point either, because we're getting the id from the document, where that's the only thing we store. Although this fact is a decent case to actually store rel="self" in the document prefs as well for future compatibility.

Also there's issue #685 for installing missing styles and [your comment](https://forums.zotero.org/discussion/47635/) on the forums suggesting storing the link too. 

I'm not sure about the prompt:

- There's no real reason to bother users with this. In 99% of cases this should just work magically.
- If we prompt most likely response from users will be confusion and/or just clicking "install" without thinking or understanding consequences if the URL is actually weird.

Limiting to zotero.org installs and prompting for foreign styles might be a reasonable approach though. Comments?

chrome/content/zotero/xpcom/style.js
adomasven: > I also don't think this is likely to be generally necessary functionality — it's pretty specific to the case of an embedded style in a stored document.

Looking at this code again, I agree. I'll simplify this.

rmzelle:
For me, it's definitely desirable to allow for auto-installing and auto-updating of styles outside of the CSL repository. Especially auto-updating. We sometimes say no to style submission to the CSL repository, and contributors are sometimes bummed that they lose the auto-update functionality already in Zotero for styles outside the official repo.

chrome/content/zotero/xpcom/style.js
rmzelle: > (The whole id situation in CSL is non-ideal — the id would ideally be non-changing, with a (changeable) rel="self" for download (in which case renamed-styles.json might still be used for creating server-side rel="link" redirects but wouldn't be necessary for client-side resolution. But that's a discussion for elsewhere.)

Yes, no disagreement there. There might be more old discussion on this, but this is what I could dig up:

settle on new base uri for styles, update ids accordingly
https://github.com/citation-style-language/styles/issues/2

settle on conventions for versioning
https://github.com/citation-style-language/styles/issues/7

Switch from "http" to "https" prefix for CSL style ID and "self"-links
http://xbiblio-devel.2463403.n2.nabble.com/Switch-from-quot-http-quot-to-quot-https-quot-prefix-for-CSL-style-ID-and-quot-self-quot-links-td7578982.html

call for comments on base URI issue
http://xbiblio-devel.2463403.n2.nabble.com/call-for-comments-on-base-URI-issue-td6097469.html

chrome/content/zotero/xpcom/style.js
```diff
@@ -238,24 +238,36 @@ Zotero.Styles = new function() {
...
 	 *     displayed in dialogs referencing the style
 	 * @param {Boolean} [silent=false] Skip prompts
 	 */
 	this.install = Zotero.Promise.coroutine(function* (style, origin, silent=false) {
 		var styleTitle;
+		var warnDeprecated;
+		if (style instanceof Components.interfaces.nsIFile) {
+			warnDeprecated = true;
+			style = {file: style};
+		} else if (Object.prototype.toString.call(style) === "[object String]") {
```
dstillman: `typeof style == 'string'` is fine. (We don't need to worry about `new String()`.)

chrome/content/zotero/xpcom/style.js
```diff
@@ -238,24 +238,36 @@ Zotero.Styles = new function() {
...
 		try {
-			if (style instanceof Components.interfaces.nsIFile) {
-				// handle nsIFiles
-				var url = Services.io.newFileURI(style);
-				var xmlhttp = yield Zotero.HTTP.request("GET", url.spec);
-				styleTitle = yield _install(xmlhttp.responseText, style.leafName, false, silent);
-			} else {
-				styleTitle = yield _install(style, origin, false, silent);
+			if (style.file) {
+				style.url = Services.io.newFileURI(style.file).spec;
```
dstillman: Should add a `style = Zotero.File.pathToFile(style);` above this line so that `file` can be a string path, which would be the preferred way to pass a file going forward. (And should update the function documentation above.)

test/tests/styleTest.js
```diff
@@ -0,0 +1,41 @@
+"use strict";
+
+describe("Zotero.Styles", function() {
+	var styleID = "http://www.zotero.org/styles/cell";
+	var styleFile = getTestDataDirectory();
+	var style;
+	styleFile.append('cell.csl');
```
dstillman: Better to avoid `nsIFile` when it's easy to do so:

`var styleFile = OS.Path.join(getTestDataDirectory().path, 'cell.csl');`

(Then at some point we'll fix `getTestDataDirectory()` to return a string path and can just take out the `.path`.)

test/tests/styleTest.js
```diff
@@ -0,0 +1,41 @@
...
+		style = yield Zotero.File.getContentsAsync(styleFile);
+	});
+	
+	describe("Zotero.Styles.install", function() {
+		afterEach(function* (){
+			assert.isOk(Zotero.Styles.get(styleID));
+			yield Zotero.Styles.get(styleID).remove();
+		});
+		
+		it("Installs the style from string", function* () {
```
dstillman: Lowercase, beginning with `should`, for all these. (Sorry, I know you dislike it, but that's the Mocha convention we've used for everything.)

dstillman:
> Limiting to zotero.org installs and prompting for foreign styles might be a reasonable approach though.

Yeah, a prompt for an install that wasn't from zotero.org, citationstyles.org, or GitHub seems reasonable. It's a pretty minor privacy issue, but if you opened a doc someone sent you, had Zotero, and used the plugin, you could end up revealing your IP address to a third-party server, which I guess shouldn't happen silently.

test/tests/styleTest.js
adomasven: This is literally testing nsIFile handling in this function (as I was changing the signature and wanted to make sure old calls will remain functional)

test/tests/styleTest.js
adomasven: > Sorry, I know you dislike it, but that's the Mocha convention we've used for everything.

It's a bit ugly, but I strongly prefer consistency with some ugliness over ideological chaos. I simply forget that we do "should" everywhere, especially when in a new testfile, so do continue pointing it out in the future.

test/tests/styleTest.js
dstillman: But this is the setup for all tests, not the `nsIFile` test, which could use `Zotero.File.pathToFile()` specifically. This way any other tests that need the file (e.g., to test passing a string path, as I suggest elsewhere) would have to convert from the deprecated `nsIFile` instead of the permanent string path.

adomasven:
Ok. I've spent quite a bit of time on integrationTests.js in this branch, which will let me be more comfortable with further messing in integration.js. Up for review

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -818,6 +818,7 @@ integration.error.noBibliography		= The current bibliographic style does not def
 integration.error.deletePipe			= The pipe that Zotero uses to communicate with the word processor could not be initialized. Would you like Zotero to attempt to correct this error? You will be prompted for your password.
 integration.error.invalidStyle			= The style you have selected does not appear to be valid. If you have created this style yourself, please ensure that it passes validation as described at https://github.com/citation-style-language/styles/wiki/Validation. Alternatively, try selecting another style.
 integration.error.fieldTypeMismatch		= Zotero cannot update this document because it was created by a different word processing application with an incompatible field encoding. In order to make a document compatible with both Word and LibreOffice, open the document in the word processor with which it was originally created and switch the field type to Bookmarks in the Zotero Document Preferences.
+integration.error.styleMissing			= The citation style used in this document is missing. Would you like to install it from %S
```
dstillman: Missing question mark at end

test/tests/integrationTests.js
```diff
@@ -0,0 +1,322 @@
...
+					displayDialogStub.reset();
+					displayAlertStub.reset();
+				});
+				
+				after(function* () {
+					displayAlertStub.restore();
+				});
+			
+				describe('when the style is not from zotero.org or citationstyles.org', function() {
+					it('should download the style and not call doc.setDocumentData if user clicks YES', function* () {
```
dstillman: Doesn't seem like this tests `setDocumentData`?

test/tests/integrationTests.js
```diff
@@ -0,0 +1,322 @@
...
+					initDoc(docID, {style});
+					displayDialogStub.reset();
+					displayAlertStub.reset();
+				});
+				
+				after(function* () {
+					displayAlertStub.restore();
+				});
+			
+				describe('when the style is not from zotero.org or citationstyles.org', function() {
```
dstillman: This isn't actually testing citationstyles.org, so shouldn't include that in the test description.

chrome/content/zotero/xpcom/integration.js
```diff
@@ -40,6 +40,19 @@ const FORCE_CITATIONS_RESET_TEXT = 2;
...
+const DIALOG_ICON_WARNING = 1;
+const DIALOG_ICON_CAUTION = 2;
+
+const DIALOG_BUTTONS_OK = 0;
+const DIALOG_BUTTONS_OK_CANCEL = 1;
+const DIALOG_BUTTONS_YES_NO = 2;
+const DIALOG_BUTTONS_YES_NO_CANCEL = 3;
+
+const NOTE_FOOTNOTE = 1;
+const NOTE_ENDNOTE = 2;
```
dstillman: A lot of this stuff is in the wrong commit. It would be better to move it into the previous one.

adomasven: I mean this stuff only needed moving once I started testing certain parts of integration.js, so arguably it makes just as much sense in this commit.

dstillman: It's unrelated to the problem being fixed, though, and a lot of it is just changes to the code that was introduced in the previous commit, so it pretty clearly belongs there. This is easy to fix, and it's important for people reviewing/merging code.

test/tests/integrationTests.js
```diff
@@ -4,28 +4,21 @@ describe("Zotero.Integration", function () {
...
-			this.noteIndex = DocumentPluginDummy.Field.noteIndex++;
-			this.wrappedJSObject = this;
-		}
+	var DocumentPluginDummy = {};
+	
+	DocumentPluginDummy.Application = function() { 
+		this.doc = new DocumentPluginDummy.Document();
+		this.primaryFieldType = "Field";
+		this.secondaryFieldType = "Bookmark";
+		this.fields = [];
```
dstillman: I meant all this stuff too. This should all go in the previous commit.

chrome/content/zotero/xpcom/integration.js
```diff
@@ -1306,8 +1320,8 @@ Zotero.Integration.Fields.prototype.addField = function(note) {
 	var field = this._doc.cursorInField(this._session.data.prefs['fieldType']);
 	if(field) {
 		if(!this._doc.displayAlert(Zotero.getString("integration.replace"),
-				Components.interfaces.zoteroIntegrationDocument.DIALOG_ICON_STOP,
-				Components.interfaces.zoteroIntegrationDocument.DIALOG_BUTTONS_OK_CANCEL)) {
+				DIALOG_ICON_STOP,
+				DIALOG_BUTTONS_OK_CANCEL)) {
```
dstillman: This should go with the same commits as the constants.

test/tests/styleTest.js
```diff
@@ -12,7 +12,7 @@ describe("Zotero.Styles", function() {
 	});
 	
 	describe("Zotero.Styles.install", function() {
-		afterEach(function* (){
+		afterEach(`${styleID} style should be installed`, function* (){
```
dstillman: This actually works?

test/tests/integrationTests.js
```diff
@@ -262,8 +262,9 @@ describe("Zotero.Integration", function () {
 					displayAlertStub.restore();
 				});
 			
-				describe('when the style is not from zotero.org or citationstyles.org', function() {
+				describe('when the style is not from a trusted source', function() {
```
dstillman: Are we actually doing something different for untrusted sources at this point?

test/tests/styleTest.js
adomasven: What? The string? Yeah, if the assertion fails in the hook it displays this string. Was reading mocha docs the other day and got surprised myself.

test/tests/integrationTests.js
adomasven: As described in the 3 following tests. Prompt if untrusted source, install without prompt otherwise.

dstillman: Ah, sorry, got my negation (and waterbirds) confused.

PR #1218, opened 2017-04-19
Feature/json pref docs

adomasven:
We should wait for proper 5.0 release (and possibly 4.0 phasing out) before merging this in. Otherwise people who are sharing documents between 4.0 and 5.0 users won't be able to collaborate, unless everyone upgrades to 5.0


KarlHegbloom:
Perhaps as long as it accepts the old format when it queries the
integration plugin, which must also support this, and then it can update
the document prefs data from that xml format to this json format.

dstillman:
@KarlHegbloom: That would break 4.0 clients. That's why this can't be merged yet.

But @adomasven, to do this smoothly we should probably roll out support for reading and writing JSON now, even if we don't start writing JSON to new documents until at least the 4.0 sync cut-off. I suspect upgrade cycles with everyone on Standalone will be a bit slower, since some users will need IT approval for upgrades in a way they didn't have now, so allowing older 5.0 clients to seamlessly handle JSON documents would be good.

adomasven:
I've updated this in accordance to @dstillman's comment

PR #1219, opened 2017-04-19
Do not return empty string only for `0` arg to #htmlSpecialChars

adomasven:
See discussion https://github.com/zotero/zotero/commit/ff74e0ff00832f9f2c80b869525fcc12a2d3ccf9

### DIFF
--- a/chrome/content/zotero/xpcom/utilities.js
+++ b/chrome/content/zotero/xpcom/utilities.js
@@ -459,7 +459,9 @@ Zotero.Utilities = {
 	 * @return {String}
 	 */
 	"htmlSpecialChars":function(str) {
-		if (str !== undefined && typeof str != 'string') {
+		if (str && typeof str != 'string') {
+			Zotero.debug('#htmlSpecialChars: non-string arguments are deprecated. Update your code', 
+				1, undefined, true);
 			str = str.toString();
 		}
 		
--- a/test/tests/integrationTests.js
+++ b/test/tests/integrationTests.js
@@ -470,4 +470,65 @@ describe("Zotero.Integration", function () {
 			});
 		});
 	});
+	
+	describe("DocumentData", function() {
+		it('should properly unserialize document data', function() {
+			var serializedXMLData = "<data data-version=\"3\" zotero-version=\"5.0.SOURCE\"><session id=\"F0NFmZ32\"/><style id=\"http://www.zotero.org/styles/cell\" hasBibliography=\"1\" bibliographyStyleHasBeenSet=\"1\"/><prefs><pref name=\"fieldType\" value=\"ReferenceMark\"/><pref name=\"storeReferences\" value=\"true\"/><pref name=\"automaticJournalAbbreviations\" value=\"true\"/><pref name=\"noteType\" value=\"0\"/></prefs></data>";
+			var data = new Zotero.Integration.DocumentData(serializedXMLData);
+			var expectedData = {
+				style: {
+					styleID: 'http://www.zotero.org/styles/cell',
+					locale: null,
+					hasBibliography: true,
+					bibliographyStyleHasBeenSet: true
+				},
+				prefs: {
+					fieldType: 'ReferenceMark',
+					storeReferences: true,
+					automaticJournalAbbreviations: true,
+					noteType: '0'
+				},
+				sessionID: 'F0NFmZ32',
+				zoteroVersion: '5.0.SOURCE',
+				dataVersion: '3'
+			};
+			// Convert to JSON to remove functions from DocumentData object
+			assert.equal(JSON.stringify(data), JSON.stringify(expectedData));
+		});
+		
+		it('should properly serialize document data', function() {
+			sinon.spy(Zotero, 'debug');
+			var data = new Zotero.Integration.DocumentData();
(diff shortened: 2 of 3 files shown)


chrome/content/zotero/xpcom/utilities.js
```diff
@@ -455,11 +455,15 @@ Zotero.Utilities = {
...
 	 *
-	 * @param {String} str
+	 * @param {String|Number} str
 	 * @return {String}
 	 */
 	"htmlSpecialChars":function(str) {
-		if (str !== undefined && typeof str != 'string') {
+		if ((str || str == 0) && typeof str != 'string') {
+			if (typeof str != 'number') {
+				Zotero.debug('#htmlSpecialChars: non-string and non-number arguments are deprecated. Update your code',
```
dstillman: There's not any reason to support number either, though. We should just warn when it's a non-string.

test/tests/integrationTests.js
```diff
@@ -470,4 +470,65 @@ describe("Zotero.Integration", function () {
...
+			data.dataVersion = "3";
+
+			// Convert to JSON to remove functions from DocumentData objects
+			assert.equal(JSON.stringify(processedData), JSON.stringify(data));
+			
+			// Make sure we are not triggering debug traces in Utilities.htmlSpecialChars()
+			assert.isFalse(Zotero.debug.calledWith(sinon.match.string, 1));
+			Zotero.debug.restore();
+		});
+	})
```
dstillman: Seems like this made it in here by accident.

chrome/content/zotero/xpcom/utilities.js
```diff
@@ -455,11 +455,15 @@ Zotero.Utilities = {
 	 *   <ZOTEROBREAK/> => <br/>
 	 *   <ZOTEROHELLIP/> => &#8230;
 	 *
-	 * @param {String} str
+	 * @param {String|Number} str
 	 * @return {String}
 	 */
 	"htmlSpecialChars":function(str) {
-		if (str !== undefined && typeof str != 'string') {
+		if ((str || str == 0) && typeof str != 'string') {
```
dstillman: Did you mean `str === 0`?

If we want to keep `true` support for now, seems like this should just be `if (typeof str == 'number' || str === true)`?

chrome/content/zotero/xpcom/utilities.js
```diff
@@ -455,11 +455,15 @@ Zotero.Utilities = {
 	 *   <ZOTEROBREAK/> => <br/>
 	 *   <ZOTEROHELLIP/> => &#8230;
 	 *
-	 * @param {String} str
+	 * @param {String|Number} str
```
dstillman: Just `String`

test/tests/integrationTests.js
adomasven: As discussed the doc prefs serialization relies on htmlSpecialChars working properly (and previously without `0` or `false` being converted it only worked properly by accident), so it made sense to include these here.

dstillman: It doesn't matter if it depends on this — it's a totally separate thing and should be in a separate commit.

chrome/content/zotero/xpcom/utilities.js
adomasven: Idk, maybe we want to keep object or array support too, because there is some broken code somewhere that relies on that.

test/tests/integrationTests.js
adomasven: But this is changing the behaviour of the function. You want to have tests to make sure the functionality which relies on the behaviour works properly.

dstillman: These tests are testing something different and should exist separately and persist even if this commit is reverted. If there were `htmlSpecialChars` tests, those would make sense in the same commit.

chrome/content/zotero/xpcom/utilities.js
adomasven: It could be extensions, but also [this piece of code](https://github.com/zotero/zotero/blob/355e25d9a95df0fa810e82c23a266ef615e0e27b/chrome/content/zotero/xpcom/utilities.js#L1293-L1293), which technically shouldn't ever be broken, but `xpath()` is used by a lot of translators

chrome/content/zotero/xpcom/utilities.js
adomasven: I've been looking into why this was here in the first place and it used to throw [before](https://github.com/rmzelle/zotero/blame/916db3fd4878c1e72f530c21ecb92bd0159a70ba/chrome/content/zotero/xpcom/utilities.js#L307-L309), but got changed [here](https://github.com/rmzelle/zotero/blame/48810f23e04229ace477713a4023ac3ccad49283/chrome/content/zotero/xpcom/utilities.js#L421). I regret ever touching this. Will update soon.

adomasven:
e272465 is not actually part of this PR, i've already pushed it to master, but github got fairly confused after a couple of forced-pushed, closing the PR automatically at one point.

dstillman:
Merged in https://github.com/zotero/zotero/commit/b077e1cd8c19969bec8dbba889a27c0ae760d3e1

PR #1222, opened 2017-05-02
Document preferences cleanup

adomasven:
Before:
<a href="https://cloud.githubusercontent.com/assets/5899315/25622282/993840d8-2f5c-11e7-9bf5-d5fdc584364e.png"></a>

Now: 
<a href="https://cloud.githubusercontent.com/assets/5899315/25622241/7aa352a2-2f5c-11e7-9f70-da6cb2735f48.png"></a>

### DIFF
--- a/chrome/content/zotero/bibliography.js
+++ b/chrome/content/zotero/bibliography.js
@@ -166,12 +166,6 @@ var Zotero_File_Interface_Bibliography = new function() {
 				document.getElementById("automaticJournalAbbreviations-checkbox").checked = true;
 			}
 		}
-		if(document.getElementById("storeReferences")) {
-			if(_io.storeReferences || _io.storeReferences === undefined) {
-				document.getElementById("storeReferences").checked = true;
-				if(_io.requireStoreReferences) document.getElementById("storeReferences").disabled = true;
-			}
-		}
 		
 		// set style to false, in case this is cancelled
 		_io.style = false;
@@ -271,7 +265,6 @@ var Zotero_File_Interface_Bibliography = new function() {
 			}
 			_io.useEndnotes = document.getElementById("displayAs").selectedIndex;
 			_io.fieldType = (document.getElementById("formatUsing").selectedIndex == 0 ? _io.primaryFieldType : _io.secondaryFieldType);
-			_io.storeReferences = document.getElementById("storeReferences").checked;
 		}
 		
 		// remember style and locale if user selected these explicitly
--- a/chrome/content/zotero/integration/integrationDocPrefs.xul
+++ b/chrome/content/zotero/integration/integrationDocPrefs.xul
@@ -84,8 +84,5 @@
 			<checkbox id="automaticJournalAbbreviations-checkbox" label="&zotero.integration.prefs.automaticJournalAbbeviations.label;"/>
 			<description class="radioDescription">&zotero.integration.prefs.automaticJournalAbbeviations.caption;</description>
 		</vbox>
-		
-		<checkbox id="storeReferences" label="&zotero.integration.prefs.storeReferences.label;"/>
-		<description class="radioDescription">&zotero.integration.prefs.storeReferences.caption;</description>
 	</vbox>
 </dialog>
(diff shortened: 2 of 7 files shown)


adam3smith:
Is it actually the case that for bookmarks "The document needs to be saved in doc/docx format"? -- I haven't tested, but I'm surprised that's the case.

adam3smith:
(otherwise yay for this)

dstillman:
Combined "The document must be saved in the .doc or .docx file format." with the previous string.

"Footnotes" should be lowercase (and should we add "endnotes" back?).

Switch order of "footnotes" and "accidentally modified".

How easy is it to modify bookmarks accidentally? Is that why we recommend ReferenceMarks?

adomasven:
> Is it actually the case that for bookmarks "The document needs to be saved in doc/docx format"? -- I haven't tested, but I'm surprised that's the case.

Probably not, but currently the plugins listen for saves and warn if you are trying to save as .odt with bookmarks, so I haven't tested. Possible that Word might lose bookmarks when opening .odt. But I'll check this out.

> How easy is it to modify bookmarks accidentally? Is that why we recommend ReferenceMarks?

We recommend ReferenceMarks/Fields for LO/Word. Yes, bookmarks are very modification non-resistant. Sometimes you can modify the bookmark by starting to type directly after the inserted citation, but the whole bookmark system in how it works is magic to me still, so I cannot comment. However, we've had people send in very broken bookmark-based documents, where once they get corrupt, they get very very badly corrupt and it's very hard to recover from. Furthermore, you cannot add field shading for bookmarks, like you can for ReferenceMarks/Fields, so you can't even tell what's up with straight-up doc examination.

dstillman:
> Possible that Word might lose bookmarks when opening .odt.

This sounds vaguely familiar, FWIW.

adomasven:
> Combine "The document must be saved in the .doc or .docx file format." with the previous string.

<a href="https://cloud.githubusercontent.com/assets/5899315/25652379/9a589fc8-2ff0-11e7-94d1-ba3f309aadfd.png"></a>

I'm not sure about this. The discussion in zotero/zotero-word-for-windows-integration#30 focused precisely on making the dialog less wordy and intimidating. This is okay:

<a href="https://cloud.githubusercontent.com/assets/5899315/25652531/56dc80ba-2ff1-11e7-933d-9343d32d44ac.png"></a>

dstillman:
All right, I guess if we have those for both it's OK — there's a line for features/limitations and then another line for the document format.

We could probably say "saved as an .odt" and "saved as a .doc or .docx" — "the .odt file format" seems a bit technical.

And maybe you're right that "or endnotes" is unnecessary. If you see that they can't be saved as footnotes, you're probably not going to be shocked when they don't work in endnotes either, right?

rmzelle:
Two more nitpicks:

* Should the window title be changed from "Document Preferences" to "Zotero Document Preferences" to indicate context?
* Dan and Sebastian both agreed with me that changing "Format Using:" to e.g. "Store Citations as:" would be an improvement. Can you make that change as well? See https://github.com/zotero/zotero-word-for-windows-integration/issues/30#issuecomment-285139567.

adomasven:
@rmzelle thanks! [Updated](https://github.com/zotero/zotero/commit/b80cdb2915cad5312a8874a67fe0b8afb2ae6ac9)

gracile-fr:
> Should the window title be changed from "Document Preferences" to "Zotero Document Preferences" or "Zotero - Document Preferences" to indicate context?
I agree (<a href="https://github.com/zotero/zotero/blob/b80cdb2915cad5312a8874a67fe0b8afb2ae6ac9/chrome/locale/en-US/zotero/zotero.dtd#L190">here's the string</a>).

rmzelle:
(@gracile-fr, @adomasven already made that change by programmatically prepending the client name (i.e. Zotero) before the original window title: https://github.com/zotero/zotero/commit/b80cdb2915cad5312a8874a67fe0b8afb2ae6ac9#diff-3af4d2a3f7cf404fa9befc1d27591570R148)

gracile-fr:
ah indeed, I had only looked at the localized files… thanks!

bwiernik:
> Is it actually the case that for bookmarks "The document needs to be saved in doc/docx format"? -- I haven't tested, but I'm surprised that's the case.

@adamsmith I looked into that—Word opens .odt Files as read-only. You have to convert to doc(x) or Save As to keep as odt. As the only reason to use Bookmarks is to move between Word and LO, it makes sense to just say that doc(x) is required.

PR #1240, opened 2017-06-03
Tweaks to make build work on Windows

tnajdek:
With this fix it should be possible to build and test on Windows with cygwin/msys.

In order for this to work on Windows, a relatively new version of git for windows is required (I believe it's >2.4.5 but I only tested with latest). Furthermore symlink support needs to be enable during the installation **as well as** when the repo is cloned with `git clone -c core.symlinks=true`.


gulpfile.js
```diff
@@ -72,56 +80,74 @@ const browserifyConfigs = [
...
-function getBrowserify() {
+function getBrowserify(exitOnError = true) {
 	const streams = browserifyConfigs.map(config => {
 		return gulp
 			.src(config.src)
 			.pipe(tap(file => {
 				file.contents = browserify(file.path, config.config).bundle();
 			}))
 			.pipe(rename(config.dest))
+			.on('error', partial(onError, exitOnError))
```
tnajdek: No, this wouldn't work, it would swallow the actual error message.

Partial is used here because `onError` needs to be executed in the context of current stream (hence `apply()` wouldn't work, it needs to know whether to exit and finally it needs to be called with the actual error message.

Replacement code, without the use of `partial` would be much less readable, something like this:

```
.on('error', function(err) { onError.bind(this)(exitOnError, err); })
```

test/runtests.sh
```diff
@@ -143,6 +143,10 @@ if [ -z $IS_CYGWIN ]; then
 	echo "`MOZ_NO_REMOTE=1 NO_EM_RESTART=1 \"$FX_EXECUTABLE\" -v`"
 fi
 
+if [ $IS_CYGWIN ]; then
+	FX_ARGS="$FX_ARGS -console"
```
tnajdek: Fair enough, should I push an updated PR?

gulpfile.js
dstillman: Oops, sorry, didn't look at the onError definition. Still, I'd prefer the replacement version — it seems clearer to me what it's doing without the extra dependency and without having to look up what `partial` does (in a library that's otherwise pretty much unneeded in the ES6+ era).

package.json
```diff
@@ -42,7 +42,9 @@
     "gulp-sass": "^3.1.0",
     "gulp-tap": "^1.0.1",
     "gulp-util": "^3.0.7",
+    "lodash.partial": "^4.2.1",
```
dstillman: Can be removed

dstillman:
Merged in https://github.com/zotero/zotero/commit/98f1ac5edbd716791f19140e90b1ea37acbefad8

PR #1242, opened 2017-06-06
Refactor integration and delay citation updates

### LINKED ISSUE #1241: Option to delay updating citations in documents
Moved from zotero/zotero-word-for-windows-integration#35

See discussion https://forums.zotero.org/discussion/64960/feature-suggestion-delay-updating-citations-in-documents

adomasven:
So integration.js needed a refactor and some tests for a long time now. There is still a lot of room to improve, as the code is still quite dirty, but it's much better and more readable now, than before. Mainly, the document and session updates are localized and not all over the place, but there are still some weird circular object dependencies and for `Zotero.Integration.execCommand()` the session and document objects are globally exposed during a transaction.

----

The PR introduces the ability to delay citation updates and closes #1241. Note that this will speed up inserts that take 10min to 2s or so. Pretty good.

When a document update takes longer than 5s, a prompt is displayed:

Clicking *Yes* or *No* opens up document preferences window with the new option highlighted:

The checkbox is then always available for that document and the prompt is not displayed again.

Clicking *Cancel* in the prompt displays the prompt the next time an update takes longer than 5s, until the user chooses either *Yes* or *No*.

I think a KB page would be useful to link to from the checkbox description.

This is what the delayed citation looks like (note the wavy underlining):
image

See other possible [RTF formatting options](http://www.biblioscape.com/rtf15_spec.htm#Heading42)

### DIFF
--- a/chrome/content/zotero/integration/addCitationDialog.js
+++ b/chrome/content/zotero/integration/addCitationDialog.js
@@ -59,10 +59,8 @@ var Zotero_Citation_Dialog = new function () {
 	this.listItemSelected = listItemSelected;
 	this.up = up;
 	this.down = down;
-	this.add = add;
 	this.remove = remove;
 	this.setSortToggle = setSortToggle;
-	this.citationSortUnsort = citationSortUnsort;
 	this.confirmRegenerate = confirmRegenerate;
 	this.accept = accept;
 	this.cancel = cancel;
@@ -373,13 +371,13 @@ var Zotero_Citation_Dialog = new function () {
 	/*
 	 * Adds an item to the multipleSources list
 	 */
-	function add(first_item) {
+	this.add = Zotero.Promise.coroutine(function* (first_item) {
 		
 		var pos, len;
 		var item = itemsView.getSelectedItems()[0]; // treeview from xpcom/itemTreeView.js
 		
 		if (!item) {
-			sortCitation();
+			yield sortCitation();
 			_updateAccept();
 			_updatePreview();
 			return;
@@ -412,11 +410,11 @@ var Zotero_Citation_Dialog = new function () {
 		_citationList.ensureElementIsVisible(selectionNode);
 
 		// allow user to press OK
-		selectionNode = sortCitation(selectionNode);
+		selectionNode = yield sortCitation(selectionNode);
 		_citationList.selectItem(selectionNode);
 		_updateAccept();
 		_updatePreview();
-	}
+	});
 	
 	/*
 	 * Deletes a citation from the multipleSources list
@@ -446,11 +444,11 @@ var Zotero_Citation_Dialog = new function () {
 	/*
 	 * Sorts preview citations, if preview is open.
 	 */
-	function citationSortUnsort() {
+	this.citationSortUnsort = Zotero.Promise.coroutine(function* () {
 		setSortToggle();
-		sortCitation();
+		yield sortCitation();
 		_updatePreview();
-	}
+	});
 
 	/*
 	 * Sets the current sort toggle state persistently on the citation.
@@ -468,7 +466,7 @@ var Zotero_Citation_Dialog = new function () {
 	/*
 	 * Sorts the list of citations
  	 */
-	function sortCitation(scrollToItem) {
+	var sortCitation = Zotero.Promise.coroutine(function* (scrollToItem) {
  		if(!_sortCheckbox) return scrollToItem;
  		if(!_sortCheckbox.checked) {
  			io.citation.properties.unsorted = true;
@@ -485,7 +483,7 @@ var Zotero_Citation_Dialog = new function () {
 		
 		// run preview function to re-sort, if it hasn't already been
 		// run
-		io.sort();
+		yield io.sort();
 		
 		// add items back to list
 		scrollToItem = null;
@@ -502,7 +500,7 @@ var Zotero_Citation_Dialog = new function () {
 		
 		if(scrollToItem) _citationList.ensureElementIsVisible(scrollToItem);
 		return scrollToItem;
-	}
+	});
 	
 	/*
 	 * Ask whether to modifiy the preview
--- a/chrome/content/zotero/xpcom/integration.js
+++ b/chrome/content/zotero/xpcom/integration.js
@@ -37,10 +37,6 @@ const FORCE_CITATIONS_REGENERATE = 1;
 // Specifies that citations should be reset regardless of whether formattedText has changed
 const FORCE_CITATIONS_RESET_TEXT = 2;
 
-// this is used only for update checking
-const INTEGRATION_PLUGINS = ["zoteroMacWordIntegration@zotero.org",
-	"zoteroOpenOfficeIntegration@zotero.org", "zoteroWinWordIntegration@zotero.org"];
-	
 // These must match the constants in corresponding word plugins 
 const DIALOG_ICON_STOP = 0;
 const DIALOG_ICON_WARNING = 1;
@@ -58,8 +54,6 @@ Zotero.Integration = new function() {
 	Components.utils.import("resource://gre/modules/Services.jsm");
 	Components.utils.import("resource://gre/modules/AddonManager.jsm");
 	
-	const INTEGRATION_MIN_VERSIONS = ["3.1.7.SOURCE", "3.5b2.SOURCE", "3.1.3.SOURCE"];
-
 	this.currentWindow = false;
 	this.sessions = {};
 	
@@ -114,8 +108,6 @@ Zotero.Integration = new function() {
 		} catch(e) {
 			Zotero.logError(e);
 		}
-		
-		Zotero.Promise.delay(1000).then(_checkPluginVersions);
 	}
 
 	/**
@@ -182,42 +174,6 @@ Zotero.Integration = new function() {
 		}
 	});
 	
-	/**
-	 * Checks to see that plugin versions are up to date.
-	 * @return {Promise} Promise that is resolved with true if versions are up to date
-	 *     or with false if they are not.
-	 */
-	var _checkPluginVersions = new function () {
-		var integrationVersionsOK;
-		
-		return function _checkPluginVersions() {
-			if(integrationVersionsOK) {
-				if(integrationVersionsOK === true) {
-					return Zotero.Promise.resolve(integrationVersionsOK);
-				} else {
-					return Zotero.Promise.reject(integrationVersionsOK);
-				}
-			}
-			
-			var deferred = Zotero.Promise.defer();
-			AddonManager.getAddonsByIDs(INTEGRATION_PLUGINS, function(addons) {
-				for(var i in addons) {
-					var addon = addons[i];
-					if(!addon || addon.userDisabled) continue;
-					
-					if(Services.vc.compare(INTEGRATION_MIN_VERSIONS[i], addon.version) > 0) {
-						deferred.reject(integrationVersionsOK = new Zotero.Exception.Alert(
-							"integration.error.incompatibleVersion2",
-							[Zotero.version, addon.name, INTEGRATION_MIN_VERSIONS[i]],
-							"integration.error.title"));
-					}
-				}
-				deferred.resolve(integrationVersionsOK = true);
-			});
-			return deferred.promise;
-		};
-	}
-	
 	this.getApplication = function(agent, command, docId) {
 		// Try to load the appropriate Zotero component; otherwise display an error
 		try {
@@ -242,10 +198,10 @@ Zotero.Integration = new function() {
 	this.execCommand = new function() {
 		var inProgress;
 		
-		return function execCommand(agent, command, docId) {
+		return Zotero.Promise.coroutine(function* execCommand(agent, command, docId) {
 			var document;
 			
-			if(inProgress) {
+			if (inProgress) {
 				Zotero.Utilities.Internal.activate();
 				if(Zotero.Integration.currentWindow && !Zotero.Integration.currentWindow.closed) {
 					Zotero.Integration.currentWindow.focus();
@@ -255,15 +211,15 @@ Zotero.Integration = new function() {
 			}
 			inProgress = true;
 			
-			// Check integration component versions
-			return _checkPluginVersions().then(function() {		
-				var application = Zotero.Integration.getApplication(agent, command, docId);
-				
-				// Try to execute the command; otherwise display an error in alert service or word processor
-				// (depending on what is possible)
-				document = (application.getDocument && docId ? application.getDocument(docId) : application.getActiveDocument());
-				return Zotero.Promise.resolve((new Zotero.Integration.Document(application, document))[command]());
-			}).catch(function(e) {
+			var application = Zotero.Integration.getApplication(agent, command, docId);
+			
+			// Try to execute the command; otherwise display an error in alert service or word processor
+			// (depending on what is possible)
+			document = (application.getDocument && docId ? application.getDocument(docId) : application.getActiveDocument());
+			try {
+				yield Zotero.Promise.resolve((new Zotero.Integration.Document(application, document))[command]());
+			}
+			catch (e) {
 				if(!(e instanceof Zotero.Exception.UserCancelled)) {
 					try {
 						var displayError = null;
@@ -301,8 +257,8 @@ Zotero.Integration = new function() {
 						Zotero.logError(e);
 					}
 				}
-			})
-			.finally(function() {
+			}
+			finally {
 				if(document) {
 					try {
 						document.cleanup();
@@ -327,8 +283,8 @@ Zotero.Integration = new function() {
 				}
 				
 				inProgress = Zotero.Integration.currentWindow = false;
-			});
-		};
+			}
+		});
 	};
 	
 	/**
--- a/chrome/content/zotero/xpcom/integration.js
+++ b/chrome/content/zotero/xpcom/integration.js
@@ -1337,7 +1337,10 @@ Zotero.Integration.Fields.prototype._updateDocument = function* (forceCitations,
  * Brings up the addCitationDialog, prepopulated if a citation is provided
  */
 Zotero.Integration.Fields.prototype.addEditCitation = Zotero.Promise.coroutine(function* (field) {
-	var newField, citation, session = this._session;
+	var newField, citation;
+	
+	// TODO: refactor citation/field preparation
+	// Citation loading should be moved into Zotero.Integration.Citation
 	
 	// if there's already a citation, make sure we have item IDs in addition to keys
 	if (field) {
@@ -1352,7 +1355,7 @@ Zotero.Integration.Fields.prototype.addEditCitation = Zotero.Promise.coroutine(f
 		
 		if (citation) {
 			try {
-				yield session.lookupItems(citation);
+				yield this._session.lookupItems(citation);
 			} catch(e) {
 				if(e instanceof Zotero.Integration.MissingItemException) {
 					citation.citationItems = [];
@@ -1391,7 +1394,41 @@ Zotero.Integration.Fields.prototype.addEditCitation = Zotero.Promise.coroutine(f
 		citation = {"citationItems":[], "properties":{}};
 	}
 	
-	var io = new Zotero.Integration.CitationEditInterface(citation, field, this, session);
+	// -------------------
+	// Preparing stuff to pass into CitationEditInterface
+	var fieldIndexPromise = this.get().then(function(fields) {
+		for (var i=0, n=fields.length; i<n; i++) {
+			if (fields[i].equals(field)) {
+				return i;
+			}
+		}
+	});
+	
+	var citationsByItemIDPromise = fieldIndexPromise.then(function() {
+		return this.updateSession();
+	}.bind(this)).then(function() {
+		return this._session.citationsByItemID;
+	}.bind(this));
+
+	// Required for preview fn
+	citation.properties.noteIndex = field.getNoteIndex();
+	var previewFn = Zotero.Promise.coroutine(function* (citation) {
+		let idx = yield fieldIndexPromise;
+		let citationsPre, citationsPost, citationIndices;
+		[citationsPre, citationsPost, citationIndices] = this._session._getPrePost(idx);
+		try {
+			return this._session.style.previewCitationCluster(citation, citationsPre, citationsPost, "rtf");
+		} catch(e) {
+			throw e;
+		}
+	}.bind(this));
+		
+	var io = new Zotero.Integration.CitationEditInterface(
+		// Clone citation
+		JSON.parse(JSON.stringify(citation)), 
+		this._session.style.opt.sort_citations, fieldIndexPromise, citationsByItemIDPromise,
+		previewFn
+	);
 	
 	if (Zotero.Prefs.get("integration.useClassicAddCitationDialog")) {
 		Zotero.Integration.displayDialog(this._doc,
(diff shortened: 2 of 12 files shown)


adomasven:
(I'd rather not squish these commits. I've found commit history and access to detailed blame super useful when figuring out the reasoning for certain choices in the code that I was refactoring. Every time you run into a "megacommit" that sort of detail is lost and you're left wondering whether the unclear code choices are just "leftovers" or a bugfix)

adomasven:
> I think a KB page would be useful to link to from the checkbox description.

It would also help reduce the textual clutter in the pref dialog, which we tried to minimize in #1222

dstillman:
Awesome. I'll look at this more later, but a few quick thoughts:

- I don't think popping up the doc prefs window is necessary — I think we can just add a new line to that prompt that says "You can change this setting later in the document preferences.", which will also make people feel more comfortable choosing an option.

- Green squiggly looks too much like spelling/grammar, so we'll need to find something else. Using RTF is clever, but we'll need to be careful that we're not making things more confusing/distracting than necessary. I'll look through the RTF options later, but possible we just want to make the foreground or background color slightly grayer.

- I think the checkbox label could just be "Delay citation updates until manual refresh", either without any text below it or maybe with a short sentence like "Delaying updates can speed up citation insertion in large documents." if we think that's necessary (which I'm not sure it is, particularly if this isn't even shown by default and it's only going to get turned on by that prompt).

- I don't think we need a specific KB link for this. We can just have a help icon in the corner that takes you to the [documentation](https://www.zotero.org/support/word_processor_plugin_usage), like we have in the main prefs. We can and probably should also use tooltips — less discoverable, but some people would still see them.

adomasven:
> I'll look through the RTF options later, but possible we just want to make the foreground or background color slightly grayer

This is default look for highlighted field codes in both Word and LO, so a bad option. There's also a side-effect of changing the colour, that when the citation gets refreshed the colour remains (because Word stores that as text style), so we'd have to be setting all non-delayed citations to default colour with RTF, when delayed insertion is enabled, to revert the colour. Could be annoying if someone uses weird foreground/background colours for some reason, also due to LO bugs inserting RTF is time consuming. Also note that track-changes uses green and red colours, and strikethrough style.

There are also other things we can do, like add `#PENDING#` prefix and suffix to the citation text.

adam3smith:
This is great. I like Dan's shorter message  "Delay citation updates until manual refresh" -- we just went through removing text clutter there, I'd keep with that spirit.

Using Word extensively for work right now, I'd be very pleased with the green squiggly. The squigglies (sp?) indicate that I need to look at something again, which is the right message here. Grammar is blue, so I don't think that's confusing. Shading is a no-go as Adomas points out. I think adding text like `#PENDING` is ugly and possibly more confusing still.

adomasven:
> This is great. I like Dan's shorter message "Delay citation updates until manual refresh" -- we just went through removing text clutter there, I'd keep with that spirit.

Yep, already updated.

> I'd be very pleased with the green squiggly

Note that the squiggly isn't green. May look like that in the screenshot, but it's black. There's no way to independently control the colour of the underlining with RTF (or word processors, for that matter). Otherwise, squigglies are universally used by grammar checkers in red and/or blue, which I'm not sure if we want to collide with, even if without colour.

adam3smith:
Ah right, not sure why I thought this was green. 
Since I'm guessing the obvious option from RTF -- Marching red ants -- is not going to find much love here, I still think that squiggly is good, but I think any of the dotted underlines would work, too. I'd probably go with the simple `\uld`
image

Shown here "conflicting" with Word's spell check.

rmzelle:
> I like Dan's shorter message "Delay citation updates until manual refresh"

I think "Disable automatic citation updates" would be clearer, with an explanation that mentions the use of the Refresh button. "Delay" doesn't seem the right word (https://www.merriam-webster.com/dictionary/delay) for switching from automatic to manual.

adomasven:
Maybe switch to "Postpone citation updates until manual refresh". The wording can be made more succinct with a synonym of "delay".

dstillman:
There's nothing wrong with "Delay citation updates until manual refresh" to describe what this is doing, and the advantage over "Disable automatic citation updates" is that it conveys that this isn't an optional step — your citations will likely be wrong until you refresh manually — without extra explanatory text. Using a negative action ("disable") for the enabled state of a checkbox is also a bit of an anti-pattern. But "until manual refresh" might be a little confusing, so mentioning "Refresh" explicitly is probably a good idea, at least in the prompt.

I'll try this in the UI later and see what works.

rmzelle:
> Using a negative action ("disable") for the enabled state of a checkbox is also a bit of an anti-pattern.

Ah, right. One more suggestion, then: "Automatically update citations" checkbox (enabled by default) for the Zotero document preferences window, with explanation "Disable automatic citation updates to speed up citation insertion in large documents. Citations can be manually updated with the Zotero Refresh button."

KarlHegbloom:
If the "Scannable cite" or "RTF scan" style is chosen, does it perform the time-consuming operations on updates? Even so, that form would work well for a temporary right-after-inserting-citation format.

It would be cool if a background thread could talk with zotero integration.js from the word processor plugin, gather what it needs, and then perform its' work in the background, as a "second typist", while the writer just keeps on going. e.g., GNU Emacs can open a window on multiple X desktops, and two people can edit the same file, with cursors in different places. That depends on the editor having a separate abstraction for the file's "buffer", the "frame" (an X Window), and the "window" (a subwindow inside of an emacs frame); the cursor point where input is inserted is associated with the "window", which is displaying a "buffer" inside a "frame", and thus it allows two people typing into different "window"s to each have their own separate "point" of text insertion.

For my [zotero-texmacs-integration](https://github.com/KarlHegbloom/zotero-texmacs-integration), in the bootstrap.js for my [propachi-texmacs](https://github.com/KarlHegbloom/propachi-texmacs) I have monkey-patched a few things that I would like to become part of the integration.js. Please UTSL look at what I did and see what and why.

I needed to be able to change the outputFormat from RTF to my own LaTeX-like format, still calling it bbl for now;

Also, I've defined a variableWrapper, in order to add support for hyperlinking that would not otherwise be possible;

And, I've added an "assertCitation" call that is effectively "editCitation" without the editting... It just sends what the citation cluster already contains, and receives back the reformatted text, in order to update ibid, etc. after my plugin's keybindings are used to cut / paste entire citation clusters or subcite's within them. The YouTube video's show what it looks like at this point; they are linked to from the README.md on the zotero-texmacs-integration page.

I have other ideas also... or had, but not all thought out or written down at this point. I'll try to stay in touch.

KarlHegbloom:
wrt renaming "Document" to "Interface": I think it's supposed to create one per document being editted... I don't care about the name changing really, but just want to be sure that it's going to keep working right when I have multiple documents open in the same instance of the editor, and perhaps it ought to warn you if you open the same document in more than one instance of it? (Accidental minimize over the weekend... Monday open the document fresh... edit it in two places...)

KarlHegbloom:
Also, there have been protocol stability issues, where when there's an error on my side (zotero-texmacs-integration) the entire Juris-M / Zotero needs to be restarted to make it reset itself. It ought to notice that something went wrong, and reset itself. TINC

KarlHegbloom:
wrt "propachi-texmacs" bootstrap.js and monkey-patches there, pertaining to outputFormat:  I'd like to be able to ask it for copies formatted via more than one outputFormat.

Use case: TeXmacs, exporting the document as HTML, I'll want the HTML outputFormat formatted citation inside an if-html; exporting as LaTeX, I'll want the "bbl" kept available for output, to avoid the translation -> TeXmacs -> other format and associated information loss.

Proposal: Ability for editor plugin to tell it what outputFormat to send the citations back as, with a primary one, and allow secondary ones to appear inside of the JSON fieldCode. {"Mockcitation": [ {"outputFormat": "bbl"  "data": "bbl formatted citation here"}, {"outputFormat": "html"  "data": "html formatted citation here"} ] }

Also to think over: What if a user has LibreOffice, LyX, Emacs, and TeXmacs all opened to documents with citations in them at the same time... while trying them out... or showing them off... or one for one purpose, another for another... So the outputFormat ought to be a document setting, but also per the particular editor's capabilities... but the plugin can set the default as a document preference or in the initial handshaking.

I wonder if it would be useful to define an ODT XML outputFormat, for use by LibreOffice? What does MS Word use internally? But they have RTF down since it's really their product... RTF is not really a "standard" per se, and it's not really easy to translate it to other formats. Also, it's a "display" markup, rather than a "structural" one, right?

Sending a "structural" XML markup that gets formatted by document styles works very well. That's pretty much what the HTML outputFormat does, and also the "bbl" one. (I'll probably rename it next iteration since it's not really the same as the output from BibTeX any longer.)

KarlHegbloom:
[Sorry so many separate comments; trying to keep them partititioned a little by topic...]

When TeXmacs has Guile 2.2 working inside of it, I think that it will be easier for me to have a separate actual thread talking to the Zotero integration. I'm having ideas about having Zotero send an "RTF scan" formatted citation first, as a quick temp value to show the typist, returning control at that point to the typist in the editor... while completing all the other work asynchronously.

Very hypothetically... I don't know enough about it and have not thought this out well yet... the zotero integration thread would then gather and potentially organize the remainder of the incoming requests into "patches" that can be applied in the background; 

The way it works now is that the typist has to  wait for "document_Complete", and it won't really let you do anything until it's all done due to single threaded design. The document tree updates can conceiveably happen all at once, and then the redisplay can happen one time after that. Right now the redisplay happens during the wait time for the network protocol to happen between TeXmacs and Zotero integration.js.

I think that fielding... er, catching and organizing... each of those transactions in a thread might work, and that the actual document tree edits they entail can happen that fast since I keep the locations of the citationCluster fields as tree-pointers, so there's no searching, just a jump straight to the right place, maintained by TeXmacs internal "observers" mechanism. But in a large document, the typist might insert a citation, the editor comes back right away with a temporary value inserted there, and the typist keeps clattering along, hopefully without lag or delay... and then moves to insert another citationCluster before the Zotero machinery is done with all the updates.

Can Zotero's integration.js handle that? Probably not yet... but the thing to think over is about how to make such thing possible. In my zotero.org collection are a few articles about using TeXmacs as an interface to mathematical software... and something like what I'm talking about here is discussed in there;

One thing at a time department is on the other line reminding me to eat...  :-)

adomasven:
@KarlHegbloom GitHub is a great place for us to triage issues, pull requests and sometimes get feedback from the community, but we appreciate if the conversation is kept to the minimal and on topic here. 

We are happy to see and discuss development around Zotero, but as described in the [support pages](https://www.zotero.org/support/dev/start), the correct place to talk about these things is on [zotero-dev mailing list](https://groups.google.com/group/zotero-dev).

adomasven:
@dstillman Ping, since 5.0 has been released

dstillman:
After pausing updates with the debugger to trigger the prompt, and then saying Yes, I got an error on the next manual refresh:

https://gist.github.com/dstillman/7c4e8f609d9a47d147326346f07eeaa8

adomasven:
> After pausing updates with the debugger to trigger the prompt

Could you detail the steps?

dstillman:
1) Insert a citation into a new LibreOffice doc, Chicago Author-Date, Reference Marks

2) Add a breakpoint at the beginning of _updateCitations().

3) Insert another citation, wait 5 seconds, continue from the breakpoint.

4) Delay prompt pops up. Say Yes.

5) Remove breakpoint, insert another citation, get squiggly lines.

6) Press Refresh. Kaboom.

chrome/content/zotero/xpcom/integration.js
```diff
@@ -869,11 +886,37 @@ Zotero.Integration.Fields.prototype._processFields = Zotero.Promise.coroutine(fu
...
+		var deferred = Zotero.Promise.defer();
+		setTimeout(deferred.resolve.bind(deferred));
+		yield deferred.promise;
+	}
+	for (let t of this._updateDocument(forceCitations, forceBibliography, ignoreCitationChanges)) {
+		// Promise.each doesn't yield for UI updates here for some reason, so we yield into the next event loop
+		var deferred = Zotero.Promise.defer();
+		setTimeout(deferred.resolve.bind(deferred));
+		yield deferred.promise;
+	}
```
dstillman: Promise.each() doesn't yield because there are no promises, just generators.

Should just make `_updateCitations()` and `_updateDocument` `async` functions and have them `await` `Zotero.Promise.delay()` (which is all you're really doing here) instead of yielding.

chrome/content/zotero/xpcom/integration.js
```diff
@@ -869,11 +886,37 @@ Zotero.Integration.Fields.prototype._processFields = Zotero.Promise.coroutine(fu
...
+
+	var diff = ((new Date()).getTime() - startTime)/1000;
+	Zotero.debug(`Integration: updateDocument complete in ${diff}s`)
+	// If the update takes longer than 5s suggest delaying citation updates
+	if (diff > 5 && !this._session.data.prefs.dontAskDelayCitationUpdates && !this._session.data.prefs.delayCitationUpdates) {
```
dstillman: Move 5 to a constant up top

chrome/content/zotero/integration/addCitationDialog.js
```diff
@@ -373,13 +371,13 @@ var Zotero_Citation_Dialog = new function () {
 	/*
 	 * Adds an item to the multipleSources list
 	 */
-	function add(first_item) {
+	this.add = Zotero.Promise.coroutine(function* (first_item) {
```
dstillman: `async` for new code

adomasven: FWIW all of this was written before we had the babel build step

chrome/content/zotero/xpcom/integration.js
dstillman: Zotero.Promise.delay(1) definitely yields to the next event loop, because it is a delay… It's literally a setTimeout() call. There's no reason to use generators here.

adomasven: > There's no reason to use generators here.
In my defence, I was trying to change as few things as possible and the generator code was one I found and didn't want to touch, but agreed.

chrome/content/zotero/integration/addCitationDialog.js
dstillman: Ah, right.

adomasven:
Updated. 

- Some citeproc/embedded citation fixes
- Loads up all fields from the doc upon first interaction even if delaying citations, so that we can prompt for already cited/embedded items in the citation dialog.
- Requested changes

dstillman:
I'm still getting the same error.

adomasven:
Responding to @fbennett comment on https://github.com/Juris-M/citeproc-js/issues/51#issuecomment-323297374, since it's not relevant to that issue and this PR is where the discussion should be taking place.

> I'll have some questions, but I'll start with some description from the citeproc-js side as context. I haven't been following the content of the changes under the zotero/zotero#1242 issue, so there may be some back-and-forth before I catch on. My first thought is that previewCitationCluster() probably isn't the thing to use here, but I may need to listen up first.
>
> previewCitationCluster() is a wrapper around processCitationCluster(), which is the bottleneck culprit that performs disambiguation and maintains citation state in the processor registry. It essentially runs the latter function, but invoking additional logic that reverses registry changes that would be caused by the insertion or edit, after the rendered sample text is produced. Invoked as intended (with pre- and post- arrays), it will take as long or a bit longer to return than a call to processCitationCluster().
> 
> For a quick static representation of a cited item, I think getCitationCluster() will be more suitable. It doesn't touch the registry, and doesn't perform any heavyweight context evaluation. It's what processCitationCluster() calls once it's worked out what the correct parameters are for the elements of a citation. Since getCitationCluster() doesn't touch the registry, it won't add the static item to the bibliography. To do that in a lightweight way that avoids disambiguation, I think you could use updateUncitedItems().
> 
> Does that roughly make sense, or am I misunderstanding how postponed rendering is meant to work?

We just need something, that formats the citation as if it were the first citation in the document according to the selected style CSL rules without messing up the processor state. This citation is inserted as a placeholder until the user presses refresh. The choice for `previewCitationCluster()` was informed by the citeproc docs I managed to get my hands on (the current ones and the older ones).

What would be the best function to use in this situation, and if it is `getCitationCluster()`, how would I properly invoke it for such use case?

fbennett:
Sorry about that.

Understood. I will update the current manual with sections on these functions soon. It may take a couple of days, please bear with.

fbennett:
Docs still to come, but here's some progress. It's been a long time since I looked carefully at the functions behind `processCitationCluster()` (years), and I was mistaken about which is meant to do this job. It's actually `makeCitationCluster()`. There was a small bug in (relatively) recent code that prevented it from running on unregistered items. I've fixed the bug, so for the function to work in this context, you'll need to use the latest tagged & bundled version of the code (1.1.174).

Re how to invoke it, I've cast a [processor test fixture](https://github.com/Juris-M/citeproc-js/blob/master/tests/citeproc-js/makecitationcluster.js) that shows it in action. The test isn't totally transparent, but the `_setCache()` operation just makes the item data available to the processor's `retrieveItem()` method. The `makeCitationCluster()` method takes a single argument, which is an array of parameter items (with `id` and `position:0`, and optionally `locator`, `label` etc). The processor will acquire the metadata for the item during execution, using its `retrieveItem()` method.

adomasven:
Updated:

- Uses `makeCitationCluster()` instead of `previewCitationCluster()` for previews when delaying is turned on
- Some adjustments on what occurs when delayed and fresh citations get edited in-text to make the logic more consistent. Tests added to [clarify logic](https://github.com/zotero/zotero/pull/1242/files#diff-99cb20105c486f0a7b94b743b6bb4b41R499)

adomasven:
Ok, so this is actually still problematic @fbennett.

`makeCitationCluster()` doesn't set `citation.sortedItems`, which we rely on to display citation items in the correct order.

fbennett:
Got it on (1) and (2). Before digging deeper, there may be a few edge cases to consider.

Disambiguation is a bottleneck, so we'll have to avoid that; but when performed, it can affect both sort ordering in 'citation.sortedItems' and citation collapsing. Some users will complain when the semantics of affixes stirred in during the initial insert are disturbed by a change in sort ordering on refresh. Not sure what the best solution is there, but it's certain to come up.

Re using `updateUncitedItems()`, you're right, that was bad advice. I'm not sure what the effect of a static-citation insert should be, though. The choices would be (a) to add the item to registry as a cited item, but without processing for disambiguation; or (b) to leave the registry untouched.

With (a) are that performance might suffer a bit, but the larger problem is that, in static-insert mode, there is no way of detecting that references have been removed from the document. Some users will find that references disappear from the bibliography on final refresh, and there will be complaints.

With (b), items added in static-insert mode will not appear in the bibliography until a full refresh. Not sure if that's a problem.

Both (a) and (b) will raise some problems for numeric styles. In (1), the correspondence between in-text numeric references and numbered entries in the bibliography would break, potentially leading to confusion. In (b), the details of a reference newly added in static-insert mode would not appear anywhere in the document until a full refresh, which would be pretty inconvenient.

(Not sure what the right choices are on these.)

fbennett:
It would also be good to know what your profiling might show about the balance of burden between extracting refs from the document (by Zotero integration layer), updating registry records (by `update[Uncited]Items()`], and generating citations (by `[make|process|preview]CitationCluster()`).

If the cost of extracting citation data from the document is very low, and that of resetting the registry is the highest of the three, maybe we should be looking at just stripping down the registry update to bare essentials in "static" mode, and otherwise just continuing to do what we do.

If the cost of extracting citation data from the document is significant, the citation/bibliography discontinuity issues will be intractable. In that case, one possible solution might be to use a separate processor instance to generate author-date reference placeholders for the static references.

fbennett:
Reading through the thread again, another thought is that generating the citation placeholder text in a separate processor instance would allow you to play arbitrary games with their formatting---adding a plain underline, say, or giving them a special <<set of affixes>> (or both).

adomasven:
Overall, the tradeoff for users here is clear - the document isn't fully and properly updated and there are no promises for e.g. the bibliography being correct. That said

> Disambiguation is a bottleneck, so we'll have to avoid that; but when performed, it can affect both sort ordering in `citation.sortedItems` and citation collapsing. Some users will complain when the semantics of affixes stirred in during the initial insert are disturbed by a change in sort ordering on refresh.

Thanks for bringing this issue up. This is perhaps the most concerning of them all, especially if multiple items are cited, and the first item in the list has a prefix as a prefix for the whole citation, and the last item in the list has a suffix, as a suffix for a whole citation. One way to address this would be to have citation-level prefixes and affixes, but that's a whole different journey and we still want to release the delaying feature to the users. We won't address it with this branch, but will have to keep in mind.

> With (b), items added in static-insert mode will not appear in the bibliography until a full refresh. 
>
> If the cost of extracting citation data from the document is very low, and that of resetting the registry is the highest of the three, maybe we could look at just stripping down the registry update to bare essentials in "static" mode, and otherwise just continuing to do what we do.

Currently this does a full read of the document and update the processor (`rebuildProcessorState()`) upon attempting to edit the bibliography, to avoid problems arising from this. Reading the field codes and text actually seems to be rather low-cost on LibreOffice and likely Word too (although I haven't tested Word as thoroughly in this regard). Either way, doing a full processor refresh for bibliography edits is a reasonable tradeoff IMO.

It still means that there will be discontinuity during citation inserts, but that's an obvious and inevitable result.

> It would also be good to know what profiling shows about the balance of burden between extracting refs from the document (by Zotero integration layer), updating registry records (by update[Uncited]Items()], and generating citations (by [make|process|preview]CitationCluster()).

I have primarily tested with author-date styles and in those cases the biggest burden seems to be actually writing the documents, not updating the processor. I assume footnote based styles will have a higher performance impact both in disambiguation and writes to document.

> Reading through the thread again, another thought is that generating the citation placeholders in a separate processor instance would allow you to play arbitrary games with the formatting of the text itself, which might help with visual hinting

I think this is what we'll go for and specifically choose a preinstalled author-date style.

-----

I'm thinking of replacing bibliography in the document with placeholder text like "Bibliography is unavailable while delayed citation mode is enabled. Disable delayed citation mode in document preferences to see the bibliography". A single delayed citation could escape the eyes of a user during a final review, but they will definitely look at the bibliography and seeing this would be a good reminder. This still doesn't quite address users who use citation styles without bibliography, but I am not sure whether displaying an incorrect bibliography is the way to go. The last thing we want is this feature causing people to submit documents with insanely wrong references.

(Comments @dstillman and others if you have thoughts?)

fbennett:
One way to handle the problem of ambiguous sort ordering might be to disable citation sorting on citations added in the "static" mode, with a notice to the user.

dstillman:
> Bibliography is unavailable while delayed citation mode is enabled. Disable delayed citation mode in document preferences to see the bibliography

Hmm, interesting. But could we not just change it to "Refresh to view bibliography" (maybe with the same wavy underline) when there's a delayed insert, and update on manual refresh? Even with this mode disabled, you have to manually refresh if you move or remove citations, so that's not actually changing, right?

adomasven:
> Hmm, interesting. But could we not just change it to "Refresh to view bibliography" (maybe with the same wavy underline) when there's a delayed insert, and update on manual refresh? Even with this mode disabled, you have to manually refresh if you move or remove citations, so that's not actually changing, right?

We would have to scan all fields to find the bibliography and replace it with the message again after another delayed citation insert, which is what we're trying to avoid in the first place.

dstillman:
You said that was fairly quick, though?

Also, is it not possible to just look at the end? Seems like it would be very rare for someone to have their bibliography elsewhere in the document.

adomasven:
> You said that was fairly quick, though?

From current experience with LibreOffice - yes. Would need to implement and test this more thoroughly.

> Also, is it not possible to just look at the end?

Not with the current state of plugin code. We retrieve all fields in a single go.

Either way, I'll work towards not showing the bibliography while delayed mode is enabled and will see if keeping bibliography there after refresh is viable.

> One way to handle the problem of ambiguous sort ordering might be to disable citation sorting on citations added in the "static" mode, with a notice to the user.

I'll not bother with this for now and we'll see how it goes once people start using the feature.

fbennett:
>> One way to handle the problem of ambiguous sort ordering might be to disable citation sorting on citations added in the "static" mode, with a notice to the user.

> I'll not bother with this for now and we'll see how it goes once people start using the feature.

Understood. Does that mean that you will need to have ``makeCitationCluster()`` extended to apply the style-specified sort to the inserted citation?

adomasven:
> Does that mean that you will need to have makeCitationCluster() extended to apply the style-specified sort to the inserted citation?

Ideally it would do that, but we can just create another citeproc instance to work around this for static inserts.

fbennett:
I've extended the ``makeCitationCluster()``[test fixture](https://github.com/Juris-M/citeproc-js/blob/master/tests/citeproc-js/makecitationcluster.js), and the returned citation is properly sorted. That's nice to know at my end, but if there is a separate processor instance anyway, it should work just as well to stick with the devil you know, and use ``processCitationCluster()`` with empty ``citationsPre`` and ``citationsPost`` arguments.

*However* ... there is another wrinkle. It just dawned on me that if you use a separate processor instance with a dedicated static-mode style, the dummy citations will be sorted by the rules of that style, and not the primary style applied to the document.

adomasven:
> However ... there is another wrinkle. It just dawned on me that if you use a separate processor instance with a dedicated static-mode style, the dummy citations will be sorted by the rules of that style, and not the primary style applied to the document

Good point, I've considered this too, since suggesting that we would use a pre-installed author-date style, so we'll just use the document style for the other instance too.

fbennett:
If it proves desirable after initial testing, it should be possible to replace the citation ``cs:layout`` node in a copy of the original style with a concise author-date construct (to avoid problems with the numeric styles).

adomasven:
Updated with refresh notice. Changed squigly to dashed underline for now.

bwiernik:
I don't think that as a user, I would expect within-citation sorting, disambiguation, collapsing, etc. to occur during static mode, particularly if you are using some sort of placeholder author-date style. I don't really see a problem with showing a citation like this in static mode:
(Miller 2012; Wilson et al. 2008; Jones 2012; Jones 2012; Jones 2015)

adomasven:
Anyone has any comments on this that have been failed to address? Can we proceed with the merge?

rmzelle:
Happy to see this accepted, but two more ramblings:

* Maybe change
  > Updating citations **in this document** is taking a long time. Would you like to delay citation updates until manual refresh?\n\nYou can change this setting later in the document preferences.

  to

  > Updating citations is taking a long time. Would you like to delay citation updates **in this document** until manual refresh?\n\nYou can change this setting later in the document preferences.
 
  That's a little bit more explicit that the setting is document-specific (although the sentence at the end already suggests as much).
  * I'm also still not very keen on the use of "delay", but happy to admit my defeat. But you could also frame it as "Would you like to switch this document from automatic to manual refresh?"
* Would it be possible and a good idea to change the Zotero toolbar somehow to indicate that the  "Delay Citation Updates"-mode is active in the document, e.g. by changing the Refresh button label and/or icon (e.g. to orange arrows). E.g. people might receive documents from a colleague with "Delay Citation Updates"-mode enabled, and a visual cue might help them figure out that the mode is active.

adomasven:
> Would it be possible and a good idea to change the Zotero toolbar somehow to indicate that the "Delay Citation Updates"-mode is active in the document, e.g. by changing the Refresh button label and/or icon (e.g. to orange arrows)? Users might e.g. receive documents with "Delay Citation Updates"-mode enabled from a colleague, and a visual cue in the toolbar might help them figure out that the mode is active.

I've been thinking about something like this myself, but maybe do it after the PR is merged.

bwiernik:
I think a different text label might be sufficient. Currently, the Word add-in uses a stock MSO icon for Refresh. If you would like the icon to change as well, there are several other stock icons that would work well (red looping arrows, arrow loopn with an X or check mark, etc.). The stock MSO icons have the advantage that they resize and recolor correctly automatically.
http://soltechs.net/customui/AllGal.asp

rmzelle:
> I think a different text label might be sufficient.

E.g. "Refresh (manual only)"?

bwiernik:
> E.g. "Refresh (manual only)"?

Yeah, something like that.

dstillman:
I'm not getting correct squiggles in Word 2011:

uldash

adomasven:
> I'm not getting correct squiggles in Word 2011

Fixed

dstillman:
There's a conflict here that needs to be resolved.

adomasven:
Updated

adomasven:
> 1) Insert a citation into a new Word doc, Chicago style, Fields.
> 
> 2) Add a breakpoint at the beginning of _updateCitations().
> 
> 3) Insert another citation, wait 5 seconds, continue from the breakpoint.
> 
> 4) Delay prompt pops up. Say Yes.
> 
> 5) Remove breakpoint, insert another citation, get dashed lines.
> 
> 6) Press Refresh.
> 
> The citations are updated, but the dashes don't go away.

Updated

dstillman:
1) Insert a citation into a new LibreOffice, Chicago style, ReferenceMarks.

2) Add a breakpoint at the beginning of _updateCitations().

3) Insert another citation, wait 5 seconds, continue from the breakpoint.

4) Delay prompt pops up. Say Yes.

5) Add a third citation.

The citation is inserted immediately but does not have any underline to indicate that it's provisional.

adomasven:
macOS LO plugin seems to fail to insert RTF properly. I'll look into it, but it's a separate issue present on master too. I wouldn't stall this branch because of that (and nobody having reported this bug indicates that perhaps macOS LO combinations are fairly low, and no wonder, given how terribly slow LO is on macOS)

dstillman:
I don't think many styles use italics/etc. in in-text citations, so it's possible that to notice this someone would've had to manually add italics to a citation in classic view, which probably isn't very common.

But at least for this we have the warning in the bibliography.

I'll do some more testing in Word.

dstillman:
I actually see the dashed line briefly — in text with no gray background — while the citation is being inserted, and then it switches to the gray field with no dashed line.

adomasven:
Actually, reinstall the plugin via Zotero and try again with LO.

dstillman:
1. I'm getting the underlines now (but still having <a href="https://github.com/zotero/zotero-libreoffice-integration/issues/35#issuecomment-357780732">other RTF trouble</a>), but now the lines aren't disappearing after the first Refresh, even though the citations are being updated. After the second or sometimes third Refresh the lines are going away.

2. It seems like, when this mode is on, Insert Bibliography will insert a bibliography even without refreshing, which seems dangerous, because you might still have placeholder citations. (If you then insert another citation, the bibliography switches to the warning.) Inserting a bibliography in delayed-update mode should probably trigger an automatic refresh.

3. I got [the same error](https://github.com/zotero/zotero/pull/1242#issuecomment-322728956) I was getting before (`this.state.registry.registry[Item.id] is undefined`) a couple times, though I can't reproduce it now. I'll keep trying.

dstillman:
Unchecking the checkbox in the prefs and clicking OK also should trigger an automatic refresh, probably.

adomasven:
> I'm getting the underlines now (but still having other RTF trouble), but now the lines aren't disappearing after the first Refresh, even though the citations are being updated. After the second or sometimes third Refresh the lines are going away.
 
Yeah, there was a bug with removing uldash styling, now fixed.

> It seems like, when this mode is on, Insert Bibliography will insert a bibliography even without refreshing, which seems dangerous, because you might still have placeholder citations. (If you then insert another citation, the bibliography switches to the warning.) Inserting a bibliography in delayed-update mode should probably trigger an automatic refresh.

The internal session is reloaded and the bibliography is inserted only from existing citations, but I've changed it so that it also updates all the delayed citations as well, since it makes sense to do that upon insertion (and bibliography editing too).
 
> Unchecking the checkbox in the prefs and clicking OK also should trigger an automatic refresh, probably.

It does, but there was a bug with removing uldash styling. Now addressed

Issue #1232, opened 2017-05-23
Symlink entire directories in build system


dstillman:
Directories in `symlinkDirs` should be symlinked directly instead of symlinking individual files. E.g., no need to symlink 400+ translators.

PR #1248, opened 2017-06-19
Better build process. Fix #1238, #1232

### LINKED ISSUE #1238: Don't clear build directory on every gulp run
gulp-watch is nice, but running gulp from scratch is still slow (and it can be necessary not only when you're sitting down to work but also if, say, you introduce a syntax error and gulp bails (though maybe we want to fix that too)). Instead of clearing the build directory each time, we could store an index with source file mtime, mode, and type (dir/file) and only reprocess the files that have actually changed, along with deleting any files in the index that are no longer present in the source.

tnajdek:
This is a new, re-written build process to make it less cumbersome. Here are the major changes:

* Remove gulp, replace with custom scripts
* Symlink entire dirs where possible (#1232)
* Significantly speed up subsequent builds (macOS and Linux only) (#1238)
* Watch process now observes new/removed files, not only changed. Should now remove any temporary/editor backup etc. files as soon as these disappear from the source 
* Better console output during builds


dstillman:
This is great. Just for house style, can we rename trailing abbreviations to be capitalized (`getJs` -> `getJS`) and add spaces after statements (`if (`, `while (`, `for (`) to differentiate from functions?

scripts/utils.js
```diff
@@ -0,0 +1,123 @@
...
+	&& ['mode', 'mtime', 'isDirectory', 'isFile'].reduce((acc, k) => {
+		return acc ? k in a && k in b && a[k] == b[k] : false;
+	}, true);
+}
+
+function getPathRelativeTo(f, dirName) {
+	return path.relative(path.join(ROOT, dirName), path.join(ROOT, f));
+}
+
+const formatDirsforMatcher = dirs => {
```
dstillman: Capital `F`

dstillman:
And should probably log a warning for earlier Node versions, so people know they're getting an suboptimal experience.

scripts/utils.js
```diff
@@ -0,0 +1,123 @@
...
+	msg += ` [${yellow(`${result.processingTime}ms`)}]`;	
+
+	console.log(msg);
+}
+
+function onProgress(sourcefile, outfile, operation) {
+	if('isError' in global && global.isError) {
+		return;
+	}
+	console.log(`${colors.blue(`[${operation}]`)} ${sourcefile} -> ${outfile}`);
```
dstillman: I think we can remove the ` -> ${outfile}`. All it's ever adding is `build/` at the beginning, right? Easier to read without that.

dstillman:
The `npm i` on Travis is [failing](https://travis-ci.org/zotero/zotero/jobs/244507026#L179) with `Cannot find module 'array-union'`. I actually saw that locally, but then I think I cleared `node_modules` to fix it. But clearing the cache on Travis doesn't seem to be doing the trick.

tnajdek:
Updated PR:

* Adjusted naming convention & keyword-spacing
* Tweaked signatures code to work correctly in node 7.x
* Simplified `onProgress` output, unless `NODE_ENV=debug
* Introduced ignoreMask for easy file-exclusion

> The npm i on Travis is failing with Cannot find module 'array-union'. I actually saw that locally, but then I think I cleared node_modules to fix it. But clearing the cache on Travis doesn't seem to be doing the trick.

I'm struggling to reproduce that locally. I can't think of the reason other than travis cache.

dstillman:
OK, working after clearing the `master` cache. (Previously I only cleared the PR cache. Not sure why that didn't work.)

PR #1266, opened 2017-07-17
Update newly introduced messages to Spanish

febrezo:
Minor changes also include `...` for `…` and opting for the use of the `«»` which are preferred in Spanish.


dstillman:
Thanks for working on this! Localization changes need to be made [on Transifex](https://www.transifex.com/zotero/zotero), so you'll need to join the Spanish team there and apply the changes that way.

PR #1278, opened 2017-08-02
Server-Side Events

adomasven:
Upgrading X-Zotero-Connector-API-Version to 3 since an addition of SSE
is a big enough difference and we'll want to push the upgrade of Zotero
more aggressively later, when the Connector expects SSE support, but the
client doesn't have it.

Various advanced connector features depend on this, so the sooner it's out, the better.


dstillman:
What are the implications for the version change? Will the old connector keep working once we push this out?

adomasven:
> What are the implications of the version change?

None at the moment. However, I'd hope that we can rely on the connectors being updated more frequently than the client and we'll later be able to push people into upgrading the client when it gets really awful to work without SSE available.

adomasven:
Requires a modification to reports endpoint to accept plaintext, because this update significantly improves the look of error and debug reports and also includes error/debug info from the client if it is running. The text of the report is generated on the client side.

See https://github.com/zotero/zotero/pull/1278/commits/7e47df470efcea5943ad7d8fad18f886191167d0#diff-740a54408ff0b5170a5695feae63baf0R269

adomasven:
Updated. Please review and let's not shelve this again for 3 years.

dstillman (changes requested):
Before I review this more fully, can we make a few changes?

1) Rename server_connectorSse.js and server_connectorSseTest.js to server_connectorSSE.

2) Remove the proxies stuff and save it for a separate PR, unless there's a very strong reason it needs to be part of this. (Among other things, schema updating is currently unreliable, for unknown reasons.)

3) Remove the stuff for logging the connector in the client, as we discussed. The connector will always log the client when available. The client will never log the connector, for privacy reasons and simplicity of submitting and viewing. When we want something that requires connector output, we'll just ask for a connector Debug ID, and it will include both.

adomasven:
> Rename server_connectorSse.js and server_connectorSseTest.js to server_connectorSSE

Done

> Remove the proxies stuff and save it for a separate PR, unless there's a very strong reason it needs to be part of this. (Among other things, schema updating is currently unreliable, for unknown reasons.)

I didn't remove most of the code, just disabled/commented it out, since after rebasing on current master the commit changes do not perfectly correspond with added features.

> Remove the stuff for logging the connector in the client, as we discussed. The connector will always log the client when available. The client will never log the connector, for privacy reasons and simplicity of submitting and viewing. When we want something that requires connector output, we'll just ask for a connector Debug ID, and it will include both.

I have disabled debug log collection from connectors from Zotero. The report log is still included if submitting a report ID. The connector report log is included after Zotero's own report and the boundary between the first and the second is very clear in the report log. If you think including that causes privacy concerns then it can be disabled too.

dstillman:
> since after rebasing on current master the commit changes do not perfectly correspond with added features

Not sure what you mean by that — you can rebase and edit the original proxy commit and use `add -p` to create a separate commit with just proxy stuff that can be pulled out — but the commit disabling stuff should at least be squashed into the original proxy commit so there's not an intermediate state where it's active.

> If you think including that causes privacy concerns then it can be disabled too.

Yes. Connector error reports could include URLs, and there's really no reason for those to be included from the client. For any problem related to the connector, we should just ask for a report from the connector and get both that way.

adomasven:
Updated

adomasven:
The branch commits reset, connector log fetching removed completely, some commits combined.

PR #1356, opened 2017-11-04
Support short DOIs in cleanDOI() and update DOI copyAsURL prefix

bwiernik:


chrome/content/zotero/xpcom/utilities.js
```diff
@@ -311,7 +311,7 @@ Zotero.Utilities = {
 			throw "cleanDOI: argument must be a string";
 		}
 
-		var doi = x.match(/10\.[0-9]{4,}\/[^\s]*[^\s\.,]/);
+		var doi = x.match(/(?:10\.[0-9]{4,}\/[^\s]*[^\s\.,])|(?:10\/[^\s]*[^\s\.,])/);
```
dstillman: Seems like we could shorten this to `(?:10(?:\.[0-9]{4,})?\/[^\s]*[^\s\.,])`? As in, just make the `.<number>` between the `10` and the `/` optional?

bwiernik: Good point. Made that change.

dstillman:
Thanks!

Issue #1361, opened 2017-11-08
Support short DOI lookup


dstillman:
https://github.com/zotero/zotero/pull/1356 added parsing for [short DOIs](http://shortdoi.org/), but lookup doesn't actually work. This probably requires our new CrossRef REST API translator, since it doesn't look like the old API supports short DOIs.

(Side note: I'm not sure what the deal is with Airiti, since it seems like it just ends up making another doi.org request, but it's URL-encoding the DOI in its ```http://doi.org/<doi>``` requests, which causes them to fail. This only shows up when the DOI isn't recognized by earlier translators.)

bwiernik:
http://doi.org has acknowledged that this is a bug in their resolver, but aren't confident that it will be fixed in any sort of timely manner.

bwiernik:
This has been fixed by http://doi.org/ now. shortDOI links resolve correctly.

dstillman:
(This might have been fixed by our recent switch to using content negotiation via doi.org instead of the old Crossref API.)

PR #1362, opened 2017-11-09
Fix shortDOI for doi.org resolver

bwiernik:
The https://doi.org/ resolver doesn't properly handle short DOIs that have been passed through encodeURIComponent()

The resolver works find for unescaped short DOIs:
https://doi.org/10/aabbe

But if the short DOI is escaped, the resolver fails:
https://doi.org/10%2Faabbe

It interprets that, for some reason, as:
https://doi.org/10/10/aabbe

This patch gets around that bug by only capturing the part of a short DOI after "10/", which is all that is needed to resolve a short DOI URL:
https://doi.org/aabbe

### DIFF
--- a/chrome/content/zotero/xpcom/utilities.js
+++ b/chrome/content/zotero/xpcom/utilities.js
@@ -311,7 +311,11 @@ Zotero.Utilities = {
 			throw "cleanDOI: argument must be a string";
 		}
 
-		var doi = x.match(/10\.[0-9]{4,}\/[^\s]*[^\s\.,]/);
+		if(/10\/[^\s]*[^\s\.,]/.test(x)) {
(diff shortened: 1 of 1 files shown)


dstillman:
I don't think this is the right approach. `cleanDOI()` can be used in many different contexts, and it should still extract a valid DOI. The place to fix this would be in any lookup code that hits doi.org (where it could not run `encodeURIComponent()` on something that resembles a short DOI) or the CrossRef API (which, as I note in https://github.com/zotero/zotero/issues/1361, probably means swapping in our new CrossRef REST API translator, which we haven't yet published).

bwiernik:
Thanks. Wasn't sure which would be the best way to fix and thought this would be the easiest way to ask. I'll fix it in the DOI URL creation to not encode short DOIs.

PR #1364, opened 2017-11-11
Ensure that short DOIs are resolved correctly

bwiernik:
An alternative would be to encode short DOIs as:
`"http://shortdoi.org/" + encodeURIComponent(doi)`

But this approach seems less prone to breakage on the part of doi.org

I'm not sure what to do with the `_mapTag` function in `openurl.js` when it is passed a DOI item. This is mostly a concern for the CrossRef locator engine, which currently breaks when it passes a short DOI to the doi.org resolver. Shoul I add a conditional to `_mapTag` to look for DOIs in `data` and then treat long and short DOIs appropriately? Will being passed unescaped `/` break things if they are present in an OpenURL object (so that I should extract just the part after "10/" in the short DOI)?

### DIFF
--- a/chrome/content/zotero/bindings/itembox.xml
+++ b/chrome/content/zotero/bindings/itembox.xml
@@ -427,7 +427,11 @@
 							// Pull out DOI, in case there's a prefix
 							var doi = Zotero.Utilities.cleanDOI(val);
 							if (doi) {
-								doi = "http://dx.doi.org/" + encodeURIComponent(doi);
+								if(/10\/[^\s]*[^\s\.,]/.test(x)) {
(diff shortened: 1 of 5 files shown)


bwiernik:
@dstillman Regarding the DOI Copy as URL function -- should this copy the non-escaped DOI-URL for all DOIs? I imagine the main use of this is to paste the DOI in text somewhere, such as in a blog post or email, and I would think a non-escaped URL would be better there?

bwiernik:
doi.org has fixed the error on their end and escaped short DOIs now resolve correctly. Question still whether you think that the Copy as URL function should copy un-escaped DOIs (I'm thinking it should)?

Issue #772, opened 2015-06-24
Support download parameters for downloading attachments


aurimasv:
We should support
- Referer header (e.g. https://forums.zotero.org/discussion/50036/zotero-connector-extension-for-chrome-cannot-get-pdf-from-httpprojecteuclidorg/)
- POST + parameters (I don't recall the example from the top of my head right now, but this exists)
- Captcha solving (this one might be quite difficult and it's not very widely needed) (e.g. https://forums.zotero.org/discussion/47644/no-attempt-to-download-pdf-from-aps-journal-website/#Item_29)
- Maybe other arbitrary headers, though I don't have an example that requires this in mind

Should not be terribly messy on 5.0 given that we're now using `options` to pass parameters to `Zotero.Attachments`

psisquared2:
Earlier today I was trying to fix Project Euclid and now I have working code for the Referer problem, which was relatively straightforward to implement.

The Captcha problem no longer exists for the provided example, and I wouldn't be sure what the best way to tackle it anyways. For now I think it's best to leave it.

Do you have a particular example to test the POST problem on? If not, I'd soon make a pull request for the Referer solution.

Concerning arbitrary headers: What would be your suggested interface in the translator? Right now I use
`item.attachments[i].referrer = url;`
which is obviously limited to Referer issues. A more general option would be to pass an object of parameters, e.g.
`item.attachments[i].special_header = {Referer: url};`
which could then directly get passed on in attachments.js like
`Zotero.Utilities.Internal.saveURI(wbp, nsIURL, tmpFile, special_header);`
I guess POST + data would need an additional field in the attachment anyways.

dstillman:
A `headers` property on `item.attachments[i]` and in the `options` objects passed to the Zotero.Attachments functions would be fine. We can ignore POST in the absence of an example.

psisquared2:
This can be closed as SSRN is fixed as https://github.com/zotero/translators/pull/1560 was merged.

Issue #523, opened 2014-08-08
Zotero cannot deal with data directory paths that contain non-ascii characters on Windows


aurimasv:
Zotero stores the user-selected data directory path that it retrieves from `nsILocalFile.persistentDescriptor`. This is supposed to return ACString, which is meant to be treated as binary data (It doesn't store encoding information and is just an array of single-byte chars). Of course we receive it as a `string`, which is transcoded to UTF-8 (although it already seems to have been UTF-8). In any case, `á` turns into `Ã¡` and we can't really restore it back.

I'm not entirely sure why we're using `nsILocalFile.persistentDescriptor`. Supposedly it's needed to retrieve files on Mac, which is why we can't use `path`. What we might be able to do, is use `nsIPrefBranch.setComplexValue` and store the `nsiLocalFile` "object" itself. On Windows, this just seems to be the value of `nsILocalFile.path` (with correct encoding), which may just be the properly interpreted `nsILocalFile.persistentDescriptor`. I'm hoping that on a Mac it's whatever the `nsILocalFile.persistentDescriptor` would be.

aurimasv:
Seems like we can use [nsIScriptableUnicodeConverter](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XPCOM/Reference/Interface/nsIScriptableUnicodeConverter) to convert ACString to AString and back again if needed. Assuming there's still any need to use persistentDescriptor

dstillman:
We now use string paths for the data directory setting, so this is probably no longer a problem.

PR #1375, opened 2017-11-26
Allow translators to specify HTTP Referers for attachments

psisquared2:
The need for this feature was discussed in #772 and in [issue #523 in the translator repository](https://github.com/zotero/translators/issues/523). It allows a translator to set
`item.attachments[i].http_referer`
which will then be used in the HTTP request downloading the attachment. This makes automatic PDF downloads possible e.g. for projecteuclid.org and papers.ssrn.com, which refuse PDF downloads if the Referer is not set. I have working translators for both of these pages, which I will submit in a separate pull request to the translator repository.

This code was compiled and tested on Linux Ubuntu 16.04 64 bit, I tested both the nativeHandlerImport and the externalHandlerImport routine extensively.

### DIFF
--- a/chrome/content/zotero/xpcom/attachments.js
+++ b/chrome/content/zotero/xpcom/attachments.js
@@ -234,7 +234,7 @@ Zotero.Attachments = new function(){
 	
 	/**
 	 * @param {Object} options - 'libraryID', 'url', 'parentItemID', 'collections', 'title',
-	 *                           'fileBaseName', 'contentType', 'cookieSandbox', 'saveOptions'
+	 *                           'fileBaseName', 'contentType', 'cookieSandbox', 'saveOptions', 'http_referer'
 	 * @return {Promise<Zotero.Item>} - A promise for the created attachment item
 	 */
 	this.importFromURL = Zotero.Promise.coroutine(function* (options) {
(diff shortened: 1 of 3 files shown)


dstillman:
Thanks — this isn't the approach [I suggested](https://github.com/zotero/zotero/issues/772#issuecomment-345559938), though.

psisquared2:
I have found no obvious way to pass arbitrary headers in `Zotero.HTTP.loadDocuments` because `loadURI` or `loadURIWithFlags` from [XUL Browser](https://developer.mozilla.org/de/docs/Mozilla/Tech/XUL/browser#m-loadURIWithFlags) do not allow for such a parameter. As far as I am aware, `externalHandlerImport` can be modified easily to support arbitrary headers but `nativeHandlerImport` cannot.

This gives us three options:
1) Implement Referers only (what I did).
2) Implement arbitrary headers and only use them when `externalHandlerImport` is called.
3) Implement arbitrary headers and always use `externalHandlerImport` when the additional header is set, disregarding whether a native handler exists or not.

Option 2) does not seem like a good choice because the header would only be applied when no native handler exists, which is confusing and intransparent to translator developers. At least I would find such a behaviour quite unexpected.

As for option 3), I can't tell which consequences or side effects it would have, though I could definitely implement it instead.

In the absence of an example where arbitrary headers are needed instead of just a Referer, I think option 1) would be the best choice for now.

dstillman:
Sorry for the delay, and thanks again for working on this. I decided to try a simpler option and simply include the referrer automatically for all downloads, which works for Project Euclid. More details in the commit. If it turns out we still need something explicit, we can add that in.

psisquared2:
Thank you for the commit! I have now created a pull request for the SSRN translator: https://github.com/zotero/translators/pull/1560

PR #1417, opened 2018-01-18
Fix PDF tools usage in tests

mrtcode:


test/tests/fulltextTest.js
```diff
@@ -133,7 +134,7 @@ describe("Zotero.Fulltext", function () {
 			toSync.push({
 				item: pdfAttachment,
 				content: "Zotero [zoh-TAIR-oh] is a free, easy-to-use tool to help you collect, "
-					+ "organize, cite, and share your research sources.\n\n",
```
mrtcode: The test was failing because the pdftotext inserts `\n` after `share` when recognizing test.pdf.

test/tests/fulltextTest.js
```diff
@@ -5,6 +5,7 @@ describe("Zotero.Fulltext", function () {
 		// Hidden browser, which requires a browser window, needed for charset detection
 		// (until we figure out a better way)
 		win = yield loadBrowserWindow();
+		initPDFToolsPath();
```
mrtcode: For some tests it's not enough to initialize `Zotero.Fulltext` paths only in `runtests.js`.

mrtcode:
["should store data in .zotero-ft-unprocessed file"](https://github.com/zotero/zotero/pull/1417/commits/758c80432097758601e1a30280b30f835ea399a4#diff-d265828da11423acf2d7b5f75b7a6254L178) is the last of failing tests. The thing is that it needs that PDF tools would be unregistered. PDF tools are uninstalled in the tests block above (#getUnsyncedContent). I'm thinking what was the real purpose of this test?

dstillman:
Merged in ca9a7c6

PR #1448, opened 2018-02-17
Set embedded metadata as Zotero item in Quick Format

fbennett:
This one is probably on the fixit list already, but here goes.

In the current beta of the client, attempting to edit a citation to an item that has been removed from the Zotero database after insertion in the document brings up the Quick Copy popup with no content. The cause was in calling `getItem()` with a single argument, which delivers embedded metadata in (I think) raw CSL-JSON form when a proper Zotero item is not available. With a second argument of `true`, the embedded data is converted to a Zotero item before delivery, and everything works.

### DIFF
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -622,7 +622,7 @@ var Zotero_QuickFormat = new function () {
 	 * Builds the string to go inside a bubble
 	 */
 	function _buildBubbleString(citationItem) {
-		var item = Zotero.Cite.getItem(citationItem.id);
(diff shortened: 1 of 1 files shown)


dstillman:
Thanks — I'll let @adomasven handle this, but the commit message should reference Quick Format, not Quick Copy.

fbennett:
Oops. Fixed. The PR gets the popup going, but it looks like the embedded ID (with slash) turns up as `NaN` in search attempts, and crashes SQLite. Not sure of the details, but in any case this is just a partial solution.

dstillman:
Is this bug still present?

fbennett:
Yes, still present in Zotero-5.0.36-beta.7+ea569b8b1.

Issue #1449, opened 2018-02-20
[Feature Request] Use a custom PDF reader for zotero


ehhc:
Hi,
it would be nice if it would be possible to open PDFs with a custom PDF reader instead of the default one provided by the OS.
When i googled how to do this i noticed that that's a feature requested by multiple users beside me.

dstillman:
Please post all feature requests and bug reports to the Zotero Forums.

PR #1450, opened 2018-02-20
Issue #1449 [Feature Request] Use a custom PDF reader for zotero

ehhc:
Hi,
it would be nice if it would be possible to open PDFs with a custom PDF reader instead of the default one provided by the OS.
When i googled how to do this i noticed that that's a feature requested by multiple users beside me. 

Unfortunately i wasn’t able to build zotero locally yet. My fix is therefore just a guess and not tested. I hope that you might have a look at it anyway. I will try to test it next weekend..


ehhc:
See https://forums.zotero.org/discussion/70494/feature-request-custom-pdf-reader#latest

According to comments there, the failing tests aren't my issue..

ehhc:
During the weekend i tried to test my implementation. Even though I installed a new linux specifically for this use case, I wasn’t able to build the client. I tried to follow this guide (https://www.zotero.org/support/dev/client_coding/building_the_standalone_client) but unfortunately I got a lot of exceptions both during ‘npm I’ and ‘npm run build’. Am I right assuming that you are using nodejs-legacy instead of the newer nodejs?

Independent of the question how it is possible to build zotero, would anybody of you be that kind and test my pull-request? Is it working? Might it be possible to merge it into the master branche?
Thanks a lot

defaults/preferences/zotero.js
```diff
@@ -186,6 +186,9 @@ pref("extensions.zotero.connector.repo.lastCheck.repoTime", 0);
 pref("extensions.zotero.fallbackLauncher.unix", "/usr/bin/xdg-open");
 pref("extensions.zotero.fallbackLauncher.windows", "");
 
+// Custom pdf reader
+pref("extensions.zotero.pdf.launcher", "");
```
dstillman: While we probably won't bother doing this for anything else, it's not impossible, so how about `extensions.zotero.fileHandler.pdf` and generalize the comment to "File handlers"? Also move above the `fallbackLauncher` prefs.

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -1142,7 +1137,25 @@ Services.scriptloader.loadSubScript("resource://zotero/polyfill.js");
...
 		}
-	}
+	};
+     /**
+      * Launch a file with the given application
+      */
+     this.launchFileWithGivenApplication = function (applicationPath, file) {
+         var exec = Components.classes["@mozilla.org/file/local;1"]
+             .createInstance(Components.interfaces.nsILocalFile);
+         exec.initWithPath(applicationPath);
```
dstillman: `var exec = Zotero.File.pathToFile(applicationPath);`

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -1142,7 +1137,25 @@ Services.scriptloader.loadSubScript("resource://zotero/polyfill.js");
...
+         if (!exec.exists()) {
+             throw new Error("'" + applicationPath + "' does not exist");
+         }
+
+         var proc = Components.classes["@mozilla.org/process/util;1"]
+             .createInstance(Components.interfaces.nsIProcess);
+         proc.init(exec);
+
+         var args = [file.path];
```
dstillman: Just `file`, once the calls above are fixed.

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -1111,20 +1118,8 @@ Services.scriptloader.loadSubScript("resource://zotero/polyfill.js");
...
-				}
-				
-				var proc = Components.classes["@mozilla.org/process/util;1"]
-								.createInstance(Components.interfaces.nsIProcess);
-				proc.init(exec);
-				
-				var args = [file.path];
-				proc.runw(true, args, args.length);
+
+                launchFileWithGivenApplication(path, file);
```
dstillman: `file.path`

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -1095,6 +1095,13 @@ Services.scriptloader.loadSubScript("resource://zotero/polyfill.js");
 	 */
 	this.launchFile = function (file) {
 		file = Zotero.File.pathToFile(file);
+
+		var customPDFReaderPath  = Zotero.Prefs.get("extensions.zotero.pdf.launcher");
+		if(file.path.toUpperCase().endsWith(".PDF") && customPDFReaderPath){
+            launchFileWithGivenApplication(customPDFReaderPath, file);
```
dstillman: Pass `file.path` instead of `file` — we're trying to avoid `nsIFile` in new code as much as possible.

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -1095,6 +1095,13 @@ Services.scriptloader.loadSubScript("resource://zotero/polyfill.js");
 	 */
 	this.launchFile = function (file) {
 		file = Zotero.File.pathToFile(file);
+
+		var customPDFReaderPath  = Zotero.Prefs.get("extensions.zotero.pdf.launcher");
+		if(file.path.toUpperCase().endsWith(".PDF") && customPDFReaderPath){
```
dstillman: - Move `customPDFReaderPath` first in the conditional
- `toLowerCase()` and `.pdf`
- Space after `if` and before `{`}

The bigger issue is that technically the extension isn't what we want to use here — we want to use `attachmentContentType` on the `Zotero.Item` and look for `application/pdf`. For now we should probably just do that test in `ZoteroPane.viewAttachment()` (where we do various other tests to figure out what to do) and call `Zotero.launchFileWithGivenApplication()` from there. Down the line we can extract that parsing logic out, but right now that's the only place that calls `launchFile()` where this is actually necessary.

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -1095,6 +1095,13 @@ Services.scriptloader.loadSubScript("resource://zotero/polyfill.js");
...
+            launchFileWithGivenApplication(customPDFReaderPath, file);
+            return;
+        }
```
dstillman: Change all spaces to tabs

defaults/preferences/zotero.js
ehhc: You're correct. That sounds better.

chrome/content/zotero/xpcom/zotero.js
ehhc: I think we still need the file.path as an argument here, won't we? I mean from commandline, i would execute it like this:
´´<<path to pdf reader>> <<path to file>>´´

if i would use the use file as an argument instead of file.path i would try to pass a file to the commandline execution. The way i understand the documentation of nsIProcess that wouldn't work as intendet, would it?

ehhc:
Hi @dstillman , thanks for your feedback. In my last commit i have fixed the issues raised by you.
Unfortunately, there are still two problems:

First and easiest: somehow my git has modified the empty lines in the header of the files during commit. I have no idea why (maybe the line-ending?). I hope that you can simply ignore that during the merge..

Second: Since I'm still not able to build zotero locally, i can test my changes. I only hope that they work. Could you test if my changes do what they are intended to do?

dstillman:
Thanks. This will be in 5.0.36. (I changed pretty much every line, so I just did it as a separate patch.)

PR #1451, opened 2018-02-20
Mendeley import

dstillman:
Accept Mendeley SQLite databases via File -> Import… and perform a direct import, including collections, timestamps, notes, and attachments.

Known fields that aren't valid for a type are stored in Extra. A few fields are not currently mapped.

Files in the Mendeley 'Downloaded' folder are stored. Files elsewhere are linked.

This should not yet be used on a production Zotero database, and imported data cannot yet be synced.

To-do:

- [ ] More testing on real databases (ideally by someone who actually has Mendeley installed)
- [ ] Review data mappings and check for additional things to be mapped
- [x] Automatic DB selection, with an intermediate window when selecting File -> Import… that allows "Mendeley" to be chosen as an import  option
- [x] Test error handling (e.g., if a non-Mendeley SQLite DB is passed)
- [ ] Format Extra fields before passing to citeproc-js (since we're storing them title-cased and with spaces, and citeproc-js supposedly only allows lowercase and hyphens)
- [x] Allow syncing of new predicates, or switch to owl:sameAs with a URI for the UUIDs
- <del>Update or skip items previously imported from Zotero (DocumentsZotero table)</del> Can't do this safely, because only local ids are available in the Mendeley database
- <del>Group support? Group library items are currently ignored.</del> Not going to worry about this.
- [ ] An easier way to move all subcollections of the imported library to the library root, either with a menu option (easy but awkward) or multi-selection for collections
- [ ] Support linked citations from documents created by Mendeley after import
- <del>Annotations/highlights?</del> These are stored in the DB in Mendeley, and we don't have a good way of applying them to the file. It's possible we could figure out a way to do this later or swap in PDFs from an export (which apparently includes the option of embedded annotations).
- [x] Extract annotations?

### DIFF
--- a/chrome/content/zotero/fileInterface.js
+++ b/chrome/content/zotero/fileInterface.js
@@ -23,6 +23,8 @@
     ***** END LICENSE BLOCK *****
 */
 
+Components.utils.import("resource://gre/modules/osfile.jsm")
+
 /****Zotero_File_Exporter****
  **
  * A class to handle exporting of items, collections, or the entire library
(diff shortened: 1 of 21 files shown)


bwiernik:
Regarding citeproc-js, the CSL variable names do indeed need to be formatted using the names as given [here](https://aurimasv.github.io/z2csl/typeMap.xml) with either all lowercase letters or all uppercase letters (e.g., DOI or PMID) and no spaces (mostly hyphens, with an underscore used for `archive_location`).

dstillman:
I added Extra field format conversion in https://github.com/zotero/zotero/commit/1f320e1f5d5fd818e2c2d532f4789e38792a77a2, but this is only for valid CSL fields, so we'll need to update `extraToCSL()` with mappings for any fields that are different in CSL from what they'll be in Extra.

bwiernik:
@dstillman Is import of Mendeley groups technically infeasible, or is there another barrier? I have some colleagues who are Mendeley users for whom this would be a blocking issue for switching.

bwiernik:
I’ll check with my colleagues about how items are handled. 

I think so long as there is a way to import groups libraries and have them indicated, it will be fine. I think an option to select a Mendeley group and select the Zotero group to import it into would be great, especially if there is a button to go to Zotero.org and make a new group.

bwiernik:
Mendeley items in groups work the same as in Zotero—updates in one library don’t change the item in the other.

Issue #1231, opened 2017-05-23
Display progress window for all document updates


adomasven:
Currently the most glaring one is refresh, but also applies to style changes.

Current progress-bar is part of quick format dialog, so this is related to #1223

adomasven:
Closed by #1476

PR #1476, opened 2018-03-30
Doc Integration Endpoint for the Connector

adomasven:
Changes:

- Moved connector communication specific code under `xpcom/connector` (directory was previously occupied by FX connector code and since got removed)
- Added `/connector/document/execCommand` and `/connector/document/respond` endpoints for integration functionality support
- Made all integration plugin invocation calls async (and added a wrapper for old integration plugins that have a sync API).
- Misc changes including support for "Http" fields, "html" citation/bibl output and plugins that do not support endnotes
- Adds a progress bar for integration interactions that do not use the quickFormat dialog

### DIFF
--- a/chrome/content/zotero/integration/progressBar.xul
+++ b/chrome/content/zotero/integration/progressBar.xul
@@ -1,0 +1,51 @@
+<?xml version="1.0"?>
+<!--
+    ***** BEGIN LICENSE BLOCK *****
+    
+    Copyright © 2018 Center for History and New Media
+                     George Mason University, Fairfax, Virginia, USA
+                     http://zotero.org
+    
+    This file is part of Zotero.
+    
+    Zotero is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Affero General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+    
+    Zotero is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Affero General Public License for more details.
+    
+    You should have received a copy of the GNU Affero General Public License
+    along with Zotero.  If not, see <http://www.gnu.org/licenses/>.
+    
+    ***** END LICENSE BLOCK *****
+-->
+<?xml-stylesheet href="chrome://global/skin/" type="text/css"?>
+<?xml-stylesheet href="chrome://global/skin/browser.css" type="text/css"?>
+<?xml-stylesheet href="chrome://zotero/skin/zotero.css" type="text/css"?>
+<?xml-stylesheet href="chrome://zotero/skin/integration.css" type="text/css"?>
+<?xml-stylesheet href="chrome://zotero-platform/content/integration.css" type="text/css"?>
+<!DOCTYPE window SYSTEM "chrome://zotero/locale/zotero.dtd">
+
+<window
+	id="quick-format-dialog"
+	class="progress-bar"
+	orient="vertical"
+	title="&zotero.progress.title;"
+	xmlns:html="http://www.w3.org/1999/xhtml"
+	xmlns="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul"
+	persist="screenX screenY">
+	
+	<script src="../include.js"/>
+	<script src="windowDraggingUtils.js" type="text/javascript"/>
+	<script src="progressBar.js" type="text/javascript"/>
+	
+	<box orient="horizontal" id="quick-format-entry"> 
+		<deck id="quick-format-deck" selectedIndex="0" flex="1">
+			<progressmeter id="quick-format-progress-meter" mode="undetermined" value="0" flex="1"/>
+		</deck>
+	</box>
+</window>
--- a/chrome/content/zotero/xpcom/connector/httpIntegrationClient.js
+++ b/chrome/content/zotero/xpcom/connector/httpIntegrationClient.js
@@ -85,8 +85,7 @@ for (let method of ["activate", "canInsertField", "displayAlert", "getDocumentDa
 
 // @NOTE Currently unused, prompts are done using the connector
 Zotero.HTTPIntegrationClient.Document.prototype._displayAlert = async function(dialogText, icon, buttons) {
-	var ps = Components.classes["@mozilla.org/embedcomp/prompt-service;1"]
-		.getService(Components.interfaces.nsIPromptService);
+	var ps = Services.prompt;
 	var buttonFlags = (ps.BUTTON_POS_0) * (ps.BUTTON_TITLE_OK)
 		+ (ps.BUTTON_POS_1) * (ps.BUTTON_TITLE_IS_STRING);
 	
@@ -172,8 +171,7 @@ Zotero.HTTPIntegrationClient.Field.prototype.getText = async function() {
 Zotero.HTTPIntegrationClient.Field.prototype.setText = async function(text, isRich) {
 	// The HTML will be stripped by Google Docs and and since we're 
 	// caching this value, we need to strip it ourselves
-	var wm = Components.classes["@mozilla.org/appshell/window-mediator;1"]
-		.getService(Components.interfaces.nsIWindowMediator);
+	var wm = Services.wm;
 	var win = wm.getMostRecentWindow('navigator:browser');
 	var doc = new win.DOMParser().parseFromString(text, "text/html");
 	this._text = doc.documentElement.textContent;
--- a/chrome/content/zotero/xpcom/connector/server_connector.js
+++ b/chrome/content/zotero/xpcom/connector/server_connector.js
@@ -484,6 +484,12 @@ Zotero.Server.Connector.SavePage.prototype = {
 	 * @param {Function} sendResponseCallback function to send HTTP response
 	 */
 	init: function(url, data, sendResponseCallback) {
+		var { library, collection, editable } = Zotero.Server.Connector.getSaveTarget();
+		if (!library.editable) {
+			Zotero.logError("Can't add item to read-only library " + library.name);
+			return sendResponseCallback(500, "application/json", JSON.stringify({ libraryEditable: false }));
+		}
+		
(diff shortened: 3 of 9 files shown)


chrome/content/zotero/integration/progressBar.xul
```diff
@@ -0,0 +1,50 @@
...
+	id="quick-format-dialog"
+	orient="vertical"
+	title="&zotero.progress.title;"
+	xmlns:html="http://www.w3.org/1999/xhtml"
+	xmlns="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul"
+	persist="screenX screenY">
+	
+	<script src="../include.js"/>
+	<script src="windowDraggingUtils.js" type="text/javascript;version=1.8"/>
+	<script src="progressBar.js" type="text/javascript;version=1.8"/>
```
dstillman: Don't need the version anymore (and they throw an error in Fx60).

chrome/content/zotero/xpcom/connector/httpIntegrationClient.js
```diff
@@ -0,0 +1,194 @@
...
+	Zotero.HTTPIntegrationClient.Document.prototype[method] = async function() {
+		return Zotero.HTTPIntegrationClient.sendCommand("Document."+method,
+			[this._documentID].concat(Array.prototype.slice.call(arguments)));
+	};
+}
+
+// @NOTE Currently unused, prompts are done using the connector
+Zotero.HTTPIntegrationClient.Document.prototype._displayAlert = async function(dialogText, icon, buttons) {
+	var ps = Components.classes["@mozilla.org/embedcomp/prompt-service;1"]
+		.getService(Components.interfaces.nsIPromptService);
```
dstillman: `var ps = Services.prompt;`

chrome/content/zotero/xpcom/connector/httpIntegrationClient.js
```diff
@@ -0,0 +1,194 @@
...
+	};
+}
+Zotero.HTTPIntegrationClient.Field.prototype.getText = async function() {
+	return this._text;
+};
+Zotero.HTTPIntegrationClient.Field.prototype.setText = async function(text, isRich) {
+	// The HTML will be stripped by Google Docs and and since we're 
+	// caching this value, we need to strip it ourselves
+	var wm = Components.classes["@mozilla.org/appshell/window-mediator;1"]
+		.getService(Components.interfaces.nsIWindowMediator);
```
dstillman: `var wm = Services.wm;`

chrome/content/zotero/xpcom/connector/server_connector.js
```diff
@@ -493,7 +491,7 @@ Zotero.Server.Connector.SavePage.prototype = {
 		} else {
 			translate.setTranslator(translators[0]);
 		}
-		translate.translate({libraryID, collections: collection ? [collection.id] : false});
+		translate.translate(libraryID);
```
dstillman: What's the reason for this? Isn't this what the `collections` property is for?

chrome/content/zotero/xpcom/connector/server_connector.js
```diff
@@ -418,12 +418,6 @@ Zotero.Server.Connector.SavePage.prototype = {
 	 * @param {Function} sendResponseCallback function to send HTTP response
 	 */
 	init: function(url, data, sendResponseCallback) {
-		var { library, collection, editable } = Zotero.Server.Connector.getSaveTarget();
-		if (!library.editable) {
-			Zotero.logError("Can't add item to read-only library " + library.name);
-			return sendResponseCallback(500, "application/json", JSON.stringify({ libraryEditable: false }));
-		}
-
```
dstillman: ?

chrome/content/zotero/xpcom/connector/server_connectorIntegration.js
```diff
@@ -0,0 +1,72 @@
...
+Zotero.Server.Endpoints['/connector/document/execCommand'] = function() {};
+Zotero.Server.Endpoints['/connector/document/execCommand'].prototype = {
+	supportedMethods: ["POST"],
+	supportedDataTypes: ["application/json"],
+	permitBookmarklet: true,
+	init: function(data, sendResponse) {
+		if (Zotero.HTTPIntegrationClient.inProgress) {
+			// This will focus the last integration window if present
+			Zotero.Integration.execCommand('http', data.command, data.docId);
+			sendResponse(503, 'text/plain', 'Integration transaction is already in progress')
```
dstillman: Is there a mechanism to avoid needing to restart if the integration client doesn't respond and the session doesn't get closed? I think I ran into that during testing.

dstillman:
The progress bar now has an ugly white section for me:

progress

chrome/content/zotero/xpcom/connector/server_connector.js
adomasven: Bad rebasing. Good catch.

chrome/content/zotero/xpcom/connector/server_connectorIntegration.js
adomasven: It was very easy to trigger with the earlier versions and I've made it much more robust since. If any of the integration plugins acts up during the transaction it makes further transactions impossible without a restart. There is no easy way to solve this. Timeouts or multiple sessions are not OK, since the document may be very big and/or take a long time to process.

dstillman: OK, we can see how it goes. If it becomes a problem, it seems like we could maybe do a timeout based on the time of the last field update or something like that.

adomasven:
Updated

dstillman:
Seeing this in the console after making a selection in Quick Format:

> Timestamp: 4/16/18, 5:09:01 PM
> Error: TypeError: this.onProgress is not a function
> Source File: chrome://zotero/content/xpcom/integration.js
> Line: 2788

dstillman:
In this version, using LibreOffice on macOS, if I go to add a citation it brings Zotero to the front and shows the Quick Format bar on top of that. If I switch back to master, the Quick Format bar shows on top of LibreOffice. (Zotero sometimes flashes to the front but it generally stays behind LibreOffice.)

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -1660,6 +1660,16 @@ Zotero.Utilities.Internal.activate = new function() {
...
+Zotero.Utilities.Internal.sendToBack = function() {
+	Zotero.Utilities.Internal.executeAppleScript(`
+		tell application "System Events"
+			if frontmost of application id "org.zotero.zotero" then
+				set visible of process "Zotero" to false
+			end if
+		end tell
+	`);
+}
+
```
adomasven: > In this version, using LibreOffice on macOS, if I go to add a citation it brings Zotero to the front and shows the Quick Format bar on top of that. If I switch back to master, the Quick Format bar shows on top of LibreOffice. (Zotero sometimes flashes to the front but it generally stays behind LibreOffice.)

Updated, but I'm not sure if we can rely on the stable name of the process "Zotero". Either way, this script should fail gracefully, so maybe not a big worry. This also fixes Zotero staying in the front completely, after interacting from Safari (which I was planning to address later, if there is enough interest/complaints from Safari gdocs users)

dstillman:
In a new doc with Word 2011, I click Add Citation, Zotero and the doc prefs come to the front, I choose APA and OK, and then Zotero becomes hidden — i.e., actually grayed out in the Dock. So I'm left with just Word and "{Citation}".

chrome/content/zotero/xpcom/integration.js
```diff
@@ -2794,6 +2799,13 @@ Zotero.Integration.Progress = class {
 			await Zotero.Promise.delay(300);
 		}
 		this.window.close();
+		// macOS switches focus to last frontmost window of that application when a window is closed,
+		// instead of switching to the last frontmost window from any application
+		if (Zotero.isMac) {
+			Zotero.Utilities.Internal.sendToBack();
+			// Wait a bit to ensure that other window activations are executed in proper order
+			await Zotero.Promise.delay(100);
+		}
```
adomasven: > In a new doc with Word 2011, I click Add Citation, Zotero and the doc prefs come to the front, I choose APA and OK, and then Zotero becomes hidden — i.e., actually grayed out in the Dock. So I'm left with just Word and "{Citation}".

I added this progress meter as an independent Zotero window for various integration operations that used to have no progress indication before. E.g. The progress meter window is displayed when retrieving fields from the document, before showing the document preferences window. The problem that surfaced here

> In this version, using LibreOffice on macOS, if I go to add a citation it brings Zotero to the front and shows the Quick Format bar on top of that. If I switch back to master, the Quick Format bar shows on top of LibreOffice. (Zotero sometimes flashes to the front but it generally stays behind LibreOffice.)

is rooted in how macOS handles closed application windows (described in the code comment above).

We can hide the Zotero window using AppleScript, but then the problem is that if AppleScript takes a long time to execute, the next window to display (the citation, preferences, bibl edit windows) may get hidden too. We can either keep the Zotero window on top while displaying the citation dialog, or we can keep on increasing this delay above (currently 100ms). We can probably go up to 500ms without a significant usability decrease, but if that is not enough, I am out of suggestions.

dstillman: I don't think we should do anything that relies on a specific delay. If we can't do this properly, we should just drop the extra progress window on macOS.

dstillman: I don't think bringing Zotero to the front with a progress bar would be the end of the world for Google Docs, but we shouldn't regress for everything else.

I'm not clear on the consequences of increasing the delay, but we definitely can't risk leaving Zotero in a hidden state with "{Citation}" displayed, and we don't want to actually slow things down further if that's a possibility.

I also still think that "{Updating}" would be clearer, and possibly obviate the need for a progress bar at all, at least in Word/LibreOffice.

adomasven: > I don't think bringing Zotero to the front with a progress bar would be the end of the world for Google Docs, but we shouldn't regress for everything else.

I think this is the correct solution.

adomasven: Updated

adomasven:
Closes #1231

adomasven:
This has been merged

PR #1477, opened 2018-03-31
Track master branch of submodules

tobiasdiez:
References the master branch of submodules so that they can easily be updated using `git submodule update --remote` as e.g. described in https://stackoverflow.com/a/18797720/873661.


dstillman:
Thanks!

PR #1480, opened 2018-04-08
Asyncify Zotero.Relations.getByObject(), called only from item merge.

fbennett:
This is the patch referred to in a recent [post to zotero-dev](https://groups.google.com/forum/#!topic/zotero-dev/6kFmaW5YhtY). To recap, merge of two items was failing for me in a large-ish account (10,000 items), where the "subordinate" partner in the merge had an `owl:sameAs` relation to another library that had not been opened during the session. The merge failed with an "item not loaded" error.

This change async-ifies the failing function `getByObject()`, and swaps out synchronous data object `get()` in favor of `asyncGet()`. Nothing needs to be altered up the execution chain, because the `merge()` function in `items.js` was already a generator, and its call to `getByObject()` is already `yield`-ed. One other call to `getByObject()` inside relations.js needed minor changes to work async.

### DIFF
--- a/chrome/content/zotero/xpcom/data/relations.js
+++ b/chrome/content/zotero/xpcom/data/relations.js
@@ -146,7 +146,7 @@ Zotero.Relations = new function () {
 	 * @return {Object[]} - An array of objects with a Zotero.DataObject as 'subject'
 	 *     and a predicate string as 'predicate'
 	 */
-	this.getByObject = function (objectType, object) {
+	this.getByObject = Zotero.Promise.coroutine(function* (objectType, object) {
 		var objectsClass = Zotero.DataObjectUtilities.getObjectsClassForObjectType(objectType);
 		var predicateIDs = [];
 		var o = _subjectPredicatesByObject[objectType]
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/xpcom/data/relations.js
```diff
@@ -156,13 +156,16 @@ Zotero.Relations = new function () {
...
 		for (let predicateID in o) {
-			o[predicateID].forEach(subjectID => toReturn.push({
-				subject: objectsClass.get(subjectID),
-				predicate: Zotero.RelationPredicates.getName(predicateID)
+			o[predicateID].forEach(Zotero.Promise.coroutine(function* (subjectID) {
+				toReturn.push({
+					subject: yield objectsClass.getAsync(subjectID),
+					predicate: Zotero.RelationPredicates.getName(predicateID)
+				})
+				return toReturn;
```
fbennett: I think the line above is unnecessary. The array accumulates as a side-effect of the function and is returned after the loop completes. The inner return value isn't used for anything.

fbennett: Oops, this breaks tests. Will replace with a *much* simpler change a bit later, plus a test of the conditions that originally triggered the failure in my environment.

fbennett:
@dstillman All tests now pass. Should be ready for a look, when you have time.

dstillman:
Thanks! (Also note that we can use `async`/`await` in new code, which is cleaner and more future-proof than `Zotero.Promise.coroutine(function *)`/`yield`.)

PR #1494, opened 2018-05-04
Adjust expected add and modify array lengths in recognizePDFTest

fbennett:
I won't pretend to understand the underlying code changes: this just adjusts the fixtures to reflect the return values in the current release. As far as I can tell, recognition and conversion are working well in a client that passes the revised fixtures.


dstillman:
Thanks — superseded by https://github.com/zotero/zotero/issues/1495

PR #1501, opened 2018-05-15
Document export-import UI and integration code

adomasven:
I need ideas for what the dialogs for Export and Import should say. Export converts all citations to their field-code representations (and appends doc prefs and bibl style at the end of the doc). Import reverses this. 

Support of this feature has to be implemented for each individual docs plugin separately (and the buttons will only appear if the plugins have support).

Addresses zotero/zotero-google-docs-integration#1

### DIFF
--- a/chrome/content/zotero-platform/mac/bibliography.css
+++ b/chrome/content/zotero-platform/mac/bibliography.css
@@ -1,0 +1,3 @@
+.chevron {
+	font-size: 1.5em;
+}
--- a/chrome/content/zotero/bibliography.js
+++ b/chrome/content/zotero/bibliography.js
@@ -32,6 +32,7 @@
 // Class to provide options for bibliography
 // Used by rtfScan.xul, integrationDocPrefs.xul, and bibliography.xul
 
+Components.utils.import("resource://gre/modules/Services.jsm");
 var Zotero_File_Interface_Bibliography = new function() {
 	var _io;
 	
@@ -174,6 +175,10 @@ var Zotero_File_Interface_Bibliography = new function() {
 				
 				document.getElementById("automaticCitationUpdates-checkbox").checked = !_io.delayCitationUpdates;
 			}
+			
+			if (_io.showImportExport) {
+				document.querySelector('#exportImport').hidden = false;
+			}
 		}
 		
 		// set style to false, in case this is cancelled
@@ -237,7 +242,38 @@ var Zotero_File_Interface_Bibliography = new function() {
 
 		window.sizeToContent();
 	};
-
+	
+	this.toggleAdvanced = function() {
+		var advancedSettings = document.querySelector("#advanced-settings");
+		advancedSettings.hidden = !advancedSettings.hidden;
+		var chevron = document.querySelector('.chevron');
+		chevron.classList.toggle('chevron-down');
+		chevron.classList.toggle('chevron-up');
+		window.sizeToContent();
+	};
+	
+	this.exportDocument = function() {
+		const importExportWikiURL = "https://www.zotero.org/support/kb/export_import_document";
+		
+		var ps = Services.prompt;
+		var buttonFlags = (ps.BUTTON_POS_0) * (ps.BUTTON_TITLE_OK)
+			+ (ps.BUTTON_POS_1) * (ps.BUTTON_TITLE_CANCEL)
+			+ (ps.BUTTON_POS_2) * (ps.BUTTON_TITLE_IS_STRING);
+		var result = ps.confirmEx(null,
+			Zotero.getString('integration.exportDocument'),
+			Zotero.getString('integration.exportDocument.description'),
+			buttonFlags,
+			null,
+			null,
+			Zotero.getString('general.moreInformation'), null, {});
+		if (result == 0) {
+			_io.exportDocument = true;
+			document.documentElement.acceptDialog();
+		} else if (result == 2) {
+			Zotero.launchURL(importExportWikiURL);
+		}
+	}
+	
 	/*
 	 * Update locale menulist when style is changed
 	 */
--- a/chrome/content/zotero/integration/integrationDocPrefs.xul
+++ b/chrome/content/zotero/integration/integrationDocPrefs.xul
@@ -26,6 +26,7 @@
 <?xml-stylesheet href="chrome://global/skin/global.css"?>
 <?xml-stylesheet href="chrome://browser/skin/preferences/preferences.css"?>
 <?xml-stylesheet href="chrome://zotero/skin/bibliography.css"?>
+<?xml-stylesheet href="chrome://zotero-platform/content/bibliography.css"?>
 <!DOCTYPE window SYSTEM "chrome://zotero/locale/zotero.dtd">
 
 <dialog
@@ -81,14 +82,27 @@
 			</radiogroup>
 		</groupbox>
 
-		<vbox id="automaticJournalAbbreviations-vbox">
+		<vbox class="pref-vbox" id="automaticJournalAbbreviations-vbox">
 			<checkbox id="automaticJournalAbbreviations-checkbox" label="&zotero.integration.prefs.automaticJournalAbbeviations.label;"/>
 			<description class="radioDescription">&zotero.integration.prefs.automaticJournalAbbeviations.caption;</description>
 		</vbox>
 		
-		<vbox id="automaticCitationUpdates-vbox">
-			<checkbox id="automaticCitationUpdates-checkbox" label="&zotero.integration.prefs.automaticCitationUpdates.label;" tooltiptext="&zotero.integration.prefs.automaticCitationUpdates.tooltip;"/>
-			<description class="radioDescription">&zotero.integration.prefs.automaticCitationUpdates.description;</description>
+		<!--<vbox id="advanced-separator" align="center">-->
+			<!--<hbox align="center" onclick="Zotero_File_Interface_Bibliography.toggleAdvanced()">-->
+                <!--<label>&zotero.general.advancedOptions.label;</label>-->
+				<!--<label class="chevron chevron-down">➤</label>-->
+			<!--</hbox>-->
+		<!--</vbox>	-->
+		
+		<vbox id="advanced-settings" hidden="false">
+			<vbox id="automaticCitationUpdates-vbox">
+				<checkbox id="automaticCitationUpdates-checkbox" label="&zotero.integration.prefs.automaticCitationUpdates.label;" tooltiptext="&zotero.integration.prefs.automaticCitationUpdates.tooltip;"/>
+				<description class="radioDescription">&zotero.integration.prefs.automaticCitationUpdates.description;</description>
+			</vbox>
+			<hbox id="exportImport" hidden="true">
+				<button label="&zotero.integration.prefs.exportDocument;" oncommand="Zotero_File_Interface_Bibliography.exportDocument()"/>
+			</hbox>
 		</vbox>
+		
 	</vbox>
 </dialog>
--- a/chrome/content/zotero/xpcom/connector/httpIntegrationClient.js
+++ b/chrome/content/zotero/xpcom/connector/httpIntegrationClient.js
@@ -59,12 +59,16 @@ Zotero.HTTPIntegrationClient.Application = function() {
 	this.secondaryFieldType = "Http";
 	this.outputFormat = 'html';
 	this.supportedNotes = ['footnotes'];
+	this.supportsImportExport = false;
+	this.processorName = "HTTP Integration";
 };
 Zotero.HTTPIntegrationClient.Application.prototype = {
 	getActiveDocument: async function() {
 		let result = await Zotero.HTTPIntegrationClient.sendCommand('Application.getActiveDocument');
 		this.outputFormat = result.outputFormat || this.outputFormat;
 		this.supportedNotes = result.supportedNotes || this.supportedNotes;
+		this.supportsImportExport = result.supportsImportExport || this.supportsImportExport;
+		this.processorName = result.processorName || this.processorName;
 		return new Zotero.HTTPIntegrationClient.Document(result.documentID);
 	}
 };
@@ -76,7 +80,7 @@ Zotero.HTTPIntegrationClient.Document = function(documentID) {
 	this._documentID = documentID;
 };
 for (let method of ["activate", "canInsertField", "displayAlert", "getDocumentData",
-	"setDocumentData", "setBibliographyStyle"]) {
+	"setDocumentData", "setBibliographyStyle", "importDocument", "exportDocument"]) {
 	Zotero.HTTPIntegrationClient.Document.prototype[method] = async function() {
 		return Zotero.HTTPIntegrationClient.sendCommand("Document."+method,
 			[this._documentID].concat(Array.prototype.slice.call(arguments)));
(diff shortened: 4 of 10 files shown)


dstillman:
I can play with the wording and layout once this is in.

adomasven:
This is OK to merge. It won't display to current users of the Word plugins, since the plugins need to be updated first. And it won't display to the google docs users, until we release the latest changes (no need to do it yet).

adomasven:
Although I guess it's no use to merge it, until there's a e.g. LO support, to be able to actually see the UI to play around with it. Let's postpone for now.

dstillman:
Just to think about the text here, when would you actually use this? Really just to go back and forth with Google Docs for now? You could in theory go between Word and LO but there's no real reason to do that, right?

adomasven:
Just for Google Docs for now. You could go between Word and LO. This might also be useful for troubleshooting broken docs, since it converts all citations to code text and removes underlying field codes (although probably not something to display in a prompt). 

My hunch is that we will need a short piece of text in the prompt saying something like 

> This will convert all your citations to their underlying code representations and remove the associated field codes. It will allow you to import the document with a different document processor that support Zotero integration.

or

> This will convert your document to a format that will allow you to import it with a different document processor that support Zotero integration. 

and then add a More Info button that leads to the wiki.

adomasven:
This is good to test. Ensure you got the latest version of zotero-standalone-build and connector google-docs submodule, and the newest apps-script is deployed (compatible with earlier versions). Currently support for going from LO to Gdocs and back exists.

adomasven:
@dstillman this is blocking proper gdocs release

adomasven:
Updated with [some text](https://github.com/zotero/zotero/pull/1501/commits/34b343c9d3253171a6f3b0e4b535c211750e2ff1#diff-850b98547696288788d2ee0883e1f077R877)

```
integration.importDocument                          = Import Document?
integration.importDocument.description              = This will attempt to import Zotero citations from the data exported in a different word processor. Do you want to proceed?
integration.exportDocument                          = Export Document?
integration.exportDocument.description              = Exporting the document will allow you to use it with a different Zotero supported word processor and retain citation links. Do you want to proceed?
```

And a More Info button on that takes you to a [kb page](https://www.zotero.org/support/kb/export_import_document):

adomasven:
There is demand for this feature https://forums.zotero.org/discussion/76155/google-docs-to-word-active-citation-transfer

adomasven:
The branch (and import/export concept) updated.

Upon export the document will be prepended with an import marker and import instructions (localized). When the user interacts with the document which has an import marker at the start they will be prompted to import the document.

image

Support for WinWord and MacWord has also been added. zotero/zotero-standalone-build is currently not updated so you will have to manually update all three (LO, WW, MW) modules to the latest master for testing. Google Docs plugin also needs updating (AppsScript to be deployed first, being backwards-compatible).

dstillman:
I tried this in Mac Word but got an error on import:

> Zotero experienced an error updating your document.
>
> Document modified during update @[setCode:field.m:661]

And I was left with corrupted fields, with the visible text beginning with this:

> I(Mobley, 1982)TEM CSL_CITATION

After toggling between fields codes a couple times, it now starts with this in non-code mode:

> I( PRINTDATE \* MERGEFORMAT }TEM CSL_CITATION

Though when I copy it I still get the version with the citation, so I have no idea what's going on.

I have the document if you want it.

adomasven:
Sorry for the brokenness, Mac Word is very very finicky with its fields. Should be fixed now.

dstillman:
The error is fixed, but I think some of https://github.com/zotero/zotero/issues/1682 may be from this branch. I'm seeing insertions happen in totally the wrong place — e.g., I insert a citation, press Return twice, insert another citation, press Return twice, insert a bibliography, and the bibliography appears between the two citations.

dstillman:
Particularly now that we have one fewer button, we should remove the disclosure triangle and just show these things when you return to the prefs.

dstillman:
> This document contains exported Zotero citations. To import it click "Refresh" in the Zotero %S plugin.

Is this right? Aren't we hard-coding this message in the document? So then don't we not want to include the name of the word processor, since you might be importing the document into one of the other programs?

adomasven:
> > This document contains exported Zotero citations. To import it click "Refresh" in the Zotero %S plugin.
> 
> Is this right? Aren't we hard-coding this message in the document? So then don't we not want to include the name of the word processor, since you might be importing the document into one of the other programs?

We are using `application.processorName`, see e.g. https://github.com/zotero/zotero-word-for-mac-integration/commit/68a523b68f7be67ffb822bfa1df43c9a8a558f98#diff-8ba7f3e2295cead220c5b76e212f4b76R378

adomasven:
> Particularly now that we have one fewer button, we should remove the disclosure triangle and just show these things when you return to the prefs.

Addressed

> > This document contains exported Zotero citations. To import it click "Refresh" in the Zotero %S plugin.
> 
> Is this right? Aren't we hard-coding this message in the document? So then don't we not want to include the name of the word processor, since you might be importing the document into one of the other programs?

Addressed

dstillman:
Bib insertion/refreshing is just failing for me now, even in a new document.

New doc, insert citation, choose APA, insert another citation, insert bib. Error:

> You must insert a citation before performing this operation.

Same refreshing an existing document with two citations and a bib.

adomasven:
> Bib insertion/refreshing is just failing for me now, even in a new document.

This was my bug on master. Fixed now.

dstillman:
I managed to break a document — permanently, as far as I can tell — by adding a few newlines between `ZOTERO_EXPORTED_DOCUMENT` and the next line before pressing Refresh. I now have fields with the visible field codes and `{ HYPERLINK https://www.zotero.org/ }` under the hood, but it doesn't seem possible to convert those back into citations no matter what I try.

We should make sure it's not possible to corrupt the document no matter what you do to the header, and you should always be able to recover an exported document by adding `ZOTERO_EXPORTED_DOCUMENT` to the top (even without the next line).

dstillman:
Same if you delete the doc prefs section at the end — that should just reset the settings on import and show the doc prefs window to let you choose a style.

adomasven:
> We should make sure it's not possible to corrupt the document no matter what you do to the header, and you should always be able to recover an exported document by adding `ZOTERO_EXPORTED_DOCUMENT` to the top (even without the next line).

Why? We should display a better error message if the import fails for some reason, but as said above, if we could, we would make this file unreadable/editable to prevent tampering. You are not supposed to edit the document in such a state no matter what and if you do its your own fault. It is very obviously not intended for edits in the state we export it.

dstillman:
> Why?

Because we should try not to corrupt documents and leave them in unfixable states. That includes if someone makes a mistake and hits Enter by mistake before pressing Refresh, but it also includes any unexpected error during the import, which isn't outside the realm of possibility when dealing with three different word processors across multiple platforms. I was already planning to adjust the text to make it clear that you shouldn't edit the document in this state, but it just doesn't seem like it should be possible to leave a document unfixably broken, and the fact that it is suggests that this might be unnecessarily brittle. (Hopefully someone wouldn't re-save the document after getting an error like this, or they'd still have the original, but I'd still rather prevent corruption like this if possible.)

adomasven:
The import code just removes the first 3 paragraphs of text in the document (export adds 3 paragraphs) as long as the first one is "ZOTERO_EXPORTED_DOCUMENT", so I don't know what's up with that, but it seems to be broken on Mac Word and LibreOffice, and we can fix that. Once you start messing with the links (i.e. removing or editing them) I would draw the line though. If you are willing to edit those, then you are at your own risk of corruption and we won't be able to prevent such edits from causing corruption.

adomasven:
> I managed to break a document — permanently, as far as I can tell — by adding a few newlines between `ZOTERO_EXPORTED_DOCUMENT` and the next line before pressing Refresh. I now have fields with the visible field codes and `{ HYPERLINK https://www.zotero.org/ }` under the hood, but it doesn't seem possible to convert those back into citations no matter what I try.

I have been unable to reproduce this if I download and save the document instead of copy-pasting. The import code just removes the 3 initial lines of the document, whether they are empty lines or anything else.

dstillman:
Nothing to do with copying and pasting — this was all in Word. But it's actually unrelated to making edits. I can reproduce this in a new Mac Word doc just by adding an APA citation, adding a bib, exporting, importing, exporting again, and importing again. Blows up on the second import.

adomasven:
> Nothing to do with copying and pasting — this was all in Word. But it's actually unrelated to making edits. I can reproduce this in a new Mac Word doc just by adding an APA citation, adding a bib, exporting, importing, exporting again, and importing again. Blows up on the second import.

Cannot reproduce with the latest version of the Word plugin and latest commit on this branch.

dstillman:
Definitely breaking for me on the second import. New doc, APA, single citation, no bib. Export → import → export → import → kaboom.

adomasven:
> Definitely breaking for me on the second import. New doc, APA, single citation, no bib. Export → import → export → import → kaboom.

MacWord plugin updated to fix this

dstillman:
Import [this file](https://github.com/zotero/zotero/files/3141568/ZOTERO.docx) — exported from Word — in Google Docs, "Refresh", place cursor after the last (Sumner) citation, "Add/edit citation". Type "umn" and choose the Sumner citation from Cited, and add to the doc. I get "Zotero experienced an error updating your document" and this error:

```
Error: InternalError: Cannot find method addNamedRange(string,number).
	at Code (Zotero Google Docs Integration (Beta)):198 (encodeRange)
	at Code (Zotero Google Docs Integration (Beta)):662
	at Code (Zotero Google Docs Integration (Beta)):437
	at Code (Zotero Google Docs Integration (Beta)):423
	at Code (Zotero Google Docs Integration (Beta)):74 (callMethod)
```

adomasven:
> Open [this file](https://github.com/zotero/zotero/files/3141568/ZOTERO.docx) — exported from Word — in Google Docs, "Refresh", place cursor below the last (Sumner) citation, "Add/edit citation". Type "umn" and choose the Sumner citation from Cited, and add to the doc. I get "Zotero experienced an error updating your document" and this error:
> 
> ```
> Error: InternalError: Cannot find method addNamedRange(string,number).
> 	at Code (Zotero Google Docs Integration (Beta)):198 (encodeRange)
> 	at Code (Zotero Google Docs Integration (Beta)):662
> 	at Code (Zotero Google Docs Integration (Beta)):437
> 	at Code (Zotero Google Docs Integration (Beta)):423
> 	at Code (Zotero Google Docs Integration (Beta)):74 (callMethod)
> ```

Should be fine now, although the source of these errors seems to be citeproc-related. See https://github.com/Juris-M/citeproc-js/issues/108

dstillman:
After updating, I could insert another Sumner reference, but I got the same error then trying to export.

dstillman:
Actually, the insert isn't necessary to reproduce. I get the error just from trying to export from Google Docs immediately after importing.

adomasven:
This is technically only a problem with multiple exports and reimports in the Google Docs processor, but I've updated the plugin to fix that too now.

dstillman:
Now working great (at least with APA) between Word and Google Docs.

Trying to import an .odt file exported from Google Docs into LibreOffice fails for me, though:

> The Zotero field code corresponding to this citation, which tells Zotero which item in your library this citation represents, has been corrupted. Would you like to reselect the item?

It seems to fail on the page boundaries, converting the parts before the page break and leaving the text after.

[zotero-gdocs-export.odt.zip](https://github.com/zotero/zotero/files/3156562/zotero-gdocs-export.odt.zip)

This is with 6.2.1.2. I'm upgrading to 6.2.3 to see if that makes a difference.

(I can actually copy the exported text right from LO to Word and import there and it imports fine.)

dstillman:
More bugs, all with Word 16.25 for Mac:

1) Open [this doc](https://github.com/zotero/zotero/files/3165011/gdocs-exported-footnotes.docx.zip) in Word, press Refresh, but click Cancel on the import.

I get an update error with this error logged:

```
TypeError: session.data is undefined

    Zotero.Integration</this.execCommand@chrome://zotero/content/xpcom/integration.js:255:7
```

2) Import the same file. It seems to work, but I get this error in the console:

```
zotero(3)(+0000006): Integration: MacWord16-refresh:'/Applications/Microsoft Word.app/' complete in 2.89s

zotero(4)(+0000000): ZoteroMacWordIntegration: cleanup

zotero(1)(+0000038): NS_ERROR_FAILURE Exception: The operation couldn't be completed. (OSStatus error -1728.) @[restoreCursor:utilities.m:242]

    checkStatus@file:///Users/dan/zotero-standalone-build/staging/Zotero.app/Contents/Resources/extensions/zoteroMacWordIntegration@zotero.org/components/zoteroMacWordIntegration.js:227:9
    Document.prototype.cleanup@file:///Users/dan/zotero-standalone-build/staging/Zotero.app/Contents/Resources/extensions/zoteroMacWordIntegration@zotero.org/components/zoteroMacWordIntegration.js:520:4
    cls.prototype[method]@file:///Users/dan/zotero-standalone-build/staging/Zotero.app/Contents/Resources/extensions/zoteroMacWordIntegration@zotero.org/components/zoteroMacWordIntegration.js:651:12
    Zotero.Integration</this.execCommand@chrome://zotero/content/xpcom/integration.js:271:17
```

3) In the imported document, add a couple new lines after 4 and insert the Sumner citation again. The style of footnote 5 is "Footnote Text", but the four previous footnotes are "Normal". I assume 5 is correct and the others aren't? (Trying this initially I somehow ended up with yet another style in a subsequent footnote, but I can't reproduce that now, nor can I remember what it was.)

dstillman:
Mostly fixed, but still seeing a variation of 3 in Mac Word:

Import the above doc. Export. Import again. For me, the 4th footnote has a style of "Hyperlink". The first three are "Footnote Text". If I insert another Sumner citation after the initial import, before exporting and reimporting, it also is "Hyperlink" after the reimport.

dstillman:
In Word for Mac, every export/import cycle adds two newlines to the end of the document.

adomasven:
> In Word for Mac, every export/import cycle adds two newlines to the end of the document.

The same is true for Google Docs. In LibreOffice a newline at the start of the document is added. The messing with the last newline character tends to throw errors, so I leave those be (although a newline needs to be inserted for every export to append document preferences). The net result is export/import cycles add newlines, but I don't think that's too much of an issue.

dstillman:
Are these not just the newlines we add above the preferences? Can we not remove those before removing the prefs?

adomasven:
When exporting we add a newline at the end of the file and on it insert the document preferences. Upon import we could remove the final paragraph, but anything new was added there, or on the doc preferences line, we would have to be do additional checks or risk removing that too. Also Google Docs disallows removing the final paragraph. Generally I don't think this is an issue at all.

dstillman:
Well, it's not a big problem, but it's clearly a bug, so it'd be nice to avoid (or limit it to a single line) if there's an easy way to do so.

Speaking of newlines, can we add a line between ZOTERO_EXPORTED_DOCUMENT and the explanatory paragraph to make it easier to read? I'll tweak the text, but I don't want to mess with the line counting/removal.

adomasven:
> Speaking of newlines, can we add a line between ZOTERO_EXPORTED_DOCUMENT and the explanatory paragraph to make it easier to read? I'll tweak the text, but I don't want to mess with the line counting/removal.

We can, but we have to decide how many lines we will be inserting at the start of the document, since we will not be able to change that later without risking breaking earlier exported documents. This also requires changing every doc processor plugin.

dstillman:
I'll keep the text a single paragraph — just need one more line than we have now with the extra newline above it.

adomasven:
> I'll keep the text a single paragraph — just need one more line than we have now with the extra newline above it.

Updated

PR #1525, opened 2018-07-04
Update link to import formats documentation

bwiernik:
The existing destination has never really been the appropriate page for this dialogue, as it has never actual listed the supported formats.


dstillman:
Thanks! (And that documentation looks great.)

PR #1535, opened 2018-07-22
Simplify contextual menu wording

gagarine:
Contextual menu do not need to specify the action apply on a "item", the user know the action apply on the selected item(s).

This is what we have now:
screen shot 2018-07-22 at 20 25 16

This is not consistent with OS contextual menu:
screen shot 2018-07-22 at 20 25 29

Not enven on XP ;)
image


gagarine:
I don't see how failed are related to my proposed changes. But I'm happy to fix it if someone can give me a pointer.

dstillman:
Don't worry about the test failures — the tests can be a little unreliable.

Thanks for this. Just so we can visualize it, here's what we'd be looking at:

simplified_options

(I've added "Duplicate Item" → "Duplicate" from zotero.dtd.)

I think these generally make sense. I'm the least sure about the actions at the bottom, which feel a little curt to me, but I may just be used to the longer names. I'd note that we do use these names regularly in support and documentation, and I think the extra precision can be useful in identifying a feature — e.g., it's clearer to describe the feature as "Create Bibliography from Items" than "Create Bibliography" — but the reduced visual clutter in the menu probably makes up for it. In any case, if we're going to change these, we should apply the same change to those three bottom actions in the collection context menu.

Not shown above, but "Create Parent Item" should stay as is, since that's not referring to the clicked item.

(Once this is ready, I'll do a follow-up commit to remove the plural strings and update the code to use the single strings so localizers don't have extra work.)

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -283,26 +283,26 @@ pane.items.remove.multiple		= Are you sure you want to remove the selected items
...
+pane.items.menu.generateReport            = Generate Report…
+pane.items.menu.generateReport.multiple   = Generate Report…
+pane.items.menu.reindexItem					= Reindex
+pane.items.menu.reindexItem.multiple		= Reindex
 pane.items.menu.recognizePDF				= Retrieve Metadata for PDF
 pane.items.menu.recognizePDF.multiple		= Retrieve Metadata for PDFs
-pane.items.menu.createParent              = Create Parent Item
-pane.items.menu.createParent.multiple     = Create Parent Items
+pane.items.menu.createParent              = Create Parent
+pane.items.menu.createParent.multiple     = Create Parent
```
dstillman: Can you revert these two?

gagarine:
Ok it's done. Sorry I forget about this pull request.
The history is quiet messy. Do you want to squash the commit or I can open a new pull request with one single commit if it's easier.

gagarine:
Please, let me know if their is something more I can do to get this merged.

PR #1568, opened 2018-09-12
Follow meta redirects and use the final URL when processing documents

mrtcode:


chrome/content/zotero/xpcom/http.js
```diff
@@ -311,6 +311,47 @@ Zotero.HTTP = new function() {
...
+				if (options.responseType === 'document' &&
+					(!options.numRedirects || options.numRedirects < 3)) {
+					let contentType = xmlhttp.getResponseHeader('Content-Type');
+					if (contentType && contentType.startsWith('text/html')) {
+						let meta = xmlhttp.response.querySelector('meta[http-equiv="refresh"]');
+						if (meta) {
+							let content = meta.getAttribute('content');
+							if (content) {
+								// Parse URL from the content tag
+								let m = content.match(/url[ ="']+(.+?)[ "']/i);
```
dstillman: I don't think this is right. If there are no quotes — as is the case in the [examples on Wikipedia](https://en.wikipedia.org/wiki/Meta_refresh) — this won't match. See [my approach in translation-server](https://github.com/zotero/translation-server/blob/ad454e02e11110a9726d777711414218af3ab03e/src/http.js#L163), which should work with or without quotes (though it doesn't allow a space after `url`).

The translation-server code also extracts the seconds so we can ignore long timeouts (which somebody pointed out can be used for things like auto-logout or long-term refresh). (It does not, however, cap the number of redirects, which seems important. [Issue added.](https://github.com/zotero/translation-server/issues/43))

test/tests/httpTest.js
```diff
@@ -16,6 +16,16 @@ describe("Zotero.HTTP", function () {
 				}
 			}
 		);
+		httpd.registerPathHandler(
+			'/test-redirect.html',
+			{
+				handle: function (request, response) {
+					response.setHeader("Content-Type", 'text/html', false);
+					response.setStatusLine(null, 200, "OK");
+					response.write("<html><head><meta http-equiv=\"refresh\" content=\"2000; url = 'test.html'\"/></head><body></body></html>");
```
dstillman: 2000 → 2 (and we should ignore anything over 15, as noted above)

I think we should also test without the spaces or quotes, since I think those are less common.

chrome/content/zotero/xpcom/http.js
```diff
@@ -311,6 +311,47 @@ Zotero.HTTP = new function() {
 				if (options.debug) {
 					Zotero.debug(xmlhttp.responseText);
 				}
+				
+				// Try to use HTTP redirect, but if url extraction fails
+				// or numRedirects is exceeded, just return the current page
```
dstillman: This can just be "Follow meta redirects"

chrome/content/zotero/xpcom/http.js
mrtcode: It works without quotes too, but, ok, I can add that code from t-s

dstillman: I don't think it would. It requires a space or quote character at the end, so for a standard value — `5; url=https://example.com/` — none of those would be present.

dstillman: (Yep. If you remove the spaces and quotes from the test, it fails.)

mrtcode: Right, I was initially regexing the whole line where content attribute always ends with some quote. That regex would need to mach the end-of-line to work.

chrome/content/zotero/xpcom/http.js
```diff
@@ -311,6 +311,47 @@ Zotero.HTTP = new function() {
...
+				if (options.responseType === 'document' &&
+					(!options.numRedirects || options.numRedirects < 3)) {
+					let contentType = xmlhttp.getResponseHeader('Content-Type');
+					if (contentType && contentType.startsWith('text/html')) {
+						let meta = xmlhttp.response.querySelector('meta[http-equiv="refresh"]');
```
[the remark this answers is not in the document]
mrtcode: `document.querySelector('meta[http-equiv="refresh"]')` in Chrome for uppercase attribute seems to work without a problem, but I still added `i` just in case.

PR #1570, opened 2018-09-15
Fix some typos

zuphilip:


zuphilip:
Let me know, if I should care about the failed checks in Travis.

chrome/content/zotero/xpcom/rdf/n3parser.js
```diff
@@ -31,7 +31,7 @@ $rdf.N3Parser = function () {
 
   var pyjslib_slice = function (str, i, j) {
       if(typeof str.slice == 'undefined')
-        throw '@@ mising.js: No .slice function for ' + str + ' of type ' + (typeof str)
+        throw '@@ missing.js: No .slice function for ' + str + ' of type ' + (typeof str)
```
dstillman: Files in the `rdf` directory aren't mostly third-party files and shouldn't be modified.

chrome/content/zotero/preferences/preferences_export.xul
```diff
@@ -66,7 +66,7 @@
 			
 			<separator/>
 			
-			<label value="&zotero.preferences.quickCopy.siteEditor.setings;" control="quickCopy-siteSettings"/>
+			<label value="&zotero.preferences.quickCopy.siteEditor.settings;" control="quickCopy-siteSettings"/>
```
dstillman: This would lose the strings on Transifex, so let's leave this and the corresponding `locale` entries as is for now.

zuphilip: Hm... I fixed it in all the locals as well see https://github.com/zotero/zotero/pull/1570/commits/9649468042bdaa723710bee73911f42fbee2dce6 . Would we still lose something then?

dstillman: Well, I guess I can do a push to Transifex. I generally don't like to push non-source (i.e., non-English) strings, but if I do a pull immediately before it should be OK.

zuphilip:
I excluded the `rdf` folder here. Please have a look at the new version.

dstillman:
Actually, sorry, no, let's remove the .dtd change. I don't want to risk messing up existing translations. We just don't do those via code.

dstillman:
Thanks!

PR #1571, opened 2018-09-15
Fix some lgtm alerts

zuphilip:
See also https://lgtm.com/projects/g/zotero/zotero/


dstillman:
Thanks, but I'm going to hold off on these — while many of these would be fine, the proper fix here is to use eslint, and many of these lines will need additional changes from that, so I'd rather not create additional churn and risk unnecessary conflicts.

I've just committed an [initial eslint config](https://github.com/zotero/zotero/commit/18054c4bc407240965f13e7d0e3f69477fa73f16). Once we're happy with that (and we've closed some old pull requests), we'll be able to automatically fix many things and start enforcing via CI.

zuphilip:
Okay, ESLint is fine. However, LGTM might be able to detect other issues as well...

PR #1573, opened 2018-09-17
Add universal progress queues

mrtcode:
Any number of progress queues can be created at Zotero client initialization time with `Zotero.ProgressQueues.createQueue`. Queues can only be used with items. `Zotero_ProgressQueue_Dialogs` creates a dialog for each queue  and zoteroPane adds a toolbar button for each queue too. Queues are also waiting for internet connection.

### DIFF
--- a/chrome/content/zotero/progressQueueDialog.js
+++ b/chrome/content/zotero/progressQueueDialog.js
@@ -23,30 +23,23 @@
     ***** END LICENSE BLOCK *****
 */
 
-/**
- * @fileOverview Tools for automatically retrieving a citation for the given PDF
- */
-
-/**
- * Front end for recognizing PDFs
- * @namespace
- */
-
-var Zotero_RecognizePDF_Dialog = new function () {
+var Zotero_ProgressQueue_Dialog = function (progressQueue) {
 	const SUCCESS_IMAGE = 'chrome://zotero/skin/tick.png';
 	const FAILURE_IMAGE = 'chrome://zotero/skin/cross.png';
 	const LOADING_IMAGE = 'chrome://zotero/skin/arrow_refresh.png';
+	
+	let _progressQueue = this.progressQueue = progressQueue;
 	
 	let _progressWindow = null;
 	let _progressIndicator = null;
 	let _rowIDs = [];
 	
-	this.open = function() {
+	this.open = function () {
 		if (_progressWindow) {
 			_progressWindow.focus();
 			return;
 		}
-		_progressWindow = window.openDialog('chrome://zotero/content/recognizePDFDialog.xul', '', 'chrome,close=yes,resizable=yes,dependent,dialog,centerscreen');
+		_progressWindow = window.openDialog('chrome://zotero/content/progressQueueDialog.xul', '', 'chrome,close=yes,resizable=yes,dependent,dialog,centerscreen');
 		_progressWindow.addEventListener('pageshow', _onWindowLoaded.bind(this), false);
 	};
 	
@@ -55,13 +48,13 @@ var Zotero_RecognizePDF_Dialog = new function () {
 	}
 	
 	function _getImageByStatus(status) {
-		if (status === Zotero.RecognizePDF.ROW_PROCESSING) {
+		if (status === Zotero.ProgressQueue.ROW_PROCESSING) {
 			return LOADING_IMAGE;
 		}
-		else if (status === Zotero.RecognizePDF.ROW_FAILED) {
+		else if (status === Zotero.ProgressQueue.ROW_FAILED) {
 			return FAILURE_IMAGE;
 		}
-		else if (status === Zotero.RecognizePDF.ROW_SUCCEEDED) {
+		else if (status === Zotero.ProgressQueue.ROW_SUCCEEDED) {
 			return SUCCESS_IMAGE;
 		}
 		return '';
@@ -93,8 +86,18 @@ var Zotero_RecognizePDF_Dialog = new function () {
 	}
 	
 	function _onWindowLoaded() {
-		let rows = Zotero.RecognizePDF.getRows();
+		let rows = _progressQueue.getRows();
 		_rowIDs = [];
+		
+		_progressWindow.document.title = Zotero.getString(_progressQueue.getTitle());
+		
+		let col1 = _progressWindow.document.getElementById('col1');
+		let col2 = _progressWindow.document.getElementById('col2');
+		
+		let columns = _progressQueue.getColumns();
+		col1.setAttribute('label', Zotero.getString(columns[0]));
+		col2.setAttribute('label', Zotero.getString(columns[1]));
+		
 		let treechildren = _progressWindow.document.getElementById('treechildren');
 		
 		for (let row of rows) {
@@ -113,7 +116,7 @@ var Zotero_RecognizePDF_Dialog = new function () {
 		_progressWindow.document.getElementById('cancel-button')
 			.addEventListener('command', function () {
 				close();
-				Zotero.RecognizePDF.cancel();
+				_progressQueue.cancel();
 			}, false);
 		
 		_progressWindow.document.getElementById('minimize-button')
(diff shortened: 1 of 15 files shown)


mrtcode:
Progress queue calls an async processor function for each item in the queue. The function should throw an error for any kind of failure, or return a text message on success. The message will be displayed next to the item, in the dialog.

When creating a progress queue, a queue id must be set which is later used to select which dialog to open. The same id is used for styling zotero pane toolbar button. I.e. for id `recognize` `#zotero-tb-pq-recognize` must be added to `overlay.css`.

chrome/content/zotero/progressQueueDialog.js
```diff
@@ -217,3 +220,20 @@ var Zotero_RecognizePDF_Dialog = new function () {
 		}
 	}
 };
+
+/**
+ * Create a dialog for each progressQueue
+ */
+var Zotero_ProgressQueue_Dialogs = new function () {
```
dstillman: Should this be Zotero.ProgressQueueDialogs within xpcom instead? The dialog isn't tied to a particular window, and if we allowed multiple Zotero windows you'd want to be able to access a dialog from the toolbar button in any window.

chrome/content/zotero/xpcom/progressQueue.js
```diff
@@ -0,0 +1,281 @@
...
+	let _columns = options.columns;
+	let _processor = options.processor;
+	
+	let _listeners = {};
+	let _rows = [];
+	let _queue = [];
+	let _queueProcessing = false;
+	
+	
+	this.getId = function () {
```
dstillman: `getId` → `getID`

chrome/content/zotero/xpcom/progressQueue.js
```diff
@@ -0,0 +1,281 @@
...
+	};
+	
+	
+	/**
+	 * Stop processing items
+	 */
+	this.cancel = function () {
+		_queue = [];
+		_rows = [];
+		if (_listeners['empty']) {
```
dstillman: `_listeners.empty`

chrome/content/zotero/xpcom/progressQueue.js
```diff
@@ -0,0 +1,281 @@
...
+		_rows.unshift(row);
+		_queue.unshift(item.id);
+		
+		if (_listeners['rowadded']) {
+			_listeners['rowadded'](row);
+		}
+		
+		if (_listeners['nonempty'] && _rows.length === 1) {
+			_listeners['nonempty']();
+		}
```
dstillman: `_listeners.rowadded`, `_listeners.rowdeleted`, `_listeners.nonempty`

chrome/content/zotero/xpcom/progressQueue.js
```diff
@@ -0,0 +1,281 @@
...
+	 * @param name Event name
+	 */
+	this.removeListener = function (name) {
+		delete _listeners[name];
+	};
+	
+	
+	/**
+	 * Adds items to the queue and triggers processing
+	 * @param items {Zotero.Item}
```
dstillman: `@param {Zotero.Item[]} items`

(Same for all of these. Format is `@param {Type} varname - Optional description`.)

chrome/content/zotero/xpcom/progressQueue.js
```diff
@@ -0,0 +1,281 @@
...
+		for (let item of items) {
+			_addItem(item);
+		}
+		_processQueue();
+	};
+	
+	
+	/**
+	 * Returns all rows
+	 * @return {Array}
```
dstillman: Should use `Type[]` for an array. So, e.g., `Object[]` if `_rows` is an array of objects.

chrome/content/zotero/xpcom/progressQueue.js
```diff
@@ -0,0 +1,281 @@
...
+	 * @return {null}
+	 */
+	function _addItem(item) {
+		for (let row of _rows) {
+			if (row.id === item.id) {
+				if (row.status > Zotero.ProgressQueue.ROW_PROCESSING) {
+					_deleteRow(row.id);
+					break;
+				}
+				return null;
```
dstillman: Just `return;`, unless there's a meaningful difference between `null` and `undefined`.

dstillman: Let's add a comment to `addItems()` clarifying this behavior — i.e., that if an item is already processed, it's reprocessed, and if it's in the queue but not done, it's ignored.

chrome/content/zotero/progressQueueDialog.js
mrtcode: One problem is that when xpcom is initializing the dialogs are not available at the time. Therefore we have to wait until UI part is initialized. Unless there is a workaround.

dstillman: Actually I think the dialog should be in xpcom/ too (similar to xpcom/progressWindow.js, the bottom-right save popups). We can just get rid of Zotero_ProgressQueue_Dialogs and move getDialog() to the Zotero.ProgressQueue object, and then lazily initialize a Zotero.ProgressQueueDialog when getDialog() is called.

dstillman: For opening a window within xpcom, whether or not there's a frontmost window:

```javascript
var win = Services.wm.getMostRecentWindow("navigator:browser");
if (win) {
	_progressWindow = win.openDialog("chrome://zotero/content/progressQueueDialog.xul",
		"", "chrome,close=yes,resizable=yes,dependent,dialog,centerscreen");
}
else {
	_progressWindow = Services.ww.openWindow(null, "chrome://zotero/content/progressQueueDialog.xul",
		"", "chrome,close=yes,resizable=yes,dependent,dialog,centerscreen", null);
}
```

(Can do it later, but we should put this in a zotero.js function at some point, next to getActiveZoteroPane.)

chrome/content/zotero/xpcom/progressQueue.js
```diff
@@ -0,0 +1,300 @@
...
+	 * @return {Zotero.ProgressQueue}
+	 */
+	this.getQueue = function (id) {
+		return _queues.find(queue => queue.getID() === id);
+	};
+	
+	/**
+	 * @return {Zotero.ProgressQueue[]}
+	 */
+	this.getAllQueues = function () {
```
dstillman: These can just be `create`, `get`, and `getAll`.

chrome/content/zotero/xpcom/recognizePDF.js
```diff
@@ -341,6 +407,8 @@ Zotero.RecognizePDF = new function () {
 	 * @return {Promise}
 	 */
 	async function _recognize(item) {
+		if(!item.isAttachment()) return null;
+
```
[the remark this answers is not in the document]
mrtcode: Well, `getFilePath()` was failing for me when PDF item recognition was triggered when the previous recognition is already in progress. But that should be solved by additional `canRecognize(item)` and `_processingItemID` checks in `this.recognizeItems`. So probably not needed.

chrome/content/zotero/xpcom/progressQueue.js
```diff
@@ -156,17 +144,16 @@ Zotero.ProgressQueue = function (options) {
...
-		_rows.unshift(row);
-		_queue.unshift(item.id);
+		_rows.push(row);
 		
 		if (_listeners.rowadded) {
 			_listeners.rowadded(row);
 		}
 		
-		if (_listeners.nonempty && _rows.length === 1) {
+		if (_listeners.nonempty) {
```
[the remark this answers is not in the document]
mrtcode: It is, but we don't have a proper event emitter that's would support multiple listeners, unless we would implement that.

chrome/content/zotero/xpcom/recognizePDF.js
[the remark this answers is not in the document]
mrtcode: Yeah, so `canRecognize(item)` fixes that, because if context menu is opened while PDF item is already in progress, until I press "recognize PDF" a new parent is created and selected which is then inserted into recognition queue.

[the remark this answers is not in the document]
dstillman: I still don't understand. Why would non-attachment items ever be passed to recognizeItems?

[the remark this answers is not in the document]
mrtcode: Because when context menu is created the selected item is PDF attachment, therefore there's an option to recognize it. But after the item is processed, and context menu is still open and the recognition is triggered again, it gets the automatically selected recognized PDF attachment  parent.

https://github.com/zotero/zotero/blob/d899134e7c53a94cbbc62e0252174e9ce71f7666/chrome/content/zotero/zoteroPane.js#L4469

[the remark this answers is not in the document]
dstillman: Ah, OK, got it.

PR #1582, opened 2018-10-09
Update metadata

mrtcode:
Uses an item identifier and URL to translate and update metadata. If the identifier is not found between item fields, it tries to resolve DOI by fields like 'title', 'authors', 'year', etc.


chrome/content/zotero/xpcom/utilities.js
```diff
@@ -144,6 +144,238 @@ var CSL_TYPE_MAPPINGS = {
...
+				newItem = item1;
+			}
+			// If only item1 is translated
+			else if (item1) {
+				newItem = item1;
+			}
+			// If only item2 is translated
+			else if (item2) {
+				newItem = item2;
+			}
```
dstillman: These two blocks can be replaced with:

```javascript
else {
    newItem = item1 || item2;
}
```

chrome/content/zotero/xpcom/utilities.js
```diff
@@ -144,6 +144,238 @@ var CSL_TYPE_MAPPINGS = {
...
+		}
+		
+		// Translate by URL if translation by identifier failed
+		if (!newItem && item.getField('url')) {
```
dstillman: It's not strictly necessary, and would be a little annoying to do, but technically we should avoid trying the URL if we already tried the DOI above. We don't have to do that in the initial implementation, but should at least add a TODO to skip URLs we tried from resolved DOIs.

chrome/content/zotero/xpcom/utilities.js
```diff
@@ -144,6 +144,238 @@ var CSL_TYPE_MAPPINGS = {
...
+						creators1[i].lastName !== creators2[i].lastName
+					) {
+						creatorsUpdated = true;
+						break;
+					}
+				}
+			}
+			
+			if (creatorsUpdated) {
+				// Empty creators, because setCreators() doesn't do that bey default
```
dstillman: That's probably a bug, though. I don't think there's a good reason for `setCreators()` to set the number of creators passed and leave the rest.

chrome/content/zotero/xpcom/utilities.js
```diff
@@ -144,6 +144,238 @@ var CSL_TYPE_MAPPINGS = {
...
+			for (let key of fields) {
+				let field1 = item.getField(key) || null;
+				let field2 = newItem[key] || null;
+				
+				if (field1 !== field2) {
+					// accessDate is always new, but it doesn't mean that metadata is changed
+					if (key !== 'accessDate') updated = true;
+					item.setField(key, field2);
+				}
+			}
```
dstillman: Is there a reason not to just use `fromJSON()` for all/most of the above?

chrome/locale/en-US/zotero/zotero.dtd
```diff
@@ -96,6 +96,7 @@
 <!ENTITY zotero.items.menu.mergeItems					"Merge Items…">
 <!ENTITY zotero.items.menu.unrecognize                "Undo Retrieve Metadata">
 <!ENTITY zotero.items.menu.reportMetadata             "Report Incorrect Metadata">
+<!ENTITY zotero.items.menu.updateMetadata             "Update Metadata">
```
dstillman: We use "metadata" in a few contexts, but it's probably a mistake — few people know what that means.

Also, doesn't have to be with this commit, but won't we want this function to retrieve PDFs if they're now available (say, because we resolved a DOI, or the PDF is now available on the page that we're translating anyway)? We could leave that separate, but I think it'd make sense that you'd basically update an item and get everything that's now available for it, both metadata and PDF.

I'd say "Update Item", but we're [removing the extraneous "Item" references](https://github.com/zotero/zotero/pull/1535) from the context menu. "Update Details" doesn't include the PDF.  "Update" alone might work, though it's a little vague.

I guess just "Update" for now, and we can see if we come up with something better? If it's just "Update", do we need multiple plural forms (e.g., for Lithuanian) to reflect the number of items selected (or at least singular vs. plural)?

(And the key can be `updateItem` — if we incorporate PDF handling, I think that becomes the internal name of the feature.)

chrome/skin/default/zotero/overlay.css
```diff
@@ -617,6 +617,15 @@
 	margin-top: 1px;
 }
 
+#zotero-tb-pq-update {
+	list-style-image: url(chrome://zotero/skin/puzzle-arrow.png);
```
dstillman: I guess that icon was added years ago for the Retrieve Metadata context menu on Windows, but the puzzle icon seems pretty specifically for plugins. (The blue arrow is also pretty hard to see against the green.) How about document-search-result.png for now, and create a ticket to replace it with a better icon?

Digression: I'm sort of wondering why we have icons for the progress queues at all. Do we just not want these windows to disappear behind the main window? Are we not able to keep windows on top? (I forget, but I think that might've been broken in Firefox.)

resource/config.js
```diff
@@ -12,6 +12,7 @@ var ZOTERO_CONFIG = {
 	API_URL: 'https://api.zotero.org/',
 	STREAMING_URL: 'wss://stream.zotero.org/',
 	RECOGNIZE_URL: 'https://recognize.zotero.org/',
+	RESOLVE_URL: 'https://resolve.zotero.org/',
```
dstillman: We can use SERVICES_URL for this, which we're already mapping to another API Gateway endpoint for OA PDF lookup.

chrome/content/zotero/xpcom/utilities.js
mrtcode: We are setting fields one by one because we don't want to overwrite the existing item with tags, notes and creation date. Therefore we only update fields that belong to specific item type. And also at the same time we are comparing fields to detect if the item was updated.

chrome/content/zotero/xpcom/utilities.js
mrtcode: Well, this is complicated because the resolved URL only exists in `Zotero.HTTP.request`. Unless the item was successfully translated and the translator set a URL field, but in this case we already have an item.

chrome/locale/en-US/zotero/zotero.dtd
mrtcode: That would be very convenient to have just one context menu option to just update/retrieve everything, but for the beginning it's better to make sure that metadata updating is working good enough.

But there are a few things to discuss.
So if an item doesn't have any PDF, we can just retrieve them, which is simple, but if there already are some PDFs we need a way to update them, which can be complicated. What if user added PDFs manually?

Then another thing to discuss is how are we going to implement undo, especially if we'll combine metadata update with PDF retrieval.

chrome/skin/default/zotero/overlay.css
mrtcode: Our initial idea was that it's just more convenient if you can close and open a progress window, but we can rethink if that's actually true from usability perspective. From technical side, I think there're issues on macOS to keep that progress window on the top.

chrome/content/zotero/xpcom/utilities.js
dstillman: The URL can exist outside of `Zotero.HTTP.request` if calling code handles redirects manually, like we're doing for PDF retrieval. See https://github.com/zotero/zotero/blob/30291328ed0469a28118e7e822c3a1d3e60e8c26/chrome/content/zotero/xpcom/attachments.js#L1691-L1729 for an example (ignoring some of the extra stuff in there).

We would still need to get the URL out of `getItemFromURL()` — e.g., with an optional `onRedirectURL(url)` callback or something like that.

chrome/skin/default/zotero/overlay.css
dstillman: Yeah, I don't really see any reason why you'd need to — or would even want to — close and reopen at least any of the existing ones. You can always move them out of the way.

But it would be nicer with always-on-top. The one difference from other windows, and perhaps the reason we added the icons, is that, if you don't know where the progress window went, there's no good way to get it back through the main UI short of triggering the action again on unprocessed items. So until we can do that, I guess we should keep the icons.

mrtcode: Sure, if cancelling all operations when closing a window isn't a problem then we can do that.

chrome/content/zotero/xpcom/utilities.js
mrtcode: Well, but the code seems intentionally made to clear creators only when passing empty array:
https://github.com/zotero/zotero/blob/master/chrome/content/zotero/xpcom/data/item.js#L1091

dstillman: Yeah, but I still think it's a bug. Calling it with 3 creators, you'd never expect creators 4 through 6 to remain on the item.

chrome/content/zotero/xpcom/utilities.js
mrtcode: So you are saying that we should add a `beforeRequest` callback option to `Zotero.HTTP.request`? That wouldn't be hard, but we are calling it through `Zotero.HTTP.processDocuments` therefore we have to add the same option parameter to this function. And also we have to workaround this: https://github.com/mrtcode/zotero/blob/ec5424d5ae5e4697ed5ecafedb04f4a809314a26/chrome/content/zotero/xpcom/http.js#L890-L896

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -32,6 +32,278 @@
 Zotero.Utilities.Internal = {
 	SNAPSHOT_SAVE_TIMEOUT: 30000,
 	
+	/**
+	 * Get an identifier from the item whether it's in 'extra' or a separate field
+	 *
+	 * @param {Zotero.Item} item
+	 * @return {Object[]} Identifiers
+	 */
+	'getItemIdentifiers': function (item) {
```
dstillman: No quotes for any of these

chrome/content/zotero/xpcom/utilities.js
dstillman: No, we wouldn't need to modify `request()` or `processDocuments()`. We could just do a `request()` with `followRedirects: false` and `responseType: 'document'`, call the callback on each URL, and use `wrapDocument()` like `processDocuments()` does on the final response.

But it's actually even simpler than that, I think. We really just care about the final resolved URL, which we can get from the document returned from `processDocuments()`. So `translateURL()` could just take an `onResponseURL()` callback that was passed the final URL from that document, whether or not it was translatable.

(The alternative would be to not worry about keeping the same response signature the same and return an object with a 'url' property and a 'data' property (which would sometimes be null), but you could imagine various functions that perform HTTP requests taking notification callbacks to monitor/abort/etc. the underlying HTTP requests, so I'm OK with the callback approach.)

mrtcode: I agree that the final URL is enough to have, but the problem is that if HTTP request fails and an exception is thrown we don't know what was the final URL which failed. Therefore if the original URL was a doi.org URL, we will likely query it again and that fail again (because item can have the same doi.org URL in URL field and also we are making the same URL by adding DOI to doi.org URL).

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -94,13 +94,17 @@ Zotero.Utilities.Internal = {
 	 * @return {Promise<Object|null>} Item metadata in translator format,
 	 * or null if nothing was translated
 	 */
-	'translateURL': async function (url) {
-		let doc = (await Zotero.HTTP.processDocuments(url, doc => doc))[0];
-		let newItems = await this.translateDocument(doc);
-		if (!newItems.length) {
+	translateURL: async function (url) {
+		try {
```
mrtcode: URL translation errors should be handled because it's expected that some publisher websites will be down therefore this shouldn't prevent to use other metadata update methods. While our identifier resolver API and other identifier APIs should fail hard and stop metadata updating for the current item.

chrome/locale/en-US/zotero/zotero.dtd
mrtcode: The last commit avoids unnecessary localizations for metadata update, only `updateMetadata.title` is left which is needed for progress window title. As I said, I agree that we should combine metadata update with PDF retrieval but that should happen after the metadata updating proves itself.

chrome/content/zotero/xpcom/utilities.js
mrtcode: I started to think that maybe retrying a URL that was already requested through resolved DOI isn't as big problem as we're thinking. On the one hand it would be good to avoid querying the same URL two times, but on the other hand, it depends on the reason why it failed. For example if `Zotero.HTTP.request` failed with `404` error then yes, no need to query again, but if it failed because of [browser offline](https://github.com/mrtcode/zotero/blob/ec5424d5ae5e4697ed5ecafedb04f4a809314a26/chrome/content/zotero/xpcom/http.js#L172) error or something else that isn't related with the specific publisher web page problem, then retrying isn't that bad. So for now I'm just leaving this with "TODO".

chrome/content/zotero/xpcom/utilities_internal.js
dstillman: I don't think the `try` statements belong in these functions. Calling code should be able to know that an attempt failed and handle it however it wants, whether that's retrying or moving on to another method.

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -32,6 +32,283 @@
...
+	},
+	
+	
+	/**
+	 * Update item metadata
+	 *
+	 * @param {Zotero.Item} item
+	 * @return {Promise<boolean>} Returns true if metadata was updated and false if not
+	 */
+	updateMetadata: async function (item) {
```
dstillman: `updateItemMetadata`

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -32,6 +32,283 @@
...
+	 * Combines an existing item with the new metadata.
+	 * The existing item usually has additional data like notes,
+	 * tags, creation date, etc. therefore we can't completely
+	 * replace it by creating a new item
+	 *
+	 * @param {Zotero.Item} item
+	 * @param {Object} newItem Item metadata in translator format
+	 * @return {Promise<boolean>} True if the item was updated and false if not
+	 */
+	combineMetadata: async function (item, newItem) {
```
dstillman: `combineItemMetadata`

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -32,6 +32,283 @@
...
+	},
+	
+	
+	/**
+	 * Resolve an identifier by using fields like title, authors, years, etc.
+	 *
+	 * @param {Zotero.Item} item
+	 * @return {Promise<Object>} Object containing resolved identifiers
+	 */
+	resolveIdentifiers: async function (item) {
```
dstillman: `resolveItemIdentifiers`

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -32,6 +32,278 @@
...
+		
+		if (identifier.DOI) {
+			item2 = await this.translateURL('https://doi.org/' + identifier.DOI);
+		}
+		
+		// If both items are translated
+		if (item1 && item2) {
+			// Combine item1 and item2
+			for (let key in item2) {
+				if (Zotero.Utilities.fieldIsValidForType(key, item1.itemType)) {
```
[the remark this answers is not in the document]
mrtcode: Good point. So we should add creators from publisher website only if creators array, that is from identifier API, is empty.

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -32,6 +32,282 @@
...
+					creatorsUpdated = true;
+					break;
+				}
+			}
+		}
+		
+		// If at least something is updated in creators, set the new creators
+		if (creatorsUpdated) {
+			// Empty creators, because setCreators() doesn't do that by default
+			item.setCreators([]);
```
mrtcode: To remove this, we should merge zotero#1594

chrome/locale/en-US/zotero/zotero.dtd
bwiernik: Perhaps "Update Details" and "Update Details and Find Available PDF" if no attachment is present?

mrtcode: Well, maybe, but in future we would like to update attachments too (if it was a preprint i.e.). Although that can be complicated to achieve in practice.

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -32,6 +32,282 @@
...
+		for (let key of fields) {
+			// Null is when a field doesn't exist
+			let field1 = item.getField(key) || null;
+			let field2 = newItem[key] || null;
+			
+			// If fields aren't identical
+			if (field1 !== field2) {
+				// accessDate is always new, but it doesn't mean that metadata is changed
+				if (key !== 'accessDate') itemUpdated = true;
+				item.setField(key, field2);
```
bwiernik: There are many cases when item metadata needs small manual corrections after translation (e.g., correcting capitalization, removing a period from the end of a title). The function as written will overwrite those changes without notice, correct? I worry that that would be very frustrating—should it trigger the resolve conflicts or merge items dialog instead?

mrtcode: Along with this metadata updating feature, we plan to also have an undo option to return the previous metadata. But of course it can be a little bit hard to track back what exactly changed, especially if it was a small change like a title fix. It might be convenient to have an option to see all changes for each updated item, but that would be difficult to fit into the current metadata update progress dialog (it looks the same as PDF recognizer dialog). Currently it just shows 'updated' and nothing else.

bwiernik: I see that being potentially really frustrating unless it triggers something like the sync conflict dialog to show the changes. 

Updating date, volume, issue, page fields aren’t likely to yield any objections ever. I think the issues might primarily come from title, creators, and book/journal title.

bwiernik: I think just showing the revised content in bold would be sufficient. Perhaps double clicking could open something like the merge dialog or the sync conflict dialog?

mrtcode: Could you please give an example of how the revised content should be presented in a progress dialog row?

I was thinking that the most convenient way to spot changes would be a diff of two citations that are generated from the old and the updated metadata.

Citation 1:
```
Wise, James B. “Argon Laser Therapy for Open-Angle Glaucoma: A Pilot Study.” Archives of Ophthalmology 97, no. 2 (1979): 319–322.
```

Citation 2:
```
Wise, James B., and Stanton L. Witter. “Argon Laser Therapy for Open-Angle Glaucoma” Archives of Ophthalmology 97, no. 2 (1980): 319–322.
```

Result:
alt text

The citations demonstrate quite common metadata differences: incomplete title, missing authors, different publication date.

But of course it doesn't include changes on other fields that are not typically used in citations i.e. Abstract, URL, etc. But from the convenience perspective that would be the quickest way to review changes, in my opinion.

Merge or conflict dialog can be considered in future if there will be a need.

bwiernik: I think that is great. If it uses the default style for Quick Copy or another user-specified style, then it will likely show the the most relevant differences for most users.

dstillman: Yeah, I think that's really nice. I assume this would need to go in a space below each row?

It would take some work, but ideal might be an edit button that opens a panel (like what's used for Add Item by Identifier) with the merge-items interface, that let you choose individual versions of fields. I can't remember how generalizable that is right now — if it takes too much work, it'd be better to reimplement it with React (which we're hopefully almost able to do).

Probably should also be possible to revert with a single button.

mrtcode:
A new update after a long break!

update-metadata

Some information about the recent updates:
1. Shows the diff between the existing item and the new metadata, where all or specific fields are then applied
2. Observes item changes (i.e. when changing fields in the item pane) and updates the diff
3. When `itemType` changes then only all or none of the fields can be applied
4. Double-click on the title to select the item in Zotero pane

Some issues to think about:
1. Diff might be unclear for some people who aren't used to it
2. Better styling/layout is necessary. Probably OS specific styling as well
3. This currently doesn't combine metadata from different sources, uses single source (Crossref, publisher website URL, DOI.org redirect, arXiv, etc.). Not "frankenstaining" metadata is safer but if the source is Crossref there won't be `abstract` field
4. Investigate what are the best ways to display the diffed fields (i.e. diffing chars vs diffing words, showing diffs in separate lines if it's too finicky)
5. Previously mentioned "all or nothing" mode for some items where `itemType` changes might be unclear for some people

mrtcode:
Updates:
- Cleaner and more aligned look which reminds item pane
- Metadata for each item can be applied individually
- Does diff on words (separated with whitespace), which should be more human readable

bwiernik:
Call Number should probably not be cleared ever. It's usually a user-specified field. It would also be really nice to be able to unstage individual field changes (e.g., to prevent a field from being deleted).

dstillman:
> It would also be really nice to be able to unstage individual field changes (e.g., to prevent a field from being deleted).

That’s already implemented. Clicking a row toggles the change off.

dstillman:
> ReferenceError: oldItem is not defined
>
>    _processQueue@chrome://zotero/content/xpcom/updateMetadata.js:235:9

(eslint would catch this, by the way.)

dstillman:
disabled-fields

Why are these disabled by default? Removing the abstract shouldn't even show up as an option, but everything else should be enabled — they're not meaningful changes, but they're what you get from a DOI lookup, and I think we want to offer that by default.

More generally, we need tests for which fields get ignored and which get disabled/enabled by default, and then we can write more intelligent rules (e.g., ignore a date change if it results in fewer parsed date parts, combine Extra values in a more sophisticated way). It looks like `_setRowFields()` and the rest of Zotero.UpdateMetadata depends on UI, but ideally these tests wouldn't need to open a window to run.

(I can provide guidance on setting up tests in general.)

mrtcode:
Most of the points from your previous comment are fixed. 
I didn't have a chance to explain this yesterday but I think it could make sense to show some fields in gray.

So here is the priority of metadata sources:
1. Try DOI, PMID, arXiv
2. If not then https://doi.org/DOI
3. If not then URL
4. If not then resolve identifiers (currently DOI only) and try them

The existing fields will be removed/replaced when:
- Different sources has different completeness metadata, so which source you will hit depends on the priority described above and the luck. But most of the time it will be Crossref and it's not super complete
- There might be your own, manually entered metadata


So my idea was to detect risky individual field changes, or the whole metadata and show them in gray. This way more advanced users have a chance to review metadata and get it as it is in the source, and those that just want to click "Apply All/Updates" will be left on the safe side and only non-risky changes will be applied.

A few examples what is non-risky:
1) Non conflicting - doesn't remove, doesn't replace the existing data
2) Longer than the existing field or even includes the existing field data
3) Item type doesn't change to `webpage`
4) Title is not too different
5) Creator first names are not shorter
6) etc.

Well, I agree, it's annoying to review the metadata and a completely automatic approach would be much nicer for the user, but at the moment I see too many unexpected situations. Maybe as we get more experience on how the old metadata conflicts with the new one and which fields are safer than others, we will have more confidence in replacing the existing fields.

**Apply all vs individual item updates**
Next, I think it might be useful to update individual items if you don't want to wait for other items to be processed or don't have time for reviewing them properly. And I guess "Apply All/Updates" shouldn't be visible when in progress, because metadata which is not reviewed yet might be applied.
In some cases if it hits publisher websites it can be very slow, or the failed requests have to be rescheduled.

mrtcode:
In the previous comment I haven't mentioned that specifically DOI identifiers are used to resolve and combine two metadata items: 1) one is translated by using it as an identifier AND 2) another is by resolving publisher's website over `https://doi.org/DOI`. Fields that doesn't exist in the first item are added from the second one. It usually only adds `abstract` field since `Crossref` doesn't have it. So we are kind of already producing "Frankensteins" but, the metadata is from reliable sources, so it shouldn't be a problem.

Field value text can be selected now.

Metadata and individual field check rules can now be easily added to: https://github.com/zotero/zotero/blob/70ae81f9229405c1fe6b17b868a830ba372a675b/chrome/content/zotero/xpcom/updateMetadata.js#L326-L388
Some basic tests are added as well.

So the questions that are still open:
1) Individual field updates vs "apply all"
2) Is new field editing in place necessary?
3) If editing necessary then show an input in a popup or in a vertically expanded space between fields?
4) How about creators editing?
5) `accessDate` - keep old or set the new one? What if only one new field from multiple was applied?

bwiernik:
> accessDate - keep old or set the new one? What if only one new field from multiple was applied?

This field is primarily used in citations to indicate the date that the item was verified to be available, so I suggest updating the date always.

janosh:
@mrtcode Thanks for working on this! Will be super useful once it lands.

zhuligs:
Does it have any '**beta**' version that I can use to test this feature?  This is awesome.

dstillman (commented):
OK, still playing with this, but some notes:

- Let's put "Update" above "Find Available PDF" if it exists, or in the same space (with separators above and below) if it doesn't. I'm still not sure what we want to call this (see discussion above), but I guess let's say "Update Item" for the moment.
- I don't think we need the Minimize button.
- I think the button and font sizes are too small — this was modeled after the PDF recognition progress window, but the small sizes are sort of less appropriate for this larger window. Might be OK for the list text, but I think at least the buttons and text outside should be normal size.
- font-family of inside buttons should be our standard Lucida Grande stack (the outside ones are correct, so maybe just missing a style sheet)
- Icons
  - Better spinner (spinning circle too small)
  - The exclamation mark for available changes seems a little alarming, though I don't have a better suggestion off-hand
- Instead of the opacity effect on hover for the title, maybe `text-decoration: underline`? Not very app-y, but I don't have a better idea. The opacity is OK for toggling fields, but doesn't seem right for a different action.
- Apply All should be disabled when nothing to apply
- Apply button to the right of Ignore
- Single-click on title to select item in items list
- The translator output is our old `toArray()` format, but `fromJSON()` takes "API JSON", which is similar but different, resulting in some unnecessary debug output lines (e.g., "Discarding unknown JSON object 'notes' for item null/null"). I'm not sure if it causes any data problems. We used to have a function that did this conversion, but it's since been [moved to translation-server](https://github.com/zotero/translation-server/blob/3d11fc9af4e2fe026a084722bb5388bd2994a86a/src/utilities.js#L50). We could restore that or just look through it to make sure we're making whatever changes we need to make to preserve data and avoid unnecessary output logging. That function might do more than we need/want, with separate items for notes, etc., so it's possible we just need to discard `notes` (which we don't need here), `seeAlso` (which won't be used), and `attachments` (which we don't want), but worth checking to make sure there's nothing else we need to do.
- I'm getting the following error for some items, maybe caused by changing item types?

```
Error: Invalid field 'itemType

    _fieldCheck@chrome://zotero/content/xpcom/data/itemFields.js:423:13
    getBaseIDFromTypeAndField@chrome://zotero/content/xpcom/data/itemFields.js:312:5
    _setRowFields/titleIndex<@chrome://zotero/content/xpcom/updateMetadata.js:498:16
    _setRowFields@chrome://zotero/content/xpcom/updateMetadata.js:497:24
    _processQueue@chrome://zotero/content/xpcom/updateMetadata.js:227:1
```

bwiernik:
> The exclamation mark for available changes seems a little alarming, though I don't have a better suggestion off-hand

Perhaps a green circle 🟢 ?

dstillman:
We currently use a green checkmark (which may or may not be right) when there are no changes, so I don't think it could be green.

bwiernik:
Blue yellow or red then?

dstillman:
https://forums.zotero.org/discussion/87881/retrieving-automatic-tags-to-an-existing-record

Are we pulling in automatic tags? We're not showing them in the diff, so I assume not currently? Not totally clear how we'd reconcile those with existing automatic tags. (Existing manual tags could be ignored.)

genghisun:
Hi~ How's it going now? This is such a great feature i want.

gdalle:
Hi @dstillman is this the PR you mentioned here https://forums.zotero.org/discussion/93777/update-metadata-in-bulk ?

dstillman:
@AbeJellinek Can you rebase so this is ready after Z6?

AbeJellinek:
@dstillman Done.

rxy0424:
Thanks for this amazing feature.

rxy0424:
No offense or prodding, but please, since Zotero 6 has been released, when will this feature be merged into the main branch? Thanks for all you.

invisprints:
Hi, I combined your PR with https://github.com/zotero/zotero/commit/7182531ec90376a676fa6fd602ddd1e570232942. I built zotero successfully but it seems not work correctly.
I got the error message:

```
[JavaScript Error: "Error: Localized string not available for pane.items.menu.updateMetadata" {file: "chrome://zotero/content/xpcom/intl.js" line: 131}]

version => 6.0.SOURCE.ad75d9d0e, platform => MacIntel, oscpu => Intel Mac OS X 10.16, locale => zh-CN, appName => Zotero, appVersion => 6.0.SOURCE.ad75d9d0e, extensions => Zotero LibreOffice Integration (5.0.25.SA.6.0.SOURCE.ad75d9d0e, extension), Zotero Word for Mac Integration (5.0.39.SA.6.0.SOURCE.ad75d9d0e, extension), DOI Manager (1.4.2, extension, disabled), Sci-Hub Plugin for Zotero (1.4.2, extension, disabled), Zotero IF (1.4.1, extension, disabled)
```

Could you help me?

dstillman:
@invisprints: Source builds need to be run as en-US.

invisprints:
@dstillman Thx! I added keywords in zh-CN files and it works well in Chinese environment!
I added a PR https://github.com/mrtcode/zotero/pull/1 and hope this could help you!

prototype99:
> > It would also be really nice to be able to unstage individual field changes (e.g., to prevent a field from being deleted).
> 
> That’s already implemented. Clicking a row toggles the change off.

I love the fact deleted fields are automatically deselected, I'd say that seems like a good default. there is one issue though: if the item's updated metadata specifically changes type, for example from a book to a journal article, the ability to individually toggle fields is lost. this can easily be tested by duplicating an item and changing the type then using the "Update" context menu option

AbeJellinek:
Right, that's fair, OK. We can let the user decide whether to accept an emptied Extra field.

F1orian:
Any news on this, or a beta build with this functionality we could test?

dawnlh:
looking forward to this useful feature~

truenomad:
Any update on this feature yet? It's a game-changer!

lucienwang1009:
Any plan to merge this feature?

flinz:
I would very much like this feature to be merged, too! Jumping on the hype train.

achennings:
looking forward to this being merged!

mfranzil:
Is there any news for this? Possibly on the roadmap for Zotero 7?

invisprints:
Hi everyone! While we are waiting for merging the feature, I find another reference management software called [Paperlib](https://paperlib.app/en/). It can scrape and update metadata from many metadata databases. Let's give it a chance.

centerofexcellence:
Is this still a WIP or is it now available?

dwaardenburg:
RIP - My dear PR...

dstillman:
Superseded by #5120

PR #1594, opened 2018-11-12
Clear the existing creators before setting the new ones

mrtcode:


mrtcode:
The last commit makes sure that `this._changed.creators[i] = true;` is not triggered when setting the same authors for an item again. Previously one test was failing because of this.

PR #1609, opened 2018-12-19
Async translators

dstillman:
This is a staging ground for full `async`/`await` support in translators. Both `detect*()` and `do*()` can now be marked as `async` and can use `await` to wait for promise-returning functions. (Import translators could already return a promise from `doImport()` and do incremental saving via promise-returning `item.complete()` functions, but it was hacky and unpleasant.)

This change is made possible by @adomasven's removal of the Mozilla sandbox in favor of a less-secure `eval()`-based sandbox like what we use in the connectors. We're comfortable doing that because 1) we now load remote pages with XHR/DOMParser rather using hidden browsers and running translators in those browsers, so we're not worried about untrusted JS on remote pages potentially getting access to privileged code and 2) we review translator code and can make sure translators are using only the translator API and aren't using `eval()`, so we don't need as strict isolation from globals or other Zotero code that might be accessible somehow. We might still want to do some initial page loads in hidden browsers, but if necessary we can always let the JS run and then reparse the DOM with DOMParser to get a safe static document, at the cost of some efficiency.

Currently, this exposes the promise-based `Zotero.HTTP.request()` as `ZU.request()`, which returns a promise for an XMLHttpRequest object. That may be a bit too verbose to expose directly to translators, so we need to figure out what request functionality translators really need and what it should look like. (CCing @adam3smith and @zuphilip for any thoughts they have on this.) Among other things:

  * `doGet()`/`doPost()` automatically fail translation on non-2xx status codes. We probably want to keep that in general, but some translators could possibly benefit from specifying that 403 or 404 isn't a fatal error. The main question is whether translators actually need a response object that allows them to check the response status code and content type. It may be enough 1) for 403/404 (when explicitly allowed?) to result in a promise resolution of `null` or to throw the promise with an `HTTP.UnexpectedStatusException` that lets you check `.status` without throwing translation and 2) for content type to be ignored or handled with string checks (e.g., does this look like RIS?).
  * We should simplify the `Zotero.HTTP.request()` signature in general (both for translators and for the rest of Zotero) to allow a URL in the first parameter, similar to `fetch()`, and do a GET if not specified in `options`. (Technically a `body` in `options` could even make it default to POST, though `method` would still be possible in `options` for other methods.)
  * `doGet()`/`doPost()` return raw text, which is useful for anything retrieving BibTeX/RIS/etc. It would be nice to be able to just do `var ris = await request(url);`.
  * We can probably stop using `processDocuments()` altogether, since `Zotero.HTTP.request()` supports `response: 'document'`. Instead you'd just do a loop and run the handler function on the `await`ed response. As above, if we didn't return a response object we'd need to decide how to convey that a response either couldn't be parsed as a document or was a 403/404. `null` is probably sufficient for the former, and the latter could be either `null` or a thrown promise. Maybe a translator doesn't care about the distinction and only cares whether it got 1) a 200 and an HTML document or 2) something else?

We won't be able to start using this in translators for probably a month after the first release that includes it, at which time we can cut off translator updates for older versions.

Down the line, it would be nice to get rid of the callback-based functions so that we can remove the horrible `incrementAsyncProcesses()`/etc. stuff in the translator framework.

### DIFF
--- a/chrome/content/zotero/xpcom/translation/translate.js
+++ b/chrome/content/zotero/xpcom/translation/translate.js
@@ -1712,17 +1712,18 @@ Zotero.Translate.Base.prototype = {
 	/**
 	 * Runs detect code for a translator
 	 */
-	"_detectTranslatorLoaded":function() {
+	_detectTranslatorLoaded: async function () {
 		this._prepareDetection();
 		
 		this.incrementAsyncProcesses("Zotero.Translate#getTranslators");
 		
-		try {
-			var returnValue = Function.prototype.apply.call(this._sandboxManager.sandbox["detect"+this._entryFunctionSuffix], null, this._getParameters());
-		} catch(e) {
-			this.complete(false, e);
-			return;
-		}
+		var maybePromise = Function.prototype.apply.call(
+			this._sandboxManager.sandbox["detect" + this._entryFunctionSuffix],
+			null,
+			this._getParameters()
+		);
+		// If detect* returns a promise, wait for it
+		var returnValue = (maybePromise && maybePromise.then) ? await maybePromise : maybePromise;
 		
 		if(returnValue !== undefined) this._returnValue = returnValue;
 		this.decrementAsyncProcesses("Zotero.Translate#getTranslators");
--- a/chrome/content/zotero/xpcom/translation/translate.js
+++ b/chrome/content/zotero/xpcom/translation/translate.js
@@ -352,7 +352,7 @@ Zotero.Translate.Sandbox = {
 			};
 			
 			var translatorsHandlerSet = false;
-			safeTranslator.getTranslators = function() {
+			safeTranslator.getTranslators = async function () {
 				if(!translation._handlers["translators"] || !translation._handlers["translators"].length) {
 					throw new Error('Translator must register a "translators" handler to '+
 						'call getTranslators() in this translation environment.');
@@ -367,7 +367,7 @@ Zotero.Translate.Sandbox = {
 			};
 			
 			var doneHandlerSet = false;
-			safeTranslator.translate = function() {
+			safeTranslator.translate = async function () {
 				translate.incrementAsyncProcesses("safeTranslator#translate()");
 				setDefaultHandlers(translate, translation);
 				if(!doneHandlerSet) {
@@ -378,7 +378,7 @@ Zotero.Translate.Sandbox = {
 					errorHandlerSet = true;
 					translation.setHandler("error", function(obj, error) { translate.complete(false, error) });
 				}
-				translation.translate(false);
+				return translation.translate(false);
 			};
 			
 			safeTranslator.getTranslatorObject = function(callback) {
--- a/chrome/content/zotero/xpcom/utilities_translate.js
+++ b/chrome/content/zotero/xpcom/utilities_translate.js
@@ -280,6 +280,18 @@ Zotero.Utilities.Translate.prototype.processDocuments = async function (urls, pr
 	translate.decrementAsyncProcesses("Zotero.Utilities.Translate#processDocuments");
 }
 
+Zotero.Utilities.Translate.prototype.request = async function (method, url, options = {}) {
+	options = Object.assign(
+		{},
+		options,
+		{
+			headers: Object.assign({}, this._translate.requestHeaders, options.headers),
+			cookieSandbox: this._translate.cookieSandbox
+		}
+	);
+	return Zotero.HTTP.request(method, url, options);
+};
+
 /**
 * Send an HTTP GET request via XMLHTTPRequest
 * 


retorquere:
While it's a much larger change, obviously, could the sandbox be further secured by running translators inside a web worker? This would en-passant separate them from the main thread, which has potential performance benefits.

dstillman:
No, DOM APIs (like DOMParser) aren't available in web workers.

retorquere:
Right. That's shimmable, but probably more effort than it's worth.

dstillman:
Not at native speeds

retorquere:
Is there a prospective timeline for this making it into a public release?

retorquere:
Is there a way to abuse `incrementAsyncProcesses` to occasionally yield in export translators?

retorquere:
Doesn't look like it.

I would really love it if I could get rid of the BBT caches ([cache invalidation](https://martinfowler.com/bliki/TwoHardThings.html) is getting to be a real nuisance), which mostly exist to ameliorate problems caused by BBT doing synchronous work locking up Zotero.

dstillman:
The eval sandbox landed in 5216f1188531627fd7543d2d149b72a9301622c5.

Despite my "somehow" above, a couple minutes of investigation reveals that translators definitely now have access to Zotero internals, so it's critical that translator code is reviewed for inappropriate access and that `eval` is never used.

retorquere:
My translators use `new Function`, for the postscripts... as far as I know, in Javascript the only way to really have safe separation is using workers.

retorquere:
Is https://github.com/zotero/zotero/commit/5216f1188531627fd7543d2d149b72a9301622c5 already in a public release? Or will it be in 5.0.74?

dstillman:
It just landed on master. It will be in 5.0.74, unless we find a problem with it.

> My translators use new Function, for the postscripts...

I'm not familiar with what you're doing, but the point is just that translators should not try to reach outside the (now very lightly enforced) sandbox, and they absolutely can't call `eval` or `new Function` on untrusted content, because that's now a path to full system access.

retorquere:
I use `new Function` for my [postscripts](https://retorque.re/zotero-better-bibtex/exporting/scripting/).

Damn. I'm going to have to think about this. Since the translators can now do async (right? `doExport` can be async in the new situation?) maybe I can shuttle off my postscripts to a web worker.

dstillman:
"Untrusted content" in this context really means content from webpages or imported files. The concern isn't really scripts that people wrote themselves, since they can also just open the Run JavaScript window and write some code to delete all files on their system if they want to do that.

If there's some mechanism for sharing these between users, you would want to be more careful, since it might not be clear that they're running privileged code that can do anything it wants to their system.

> Since the translators can now do async (right? doExport can be async in the new situation?)

This PR hasn't yet landed. So far, we've only changed the sandbox.

retorquere:
Ah OK, none of that applies to me.

Still, since I don't do any DOM work, it could be interesting to see if I can shuttle off work to a worker. Should make auto-exports a lot nicer.

retorquere:
I've recently learned that `new Function` compiles the JS passed in in such a way that the code inside can't see the global context. Translators loaded this way can probably still access `Components` so it's not really a hard sandbox but it'd at least make the surface smaller (and the list of things to check for more limited).

Has this PR made it to 5.0.74? Or does "has not landed yet" stretch beyond that?

AbeJellinek:
Can we target this for 6.0? In my view, the work here is nearly all the way there.

1. We probably want `request` to cause translation to fail in the same cases as `doGet`/`doPost` do now - new behavior would lead to bugs, especially when updating old translators.
2. If this won't be too messy, we might want to add a few functions (`requestText`, `requestJSON`, `requestDocument`?) to the global scope in order to handle the simple `var ris` case which @dstillman mentioned in the PR body.

I am really very excited to be able to write translators without callbacks.

retorquere:
> I'm not familiar with what you're doing, but the point is just that translators should not try to reach outside the (now very lightly enforced) sandbox, and they absolutely can't call `eval` or `new Function` 

Wait -- does that mean I can call all of Zotero from inside import/export translators? That would simplify a few things for me. But the `Zotero` object inside translators isn't the `Zotero` object brought in by `chrome://zotero/content/include.js` right?

dstillman:
I don't think we want to expose the existing `request()` directly — there's too much internal stuff in the options, and translators don't need a real XHR object.

Are we aware of anywhere that translators would need to check either the status code or header (e.g., `Content-Type`) of a `2xx` request? I assume not, in which case we could avoid having a response object altogether and just resolve the promise with a string (response body) or object (parsed JSON/Document).

Thinking about this more, I'm also not sure the concept of a non-2xx error code failing translation makes sense in the context of async translators. In the current functions, [we call `translate.complete(false…)`](https://github.com/zotero/zotero/blob/3584853f3926bfd735700e3430366fa48ad530cc/chrome/content/zotero/xpcom/utilities_translate.js#L321) and just don't call the callback, but here the async request function will need to be resolved or rejected anyway, so maybe an unexpected status code just rejects the promise, and normally a translator wouldn't catch it and that would be a translation failure. But if a translator wanted to catch it and continue, it could, and we can put `status`, `headers`, and `body` on the error object to give the translator more control.

So, taking all that into account:

Required options:

- `headers`
- `body`
- `responseCharset` (```xmlhttp.overrideMimeType(`text/plain; charset=${responseCharset}`)```)

Possible options:

- `method`
  - Or if we only need `GET` and `POST`, `body` can just imply `method: 'POST'` 
- `responseType` (`text` (default), `json`, or `document`)
  - `json` and `document` would be syntactic sugar for `JSON.parse()` and `DOMParser`
  - We probably don't need `arraybuffer` or `blob`

Without a response object, we could do `await request(url[, options])`, which would by default make a GET request and resolve to a string, and we could optionally add `requestJSON(url)` and `requestDocument(url)`, which would be syntactic sugar for the syntactic sugar that is `responseType`.

(We could also choose a name other than `request()` for the translator-specific method, to avoid confusion with the rest of the codebase, but I don't have any great ideas. `fetch()` would be misleading. We could just have separate `get()` and `post()` functions with identical signatures, but those are a little vague. Most people working on translators won't be looking at the main Zotero codebase, so maybe this doesn't matter.)

dstillman:
Note to self: this will need to call `translate.resolveURL()`, as `doGet()` and friends do.

AbeJellinek:
>Are we aware of anywhere that translators would need to check either the status code or header (e.g., `Content-Type`) of a `2xx` request?

In rare cases, making a request and grabbing the `Set-Cookie` header is useful. And having `Content-Type` could be good in a case where one endpoint might return a number of different types of data. These aren't concrete cases, but not allowing translators to have access to headers at *all* seems like unnecessary loss of information. Ditto for the status code. If we're exposing those values in the error object when a request fails, we should expose them when it succeeds.

We definitely shouldn't give the caller the whole XHR object, but we could return an object with `status`, `headers`, and `body` fields and make the recommended pattern for calling it look like:

```js
let { body } = await request("https://example.com/");
```

It's just two curly braces away from what we'd have if we only returned the body, and it lets us add more properties if needed in the future without breaking backwards compatibility. (If we only return a string, `Document`, or JSON object right now but eventually decide that we want to return more data, then we'll have to add a new method.)

>Possible options:

These look good. I think that we should accept a `method` option; it's reasonable to want to make a POST request with an empty body (a few translators do), and it's unintuitive and visually unclear for `body: null` to imply `GET` and `body: ''` to imply `POST`.

I like the idea of having `requestJSON` and `requestDocument`.

AbeJellinek:
Oh, and I really like the idea of just rejecting the promise. Much cleaner than taking an array of valid status codes.

AbeJellinek:
OK, [see new commits](https://github.com/zotero/translate/pull/4/files/325b36d36d4496cb3ef78441a9a7fc909250335e..6f24f7922a03908854c4ff67a40725dcf4fa8368).

Deprecating the callback-based request methods comes under the assumption that we'll want new translators (starting a few months after 6.0 is released) to be async, meaning an overhauled web translator template and a gradual process of updating old translators to use `request`. We should (of course) keep the old methods around as we have even older methods like `loadDocument`, but the documentation should make it clear that new translators need to avoid callbacks. Deprecating methods used by 500+ translators is big, but I would argue that it's appropriate here.

retorquere:
Can I get a ping when async support lands in a beta. Having async in export translators would allow me to do away with some really ugly monkey patches.

What does this mean for drag and drop of translation results? ISTR that translators needed to be sync for this to work.

dstillman:
@retorquere: This PR (and the newer linked version) won't actually affect export — it just makes detect* and subtranslators async, and it adds async HTTP methods for web translation.

do* functions should already be awaited if they return a promise, and some import translators take advantage of that, but I don't know that anyone has looked into whether exports can be made async.

retorquere:
I see. For export there's at least the drag and drop issue, but I had hoped the move to Electron would have made it possible and I'd hoped this was an early Christmas gift.

dstillman:
Superseded by zotero/translate#4

PR #1618, opened 2019-01-03
Print highlight colors in Reports

bwiernik:
https://forums.zotero.org/discussion/72505/color-highlights-disappear-when-i-print-a-report#latest

### DIFF
--- a/chrome/skin/default/zotero/report/detail_print.css
+++ b/chrome/skin/default/zotero/report/detail_print.css
@@ -13,8 +13,6 @@ h1, h2, h3, h4, h5, h6 {
 
 ul, ol, dl {
 	page-break-inside: avoid;
-	-webkit-print-color-adjust: exact;
-	printer-colors: exact;
 	color-adjust: exact;
 }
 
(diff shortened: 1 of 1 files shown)


chrome/skin/default/zotero/report/detail_print.css
```diff
@@ -13,6 +13,9 @@ h1, h2, h3, h4, h5, h6 {
 
 ul, ol, dl {
 	page-break-inside: avoid;
+	-webkit-print-color-adjust: exact;
+	printer-colors: exact;
+	color-adjust: exact;
```
dstillman: We only need `color-adjust`, which is [supported in current browsers](https://developer.mozilla.org/en-US/docs/Web/CSS/color-adjust#Browser_compatibility). (The `-webkit` prefix is no longer required in Chrome, and as far as I know `printer-colors` was never in an actual spec.)

bwiernik: Fixed.

PR #1638, opened 2019-02-09
Fix integration for citeproc 1.1.215

fbennett:
This pull request addresses test suite failures triggered by `citeproc-js` version 1.1.215, and reported at https://github.com/Juris-M/citeproc-js/issues/98. The anomalies shown in the trace posted to the linked issue have two causes: the marking of an existing citation as a duplicate (it looks as though this affected all citation edits); and multiple iterations of `citeproc-js` on the same index and citation data. I don't fully understand the integration layer, but all integration tests pass with these changes:
* Mark citations as duplicates only if they are at a different index from the citation to be inserted.
* Skip citation indices present in `newIndices` when processing `updateIndices`, to avoid redundant runs of `citeproc-js`.

(**Edit:** changed "Remove" to "Skip" in the second bullet-point, with other edits. See further notes below.)

### DIFF
--- a/chrome/content/zotero/xpcom/integration.js
+++ b/chrome/content/zotero/xpcom/integration.js
@@ -1661,7 +1661,7 @@ Zotero.Integration.Session.prototype._updateCitations = async function () {
 	
 	var [citations, fieldToCitationIdxMapping, citationToFieldIdxMapping] = this.getCiteprocLists();
 
-	var updateIndices = {};
+	var updateIndices = Object.assign({}, this.updateIndices);
 	for (let indexList of [this.newIndices, this.updateIndices]) {
 		for (let index in indexList) {
 			// Jump to next event loop step for UI updates
--- a/chrome/content/zotero/xpcom/integration.js
+++ b/chrome/content/zotero/xpcom/integration.js
@@ -1212,8 +1212,12 @@ Zotero.Integration.Fields.prototype.addEditCitation = async function (field) {
 		for (var nextIdx = idx+1; nextIdx < fields.length; nextIdx++) {
 			if (nextIdx in fieldToCitationIdxMapping) break;
 		}
+		var nextMappedIdx = fieldToCitationIdxMapping[nextIdx];
+		if (nextMappedIdx === undefined) {
+			nextMappedIdx = fields.length;
+		}
 		let citationsPre = citations.slice(0, fieldToCitationIdxMapping[prevIdx]+1);
-		let citationsPost = citations.slice(fieldToCitationIdxMapping[nextIdx]);
+		let citationsPost = citations.slice(nextMappedIdx);
 		let citationID = citation.citationID;
 		try {
 			var result = this._session.style.previewCitationCluster(citation, citationsPre, citationsPost, "rtf");
--- a/chrome/content/zotero/xpcom/integration.js
+++ b/chrome/content/zotero/xpcom/integration.js
@@ -1600,7 +1600,7 @@ Zotero.Integration.Session.prototype.addCitation = Zotero.Promise.coroutine(func
 	}
 	
 	// We need a new ID if there's another citation with the same citation ID in this document
-	var duplicateIndex = this.documentCitationIDs[citation.citationID];
+	var duplicateIndex = this.documentCitationIDs[citation.citationID] != citation.properties.zoteroIndex ? this.documentCitationIDs[citation.citationID] : undefined;
 	var needNewID = !citation.citationID || duplicateIndex != undefined;
 	if(needNewID) {
 		if (duplicateIndex != undefined) {
@@ -1656,6 +1656,12 @@ Zotero.Integration.Session.prototype._updateCitations = async function () {
 	Zotero.debug(Object.keys(this.updateIndices));
 	
 	var [citations, fieldToCitationIdxMapping, citationToFieldIdxMapping] = this.getCiteprocLists();
+
+	for (var index in this.newIndices) {
+		if (this.updateIndices[index] !== undefined) {
+			delete this.updateIndices[index];
+		}
+	}
 	
 	for (let indexList of [this.newIndices, this.updateIndices]) {
 		for (let index in indexList) {
--- a/chrome/content/zotero/xpcom/integration.js
+++ b/chrome/content/zotero/xpcom/integration.js
@@ -1657,14 +1657,16 @@ Zotero.Integration.Session.prototype._updateCitations = async function () {
 	
 	var [citations, fieldToCitationIdxMapping, citationToFieldIdxMapping] = this.getCiteprocLists();
 
-	for (var index in this.newIndices) {
-		if (this.updateIndices[index] !== undefined) {
-			delete this.updateIndices[index];
-		}
-	}
-	
+	var outerLoopCounter = 0;
+	var newIndicesCopy = JSON.parse(JSON.stringify(this.newIndices));
 	for (let indexList of [this.newIndices, this.updateIndices]) {
+		outerLoopCounter++;
 		for (let index in indexList) {
+			if (outerLoopCounter === 2) {
+				if (newIndicesCopy[index]) {
+					continue;
+				}
+			}
 			// Jump to next event loop step for UI updates
 			await Zotero.Promise.delay();
 			index = parseInt(index);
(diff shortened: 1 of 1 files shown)


fbennett:
The first commit did not avoid redundant runs of the processor as I intended. Looking more carefully at the code, in the loop, I see that it's already trying to do something similar---that is, to remove `newIndices` index values from `updateIndices`. But both the inner-loop `delete` operation and my "fix" were foiled by the final operation in the loop, which in iteration 1 (`newIndices`) includes the current citation index in values set on `updateIndices`...and that forces a second run of the processor in iteration 2 (`updateIndices`).

My fix in the second commit may look kind of klutzy, but it does what I intended now.

fbennett:
Third time's the charm, maybe. The third commit is cleaner still. Rather than skipping with continue, it casts a local-to-function `updateIndices` object to memo indexes added from processor output. Setting those values on `this.updateIndices` after the outer loop has completed allows the original loop code to work as was intended.

fbennett:
The fourth commit addresses the first anomaly noted in the long trace shown in https://github.com/Juris-M/citeproc-js/issues/98. The original behavior (calling the same citationID in `citationsPre` and `citationsPost` when generating a citation preview) may be harmless, but runs the risk of confusing the processor internals with irrational input.

fbennett:
The fifth commit tweaks the private `updateIndices` variable slightly to avoid losing data carried by the original `this.updateIndices`.

fbennett:
In the test suite, at least, the changes seem to yield a modest improvement in the speed of citation updates. The 1.1.215 processor will consume less memory, but doesn't seem to have much effect on speed.

Here are timings with current master -> this PR -> this PR + citeproc-js 1.1.215.
```
  Zotero.Integration
    Interface
	  [...]
      #addEditCitation
      ✓ should insert citation if not in field (137 ms -> 111 ms -> 110 ms)
      ✓ should edit citation if in citation field (219 ms -> 187 ms -> 186 ms)
      ✓ should update bibliography if present (203 ms -> 185 ms -> 181 ms)
        when original citation text has been modified
        ✓ should keep modification if "Cancel" selected in editCitation triggered alert (129 ms -> 119 ms -> 109 ms)
        ✓ should display citation dialog if "OK" selected in editCitation triggered alert (140 ms -> 125 ms -> 122 ms)
        ✓ should set dontUpdate: true if "yes" selected in refresh prompt (122 ms -> 112 ms -> 116 ms)
        ✓ should reset citation text if "no" selected in refresh prompt (259 ms -> 107 ms -> 112 ms)
        when there are copy-pasted citations
        ✓ should resolve duplicate citationIDs and mark both as new citations (197 ms -> 129 ms -> 126 ms)
        ✓ should successfully process citations copied in from another doc (166 ms -> 178 ms -> 185 ms)
        when delayCitationUpdates is set
        ✓ should insert a citation with wave underlining (200 ms -> 282 ms -> 278 ms)
        ✓ should not write to any other fields besides the one being updated (172 ms -> 171 ms -> 162 ms)
        [...]
```

fbennett:
If it is possible, running `previewFn` on demand rather than for every add/edit operation would eliminate another run of the processor, and speed things up a little more.

adomasven:
If the document has citations [a, b, c] and two new citations are added [a, b, c, **d**, **e**], when we call `processCitationCluster()` with d and specify [e] in citationsPost, will the processor return text for e, even if items in d have no relation to e? If so, why?

If we have 30 citations in a document and upon updating one of them the processor returns text for 2 other citations, do we need to run `processCitationCluster()` with those citations as the first argument or not? What about if we add a new citation and the processor returns text for 2 other citations?

fbennett:
> If the document has citations [a, b, c] and two new citations are added [a, b, c, **d**, **e**], when we call `processCitationCluster()` with d and specify [e] in citationsPost, will the processor return text for e

This will crash the processor, because it has no internal record of **e**.

> If we have 30 citations in a document and upon updating one of them the processor returns text for 2 other citations, do we need to run `processCitationCluster()` with those citations as the first argument or not?

The returned strings can be inserted into the document directly. There is no need to run `processCitationCluster()` over the additional citations.

> What about if we add a new citation and the processor returns text for 2 other citations?

Again, the strings returned for the additional citations can be used directly. When the new citation is added, its side effects are all reflected in the returned strings.

----

Commute time here, I need to run - but meanwhile, I should mention that I've come up with a test fixture that will break under this PR + citeproc-js 1.1.215. I need to study some more.

adomasven:
> > If the document has citations [a, b, c] and two new citations are added [a, b, c, **d**, **e**], when we call `processCitationCluster()` with d and specify [e] in citationsPost, will the processor return text for e
> 
> This will crash the processor, because it has no internal record of **e**.

How about if we have [a, b, c] and we make inserts to produce [a, **d**, b, **e**, c]. Should `processCitationCluster()` call for **d** have [b, c] in postCitations?

> Commute time here, I need to run - but meanwhile, I should mention that I've come up with a test fixture that will break under this PR + citeproc-js 1.1.215. I need to study some more.

Thanks for this work you've done over the weekend. I have some ideas about not quite proper use of citeproc in the integration layer based on wrong assumptions about how the citeproc API works (although I haven't touched this myself in a while, so my memory is hazy too) and will be updating this code some more myself. It would be good to check if your text fixtures pass after that.

fbennett:
I thought about this on the train, and I think steps like this should work. I haven't looked at the existing code in detail, so this is just my own musing.
* Start with two values: `fieldIDs` (an array); `citeproc.registry.citationreg.citationById` (an object). The aim is to update the processor so that the sequence of `fieldIDs` is reflected in `citeproc.registry.citationreg.citationByIndex`---but the latter will be updated implicitly based on the IDs provided in `citationsPre` and `citationsPost`.
* Prepare two objects to hold field index values: `newIndices` and `updateInfo`.
* Prepare on tracking object: `seenIDs`.
* If inserting a citation, memo its field index as `insertIndex`. If editing a citation, memo its field index as `editIndex`.
* Step through `fieldIDs`, checking each for a corresponding value in `citeproc...citationById`. If a value does not exist, or if the index matches `insertIndex`, or if the ID exists in `seenIDs`, add the document index of the field to `newIndices`. Memo each ID to `seenIDs`.
* To reconcile, first process `newIndices` in ascending order. Here things get a little tricky. For each processor run:
  - `citationsPre` is slice of `fieldIDs` up to the target index.
  - `citationsPost` is a slice of `fieldIDs` after the target index, but with all IDs held in `newIndices` removed.
  - Index/string/citationID triples returned from `processCitationCluster()` are written to `updateInfo`, but with one proviso.
  - Before the second and all subsequent runs, an adjustment must be made to field index values in `updateInfo` to keep citations aligned. Each index that is equal to or greater than the field index value of the *the current citation to be processed from* `newIndices` must be incremented by one. With that adjustment, overwriting existing entries in `updateInfo` will be safe, and will capture any changes due to disambiguation and the like that arise as citations are added.
* If the operation is an edit to an existing citation, that can be performed after all `newIndices` fields have been processed. Any index/string/citationID triples returned can be written to `updateInfo` directly, as the index values will align.

Finally, if the user adds or removes one or more fields in the document before all of that is complete, cancelling the operation and restarting the steps from the beginning should be safe.

That's my thinking-out-loud pseudo-code for it anyway.

adomasven:
> That's my thinking-out-loud pseudo-code for it anyway.

This is describing something like how we use citeproc, although I am not sure what this response is to. What is the answer to:

> How about if we have [a, b, c] and we make inserts to produce [a, **d**, b, **e**, c]. Should `processCitationCluster()` call for **d** have [b, c] in postCitations?

fbennett:
I was just thinking things through to be sure I understood. Documents are slippery and hard.
> How about if we have [a, b, c] and we make inserts to produce [a, **d**, b, **e**, c]. Should `processCitationCluster()` call for **d** have [b, c] in postCitations?

That's right. Then you would have to follow with a call to insert **e**, with `citationsPre` set to [a, d, b], and `citationsPost` set to [c]. It isn't possible to insert multiple citations in one go in the `citeproc-js` API.

adomasven:
Ok, so the integration code does two passes through citeproc: one for newly inserted (copied in from somewhere) citations and one for citations with Zotero updated metadata. On the newly inserted pass we provide `[]` for [citationsPost](https://github.com/zotero/zotero/blob/a931c2ba51b197f5658a9a5904798b8410475e53/chrome/content/zotero/xpcom/integration.js#L1672). When citeproc used to maintain citation IDs no longer in the registry this worked fine (and used to make me wonder why citationsPre/citationsPost stuff is even required). Now that you've updated citeproc we'll have to change it. We're also doing some redundant invocations on the second pass that I'll fix.

Thanks for the investigation, explanation and finding some other minor bugs. I'll push a few more commits to master, close this PR and see whether we still have problems.

fbennett:
Excellent - many thanks for your patience!

adomasven:
Closing this for now. Thanks again for putting me on the right track!

PR #1640, opened 2019-02-12
Test of implicit citation update

fbennett:
This adds a test to exercise implicit citation updates triggered by disambiguation. With https://github.com/zotero/zotero/commit/9410094ad87193a462e705e955856779470add40 the test currently fails (because only the index of the current citation being added/edited is set on `processIndices`).

### DIFF
--- a/test/tests/integrationTest.js
+++ b/test/tests/integrationTest.js
@@ -486,8 +486,6 @@ describe("Zotero.Integration", function () {
 				setAddEditItems(testItems[4]);
 				yield execCommand('addEditCitation', docID);
 				assert.equal(doc.fields.length, 3);
-				assert.equal(Zotero.Integration.currentSession.style.registry.registry[testItems[4].id].disambig.year_suffix, '0');
-				assert.equal(Zotero.Integration.currentSession.style.registry.registry[testItems[3].id].disambig.year_suffix, '1');
 				assert.equal(doc.fields[0].text, "(Smith, 2019a)");
 				assert.equal(doc.fields[2].text, "(Smith, 2019b)");
 			});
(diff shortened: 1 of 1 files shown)


fbennett:
Oops. In the first commit of the PR I'm setting the date to 'today', then assuming in the test that the year is 2019.

fbennett:
Fixed.

fbennett:
The third commit adds another test of implicit updates. The test sets up three citations, then:
* sets the first two as duplicates, then
* updates the first to require disambiguation against the third.

That forces integration to run the processor twice:
* `processCitationCluster(a, [], [c])`; then
* `processCitationCluster(b, [a], [c])`

The first run returns strings for field positions `0` and `1` (because the processor has not yet seen "b"), and the second returns a string for field position `1`. For the insertion to work correctly, the index on the return string at `1` must be incremented by one before the second run of the processor.

I hope that makes sense---the basic problem is that the processor cannot return the correct target index because it has not yet seen the full field list, so things need to be adjusted as each `newIndices` citation is processed in turn.

test/tests/integrationTest.js
```diff
@@ -492,6 +492,36 @@ describe("Zotero.Integration", function () {
...
+				doc.fields[1].text = doc.fields[0].text;
+
+				testItems[4].setCreator(0, {creatorType: 'author', lastName: 'Smith', firstName: 'Robert'});
+				testItems[4].setField('date', '2019-01-01');
+
+				setAddEditItems(testItems[4]);
+				yield execCommand('addEditCitation', docID);
+				assert.equal(doc.fields.length, 3);
+				assert.equal(Zotero.Integration.currentSession.style.registry.registry[testItems[4].id].disambig.year_suffix, '0');
+				assert.equal(Zotero.Integration.currentSession.style.registry.registry[testItems[3].id].disambig.year_suffix, '1');
```
adomasven: We do not want Zotero tests to be asserting things within citeproc state (or the correctness of state of any library that we use), because none of us knows how the internals of it work and if the test breaks we won't be able to fix it ourselves or figure out what's gone wrong. Aren't the lines below essentially checking the same thing?

fbennett:
Yes. Those checks of processor state were just for my own reference - they aren't essential to the test, I'll remove them.

adomasven:
Merged via 4a61cffc99e22de3d32b125e9df58e012da8bc7f

PR #1641, opened 2019-02-13
Set and adjust processIndices

fbennett:
With this PR, the tests at https://github.com/zotero/zotero/pull/1640 will pass.


fbennett:
I had this wrong. The adjustment should happen only when processing `newIndices`. The second commit fixes it.

fbennett:
No, there are serious problems with this. Closing the PR, will open another if I get things sorted out.

PR #1668, opened 2019-03-21
Handle handingindent return from citeproc-js as boolean

fbennett:
This PR addresses an issue raised by Derek Sifford a couple of years ago that I've been slow to move on, concerning a spec violation by the processor:

* **citeproc-js issue:** https://github.com/Juris-M/citeproc-js/issues/15
* **CSL spec language:** http://docs.citationstyles.org/en/1.0.1/specification.html#whitespace


fbennett:
This change will go in at month's end, with a toggle documented in CHANGES that you can use to get the old return value of "2".

dstillman:
Thanks!

PR #1672, opened 2019-03-27
Display a more user-friendly error for integration errors

### LINKED ISSUE #1666: Better error dialog for integration errors
We should remove the raw citeproc-js error and include a link to the debugging steps.

(This presumably has to happen in the three plugin repos separately, but this can be a tracking issue.)

adomasven:
image

We cannot use Zotero to display prompts because those tend to not focus the Zotero window properly. We cannot display prompts with options other than Yes/No/Ok/Cancel in Word and LibreOffice, so no "More Info" buttons either. To test add 

```
throw new Error("test");
```
at https://github.com/adomasven/zotero/blob/0cb056c994ccd4fa493a60d473d36638781afe0b/chrome/content/zotero/xpcom/integration.js#L248

Closes #1666


chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -873,6 +873,7 @@ integration.error.styleNotFound = The citation style %S could not be found.
 integration.error.macWordSBPermissionsMissing.title = Missing Permission
 integration.error.macWordSBPermissionsMissing = Zotero does not have permission to control Word. To grant this permission:\n\n1) Open System Preferences\n2) Click on “Security & Privacy”\n3) Select the “Privacy” tab\n4) Find and select “Automation” on the left\n5) Check the checkbox for “Microsoft Word” under “Zotero”\n6) Restart Word
 integration.error.macWordSBPermissionsMissing.pre2016 = If “Microsoft Word” does not appear under “Automation”, make sure you are running Word 2011 version 14.7.7 or later.
+integration.error.seeTroubleshootingInfo	= Would you like to see the troubleshooting instructions?
```
dstillman: "Would you like to view troubleshooting instructions?"

Issue #1338, opened 2017-10-10
Don't recommend "Attach Link to File" for file: URIs in group libraries


dstillman:
https://forums.zotero.org/discussion/68075/how-attach-links-to-files-using-link-to-uri-in-groups-zotero5

PR #1673, opened 2019-03-29
Add per library linked attachment base directory support

sbennett18:
Basically the two preferences that related to base directory attachment paths were changed to stringified objects that use the `libraryID` as the key and store a unique value per library.  The same `BASE_ATTACHMENT_PATH` prefix is used.
The logic for changing or clearing base directories now only modifies the `attachment.attachmentPath` for an `attachment` who's `attachment.libraryID` matches the specified `libraryID`.

The attachment base directory selection window UI is based off the *Libraries to Sync UI*.

I tested the following additional use cases (for which unit tests need to be created):
- Set attachment base paths for two libraries to two different paths, then checked that all attachments still open.
- Set attachment base paths for two libraries to the same path, then checked that all attachments still open.
- Set attachment base paths for two libraries, then removed one of them and checked that all attachments still open and that the other library still uses the linked files.
- Others I can't think of right now...

| Old | New |
| --- | --- |
| `Zotero_Preferences.Attachment_Base_Directory.getPath` | `Zotero.Attachments.getBaseDirByLibrary` |
| `Zotero_Preferences.Attachment_Base_Directory.choosePath` | `Zotero_Preferences.Attachment_Base_Directory.getNewBaseDir` `Zotero_Preferences.Attachment_Base_Directory.manageBaseDirs` |
| `Zotero_Preferences.Attachment_Base_Directory.changePath` | `Zotero_Preferences.Attachment_Base_Directory.changeBaseDirByLibrary` |
| `Zotero_Preferences.Attachment_Base_Directory.clearPath` | `Zotero_Preferences.Attachment_Base_Directory.clearBaseDirByLibrary` |

### To Do
- [ ] Reactify (Blocked by #1674)
- [x] Only set attachment base paths for user and group libraries
- [x] Get all tests to pass
- [ ] Add new tests for group specific base paths - should only change files in specified library
- [ ] _Assistance Needed:_ Add localization support for new UI window

### Unanticipated Changes
- `item.js`: Added a default `libraryID` of `Zotero.Libraries.userLibraryID` if the `linkMode` is `LINK_MODE_LINKED_FILE`.  This was an assumption being made by *many* tests and has the same effect as if the attachment had first been `.save()`'d and had its `libraryID` defaulted there.  The tests could all be updated to pass an explicit `libraryID`, but the other issue is generally that the many test parent items also don't have a `libraryID` set.
- Added additional optional parameter `libraryID` to several functions in `attachments.js`.

### Related Issues / Commits / PRs
- #1674 - Reactify preferences
- #1675 - Show X for unsynced libraries
- #1338 - Don't recommend "Attach Link to File" for file: URIs in group libraries
- zotero#51 - Relative paths to linked files
- https://forums.zotero.org/discussion/comment/317328#Comment_317328
- https://forums.zotero.org/discussion/36857/group-sharing-through-webdav-and-linked-base-directories
- https://forums.zotero.org/discussion/55253/attach-link-to-file-grayed-out

### Screenshots
image
image

### DIFF
--- a/chrome/content/zotero/preferences/preferences_advanced.js
+++ b/chrome/content/zotero/preferences/preferences_advanced.js
@@ -535,27 +535,117 @@ Zotero_Preferences.Advanced = {
 
 
 Zotero_Preferences.Attachment_Base_Directory = {
-	getPath: function () {
-		var oldPath = Zotero.Prefs.get('baseAttachmentPath');
-		if (oldPath) {
-			try {
-				return OS.Path.normalize(oldPath);
+
+
+	manageBaseDirs: function () {
+		var io = {};
+		window.openDialog('chrome://zotero/content/preferences/libraryAttachmentBaseDirs.xul',
+			"zotero-preferences-libraryAttachmentBaseDirsDialog", "chrome,modal,centerscreen", io);
+	},
+
+
+	dblClickLibraryAttachmentBaseDir: async function (event) {
+		var tree = document.getElementById("library-attachment-base-dirs-tree");
+		var row = {}, col = {}, child = {};
+		tree.treeBoxObject.getCellAt(event.clientX, event.clientY, row, col, child);
+
+		// Below the list
+		if (!col.value) {
+			return;
+		}
+		var index = row.value;
+
+		var treechildren = document.getElementById("library-attachment-base-dirs-rows");
+		if (index >= treechildren.childNodes.length) {
+			return;
+		}
+
+		var treeRow = treechildren.childNodes[index];
+
+		var libraryID = treeRow.firstChild.childNodes[1].getAttribute("value");
+		if (!libraryID) {
+			return;
+		}
+		libraryID = parseInt(libraryID);
+
+		var checkboxCell = treeRow.firstChild.childNodes[0];
+		var pathCell = treeRow.firstChild.childNodes[2];
+		var oldPath = pathCell.getAttribute("value");
+
+		var newPath = await this.getNewBaseDir(oldPath);
+		if (!newPath) {
+			return;
+		}
+
+		if (await this.changeBaseDirByLibrary(libraryID, newPath)) {
+			checkboxCell.setAttribute("value", true);
+			pathCell.setAttribute("label", newPath);
+			pathCell.setAttribute("value", newPath);
+		}
+	},
+
+
+	clickLibraryAttachmentBaseDir: function (event) {
+		var tree = document.getElementById("library-attachment-base-dirs-tree");
+		var row = {}, col = {}, child = {};
+		tree.treeBoxObject.getCellAt(event.clientX, event.clientY, row, col, child);
+
+		// Below the list or not on checkmark column
+		if (!col.value || col.value.element.id != "library-attachment-base-dirs-checked") {
+			return;
+		}
+
+		this.toggleLibraryAttachmentBaseDir(row.value);
+	},
+
+
+	toggleLibraryAttachmentBaseDir: async function (index) {
+		var treechildren = document.getElementById("library-attachment-base-dirs-rows");
+		if (index >= treechildren.childNodes.length) {
+			return;
+		}
+
+		var row = treechildren.childNodes[index];
+		var libraryID = row.firstChild.childNodes[1].getAttribute("value");
+		if (!libraryID) {
+			return;
+		}
+		libraryID = parseInt(libraryID);
+		var checkboxCell = row.firstChild.childNodes[0];
+		var checked = checkboxCell.getAttribute("value") === "true";
+		var pathCell = row.firstChild.childNodes[2];
+		var oldPath = pathCell.getAttribute("value");
+
+		if (checked) {
+			var newPath = await this.getNewBaseDir(oldPath);
+			if (!newPath) {
+				checkboxCell.setAttribute("value", false);
+				return;
 			}
-			catch (e) {
-				Zotero.logError(e);
-				return false;
+			var changed = await this.changeBaseDirByLibrary(libraryID, newPath)
+			checkboxCell.setAttribute("value", changed);
+			if (changed) {
+				pathCell.setAttribute("label", newPath);
+				pathCell.setAttribute("value", newPath);
+			}
+		} else {
+			var cleared = await this.clearBaseDirByLibrary(libraryID)
+			checkboxCell.setAttribute("value", !cleared);
+			if (cleared) {
+				pathCell.setAttribute("label", "");
+				pathCell.setAttribute("value", "");
 			}
 		}
 	},
-	
-	
-	choosePath: async function () {
-		var oldPath = this.getPath();
-		
-		//Prompt user to choose new base path
+
+	getNewBaseDir: async function (oldBaseDir) {
+		// Prompt user to choose new base directory
+		if (oldBaseDir) {
+			var oldPathFile = Zotero.File.pathToFile(oldBaseDir);
+		}
 		var fp = new FilePicker();
-		if (oldPath) {
-			fp.displayDirectory = oldPath;
+		if (oldPathFile) {
+			fp.displayDirectory = oldPathFile;
 		}
 		fp.init(window, Zotero.getString('attachmentBasePath.selectDir'), fp.modeGetFolder);
 		fp.appendFilters(fp.filterAll);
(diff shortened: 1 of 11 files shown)


chrome/content/zotero/preferences/libraryBaseAttachmentPaths.xul
```diff
@@ -0,0 +1,67 @@
...
+	<groupbox>
+		<caption label="Library Base Attachment Paths"/>
+
+		<tree id="library-base-attachment-paths-tree" flex="1" width="415" hidecolumnpicker="true" rows="10" seltype="single" editable="true"
+			ondblclick="Zotero_Preferences.Attachment_Base_Directory.dblClickLibraryBaseAttachmentPath(event);"
+            onclick="Zotero_Preferences.Attachment_Base_Directory.clickLibraryBaseAttachmentPath(event);">
+            <treecols>
+                <treecol editable="true" type="checkbox" id="library-base-attachment-paths-checked" label="Relative"/>
+                <treecol editable="false" id="library-base-attachment-paths-name" label="&zotero.preferences.sync.librariesToSync.library;"/>
+                <treecol editable="false" id="library-base-attachment-paths-path" flex="1" label="Path"/>
```
sbennett18: Localisation needed for `label="Path"`

chrome/content/zotero/preferences/libraryBaseAttachmentPaths.xul
```diff
@@ -0,0 +1,67 @@
...
+	<script src="preferences_advanced.js"/>
+	
+	<groupbox>
+		<caption label="Library Base Attachment Paths"/>
+
+		<tree id="library-base-attachment-paths-tree" flex="1" width="415" hidecolumnpicker="true" rows="10" seltype="single" editable="true"
+			ondblclick="Zotero_Preferences.Attachment_Base_Directory.dblClickLibraryBaseAttachmentPath(event);"
+            onclick="Zotero_Preferences.Attachment_Base_Directory.clickLibraryBaseAttachmentPath(event);">
+            <treecols>
+                <treecol editable="true" type="checkbox" id="library-base-attachment-paths-checked" label="Relative"/>
```
sbennett18: Localisation needed for `label="Relative"`

chrome/content/zotero/xpcom/attachments.js
```diff
@@ -2024,60 +2024,153 @@ Zotero.Attachments = new function(){
...
+			basePath, path.substr(Zotero.Attachments.BASE_PATH_PLACEHOLDER.length)
 		));
-	}
+	};
+
+
+	/**
+	 * Get the base attachment path for ``libraryID``, if its set
+	 *
+	 * @param {Number} libraryID
```
sbennett18: `Number` or `Integer` here?  Be consistent at least...

chrome/content/zotero/xpcom/attachments.js
```diff
@@ -2024,60 +2024,153 @@ Zotero.Attachments = new function(){
...
+		Zotero.Prefs.set("libraryBaseAttachmentPaths", JSON.stringify(prefValue));
+		return true;
+	};
+
+
+	/**
+	 * Get whether attachments within ``libraryID``'s base attachment directory
+	 * should be saved as relative paths.
+	 *
+	 * @param {Integer} libraryID
```
sbennett18: `Number` or `Integer` here?  Be consistent at least...

sbennett18 (commented):
Add tests for multiple libraries, should only change files in specified library

chrome/content/zotero/xpcom/attachments.js
```diff
@@ -604,6 +628,10 @@ Zotero.Attachments = new function(){
 		var contentType = options.contentType;
 		var title = options.title;
 		var collections = options.collections;
+
+		if (parentItemID && collections) {
```
sbennett18: This check exists in all other similar functions.

chrome/content/zotero/xpcom/attachments.js
```diff
@@ -2449,13 +2593,14 @@ Zotero.Attachments = new function(){
...
 			attachmentItem.attachmentCharset = charset;
-			if (file) {
-				attachmentItem.attachmentPath = typeof file == 'string' ? file : file.path;
-			}
 			
 			if (collections) {
 				attachmentItem.setCollections(collections);
 			}
+
+			if (file) {
```
sbennett18: Since `setCollections()` sets a default `libraryID` if one isn't already set, this re-ordering is just to give it one more chance to be set before possibly using it in `.attachmentPath`.

chrome/content/zotero/xpcom/data/item.js
```diff
@@ -2889,20 +2891,24 @@ Zotero.defineProperty(Zotero.Item.prototype, 'attachmentPath', {
...
-				let libraryType = Zotero.Libraries.get(this._libraryID).libraryType;
+			let libraryID = this._libraryID;
+			if (libraryID) {
+				let libraryType = Zotero.Libraries.get(libraryID).libraryType;
 				if (libraryType != 'user') {
 					throw new Error("Linked files can only be added to user library");
 				}
+			} else {
+				Zotero.debug(`'this._libraryID' has not been set, defaulting to 'userLibraryID'`);
+				libraryID = Zotero.Libraries.userLibraryID;
```
sbennett18: This is one of the biggest changes.  This was done to appease a lot of the unit tests that don't otherwise specify any `libraryID` or provided a fully-fleshed out `parentItem` that also doesn't have a `libraryID` setup.

chrome/content/zotero/preferences/preferences_sync.js
```diff
@@ -213,11 +213,22 @@ Zotero_Preferences.Sync = {
 
 		this.displayFields();
 		Zotero.Prefs.clear('sync.librariesToSync');
+		var userLibraryID = Zotero.Libraries.userLibraryID;
```
sbennett18: I don't believe this is actually needed anymore with the pruning done in `initLibraryAttachmentBasePaths`.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -121,6 +121,9 @@ errorReport.repoCannotBeContacted	= Repository cannot be contacted
 
 
 attachmentBasePath.selectDir                                   = Choose Base Directory
+attachmentBasePath.libraryAttachmentBasePaths.label            = Library Attachment Base Directories
+attachmentBasePath.libraryAttachmentBasePaths.relative         = Relative
+attachmentBasePath.libraryAttachmentBasePaths.path             = Path
```
sbennett18: I must be missing something with the localization since I got an error when I tried to reference these in the xul.

chrome/content/zotero/preferences/preferences_advanced.js
```diff
@@ -579,30 +673,80 @@ Zotero_Preferences.Attachment_Base_Directory = {
...
-		];
-		var oldRelativeAttachmentIDs = yield Zotero.DB.columnQueryAsync(sql, params);
-		
-		//Find all attachments on the new base path
+		// Find all attachments on the new base path
 		var sql = "SELECT itemID FROM itemAttachments WHERE linkMode=?";
 		var params = [Zotero.Attachments.LINK_MODE_LINKED_FILE];
-		var allAttachments = yield Zotero.DB.columnQueryAsync(sql, params);
+		var allAttachmentIDs = yield Zotero.DB.columnQueryAsync(sql, params);
+		var isLibraryBasePathSet = Zotero.Attachments.getBasePathByLibrary(libraryID);
```
sbennett18: Why was `getPath` (see `Zotero.Attachments.getBasePathByLibrary`) called every loop when it shouldn't change from one iteration to the next?
It's fair to check that a base attachment path is set for the library, but it doesn't need to be checked every time...

chrome/content/zotero/preferences/preferences_sync.js
sbennett18: Testing showed this cleanup is not necessary.

chrome/content/zotero/preferences/preferences_advanced.js
```diff
@@ -719,11 +721,12 @@ Zotero_Preferences.Attachment_Base_Directory = {
...
-				).map(id => parseInt(id));
+			Object.keys(
+				JSON.parse(
+					Zotero.Prefs.get("librarySaveRelativeAttachmentPaths") || "{}"
+				)
+			).map(id => parseInt(id));
+
 		savedLibraryIDs.forEach(function (libraryID) {
 			if (existentLibraryIDs.indexOf(libraryID) == -1) {
 				Zotero.debug(`Pruning attachment base path preferences for non-existent library '${libraryID}'`);
```
sbennett18: Hmm...  Should `clearPath` be called on every `libraryID` that's being pruned?  Or is it safe to assume that all attachments have been removed already, so there wouldn't be any attachment paths that need to be updated?
Is it possible for any attachments to have a `libraryID` of a library that `Zotero.Libraries.getAll()` doesn't return?

sbennett18:
Let me know if you want me to squash any of these commits.

dstillman:
Thanks for working on this!

Unfortunately this will need to be reworked a bit, since we're in the process of moving Zotero from XUL to HTML/React as part of the eventual port to Electron, and we can't accept any new XUL UI.

We'll be redoing the preferences in React in the near future, which will likely involve some significant UI changes, so I'd recommend holding off until that happens, after which you'll be able to base the UI for this on the work we do for that. I've created https://github.com/zotero/zotero/issues/1674 so that you can subscribe and get notified when the preferences have been redone.

chrome/content/zotero/xpcom/attachments.js
```diff
@@ -604,6 +640,10 @@ Zotero.Attachments = new function(){
 		var contentType = options.contentType;
 		var title = options.title;
 		var collections = options.collections;
+
+		if (parentItemID && collections) {
```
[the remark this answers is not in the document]
sbennett18: Consider pulling these checks out to a separate function so this kind of thing isn't missed again.
Also, why don't all of these functions use `_addToDB`?  These checks could be implemented there instead.

fredericky123:
when the work will restart?

sbennett18:
I'll let @dstillman decide if this should remain in a holding pattern in light of his recent comment on #1674.

fredericky123:
should this be remained? @dstillman

fredericky123:
can the group library be synced through webdav?

sbennett18:
It looks like some changes were made that now use async functions.  Is the idea to move away from Bluebird coroutines and just use built in language features going forward? I'll have to re-remember what I did and learn a bit of the new (to me) syntax for async function usage.

dstillman:
Nothing to learn — it's a direct conversion. `coroutine(function*` becomes `async function`, and `yield` becomes `await`. (Until somewhat recently, we were converting `async`/`await` to Bluebird `coroutine` and `yield` with Babel, though now we just use the native promises.)

PR #1678, opened 2019-04-06
Fix data type

fbennett:
I don't know whether this is related to the ``c[0]`` bug in some way, but it seems to need a tweak.

Values of `fieldToCitationIdxMapping` returned by `getCiteprocLists` are assumed to be numbers in some places. The `idx` value used in the function is an object key, this converts it to an integer so that adding to it produces a sane result.


fbennett:
Closing this PR, and will file a separate one with a few more changes related to the c[0] bug.

PR #1688, opened 2019-04-26
Make CSL extraction from Extra more flexible

bwiernik:
1) Allow 'type' as a CSL field
2) Transform common field variants to CSL variables
3) Transform common item type variants to CSL types

Follow up on https://github.com/zotero/zotero/commit/1f320e1f5d5fd818e2c2d532f4789e38792a77a2#diff-9b1aaab8c13feb794bfd72e10de8b3d8

### DIFF
--- a/chrome/content/zotero/xpcom/cite.js
+++ b/chrome/content/zotero/xpcom/cite.js
@@ -387,6 +387,7 @@ Zotero.Cite = {
 			case 'title':
 			case 'title-short':
 			case 'translator':
+			case 'type':
 			case 'version':
 			case 'volume':
 			case 'year-suffix':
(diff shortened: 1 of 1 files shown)


dstillman:
Could you say more about this? What makes these "common"? Are these defined anywhere, or are these just things you think people try to use?

We're also close to being able to add new fields, so hopefully the vast majority of Extra field hacks will go away in the near future.

dstillman:
Part of the reason this is important is because we're going to migrate the fields we can from Extra to regular fields, and the code I have so far uses the [CSL mappings](https://github.com/zotero/zotero/blob/bd7cdbdc81a8dc26e6ebc638b81d263e674e68f2/chrome/content/zotero/xpcom/utilities.js#L29-L142) for that. If we add other fields here, they won't be migrated properly.

bwiernik:
The first change regarding `type` is just to make this CSL variable capable of being stored in uppercase like all the rest. 

The others are based primarily on the English UI labels for fields in Zotero. Happy to remove these and make the change just the minimal `type` compatibility stuff.

bwiernik:
@dstillman For now, would you prefer to just have a PR to accept `Type` in addition to `type`, to be consistent with all other CSL variables?

Issue #1703, opened 2019-06-11
Detect existing retracted items in documents


dstillman:
The retraction functionality so far will detect retracted items in your library and warn you before citing one of those, but it doesn't help if you already have a retracted item in your document — and, since it encourages you to delete retracted items, risks orphaning those items in your document.

When refreshing a document, for each item that's retracted (see Checking Items below), scroll to the citation and show an alert:

```
Warning

A citation in your document has been retracted:

{{ bib entry }}

{{ if item is in library, add retraction.citeWarning.text2,
  but no view button because we're in a loop }}

[ ] Don’t warn me about this citation again

[ OK ]
```

If they check the box, add a flag to the citation metadata indicating that the item shouldn't be checked or alerted on again. If they don't, we'll just show the same alert on the next refresh.

## Checking Items

If the item is in the library (including the trash), use `Zotero.Retractions.isRetracted(item)`.

If not in the library, use `Zotero.Retractions.isRetractedData(data)`, which will optimize checking and caching fields to avoid repeated lookups. [Not yet implemented]

dstillman:
If the same item appears more than once, only show the prompt for the first one in the loop. If the checkbox is checked and there are multiple of the same citation, mark all subsequent citations as permanently ignored.

dstillman:
Update to the above: instead of running `async Zotero.Retractions.isRetractedData(data)` on every item, pass an array of objects to `async Zotero.Retractions.getRetractionsFromJSON(data)`, which will return an array of indexes of retracted items in the input array. (This lookup can result in an API request, so it doesn't make sense to do them individually.)

adomasven:
I recommend to instead to make `Zotero.Retractions.isRetractedData()` synchronous and immediately return a boolean, and also potentially issue async actions in the background. We don't want citing to be slowed down by network requests, especially since not everyone using Zotero is on a good network.

PR #1704, opened 2019-06-12
Add a checks and prompts for retracted items within an existing doc

adomasven:
See #1703

Currently `Retractions.getRetractionsFromJSON()` fails with a HTTP 403 for me.

### DIFF
--- a/chrome/content/zotero/xpcom/integration.js
+++ b/chrome/content/zotero/xpcom/integration.js
@@ -967,6 +967,13 @@ Zotero.Integration.Fields.prototype.updateSession = Zotero.Promise.coroutine(fun
 		this._session.regenAll = true;
 	}
 	yield this._processFields();
+	try {
+		yield this._session.handleRetractedItems();
+	}
+	catch (e) {
+		Zotero.debug('Retracted item handling failed');
+		Zotero.logError(e);
+	}
 	this._session.regenAll = false;
 
 	var updateTime = timer.stop();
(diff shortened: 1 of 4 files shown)


chrome/content/zotero/xpcom/integration.js
```diff
@@ -1928,6 +1929,68 @@ Zotero.Integration.Session.prototype.getItems = function() {
 	return Zotero.Cite.getItem(Object.keys(this.citationsByItemID));
 }
 
+Zotero.Integration.Session.prototype.handleRetractedItems = async function () {
+	const dealWithRetracted = (citedItem) => {
+		let dontPromptAgain = this.promptForRetraction(citedItem);
```
dstillman: Don't we need `inLibrary` to be passed to this?

chrome/content/zotero/xpcom/integration.js
```diff
@@ -1928,6 +1929,68 @@ Zotero.Integration.Session.prototype.getItems = function() {
...
+	for (let index of retractedIndices) {
+		dealWithRetracted(embeddedCitedItems[index]);
+	}
+};
+
+Zotero.Integration.Session.prototype.promptForRetraction = function (citedItem, inLibrary) {
+	let ps = Services.prompt;
+	let buttonFlags = (ps.BUTTON_POS_0) * (ps.BUTTON_TITLE_OK);
+	let creator = citedItem.getCreator(0);
+	let itemString = (creator ? creator.lastName + ", " : "")
```
dstillman: Can use `item.firstCreator`

dstillman:
I moved a retracted item to the trash and clicked Refresh and got this:

```
Error: citedItem.getField is not a function
Source File: chrome://zotero/content/xpcom/integration.js
Line: 1974
```

dstillman:
We should probably put all of `handleRetractedItems` in a try/catch (instead of just the `getRetractionsFromJSON()` call, since there's no reason for document updates to fail if something goes wrong.

PR #1712, opened 2019-07-14
Better wording for attach file and attach link

csicar:
The wording of the "attach ..." context menu is a bit confusing:
"Attach Link to File…" attaches a file to a link.

The PR simplifies the wording.


dstillman:
Thanks, but the wording here is intentional.

"Attach Link to File" is necessary to differentiate the option from "Attached Stored Copy of File" — one is for linked files, and one is for stored files, and the difference is quite important.

"Attach Link to URI", similarly, is intended to differentiate a web link from an HTML snapshot (which is a URI with a copy of the webpage). This one is slightly more debatable because there's no longer an option in Zotero itself to attach a snapshot (as there was when Zotero was a Firefox extension), but I think it's still clearer and wouldn't particularly benefit from being changed.

PR #1726, opened 2019-08-27
Update zotero.properties

junquera:
"FirstName" is "nombre" in Spanish, and "LastName" is "apellido"


junquera:
Hi! Sorry, I didn't know the correct way to report the mistake, I will follow the procedure next time!

Yes, in Spanish it's never told "first" or "second" to refer "name" or "last name". I've realized exporting the bibliography that some authors were wrong, and the reason is that in this context "first" and "second" can be interpreted as "first author" and "second author".

PR #1735, opened 2019-09-17
Update zotero.dtd

pmartinolli:
I changed the translation "éditeur" to "maison d'édition" which is the exact translation of publisher in French. In French, there is a confusion between editor/publisher because "éditeur" can be both. With this translation, the two roles are more clear.


dstillman:
Thanks. These changes have to be made on [Transifex](https://www.transifex.com/zotero/zotero), so I've applied this (and the equivalent for the item pane in zotero.properties) there. I believe our usual French contributor is no longer around, so you're welcome to join the French team there if you'd like to make other improvements.

PR #1737, opened 2019-10-12
Fix UnhandledPromiseRejection in translators

larsgw:
If `maybePromise` resolves, `this.decrementAsyncProcesses` is run which returns `undefined`, so no different behaviour. If `maybePromise` rejects, it now bubbles up through the call stack instead of becoming unhandled.

### DIFF
--- a/chrome/content/zotero/xpcom/translation/translate.js
+++ b/chrome/content/zotero/xpcom/translation/translate.js
@@ -1359,9 +1359,8 @@ Zotero.Translate.Base.prototype = {
 				// doImport can return a promise to allow for incremental saves (via promise-returning
 				// item.complete() calls)
 				if (maybePromise) {
-					maybePromise
-						.then(() => this.decrementAsyncProcesses("Zotero.Translate#translate()"))
-					return;
+					return maybePromise
+						.then(() => this.decrementAsyncProcesses("Zotero.Translate#translate()"));
 				}
 			} catch (e) {
 				this.complete(false, e);
--- a/chrome/content/zotero/xpcom/translation/translate.js
+++ b/chrome/content/zotero/xpcom/translation/translate.js
@@ -1344,7 +1344,7 @@ Zotero.Translate.Base.prototype = {
 			return loadPromise.then(() => rest.apply(this, arguments))
 		}
 		
-		function rest() {
+		async function rest() {
 			Zotero.debug("Translate: Beginning translation with " + this.translator[0].label);
 
 			this.incrementAsyncProcesses("Zotero.Translate#translate()");
@@ -1359,8 +1359,9 @@ Zotero.Translate.Base.prototype = {
 				// doImport can return a promise to allow for incremental saves (via promise-returning
 				// item.complete() calls)
 				if (maybePromise) {
-					return maybePromise
+					await maybePromise
 						.then(() => this.decrementAsyncProcesses("Zotero.Translate#translate()"));
+          return;
 				}
 			} catch (e) {
 				this.complete(false, e);


dstillman:
Do you have code to trigger this?

This change alone wouldn't be correct. If the function is meant to return a promise, it should be marked as `async`, since it currently returns a non-promise `false` below. This change would [release Zalgo](https://blog.izs.me/2013/08/designing-apis-for-asynchrony).

larsgw:
[This is the code](https://github.com/citation-js/bibtex-parser-experiments/blob/master/src/zotero/index.js) (if there's some better trick to run the translator, let me know):

```js
import path from 'path'

process.env.NODE_CONFIG_DIR = path.resolve(__dirname, '../../node_modules/translation-server/config')

require('translation-server/src/zotero')
const Translators = require('translation-server/src/translators')
const Translate = require('translation-server/src/translation/translate')

export async function init () {
  await Translators.init()
}

export async function parse (text) {
  const translate = new Translate.Import()
  translate.setString(text)
  const translators = await translate.getTranslators()
  translate.setTranslator(translators[0])
  return translate.translate({ libraryID: 1 })
}
```

When the `parse` function is run with [this text file](https://github.com/citation-js/bibtex-parser-experiments/blob/master/test/files/syntax.bib):

```bibtex
@String {meta:maintainer = "Xavier D\\'ecoret"}

@
  %a
preamble
  %a
{ "Maintained by " # meta:maintainer }
@String(Stefan = "Stefan Swe{\\i}g")
@String(and = " and ")

@Book{sweig42,
  Author =	 stefan # and # meta:maintainer,
  title =	 { The {impossible} TEL---book },
  publisher =	 { D\\"ead Po$_{eee}$t Society},
  yEAr =	 1942,
  month =        mar
}
```

the parser errors. Consequently, `maybePromise` rejects which is unhandled.

larsgw:
And this is the warning/error, by the way:

```
node:1312) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'trim' of undefined
    at processField (eval at <anonymous> (/node_modules/translation-server/src/translation/sandboxManager.js:70:4), <anonymous>:339:41)
    at beginRecord (eval at <anonymous> (/node_modules/translation-server/src/translation/sandboxManager.js:70:4), <anonymous>:906:5)
    at readString (eval at <anonymous> (/node_modules/translation-server/src/translation/sandboxManager.js:70:4), <anonymous>:979:25)
    at eval (eval at <anonymous> (/node_modules/translation-server/src/translation/sandboxManager.js:70:4), <anonymous>:955:4)
    at new Promise (<anonymous>)
    at doImport (eval at <anonymous> (/node_modules/translation-server/src/translation/sandboxManager.js:70:4), <anonymous>:954:10)
    at Zotero.Translate.Import.rest (/node_modules/translation-server/modules/zotero/chrome/content/zotero/xpcom/translation/translate.js:1406:49)
    at loadPromise.then (/node_modules/translation-server/modules/zotero/chrome/content/zotero/xpcom/translation/translate.js:1396:39)
(node:1312) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 4)
(node:1312) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

larsgw:
> Also, if the function might return a promise, we'd want to mark it as `async`

I thought `Zotero.Promise.method()` did that.

larsgw:
I forgot the statement was inside `rest`, not `_translateTranslatorLoaded`. This commit should also make sure `this.complete` is called instead of passing the error to the parent, as seems to be done for synchronous parsers.

PR #1743, opened 2019-10-26
Use the label Programming Language

zuphilip:
This should use the same label as in searchConditions.programmingLanguage.

See also https://forums.zotero.org/discussion/78907/programming-language-is-labeled-as-language-only


dstillman:
Maybe we should abbreviate this as "Prog. Language"? Not that elegant, but neither is this:

field1

vs.

field2

zuphilip:
Yeah, I guess that would be better. There is also `Loc. in Archive` which is abbreviated. However, in other languages it might be different, e.g. `Programmiersprache` is equally long as `Standort im Archiv` or `Heruntergeladen am`.

Should I change to `Prog. Language`?

dstillman:
> Should I change to Prog. Language?

Yeah, let's do that.

zuphilip:
Done

PR #1748, opened 2019-10-31
Show bullet instead of '1' when a note exists

### LINKED ISSUE #1404: Show bullet instead of 1 for child attachment notes
https://github.com/zotero/zotero/issues/1329#issuecomment-355527868

No reason to show a "1" when it's just indicating the presence of a note.

nizarion:
Fixes #1404
This is my first pr in this repo :)

### DIFF
--- a/chrome/content/zotero/xpcom/itemTreeView.js
+++ b/chrome/content/zotero/xpcom/itemTreeView.js
@@ -1060,6 +1060,7 @@ Zotero.ItemTreeView.prototype.getCellText = function (row, column)
 	}
 	
 	var val;
+	let suffix = Zotero.hiDPISuffix;
 	
 	// Image only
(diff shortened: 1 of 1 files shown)


dstillman:
Thanks for contributing. You'll want to make sure you test PRs in a [Zotero build](https://www.zotero.org/support/dev/client_coding/building_the_standalone_client), though. This would return `chrome://zotero/skin/bullet_blue${suffix}.png` as literal column text.

The issue is also to show a bullet specifically for child attachment notes, not for parent items, as this would do. And we would probably want to show a text bullet, not the blue attachment icon.

In any case, the items list is currently being rewritten in React, so a fix for the XUL tree won't be relevant anyway. We'll need to wait to fix this until the React tree is done.

Issue #1702, opened 2019-06-08
DOI Manager functions


bwiernik:
Follow up from https://forums.zotero.org/discussion/77711/available-for-beta-testing-retracted-paper-notifications

The DOI Manager plugin (https://github.com/bwiernik/zotero-shortdoi/) has three functions:
1. Retrieve a DOI for items that don't have one.
1. Verify that a stored DOI is valid and clean it of URL components (cf. https://github.com/zotero/zotero/issues/556)
1. Retrieve shortDOIs (http://shortdoi.org/) for items.

The plugin can automatically be set to retrieve short or long DOIs (or verify DOIs) on import. 

If no DOI can be found, an invalid DOI is stored, or DOI lookup results in multiple possible hits, corresponding tags are added to the affected items.

Currently, the plugin stores shortDOIs by replacing the contents of the DOI field with the shortDOI. This works for citing and resolving DOI links by clicking on the DOI field or using the CrossRef lookup engine, but it breaks other functions, such as lookup through other APIs (e.g., Unpaywall, Web of Science), the Retrieve PDF function, and the RetractionWatch integration.

It would be nice to avoid this conflict, as well as integrate some or all of DOI Manager's features into Zotero itself. 

For storing shortDOIs, could we add `shortDOI: ` as a parse-able field to Extra? If present, this would replace the contents of the DOI field (or DOI stored in Extra) for export/citing. This could eventually be moved to a proper shortDOI field to permit copying as a URL or left as something parsed from Extra. This could be a hidden preference if a user using shortDOIs for some reason wanted to temporarily use the long DOIs.

Regarding the other functions:
1. DOI lookup and verifying the validity of DOIs could be incorporated into the forthcoming metadata update function
1. Retrieving shortDOI (manually or automatically on import) could also be added a function directly into Zotero

@retorquere

retorquere:
Wouldn't `Short DOI: ...` be more in keeping with how Zotero stores other extra-fields?

bwiernik:
The official name of these is "shortDOI". We would probably want to parse any casing and with or without spaces.

retorquere:
I was just thinking that the citation key is stored in the `extra` field as `Citation Key: ...` where it initially was `citationkey: ...` or somesuch, and I changed it at direction of Dan to `Citation Key`. The general rule seemed to be that it would go into the `extra` field as it would be expected to display in the UI of the English version of Zotero.

bwiernik:
[...]

PR #1758, opened 2019-12-11
Add option to cite shortDOIs instead of full DOIs

bwiernik:
Closes https://github.com/zotero/zotero/issues/1702

### DIFF
--- a/chrome/content/zotero/preferences/preferences_cite.xul
+++ b/chrome/content/zotero/preferences/preferences_cite.xul
@@ -40,6 +40,7 @@
 		<preferences id="zotero-preferences-cite">
 			<preference id="pref-cite-useClassicAddCitationDialog" name="extensions.zotero.integration.useClassicAddCitationDialog" type="bool"/>
 			<preference id="pref-cite-citePaperJournalArticleURL" name="extensions.zotero.export.citePaperJournalArticleURL" type="bool"/>
+			<preference id="pref-cite-useShortDOI" name="extensions.zotero.export.useShortDOI" type="bool"/>
 		</preferences>
 		
 		<tabbox>
@@ -80,6 +81,7 @@
 						<label id="export-citePaperJournalArticleURL" width="45em">
 							&zotero.preferences.export.citePaperJournalArticleURL.description;
 						</label>
+						<checkbox label="&zotero.preferences.export.useShortDOI;" preference="pref-cite-useShortDOI"/>
 					</groupbox>
 					
 					<groupbox>
--- a/chrome/content/zotero/xpcom/cite.js
+++ b/chrome/content/zotero/xpcom/cite.js
@@ -651,6 +651,14 @@ Zotero.Cite.System.prototype = {
 				delete cslItem.accessed;
 			}
 		}
+
+		if (!Zotero.Prefs.get("export.useShortDOI")) {
+			// substitute shortDOI for DOI if supplied
+			shorDOI = cslItem.extra.match(/^short\s?doi:\s*(.+)/mi);
+			if (shortDOI) {
+				cslItem.DOI = shortDOI[1];
+			}
+		}
 		
 		return cslItem;
 	},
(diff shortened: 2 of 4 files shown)


bwiernik:
Note that this will work with citeproc-js for both item types with proper DOI fields and DOI stored in Extra.

bwiernik:
@dstillman Do you have any comments on this? I'd like to update DOI Manager to store shortDOIs in Extra rather than the DOI field to stop it from interfering with Retraction notices, etc., but I can't really do that without this being merged/

dstillman:
Do shortDOIs still have any institutional backing, beyond a service that hasn't been taken down ("Updated 2 April 2013")? Wondering about, e.g., [this thread](https://twitter.com/thunkenizer/status/1043180863155056641), where @mfenner questions their long-term future.

DOI Manager should certainly switch to storing shortDOIs in Extra, and this option could be implemented in DOI Manager as well, with fairly trivial monkey-patching of `retrieveItem()`.

(There's also a typo in this patch, such that it will never work. `shorDOI = […]` should be `let shortDOI = […]`.)

mfenner:
shortDOIs have been deprecated by the DOI Foundation. I haven't seen any significant activity around using them, and I personally think supporting them leads to more confusion than being helpful to users.

bwiernik:
Monkey-patching is a bit beyond my knowledge. Do you have a toy example of a monkey patch I could look as a model?

dstillman:
Most of BBT, I believe. We don't particularly recommend it, since it's inherently fragile, but it's how @retorquere makes changes to lots of native functionality. The basic idea is that you save a reference to a function and replace it with a wrapper method that calls the original and modifies the input or (more likely) the output. In this case, you'd overwrite `retrieveItem()`, call your stored reference to the original function with the same arguments, and just apply the same modification to its return value that you're doing now, and then return that.

Untested, but it'd be something like this:

```javascript
function wrapRetrieveItem() {
	var origFunc = Zotero.Cite.System.prototype.retrieveItem;
	Zotero.Cite.System.prototype.retrieveItem = function () {
		var cslItem = origFunc.apply(Zotero.Cite.System.prototype, arguments);
		if (!Zotero.Prefs.get("export.useShortDOI")) {
			// substitute shortDOI for DOI if supplied
			let shortDOI = cslItem.extra.match(/^short\s?doi:\s*(.+)/mi);
			if (shortDOI) {
				cslItem.DOI = shortDOI[1];
			}
		}
		return cslItem;
	};
}
```

dstillman:
(Thanks, @mfenner.)

bwiernik:
Thanks @dstillman

retorquere:
BBT has enough of them that I've written a function to do it 🙄 .

I'd go as far as actively recommending against monkey-patching if it can be at all avoided. It makes the work for Zotero support harder, and you need to keep vigilant yourself that you understand very keenly the assumptions that the code you're patching makes, and what assumptions code that calls it makes, and that you can always still satisfy those. I've made a fair bit of gaffes, and that cost more often than it should falls to the Zotero developers.

That said, the basics are very simple. If you have

```
const marker = 'MyExtensionHasMonkeyPatched'

export function patch(object, method, patcher) {
  if (object[method][marker]) return
  object[method] = patcher(object[method])
  object[method][marker] = true
}
```

you can then do

```
patch(Zotero.Item.prototype, 'getField', original => function (field, unformatted, includeBaseMapped) {
  try {
    if (field === 'citekey') return 'fixed'
  } catch (err) {
    Zotero.debug(`patched getField:${err.message}`)
  }

  return original.apply(this, arguments)
})
```

it works the same for `async` functions (or any `coroutine`-wrapped equivalent):

```
patch(Some.object, 'method', original => async function (arg1, arg2) {
  await someStuff()
  return await original.apply(this, arguments) // or just return original.apply(this.arguments) if it's last
})
```

retorquere:
("Most of BBT" is fortunately no longer the case. I have 15 monkey patches in total currently)

bwiernik:
Thanks Emiliano! Much appreciated.

PR #1785, opened 2020-02-10
Expose ViewOptions from Zotero_LocateMenu

wshanks:
There have been some recent requests to add shortcut functions to Zutilo for items from the locate menu. Would it be okay to expose ViewOptions like this so it is easy to create the shortcuts? Without this, the easiest way I saw to make the shortcuts is to create a dummy menu element, run `Zotero_LocateMenu.buildContextMenu` on it, and then use `dispatchEvent(new Event("command")` on the appropriate menu item. That is pretty convoluted compared to calling `handleItems()` on the currently selected items.

### DIFF
--- a/chrome/content/zotero/locateMenu.js
+++ b/chrome/content/zotero/locateMenu.js
@@ -584,4 +584,7 @@ var Zotero_LocateMenu = new function() {
 			ZoteroPane_Local.loadURI(urls, event);
 		});
 	};
+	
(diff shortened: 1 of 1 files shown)


vitlib:
Is there any news on this front? I'd like to setup a shortcut for "view online" too

Yehuda-Bergstein:
Would be great if this is implemented!

PR #1786, opened 2020-02-15
Allow session history in a hidden browser

tnajdek:
Currently loading any page that uses history object `window.history` triggers `NS_ERROR_FAILURE`. Many single-page apps, including the new web-library, depend on availability of `window.history` and will crash without rendering anything if unexpectedly `NS_ERROR_FAILURE` is thrown when accessing `window.history`. 

This small change ensures that page loaded in the hidden browser can access `window.history`. However I'm not familiar enough with the codebase to say how would this affect other translators and other parts of Zotero that use `Zotero.Browser`.


dstillman:
Seems safe enough to me — we can revisit if there are any problems.

PR #1814, opened 2020-04-23
Added Zotero.Utilities.setTimeout routine to utilities_translate.js

pchemguy:
This utility has identical signature and functionality with setTimeout(). Its purpose is to be used by the web translators in situations when additional server requests are necessary for retrieving additional data/information and the target server or the nature of the request requires certain delays between successive requests. Basically, this routine wraps setTimeout() adding calls to incrementAsyncProcesses/decrementAsyncProcesses routines.


dstillman:
Can you provide an example where this is needed? We've gotten pretty far without needing this, and we're also trying to migrate to [async requests](https://github.com/zotero/zotero/pull/1609) in translators (where, if we really needed this, it could be done with a promise-returning delay() function), so I'd rather not add another callback-based function.

dstillman:
(I guess one could argue that due to <a href="https://www.zotero.org/support/kb/site_access_limits">site access limits</a> we haven't actually gotten this far without needing this, though at least in the GS case we don't have any way of knowing when it's safe to make additional requests, so adding arbitrary delays isn't exactly a great solution.)

pchemguy:
I submitted a translator for a Russian legal reference service. The submitted version does not include the possibility to save pdfs at present, but this is a work in progress. That website (docs.cntd.ru) offers pdfs for some of the documents for free. They do not provide direct links, however.  Apparently, pdfs they provide are dynamically generated upon request, unless still in cache. I inspected their web interface and found the protocol. A POST request must be made first to request a pdf. Then, the server is polled using GET requests every 2 seconds until the "ready" response is received. Only then a pdf can be downloaded.

PR #1846, opened 2020-07-17
Use SingleFile to create snapshots of web pages

fletcherhaz:


chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -519,6 +519,82 @@ Zotero.Utilities.Internal = {
...
+		// Run all the scripts of SingleFile scripts in Sandbox
+		SCRIPTS.forEach(script => {
+			loadSubScript('resource://zotero/SingleFile/' + script, sandbox);
+		});
+
+		// Wait 5s for everything to load and execute
+		function sleep(ms) {
+			return new Promise(resolve => setTimeout(resolve, ms));
+		}
+		await sleep(5000);
```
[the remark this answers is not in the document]
dstillman: Based on an earlier discussion, I believe this was to allow possible JS on the page to finish running and doing dynamic layout, not for SingleFile to run. (I'm not sure any of the SingleFile stuff above here is even async.)

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -519,6 +519,82 @@ Zotero.Utilities.Internal = {
...
+
+	/**
+	 * Takes in a document in a hidden web browser, creates a JS Sandbox and
+	 * executes the SingleFile extension to save the page as one single file
+	 * without JavaScript.
+	 *
+	 * @param {Object} document Document object inside hidden browser
+	 * @param {String|nsIFile} destFile Path for file to write to using Zotero.File
+	 */
+	 saveHTML: async function (document, destFile) {
```
dstillman, line 532: `saveHTMLDocument`?

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -519,6 +519,82 @@ Zotero.Utilities.Internal = {
...
 	},
+
+
+	/**
+	 * Takes in a document in a hidden web browser, creates a JS Sandbox and
+	 * executes the SingleFile extension to save the page as one single file
+	 * without JavaScript.
+	 *
+	 * @param {Object} document Document object inside hidden browser
+	 * @param {String|nsIFile} destFile Path for file to write to using Zotero.File
```
dstillman, line 530: Since the signature is changing anyway, this should only take a string path. nsIFile will be going away eventually.

No need to mention Zotero.File

chrome/content/zotero/xpcom/attachments.js
```diff
@@ -722,23 +722,31 @@ Zotero.Attachments = new function(){
...
+
 		var libraryID = options.libraryID;
 		var document = options.document;
 		var parentItemID = options.parentItemID;
 		var title = options.title;
 		var collections = options.collections;
 		var saveOptions = options.saveOptions;
+		var content = options.content;
+		var url = options.url;
+		var contentType = options.contentType;
```
dstillman, line 734: I assume these lines and the conditional below were from an earlier version (before `importFromPageData()`?) and can be removed?

.gitmodules
```diff
@@ -26,3 +26,9 @@
 	path = resource/schema/global
 	url = git://github.com/zotero/zotero-schema.git
 	branch = master
+[submodule "resource/SingleFile"]
+	path = resource/SingleFile
+	url = https://github.com/gildas-lormeau/SingleFile.git
+[submodule "resource/SingleFileZ"]
+	path = resource/SingleFileZ
+	url = https://github.com/gildas-lormeau/SingleFileZ.git
```
dstillman, line 34: Why do we need both?

chrome/content/zotero/xpcom/connector/server_connector.js
```diff
@@ -958,20 +962,43 @@ Zotero.Server.Connector.SaveSnapshot.prototype = {
...
-				parentItemID: itemID,
-				pageData: data.pageData
-			});
-			
-			// await Zotero.Attachments.importFromDocument({
-			// 	document: doc,
-			// 	parentItemID: itemID
-			// });
+			if (data.pageData) {
+
```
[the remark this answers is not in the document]
dstillman: (Not sure if you have ESLint set up in your editor, but it should flag this sort of thing (`padded-blocks`, in this case). A lot of our existing code will show warnings or errors, but it's good to make sure new code conforms.)

chrome/content/zotero/xpcom/server.js
```diff
@@ -605,35 +605,42 @@ Zotero.Server.DataListener.prototype._requestFinished = function (response, opti
...
-			for (let nameVal of contentDisposition[1].split(';')) {
-				nameVal.split('=');
-				fieldData[nameVal[0]] = nameVal.length > 1 ? nameVal[1] : null;
+		// Extract the name to for this part from the header
+		// e.g. Content-Disposition: form-data; name="uri"
+		let contentDisposition = contentDispositionRe.exec(header);
+		if (contentDisposition && contentDisposition.length > 1) {
+			// Remove Content-Disposition prefix
+			contentDisposition.shift();
+			for (let pair of contentDisposition) {
```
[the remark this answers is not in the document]
dstillman: But this sort of thing should also have tests in serverTest.js. See the examples there for creating test endpoints. We should have tests for various kinds of multipart/form-data requests.

chrome/content/zotero/xpcom/translation/translate_item.js
```diff
@@ -857,19 +858,35 @@ Zotero.Translate.ItemSaver.prototype = {
...
 		}
 		
-		Zotero.debug('Importing attachment from URL');
 		attachment.linkMode = "imported_url";
 		
 		attachmentCallback(attachment, 0);
-		
+
+		// Import from SingleFileZ Page Data
+		if (attachment.pageData && this._pageData) {
```
dstillman, line 873: Why both?

fletcherhaz: No particular reason. Just a habit of checking to try and prevent errors. `attachment.pageData` is simply a boolean indicating the attachment should have pageData. `this._pageData` is the global object holding the pageData itself. If any attachment has pageData the global object should exist.

dstillman: But is there any reason for the global object, as opposed to just passing the data with the attachment it pertains to?

dstillman: While it wouldn't be the case now if we're only running SingleFile on the current page in the connector, couldn't there in theory be multiple attachments with page data?

fletcherhaz: The reason I had the global object was because I was accounting for the scenario where the connector saves multiple items with the same page data. For example [https://ui.adsabs.harvard.edu/search/p_=0&q=test&sort=date%20desc%2C%20bibcode%20desc](this page) saves multiples items with a snapshot of the search text and a pdf from the result. Passing 10 copies of the snapshot data seemed to be compounding the problem of size so I passed one as a global.

SingleFile for multiple attachments only makes sense if we are loading those in the browser somehow to retrieve the page contents and I thought we would handle that in the client with a hidden browser. That may very well be the wrong approach though!

chrome/content/zotero/xpcom/server.js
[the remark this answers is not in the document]
fletcherhaz: I was breaking some old endpoints when I did this. So I moved my work to the Utilities object I created. This content disposition code does not produce coherent results for my requests and I'm not sure what the requests look like that this was originally for.

chrome/content/zotero/xpcom/connector/server_connector.js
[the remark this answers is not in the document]
fletcherhaz: I now have ESLint set up. Sorry for not getting that done earlier.

chrome/content/zotero/xpcom/translation/translate_item.js
dstillman: Oh, yeah, that's just bad behavior on the part of the translator — it uses the same scrape() function for both single and multiple results, and it's passing the parent doc to every item, which makes sense for single items but not for multiple. So we should just fix that translator, and we definitely don't need to optimize for that here. Better to just keep this scoped to the attachment.

fletcherhaz: Ok that makes sense. Thanks! I will switch that back now.

dstillman:
Is there a test that covers the getBoundingClientRect() issue?

dstillman:
Looks like `-g "should save a document with embedded files"` fails for a different reason from the `getBoundingClientRect()` issue: it's looking for img.gif from the snapshot, but SingleFile saves images/2.gif. (I'm not sure what happened to 1.gif.) We still need to update this to save ZIP files instead and then render those in-app, which I can help with, but for now just to get tests passing if it's consistently going to be images/2.gif we should update the test accordingly (though I'm still curious why it's 2.gif).

fletcherhaz:
Ok. I'll look at getting the current tests working. I was able to replicate the `getBoundingClientRect` issue by changing the url on the saved attachment in this test: `should save a translated item to the current selected collection` to `https://www.nytimes.com/2020/08/04/world/coronavirus-cases.html` so it must be something particular with that page. 

It appears that `https://www.example.org` causes the test to pass and `https://www.zotero.org` fails for a different reason altogether!

fletcherhaz:
Regarding `-g "should save a document with embedded files"`, SingleFile renames every attached resource to numerically increasing numbers, sometimes with extensions, sometimes without. I'm not sure why it skipped 1 though. How are you viewing what files are created in that test?

dstillman:
> How are you viewing what files are created in that test?

I was just going by debug output, which lists images/2.gif, but you can also add something like this to the test:

```javascript
this.timeout(1000000);
yield Zotero.Promise.delay(1000000); // or await if the test uses async
```

And then just look in the temporary data directory path (e.g., /var/folders/ss/…) it outputs near the top of the debug output.

fletcherhaz:
@dstillman I tried to use FileUtil and NetUtil for the binary write, but couldn't quite get it to work. I found the example snippet (linked in the documentation), but if you know a better way to do this please let me know.

chrome/content/zotero/xpcom/file.js
```diff
@@ -447,6 +447,46 @@ Zotero.File = new function(){
...
+					stream.close();
+				}
+				resolve();
+			}
+			catch (e)
+			{
+				reject(e);
+			}
+		});
+	};
```
dstillman, line 487: Nothing in here is async, so if we were to keep this, it would be `putBinaryContents()`, and the `async`/`await`/`try`/`catch` would need to be removed. But `putContentsAsync()` should already handle this. It can take an input stream, ArrayBuffer, or Blob, and there's an existing test, "should save a Blob", for writing a binary blob.

fletcherhaz:
Here is where I switched to using `NetUtil.readInputStreamToString` which can read arbitrary binary data without decoding it:
https://github.com/zotero/zotero/pull/1846/files#diff-a62120bad5179c0ece7cab1f12a9cb15L195

I found that on MDN. I'm not super familiar with how JS handles binary data in strings and what kind of automatic decoding it is doing. If that is not safe enough I can simply read using UTF-8 charset until `headerFinished` then read the body using no charset.

The bigger problem I didn't know how to solve was that each part of the multipart request can contain normal text in the header and binary or normal text in the body. I didn't see a way to operate on those both without both types of data in the same structure. JS strings seemed to be happy with arbitrary binary nonsense in the middle so I went with that. However, if that assumption is poor and/or bad practice I would definitely appreciate a pointer on how to handle that. If that's not clear let me know and I can try to explain further.

chrome/content/zotero/xpcom/file.js
fletcherhaz: I spent some time on the MDN site looking for a stream that I could use to write a binary string using putContentsAsync, but had no luck. I tried this:
```
Components.utils.importGlobalProperties(["Blob"]);
let blob = new Blob([resource], { type: contentType });
await Zotero.File.putContentsAsync(
	filePath,
	blob
);
```
but the file was corrupted. This will depend on the data coming out of the processEndpoint so maybe I will wait until we get that settled first unless you see something obvious I'm missing?

dstillman:
The first step would be to create a test in serverTest.js that 1) creates a new endpoint that expects binary data, 2) replicates a multipart/form-data request, with an embedded File from test.png in the test/tests/data directory (`File.createFromFileName(path)`), and 3) extracts the file from the form data in the endpoint and verifies it against the file. I believe you should be able to do the last part with `Zotero.File.getBinaryContentsAsync()` and `[FileReader.readAsBinaryString(blob)`, though I haven't checked that.

See [here](https://developer.mozilla.org/en-US/docs/Web/API/DOMString/Binary) for a bit of information on binary strings in JS. They're a bit of a hack, but they make it easy to compare binary data, among other things.

It's OK if the test fails or is incorrect, but it will make it much easier and quicker to debug this, and it needs to exist anyway.

test/tests/serverTest.js
```diff
@@ -141,6 +141,99 @@ describe("Zotero.Server", function () {
...
+							},
+							body: formData
+						}
+					);
+
+					assert.ok(called);
+					assert.equal(req.status, 204);
+				});
+				
+				it("should support binary", function* () {
```
dstillman, line 192: For new code, use `async` and `await` rather than generators and `yield`.

chrome/content/zotero/xpcom/connector/server_connector.js
```diff
@@ -384,6 +384,70 @@ Zotero.Server.Connector.SaveSession.prototype._updateRecents = function () {
...
+		let utf8Stream = Components.classes["@mozilla.org/intl/converter-input-stream;1"]
+					.createInstance(Components.interfaces.nsIConverterInputStream);
+		utf8Stream.init(dataStream, "UTF-8", 4096, "?");
+
+		let decoded = "";
+		let string = {};
+		while (utf8Stream.readString(data.length, string)) {
+			decoded += string.value;
+		}
```
dstillman, line 444: I know this is copied from `_decodeRawData()`, but these days we can do this in a more standards-compliant fashion with TextDecoder:

```javascript
var buf = new ArrayBuffer(data.length);
var bufView = new Uint8Array(buf);
for (let i = 0; i < data.length; i++) {
	bufView[i] = data.charCodeAt(i);
}
var decoder = new TextDecoder();
return decoder.decode(bufView);
```

(Not fully tested, but something like that should work.)

Probably better to add that to a Utilities.Internal.decodeUTF8(data) function (maybe above `byteLength()`) that we can just call from both of these places.

chrome/content/zotero/xpcom/file.js
dstillman: `putContentsAsync()` will happily take a Blob with valid data. E.g., this will create an exact copy:

```javascript
var data = await File.createFromFileName('/Users/dan/Desktop/z.png');
await Zotero.File.putContentsAsync('/Users/dan/Desktop/z2.png', data);
```

If `resource` is a binary string in `new Blob([resource]…)`, that won't work — you would have to convert that to a typed array, I think, to keep it from being interpreted as UTF-8 — but don't you have a Blob from the Connector now anyway?

dstillman (changes requested):

chrome/content/zotero/xpcom/file.js
fletcherhaz: Ah right. `resource` is just a binary string from the multipart data so I can use charCodeAt and then convert it to a Blob. Thanks.

fletcherhaz:
@dstillman The one issue I forgot about is I still haven't solved the `getBoundingClientRect` issue (https://github.com/zotero/zotero/pull/1846#issuecomment-678550953).

Changing this line (https://github.com/fletcherhaz/zotero/blob/snapshot/test/tests/attachmentsTest.js#L312) to:
```
var uri = "https://www.twitter.com/zotero";
```
replicates the issue. The normal test succeeds so I'm not exactly sure what that means.
```
./test/runtests.sh -f -b -d 1 -g "#importFromDocument()" attachments
```

I want try and fix one issue with the Progress Window and look over all the code changes again, but I think those are my only outstanding items.

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -360,6 +360,35 @@ Zotero.Utilities.Internal = {
...
+	},
+	
+	
+	/**
+	 * Decode a binary string to UTF-8 string
+	 *
+	 * @param {String} data - Binary string to decode
+	 * @return {String} UTF-8 encoded string
+	 */
+	decodeToUTF8: function (data) {
```
dstillman, line 385: I think `decodeUTF8` would be more accurate. It's not really decoding _to_ UTF-8 — the bytes are UTF-8-encoded text, and we're decoding that to a JavaScript string.

PR #1827, opened 2020-06-03
Virtualized tree

adomasven:
See the first commit message. Sorry for the single commit. Makes #1636 obsolete

PR #1849, opened 2020-07-20
Migrate Wizards from XUL to React

tnajdek:
See #2311

- [x] Import
- [x] Publications
- [x] Report Error
- [x] Rtf Scan (requires item tree ported)
- [ ] Merge (requires item box)


chrome/content/zotero/xpcom/zotero.js
```diff
@@ -1368,6 +1368,13 @@ Services.scriptloader.loadSubScript("resource://zotero/polyfill.js");
 	this.randomString = function(len, chars) {
 		return Zotero.Utilities.randomString(len, chars);
 	}
+
+	this.nextHtmlId = (prefix = 'id-') => {
+		if (!this._htmlId) {
+			this._htmlId = 1;
+		}
+		return prefix + this._htmlId++;
+	}
```
adomasven, line 1377: I've been putting things needed for React UI here https://github.com/zotero/zotero/blob/7b77c0cd27c9c5c78ce4f14bb517024ccf72de2f/chrome/content/zotero/components/utils.js

(There's more stuff coming to that file with the tree PR). We may need to figure out a better, more appropriately named place for those in the future, but a function like this should certainly not live on the main Zotero object.

adomasven:
Dan almost always wants things to look exactly the same as they did in the previous versions of Zotero, and while this Wizard looks slightly different on Ubuntu than the XUL one, I think it is better styled here. Except for the checkbox/radiobutton labels which seem to not be properly vertically-aligned/centered.

tnajdek:
Hi, thanks for reviewing this! 

On macOS layout & styling is pretty much identical between the XUL and React version, but I didn't test other platforms. The only difference as far as I can tell is HTML version handles resizing better on the page where you select database from the list when importing from Mendeley.

Also, from what I can see, "Go Back" was there in XUL version on macOS, perhaps there are functional differences between platforms? I'll definitely have a look how it looks like on Window/Linux.

tnajdek:
Updated PR to include Windows & Linux specific shenanigans. I've removed "Back" button on Linux to match native wizards on that OS, not sure if that's a "feature" we want to include. 

On Windows, it's no longer possible to press alt+n/alt+b to navigate next/previous - preserving this feature is super-tricky because of internationalization.

Please note that changes in this branch currently only work if locale is set to `en_US`, otherwise `Zotero.getString` crashes and import wizard renders empty window.

tnajdek:
I've now moved `nextHtmlId()` to `utils.js` as suggested.

tnajdek:
I've now also ported Publications Dialog, forced pushed commits and updated the PR to reflect that. This PR also now fixes #1868 currently present in XUL version.

tnajdek:
Just realized RtfScan Wizard feature needs a table and the item selector so it will require #1827

chrome/content/zotero/components/importDatabaseTable.jsx
```diff
@@ -0,0 +1,160 @@
+/*
```
dstillman, line 1: Move to an `import` subfolder

chrome/content/zotero/components/utils.js
```diff
@@ -1,5 +1,83 @@
...
+	
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero.  If not, see <http://www.gnu.org/licenses/>.
+	
+	***** END LICENSE BLOCK *****
+*/
+
 const noop = () => {};
 
+var _htmlId = 1;
```
dstillman, line 28: Style: `_htmlID`

chrome/content/zotero/components/utils.js
```diff
@@ -1,5 +1,83 @@
...
+
+const nextHtmlId = (prefix = 'id-') => prefix + _htmlId++;
```
dstillman, line 30: Style: `nextHTMLID`

chrome/content/zotero/components/utils.js
```diff
@@ -1,5 +1,83 @@
...
+	if (elementTop < containerTop || elementBottom > containerBottom) {
+		const before = container.scrollTop;
+		element.scrollIntoView(opts);
+		const after = container.scrollTop;
+		return after - before;
+	}
+	return 0;
+};
+
+const getLicenseData = (license) => {
```
dstillman, line 47: Move to something in the `publications` folder — no real reason to think this will be a general thing

dstillman (changes requested):
Sorry for the delay on this. Let's rebase this and get it in.

Made a few notes, but haven't run it yet — I'll do that after the rebase.

tnajdek:
Thanks for the feedback, I'll look into this!

Also, now that #1827 has been merged, I should be able to migrate one more wizard (but I'll get this up to date first).

dstillman:
Did you take out the local Mendeley importer altogether? We still want that — we [link to that](https://www.zotero.org/support/kb/mendeley_import#alternative_method) as an alternative method for when people have trouble with the online installer (which happens sometimes, for unclear reasons related to authentication).

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -1250,6 +1275,27 @@ publications.buttons.choose-sharing = Sharing
 publications.buttons.choose-license = Choose a License
 publications.buttons.addToMyPublications = Add to My Publications
 
+publications.my_publications = My Publications
+
+publications.intro = Items you add to My Publications will be shown on your profile page on org. If you choose to include attached files, they will be made publicly available under the license you specify. Only add work you yourself have created, and only include files if you have the rights to distribute them and wish to do so.
```
dstillman, line 1280: "zotero." got lost?

chrome/content/zotero/error/errorReport.jsx
```diff
@@ -0,0 +1,182 @@
...
+				<div className="report-result">
+					<label htmlFor={ id.current + '-report-result' }>
+						{ Zotero.getString('errorReport.reportID') }
+					</label>
+					<input
+						className="report-id"
+						id={ id + '-report-result' }
+						readOnly={ true }
+						value={ reportID }
+					/>
```
dstillman, line 159: Why not just wrap the `input` in the `label` and skip the `for` (here and elsewhere)?

tnajdek:
> Did you take out the local Mendeley importer altogether? We still want that — we [link to that](https://www.zotero.org/support/kb/mendeley_import#alternative_method) as an alternative method for when people have trouble with the online importer (which happens sometimes, for unclear reasons related to authentication).

That still works via file import as described. I've removed a radio button for local Mendeley import and UI to deal with finding and selecting the Mendeley DB that we've previously ported to react and that, I believe, we don't need anymore.

dstillman:
Ah, yeah, OK — that's right.

chrome/locale/en-US/zotero/zotero.properties
tnajdek: I've reinstituted `zotero.` prefix for locale strings ported from `publications.dtd` (which `Zotero.getString` doesn't retrieve from). There is a slight drawback of this: previously prefixed strings would only appear in `.xul` file, `.js` files used non-prefixed ones, now both kinds are mixed in `publicationsDialog.jsx`.

chrome/content/zotero/components/licenseInfo.jsx
```diff
@@ -0,0 +1,89 @@
+/*
```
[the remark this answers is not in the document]
tnajdek: I've ended up putting that component in the same file as `PublicationsDialog` as it is unlikely it will be used on its own.

chrome/content/zotero/components/importDatabaseTable.jsx
tnajdek: File removed as no longer needed.

chrome/content/zotero/components/utils.js
tnajdek: Merged into `publicationsDialog.jsx`

chrome/content/zotero/error/errorReport.jsx
tnajdek: Both are perfectly valid, however few reasons to keep explicit `for`:

* Recommended by [W3C for accessibility reasons](https://www.w3.org/WAI/tutorials/forms/labels/#associating-labels-implicitly)
* Generally easier to style this way
* We're [using](https://github.com/zotero/web-library/issues/134#issuecomment-481601752) explicit `for` in Web Library 

I actually came to prefer it this way, not only for the above reasons, but also because, semantically, it makes little sense for an input to be part of a label that describes it (other way round would actually make more sense!)

I'd recommend keeping explicit `for` but can change it if it's a problem.

chrome/locale/en-US/zotero/zotero.properties
dstillman: Oh, no, sorry — we don't need (and don't want to) to prefix the string keys with `zotero.`. I was talking about `will be shown on your profile page on org` in the text. Presumably just a find-and-replace gone wrong.

chrome/content/zotero/error/errorReport.jsx
dstillman: Yeah, I've seen the better-supported-by-accessibility-tools argument. That seems like a problem with the tools, not the method, and hopefully one that's been fixed in the years since all the mentions of those limitations. It's just particularly annoying in the context of React because of the need for id mechanism, but maybe worth it for the easier styling. (And the semantic argument is fair! XUL did get this right…)

dstillman:
Just as a general question, do we actually want to rename everything to .jsx? Babel doesn't require it, and we have some JSX in .js files already. Everything we put in a file is dependent on what kind of transpilation we do, and one could imagine other technologies that we might integrate, so it seems a little arbitrary to rename the whole file just because we use some JSX in it.

chrome/locale/en-US/zotero/zotero.properties
tnajdek: ohh... I see it now, thanks, fixed!

dstillman:
Re: renaming, maybe the argument is that the things we name .jsx are basically just React components, and .jsx is a way of communicating that? But if we just add some JSX into a file that's not a dedicated component, not renaming it makes sense.

tnajdek:
Re: renaming, my thinking was that maybe in future we don't process .js files during the build (because browser provides all the required features) but we still process .jsx files because of JSX syntax.

tnajdek:
@adomasven @dstillman I think this one is ready for another round of reviews.

Some notes & known issues:

* I didn't port "Merge" wizard yet as it probably makes more sense to do item box react port first.
* On macOS, in rtfscan form, on the citations review step, the column divider is off in the virtualised table. This also happens for me in non-react version so probably should be a separate issue.
* Also the virtualised table is not accessible via keyboard so rtfscan wizard cannot be completed without a pointer device.
* On Windows tab-order is the same as on other platforms but visually "cancel" follows back/next so it feels off. This can be fixed by producing platform-specific markup rather than repositioning via css. (I'll fix this one)
* It seems we have files named both `camelCase.js` and `dash-case.js`. `camelCase.js` seems to be dominant so I've followed this but we should probably stick to one. This is even more confusing when it comes to `.scss` files. Happy to change both to whatever we decide
* Also all issues listed in my previous comment are still valid: https://github.com/zotero/zotero/pull/1849#issuecomment-667110025

tnajdek:
Closing this in favor of #2311 which is all this + bulk folder import, adapted for fx102 branch.

PR #1857, opened 2020-07-28
Fix babel worker file-specific fixes check on Windows

tnajdek:
Everywhere else we use `glob` for paths which deals with forward/backward slash apart from `babel-worker.js` where paths were compared verbatim, hence file-specific fixes weren't applied during the build on windows. This PR resolves this issue by using regex check instead.


dstillman:
Just to keep things cleaner and make it easier to add others later, how about abstracting these into something like `pathContains(sourcefile, 'resource/react-dom.js')`?

tnajdek:
Abstracted logic to `comparePaths` util function, rebased on top of current master.

PR #1873, opened 2020-09-04
Update engines.json, add 3 search engines

qizidog:
add 3 search engines, cnki, sci-hub and web of science, hope to help.


yhu266:
I'd like to point out that sci-hub is not legal in many jurisdictions.

PR #1883, opened 2020-09-28
Switch Advanced Search window to React

fletcherhaz:


fletcherhaz:
Currently status of things still not done/tested:
- Saving search/editing later
- Submenus/tooltips
- Autocomplete
- Styling needs to be cleaned up and tested on all platforms
- The files are sort of all over the place
- Go over zoterosearch.xml and double check we have everything

adomasven:
This will introduce merge conflicts as far as changes to how the item tree in the advanced search dialog works

fletcherhaz:
I've left the item tree entirely alone and didn't see anything in #1827 that would interfere. The window is still XUL it just has two HTML elements inside it for the search conditions and buttons. Can you be more specific about what conflicts you see?

AbeJellinek:
Superseded by #2631.

PR #1892, opened 2020-10-10
increase color of tags

senz1024:
Not limiting the choices to 9 colors.


senz1024:
I'm sorry This is mistake.

PR #1901, opened 2020-10-13
Add an option to create parent item from identifier

fletcherhaz:


chrome/content/zotero/lookup.js
```diff
@@ -151,7 +199,9 @@ var Zotero_Lookup = new function () {
 		var search = multiline ? event.shiftKey : !event.shiftKey;
 		if(keyCode === 13 || keyCode === 14) {
 			if(search) {
-				Zotero_Lookup.accept(textBox);
+				// Copy the child item in case the user clicks focus away and wants
+				// to move on to a new item
+				Zotero_Lookup.accept(textBox, this.childItem ? { ...this.childItem } : false);
```
dstillman, line 204: Why destructure instead of just passing `this.childItem`?

chrome/content/zotero/zoteroPane.js
```diff
@@ -4550,6 +4552,28 @@ var ZoteroPane = new function()
...
+		let items = this.getSelectedItems();
+		for (let item of items) {
+			if (!item.isTopLevelItem() || item.isRegularItem()) {
+				throw new Error('Item ' + item.itemID + ' is not a top-level attachment or note in ZoteroPane_Local.createParentItemFromIdentifier()');
+			}
+
+			Zotero_Lookup.showPanelXY(node.popupBoxObject.x, node.popupBoxObject.y, {
+				libraryID: item.libraryID,
+				collections: item.getCollections(),
+				id: item.id
```
dstillman, line 4571: Can't we just pass `item` here as `childItem`, store the actual item, and get the libraryID and collections from that in Zotero_Lookup?

chrome/content/zotero/zoteroPane.xul
```diff
@@ -286,6 +286,7 @@
 					<menuitem class="menuitem-iconic zotero-menuitem-unrecognize" label="&zotero.items.menu.unrecognize;" oncommand="ZoteroPane.unrecognizeSelected()"/>
 					<menuitem class="menuitem-iconic zotero-menuitem-report-metadata" label="&zotero.items.menu.reportMetadata;" oncommand="ZoteroPane.reportMetadataForSelected()"/>
 					<menuitem class="menuitem-iconic zotero-menuitem-create-parent" oncommand="ZoteroPane_Local.createParentItemsFromSelected();"/>
+					<menuitem class="menuitem-iconic zotero-menuitem-create-parent-from-identifier" label="&zotero.items.menu.parentFromIdentifier;" oncommand="ZoteroPane_Local.createParentItemFromIdentifier(this.parentNode);"/>
```
dstillman, line 289: Just `ZoteroPane` is fine. (`ZoteroPane_Local` is an old thing.)

chrome/locale/en-US/zotero/zotero.dtd
```diff
@@ -100,6 +100,7 @@
 <!ENTITY zotero.items.menu.mergeItems					"Merge Items…">
 <!ENTITY zotero.items.menu.unrecognize                "Undo Retrieve Metadata">
 <!ENTITY zotero.items.menu.reportMetadata             "Report Incorrect Metadata">
+<!ENTITY zotero.items.menu.parentFromIdentifier             "Create Parent Item From Identifier">
```
dstillman, line 103: Lowercase `from` is house title-case style

dstillman (changes requested):
This is great, but trying this out, I don't think we should have both "Create Parent Item" and "Create Parent Item from Identifier". We should just have a single "Create Parent Item…" option that lets you enter an identifier, with a primary "Add Item" (or something) button in the bottom right and secondary "Manual Entry" button in the bottom left. This will help convey that you really shouldn't enter data manually when you have a choice.

Also, as nice as it is to reuse the lookup panel as is, we really shouldn't suggest that you can enter multiple identifiers, so we need an alternative version of the caption text without plural identifiers.

fletcherhaz:
@dstillman I reworked this. Could you take a look at this and let me know what you think?

chrome/content/zotero/zoteroPane.js
```diff
@@ -4517,20 +4504,33 @@ var ZoteroPane = new function()
...
-		for (var i=0; i<items.length; i++) {
-			var item = items[i];
-			if (!item.isTopLevelItem() || item.isRegularItem()) {
-				throw('Item ' + itemID + ' is not a top-level attachment or note in ZoteroPane_Local.createParentItemsFromSelected()');
-			}
-			
-			yield Zotero.DB.executeTransaction(function* () {
+		let item = this.getSelectedItems()[0];
+		if (!item.isAttachment() || !item.isTopLevelItem()) {
+			throw('Item ' + itemID + ' is not a top-level attachment');
```
dstillman, line 4515: `throw` → `throw new Error` (that's why we can remove the function name)

dstillman:
Let's try this:

mockup

- Just the PDF as the title. "Create parent item for:" with a colon is a bit weird.
- I think "Create Empty Parent" → "Manual Entry" makes more sense. "Create Empty Parent" is technically more accurate — it's the creation of the empty item that's happening immediately — but we don't really say "Parent" without "Item", and "Manual Entry" makes it a little clearer what you're setting yourself up for by creating an empty item.
- Add `<?xml-stylesheet href="chrome://zotero-platform/content/overlay.css" type="text/css"?>` to get proper the font for the input box on macOS
- `font-size: 14px` for the input
- Slight change in order of identifiers in placeholder, and no colon
- Narrower width
- Align buttons with input box

dstillman:
Also, for what it's worth, we kind of gave up on the container/component split thing. If there's no compelling reason to keep that, it's fine to combine into just a component. In the past I've found it a little useful to develop a complicated component outside of Zotero, with nothing Zotero-specific in the component, but that's probably not necessary here.

chrome/content/zotero/components/createParent/createParent.jsx
```diff
@@ -26,33 +26,37 @@
 'use strict';
 
 import React, { memo } from 'react';
+import ReactDom from "react-dom";
```
dstillman, line 29: `ReactDom` → `ReactDOM`

dstillman:
A few more things:

- Let's go with `padding: 20px 15px;` instead of `2em` for the dialog padding
- "to identify this item" → "to identify this file" — I guess this depends on whether you're identifying the file or identifying the parent item you're about to create, but given that you start with a file and we're putting the filename above the text box in bold, "file" seems more right to me
- We should disable the accept button when the input box is empty

PR #1904, opened 2020-10-20
Switch back to SingleFile from SingleFileZ

fletcherhaz:
Our SingleFileZ integration would save images inside directories following the
SingleFileZ format. However, Zotero does not support syncing sub-directories of
attachments. This commit switch back to a single HTML file with base64 encoded
resources. We think that the 33% increase in resources will be offset by the
compression of HTML and removal of JavaScript and unused CSS.

This commit does not fix past snapshots that were saved using SingleFileZ.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -29,3 +29,6 @@
 [submodule "resource/SingleFileZ"]
 	path = resource/SingleFileZ
 	url = https://github.com/gildas-lormeau/SingleFileZ.git
+[submodule "resource/SingleFile"]
+	path = resource/SingleFile
+	url = https://github.com/gildas-lormeau/SingleFile.git
--- a/chrome/content/zotero/xpcom/attachments.js
+++ b/chrome/content/zotero/xpcom/attachments.js
@@ -349,8 +349,8 @@ Zotero.Attachments = new function(){
 		}
 		return attachmentItem;
 	});
-	
-	
+
+
 	/**
 	 * @param {Object} options
 	 * @param {Integer} options.libraryID
@@ -400,14 +400,6 @@ Zotero.Attachments = new function(){
 				var browser = Zotero.HTTP.loadDocuments(
 					url,
 					Zotero.Promise.coroutine(function* () {
-						let channel = browser.docShell.currentDocumentChannel;
-						if (channel && (channel instanceof Components.interfaces.nsIHttpChannel)) {
-							if (channel.responseStatus < 200 || channel.responseStatus >= 400) {
-								reject(new Error("Invalid response " + channel.responseStatus + " "
-									+ channel.responseStatusText + " for '" + url + "'"));
-								return false;
-							}
-						}
 						try {
 							let attachmentItem = yield Zotero.Attachments.importFromDocument({
 								libraryID,
@@ -764,8 +756,11 @@ Zotero.Attachments = new function(){
 					&& Zotero.Translate.DOMWrapper.unwrap(document) instanceof Ci.nsIDOMDocument) {
 				if (document.defaultView.window) {
 					// If we have a full hidden browser, use SingleFile
-					Zotero.debug('Saving document with saveHTMLDocument()');
-					yield Zotero.Utilities.Internal.saveHTMLDocument(document, tmpFile);
+					Zotero.debug('Getting snapshot with snapshotDocument()');
+					let snapshotContent = yield Zotero.Utilities.Internal.snapshotDocument(document);
+
+					// Write main HTML file to disk
+					yield Zotero.File.putContentsAsync(tmpFile, snapshotContent);
 				}
 				else {
 					// Fallback to nsIWebBrowserPersist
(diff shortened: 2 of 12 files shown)


chrome/content/zotero/xpcom/attachments.js
```diff
@@ -349,6 +349,39 @@ Zotero.Attachments = new function(){
 		}
 		return attachmentItem;
 	});
+
+
+	this.hiddenBrowserImport = function (url, cookieSandbox, importFunction) {
```
dstillman, line 354: I'm not sure this belongs here — it's not really about attachments or importing. This is really just a version of loadDocuments() that enforces a 2xx response code, right? Why don't we just put that in loadDocuments itself, and have it fail (so, throw Zotero.HTTP.UnexpectedStatusException (with a fake `xmlhttp` that has a status) or call onError()) instead of calling the processor if there's a non-2xx code? I doubt it's ever going to be necessary or desirable to process an error page.

chrome/content/zotero/xpcom/attachments.js
```diff
@@ -846,22 +882,22 @@ Zotero.Attachments = new function(){
 	
 	
 	/**
-	 * Save a snapshot from a page data given by SingleFileZ
+	 * Save a snapshot from a HTML page content given by SingleFile
```
dstillman, line 885: Typo: "from a HTML" → "from HTML"

test/tests/server_connectorTest.js
```diff
@@ -792,6 +792,182 @@ describe("Connector Server", function () {
...
+				snapshotContent: await Zotero.File.getContentsAsync(indexPath)
+			}));
+
+			await Zotero.HTTP.request(
+				'POST',
+				connectorServerPath + "/connector/saveSingleFile",
+				{
+					headers: {
+						"Content-Type": "application/json",
+						"zotero-allowed-request": "true"
```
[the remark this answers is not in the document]
fletcherhaz: Yes. I switched back from multipart requests, which needed that but did not remove them.

chrome/content/zotero/xpcom/attachments.js
fletcherhaz: Sounds good. This was mostly because I felt bad about copying so much code from importFromURL, but didn't have the gumption to refactor that and did not explore `loadDocuments`. Thanks.

dstillman:
(Better not to rebase and force-push while we're reviewing. We can always squash it before merging if necessary, but it makes it easier to review with new commits.)

I'm getting a "snapshot is undefined" error running this locally:

```
./test/runtests.sh -f -d 5 -g "web translators should save attachments" translate
```

And there's an error above in the log:

```
zotero(2)(+0000007): Saving attachment failed

zotero(2)(+0000000): Error: 'contentType' not provided

    Error: 'contentType' not provided
        Zotero.Attachments</this.createURLAttachmentFromTemporaryStorageDirectory@chrome://zotero/content/xpcom/attachments.js:544:37
        Zotero.Attachments</this.importFromURL</externalHandlerImport<@chrome://zotero/content/xpcom/attachments.js:472:32
```

Are you seeing this? I'm seeing it on master too, so doesn't seem specific to this commit, but I'm wondering if it's from a recent change. Also doesn't seem to happen in Travis.

I saw this checking another test that is failing in Travis, which presumably is due to the loadDocuments() change:

```
./test/runtests.sh -d 5 -g "web translators should save attachment from browser document" translate
```

```
zotero(3)(+0000006): Zotero.HTTP.loadDocuments: http://127.0.0.1:23119/test/translate/test.html loaded

JavaScript error: chrome://zotero/content/xpcom/http.js, line 1218: Error: Invalid response 404 Not Found for 'http://127.0.0.1:23119/test/translate/test.html'
```

(I didn't look that closely, but this may actually be showing that that was saving a 404 and we just weren't checking that in the test. So, good change!)

fletcherhaz:
Sorry about the force push. I won't do that in the future.

That first test is working on my machine right now. I've had it intermittently fail before and I'm not sure why.

For the second one, in the build I am looking at, that succeeds [https://travis-ci.org/github/zotero/zotero/builds/738414649](url). The one that fails is `web translators should ignore attachments that return error codes` and that fails because I broke that functionality, by not rejecting the promise in `nativeHandlerImport`. I fixed that in the last commit.

dstillman:
Do you not see a 404 in the debug output of "web translators should save attachment from browser document"?

dstillman:
Oh, sorry, I have another copy of Zotero open. I always forget about that. (We should probably run tests on a different port.)

dstillman:
All translate tests are now passing for me.

fletcherhaz:
Ok great!

.gitmodules
```diff
@@ -29,3 +29,6 @@
 [submodule "resource/SingleFileZ"]
 	path = resource/SingleFileZ
 	url = https://github.com/gildas-lormeau/SingleFileZ.git
```
dstillman, line 31: We can remove this now, right?

fletcherhaz: Yes. I didn't know if there was a good or bad way of doing that so left it.

PR #1909, opened 2020-10-28
Update SingleFile for bug fixes

### LINKED ISSUE #1908: SingleFile: Patch setTimeout() → window.setTimeout() and similar after recent update
See daefc9c0aeea473851b9c4be8f6abd690b680f93

No rush (and we probably should wait, because the code could change), but we'll need to do this the next time we upgrade SingleFile.

fletcherhaz:
- Using `sandboxPrototype` properly uses window as prototype
- This commit removes the need for our patch in babel-worker.js
https://github.com/gildas-lormeau/SingleFile/commit/3d0bc4cf9fd23e4fe8f5b684a89c69521c306b60

This fixes #1908 and other similar issues like it.

@dstillman I also turned off `removeFrames` and included an extra commit that properly injects scripts to retrieve frame contents in the client. So frames should be working in both the connector and the client without the 10 second delay we saw before.

Here are two forum posts which both now work with these commits:
https://forums.zotero.org/discussion/85726/error-report-zetero-connector-failed-to-upload-webpage-snapshot
https://forums.zotero.org/discussion/85603/bug-in-zotero-beta-video-iframe-image-not-saved-in-snapshot

### DIFF
--- a/chrome/content/zotero/xpcom/attachments.js
+++ b/chrome/content/zotero/xpcom/attachments.js
@@ -869,6 +869,15 @@ Zotero.Attachments = new function(){
 		if (parentItemID && collections) {
 			throw new Error("parentItemID and parentCollectionIDs cannot both be provided");
 		}
+
+		// If no title was provided, pull it from the document
+		if (!title) {
+			let parser = Components.classes["@mozilla.org/xmlextras/domparser;1"]
+				.createInstance(Components.interfaces.nsIDOMParser);
+			parser.init(null, Services.io.newURI(url));
+			let doc = parser.parseFromString(snapshotContent, 'text/html');
+			title = doc.title;
+		}
 		
 		let tmpDirectory = (await this.createTemporaryStorageDirectory()).path;
 		let destDirectory;
--- a/chrome/content/zotero/xpcom/connector/server_connector.js
+++ b/chrome/content/zotero/xpcom/connector/server_connector.js
@@ -855,17 +855,30 @@ Zotero.Server.Connector.SaveItems.prototype = {
 			cookieSandbox,
 			proxy
 		});
+		// This is a bit tricky. When saving items, the call back`onTopLevelItemsDone` will
+		// return the HTTP request to the connector. Then it may spend some time fetching
+		// PDFs. In the meantime, the connector will create a snapshot and send it along to
+		// the `saveSingleFile` endpoint, which quickly adds the data to the session and
+		// then saves the pending attachments, without removing them (we need them in case
+		// the session switches libraries and we need to save again). So the pending
+		// attachments exist and have already been saved by the time this `saveItems`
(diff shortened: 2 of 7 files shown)


fletcherhaz:
The failing test succeeds on my local machine, but I believe I can replicate by changing this line to `zoomFactor <= 1` here [https://github.com/gildas-lormeau/SingleFile/blob/369c194a945cfd2d442783d5b4f73a2ca5e54f18/lib/single-file/processors/hooks/content/content-hooks-frames-web.js#L143](url). I think it has something to do with the virtual machine that Travis is running.

Either way, I believe this is caused by a bug upstream in SingleFile so I've opened a pull request for it [https://github.com/gildas-lormeau/SingleFile/pull/534](url). So we can wait until that fix lands before merging this pull request.

dstillman:
Is there a performance difference between frames and no frames, on ad-heavy pages?

fletcherhaz:
Did most of my testing with NYT articles and the results are very anecdotal and non-scientific. It seems like a relatively good test case since the content does not change much between reloads and the ads are a relatively significant. This was done by timing this line (https://github.com/zotero/zotero-connectors/blob/master/src/common/inject/inject.jsx#L491)[url]. SingleFile takes up most of that time. In my experience everything else takes about 500-600ms. The reason I did it that way though is that increased size of page increases the time of transmission, etc. so it seemed fair.

Most NYT pages seem to take ~4.8 seconds.

`removeFrames` does not make make much of a difference. At least, the variation is small enough that differences in caching, latency, etc. cover up any noticeable visual difference in the timing. However, this was only loading the page, not scrolling down. If you do that it get's much more interesting. If you patiently wait for everything to load on NYT including images and frames and make it all the way to bottom seems to be about 7.0 - 7.5 seconds with frames and 3.0 - 3.5 seconds without.

With and without frames and not scrolling down the file sizes are about 2.5 to 3.0 MB. Since it is really only loading that first frame ad they are basically the same. If you scroll down and load all the ad frames it seems to be 6.5 MB to 8.5 MB depending on ad sizes.

I can also do a more automated timing of SingleFile in the client with larger numbers of test cases and actual averages, but I'm not sure how accurate those will be because the scrolling seems to be critical unless we have another website we think will be a good test.

dstillman:
And what's the difference in the saved content? Just ads? When you scroll down, is it loading ads that get saved, or other things?

dstillman:
So then I guess the question is, do we want what's mostly going to be ad content? A video still is nice, though as I mention in the linked NYT thread, in that case since there's auto-play it's totally arbitrary which still gets saved.

I'd be inclined to leave `removeFrames` off and maybe add a hidden pref to turn it on. We might even consider exposing a flag in translators, so if there are sites where we know frames are useful we can keep them. (We should wait until we find such a site, though.)

PR #1919, opened 2020-11-23
Automatically limit synced local storage size

fletcherhaz:
- WARNING: This may not properly check if the file is synced before deleting!
- Allows for setting different cache size limits for My Library and Group Libraries
- Currently the process to identify/delete files only is kicked off by a context menu item for libraries in the left pane. It will eventually be started by a change in preference (should be debounced somehow) and when new files are downloaded through automatic syncing.
- Only affects attachment files
- Provides a separate "Manual Cache", which provides Download/Remove options for both individual attachments or entire items
- In preferences dialog there is a View Storage Breakdown, which scans your entire storage folder and identifies cache sizes for automatic and manual cache (This should probably be broken out by library as well).
- I added a temporary size field to the attachment pane. We don't have to keep it (and if we do, I'll change it be nicely formatted), but it was useful for testing.
- If we keep the Manual Cache we will need to add a way to display all those files because inevitably, that cache will grow in size. We have the indexed DB field so just need a UI component for that.
- I'm not entirely convinced on the concept of a manual cache (terrible terminology not withstanding). I think intuitively, I want to be able to just download things ahead of time for when I know I will not have internet access, but once I haven't looked at them for a month and they make their way down the lastAccessed list, I'm not sure I want them treated any differently. I like the idea in principle, but once I start using it, it feels wrong.


dstillman:
Haven't looked at the code yet, but some initial questions/thoughts:

1) In the storage breakdown, "items" seems to actually mean files, right? But which files? It says "52633 items totally 6GB". `find . -type f | wc -l` within 'storage' shows 53384 and  `du -sh` shows 4.1 GB.

2) What counts as "Zotero Data"? Is that files beginning with `.zotero-`?

3) We'll definitely want to break down the breakdown by library, yes.

4) This is counting all files, but "attachments" probably makes more sense as the granularity. The fact that old snapshots have multiple files shouldn't matter.

5) This seems to be including orphaned files (i.e., files not linked to items in the database) in the breakdown. The ability to purge orphaned files in the 'storage' directory is a long-requested feature, but I don't think it belongs with this, since this is about syncing, whereas people can end up with orphaned files for other reasons. Not sure if you're doing anything else with orphaned files now, but deleting them requires different behavior — those files will never exist online, so they won't pass an existence check, and we also don't want to delete them unless they were modified on disk earlier than (some number of days earlier than?) the last sync, since otherwise some special configurations (e.g., where 'storage' is synced with an outside service) could result in new files being deleted before the associated items are synced. They also just won't work when this is broken down by library, since they don't exist in given library (at least until we start breaking out storage on disk by library, which we should've done a decade or so ago). So I think orphaned files should be ignored here and just be handled later via a separate function under Tools → Manage Attachments.

6) If it's not including orphaned files, the breakdown generation can probably be much faster, since it just needs to get all stored files by library and then calculate the size per folder on disk rather than scanning the whole 'storage' folder.

7) This is using only size for cache control, right? Nothing about deleting files not opened in x days? We discussed both previously, but thinking more about this, a TTL actually seems like the more human-friendly metric, and it's also safer — if it's size, we can't guarantee that all the files you just downloaded would even remain available, if you set the cache size low enough or there was one large file, neither of which are things you should have to think about. So I kind of think we should show the space usage in the breakdown but have the cache actually be controlled by number of days. (If we wanted to get fancy, down the line we could even dynamically show how much space would be cleared as you lowered the number of days.)

8) Re: manual caching, you may be right that having files that exist outside of the normal purging mechanism is unnecessarily complicated, particularly if we switch to a more predictable cache-control metric. But is it possible to download files without opening them? Because I'm pretty sure we at least want that. If you're getting on a plane without wifi and want to do all your reading on a topic, you should be able to select all items in a collection or all recently added items or all items that match a search and quickly download them to the cache without opening 30 files. And if they're already downloaded, we should touch lastAccessed.

9) Unfortunately, I think we're going to need to break out all file sync settings by library in the prefs, not just My Library vs. Groups. It's always been something we should do, and I don't think it makes sense to add another setting that applies to all groups together. Breaking it out starts to get pretty complicated, though. Presenting per-library settings would be a lot easier with the Firefox/Chrome-style scrollable prefs that we're probably going to switch to eventually, though we could probably do it for now with a scrollable box. We'd also have to migrate the existing groups prefs into a single pref, similar to `sync.librariesToSync`. In the interest of keeping this PR manageable, maybe we should implement this for My Library only now and break out group file sync settings separately? Unless we'd want a global setting anyway — though would that be for groups, or all libraries? — if we could figure out a non-confusing way to let you switch between a global setting and per-library settings? Probably worth looking for some examples of similar settings panes here.

fletcherhaz:
I focused on files not items in the storage breakdown because I felt that once we started displaying information about how much disk space Zotero was using we should include all the disk space Zotero was using. I thought that once Zotero started providing ways to manage disk usage it should no longer require the user to go to their file browser (i.e. we should be honest about total space Zotero is using, provide a breakdown of how that storage is broken up, and provide options to manage the usage). But you are right, that this does not really belong in the Syncing section. So, for now, I can focus on providing just the attachment storage breakdown.

Since I don't think it's really a breakdown anymore, simply just for each library here is the size on disk, I'll work on different language and display. Perhaps Tools → Manage Storage can allow for discovering orphaned files, viewing total file sizes (including db, full-text content, etc.), and other tools of that nature.

8. Yes. You can already right click and download attachment(s) for both single attachment and top level items. I'll add that for collections as well. I'm going to remove the separate manual cache though. Not sure what to do about touching items since context menu item doesn't show up if the item is already downloaded. I guess we could show it no matter what?

dstillman:
I don't think replacing the functions of a file manager is really necessary — you can always check how much space the data directory is using in the filesystem, and it's not particularly useful to see the size of, say, the database if there's nothing you can do about that. I think the relevant parts are 1) showing how much space the file sync cache is taking up, since that's under your control, and 2) showing how much space orphaned files are taking up, since you can't determine that or fix that from the filesystem.

> Not sure what to do about touching items since context menu item doesn't show up if the item is already downloaded. I guess we could show it no matter what?

Hmm, yeah, I was imagining we'd always show it for multiple items when at least some weren't locally cached, and in that case touching `lastAccessed` makes sense. But for single cached items, or multiple cached items where all files existed locally (up to some reasonable limit of stat calls), we'd want to (and it sounds like already do) show Remove Download.

If we're using a TTL, when does the deleting happen? I guess it would still only happen during syncs, right? So if you're offline, it's not like it's going to delete files that you were counting on having just because they expired. And if you're syncing, hopefully you notice that something you were counting on just expired. And we'd default to something like 30 days anyway, so the odds of your not having even opened something you were counting on in the last month should be pretty low.

fletcherhaz:
These last commits:

1 - 4. Use attachments as the unit, only count how many storage attachments you have, ignore `.zotero-` files.

5 - 6. Ignores orphaned attachments.

7. Uses TTL as the preference (but not does not yet get fancy and show an estimate.

8. Removes special manual cache, but still allows downloading of attachments and items. It does not yet touch lastAccessed if it is already downloaded. I'm mulling over using Cache instead of Download to properly convey that it may disappear. Also, those should only be enabled if we have on-demand preference set for that library.

9. I'll go ahead and start working on this. I'm going to leave both My Library and Groups both working for now since I've been testing with groups and at the end I can turn it off for groups until we get preferences all sorted.

fletcherhaz:
One other thought, if we are only doing a TTL based cache do we want to keep storage sizes in the database? It will provide a faster estimate of how much space will be freed by changing the cache limit, but otherwise doesn't have any benefit for this pull request.

dstillman:
Yeah, probably not worth storing the size. Depending on how fast the stat calls are, we might be able to do a dynamic estimate even without it.

fletcherhaz:
Here is a concept for individual group settings. We can do this in a separate pull request, but I feel it's worth discussing in the context of needing to add another group preference. Do we foresee more group preferences being added? If not, I don't see much value in creating a more complicated pane. I think the best direction to go then would be a split left/right selection/detail component, but that seems overkill for two related settings.

Screen Shot 2021-01-08 at 3 07 52 PM

dstillman:
That's nice and simple, but I think we'll want a bit more:

1) It'd be good to find a way to avoid the "0" thing. "0" meaning unlimited is always a bit of a hack (hence the need for the extra text), and here it's more ambiguous than usual, because you could imagine not wanting to keep something at all beyond the initial download/open.

2) Maybe we should just give a few options in a drop-down — 1 day, 1 week, 1 month, 3 months — instead of a number field? Less control, but maybe quicker to set a bunch one after another? Not sure.

3) It would be nice to hide the TTL option when "as needed" isn't selected.

4) It's possible we'll have more per-group settings in the future.

5) If the stat calls are fast enough, we'll want to show how much disk space you're using per library and how different TTL settings would change that, and we'll need space for that.

So I think it'll need to be some other sort of presentation, and the TTL should be a checkbox with something like "Remove downloaded files after [drop-down]". If unchecked you'd just get the current behavior of keeping forever.

Not sure what the default should be. For existing users with "as needed" selected, I think we have to stick to that being unchecked, but maybe we want to default to something different when switching to "as needed" going forward? An "as needed" mode without any sort of purging never made a ton of sense. (We've also talked about using "as needed" as the default in general, which I believe we're not doing now.)

We also need to communicate what exactly it's purging "after". How do we want to phrase that? It's from download or last open, right? We also need to convey that we're only deleting the local files, since people are going to worry that it means it'd delete the files in the online library.

fletcherhaz:
I think I touched on all the points you mentioned with this version. We can easily hide sections when checkboxes are disabled. and there is room for more settings to be enabled. Language and styling can be tweaked as we go, but if this seems like a good starting point, I can flesh it out more.

If we are going to do this a separate pull request do we want to do it first? And then rebase that into this pull request and add the new setting?

Screen Shot 2021-01-13 at 3 27 52 PM

dstillman:
I think a two-column layout is going to be too wide, and it also prevents you from quickly reviewing your settings across all groups, even just to see which groups have file syncing enabled.

A couple other possibilities:

1) A scrollable list that just has taller sections that are big enough for the necessary controls.

2) A read-only summary of the settings for each group, with a button to modify each group in a separate window.

Not sure which makes more sense, but maybe we can try the first to start.

fletcherhaz:
How is this or did you mean the whole pane should be scrollable? On my 15" laptop this window takes up the entire height. We could make it shorter, but then only one item will be shown. If we add more settings you won't even be able to see one entire group of settings, which doesn't seem right.

Screen Shot 2021-01-14 at 6 21 17 PM

fletcherhaz:
> I think a two-column layout is going to be too wide, and it also prevents you from quickly reviewing your settings across all groups, even just to see which groups have file syncing enabled.

We could put the enable/disable checkbox in the left list or least indicate which are enabled visually to solve the quickly review problem.

The other option here is just put the whole thing in a new dialog and then we have all sorts of space. If we have `Choose Libraries` in a new window maybe we're just trying to hard to fit it in here.

dstillman:
> The other option here is just put the whole thing in a new dialog and then we have all sorts of space. If we have Choose Libraries in a new window maybe we're just trying to hard to fit it in here.

We could do both — keep what we have in production now as a global control for all groups, with a new global TTL setting, but have a "Customize…" button that opened a dialog to override group settings. So we could default to syncing group files in "as needed" mode, but if you wanted to fine-tune it for specific libraries, you could. We'd need some way to showing that the global setting was being overridden in favor of the group-specific settings, and of reverting to the global setting. (Reverting would probably happen in the dialog.)

fletcherhaz:
What do we think about this?

Screen Shot 2021-01-15 at 1 34 17 PM
Screen Shot 2021-01-15 at 1 34 09 PM

dstillman:
Yeah, I think this is the general direction. A couple things:

1) White background for the scrolly section
2) I think maybe more of a block-based approach than the groupbox-caption approach.

Basic principle I'm going by is that preferences that look like chrome shouldn't be dynamic — they should be in sections that look like they're dynamically populated.

Also, this whole window should be HTML, not XUL. (I'm not sure how well we're set up to display dialog buttons in HTML windows, but we'll need it one way or another.)

Also, no asterisk/italics. Maybe just take out that explanatory line and just go with "Remove downloaded files if not opened for".

fletcherhaz:
I'm not 100% sure I understand what you mean by "block-based approach" but is this the idea?

Screen Shot 2021-01-15 at 3 24 01 PM

dstillman:
Sorry, I just mean using a border to separate the sections rather than the native-looking groupbox + caption approach. So just imagine what you have above but with no margins and a white background, with a gray border between the blocks.

I can help tweak the styling later.

Maybe group name in the top left and then a "Use custom settings" on the right-hand side that would show the additional controls.

fletcherhaz:
Thank you. That makes a lot more sense. I understand the difference you are trying to convey between dynamic and static. I'll implement all the functionality and submit it in a separate pull request for more details, linking to this one.

Screen Shot 2021-01-15 at 3 48 19 PM

dstillman:
Sorry, I left out a word — rather than two separate buttons, there can just be a "Use custom settings" <strong>checkbox</strong> in that position, which would just show/hide the additional controls.

chrome/content/zotero/xpcom/storage/storageCache.js
```diff
@@ -0,0 +1,319 @@
...
+	 * Identify items with files currently downloaded that we should delete to limit cache
+	 *
+	 * Called when the limit cache preference is turned on or when a new item is downloaded or
+	 * added.
+	 *
+	 * @param libraryID {integer} - Library ID to search for items to free
+	 *
+	 * @returns {Promise<void>}
+	 */
+	identifyItemsToFreeForLibrary: async function (libraryID) {
```
dstillman, line 159: `identify` is a little misleading here — this is actually doing the deletion, right?

For testing purposes and general cleanliness, maybe it makes sense to split this into two functions, with one that just returns a specified number of items with files to delete.

chrome/content/zotero/xpcom/storage/storageCache.js
```diff
@@ -0,0 +1,319 @@
...
+			}
+			
+			sql += " ORDER BY lastAccessed DESC, dateModified DESC LIMIT ?,1000";
+			let params = [
+				libraryID,
+				Zotero.Attachments.LINK_MODE_IMPORTED_FILE,
+				Zotero.Attachments.LINK_MODE_IMPORTED_URL,
+				Zotero.Sync.Storage.Local.SYNC_STATE_IN_SYNC,
+				cacheTime,
+				offset
```
dstillman, line 212: Is using an offset correct here? If you're changing syncState when deleting the files, won't they not be matched in subsequent queries, such that an offset would skip thousands of unprocessed files?

dstillman (changes requested):

chrome/content/zotero/xpcom/file.js
```diff
@@ -603,7 +603,33 @@ Zotero.File = new function(){
...
+	this.getFileSize = async function (path) {
+		return OS.File.stat(path)
+			.then(
+				function (info) {
+					return info.size;
+				},
+				function (e) {
+					// Can happen if there's a symlink to a missing file
+					if (e instanceof OS.File.Error && e.becauseNoSuchFile) {
+						return 0;
```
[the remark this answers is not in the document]
fletcherhaz: I was trying to abstract a bit more and save copying this try/catch everywhere we use this function. Since Zotero never knows whether a file exists on disk without checking, it made sense to me to abstract out that one error. It just seemed like a nicer interface to deal with, having the function handle that specific case. Otherwise, I envisioned writing that same try/catch every we use this function because you probably care about that exception, but none of the other (can't read disk, etc.) exceptions that can get handled a higher level since they are more serious.

But, that's an assumption which can always be wrong It's only one line without the exception handling so I'll just remove the whole function and each calling function can decide what it needs.

fletcherhaz:
To summarize for myself here is what I know is left before another review:

- Implement and test ZFS `checkFileExists`
- Implement and test WebDAV `checkFileExists`
- Clear up several TODOs covering my uncertainty
- Decide how to start cache clearing when preferences are changed
- Decide when cache clearing will be kicked off at any other time (sync, manually, etc.)
- Styling on Windows and Linux

@dstillman What are your thoughts on when we should run the cache clearing process? Do we just run it at the end of the sync? Do we run it when you close the preferences window?

dstillman:
> What are your thoughts on when we should run the cache clearing process? Do we just run it at the end of the sync? Do we run it when you close the preferences window?

I was wondering about that (and wondering what it was now — because I was afraid to start this up, lest it start deleting things).

Can you remind me what we need to do to determine whether anything needs to be cleared?

fletcherhaz:
> I was wondering about that (and wondering what it was doing now — because I was afraid to start this up, lest it start deleting things — though I guess it won't do anything unless I turn these settings on?).

Currently, this is the only entry point: https://github.com/zotero/zotero/pull/1919/files#diff-5cc5a652c4e7d6255cc45736617bbda3ed8b6c4e1ebbbde67efc375ca5eb72f2R102

The only time that is called is via the collection context menu here: https://github.com/zotero/zotero/pull/1919/files#diff-81a2ccc004bf363ce984ac3618a65eeeda6c0a5f695ae89ed7d30bedf53fe29fR2358

I added the context menu option to make it easy to test on a group instead My Library. We can remove that if we don't want it in production. There are also context menu options on individual items/attachments to cache/evict downloads.

> Can you remind me what we need to do to determine whether anything needs to be cleared?

Currently, items are selected from the database here https://github.com/zotero/zotero/pull/1919/files#diff-5cc5a652c4e7d6255cc45736617bbda3ed8b6c4e1ebbbde67efc375ca5eb72f2R149

They are selected if they: 
- Are in the right library
- Are one of `Zotero.Attachments.LINK_MODE_IMPORTED_FILE` or `Zotero.Attachments.LINK_MODE_IMPORTED_URL`
- Have an attachment sync state of `Zotero.Sync.Storage.Local.SYNC_STATE_IN_SYNC`
- Have a `lastAccessed` timestamp older than the preference
- If there is no `lastAccessed` parameter then check if `dateModified` is older than preference

Once the items are selected they are passed to the delete function here: https://github.com/zotero/zotero/pull/1919/files#diff-5cc5a652c4e7d6255cc45736617bbda3ed8b6c4e1ebbbde67efc375ca5eb72f2R247

They are deleted if they:
- Are an imported attachment
- Exist on the server (Return 2000 and match both `attachmentModifiedTime` and `attachmentHash`)
  - WebDAV: https://github.com/zotero/zotero/pull/1919/files#diff-2f104a190e72fb83c8510983b05c6186d93f6877783cc85d1e2b0f6e83038559R603
  - ZFS: https://github.com/zotero/zotero/pull/1919/files#diff-dd2edaacb90038c62241180d7e1a016e24b2f9e2dd3a7b6797c777d448abc952R331

Tests for those four functions are here:
- Select Expired Items: https://github.com/zotero/zotero/pull/1919/files#diff-660fe533862b61c6ba2c02401fda38d4ee536abf8e922648a383dd749d02a444R251
- Deletion: https://github.com/zotero/zotero/pull/1919/files#diff-660fe533862b61c6ba2c02401fda38d4ee536abf8e922648a383dd749d02a444R134
- WebDAV: https://github.com/zotero/zotero/pull/1919/files#diff-ff10bde2f02b260ab505ca649de73d260b7b0fc698180edeff71eedcf3c49244R1223
- ZFS: https://github.com/zotero/zotero/pull/1919/files#diff-c0c6f4fbbbcc4a6766be0fef0e6247d92f6cb0481e0ef7df18a4af11b085ca2eR1050

One more note is that `deleteItemFiles` can be called from the individual `removeAttachmentFilesForItems`, which is called via the context menu and so we may need to add more checking for that route.

test/tests/zfsTest.js
```diff
@@ -1046,4 +1046,107 @@ describe("Zotero.Sync.Storage.Mode.ZFS", function () {
...
+				}
+			});
+
+			let result = await zfs.checkFileExists({
+				name: item.libraryKey
+			});
+			assert.isFalse(result);
+		});
+
+		it('should check for attachment hash', async function () {
```
dstillman, line 1094: Check what?

test/tests/zfsTest.js
```diff
@@ -1046,4 +1046,107 @@ describe("Zotero.Sync.Storage.Mode.ZFS", function () {
...
+						}
+					);
+				}
+			});
+
+			let result = await zfs.checkFileExists({ name: item.libraryKey });
+			assert.isFalse(result);
+		});
+
+		it('should check for attachment modification time', async function () {
```
dstillman, line 1113: Check what?

dstillman (commented):
A few more notes — more to come

dstillman:
OK, and yeah, so if most of the work here is done by a DB query, with stat()s and HTTP requests only as confirmation for files that are older, then just kicking this off after a sync for a given library probably makes sense. But only when there's a file sync for a given library? But only once every x hours, if you haven't changed the prefs?

Also, does lastAccessed currently get updated if you open and modify a file outside of Zotero (but not via the 'open' trigger), and then sync and it uploads? That should count as an access.

dstillman:
https://forums.zotero.org/discussion/87458/remove-local-files-without-affecting-online-libraries

I guess this is another use case — disabling file sync altogether and clearing local files. If we did this, it seems like we'd need to prompt when you actually unchecked — we don't want to delete files without checking that they exist online, but doing it once file syncing is disabled is a little weird (particularly in the case of WebDAV, where the entered details may no longer even be accurate). Showing a prompt when unchecking a checkbox is a little awkward, but maybe it's OK: "Do you want to delete local files in this library that exist online?\n\nYou can download the files later by re-enabling file syncing."

chrome/content/zotero/xpcom/storage/storageLocal.js
```diff
@@ -563,6 +568,32 @@ Zotero.Sync.Storage.Local = {
...
+	 *
+	 * @param {Zotero.Item[]} items
+	 * @return {Promise}
+	 */
+	_updateLastAccessed: async function (items) {
+		let timestamp = Date.now();
+
+		await Zotero.Utilities.Internal.forEachChunkAsync(
+			items,
+			1000,
```
dstillman, line 584: This should actually be set to `Zotero.DB.MAX_BOUND_PARAMETERS` (which is 999), as should `updateSyncStates()`.

fletcherhaz:
> Also, does lastAccessed currently get updated if you open and modify a file outside of Zotero (but not via the 'open' trigger), and then sync and it uploads? That should count as an access.

For documentation purposes, `lastAccessed` is currently updated in three places:
- When you click `Download Attachment` if you already have the file (https://github.com/zotero/zotero/pull/1919/files#diff-5cc5a652c4e7d6255cc45736617bbda3ed8b6c4e1ebbbde67efc375ca5eb72f2R78)
- Whenever the file is downloaded (https://github.com/zotero/zotero/pull/1919/files#diff-b4e2f596d893d261f4cb72684e23b7be80d8ffe1f59dcc212a7e7743c8357c53R689)
- When `StorageFileOpen` notification goes off (https://github.com/zotero/zotero/pull/1919/files#diff-b4e2f596d893d261f4cb72684e23b7be80d8ffe1f59dcc212a7e7743c8357c53R689)

And now (after a little back and forth) update `lastAccessed` when we upload a file to the server (ZFS or WebDAV):  https://github.com/zotero/zotero/pull/1919/commits/88f47fe9a91c069875fa780a14b26044063ff512

fletcherhaz:
> https://forums.zotero.org/discussion/87458/remove-local-files-without-affecting-online-libraries
> 
> I guess this is another use case — disabling file sync altogether and clearing local files. If we did this, it seems like we'd need to prompt when you actually unchecked — we don't want to delete files without checking that they exist online, but doing it once file syncing is disabled is a little weird (particularly in the case of WebDAV, where the entered details may no longer even be accurate). Showing a prompt when unchecking a checkbox is a little awkward, but maybe it's OK: "Do you want to delete local files in this library that exist online?\n\nYou can download the files later by re-enabling file syncing."

Another possibility would be to have a context menu option similar to kicking off the cache clearing process (as implemented right now). So you would disable the file syncing for a group, close preferences, and then right click on a group and click `Remove All Synced Attachment Files`. It would use the same codebase, just pass in an override parameter to not compare to `lastAccessed`.

dstillman:
> Another possibility would be to have a context menu option similar to kicking off the cache clearing process (as implemented right now). So you would disable the file syncing for a group, close preferences, and then right click on a group and click Remove All Synced Attachment Files.

I think that's a bit too out of the way. It'd be better to keep this closer to the settings, one way or another.

If we didn't do it when unchecking, another option would be showing the size breakdown for each library in each section and having a Clear button next to it. Or we could just have Clear buttons in the storage breakdown dialog.

On that front: I haven't yet tried a recent version of this, but now that we have a separate dialog for these settings, is there any reason for the storage breakdown to be a separate window instead of just in here? I guess because My Library isn't in here, but it seems much more useful to see the breakdown in the context of settings where you can control it and clear it.

fletcherhaz:
> I think that's a bit too out of the way. It'd be better to keep this closer to the settings, one way or another.

Yea. That makes sense. We can definitely add a clear button to the Storage Breakdown or merge the dialogs.  Here is the current state of all three.

For my library, which has a good number of snapshots it takes just over 20 seconds to scan 4.4 GB. Not sure if that affects the choice to combine them or not.

Screen Shot 2021-02-04 at 9 01 10 AM
Screen Shot 2021-02-04 at 9 01 27 AM
Screen Shot 2021-02-04 at 9 14 24 AM

fletcherhaz:
This is now hooked into the file sync: https://github.com/zotero/zotero/pull/1919/files#diff-1cdb47456e54d5fec2e26966e4782a1ceb688487fe80effda090c883a9932b50R291

It runs after a file sync for a library completes, but never more than once per day. A restart or preference change will reset the clock.

chrome/content/zotero/preferences/storageBreakdown.jsx
```diff
@@ -88,7 +88,10 @@ const StorageBreakdown = () => {
 
 						await Zotero.File.iterateDirectory(
 							Zotero.Attachments.getStorageDirectory(item).path,
-							getFileSize
+							(attachmentFile) => {
+								getFileSize(attachmentFile);
```
dstillman, line 92: Instead of ignoring the promise, probably better to add to an array and use `Promise.all()` after?

chrome/content/zotero/preferences/storageBreakdown.jsx
```diff
@@ -0,0 +1,182 @@
...
+							let size = (await OS.File.stat(attachmentFile.path)).size;
+							libraries[index].size += size;
+						}
+						catch (e) {
+							if (e instanceof OS.File.Error && e.becauseNoSuchFile) {
+								// File may or may not exist on disk, but we
+								// don't care so swallow this error
+							}
+							else {
+								throw e;
```
dstillman, line 71: Instead of throwing, maybe just `Zotero.logError(e)` and ignore? One way or another, we should account for the possibility of permissions or I/O errors and not break the dialog if those occur.

chrome/content/zotero/preferences/storageBreakdown.jsx
```diff
@@ -0,0 +1,182 @@
...
+			});
+
+			setPartitions(libraries.map(library => library));
+			let update = 0;
+
+			await Zotero.Promise.all(libraries.map(async (partition, index) => {
+				let items = await Zotero.Items.getAll(partition.libraryID);
+
+				const getFileSize = async (attachmentFile) => {
+					if (!attachmentFile.name.startsWith('.')) {
```
dstillman, line 60: Reverse this and return, just to avoid the extra indentation below

dstillman:
I think we kind of have to combine the dialogs — it doesn't really make sense to show the usage in a different place from where you can configure it, and whether you should be able to clear also depends on the settings. (You shouldn't be able to clear if it's set to "at sync time".)

But looking at this dialog, I guess I'm now wondering whether it actually makes sense to configure different TTLs for different libraries. Choosing whether to sync files for a group certainly makes sense to me, and choosing the download mode also makes sense. But what's a case where you would actually want to configure the TTL differently? And if they're all different, how could you have any sort of mental model for when files would exist locally based on your recent usage and you could count on their being offline? I guess if it was a group that was really big, or that you knew you worked on less frequently, you could decrease the TTL for that one, but…maybe you should just choose a global TTL that balances that out based on your disk space, and/or just re-"download" files in libraries you know you want offline before getting on a plane?

Removing it would free up some space that we could use for displaying the storage usage here and help to keep this dialog from getting too out of hand.

If we do hide, we could keep the functionality in place (making sure it falls back to the global TTL), and if there was somehow a lot of demand for that level of configurability we could reconsider.

fletcherhaz:
I think the clearing/at sync time issue definitely means we need to combine. Some thoughts that may or may not be coherent:

- If we remove custom TTL then we need to always show the global TTL settings (or at least show them whenever a group has custom settings enabled that may require a global TTL setting). That seems weird.
- Part of the issue here is that `Download files as needed` without a TTL doesn't make much sense once you have a TTL option. So, we could commit to `as needed` implying removing files and not add the additional complexity of TTL enable/disable.
- We probably need some feedback for the clearing action. If you click a button that says `Remove All Synced Attachment Files` and then it just silently disappears into the background, that seems unfriendly to me. Maybe we can just disable the button and change the text while it is running? Is that breaking any user design guidelines?

All that being said, here is the concept I came up with.
Screen Shot 2021-02-06 at 9 26 18 AM

dstillman:
Oh, right, the problem with removing group TTL is that there's no guarantee that "as needed" is set globally, and if it's not, showing the global TTL setting is confusing, but if there's no global TTL then you can't remove the per-group TTL…

We could try to separate/rephrase the global TTL setting so that it's clear it applies whenever "as needed" is used. I guess one question is whether we even need it in this window. Arguably it could be truly global — i.e., one setting for all libraries, including My Library. Is there any reason you'd really care to have different TTLs for My Library and any groups?

> Maybe we can just disable the button and change the text while it is running? Is that breaking any user design guidelines?

No, that's fine. We do that, with or without changing the text, in various places (e.g., updating translators and styles from the Advanced prefs).

But I think a better model would be a browser cache, where it displays the amount used with a "Clear" button to the right. So I would move the usage down to the bottom of the section and put a "Clear…" button across from it, with a confirmation prompt — something like "Local files in “[group name]” that have been synced to the online library will be removed from this computer." with `general.continue` for the main button. Since it's not actually deleting any data permanently, the confirmation prompt would mainly be to explain what you're doing and set your mind at ease. Not sure if that's totally necessary, but deleting a few gigs of local files feels like something that probably merits a confirmation prompt…

fletcherhaz:
> We could try to separate/rephrase the global TTL setting so that it's clear it applies whenever "as needed" is used. I guess one question is whether we even need it in this window. Arguably it could be truly global — i.e., one setting for all libraries, including My Library. Is there any reason you'd really care to have different TTLs for My Library and any groups?

From the few forum posts I've read, it seems like mostly people want to a) save storage space locally and b) turn off file syncing for certain groups. Neither of those things seems to really depend on specific TTL settings so a global setting seems ok.

How do we feel about this? If this is the direction we like, I think I can run with this for a bit.
Screen Shot 2021-02-06 at 10 59 55 AM

Screen Shot 2021-02-06 at 10 49 12 AM

dstillman:
Yeah, I think that's the right direction. And I think we can probably drop the "and remove after 16 days". I don't think we need to remind people what the global setting is every time that's selected for a group.

fletcherhaz:
This last commit updates the UI to the previous images and the functionality is fully implemented (except for the `Clear` dialog, which I just remembered).

Two things, which were available before, but are no longer:
- Viewing how much space attachment files take up for `My Library`
- Clearing `My Library`

dstillman:
> I can put those below the download setting for My Library. The only drawback is that if you have a large library whenever you open that preference pane you will kick off a file scan, which probably won't happen that often.

Yeah, that's what I figured we'd need to do, though I don't know if we can do it when the Sync pane is opened and not the preferences — the way the Mozilla preferences are rendered it's possible that's a little tricky, but I don't remember. Hopefully we can.

I think we actually want to show the cumulative total for groups, too, so you can see it all without going into Customize (which isn't clearly where you'd expect to need to go to see this). And we should probably also have a global Clear button for groups. If you're using "at sync time" for a library, you see the usage but just don't get the Clear button, right? So a global Clear would appear for groups if any libraries were using "as needed", and clicking it would decrease the usage by those libraries but leave "at sync time" libraries counted?

And then for caching, ideally we would cache the group calculations in memory such that when you opened the customize dialog it wouldn't need to recalculate. And then probably we want to cache all of this for a few minutes so if you repeatedly open the sync prefs it doesn't keep doing all this disk I/O.

Actually, it occurs to me — we stat all primary attachment files during the first sync of the session, on auto-sync every 3 or 4 hours, and on every manual sync. There's no reason we couldn't track the sizes at that time. For old snapshots we'd have to check other files, but we should have recent sizes for all PDFs and single-file snapshots. Maybe it's not worth trying to reuse those counts — would have to keep them updated, track item and file deletions, etc. — but it's at least a possibility down the line if this becomes a performance problem.

fletcherhaz:
This now includes attachment counts/sizes for My Library and Groups along with Clear buttons for both. They are cached for one hour and updated if you clear a library.

> Yeah, that's what I figured we'd need to do, though I don't know if we can do it when the Sync pane is opened and not the preferences — the way the Mozilla preferences are rendered it's possible that's a little tricky, but I don't remember. Hopefully we can.

We do this on individual pane load so it shouldn't be too bad.

Currently, we only allow clearing if you have the TTL setting enabled. Is that something we want to keep?

Screen Shot 2021-02-11 at 4 34 11 PM
Screen Shot 2021-02-11 at 4 49 56 PM

chrome/content/zotero/xpcom/storage/zfs.js
```diff
@@ -818,6 +883,10 @@ Zotero.Sync.Storage.Mode.ZFS.prototype = {
 			item.attachmentSyncedModificationTime = params.mtime;
 			item.attachmentSyncedHash = params.md5;
 			item.attachmentSyncState = "in_sync";
+			if (!item.attachmentLastAccessed
+				|| params.mtime > item.attachmentLastAccessed) {
+				item.attachmentLastAccessed = params.mtime;
```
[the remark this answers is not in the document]
dstillman: Sorry, you're totally right — these can't be the same property after all. I guess that means the one in #2854 should be renamed to `attachmentLastRead` (which will cover opening and interacting with but not syncing) and can again be limited to My Library.

chrome/content/zotero/xpcom/storage/zfs.js
```diff
@@ -604,6 +665,10 @@ Zotero.Sync.Storage.Mode.ZFS.prototype = {
 				item.attachmentSyncedModificationTime = fileModTime;
 				item.attachmentSyncedHash = fileHash;
 				item.attachmentSyncState = "in_sync";
+				if (!item.attachmentLastAccessed
+					&& fileModTime > item.attachmentLastAccessed) {
```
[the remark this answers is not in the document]
dstillman: Yeah, this should be `||` (or I suppose the first condition could just be omitted).

resource/schema/userdata.sql
```diff
@@ -105,6 +105,7 @@ CREATE TABLE itemAttachments (
     storageModTime INT,
     storageHash TEXT,
     lastProcessedModificationTime INT,
+    lastAccessed INT,
```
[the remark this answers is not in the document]
dstillman: I guess `lastAccessed` and (now) `lastRead` might as well both be stored as seconds, which is what we're already using for `lastProcessedModificationTime`. No reason to use 8601 string or milliseconds. (We use 8601 for Date Added/Modified, but those are fairly regularly looked at by human eyes when debugging sync issues.)

chrome/content/zotero/xpcom/storage/zfs.js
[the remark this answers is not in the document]
dstillman: I was thinking of "accessed" in the sense of filesystem "access time", which seems like the closest analogue to this. "Used" is OK but is a little…user-y?

dstillman:
When we return to this, we should probably also make sure there's a way to keep files downloaded:

https://www.macrumors.com/2024/06/26/icloud-keep-downloaded-option-ios-18-macos-sequoia/

PR #1924, opened 2020-12-10
Inform users of which identifiers failed in add by identifier.

### LINKED ISSUE #1911: Show alert with identifiers that couldn't be resolved in Add Item by Identifier
https://github.com/zotero/zotero/issues/1882 was a bug (and I'm not sure what happens now after a translator fails), but we're still just silently ignoring identifiers that don't resolve. Instead,

> we should show an error dialog at the end with a list of failed ids, truncated at a certain number of lines.
>
> Alternatively, we could display the progress dialog and show imported title or error for each row, but that doesn't provide a great way to copy ids that failed.

fletcherhaz:
Closes #1911

### DIFF
--- a/chrome/content/zotero/lookup.js
+++ b/chrome/content/zotero/lookup.js
@@ -99,34 +99,30 @@ var Zotero_Lookup = new function () {
 			}
 			// Continue with other ids on failure
 			catch (e) {
-				let service = Object.keys(identifier)[0];
-				failedIDs.push(service + ': ' + identifier[service]);
+				failedIDs.push(identifier[Object.keys(identifier)[0]]);
 				Zotero.logError(e);
 			}
 		}));
 
 		toggleProgress(false);
-		if (!newItems) {
-			if (identifiers.length === 1) {
-				Zotero.alert(
-					window,
-					Zotero.getString("lookup.failure.title"),
-					Zotero.getString("lookup.failure.description")
-				);
-			}
-			else {
-				Zotero.alert(
-					window,
-					Zotero.getString("lookup.failure.title"),
-					Zotero.getString("lookup.failureMultiple.description", failedIDs.join('\n'))
-				);
-			}
-		}
-		else if (failedIDs) {
+		if (!newItems && identifiers.length === 1) {
 			Zotero.alert(
 				window,
 				Zotero.getString("lookup.failure.title"),
-				Zotero.getString("lookup.failureMultiple.description", failedIDs.join('\n'))
+				Zotero.getString("lookup.failure.description")
+			);
+		}
+		else if (failedIDs.length) {
+			let errors = '\n\n' + failedIDs.slice(0, 10).join('\n');
+			if (failedIDs.length > 10) {
+				errors += '\n...';
+			}
+			Zotero.alert(
+				window,
+				Zotero.getString("lookup.failure.title"),
+				Zotero.getString("lookup.failureMultiple.description")
+					+ errors + '\n\n'
+					+ Zotero.getString("lookup.failureMultiple.prompt")
 			);
 		}
 
--- a/chrome/content/zotero/lookup.js
+++ b/chrome/content/zotero/lookup.js
@@ -79,6 +79,7 @@ var Zotero_Lookup = new function () {
 		}
 
 		let newItems = false;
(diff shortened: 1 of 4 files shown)


chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -1148,6 +1148,7 @@ lookup.failure.title				= Lookup Failed
 lookup.failure.description			= Zotero could not find a record for the specified identifier. Please verify the identifier and try again.
 lookup.failureToID.description		= Zotero could not find any identifiers in your input. Please verify your input and try again.
 lookup.failureTooMany.description	= Too many identifiers. Please enter one identifier and try again.
+lookup.failureMultiple.description	= Zotero could not find a record for the following identifier(s):\n\n%s\n\nPlease verify the identifier and try again.
```
dstillman, line 1151: It's best to avoid newlines in messages, which are more likely to mangled during localization. Better to split into separate strings and hard-code the newlines in the code.

Also, we can just do plural, no "(s)".

And the second sentence should be plural.

chrome/content/zotero/lookup.js
```diff
@@ -98,19 +99,36 @@ var Zotero_Lookup = new function () {
 			}
 			// Continue with other ids on failure
 			catch (e) {
+				let service = Object.keys(identifier)[0];
+				failedIDs.push(service + ': ' + identifier[service]);
```
dstillman, line 103: Technically this isn't localized — not all languages use a colon in this way. But it's probably fine.

dstillman (commented):
What if there are dozens of identifiers? The dialog could theoretically run off the screen. Do we want to limit the number of lines displayed? Use a custom dialog that scrolls? Assume this won't happen much and just display the prompt, which would at least be copyable, even if unpleasant?

chrome/content/zotero/lookup.js
fletcherhaz: Ok. I wasn't really sure what to do here because there is no easy way to actually display what the user typed in and relate that to the parsed identifier. This would in theory only be DOI, ISBN, or arXiv so the colon seemed ok.

fletcherhaz:
@dstillman Can we clarify what we want the error dialog to show? In particular, currently we do not allow multiple types of identifiers in the same input. Using your example:
```10.1103/abcd.95.052301 98712341515 foobar```

`extractIdentifiers` pulls out the first string as a DOI, but then discards the rest? In this example we want to show this:

```
98712341515
foobar
```
If so should we clarify in the error message that we are extracting only DOIs since the next two could be valid identifiers of something else? If we return the offsets of the valid IDs then we are just left with strings? How do we split those up? For DOI we are using spaces, but for ISBNs that seems a bit harder. Finally, if I paste in this:
```
arXiv: 1608.12345
```
I don't really want to be told that `arXiv: ` was not a valid identifier, right?

I think the best solution here is not to worry about showing errors in the parsing. We should only show parsed identifiers that do not resolve. That may have been what you were saying all along, I just didn't understand, but that is what the code does as of this comment. Furthermore, if we have not had problems with multiple types of identifiers in the same input before, then that does not seem to be a problem we need to solve.

fletcherhaz:
> Is that not just a bug/oversight?

I think it was intentionally done. For example, the ISBN modifies the input string to remove hyphens, dashes, and eventually spaces in an attempt to correct spacing errors and different formats. That seems like a very user friendly feature. However, it becomes hard to remove things from the original string one your indexes change and the code/logic becomes more convoluted and error prone.

This seems to be a judgement call. It does lead to some unusual behavior as you showed with the PMID. However, I think given the simplicity and maintainability of the code and the lack of complaints? the current logic is sound enough. An alternative solution to alleviate some of the worry would be to run the extraction again, skipping whatever type of identifier was found the first time, and if we found a second type of identifier display an error saying we only accept one type of identifier at a time.

fletcherhaz:
This latest version extract all types of identifiers and only shows errors for parsed identifiers that do not resolve.

rbpasker:
this problem is plaguing me -- I add bunches of items by ISBN using the add wand tool, and some of them fail silently.

do the maintainers intend to rebase and merge, or does the behavior have a preferred alternative implementation?

PR #1930, opened 2020-12-15
Use Ace editor for Run JavaScript

fletcherhaz:
Refactor the Ace codebase out of the scaffold directory and use it for the `Run JavaScript` window.


fletcherhaz:
@dstillman I did this a three commits since I felt it nicely preserved the changes. Happy to rebase into one or two depending on what you think , especially since the first commit breaks everything so you don't want to exist at that version. Not sure what your philosophy on this is.

dstillman:
Yeah, I think two commits would be appropriate — one that moves it out of the Scaffold dir and updates Scaffold to work with the moved version, and another that adds the editor to the Run JS window.

dstillman:
Is something missing? I get this in the Run JS window:

> The file chrome://ace/content/ace.html cannot be found. Please check the location and try again.

This doesn't seem to have a change to chrome.manifest that would be necessary for that URL to work.

But for an unmodified third-party library like this, generally we'd want to put it in resource:// anyway. Is that possible here?

I also wonder if there's an NPM build that we want to use instead, with the usual symlinking from within resource/. Not necessary, but it's nice to be able to just update package.json for new versions of things.

fletcherhaz:
My apologies. I'm not sure how this worked other than I didn't actually rebuild Zotero when testing. I switched to the NPM Package inside resource, but still left the ace editor in its own chrome (with the appropriate `chrome.manifest` entry).

dstillman:
> but still left the ace editor in its own chrome

Was there are reason for that? Can we not load it all from `resource://`? (This sort of Mozilla stuff is more likely to go away, so best to avoid when possible.)

fletcherhaz:
> > but still left the ace editor in its own chrome
> 
> Was there are reason for that? Can we not load it all from `resource://`? (This sort of Mozilla stuff is more likely to go away, so best to avoid when possible.)

`ace.html` and `aceWrapper.js` are Zotero custom wrappers (not part of the Ace editor) so I didn't know where to put them (I figured `resource` was not the place). We can leave them in the scaffold or move them to Zotero. What do you think is best?

dstillman:
> ace.html and aceWrapper.js are Zotero custom wrappers (not part of the Ace editor) so I didn't know where to put them (I figured resource was not the place)

I assume they can just stay where they are in `chrome/content/ace/` and be referenced via `chrome://zotero/content/ace/`. `aceWrapper.js` is in the same directory as `ace.html`, so that probably doesn't need a prefix at all.

There might be more we want to do to load the Ace code unprivileged, but we can deal with that separately.

fletcherhaz:
> I assume they can just stay where they are in `chrome/content/ace/` and be referenced via `chrome://zotero/content/ace/`. `aceWrapper.js` is in the same directory as `ace.html`, so that probably doesn't need a prefix at all.

Does not this line [https://github.com/zotero/zotero/blob/master/chrome.manifest#L1](url) mean that the `ace` directory would have to be under `chrome/content/**zotero**/ace`? Unless I misunderstand how this works. I just tried this and it resulted in a `File Not Found` error.

Instead I moved it under `chrome/content/zotero/ace` and reference it as you suggested above in both `runJS.html` and `scaffold.xul`.

PR #1937, opened 2020-12-26
Fix double saving snapshots

fletcherhaz:
https://forums.zotero.org/discussion/86796/duplicated-snapshots

I was able to replicate it by adding a 5 second delay here:
https://github.com/zotero/zotero/blob/a72ae1481631dcefe603ef9ff3a1a9f7f5c747d8/chrome/content/zotero/xpcom/translation/translate_item.js#L196

This was caused from a race condition and (a72ae14) did not fully
solve the problem. Now this is tested and fixed.

### DIFF
--- a/chrome/content/zotero/xpcom/connector/server_connector.js
+++ b/chrome/content/zotero/xpcom/connector/server_connector.js
@@ -853,7 +853,7 @@ Zotero.Server.Connector.SaveItems.prototype = {
 			cookieSandbox,
 			proxy
 		});
-		// This is a bit tricky. When saving items, the call back `onTopLevelItemsDone` will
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/xpcom/connector/server_connector.js
```diff
@@ -855,30 +853,35 @@ Zotero.Server.Connector.SaveItems.prototype = {
 			cookieSandbox,
 			proxy
 		});
-		// This is a bit tricky. When saving items, the call back`onTopLevelItemsDone` will
+		// This is a bit tricky. When saving items, the call back `onTopLevelItemsDone` will
```
dstillman, line 856: "call back" → "callback"

PR #1939, opened 2020-12-28
Fix Input component doesn't work if autoComplete=false

diegodlh:
The Input component doesn't work if its autoComplete prop is set to false, as described here: https://groups.google.com/g/zotero-dev/c/vKdxvZJL8UQ/m/VLqu-3VGBQAJ
The commit proposed seems to solve the issue.

### DIFF
--- a/chrome/content/zotero/components/form/input.jsx
+++ b/chrome/content/zotero/components/form/input.jsx
@@ -69,7 +69,7 @@ class Input extends React.PureComponent {
 	}
 
 	handleChange({ target }, { newValue } = {}) {
-		var newValue = newValue || target.value;
+		newValue = newValue || target.value;
 		this.setState({
 			value: newValue,
 		});
--- a/chrome/content/zotero/components/form/input.jsx
+++ b/chrome/content/zotero/components/form/input.jsx
@@ -69,7 +69,7 @@ class Input extends React.PureComponent {
 	}
 
 	handleChange({ target }, options) {
-		var newValue = options.newValue || target.value;
+		var newValue = options && options.newValue || target.value;
 		this.setState({
 			value: newValue,
 		});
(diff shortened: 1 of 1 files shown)


chrome/content/zotero/components/form/input.jsx
```diff
@@ -69,7 +69,7 @@ class Input extends React.PureComponent {
 	}
 
 	handleChange({ target }, options) {
-		var newValue = options.newValue || target.value;
+		var newValue = options && options.newValue || target.value;
```
dstillman, line 72: How about just `{ newValue }` for the second argument instead, and check for `newValue` directly?

diegodlh: Hi, Dan. Thanks for your reply. Do you mean this?
```
handleChange({ target }, { newValue }) {
...
```

I tried it, but it fails with `TypeError: (destructured parameter) is undefined` when `handleChange` is called without a second parameter.

dstillman: Sorry, try `{ newValue } = {}`

diegodlh: Oh, I see. Like a default parameter. It works, yes. Just posted it as a new commit.

chrome/content/zotero/components/form/input.jsx
```diff
@@ -68,8 +68,8 @@ class Input extends React.PureComponent {
 		}
 	}
 
-	handleChange({ target }, options) {
-		var newValue = options.newValue || target.value;
+	handleChange({ target }, { newValue } = {}) {
+		var newValue = newValue || target.value;
```
dstillman, line 72: No `var` here — the variable is already declared by its use as an argument.

diegodlh: Oh, sure! I was using the online editor and missed that.

dstillman:
Thanks!

diegodlh:
Thank you for guiding me through the process!

PR #1941, opened 2021-01-01
Add ability to open/save scripts in Run JS window

fletcherhaz:
https://forums.zotero.org/discussion/76808/feature-request-more-options-for-the-javascript-runner

Here is what it looks like on MacOS now:
Screen Shot 2021-01-01 at 4 40 08 PM


fletcherhaz:
@dstillman I started working on assigning shortcut keys, but I wasn't sure the best place to store that data and I wanted to get your thoughts on this design/functionality.

For shortcut keys, the best I came up with was a hidden JSON file in the `scripts` directory, which was read upon startup. I don't 100% understand how the pref panes work. Do you have any better ideas? Are we going to have to move away from the pref pane in the future?

PR #1945, opened 2021-01-22
Individual sync preferences for groups

fletcherhaz:
Moving this discussion over from #1919.

Here is the same concept as the preferences discussed there, but more polished, functional and the dialog is now all HTML.
I'm not exactly sure how we should combine the two pull requests, but probably the best way to to temporarily hide the time to live settings.

Screen Shot 2021-01-22 at 4 24 35 PM
Screen Shot 2021-01-22 at 4 25 11 PM

### DIFF
--- a/chrome/content/zotero/xpcom/sync/syncLocal.js
+++ b/chrome/content/zotero/xpcom/sync/syncLocal.js
@@ -144,9 +144,9 @@ Zotero.Sync.Data.Local = {
 			if (io.accept) {
 				var resetDataDirFile = OS.Path.join(Zotero.DataDirectory.dir, 'reset-data-directory');
 				yield Zotero.File.putContentsAsync(resetDataDirFile, '');
-				
+
+				Zotero.Prefs.resetBranch([], 'extensions.zotero.sync.storage.groups.');
 				Zotero.Prefs.clear('sync.storage.downloadMode.groups');
-				Zotero.Prefs.clear('sync.storage.groups.enabled');
 				Zotero.Prefs.clear('sync.storage.downloadMode.personal');
 				Zotero.Prefs.clear('sync.storage.username');
 				Zotero.Prefs.clear('sync.storage.url');
--- a/chrome/locale/en-US/zotero/preferences.dtd
+++ b/chrome/locale/en-US/zotero/preferences.dtd
@@ -61,15 +61,19 @@
 <!ENTITY zotero.preferences.sync.fileSyncing.myLibrary				"Sync attachment files in My Library using">
 <!ENTITY zotero.preferences.sync.fileSyncing.groups				"Sync attachment files in group libraries using Zotero storage">
 <!ENTITY zotero.preferences.sync.fileSyncing.groups.global			"Global Settings">
-<!ENTITY zotero.preferences.sync.fileSyncing.groups.custom			"Use Custom Settings">
+<!ENTITY zotero.preferences.sync.fileSyncing.groups.custom			"Group-Specific Settings">
+<!ENTITY zotero.preferences.sync.fileSyncing.groups.custom.use		"Use Custom Settings">
 <!ENTITY zotero.preferences.sync.fileSyncing.groups.sync			"Sync attachment files">
-<!ENTITY zotero.preferences.sync.fileSyncing.groups.customize		"Customize...">
-<!ENTITY zotero.preferences.sync.fileSyncing.groups.revert			"Revert All Groups to Global Settings">
+<!ENTITY zotero.preferences.sync.fileSyncing.groups.customize		"Customize Group Settings…">
+<!ENTITY zotero.preferences.sync.fileSyncing.groups.revert			"Use Global Settings for All Groups">
 <!ENTITY zotero.preferences.sync.fileSyncing.download				"Download files">
 <!ENTITY zotero.preferences.sync.fileSyncing.download.atSyncTime	"at sync time">
 <!ENTITY zotero.preferences.sync.fileSyncing.download.onDemand		"as needed">
+<!ENTITY zotero.preferences.sync.fileSyncing.ttl.day				"1 day">
+<!ENTITY zotero.preferences.sync.fileSyncing.ttl.week				"1 week">
+<!ENTITY zotero.preferences.sync.fileSyncing.ttl.month				"1 month">
+<!ENTITY zotero.preferences.sync.fileSyncing.ttl.multipleMonths		"3 months">
 <!ENTITY zotero.preferences.sync.fileSyncing.remove				    "Remove downloaded files if not opened for">
-<!ENTITY zotero.preferences.sync.fileSyncing.remove.suffix			"day(s)">
 <!ENTITY zotero.preferences.sync.fileSyncing.verifyServer         "Verify Server">
 <!ENTITY zotero.preferences.sync.fileSyncing.tos1					"By using Zotero storage, you agree to become bound by its">
 <!ENTITY zotero.preferences.sync.fileSyncing.tos2					"terms and conditions">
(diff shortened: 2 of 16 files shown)


fletcherhaz:
Should the `Revert All` button simply turn off all the `Use custom settings` checkboxes, so your preferences are saved, just temporarily turned off, or actually clear all those preferences?

Currently it implements the first option since it is less jarring and relatively easily undone.

dstillman:
> Currently it implements the first option since it is less jarring and relatively easily undone.

Yeah, I think that's right.

chrome/locale/en-US/zotero/preferences.dtd
```diff
@@ -60,9 +60,16 @@
...
+<!ENTITY zotero.preferences.sync.fileSyncing.groups.global			"Global Settings">
+<!ENTITY zotero.preferences.sync.fileSyncing.groups.custom			"Use Custom Settings">
+<!ENTITY zotero.preferences.sync.fileSyncing.groups.sync			"Sync attachment files">
+<!ENTITY zotero.preferences.sync.fileSyncing.groups.customize		"Customize...">
+<!ENTITY zotero.preferences.sync.fileSyncing.groups.revert			"Revert All Groups to Global Settings">
 <!ENTITY zotero.preferences.sync.fileSyncing.download				"Download files">
 <!ENTITY zotero.preferences.sync.fileSyncing.download.atSyncTime	"at sync time">
 <!ENTITY zotero.preferences.sync.fileSyncing.download.onDemand		"as needed">
+<!ENTITY zotero.preferences.sync.fileSyncing.remove				    "Remove downloaded files if not opened for">
+<!ENTITY zotero.preferences.sync.fileSyncing.remove.suffix			"day(s)">
```
dstillman, line 72: I prefer to avoid this, which is a little clumsy and also possibly not easily localizable. That's why it might be better to put a few time and unit options in the drop-down ("1 day", "1 week", "1 month", "3 months"), even if it's less customizable. That also has the advantage of letting you quickly configure everything with the mouse, instead of having to keep switching between the mouse and the keyboard across many groups.

Assuming we still stored the pref as days, someone could still customize in the Config Editor, and then I guess we'd want to display a "[x] days" value in the list, at the top — a few minutes' extra work, but why not? (That should be localized in zotero.properties as `%1$S day;%1$S days`, with the number given a second time as the third parameter to `getString()` — see `findPDF.pdfsAdded` in attachments.js for an example. Other languages with different numbers of plural forms will localize differently.)

(For emptying the trash in the General pane, we do use "days", since that defaults to 30 and you probably don't want it to be 1 anyway, but for configuring multiple times here the drop-down seems a little easier to me.)

defaults/preferences/zotero.js
```diff
@@ -162,6 +162,10 @@ pref("extensions.zotero.sync.storage.groups.enabled", true);
 pref("extensions.zotero.sync.storage.downloadMode.personal", "on-sync");
 pref("extensions.zotero.sync.storage.downloadMode.groups", "on-sync");
 pref("extensions.zotero.sync.fulltext.enabled", true);
+pref("extensions.zotero.sync.storage.groups.ttl", false);
+pref("extensions.zotero.sync.storage.groups.ttl.value", 30);
+pref("extensions.zotero.sync.storage.personal.ttl", false);
+pref("extensions.zotero.sync.storage.personal.ttl.value", 30);
```
dstillman, line 168: `personal` above `groups`

scss/mac/_preferences.scss
```diff
@@ -0,0 +1,52 @@
+.group-files-sync-container {
+	font-family: Lucida Grande, Lucida Sans Unicode, Lucida Sans, Geneva, -apple-system, sans-serif;
+	
+	select {
+		font-family: Lucida Grande, Lucida Sans Unicode, Lucida Sans, Geneva, -apple-system, sans-serif;
+		font-size: 13px;
+	}
+	
+	input {
+		font-family: Lucida Grande, Lucida Sans Unicode, Lucida Sans, Geneva, -apple-system, sans-serif;
```
dstillman, line 10: This seems like something that should be in a global variable?

chrome/content/zotero/xpcom/sync/syncLocal.js
```diff
@@ -144,9 +144,9 @@ Zotero.Sync.Data.Local = {
 			if (io.accept) {
 				var resetDataDirFile = OS.Path.join(Zotero.DataDirectory.dir, 'reset-data-directory');
 				yield Zotero.File.putContentsAsync(resetDataDirFile, '');
-				
+
+				Zotero.Prefs.resetBranch([], 'extensions.zotero.sync.storage.groups.');
```
dstillman, line 148: In addition to this, we should clear the groupID-specific one when a group is erased — probably in `_finalizeErase()` in group.js.

dstillman (commented):
- Maybe "Use Global Settings for All Groups"? Less technical and more in line with "Use Custom Settings".
- If we're going to display the global settings in the dialog, I think we might as well leave them editable, even if it's redundant?
- There should be a "Group-Specific Settings" label above the scroll box.
- Some of the element sizes (menu lists, OK button) are a little weird. Should be the same size as the main pref window. (I know that's not the case for the "Choose Libraries…" window.)
- "Customize..." → "Customize Group Settings…"? (Including proper ellipsis)
- I'll tweak the styling, but can remove the gray box within each group to start

chrome/locale/en-US/zotero/preferences.dtd
fletcherhaz: This makes sense. I was trying to be clever and minimizing the number of translation, I'm not sure why. Thanks for the detail.

fletcherhaz:
I worked more on trying to replicate the styles exactly. I think this definitely hits a few more of those details. The big one I have not solved is the `select` dropdown, which looks terrible and is not styled at all. There was talk about using a custom element for that? I ran into this problem with the Advanced Search dialog as well, but I don't think anyone has used a `select` yet in converted code.

Screen Shot 2021-01-25 at 1 07 33 PM
Screen Shot 2021-01-25 at 1 24 48 PM

dstillman (commented):
Haven't tried this yet but it's looking good. I'll work on the styling, including the drop-downs. Probably can just add this to the original PR and hook up the prefs, and we'll merge it all together?

chrome/locale/en-US/zotero/preferences.dtd
```diff
@@ -61,15 +61,19 @@
...
-<!ENTITY zotero.preferences.sync.fileSyncing.groups.revert			"Revert All Groups to Global Settings">
+<!ENTITY zotero.preferences.sync.fileSyncing.groups.customize		"Customize Group Settings…">
+<!ENTITY zotero.preferences.sync.fileSyncing.groups.revert			"Use Global Settings for All Groups">
 <!ENTITY zotero.preferences.sync.fileSyncing.download				"Download files">
 <!ENTITY zotero.preferences.sync.fileSyncing.download.atSyncTime	"at sync time">
 <!ENTITY zotero.preferences.sync.fileSyncing.download.onDemand		"as needed">
+<!ENTITY zotero.preferences.sync.fileSyncing.ttl.day				"1 day">
+<!ENTITY zotero.preferences.sync.fileSyncing.ttl.week				"1 week">
+<!ENTITY zotero.preferences.sync.fileSyncing.ttl.month				"1 month">
+<!ENTITY zotero.preferences.sync.fileSyncing.ttl.multipleMonths		"3 months">
```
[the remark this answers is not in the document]
dstillman: And probably `.day` → `.oneDay`, `.week` → `.oneWeek`, etc, for clarity and in case we want to add others?

fletcherhaz:
Merging this into #1919 and closing.

PR #1946, opened 2021-01-23
Issue99

### LINKED ISSUE #99: Use filename components in metadata retrieval
http://forums.zotero.org/discussion/22882/

patricknaughton01:
Hello,

I took a crack at issue 99. I added three functions to the pdf recognition process that try to extract the DOI, ISBN, and/or year from the file name of an added pdf if the information can't be found in the pdf metadata. Right now these are three separate functions, but they could easily be merged into one function that just also accepts a regular expression. I wasn't sure what would work better stylistically (I thought it would be ugly to have regular expressions floating around in the normal pdf recognition code, but the current implementation is also quite repetitive). 

I looked at the contribution guide but didn't see anything about formatting the commit history. Right now these edits are in three separate commits (and there's a merge commit that I pulled from the upstream master). Would you prefer that I squash these down into just one change?

Thanks for keeping this project open source. I use Zotero pretty much everyday and really appreciate the work you all do.

P.S. I think my IDE also inadvertently eliminated a lot of trailing whitespace... 

Closes #99


chrome/content/zotero/xpcom/recognizePDF.js
```diff
@@ -575,16 +588,72 @@ Zotero.RecognizePDF = new function () {
...
+	/**
+	 * Tries to extract ISBN from a file name
+	 * @param {str} fileName - File name to examine
+	 * @return {str} - Found ISBN or null if nothing found.
+	 */
+	function _extractISBNFromFileName(fileName){
+		// Modified the expression given at https://stackoverflow.com/a/41271701
+		// Currently only searches for 13 digit ISBNs
+		let isbnRe = /(?=(?:\D*\d){13})((?:(?:978-[01])|(?:979-8))[\d-]+)/;
+		let isbnRes = fileName.match(isbnRe);
```
dstillman, line 630: Use `Zotero.Utilities.cleanISBN()` instead

chrome/content/zotero/xpcom/recognizePDF.js
```diff
@@ -575,16 +588,72 @@ Zotero.RecognizePDF = new function () {
...
+	}
+
+	/**
+	 * Tries to extract Year from a file name
+	 * @param {str} fileName - File name to examine
+	 * @return {str} - Found year or null if nothing found.
+	 */
+	function _extractYearFromFileName(fileName){
+		let yearRe = /(?=(?:\d){4})((?:(?:20)|(?:19))[\d]{2})/;
+		let yearRes = fileName.match(yearRe);
```
dstillman, line 647: You're going to make some historians sad here. I would think some people might have PDFs with archival documents from further back, no?

But also, 1) this will match 4 digits within a larger number and 2) it seems way too complicated. Can't it just be `\b((?:19|20)\d{2})\b`?

dstillman (changes requested):
@mrtcode: Any problems you see with this? If a DOI or ISBN is recognized by the server, or if a year is recognized/retrieved, these won't be used.

chrome/content/zotero/xpcom/recognizePDF.js
```diff
@@ -1,39 +1,39 @@
 /*
     ***** BEGIN LICENSE BLOCK *****
-
```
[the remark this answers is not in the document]
patricknaughton01: Just, curious, is this just to avoid having a bunch of changed lines in the repo, or is there a specific reason you want the spaces on the empty lines?

chrome/content/zotero/xpcom/recognizePDF.js
patricknaughton01: Yes, that would probably be simpler. In terms of point 1, I wasn't sure what a good fix was because you could have for example "author2020.pdf" which would not have a word border at the beginning. I guess maybe something like `\D((?:19|20)\d{2})\D`?

In terms of catching older years, do you think it makes more sense to include a few more prefixes to catch (e.g. 18, 17, ...) or just look for an isolated 4 digit number?

dstillman: > I guess maybe something like \D((?:19|20)\d{2})\D?

That seems reasonable.

> In terms of catching older years, do you think it makes more sense to include a few more prefixes to catch (e.g. 18, 17, ...) or just look for an isolated 4 digit number?

It probably doesn't really matter — I don't actually have any idea if people have PDFs of archival documents with years in them, and this is a last-resort feature anyway, so if it doesn't work, that's fine. And the lower the number, the more likely there'd be false positives. Maybe we should just stick to 19 and 20.

chrome/content/zotero/xpcom/recognizePDF.js
```diff
@@ -613,17 +610,17 @@ Zotero.RecognizePDF = new function () {
 		// Replace `@` sign with `/` (`@` sign commonly used because `/` not
 		// allowed in file names).
 		// Grab first element of array (first captured group)
-		doiString = doiRes[1].replace("@", "/");
-		Zotero.debug(`Found DOI (${doiString}) in file name`);
+		let doiString = doiRes[1].replace("@", "/");
+		Zotero.debug(`Found DOI ${doiString} in file name`);
```
dstillman, line 614: "file name" → "filename", here and elsewhere

chrome/content/zotero/xpcom/recognizePDF.js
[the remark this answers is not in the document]
dstillman: Both. Patches shouldn't have irrelevant changed lines, and Zotero style calls for (or at least allows for) indented newlines, which make editing less annoying in some editors.

You can revert these by doing `git reset HEAD^` and using `git checkout -p` to undo these changes, and then commit everything else and force-push.

dstillman (changes requested):

patricknaughton01:
Sorry these changes took so long, if you're still considering this PR, I went through and made the whitespace changes. It looks like the testing is not running automatically anymore but I merged the latest updates to the master branch in so that there are no conflicts. Thanks.

PR #1950, opened 2021-02-05
Do not send back `saveSingleFile: true` if saving a PDF snapshot.

fletcherhaz:
Fixes and tests for https://github.com/zotero/zotero-connectors/issues/343

### DIFF
--- a/chrome/content/zotero/xpcom/connector/server_connector.js
+++ b/chrome/content/zotero/xpcom/connector/server_connector.js
@@ -1143,7 +1143,9 @@ Zotero.Server.Connector.SaveSnapshot.prototype = {
 			return 500;
 		}
 		
-		return [201, "application/json", JSON.stringify({ saveSingleFile: !data.skipSnapshot })];
+		return [201,
+			"application/json",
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/xpcom/connector/server_connector.js
```diff
@@ -1143,7 +1143,11 @@ Zotero.Server.Connector.SaveSnapshot.prototype = {
 			return 500;
 		}
 		
-		return [201, "application/json", JSON.stringify({ saveSingleFile: !data.skipSnapshot })];
+		return [201,
+			"application/json",
+			JSON.stringify({
+				saveSingleFile: !data.skipSnapshot && !(data.pdf && library.filesEditable)
```
dstillman, line 1149: Won't this be true if it's a PDF but the library doesn't have file editing enabled? `!(true && false) == true`

(It looks like in that case we now create an empty webpage item with the PDF URL, which is not ideal, but I'm not sure there's a better option. But we still don't want /saveSingleFile to be called.)

PR #1962, opened 2021-02-23
Add context menu for React textboxes to Create Parent Dialog

### LINKED ISSUE #1961: Context menu doesn't show in Create Parent Item
https://forums.zotero.org/discussion/comment/375888/#Comment_375888

fletcherhaz:
Fixes #1961

### DIFF
--- a/chrome/content/zotero/createParentDialog.js
+++ b/chrome/content/zotero/createParentDialog.js
@@ -78,3 +78,27 @@ function doManualEntry() {
 	io.dataOut = { parent: false };
 	window.close();
 }
+
+// Support context menus on HTML text boxes
+//
+// Adapted from editMenuOverlay.js in Fx68
+window.addEventListener("contextmenu", e => {
+	const HTML_NS = "http://www.w3.org/1999/xhtml";
+	let needsContextMenu =
+		e.target.ownerDocument == document &&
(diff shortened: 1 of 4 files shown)


fletcherhaz:
@dstillman We are probably going to need this in more dialogs as we convert to React. Do we want to pull these pieces out into a separate XUL and JavaScript file? This may also be a problem for HTML only dialogs, but I have not thought extensively about that.

dstillman:
Pull out, yes. XUL overlay, no, because those are gone for sure (whereas some custom elements, XUL or otherwise, will remain). Ideally we'd have a JS file that we can pull in via &lt;script&gt; that gives us automatic context menu support in any input field in the window.

fletcherhaz:
I bundled it all up in a single JS file. I don't know if there is nice way to create XUL fragment like (https://searchfox.org/mozilla-esr68/source/toolkit/content/editMenuOverlay.js)[this], but it works this way.

Two things to look at are:
1) Localization: I'm not sure if that is the right way to load the strings (to prevent having to translate them again) or we should move them over to zotero.properties.
2) Includes: I don't know if there is a better way to include the editMenuOverlay.js and globalOverlay.js so you only have one include to add this functionality.

Issue #1842, opened 2020-06-30
Quick Format dialog: White square in dark theme on Linux


dstillman:
https://forums.zotero.org/discussion/83921/quick-format-citation-dialog-has-white-square-in-dark-theme

Might be an easy fix

PR #1965, opened 2021-02-25
Fix for the centering issue for the logo on the Quick Format window.

WesleyFord:
A fix for [zotero/zotero#1842](https://github.com/zotero/zotero/issues/1842)

The original issue on the forums stated that it looked like the logo was off-centered on a white background. Upon further investigation, I found that the white background wasn't part of the logo or background but rather the toolbarbutton used for the dropdown menu to navigate to Classic View. I increased the left margin on the icon slightly to put it at the exact center of the white box.

**Before:**

**After:**

### DIFF
--- a/chrome/skin/default/zotero/integration.css
+++ b/chrome/skin/default/zotero/integration.css
@@ -51,7 +51,6 @@
 }
 
 #add[disabled="true"] {
-	list-style-image: url('chrome://zotero/skin/citation-down.png');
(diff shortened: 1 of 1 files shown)


dstillman:
Thanks — but to be clear, the issue being referred to is the white background. There's not supposed to be a white box, so it doesn't make sense to center something within it (which would also affect the layout in non-dark themes).

dstillman:
Closing this, but feel free to open a separate pull request for that issue if you want to look into it.

PR #2009, opened 2021-03-22
Fix background color of the dropdown menu.

WesleyFord:
This is another go at issue #1842. I took your suggestion about fixing the color and found that this was a simple fix if you remove the hardcoded white color of the dropdown menu. This fixes the issue in dark theme in Linux without affecting the appearance of light mode, as shown below. There's also no noticeable difference when this change is applied to the Windows build.

Original appearance in dark theme:

Fix in dark themed Linux:
ZoteroDarkThemeFix

Fix in light themed Linux:
ZoteroLightThemeFix


dstillman:
Looks good — thanks.

PR #2015, opened 2021-03-27
Fixed dark theme on "About Zotero" page

Blychro:
[Report: 1517701677](https://forums.zotero.org/discussion/88473/zotero-linux-dark-theme-issues)

Under the help tab, the "About Zotero" page background now reacts to the Linux theme instead of being hard coded to have a white background.

Before:
Before

After:
After

### DIFF
--- a/chrome/skin/default/zotero/about.css
+++ b/chrome/skin/default/zotero/about.css
@@ -7,7 +7,6 @@ dialog
 
 #aboutcontent
 {
-	background: white;
 	padding: 10px;
 }
 
--- a/chrome/skin/default/zotero/about.css
+++ b/chrome/skin/default/zotero/about.css
@@ -7,22 +7,14 @@ dialog
 
 #aboutcontent
 {
-	padding: 10px;
+	background: white;
+	padding: 10px 16px;
+	font-size: 13px;
+	margin-bottom: 10px;
 }
 
-#column1 {
-	width: 23em;
-	padding-right: 2em;
-}
-
-#column2 {
-	margin-right: 1em;
-}
-
-#name
-{
-	font-size: large;
-	font-weight: bold;
+#logo {
+	margin: 7px 0 0 7px;
 }
 
 #version
--- a/chrome/skin/default/zotero/about.css
+++ b/chrome/skin/default/zotero/about.css
@@ -7,6 +7,7 @@ dialog
 
 #aboutcontent
 {
+	background: white;
 	padding: 10px;
 }
 
--- a/chrome/skin/default/zotero/about.css
+++ b/chrome/skin/default/zotero/about.css
@@ -7,7 +7,6 @@ dialog
 
 #aboutcontent
 {
-	background: white;
 	padding: 10px 16px;
 	font-size: 13px;
 	margin-bottom: 10px;


dstillman:
Thanks for working on this.

The About pane has just been updated, so you'll want to check with that, since there are now conflicts.

A few notes:

1) Your diff can't have whitespace changes, so you'll need to undo those.

2) Any UI changes that aren't limited to a specific platform's CSS need to be tested across all three supported platforms.

3) Simply removing the white background isn't acceptable here, since that changes the design. If you want this to look right with a dark theme, you'll need to figure out a way to do that without affecting the default light theme.

Blychro:
As far as I know, I only removed the background hard coding, so unless I am misunderstanding you, I don't know what added whitespace I made. If you're referring to the multiple commits, this was my first pull request and I was trying to figure out how to comment on the request, and I kept making the fix and undoing it.

I checked the fix against Linux's other themes and it still looks right:

Light Theme:
AfterLight

Standard Theme:
AfterStandard

I'll check the updated CSS file and check the change on the other platforms, hopefully either today or tomorrow.

dstillman:
> As far as I know, I only removed the background hard coding, so unless I am misunderstanding you, I don't know what added whitespace I made.

Your editor likely did it automatically. You don't have to take my word for it — just look at the [diff for this PR](https://github.com/zotero/zotero/pull/2015/files). You should never submit a pull request without reviewing the diff first locally to make sure it's what you expect.

> I checked the fix against Linux's other themes and it still looks right

No it doesn't — you changed the design.

Blychro:
I'm not sure if it's a platform thing (I haven't had a chance to check the file on non Linux) but I rebuilt the code with the new About.css both with my change and without it and it looks like this with before on the left and after on the right:

Dark Theme:
Compare Dark

Standard Theme:
Compare Standard

Light Theme:
Compare Light


I'm not trying to be difficult, honest, but I do not see how the other themes were changed. As for the whitespace and changing it to match the new About.css, should I make a new pull request once I check the three platforms and make sure the other two concerns are taken care of?

dstillman:
Don't open new pull requests for the same issue. Anything that needs to be fixed should be fixed here.

> I do not see how the other themes were changed

You removed the white background. So when you remove the white background, you get a gray background, and that's not the design. And that's exactly what your screenshots show (though you have some of those images reversed — for the last two pairs, the ones you described as "after" are really "before").

dstillman:
Also, you shouldn't be manually applying changes from master. You would need to rebase on master. We can't help you with that, but there are plenty of resources on rebasing pull requests online.

Blychro:
I've never used CSS files before so I'm trying my best to figure this out. Should I find a way to respond to the theme and set accordingly, or should I hard code the entire page to not respond to the theme?

Blychro:
Would this be acceptable or does it need a dark background?
image

dstillman:
Before doing anything else, you'll need to rebase on the current master, and use that for all changes and screenshots. (And I don't know what why your later screenshots don't show proper text wrapping, but obviously that can't change.)

Beyond that, the main requirement here is that the current design does not change on themes where it looks correct. If you manage to improve rendering on a theme where it currently looks incorrect (e.g., invisible text), we'll consider the patch.

Blychro:
Unless I misunderstood what rebasing is, I believe I have set it up. My new commits are going towards a branch on my page called "aboutBranch".  As for the text wrapping that stopped occurring after I started using the new code that was added last night so I assumed it was intentional. The hard coding of the text to black leaves the correct themes unchanged. I also made the same changes I made to the most recent About.css to the errorReport.css so it will be seen clearly in the dark theme. Likewise, it remains unchanged in the already correct themes.

dstillman:
Yes, you are misunderstanding. As I said, there are many resources online that explain how to rebase a pull request.

I'm afraid we're not going to be able to help you here. We're happy to accept an appropriate patch that improves rendering without changing existing designs, but what you've provided is not correct or useful.

PR #2016, opened 2021-03-27
Fix dark theme on Item Type

Dylan-Evan:
[Report: 1517701677](https://forums.zotero.org/discussion/88473/zotero-linux-dark-theme-issues)
Number 1, The item type was hard coded to black therefore it did not look well when used in dark mode. I removed the hard code so now it should look nicer when using dark mode.

Before fix:
Item Type pre fix

After fix:
Item Type color fix


dstillman:
Thanks!

PR #2020, opened 2021-03-30
Use only year in OpenURL date

bwiernik:
CrossRef's OpenURL processor chokes on full ISO dates https://forums.zotero.org/discussion/comment/379397/#Comment_379397


dstillman:
I assume you didn't test this? I don't see how this would be a valid change — you're taking a JavaScript string and trying to access a `year` property.

And as adamsmith says in the forums, there's nothing wrong with this function. If it needs to be post-processed before being sent to a given service, that should happen elsewhere.

PR #2021, opened 2021-03-30
error report page dark theme compatible

Blychro:
Report [1517701677](https://forums.zotero.org/discussion/88473/zotero-linux-dark-theme-issues):

Originally, if the user used the dark theme on Linux with Zotero, the error report page would display extremely light gray text on a white background.  I set the text to match the default text color of the light and standard themes. This keeps the already correct windows from being changed while making the dark theme readable. By keeping the window the error message is displayed in white with a darker text, it makes it easier for the user to see the error window, separating the message from the surrounding UI.

Here are some comparison screenshots (before on the left and after on the right):

Dark Theme:
CompareDark

Light Theme:
CompareLight

Standard Theme:
CompareStandard

### DIFF
--- a/chrome/content/zotero/errorReport.xul
+++ b/chrome/content/zotero/errorReport.xul
@@ -2,6 +2,7 @@
 
 <?xml-stylesheet href="chrome://global/skin/" type="text/css"?>
 <?xml-stylesheet href="chrome://zotero/skin/errorReport.css" type="text/css"?>
+<?xml-stylesheet href="chrome://zotero-platform/content/zotero-react-client.css"?>
 
 <!DOCTYPE window SYSTEM "chrome://zotero/locale/zotero.dtd">
 
--- a/chrome/skin/default/zotero/errorReport.css
+++ b/chrome/skin/default/zotero/errorReport.css
@@ -5,7 +5,6 @@ description {
 /* Intro pane */
 #zotero-error-message {
 	background: #FFF;
-	color: rgb(61, 61, 61);
 }
 
 #zotero-error-message .textbox-textarea {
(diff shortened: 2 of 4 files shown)


Blychro:
I have not committed the changes to the Pull Request yet because I wanted to check with you first.

These are the changes that will be made by the new pull request (The original change I made will be reversed:
image

Here is Dark Mode:
image

Here is Light Mode:
image

dstillman:
You can just push the changes here. No need for a new pull request, no need for screenshots of code.

dstillman:
Right, that's not correct at all. Again, you need to read up on rebasing, and review your diff before pushing it.

Blychro:
I think I updated the pull request badly

Blychro:
I did rebase it, just not the right branch

Blychro:
It appears when I rebased the branch waiting on the pull request, it added the commits done in the last week to this pull request and then added my commit to the end of the list. Is there a way to undo this or should I just close the request and make a new one?

dstillman:
You need to rebase against the current master. Again, there are plenty of resources online that you can review for this. This has nothing to do with Zotero, and while we're grateful for contributions, we can't walk people through using git. This would be the same process contributing to any project on GitHub, so it's worth investing the time in understanding how to do it properly.

Blychro:
I did rebase against the current master. I think what happened was I rebased it against the current master from my master. Then, when I went to update the branch I was making changes to, instead of rebasing the origin again, I rebased using my rebase as a reference. When I use the "git rebase upstream/master" command on the right branch, now it just says I'm already up to date. The strange part is when I compare my branch with the zotero:main branch, none of the in between is given, only my changes. That is why I asked if I should just reset the pull request

dstillman:
No, you're not rebasing against the current master or the intervening commits wouldn't show up. Your local branches likely aren't up to date.

Blychro:
I wasn't pushing the rebase. It should be good to go now

PR #2029, opened 2021-04-08
HTML tree merge

dstillman:
Follow-up to #1827 with [a bunch of conflicts resolved](https://gist.github.com/dstillman/d1e1bcbf6eb9180e9702319ab6b7493f), so it can be tested with current databases. I think it'll be easiest to leave the original commits alone and just push additional fixes here.

This is awesome! I haven't really studied the code yet beyond what was necessary for fixing conflicts, but some things I found while trying it:

- Page Down moves the selection instead of just scrolling
- "Title" shouldn't be an option in the column picker. It contains the twisty and shouldn't be able to be hidden.
- "Restore Column Order" is missing in the column picker, so there's just a separator at the bottom:

columnpicker-separator

- Header separators are sometimes misaligned:

misaligned-column

advanced-search-column-misalignment

- For the warning about pref size, can/should we just persist to a dedicated file, as it says?
- We should plan to support configurable columns in Advanced Search, which is a common, extremely longstanding request. Where should we persist that?
- The Advanced Search window opened for me with the tree hidden. I can expand it, but that should have a minimum height — it shouldn't even be possible to collapse that completely:

collapsed-advanced-search-tree

- When pressing "-" with a child item selected, the parent row should be selected and kept in view. Currently selection is lost.
- The selected row isn't being calculated properly after pressing "+" to expand all items. Down-arrow scrolls way up and selects the wrong row.
- +/- in the collections tree is extremely slow for me for some reason — takes almost 5 seconds, with pinwheel. Items tree, with thousands more rows, is much faster.
- "Loading items…" in the Advanced Search window shows up in the top-left corner
- Something weird is going on when I try to drag a child note out from under a parent item. The hover effects for both making a standalone item and for dragging to another item seem wrong:

row-hover

- The alternate-row shading of the first row changes when I expand an item lower in the list:

alt1
alt2

- After changing the column order such that the selected row isn't at the top, Home doesn't select the first row (though down-arrow does select the second row)
- Double-clicking a row isn't opening the PDF


dstillman:
Failing test:

```
     Attachment pane
    ✖ [FAIL] should refresh on file rename
      expected 'test.png' to equal 'test2.png' at
@resource://zotero-unit-tests/itemPaneTest.js:271:7
```

dstillman:
Feeds aren't indented:

feeds-indent

dstillman:
This is missing a key feature of the existing tree: when doing Select All, only search matches should be selected. Gray context rows should not. (We might not have a test for this, but it'd be good to have one.)

dstillman:
One detail of the above: if you do a search that matches some child items, press "-", and then do Select All, it will expand parent items as necessary to select the matching child items.

chrome/content/zotero/components/icons.jsx
```diff
@@ -53,7 +70,133 @@ function i(name, svgOrSrc, hasDPI=true) {
...
+
+/**
+ * Returns a DOM element for the icon class
+ *
+ * To be used in itemTree where rendering is done without react
+ * for performance reasons
+ * @param {String} icon
+ * @returns {Element}
+ */
+module.exports.getDomElement = function (icon) {
```
dstillman, line 192: `getDomElement` → `getDOMElement`

chrome/content/zotero/components/virtualized-table.jsx
```diff
@@ -0,0 +1,1350 @@
...
+	
+	***** END LICENSE BLOCK *****
+*/
+
+'use strict';
+
+const React = require('react');
+const PropTypes = require('prop-types');
+const cx = require('classnames');
+const JSWindow = require('./js-window');
```
dstillman, line 31: I know `js-window` is an homage to `react-window` but I think out of context "JS Window" is mostly just very confusing. Can we just call it 'virtualized-list` or something?

chrome/content/zotero/components/virtualized-table.jsx
```diff
@@ -0,0 +1,1350 @@
...
+		this.selected = new Set();
+		if (this._tree.invalidate) {
+			this._tree.invalidate();
+		}
+		this._updateTree();
+	}
+
+	/**
+	 * @param index
+	 * @returns {boolean} False if nothing to select and select handlers won't be called
```
dstillman, line 88: - Comment format here and elsewhere is missing the variable type
- Missing `shouldDebounce` explanation, here and elsewhere

chrome/content/zotero/components/virtualized-table.jsx
```diff
@@ -0,0 +1,1350 @@
...
+	 *        The index of the item in a full DFS traversal (ignoring collapsed
+	 *        nodes). Ignored if `item` is undefined.
+	 *
+	 * @param {Boolean} shiftSelect
+	 * 		  If true will select from focused up to index (does not update pivot)
+	 *
+	 * @param {Boolean} toggleSelection
+	 * 		  If true will add to selection
+	 *
+	 * @param {Boolean} movePivot
```
dstillman, line 616: Remove extra newlines

chrome/content/zotero/components/virtualized-table.jsx
```diff
@@ -0,0 +1,1350 @@
...
+		}
+	}
+}
+
+let TreeSelectionStub = {};
+for (const key of Object.getOwnPropertyNames(TreeSelection.prototype)) {
+	TreeSelectionStub[key] = () => 0;
+}
+TreeSelectionStub = Object.assign(TreeSelectionStub, {
+	pivot: 0,
```
dstillman, line 212: Define what the pivot is somewhere. (What's the pivot?)

dstillman:
- When clicking on an item in Duplicates Items, there's a noticeable delay before the other items in the set are selected. In the previous version it was instantaneous.
- Previously in Duplicate Items you could use the arrow keys to deselect a set and navigate the tree. In this version, the arrow keys cause sets to be selected. That might be actually be preferable in terms of accessibility, but the visual selection isn't totally reliable — when you move onto a new set, sometimes only the initial item is shown as selected, even though it shows the full set in the right-hand pane, and it's not until you move to a second item in the set that the full set is selected.

dstillman:
I'm not sure what the Attachments column is doing for sorting, but it's not quite right. Maybe it's ignoring inherited state from child items of collapsed parents? The end with filled circles is correct, but then I'm seeing a mixing of no-circle and empty-circle on the other side.

dstillman:
Since we're in control now, column headers should grow in height when font size is increased. (It was a bug that they didn't before.)

dstillman:
Newlines aren't being properly rendered in the items pane message. (Looks like we're passing as `\n\n`, but it's being interpreted as HTML.)

Previous (good):

Screen Shot 2021-04-09 at 1 46 10 AM

Now (bad):

Screen Shot 2021-04-09 at 1 46 15 AM

chrome/content/zotero/containers/itemTree.jsx
```diff
@@ -0,0 +1,3430 @@
+/*
```
dstillman, line 1: Any particular reason these are in `containers`? Because they call privileged code?

I think we're mostly done with the container/component distinction. Most of the "components" aren't really going to be usable outside of Zotero past initial development, and having the collection tree and item tree not be "components" is mostly just going to be frustrating.

chrome/content/zotero/containers/itemTree.jsx
```diff
@@ -0,0 +1,3430 @@
...
+	***** BEGIN LICENSE BLOCK *****
+	
+	Copyright © 2019 Center for History and New Media
+					George Mason University, Fairfax, Virginia, USA
+					http://zotero.org
```
dstillman, line 6: Should update all of these new files to say CDS instead:

https://github.com/zotero/zotero/blob/328865299b9f87eee7b40ee944f720b09fe0a089/chrome/content/zotero/components/search.jsx#L4-L6

(Maybe worth squashing those, if it doesn't totally blow up the merge commit.)

dstillman (commented):
Still going through it, but some things so far

dstillman:
The crash message (`Zotero.crash()`) should be permanent. Instead, it's now cleared if you switch to another collection. (I can't remember if we're doing something special for the crash message or if this is just a general property of an items pane message.)

chrome/content/zotero/containers/itemTree.jsx
adomasven: There is value in a separation between React components that are divorced from any business logic and just display stuff. Containers/components is the current naming/paradigm. The files in the containers folder could just sit in the main zotero folder, although there might also be value in keeping the Reactified UI stuff separated for organization purposes?

dstillman: Display-only components make sense if we have something like https://github.com/zotero/zotero/issues/2030.

I don't see a lot of value in separating `containers`. Most of the things in the parent directory are already creating UI. React happens to be one technology we use, and we'll likely use Custom Elements for some things as well, but who cares? I think it's more important to not have to think about or know what specific technologies something is currently using and just be able to look at _a list of files that deal with UI with some business logic in them_.

chrome/content/zotero/zoteroPane.js
```diff
@@ -1084,18 +983,62 @@ var ZoteroPane = new function()
...
+			const ItemTree = require('containers/itemTree');
+			var itemsTree = document.getElementById('zotero-items-tree');
+			ZoteroPane.itemsView = await ItemTree.init(itemsTree, {
+				id: "main",
+				dragAndDrop: true,
+				persistColumns: true,
+				columnPicker: true,
+				onSelectionChange: selection => ZoteroPane.itemSelected(selection),
+				onContextMenu: event => ZoteroPane.onItemsContextMenuOpen(event),
+				onActivate: (event, items) => ZoteroPane.onItemTreeActivate(event, items),
```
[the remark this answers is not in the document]
adomasven: Yep. We could have two handlers, but this is clear enough IMO and I don't think we'd ever want to do different things for Enter vs double-click.

resource/react-dom-server.js
```diff
@@ -0,0 +1 @@
+../node_modules/react-dom/umd/react-dom-server.browser.development.js
```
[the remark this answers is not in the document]
adomasven: The development version has detailed errors messages.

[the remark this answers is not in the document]
dstillman: Yeah, but it's slower, at least in theory. May not make a difference for this, but we do have an [open ticket](https://github.com/zotero/zotero/issues/1700) to use the production versions of React/ReactDOM.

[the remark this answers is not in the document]
adomasven: I don't think it's slower, just bigger. Unless we can identify this as a significant performance issue I'd rather include the development version.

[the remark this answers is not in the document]
dstillman: React website says slower as well, though it's possible they're talking largely about download time:

> By default, React includes many helpful warnings. These warnings are very useful in development. However, they make React larger and slower so you should make sure to use the production version when you deploy the app.

Also:

> It is expected that you use the development mode when working on your app, and the production mode when deploying your app to the users.

I think that was the idea behind #1700.

adomasven:
> For the warning about pref size, can/should we just persist to a dedicated file, as it says?

Idk what you mean here. Storing to prefs is persisting to a file - the prefs file. My comment in the code suggests that might get slow and my idea would be to persist to the DB instead (although that's also persisting to a file, just with some extra steps).

> We should plan to support configurable columns in Advanced Search, which is a common, extremely longstanding request. Where should we persist that?

The tree is designed such that it stores/loads column prefs based on tree id, so that's easy to enable. The question is whether we want to actually persist that, or if allowing to toggle column visibility there is enough? Either way this can always be considered after the merge.

> I'm not sure what the Attachments column is doing for sorting, but it's not quite right. Maybe it's ignoring inherited state from child items of collapsed parents? The end with filled circles is correct, but then I'm seeing a mixing of no-circle and empty-circle on the other side.

I just saw the same with the current master, and now I cannot reproduce it anymore with either. I didn't change the sorting code, so whatever is causing that isn't a new bug.

-----

All of the above comments have been addressed.

dstillman:
> Idk what you mean here. Storing to prefs is persisting to a file - the prefs file. My comment in the code suggests that might get slow and my idea would be to persist to the DB instead (although that's also persisting to a file, just with some extra steps).

We definitely don't want to persist to the DB — this is purely profile data, not database data. I'm mostly asking why Mozilla logs a warning here, and specifically recommends saving to a separate file rather than using the prefs system. Is there something about the prefs system that doesn't handle large chunks of data that causes them to recommend that?

> The tree is designed such that it stores/loads column prefs based on tree id, so that's easy to enable. The question is whether we want to actually persist that, or if allowing to toggle column visibility there is enough?

I don't see any reason we wouldn't persist. People will likely have their own preference for what they usually want to see when searching. But yes, we can address that after.

dstillman:
(If nothing else, constantly logging warnings in the console is going to get annoying and hide other valid messages, which would be sufficient reason to store in a separate file.)

adomasven:
> Is there something about the prefs system that doesn't handle large chunks of data that causes them to recommend that?

Idk, but it probably doesn't like writing so much text for a pref store at once and I also have a comment in the codebase that we shouldn't overuse pref storing or it might lead to degradation in performance. Either way I'll move that to a separate file in the profile dir to prevent the warning.

chrome/content/zotero/itemTree.jsx
```diff
@@ -40,6 +40,7 @@ Cu.import("resource://gre/modules/osfile.jsm");
 const TYPING_TIMEOUT = 1000;
 const CHILD_INDENT = 20;
 const COLORED_TAGS_RE = new RegExp("^[0-" + Zotero.Tags.MAX_COLORED_TAGS + "]{1}$");
+const COLUMN_PREFS_FILEPATH = OS.Path.join(Zotero.Profile.dir, "treePrefs.js");
```
dstillman, line 43: `.js` → `.json`

jryans:
> Is there something about the prefs system that doesn't handle large chunks of data that causes them to recommend that?

I believe the main concern is the I/O traffic, since the entire prefs file is written (off the main thread) any time any pref is changed. If you happen to use prefs to store many large objects, it would mean lots of I/O any time any of them changes, so I believe it's more about how the costs stack up over time, rather the effect of a single large pref in isolation. See the [libpref docs](https://firefox-source-docs.mozilla.org/modules/libpref/index.html#loading-and-saving) for more details.

PR #2039, opened 2021-04-13
Made About page text visible across default Linux themes

HollandeP:
Fix for issue in Report [1517701677](https://forums.zotero.org/discussion/88473/zotero-linux-dark-theme-issues)

I was unable to change the dark theme background since since XUL doesn't support prefers-color-scheme, so I made a Linux specific fix to ensure the text was visible across themes. 

Dark Theme Before/After:
Zotero About - Dark Before-After

Light Theme Before/After:
Zotero About - Light Before-After


dstillman:
Thanks!

PR #2052, opened 2021-05-07
Spell checker

mrtcode:
Instructions to try this out:
1. Build https://github.com/zotero/zotero-build/pull/8 and upload to `https://download.zotero.org/dictionaries/`.
2. Run `Zotero.Dictionaries.installByLocale('fr')` or `Zotero.Dictionaries.installByLocale('es-MX')`.

Some thoughts:
- Context menu option `Languages -> Add dictionaries...` just opens preferences, but we need to implement dictionaries management UI there.
- In theory, we could just store XPI files (in `<profile>/dictionaries/`) instead of extracting them, although in my quick test to mount XPI file directly, spell checker was complaining...
- Whether it's good or bad, but currently all available (156) dictionaries are downloaded by the Python script. Although in https://addons.mozilla.org/en-US/firefox/language-tools/ there are less and unclear on what criteria they are selected.
- `average_daily_users` count is included in our dictionaries index which is later used to determine the most popular dictionary targeting specific locale.
-  Dictionaries can be installed by Firefox extension id i.e. `Zotero.Dictionaries.install('en-NZ@dictionaries.addons.mozilla.org')`, or manually extracted to `<profile>/dictionaries/`.
- All dictionaries can later be updated with `Zotero.Dictionaries.update()`.

### DIFF
--- a/scss/abstracts/_variables.scss
+++ b/scss/abstracts/_variables.scss
@@ -21,7 +21,7 @@ $space-thumb: 42px;
 // Typography
 // --------------------------------------------------
 
-$font-family-sans-serif: -apple-system, BlinkMacSystemFont, "Segoe UI", "Helvetica Neue", Helvetica, Arial, sans-serif;
+$font-family-sans-serif: Lucida Grande, Lucida Sans Unicode, Lucida Sans, Geneva, -apple-system, BlinkMacSystemFont, "Segoe UI", "Helvetica Neue", Helvetica, Arial, sans-serif;
 $font-family-base:       $font-family-sans-serif;
 
 $font-size-base:   13px;


dstillman (commented):
A few small style things but otherwise this is amazing. Let's localize the unlocalized strings.

I'll add the prefs stuff on top.

dstillman:
> In theory, we could just store XPI files (in <profile>/dictionaries/) instead of extracting them, although in my quick test to mount XPI file directly, spell checker was complaining...

This would certainly be better if we could figure it out. But if it's just not supported, this is OK.

mrtcode:
> > In theory, we could just store XPI files (in /dictionaries/) instead of extracting them, although in my quick test to mount XPI file directly, spell checker was complaining...
> 
> This would certainly be better if we could figure it out. But if it's just not supported, extracting is OK.

How about storing XPI but extracting to tmp directory when loading it, if we can't load it directly?

dstillman:
On every startup? I think it would be better to just extract once in that case.

mrtcode:
Nice! Thanks for the fixes!

mrtcode:
Ok, I tried once again, and I'm still unable to load dictionary directly from XPI.

mrtcode:
The problem is that our protocol handler doesn't support this.

dstillman:
Which protocol handler?

mrtcode:
Well, our protocol handler is just for `zotero://`, so maybe the issue is not there.

In theory, we should be able to do [this](https://searchfox.org/mozilla-esr60/source/toolkit/components/extensions/Extension.jsm#1903-1907) with XPI file directly. I.e.:
```js
let uri = Services.io.newURI('jar:file:///some/path/french_spelling_dictionary-6.3.1webext.xpi!/dictionaries/fr-classic.aff', null, null);
let { file } = uri.QueryInterface(Ci.nsIFileURL);
spellCheck.addDictionary('fr', file);
```
But `nsIFileURL` is not supported for the JAR URI.

dstillman:
Yeah, that won't work. All that's doing is using the `file` property on an `nsIFileURL` to get an `nsIFile` — which is what `addDictionary()` needs — but an `nsIFile` can't point to a file within a ZIP.

dstillman:
OK, I've added a UI for managing dictionaries, along with various other changes.

One bug I've seen: when loading the Languages menu after clicking on a note, I usually/always don't get a checkbox for the active language. If I right-click a second time, the checkbox shows up. If I click to another note and back, it disappears again the first time.

A couple other things we're missing:

1) Automatic updates. Available updates are shown in the dictionary manager and are downloaded when clicking OK, but otherwise there's no checking for updates. This is as simple as calling `Zotero.Dictionaries.update()`, so maybe we want to just do that on startup if, I don't know, a week has passed since the last check?

2) Any sort of prompt to install a dictionary for a non-English locale. We can roll this out without that — just having other dictionaries available is a huge improvement — but it might be nice to prompt to install a non-English dictionary the first time we detect a high proportion of misspellings in a note, at least when the locale is set to non-en-US. Doing that in a non-annoying way (e.g., not when the user is actively typing) might be a little tricky.

mrtcode:
> One bug I've seen: when loading the Languages menu after clicking on a note, I usually/always don't get a checkbox for the active language. If I right-click a second time, the checkbox shows up. If I click to another note and back, it disappears again the first time.

It seems spell checking engine is only initialized when `contenteditable` area gets focus. And with right-click the popup is opened before `contenteditable` is focused. I'll see what we can do here.

> 1. Automatic updates. Available updates are shown in the dictionary manager and are downloaded when clicking OK, but otherwise there's no checking for updates. This is as simple as calling `Zotero.Dictionaries.update()`, so maybe we want to just do that on startup if, I don't know, a week has passed since the last check?

Yeah we can do that, I think.



> 2\. Any sort of prompt to install a dictionary for a non-English locale. We can roll this out without that — just having other dictionaries available is a huge improvement — but it might be nice to prompt to install a non-English dictionary the first time we detect a high proportion of misspellings in a note, at least when the locale is set to non-en-US. Doing that in a non-annoying way (e.g., not when the user is actively typing) might be a little tricky.

We could show a notification bar similar to the outdated note-editor message, just in another color and with 'x' button to close it. So if we would do that for non-en-US locales, for spell checking language with the same locale, and store a pref to only show this notification once (until it's closed), that doesn't seem to be too annoying.

dstillman:
Is it possible to test whether the delay is necessary and only use setTimeout when it is? Or only use it the first time after opening a note or something? 50 ms isn't a lot but it'd be better not to make that permanent.

mrtcode:
> Is it possible to test whether the delay is necessary and only use setTimeout when it is? Or only use it the first time after opening a note or something? 50 ms isn't a lot but it'd be better not to make that permanent.

Actually, I had the main part of this code before I opted in for a simpler `setTimeout` solution. Partially because I was somehow thinking that we will need to reimplement a similar popup for `pdf-reader` comments spell checking. But do we care about that?

dstillman:
I think people will expect spell check in the PDF reader, yes. But what's the connection to this?

The context menu will almost certainly be more heavily used in notes, so it's worth avoiding the delay there at least.

mrtcode:
No connection actually. And you are right about the popup.

Issue #2060, opened 2021-05-17
Closing tab can leave tooltip showing


dstillman:
https://forums.zotero.org/discussion/comment/383264/#Comment_383264

jryans:
Hmm, I must have been mousing very carefully when testing the first version... Somehow I did not manage to trigger this at the time. 😅 Anyway, should be easy to fix.

PR #2062, opened 2021-05-17
Hide tab title tooltips whenever the tab bar changes

jryans:
This hides tooltips (such as the tab title tooltip) whenever the tab bar state changes, such as when adding or removing a tab, to ensure we don't leave any tooltips behind that no longer match reality.

Additionally, this changes to `mousemove` to start the tab title tooltip, so that the behavior matches applications like Firefox more closely: you can always get the latest tooltip back after changing state by nudging the mouse (without having to exit and re-enter anything).

tab-title-mousemove

Fixes https://github.com/zotero/zotero/issues/2060

### DIFF
--- a/chrome/content/zotero/components/tabBar.jsx
+++ b/chrome/content/zotero/components/tabBar.jsx
@@ -36,15 +36,15 @@ const TabBar = forwardRef(function (props, ref) {
 	const mouseMoveWaitUntil = useRef(0);
 
 	useEffect(() => {
-		window.addEventListener('mouseup', handleMouseUp);
+		window.addEventListener('mouseup', handleWindowMouseUp);
 		return () => {
-			window.removeEventListener('mouseup', handleMouseUp);
+			window.removeEventListener('mouseup', handleWindowMouseUp);
 		};
 	}, []);
 
 	useImperativeHandle(ref, () => ({ setTabs }));
 
-	function handleMouseDown(event, id, index) {
+	function handleTabMouseDown(event, id, index) {
 		if (event.target.closest('.tab-close')) {
 			return;
 		}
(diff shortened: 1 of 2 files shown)


dstillman:
> This behavior seems to match tab bars from other applications (e.g. Firefox).

That's not what I'm seeing in Firefox 89 on macOS, for what it's worth. If I hover over a tab to show a tooltip, and then press Cmd-W, the tooltip goes away, and then I only need to move the mouse slightly for a tooltip to reappear. If that's not happening here (I haven't tested), it would be nice to fix — it's not a big deal, but if you're closing a whole bunch of tabs, it's nicer to just have to move the mouse slightly to see what the tab is rather than to have to move out and back in.

jryans:
> That's not what I'm seeing in Firefox 89 on macOS, for what it's worth. If I hover over a tab to show a tooltip, and then press Cmd-W, the tooltip goes away, and then I only need to move the mouse slightly for a tooltip to reappear. If that's not happening here (I haven't tested), it would be nice to fix — it's not a big deal, but if you're closing a whole bunch of tabs, it's nicer to just have to move the mouse slightly to see what the tab is rather than to have to move out and back in.

Ah you're right, it's a bit more subtle in Firefox than I thought. The current effect with this PR is perhaps too context dependent: if your action (such as Cmd+W to close the active tab) changes the tab under your cursor (a different one slides under it as a result of the close), then a small nudge is enough. But if you remain on the same tab, you have to move out and back in.

Anyway, I agree that doesn't make much sense as user, so I'll work on adjust this like you described.

jryans:
@dstillman Okay, ready for another look. Now you only need a small nudge over a tab to get the tooltip back in all cases.

dstillman:
OK, pushing this now, since it fixes the bug people were seeing.

This is now fairly closely to Firefox — in both apps there's some weird hiding on mousemove, albeit in different ways — but at some point we may want to consider something closer to what Chrome and Safari do, which is to keep a tooltip open once one appears as long as you remain in the tab bar, with immediate changing as you hover over different tabs. A bit non-standard, but it seems pretty clearly better for actually finding a tab you're looking for.

PR #2070, opened 2021-05-24
Restore `contextPane` `tags` and `related` tabs, display note links box

mrtcode:


chrome/content/zotero/bindings/relatedbox.xml
```diff
@@ -279,7 +279,8 @@
 								
 								p = win.ZoteroPane;
 							}
-								
+							
+							(win || window).Zotero_Tabs.select('zotero-pane');
```
dstillman, line 283: `zotero-pane` is the library pane?

This should probably just be part of `selectItem()`? Any code that selects items (e.g., a `zotero://select` link) presumably wants to select the library pane.

scss/components/_tagsBox.scss
```diff
@@ -65,6 +65,7 @@
 		flex-grow: 1;
 		margin: 0 2px;
 		//width: $item-pane-width - $icon-width - $delete-button-width - ($li-side-margin * 2);
+		width: 0;
```
dstillman, line 68: Add comment to explain this?

chrome/content/zotero/contextPane.js
```diff
@@ -773,17 +760,140 @@ var ZoteroContextPane = new function () {
...
 		var itemBox = document.createElement('zoteroitembox');
 		itemBox.setAttribute('flex', '1');
 		panelInfo.append(itemBox);
-		container.append(panelInfo);
+		// Notes panel
+		var panelNotes = document.createElement('tabpanel');
+		panelNotes.setAttribute('flex', '1');
+		panelNotes.setAttribute('orient', 'vertical');
+		var deck = document.createElement('deck');
+		deck.className = 'notes-deck';
```
dstillman, line 807: Unused?

chrome/content/zotero/contextPane.js
```diff
@@ -773,17 +760,140 @@ var ZoteroContextPane = new function () {
...
+		var panelRelated = document.createElement('tabpanel');
+		var relatedBox = document.createElement('relatedbox');
+		relatedBox.setAttribute('flex', '1');
+		relatedBox.className = 'zotero-editpane-related';
+		panelRelated.addEventListener('click', (event) => {
+			if (event.originalTarget.closest('.zotero-clicky')) {
+				Zotero_Tabs.select('zotero-pane');
+			}
+		});
+		panelRelated.append(relatedBox);
```
dstillman, line 886: This is all recreating layout from elsewhere, right? This is going to get out of sync — if we're not going to create them together at this time (e.g., because the original layout is created in .xul files), we should add a bunch of "Keep in sync with…" comments.

dstillman (changes requested):

chrome/content/zotero/bindings/relatedbox.xml
mrtcode: > `zotero-pane` is the library pane?

Yes.

> This should probably just be part of `selectItem()`? Any code that selects items (e.g., a `zotero://select` link) presumably wants to select the library pane.

A good point. Fixed.

scss/components/_tagsBox.scss
mrtcode: Added. Basically `width: 0` and `height: 0` are need to make flexbox in XUL work more similar to regular flexbox. Otherwise child container forces its width/height to parent container by default, which never happens in regular HTML.

chrome/content/zotero/contextPane.js
mrtcode: Yes, removed.

chrome/content/zotero/contextPane.js
mrtcode: Yes, from itemPane.xul. Added a comment to keep things on both sides (itemPane.xul and contextPane.js) in sync.

The way how context pane works is a little bit hacky in many aspects. I expect that we extract item pane into a separate React container at some point and will be able to reuse it in context pane.

mrtcode:
Also, I see that Related item titles aren't updated because relatedbox is only looking for changes for the current item.
This is not a big problem for library view, but in PDF tab case it has to be closed and reopened to see the updated title.
So I guess we should change this?

https://github.com/zotero/zotero/blob/e4a957ed82cddd70c4a99eb28242337dd51c3e86/chrome/content/zotero/bindings/relatedbox.xml#L118-L128

PR #2083, opened 2021-06-07
Run JavaScript: Improve `await` regex

qqobb:
The current regex doesn't seem right. Most likely there's a typo, see "intended?" below.

Here are some variants:
```
(^|[^=([]\s*)await\s  //current
(^|[=([]\s*)await\s   //intended?
(^|[=([{!]\s*)await\s //improved
(^|\W)await\s         //simpler, suggested in this PR
```

Tests: https://regex101.com/r/eQYCIp/1


dstillman:
Thanks!

PR #2088, opened 2021-06-12
Blur active field in switchCreatorMode. Fix #2087

### LINKED ISSUE #2087: Using button to toggle single/dual creator wipes the value of the focused input field
https://forums.zotero.org/discussion/90272/changes-to-the-title-arent-saved-in-some-cases#latest

https://user-images.githubusercontent.com/214628/121773511-b51ea600-cb7c-11eb-8abb-81d323a0ad88.mov

tnajdek:
After this change `activeField` is blurred just before item is saved to preserve the value entered by the user, `switchCreatorMode()` has been converted to async to allow for this.

### DIFF
--- a/chrome/content/zotero/bindings/itembox.xml
+++ b/chrome/content/zotero/bindings/itembox.xml
@@ -1068,7 +1068,12 @@
 						fields.fieldMode = fieldMode;
 						this.modifyCreator(index, fields);
 						if (this.saveOnEdit) {
-							this.item.saveTx();
+							let activeField = this._dynamicFields.querySelector('textbox');
+							if (activeField !== null && activeField !== firstName && activeField !== lastName) {
+								this.blurOpenField();
+							} else {
+								this.item.saveTx();
+							}
 						}
 					}
 				]]>


chrome/content/zotero/bindings/itembox.xml
```diff
@@ -1068,9 +1069,14 @@
 						fields.fieldMode = fieldMode;
 						this.modifyCreator(index, fields);
 						if (this.saveOnEdit) {
-							this.item.saveTx();
+							let activeField = this._dynamicFields.querySelector('textbox');
+							if (activeField !== null && activeField !== firstName && activeField !== lastName) {
+								await this.blurOpenField();	
+							}
+							await this.item.saveTx();
```
dstillman, line 1076: I think this can be in an `else`, since if a save is necessary it would've been triggered by blurring. (I know we have `blurOpenField` followed by `save` elsewhere, but in those cases we don't necessarily check whether a field is open. Here we know that a field is open.)

tnajdek: Makes sense, thanks! This also means this function doesn't need to be `async`. I guess it works either way but I've changed it back to sync to keep the changes to minimum.

PR #2091, opened 2021-06-14
Add `cleanEdition` translator utility

AbeJellinek:
I've found this function quite useful for parsing relatively unstructured sites, and I think it would make sense to have it available globally.

### DIFF
--- a/chrome/content/zotero/xpcom/utilities.js
+++ b/chrome/content/zotero/xpcom/utilities.js
@@ -1989,7 +1989,54 @@ Zotero.Utilities = {
 	 * Provides unicode support and other additional features for regular expressions
 	 * See https://github.com/slevithan/xregexp for usage
 	 */
-	 "XRegExp": typeof XRegExp !== "undefined" ? XRegExp : null
+	"XRegExp": typeof XRegExp !== "undefined" ? XRegExp : null,
+
+	/**
+	 * Cleans an edition string ("second edition", "3e", "4."), converting
+	 * spelled-out ordinals and returning only the numeric part ("2", "3", "4").
+	 * 
+	 * If text is null or empty, or no edition can be found, returns the passed text.
+	 * If the resolved edition is "1" ("first edition", "1. Auflage"), returns null.
+	 * The Zotero.Item edition field should be left empty for first-edition items.
+	 * 
+	 * @param {string} text 
+	 * @param {boolean} allowOtherText If true, match even strings with extra content
+	 * ("3rd edition of the book", "there are 3 fish" -> "3"). More flexible, but can
+	 * cause false positives or lose information ("2nd edition with the author's signature" -> "2").
+	 * @return {string}
+	 */
+	"cleanEdition": function (text, allowOtherText) {
+		if (typeof text === "string" && text) {
+			const ordinals = {
+				first: "1",
+				second: "2",
+				third: "3",
+				fourth: "4",
+				fifth: "5",
+				sixth: "6",
+				seventh: "7",
+				eighth: "8",
+				ninth: "9",
+				tenth: "10"
+			};
+
+			const re = allowOtherText
+				? /(?:(?:([0-9]+)(?:st|nd|rd|th|er|re|d?e|d|ère|\.)?)|(first|second|third|fourth|fifth|sixth|seventh|eighth|ninth|tenth))\s*(?:ed?\.?|(?:e|é)d(ition)?|auflage|editie)?/i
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/xpcom/utilities.js
```diff
@@ -1989,7 +1989,54 @@ Zotero.Utilities = {
...
+	 * If the resolved edition is "1" ("first edition", "1. Auflage"), returns null.
+	 * The Zotero.Item edition field should be left empty for first-edition items.
+	 * 
+	 * @param {string} text 
+	 * @param {boolean} allowOtherText If true, match even strings with extra content
+	 * ("3rd edition of the book", "there are 3 fish" -> "3"). More flexible, but can
+	 * cause false positives or lose information ("2nd edition with the author's signature" -> "2").
+	 * @return {string}
+	 */
+	"cleanEdition": function (text, allowOtherText) {
```
dstillman, line 2008: Let's move this up next to the other `clean*` functions (and move the test up accordingly).

test/tests/utilitiesTest.js
```diff
@@ -585,4 +585,35 @@ describe("Zotero.Utilities", function() {
 			});
 		});
 	});
+
+
+	describe("cleanEdition", function() {
+		it('should clean valid editions (allowOtherText == false)', function() {
+			for (let [text, expected] of [
+				['2. Auflage', '2'],
+				['1. Auflage', null],
```
dstillman, line 594: Should this be an empty string instead? Either probably works, but technically if we're setting this directly do a field it should probably be an empty string.

chrome/content/zotero/xpcom/utilities.js
```diff
@@ -1989,7 +1989,54 @@ Zotero.Utilities = {
...
+				sixth: "6",
+				seventh: "7",
+				eighth: "8",
+				ninth: "9",
+				tenth: "10"
+			};
+
+			const re = allowOtherText
+				? /(?:(?:([0-9]+)(?:st|nd|rd|th|er|re|d?e|d|ère|\.)?)|(first|second|third|fourth|fifth|sixth|seventh|eighth|ninth|tenth))\s*(?:ed?\.?|(?:e|é)d(ition)?|auflage|editie)?/i
+				: /^(?:(?:([0-9]+)(?:st|nd|rd|th|er|re|d?e|d|ère|\.)?)|(first|second|third|fourth|fifth|sixth|seventh|eighth|ninth|tenth))\s*(?:ed?\.?|(?:e|é)d(ition)?|auflage|editie)?$/i;
```
dstillman, line 2025: Are we interested in localizing this? We technically have ordinals and edition strings in the CSL locales, which we could make available for use here with a little bit of effort:

https://github.com/citation-style-language/locales/blob/bd8d2dbc85713b192d426fb02749475df30f0d2c/locales-fr-FR.xml#L79-L89

https://github.com/citation-style-language/locales/blob/bd8d2dbc85713b192d426fb02749475df30f0d2c/locales-de-DE.xml#L46-L49

https://github.com/citation-style-language/locales/blob/master/locales-zh-CN.xml#L74-L84

AbeJellinek: I think that's a good idea.

Should it be something like the existing [`dateFormats.json`](https://github.com/zotero/zotero/blob/f1414731daa749c1de7b5946e9c08b571c710d1e/resource/schema/dateFormats.json) but for number formats?

AbeJellinek: If we make `cleanEdition` locale-specific, I think it should default to the user's locale but have an optional `language` parameter to override that guess.

dstillman: > Should it be something like the existing dateFormats.json but for number formats?

Yup. So I think we'd want a script in the zotero-build repo that, given the path to root of the zotero/zotero repo, generated that file.

> If we make cleanEdition locale-specific, I think it should default to the user's locale but have an optional language parameter to override that guess.

Generally we parse 1) English and 2) the user's locale. And then a language parameter would override both, I suppose.

dstillman: (We usually use Python for this sort of thing.)

AbeJellinek: Alright, I have a script: https://github.com/zotero/zotero-build/pull/9

dstillman:
This will need to be folded into the zotero/utilities submodule, I assume. And the instructions for that repo will need to be updated to include passing in cslLocaleStrings.json.

dstillman:
(And we'll have to figure out the best way to load this. For dateFormats.json we use `Zotero.Date.init(json)`, but there's no direct equivalent here.)

PR #2117, opened 2021-07-08
Scaffold: Add action to open test URL in internal browser

AbeJellinek:
The dropdown in the Browser tab is fine, but I find it more natural to work from the Testing tab. This makes that an option.


chrome/locale/en-US/scaffold/scaffold.dtd
```diff
@@ -74,3 +74,5 @@
 <!ENTITY scaffold.testing.edit.import			"Edit Import">
 <!ENTITY scaffold.testing.copyToClipboard			"Copy to Clipboard">
 <!ENTITY scaffold.testing.openUrl			"Open URL">
+<!ENTITY scaffold.testing.openUrl.externally			"In default browser">
+<!ENTITY scaffold.testing.openUrl.internally			"Here">
```
dstillman, line 78: `In Browser Tab` and `In External Browser`?

Issue #2025, opened 2021-04-04
Support additional keyboard shortcuts for switching tabs


dstillman:
At least on macOS, most apps support multiple shortcuts. E.g., Firefox and Chrome both support 1) `Cmd-Shift-[` / `]`, 2) `Ctrl-Tab` / `Ctrl-Shift-Tab`, and 3) `Ctrl-PgUp` / `PgDn` (`Ctrl-Fn-up-arrow` / `down-arrow`). Safari/Finder support 1 and 2. VS Code supports 1 and 2, with 2 showing an extra selection popup. Atom supports all three, plus a bonus `Cmd-Option-left` / `right` for good measure.

We should support 1–3.

https://forums.zotero.org/discussion/88835/zotero-beta-on-macos-tab-switching-should-use-common-os-shortcuts#latest

jryans:
It's also common in Firefox, Chrome, and other macOS apps for `Cmd-1` (`Ctrl-1`) through `Cmd-8` (`Ctrl-8`) to jump to tabs 1-8 respectively, and also for `Cmd-9` (`Ctrl-9`) to jump to the last tab no matter how many there are.

PR #2121, opened 2021-07-11
Add more tab navigation shortcuts

jryans:
This adds several more tab navigation shortcuts that users commonly expect to work across any program with tabs. With this PR, we now support all of the following:

* `Ctrl-PageDown` / `PageUp`
* `Cmd-Shift-[` / `]` (macOS only)
* `Ctrl-Tab` / `Ctrl-Shift-Tab`
* `Cmd` / `Ctrl-1` through `9`

Fixes https://github.com/zotero/zotero/issues/2025


dstillman:
This is great.

PR #2128, opened 2021-07-16
Add test for ADS Bibcode extraction

AbeJellinek:
Completes zotero/translators#2529 by adding support for ADS bibcodes to the Add Item(s) by Identifier box. I suppose we should also change the label in the UI (`zotero.lookup.description` in the DTDs), but I don't know how to go about doing that (since I'm assuming it involves Transifex).

### DIFF
--- a/chrome/locale/en-US/zotero/zotero.dtd
+++ b/chrome/locale/en-US/zotero/zotero.dtd
@@ -176,7 +176,7 @@
 <!ENTITY zotero.tagColorChooser.setColor       "Set Color">
 <!ENTITY zotero.tagColorChooser.removeColor    "Remove Color">
 
-<!ENTITY zotero.lookup.description				"Enter ISBNs, DOIs, PMIDs, arXiv IDs, or ADS bibcodes to add to your library:">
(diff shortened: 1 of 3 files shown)


dstillman:
Label is here:

https://github.com/zotero/zotero/blob/6a6e4fa92c722a2ac46109199425c7dbe37eac0c/chrome/locale/en-US/zotero/zotero.dtd#L179

And a related string for Create Parent here:

https://github.com/zotero/zotero/blob/6a6e4fa92c722a2ac46109199425c7dbe37eac0c/chrome/locale/en-US/zotero/zotero.properties#L1182

(Non-English strings get updated separately.)

(At some point we're going to have to change how we list the possible identifiers, but we might still be OK, barely.)

AbeJellinek:
(This test will fail until the two PRs referenced above are merged and the submodules are updated.)

dstillman:
Let's capitalize the 'b' in Bibcode. They don't get to make up a common noun.

dstillman:
Can merge with that change

AbeJellinek:
Frontline soldiers in the battle for the sanctity of the English language.

PR #2129, opened 2021-07-19
Download embedded images when needed, or show error message

mrtcode:
- Copies images when note is duplicated or item/collection is dragged to another library.
- Downloads images when note is opened (in future we should put a placeholder indicating whether specific image is loading or missing).
- Downloads images or shows error when note is being duplicated.
- Currently, It doesn't download images when dragging item or collection to another library and doesn't show an error if images don't exist. At least for the downloading part, it would probably make sense to implement that first for all attachments.
- It only copies embedded image attachments that currently exist, and doesn't look into note body if there are missing or not yet synced image attachment items.


chrome/content/zotero/zoteroPane.js
```diff
@@ -4563,6 +4573,13 @@ var ZoteroPane = new function()
 	}
 	
 	
+	this.displayMissingImagesMessage = function () {
+		var ps = Components.classes["@mozilla.org/embedcomp/prompt-service;1"]
+								.getService(Components.interfaces.nsIPromptService);
```
dstillman, line 4578: `Services.prompt`

chrome/content/zotero/zoteroPane.js
```diff
@@ -4563,6 +4573,13 @@ var ZoteroPane = new function()
...
+		ps.alert(null, "", Zotero.getString('pane.item.attachments.missingImages.text'));
```
dstillman, line 4579: Since we're showing a prompt anyway, maybe it should just be `confirmEx()` and we should ask whether to continue or cancel.

chrome/content/zotero/xpcom/data/notes.js
```diff
@@ -190,6 +190,77 @@ Zotero.Notes = new function() {
 		return doc.body.innerHTML;
 	};
 
+	/**
+	 * Download embedded images if they don't exist locally
+	 *
+	 * @param {Zotero.Item} item
+	 * @returns {Promise<boolean>}
+	 */
+	this.ensureEmbeddedImagesAvailable = async function (item) {
```
dstillman, line 199: Might as well be grammatical and go with `ensureEmbeddedImagesAreAvailable()

chrome/content/zotero/xpcom/data/notes.js
```diff
@@ -190,6 +190,77 @@ Zotero.Notes = new function() {
...
+		var attachments = Zotero.Items.get(item.getAttachments());
+		for (let attachment of attachments) {
+			let path = attachment.getFilePath();
+			if (!path || !await OS.File.exists(path)) {
```
dstillman, line 203: Just use `let path = await attachment.getFilePathAsync()`, which does an existence check

chrome/content/zotero/xpcom/data/notes.js
```diff
@@ -190,6 +190,77 @@ Zotero.Notes = new function() {
...
+		for (let attachment of attachments) {
+			let copiedAttachment = await Zotero.Attachments.copyEmbeddedImage({ attachment, note: toNote });
+			let node = doc.querySelector(`img[data-attachment-key="${attachment.key}"]`);
+			if (node) {
+				node.setAttribute('data-attachment-key', copiedAttachment.key);
+			}
+		}
+
+		note = doc.body.innerHTML;
+		note = note.trim();
```
dstillman, line 259: `setNote()` trims, so this shouldn't be necessary

chrome/content/zotero/zoteroPane.js
```diff
@@ -4563,6 +4573,13 @@ var ZoteroPane = new function()
 	}
 	
 	
+	this.displayMissingImagesMessage = function () {
```
dstillman, line 4576: `displayMissingNoteImageMessage` (+ "Note" and singular "Image", since this appears if there's a single missing image)

chrome/content/zotero/zoteroPane.js
mrtcode: It means we need to tweak the prompt message. Currently it's "Notes with missing images cannot be copied.", though confirming still copies them. How about "Some note images are missing and cannot be copied. Do you want to continue?"

mrtcode: Also, is there any specific reason to not just use `confirm()`?

dstillman: `confirmEx()` lets us have clearer button labels like "Copy Anyway". (I don't know how many different labels we'd need here for different situations. But even `confirmEx()` with `general.continue` would be better than the `OK` we get with `confirm()`.

mrtcode:
It should be ok now.

mrtcode:
This doesn't load images after they are downloaded, but I'm going to merge this anyway, because this doesn't make things worse and the image loading code rewrite will be come soon, which will involve changes on Zotero client and `note-editor`.

PR #2132, opened 2021-07-26
Translation modularization

adomasven:
Use the modularized zotero utilities and translation architecture code.

Changes were tested by running the test suite and manual testing with the client.

Note that modularized/external code is still tested with the main zotero test suite. I.e. utilities and translate code. A new test file `utilities_itemTest.js` was added.


dstillman:
Tests are failing

adomasven:
Fixed!

PR #2155, opened 2021-08-24
Add `disableFontSizeScaling` prop to HTML tree

dstillman:
Removes `defaultRowHeight` prop added in adb8aa39f in favor of a prop
that disables font size scaling. A non-default row height can still be
specified with `rowHeight`.

Most of our existing trees need to disable font size scaling, but the
idea is that pretty much everything _should_ scale with font size for
accessibility, and it's a limitation of the current prefs and other UIs
that they don't currently, so it's better to default to scaling and
gradually remove uses of this prop.

This fixes a bug where the HTML trees other than the collection and item
trees would have larger rows but without larger text when font size was
increased.

### DIFF
--- a/chrome/content/zotero/components/virtualized-table.jsx
+++ b/chrome/content/zotero/components/virtualized-table.jsx
@@ -294,6 +294,7 @@ class VirtualizedTable extends React.Component {
 		if (!props.disableFontSizeScaling) {
 			this._rowHeight *= Zotero.Prefs.get('fontSize');
 		}
+		// A bit less row spacing on macOS
 		if (Zotero.isMac && this._rowHeight >= (props.rowHeight || DEFAULT_ROW_HEIGHT)) {
 			this._rowHeight -= 2;
 		}
--- a/chrome/content/zotero/components/virtualized-table.jsx
+++ b/chrome/content/zotero/components/virtualized-table.jsx
@@ -293,11 +293,11 @@ class VirtualizedTable extends React.Component {
 		this._rowHeight = props.rowHeight || DEFAULT_ROW_HEIGHT;
 		if (!props.disableFontSizeScaling) {
 			this._rowHeight *= Zotero.Prefs.get('fontSize');
-			if (Zotero.isMac && this._rowHeight > props.rowHeight || DEFAULT_ROW_HEIGHT) {
-				this._rowHeight -= 2;
-			}
 		}
-			
+		if (Zotero.isMac && this._rowHeight >= (props.rowHeight || DEFAULT_ROW_HEIGHT)) {
+			this._rowHeight -= 2;
+		}
+		
 		this.selection = new TreeSelection(this);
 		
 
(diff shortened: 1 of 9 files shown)


chrome/content/zotero/components/virtualized-table.jsx
```diff
@@ -1062,10 +1063,12 @@ class VirtualizedTable extends React.Component {
 			Zotero.debug("Attempting to update virtualized-table font size with a prop-specified rowHeight."
 				+ "You should change the prop on the React component instead");
```
adomasven, line 1064: `updateFontSize()` is intended to be called only when `fontSize` pref is changed on trees that have font size scaling enabled, otherwise if a row height change is expected it should be done manually by changing the prop. So this is a warning for future ourselves to not expect anything to happen. But the if statement is no longer correct above and the logged message needs to be altered. 

This method also probably needs a comment for further clarification.

chrome/content/zotero/components/virtualized-table.jsx
```diff
@@ -290,11 +290,10 @@ class VirtualizedTable extends React.Component {
...
-		this._rowHeight = props.rowHeight;
-		if (!this._rowHeight) {
-			this._rowHeight = props.defaultRowHeight || DEFAULT_ROW_HEIGHT;
+		this._rowHeight = props.rowHeight || DEFAULT_ROW_HEIGHT;
+		if (!props.disableFontSizeScaling) {
 			this._rowHeight *= Zotero.Prefs.get('fontSize');
-			if (Zotero.isMac && this._rowHeight > (props.defaultRowHeight || DEFAULT_ROW_HEIGHT)) {
+			if (Zotero.isMac && this._rowHeight > props.rowHeight || DEFAULT_ROW_HEIGHT) {
 				this._rowHeight -= 2;
 			}
```
adomasven, line 298: In relation to #2157 this conditional was supposed to get the 18px font size for macOS and 20px for Windows and Linux, but only when font scaling is 1 or higher. I think this is always true now for every OS, since `DEFAULT_ROW_HEIGHT` is always true and `&&` is evaluated before `||`.

adomasven: (Although my original condition may have been wrong too. It should have been `>=` there for it to work correctly)

Rosmaninho:
If I want to specify a specific row height in the collection or item trees in a userChrome.css file how should I do it?

.virtualized-table .rowHeight { height: 30px !important;} could work?

dstillman:
You don't — it's just controlled by the main font size pref, which is configurable from the View menu.

Rosmaninho:
In current 5.0.96 I can specifiy it with: 
treechildren::-moz-tree-row { height: 30px !important; }

I would be thankful if you could consider exposing such a parameter because I prefer the default font size but increased spacing between items.
Font size and row spacing are two very distinct parameters, so it would be helpful if you could maintain that functionality in the future release version. The spacing of items in the Windows version is too dense for some.

Rosmaninho:
I really would like to ask that instead of removing this property you could maintain it or provide a way to alter density of items shown on the collection and item trees please.

dstillman:
We didn't remove a property — you were relying on a Mozilla internal, and the tree is no longer based on Mozilla technologies. There's probably still a way to control this via CSS, but we don't provide support for userChrome.css customizations, so you'd be on your own for that. (You can view the innerHTML for the tree from the Run JavaScript window.)

If you'd like to ask for an exposed density setting, that's fine, but that should go to the forums, not the dev list.

Rosmaninho:
I asked in the Google group for that. Maybe it's not the best place? The forum is a mess because there's no sub-forum for technical requests but I'll post it there. 

Thank you for the reply! :)
I'm adapting my Dark theme to work with 5.0.97 and this is the only thing that I wasn't able to fix.

Congratulations for the PDF reader and the Tabbed UI, by the way. It's looking great.

dstillman:
It's not a technical request — the feature request would be for a density setting in the UI. That's why I'm saying it belongs in the forums and not the dev list. Again, you're on your own for figuring out CSS based on Zotero's internals.

Rosmaninho:
So there's no pointers you can give on how to adapt this?

Rosmaninho:
Ok, I added such a request to the forums.

Rosmaninho:
When I change
const DEFAULT_ROW_HEIGHT = 30; // px

from virtualized-table.js in zotero\chrome\content\zotero\components\ I manage to change the row height. Now, how to reach this via css... 

Do you have any idea?

PR #2176, opened 2021-08-30
HTML Tree: make find as you type a virtualized table functionality

### LINKED ISSUE #2168: HTML tree: Find as you type doesn't take sort column into account
https://github.com/wshanks/Zutilo/issues/203#issuecomment-907712042

"in the items list, "search as you type" seems to apply to the Title column regardless of the sort column"

### LINKED ISSUE #2169: HTML tree: Find as you type should apply to all trees
FAYT should be a feature of the tree itself, not of the collection and item trees specifically

adomasven:
Closes #2168, closes #2169.
Adds find-as-you-type to locate, preference style and export
formats managers.

To enable find as you type you need to specify the getRowString(index) prop
on the VirtualizedTable. See prop comment for more info.

### DIFF
--- a/chrome/content/zotero/collectionTree.jsx
+++ b/chrome/content/zotero/collectionTree.jsx
@@ -35,7 +35,6 @@ const { getDragTargetOrient } = require('components/utils');
 const { Cc, Ci, Cu } = require('chrome');
 
 const CHILD_INDENT = 15;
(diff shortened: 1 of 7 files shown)


dstillman:
1) Missing from libraries-to-sync tree
2) It throws when the prop is missing, instead of doing nothing

adomasven:
Updated

PR #2195, opened 2021-09-05
Correct Lookup Engine description

AbeJellinek:
The Locate dropdown isn't in the Info pane, and the text was a bit more verbose than it needed to be.

### DIFF
--- a/chrome/content/zotero/locateManager.xul
+++ b/chrome/content/zotero/locateManager.xul
@@ -63,7 +63,7 @@ To add a new preference:
 		<groupbox flex="1">
 			<caption label="&zotero.preferences.locate.locateEngineManager;"/>
 			
-			<label id="locateEngineDescription" width="45em" style="font-size: 10px"/>
+			<label id="locate-engine-description" width="45em" style="font-size: 12px"/>
 			
 			<separator class="thin"/>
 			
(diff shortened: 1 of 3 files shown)


dstillman:
While we're changing this, how about some other tweaks:

> Lookup engines help you find resources in your %S library on the web. Engines enabled in this list will appear in the Locate drop-down in the toolbar.

And then pass `Zotero.appName` in the `getString()` called.

AbeJellinek:
OK, see latest commit. I made a couple assumptions about strings/localization here:

1. I was getting nothing but errors trying to parameterize DTD strings, and it looks like all parameterized strings in the codebase are in `.properties` files, so I moved the relevant string to `zotero.properties`.
2. It doesn't appear that there's any way to pass parameters to `getString` with the normal XUL string entity syntax, so I now fill the label in using inline JavaScript (based roughly on [the approach](https://github.com/zotero/zotero/blob/f0a8c9ada38bae33593f331b36384d900e7f4d63/chrome/content/zotero/about.xul#L43-L49) in `about.xul`).

Is all of that correct? First time really working with strings in the client, so I want to be sure I'm not missing a better approach.

dstillman:
Oh, sorry, I missed that this was a .dtd — parameters indeed not possible.

PR #2208, opened 2021-09-20
Create SECURITY.md

zidingz:
A simple instruction for security researchers.

(PR is to verify write-access!)


dstillman:
Thanks, but I don't want to clutter the repo with this. Zotero is a large ecosystem with software that spans many repos. Genuine reports will come from people who are familiar with Zotero actually using the software, and those people are capable of finding our contact info.

PR #2214, opened 2021-09-24
Markdown Note Export

mrtcode:
This is still WIP. Here are the things to think about:
1) Note QuickCopy is named everywhere as `noteQuickCopy`, while item QuickCopy remains `quickCopy`.
2) Added a new pref called `export.noteQuickCopy.setting` which reuses the same unserialization logic, although only supports JSON (while regular QuickCopy can use "bibliography=http://.." format).
3) Currently no additional configuration options will be added. I.e. no "Include Annotation Links" (they will be included by default). Let's see if there will be a demand for more options.
4) Currently note QuickCopy is only triggered when only one note item is selected. In all other cases it behaves like before. Is that right?
5) Translator (currently only Markdown) exclusion from item formats and inclusion into note formats is implemented by hardcoding translator ID. Would it make sense to add a flag to translator `configOptions` to allow to filter it?
6) We said we want to move site-specific setting into a separate dialog. How exactly it should look like?
7) What should be the default option for note format? I guess it should just set both `text/plain` and `text/html` and then the target textbox will pick the right one. But how this default option should be named if so? Or maybe we want plain text and HTML options to be separate?
8) `export.quickCopy.dragLimit` probably doesn't make sense for notes?

No need to try out this PR yet, because only drop-down is visible now:

Screenshot from 2021-09-24 17-39-31


dstillman:
> Currently note QuickCopy is only triggered when only one note item is selected. In all other cases it behaves like before. Is that right?

No, if multiple notes are selected, it should concatenate them. We already do this for drag-and-drop with a line of "---" between them, which will [create a horizontal rule](https://daringfireball.net/projects/markdown/syntax#hr) in Markdown.

> Translator (currently only Markdown) exclusion from item formats and inclusion into note formats is implemented by hardcoding translator ID. Would it make sense to add a flag to translator `configOptions` to allow to filter it?

Yeah, I suppose so. Otherwise it'd be a problem if we ever pushed a new note format.

> What should be the default option for note format? I guess it should just set both `text/plain` and `text/html` and then the target textbox will pick the right one. But how this default option should be named if so? Or maybe we want plain text and HTML options to be separate?

Markdown is a plain-text format that includes raw HTML markup, so it only makes sense to copy it as text/plain, and "Copy as HTML" shouldn't be settable.

But we should have a format — the default one, in fact — that's equivalent to what we do now for drag-and-drop, which copies both plain-text and HTML, and it should also be possible to get raw HTML out of notes. So, two options:

1) We have a default format called "Text" or "Plain Text and Rich Text" or "Rich Text" that puts plain-text and HTML on the clipboard, and another called "HTML" that puts raw HTML on the clipboard as text. Since these are translators, the "HTML" one could hopefully just call the default one and then return the HTML output as text.

2) We have the same default format, and it has a config option that enables the "Copy as HTML" option.

I'd be inclined to go with the first option, for a cleaner and clearer UI. I can't think of other formats we'd create where "Copy as HTML" would make sense, and I suspect most third-party variants would want raw HTML only, because you wouldn't necessarily expect anything beyond basic formatting to survive pasting into an HTML-accepting editor. If someone did want to make a format that worked both ways (e.g., with different basic formatting), they could just create a second raw-HTML translator.

With two translators, if we added any additional options to the default translator, we'd need to mirror those in the HTML version and forward them to the first translator, but that's OK.

> export.quickCopy.dragLimit probably doesn't make sense for notes?

Why not? Just because it's harder to accidentally select a thousand notes than a thousand regular items?

mrtcode:
> But we should have a format — the default one, in fact — that's equivalent to what we do now for drag-and-drop, which copies both plain-text and HTML, and it should also be possible to get raw HTML out of notes. So, two options:

@dstillman Are you saying we should ditch the existing note text/html export logic and reimplement as separate translators? That code is doing many things.

https://github.com/zotero/zotero/blob/7814efcfe2e097e8aa8eae434d45792beabf3102/chrome/content/zotero/xpcom/quickCopy.js#L280-L419

dstillman:
Well, I'm saying there needs to be an option to make that behavior the default Quick Copy behavior for notes, one way or another. Maybe it doesn't all work within a translator — it looks like at least the existing preferences would have to change, though I'm not sure anyone even knows those exist — but that would be the easiest way to offer the same output via the API, on iOS, etc. Other than the prefs, it seems like `nsIDocumentEncoder` would be the main problem…

mrtcode:
Wrapped paragraphs output provided by `nsIDocumentEncoder` is nice, but I'm not sure how necessary it is, and we probably want to reduce our dependence on internal Firefox stuff. Now we have to figure out which [prefs](https://www.zotero.org/support/preferences/hidden_preferences#citation_quickcopy_settings) are still relevant if any. Additional quotes and indentation for blockquotes probably isn't import as well. I'm not sure about `export.quickCopy.compatibility.word`.

dstillman:
I think it's safe to say that the number of people who use these prefs is somewhere between vanishingly small and zero, and besides, if we wanted to preserve any customizability (e.g., for better compatibility with a given program), we should just add translator options for those that could get exposed in the UI. (I have no recollection of what exactly the Word compatibility pref did, except that a decade or so ago I couldn't manage to get the same HTML to paste consistently in various programs.)

mrtcode:
A few more things to discuss:
1) Do we want different HTML outputs for the raw HTML output and for the clipboard? Currently the only difference is that for the clipboard an additional `-----` block is inserted between notes. We can consider `hr` as well. Also what is the typical use case for the raw HTML output?
2) In theory we could embed images as base64 URLs, but I guess that won't work in translator and probably not necessary as well.
3) For the rich text output citations and annotations will have links that navigate to Zotero item or annotation, but for plain text we probably don't want to clutter the output with links right?
4) In theory, dragging note item should have the same output as selecting the whole note content right? But we won't be able to match those outputs because copying from `note-editor` can do more things than from translator (i.e. embed images).
5) Dragging HTML from Zotero (and Firefox, even the latest) is completely broken on Linux (Ubuntu 20). It ends up dropped like a raw HTML with some random question marks.
6) Maybe it would make sense to have one translator that exports plain text, HTML (and maybe HTML for raw output if it has to be different), and then to make life easier for everyone to have not only a separate intermediate HTML translator but also a separate plain text translator to export plain text without additional logic to parse JSON.

dstillman:
> Do we want different HTML outputs for the raw HTML output and for the clipboard?

I don't think there should be any difference.

> In theory we could embed images as base64 URLs, but I guess that won't work in translator and probably not necessary as well.

I think text is fine for now. At some point we can explore how different potential target tools actually handle images.

> For the rich text output citations and annotations will have links that navigate to Zotero item or annotation, but for plain text we probably don't want to clutter the output with links right?

I think for now the `zotero://open-pdf` links should be limited to Markdown, and ideally we'll make that configurable later.

For the plain-text output, we certainly don't need them, and I don't think we should have them for HTML output either, at least by default, since I think that's much less likely to be for local use.

> In theory, dragging note item should have the same output as selecting the whole note content right?

Yes, in theory.

> Dragging HTML from Zotero (and Firefox, even the latest) is completely broken on Linux

Huh. Oh well.

> separate intermediate HTML translator but also a separate plain text translator to export plain text

I don't really see the point of that. It would need manual overrides either way — to hard-code either hiding of some translators or processing of JSON. A translator with just `return data.html;` seems dumb. I think we can just have text/HTML, and then, contrary to what I say above, hard-code an extra entry in the Quick Copy prefs for an HTML option, that uses the same underlying translator but takes just the HTML and treats it as text.

mrtcode:
Ok, now we can try this out. Note Text and Note HTML translators are [here](https://github.com/zotero/translators/pull/2749), and Note Markdown translator has to be build [here](https://github.com/zotero/markdown-translator-builder). Or just use the zipped ones [translators.zip](https://github.com/zotero/zotero/files/7564033/translators.zip).

O few more things to consider:
1) HTML output currently keeps all the annotation and citation data in attributes and note `div` container. Do we want to strip that data?
2) Those three translators have `"noteTranslator": true` config option that allows to filter them. Do we want to rename it?
3) Plain text output is now inferior comparing to how it looked before, because `nsIDocumentEncoder` was doing lots of formatting for that plain text. E.g. lists had numbers, bold had asterisks and underline had underscores. But now we have Markdown output which is even better. Although if there is still a need for advanced plain text formatting we can improve Note Text translator.
4) How about exporting attachment notes? Don't allow when dragging, but allow over export dialog?
5) We said we want to move "Site-Specific Settings" into a separate dialog. Do we want to do that now and if so how exactly that should look like?

dstillman:
> 1. HTML output currently keeps all the annotation and citation data in attributes and note div container. Do we want to strip that data?

I think we can strip, as long as we have various formatting options (annotation colors, links back to Zotero, etc.). For now, we can strip and plan to add those options later.

> 2. Those three translators have "noteTranslator": true config option that allows to filter them. Do we want to rename it?

The "Translator" is a little redundant, but `note: true` is probably too vague. I don't have a better suggestion.

> 3. Plain text output is now inferior comparing to how it looked before, because nsIDocumentEncoder was doing lots of formatting for that plain text.

I think we'll need to keep the old non-translator code for now and just hack it into Quick Copy, and not do a Note Text translator for now. That will mean that plain text won't be available from the API or translation-server, but it's not worth making plain-text output from the client much worse for that. And as you say, Markdown is available.

> 4. How about exporting attachment notes? Don't allow when dragging, but allow over export dialog?

Yeah, allow via export dialog and keyboard-shortcut Quick Copy, but not dragging (since any attempt to include the text would be likely to mess up file copying (e.g., to email clients)).

> 5. We said we want to move "Site-Specific Settings" into a separate dialog. Do we want to do that now and if so how exactly that should look like?

I think we'd basically just move what's there into a dialog. But we'd need to HTMLify it. For now, I think we can just shrink the list in the main prefpane way down, to like three scrollable lines.

mrtcode:
> > 1. Plain text output is now inferior comparing to how it looked before, because nsIDocumentEncoder was doing lots of formatting for that plain text.
> 
> I think we'll need to keep the old non-translator code for now and just hack it into Quick Copy, and not do a Note Text translator for now. That will mean that plain text won't be available from the API or translation-server, but it's not worth making plain-text output from the client much worse for that. And as you say, Markdown is available.
> 
> >

Are you saying we shouldn't have Note Text translator at all for now? If so then does it make sense to have Note HTML translator, if we are keeping the original logic in the client? How about note export on iOS?

dstillman:
> Are you saying we shouldn't have Note Text translator at all for now?

Yeah — I just don't think we should degrade plain-text copying that much.

An HTML translator still makes sense because that'll be useful from the API, along with Markdown. I'm also guessing iOS apps that don't want Markdown would mostly want HTML. The plain-text output on the desktop is mainly there as a fallback, but for better or for worse (for worse), it's hard to paste plain text into anything on iOS. (I'm not sure what happens if you copy HTML only on iOS and then paste somewhere that doesn't use rich text, like Messages. Not sure if there's a built-in equivalent to `nsIDocumentEncoder`. In any case, we can revisit when we add this on iOS.)

mrtcode:
Ok, I made the fixes that more or less represent what we've discussed:

1) It still uses Note Text translator id to store it in prefs and assumes it's a regular translator even though it is manually inserted into translators list. 
2) In future we might want to add the actual Note Text translator, although the hard-coded translator would override it, until the logic is removed.
3) It has hard-coded "Note Text" translator label.
4) Export dialog, at the moment, doesn't have "Note Text" translator, but we can add it if needed.
5) QuickCopy settings are still removed.

But I'm not entirely convinced we need all the things that `nsIDocumentEncoder` provides, because for the plain text output I personally wouldn't want additional underscores, asterisks and paragraph wrapping on some arbitrary width, though list numbers/bullets is a good thing. Therefore if only a few corrections are needed for the separate Note Text translator, I would reconsider to just improve and use it separately. That would allow to have a much cleaner implementation.

dstillman:
> But I'm not entirely convinced we need all the things that nsIDocumentEncoder provides, because for the plain text output I personally wouldn't want additional underscores, asterisks and paragraph wrapping on some arbitrary width, though list numbers/bullets is a good thing.

Hmm. I agree that line wrapping is undesirable, and that underscores and asterisks are a little dumb when Markdown is an option.

Are there things our Markdown output will leave as raw HTML, in the course of normal usage? Maybe just copying Markdown as the text for Text mode makes sense, but it's a little weirder if we're actually leaving chunks of HTML in the output…

mrtcode:
> Hmm. I agree that line wrapping is undesirable, and that underscores and asterisks are a little dumb when Markdown is an option.

It would be a bit strange to export Markdown for the plain text output, but `nsIDocumentEncoder` output is already strange, therefore it's unlikely that Markdown would make things worse. Therefore yes we can consider that.

Turndown (the underlying Markdown library) by default doesn't keep HTML except for tables if they can't be transformed (but  even if they can, the produced tables are poor, although I hope that gets improved in one or another way in future).

dstillman:
I'm not sure it's even all that weird, particularly if it's stripping most HTML. Markdown is designed to be readable above all else, after all. Maybe links are slightly weird to the uninitiated (though really just the extra square brackets), but I think most other things are pretty clear. And nsIDocumentEncoder throws our links, right? Probably worth switching for that reason alone.

mrtcode:
> And `nsIDocumentEncoder` throws out links, right? Probably worth switching for that reason alone.

It keeps them i.e. `test <http://test.com>`. But there is no point to have another syntax that is so similar to Markdown.

Ok, then I'll change QuickCopy to use Markdown translator for plain text output.

Export dialog will have only "Note HTML" and "Note Markdown".

mrtcode:
It should be ok now. Just a few last things:

1) For the sake of consistency, it reserves `a45eca67-1ee8-45e5-b4c6-23fb8a852873` translator ID for the virtual Note Text translator. It's set in note `extensions.zotero.export.noteQuickCopy.setting` by default.
2) "should copy note content" test is removed, because we either have to test the whole `onDragStart` [function](https://github.com/zotero/zotero/blob/4051ebaf377e33eac3acbf138c9c4bd37182deea/chrome/content/zotero/itemTree.jsx#L1862) or otherwise it would just be testing translators, if we would call `Zotero.QuickCopy.getContentFromItems()` like it was before.

chrome/content/zotero/itemTree.jsx
```diff
@@ -1934,22 +1934,46 @@ var ItemTree = class ItemTree extends LibraryTree {
...
+		if (!items.some(item => !item.isNote())) {
+			format = Zotero.QuickCopy.getNoteFormat();
 		}
 
+		Zotero.debug("Dragging with format " + format);
 		format = Zotero.QuickCopy.unserializeSetting(format);
 		try {
 			if (format.mode == 'export') {
-				Zotero.QuickCopy.getContentFromItems(items, format, exportCallback);
+				// If exporting with Note Text translator, call Note HTML and Note Markdown translators instead
```
dstillman, line 1946: "Note HTML and Note Markdown" → "Note Markdown and Note HTML"

chrome/content/zotero/preferences/preferences_export.jsx
```diff
@@ -82,6 +85,65 @@ Zotero_Preferences.Export = {
...
+		
+		// Add/override Note Text translator
+		translators = translators.filter(x => x.translatorID != 'a45eca67-1ee8-45e5-b4c6-23fb8a852873');
+		translators.push({
+			translatorID: 'a45eca67-1ee8-45e5-b4c6-23fb8a852873',
+			label: 'Note Text',
+			configOptions: {
+				noteTranslator: true
+			}
+		});
```
dstillman, line 122: Now that Markdown is the text format, it doesn't really make sense to show that separately, so I think the visible options for Quick Copy should be "Markdown + Rich Text" and "HTML", in that order. No "Note" prefix. The "Rich Text" part should probably be localized.

The export dialog should have "Markdown" and "HTML", with no "Text". I don't think we need the "Note" prefixes there (and then we don't need to bother localizing them).

chrome/content/zotero/xpcom/quickCopy.js
```diff
@@ -458,6 +315,19 @@ Zotero.QuickCopy = new function() {
 	});
 	
 	
+	var _loadNoteOutputFormat = async function () {
+		var format = Zotero.Prefs.get("export.noteQuickCopy.setting");
+		format = Zotero.QuickCopy.unserializeSetting(format);
+		// If Note Text translator is selected, preload Note HTML and Note Markdown translators
```
dstillman, line 321: "Note HTML and Note Markdown" → "Note Markdown and Note HTML"

dstillman (commented):
It looks like we're not using the note format for Cmd-Shift-C. When only notes are selected, we should — dragging and using the keyboard shortcuts should be identical. Technically you could have Zotero RDF (and I think only Zotero RDF) as the export format for notes and get output, but that's not necessary, and I don't think any other export formats would even work, since they're designed for metadata items with child notes, not standalone notes.

On that subject, we should probably also exclude all non-`noteTranslator` translators other than Zotero RDF from the export dialog when only notes are selected.

chrome/content/zotero/preferences/preferences_export.jsx
mrtcode: Do you mean we should remove "Note" prefix from the actual translators?

dstillman: No, in the display.

chrome/content/zotero/itemTree.jsx
```diff
@@ -1934,22 +1934,46 @@ var ItemTree = class ItemTree extends LibraryTree {
...
+				if (format.id === 'a45eca67-1ee8-45e5-b4c6-23fb8a852873') {
+					let markdownFormat = { mode: 'export', id: '154c2785-ec83-4c27-8a8a-d27b3a2eded1' };
+					Zotero.QuickCopy.getContentFromItems(items, markdownFormat, (obj, worked) => {
+						if (!worked) {
+							Zotero.log(Zotero.getString('fileInterface.exportError'), 'warning');
+							return;
+						}
+						// Existing `items` can't be reused in another translator
```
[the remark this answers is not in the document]
mrtcode: I think it's shifting items [here](https://github.com/zotero/translate/blob/c909a38d37bc2e0c68ada2bd75ca0a3af2b31f4e/src/translation/translate_item.js#L89), but I can add `items: items.slice()` [here](https://github.com/zotero/translate/blob/a9308c0e8632846ca2dc069a1b72db0a33f99ca6/src/translation/translate.js#L2424), if we're sure this wasn't intentional.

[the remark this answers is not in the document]
dstillman: Probably wasn't intentional, though maybe it saves a little memory for large exports. Could also just clone before passing to translation.

mrtcode:
> It looks like we're not using the note format for Cmd-Shift-C. When only notes are selected, we should — dragging and using the keyboard shortcuts should be identical.

Notes and attachments, unlike regular items, don't have Edit menu option for Cmd-Shift-C. Are you saying we should only allow to do QuickCopy with keyboard shortcut but not over the menu?

chrome/content/zotero/preferences/preferences_export.jsx
mrtcode: I agree that not having a separate Markdown option is simpler and better, although in some cases user might want Markdown even when dragging into a text box that supports Rich Text. But probably no need to worry about that.

dstillman:
Notes didn't have the menu option before because it didn't make sense before. But now there should be a "Copy Note" option when only notes are selected, and Cmd-Shift-C should work.

chrome/content/zotero/preferences/preferences_export.jsx
dstillman: Yeah that occurred to me but I don't think it's necessary. Seems pretty unlikely, and anyone in that situation can probably figure out how to get the Markdown.

mrtcode:
All discussed things should be fixed now. A few more things:

1) It's a bit annoying now when export dialog overrides the last translator used for regular items. So we either have to not save the last translator when exporting notes, or there should be two prefs.
2) In export dialog, attachments currently have the same translators as notes.

Updated translators: [translators.zip](https://github.com/zotero/zotero/files/7595291/translators.zip).

dstillman:
>  So we either have to not save the last translator when exporting notes, or there should be two prefs.

Yeah, we should use a different pref for the last note translator.

> In export dialog, attachments currently have the same translators as notes.

In all cases, or only if the attachments have embedded notes? There's not really much of a use case for exporting attachments directly without their parent items, so I'm not sure it matters that the non-note translators aren't available, but maybe we should disable the export option altogether if no items have embedded notes?

mrtcode:
> In all cases, or only if the attachments have embedded notes?

In all cases.

> maybe we should disable the export option altogether if no items have embedded notes?

Disable in context and edit menus but still leave the options grayish right? Maybe. I guess we want the same behaviour for note items as well if all of them are empty?

dstillman:
> Disable in context and edit menus but still leave the options grayish right?

Yeah.

> I guess we want the same behaviour for note items as well if all of them are empty?

Sure.

mrtcode:
Fixed. Two new prefs were introduced: `export.lastNoteTranslator` and `export.noteTranslatorSettings`.

dstillman:
- If we're exporting an HTML file, should it be a proper HTML document? For Quick Copy, no, but if we're saving an .html file to disk, an excerpt doesn't seem totally correct. The main reason someone would want an excerpt would be for pasting into an existing document, and if you want to do that you should just use Quick Copy.
- When generating output for multiple notes, the output should match the order in the items list. (It doesn't really matter for regular item export, which is probably why we don't do this now.)

Looking good otherwise.

mrtcode:
> * If we're exporting an HTML file, should it be a proper HTML document? For Quick Copy, no, but if we're saving an .html file to disk, an excerpt doesn't seem totally correct. The main reason someone would want an excerpt would be for pasting into an existing document, and if you want to do that you should just use Quick Copy.

Makes sense. But I guess we should return a proper HTML document from translator, because it will be used outside of Zotero client where it also makes sense to have a proper HTML document. For QuickCopy we would add a hard-coded logic to extract the body content.

dstillman:
Yeah, that makes sense. And any translator consumer that wants the excerpt can easily do the same.

mrtcode:
[This](https://github.com/zotero/zotero/blob/7814efcfe2e097e8aa8eae434d45792beabf3102/chrome/content/zotero/xpcom/translation/translate_item.js#L1015) changes items order when setting them for translator. What is the purpose of it? Should we remove it or maybe just add a parameter to skip this sorting?

Do we want any title for Note HTML output? If so, it will need to be localized..

dstillman:
See https://github.com/zotero/zotero/commit/32c6e1edb35c7c5fbac6cd900c1cb71ac52ee896 and the linked forums thread for an explanation. We want it for regular item export, but we can skip it for note export.

> Do we want any title for Note HTML output? If so, it will need to be localized..

I don't think we need one. It's optional according to https://html.spec.whatwg.org/#the-title-element

mrtcode:
Should we pass a parameter i.e. `skipSort` to `setItems`, which also needs `translate` module modifications, or instead just check if all items are notes just before this [line](https://github.com/zotero/zotero/blob/7814efcfe2e097e8aa8eae434d45792beabf3102/chrome/content/zotero/xpcom/translation/translate_item.js#L1015)?

dstillman:
We can try just checking before that line. I think the only way it would be a problem is if you were exporting 1) only notes 2) to Zotero RDF and 3) really needed the items to stay in a consistent order which 4) I'm not sure Zotero RDF even honors.

mrtcode:
Ok, fixed now.

Updated translator: [Note HTML.zip](https://github.com/zotero/zotero/files/7618989/Note.HTML.zip)

dstillman:
Any easy translator options we can add here to Markdown and/or HTML so that the dialog doesn't look quite so silly?

Screen Shot 2021-11-30 at 6 05 15 AM

mrtcode:
I can't come up with anything useful and easy to do for now. There will probably be an option for images or annotations but not clear how it would work yet.

dstillman:
There'll certainly be an option not to include zotero:// links in Markdown export.

mrtcode:
Ok, I can add it. "Include Links to Zotero"?

dstillman:
"Include Zotero Links"?

mrtcode:
Fixed. Although, in theory, I would like to have this option for dragged/copied notes.

dstillman:
Should actually use a placeholder in that string and `Zotero.appName`. And I guess call it `includeAppLinks`.

And the `zotero://` should be based on `ZOTERO_CONFIG.ID`.

dstillman:
Oh, but the scheme is in the Markdown translator, so we'd have to pass that in somehow.

mrtcode:
Is `ZOTERO_CONFIG` available in all environments? I.e. translation-server?

mrtcode:
On the client I already can access `ZOTERO_CONFIG` from a translator. And on translation-server we can just use hard-coded `zotero://` if `ZOTERO_CONFIG` doesn't exist.

dstillman:
We don't really want to access `ZOTERO_CONFIG` if there's any way to avoid it. The sandbox is more a sandbox of the mind these days, but we still try to stick to the intended API.

mrtcode:
Any suggestion how we could pass it? I don't see any non-super-hacky way to do that.

mrtcode:
Well maybe replace in the HTML after translator returns the result.

dstillman:
So return `zotero://`, and sub in `ZOTERO_CONFIG.ID` (which in Zotero itself will be a no-op)? That works.

mrtcode:
> Should actually use a placeholder in that string and `Zotero.appName`. And I guess call it `includeAppLinks`.

Wait, but we can't easily pass `Zotero.appName` to that placeholder as well.

dstillman:
Why not?

mrtcode:
Because `Zotero.appName` has to be somehow included [here](https://github.com/zotero/zotero/blob/712aae0a813dce98f440117aff7734de110dfcb3/chrome/content/zotero/exportOptions.js#L76).

dstillman:
We can just add a special hard-coded case for that key.

mrtcode:
Ok, should be good now. Thanks for the patience explaining all this.

[Note Markdown.zip](https://github.com/zotero/zotero/files/7632921/Note.Markdown.zip)

mrtcode:
Wait, I have to fix one thing.

mrtcode:
It's ready now.

chrome/content/zotero/fileInterface.js
```diff
@@ -122,33 +122,47 @@ Zotero_File_Exporter.prototype.save = async function () {
...
+			return;
+		}
+
+		// For Note Markdown translator replace zotero:// URI scheme,
+		// if the current app is not Zotero
+		if (io.selectedTranslator.translatorID == '154c2785-ec83-4c27-8a8a-d27b3a2eded1'
+			&& ZOTERO_CONFIG.ID != 'zotero') {
+			let text = obj.string;
+			text = text.replace(/zotero:\/\//g, ZOTERO_CONFIG.ID + '://');
+			Zotero.File.putContentsAsync(fp.file, text);
```
dstillman, line 144: Hmm, is there not a way to do this without two separate code paths for saving the file (`setLocation()` and `putContentsAsync`)?

If not, it seems like `_exportDone()` should be `async`, this should be `await`, and this should run before we close the progress dialog if `worked` is true, since the save could theoretically take some time.

But it would be better not to have to have to handle this in two different places.

dstillman: Oh, and actually, is the progress window already saying the export is done by this point, and we're just closing it here? So all the more reason to do this substitution during the actual processing if we can.

mrtcode: It should be fixed now.

> Hmm, is there not a way to do this without two separate code paths for saving the file (`setLocation()` and `putContentsAsync`)?

`setLocation` uses streaming (if I'm right) which is useful for other translators, and for the replacement we need to take over and replace the whole output, therefore, no, we probably can't avoid two code paths.

> Oh, and actually, is the progress window already saying the export is done by this point, and we're just closing it here?

Popup closing was already here before. So, I guess, the export should be done at that point. Although when using RDF translator and including PDFs and annotations, the dialog closes much earlier that it actually finishes (pdf-worker is doing export) and we aren't waiting for [that](https://github.com/zotero/zotero/blob/3f37786ffd4fc1c977d4b16eb841f6619680372a/chrome/content/zotero/xpcom/translation/translate_item.js#L1223) I guess.

mrtcode: Just one thing - I still miss a way to export multiple individual notes.

chrome/content/zotero/fileInterface.js
```diff
@@ -89,33 +122,49 @@ Zotero_File_Exporter.prototype.save = async function () {
...
+				Zotero.getString('fileInterface.exportError')
+			);
+			Zotero_File_Interface.Progress.close();
+			return;
+		}
+
+		// For Note Markdown translator replace zotero:// URI scheme,
+		// if the current app is not Zotero
+		if (io.selectedTranslator.translatorID == '154c2785-ec83-4c27-8a8a-d27b3a2eded1'
+			&& ZOTERO_CONFIG.ID != 'zotero') {
```
dstillman, line 139: Nit: Indent additional tab stop (here and below)

chrome/content/zotero/fileInterface.js
dstillman: OK, I have a not-totally-hacky way to do this. Will commit in a bit.

dstillman:
OK, merging with this if no objections

dstillman:
Oh, missing the part that actually calls this, which is in `rest()` in translate.js, below the `init()` call:

```javascript
			// For the Note Markdown translator, replace the zotero:// URI scheme in the output if
			// not the official Zotero app
			if (this.translator.translatorID == '154c2785-ec83-4c27-8a8a-d27b3a2eded1'
					&& ZOTERO_CONFIG.ID != 'zotero') {
				this._io.setDataProcessor((data) => {
					return data.replace(/zotero:\/\//g, ZOTERO_CONFIG.ID + '://');
				});
			}
```

mrtcode:
Yes, that's probably ok. I'll test once again when everything is merged.

dstillman:
-

PR #2229, opened 2021-10-23
Add tests for async web translation

AbeJellinek:
See zotero/translate#4.


dstillman:
OK, can add the updated translate submodule

PR #2234, opened 2021-11-02
Remove code to patch after earlier imports

### LINKED ISSUE #2233: Mendeley import: Invalid-field-for-type error
https://forums.zotero.org/discussion/92600/please-help-zotero-sync-not-preserving-folder-structures-after-mendeley-import

This person is getting an import error, but we're seemingly not logging it properly. There's an invalid-field-for-type error in the error report:

```
[JavaScript Error: "Error: 'publicationTitle' is not a valid field for type 32" {file: "chrome://zotero/content/xpcom/data/item.js" line: 776}]
```

But no error is logged with a trace in the debug output (e.g., with `Zotero.debug(e, 1)`). There are various warnings logged for invalid fields moved to Extra, including a publicationTitle/report, but there's only one of these in the error report, so this seems to be a separate problem later in the import.

In two separate runs, the import ended with these lines:

```
(3)(+0000004): Found existing file 1/HN3QX677 for mendeleyDB:fileHash b6f728a8bcdc4cfc3c40ffef955cd76e5392e5d4

(3)(+0000007): Found existing item 1/7MX3FETG for mendeleyDB:documentUUID 76ce596d-4b81-3d84-bedc-55f45122d108
```

7MX3FETG is a report, so that's likely the item that's causing the problem.

So we should 1) figure out why we're not logging an error properly in the debug output and 2) figure out where this error is coming from.

tnajdek:
This removes code that patches after earlier imports which fixes #2233 but please see https://github.com/zotero/zotero/issues/2233#issuecomment-957382877.

Related commits: https://github.com/zotero/zotero/commit/d27b622f9f8fe7b9c2fea25e678ef232c5ca032e, https://github.com/zotero/zotero/commit/ec2ad91938492469cc00318f61b391af40b733a4


dstillman:
Huh. I guess it's possible we forgot how this worked and added those patches without testing to see if they were actually necessary.

Did you try with these removed to confirm that we just overwrite the existing item with the new data?

tnajdek:
I did test it and it overwrites existing item with new values.

I'm going to add a few automated tests for the importer to cover this and a few common import scenarios.

tnajdek:
I've included a test showing that on re-import matched items are updated. Also this test will fail without 6cf0aba because of #2233.

PR #2235, opened 2021-11-02
fix Japanese translation

pn11:
"ファイルをダウンロードする" (means "download files")  and "必要に応じて" (means "on demand") looks swapped.


dstillman:
Thanks — I've applied this [on Transifex](https://www.transifex.com/zotero/zotero). All localization changes need to be made there, so you can sign up to join the translation team if you have other fixes.

PR #2244, opened 2021-11-10
Additional tests for Mendeley importer

tnajdek:


test/tests/mendeleyImport.js
```diff
@@ -1,12 +1,34 @@
...
+		const emptyPDF = await Zotero.File.getContentsFromURLAsync('resource://zotero-unit-tests/data/empty.pdf');
+		const port = 16213;
+		httpd = new HttpServer();
+		httpdURL = `http://127.0.0.1:${port}`;
+		httpd.start(port);
+		httpd.registerPathHandler('/file1.pdf',
+			{
+				handle: function (request, response) {
+					response.setStatusLine(null, 200, "OK");
+					response.setHeader('Content-Type', 'application/json', false);
```
dstillman, line 22: Is this supposed to be `application/json`?

If you just want to serve the file, you can probably just use `httpd.registerFile()` ([example](https://github.com/zotero/zotero/blob/3b993b2af346ce8c1d37282808f9b3ab2d16c5a0/test/tests/attachmentsTest.js#L794-L797)).

test/tests/mendeleyImport.js
```diff
@@ -1,12 +1,34 @@
...
+					response.write(emptyPDF);
+				}
+			});
+	});
+
+	after(function* () {
+		var defer = new Zotero.Promise.defer();
+		httpd.stop(() => defer.resolve());
+		yield defer.promise;
```
dstillman, line 31: I assume this is copied and pasted, but might as well write it how we would these days:

```javascript
after(async function () {
    await new Zotero.Promise((resolve) => {
        httpd.stop(resolve);
    });
}

Comment (tnajdek) [reply test/tests/mendeleyImport.js]: I've used a bit shorter version (below) but happy to change if that's not in a right coding style

````javascript
after(async () => new Zotero.Promise(resolve => httpd.stop(resolve)));
````

File: test/tests/mendeleyImport.js (hunk 1 above)
Comment (tnajdek) [reply test/tests/mendeleyImport.js]: Thanks for point to the example, that's much nicer. (also 100% shouldn't have been `application/json`)

File: test/tests/mendeleyImport.js (hunk 2 above)
Comment (dstillman) [reply test/tests/mendeleyImport.js]: Even better, except we usually leave these as multi-line so they're easier to extend. So let's split the difference:

```javascript
after(async () => {
    await new Zotero.Promise(resolve => httpd.stop(resolve));
}
```

dstillman:
Oh, and the file should be mendeleyImportTest.js for consistency.

tnajdek:
Updated & renamed

Issue #2181, opened 2021-09-02
Allow to batch renumber annotation page labels


mrtcode:
Annotation pages have to be manually renumbered when page label can't be correctly determined because:
- PDF doesn't have page labels information.
- PDF has page labels information but it's incorrect.
- Page label extraction from page fails:
  - Too few pages to determine page label position in PDF page. Currently at least 4 numbered pages are necessary (different label position between odd/even pages is supported).
  - PDF pages aren't continuous - gaps or additional pages between pages.
  - PDF is OCRed which means page label position is drifting.
- Single PDF page has multiple pages (more common for OCRed books).

Possible solutions:
- There could be an option called "Renumber annotation page labels starting from this page" in page context menu. When clicked it would show a popup with the current page label and suggest to enter a new page label.
- Selecting multiple annotations in annotations sidebar could also show a context menu option to manually set labels for selected annotations, starting from the first annotation page.

mrtcode:
@dstillman

As annotations on iOS app and Zotero client can sometimes have different page labels, I think it's now even more important to have a way to re-calculate them.

Here is what I suggest:
- Double clicking annotation page label or clicking annotation context menu option "Update Page Numbers" would prompt a popup.
- The popup would have an input box with pre-entered page label and a checkbox "Update annotations in other pages accordingly".
- The popup will replace the current page label input that currently appears on double click.
- We were discussing about only updating the further pages after the current page, but I think updating all pages is a simpler concept for most users.
- For advanced use cases like updating annotations range in a single page (one PDF page has multiple physical pages) or multiple pages between gaps, you would need to select those annotations and click "Update Page Numbers" context menu option. There won't be any additional checkbox, but if selected annotations are from multiple PDF pages, they will be automatically re-calculated accordingly.

I think this would be a sweet spot between the most common simple cases and rare advanced cases.

Offtopic: I also suggest to add another annotation context menu option called "Tweak Highlight Text", because double clicking on highlight text is too difficult to discover.

dstillman:
> We were discussing about only updating the further pages after the current page, but I think updating all pages is a simpler concept for most users.

Not sure about this as the only option. To me, updating page numbers on previous pages is a little unsettling — you're already past those, and it would feel like you had to scroll back and lose your place to review those, while you're already going to continue scrolling forward to deal with later pages.

We could just give all three options, in a drop-down or radio: "Update:" "this page"/"this page and later pages"/"all pages"

mrtcode:
[...]

PR #2267, opened 2021-12-08
Add popup to update annotations in batch

mrtcode:
Fixes https://github.com/zotero/zotero/issues/2181. We should review this ASAP because this blocks further `pdf-reader` development.

Some notes:
1) The popup can be opened when double clicking page label or over annotation context menu (although it's possible that  in future all pdf-reader XUL context menus will be replaced with with internal `pdf-reader` popups, because 1) we need a better flexibility for color picking, 2) `pdf-reader` already has various popups, 3) for the web version of `pdf-reader` it will still need to have its own popups).
2) The popup has macOS look, which isn't good for other platforms. But we have the same problem for the whole `pdf-reader`.


chrome/content/zotero/xpcom/reader.js
```diff
@@ -555,35 +568,41 @@ class ReaderInstance {
...
+							await Zotero.Annotations.saveCacheImage(savedAnnotation, blob);
 						}
-					};
-					let savedAnnotation = await Zotero.Annotations.saveFromJSON(attachment, annotation, saveOptions);
-					
-					if (annotation.image && !await Zotero.Annotations.hasCacheImage(savedAnnotation)) {
-						let blob = this._dataURLtoBlob(annotation.image);
-						await Zotero.Annotations.saveCacheImage(savedAnnotation, blob);
 					}
+					await Zotero.Notifier.commit(notifierQueue);
```
dstillman, line 589: This (and others) should be in a `finally`, or else annotations that are saved successfully won't be updated properly and the UI will get out of sync.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -1355,6 +1356,12 @@ pdfReader.promptTransferFromPDF.text = Annotations stored in the PDF file will b
 pdfReader.promptTransferToPDF.title = Store Annotations in File
 pdfReader.promptTransferToPDF.text = Annotations will be transferred to the PDF file and will no longer be editable in %S.
 pdfReader.promptPasswordProtected = The operation is not supported for password-protected PDF files.
+pdfReader.updatePageNumber = Update Page Number
```
dstillman, line 1359: → `Change Page Number…`

chrome/content/zotero/xpcom/reader.js
```diff
@@ -489,6 +488,20 @@ class ReaderInstance {
...
+		menuitem.addEventListener('command', () => {
+			let data = {
+				action: 'popupCmd',
+				cmd: 'openPageLabelPopup',
+				standalone,
+				id: currentID
+			};
+			this._postMessage(data);
+		});
+		popup.appendChild(menuitem);
```
dstillman, line 504: Without a confirmation on Delete and without Undo, this can't be next to Delete. Undo would be best, but for now let's add a menu separator below.

dstillman (commented):
1) The radio buttons aren't vertically aligned with the text:
  
  Screen Shot 2021-12-13 at 6 36 42 AM

2) The positioning of the popup isn't right. It looks like it's being centered in the frame, but if you have the right-hand pane open, it's off-center to the left. As long as it's XUL, it would be better for it to be a `<panel>` anchored on the number, like the Add Item by Identifier popup [pointing to the toolbar button](https://github.com/zotero/zotero/blob/644c4e592580f4007ab590dcfe460d94934668ca/chrome/content/zotero/lookup.js#L136-L137). (And if this gets moved into the reader HTML later, we can position it similarly.)

3) It can also be much narrower (which will make it look better as a panel).

4) It should be possible to reset the label to the auto-detected value. A separate "Reset Page Number" menu option, displaying the same popup but with the label hard-coded? (And those options would reset later ones back to their auto-detected values too, not just update them based on the reset value of the current annotation.)

dstillman:
> Not related to this, but speaking about annotation context menu, I think we should add "Update Highlight Text" or "Update Extracted Text", because it's too difficult to discover that you can double click on extracted text.

Sure. "Edit Highlighted Text"? Might be necessary for accessibility also. All the more reason for the menuseparator.

mrtcode:
> 2\. The positioning of the popup isn't right. It looks like it's being centered in the frame, but if you have the right-hand pane open, it's off-center to the left. As long as it's XUL, it would be better for it to be a `<panel>` anchored on the number, like the Add Item by Identifier popup [pointing to the toolbar button](https://github.com/zotero/zotero/blob/644c4e592580f4007ab590dcfe460d94934668ca/chrome/content/zotero/lookup.js#L136-L137). (And if this gets moved into the reader HTML later, we can position it similarly.)

The popup is in HTML and inside pdf-reader already. Do you mean it should be positioned somewhere close to an annotation?

dstillman:
Yes — ideally it would be the same as the Add Item by Identifier panel, with an arrow.

mrtcode:
Sorry, but I still don't understand how do you imagine this. "Add Item(s) by Identifier" is a single button and anchoring popup is simple. But in the pagel label updating popup case there can be many ways to open it from sidebar or page. And it would be anchored on a page label? That is especially strange if you open the popup over page context menu and it gets anchored in the sidebar. Also if sidebar is closed, that would mean the popup would be anchored to annotation popup pagel label which is also strange.

dstillman:
None of those things seem weird to me, and all seem better and less weird than appearing in a modal (particularly an off-center modal, but even a centered modal). The page label is only displayed in one place: the annotation widget. There are context menu options for discoverability/accessibility, but the primary and quickest way to edit it is by double-clicking on the label. Showing the editing interface as a popup pointing to that value seems totally standard and logical to me. It's no different from going to Bookmarks → Bookmark Current Tab (or pressing Cmd-D) in Firefox and having the popup point to the star in the address bar, which is the primary indicator of bookmark status and the fastest way to access it.

mrtcode:
All right. Now how about the popup styling. It now really looks like a modal. I guess we want to remove the grayish background and the primary button shouldn't be blue? Because in annotation or selected text popups we don't have this as well.

mrtcode:
Cancel is probably also not necessary. Only one button is enough. It will be closed when clicking outside of popup.

dstillman:
Yes to all of the above, except maybe the button should still be blue. It should still trigger on Return, and maybe that helps indicate that it will.

mrtcode:
Where we could hook the page label popup if sidebar is closed and multiple annotations are selected (no annotation popup is visible)?

mrtcode:
I'll adjust styling a little bit, but here is how it looks now (radio input is in a different color because screenshots were taken on Linux):

Screenshot from 2021-12-15 17-35-16

Screenshot from 2021-12-15 17-36-18

dstillman:
> Where we could hook the page label popup if sidebar is closed and multiple annotations are selected (no annotation popup is visible)?

In that case it seems fine to center it, assuming we can center it in the window, taking the right-hand sidebar into account. If we can't, in the position of the right-click that opened the context menu?

dstillman:
Looks great, but it should point to the number itself, not the whole "Page 5" label.

mrtcode:
> 4\. It should be possible to reset the label to the auto-detected value. A separate "Reset Page Number" menu option, displaying the same popup but with the label hard-coded? (And those options would reset later ones back to their auto-detected values too, not just update them based on the reset value of the current annotation.)

Yes, an option to reset labels is really important and would probably be the main thing to do to fix page labels for annotations synced from iOS. But I don't think it needs to show the same popup with all the options because that would be an unnecessary complexity. I expect that automatic page label recognition will work well almost all the time, and in future it even get better, therefore just a single option to reset all annotation page labels should be sufficient. Although in this case I'm not sure where it should appear, because if it resets all annotations it doesn't make sense to be in annotation popup. Maybe in Tools menu? Or page context menu?

> > Where we could hook the page label popup if sidebar is closed and multiple annotations are selected (no annotation popup is visible)?
> 
> In that case it seems fine to center it, assuming we can center it in the window, taking the right-hand sidebar into account. If we can't, in the position of the right-click that opened the context menu?

Probably not worth to center it because it will start getting similar to modal dialog. It now points to the right-click position, which is always an annotation. It does that even if sidebar was open. When there is an annotation widget it points to the page label, of course.

Screenshot from 2021-12-17 18-10-10


Also added "Edit Highlighted Text". It appears even when opening annotation popup on a page. When clicked it opens sidebar and starts editing. Although "Edit Highlighted Text" might be confused with highlight annotation position "resizing".

dstillman:
> Although in this case I'm not sure where it should appear, because if it resets all annotations it doesn't make sense to be in annotation popup.

Or just an option in the widget context menu, and if you want to reset them all, you select them all? Tools would also be fine, if we think we need a more discoverable option (particularly given the iOS problem), but it should be in the context menu too.

> It now points to the right-click position, which is always an annotation. It does that even if sidebar was open.

I don't think this is right. I think the problem is that it's showing the option there at all. It shouldn't. If you can't see the page label, it shouldn't be an option. Same for "Edit Highlighted Text". There's no reason the context menu should be the same in the PDF view and in the sidebar, where the widget is already open — those are different contexts, and in the former case there's nothing visible for you to want to change. "Add to Note", the colors, and "Delete" all make sense in the PDF view, but the others don't.

So I think we remove "Change Page Number…" and "Edit Highlighted Text" in the PDF view context menu and have them appear only for the widget context menu (in both PDF view and sidebar), and then the popup always appears pointing to the visible label.

mrtcode:
> > Although in this case I'm not sure where it should appear, because if it resets all annotations it doesn't make sense to be in annotation popup.
> 
> Or just an option in the widget context menu, and if you want to reset them all, you select them all? Tools would also be fine, if we think we need a more discoverable option (particularly given the iOS problem), but it should be in the context menu too.

I guess there should be a more convenient way to reset all annotation labels that would avoid the inconvenient selection step, because this operation, currently, is more important than "Change Page Number" which doesn't need that step.


> So I think we remove "Change Page Number…" and "Edit Highlighted Text" in the PDF view context menu and have them appear only for the widget context menu (in both PDF view and sidebar), and then the popup always appears pointing to the visible label.

There is one case where we need it in the right-click context menu - updating page label for multiple selected annotations. Although it will be needed only on rare occasions (i.e. scanned books with two pages in a single PDF page). Therefore, I'll leave "Change Page Number…" in sidebar annotation right-click context menu and remove it from PDF page view.

mrtcode:
Additionally, there is no point to reset page label for single annotation. And that's another reason why this option, probably, shouldn't be in annotation context menu. Unless that option would show a confirmation popup with three buttons: 1) Cancel, 2) Reset all, 3) Reset this. But again, option 3) won't be used.

dstillman:
> There is one case where we need it in the right-click context menu - updating page label for multiple selected annotations. Although it will be needed only on rare occasions (i.e. scanned books with two pages in a single PDF page). Therefore, I'll leave "Change Page Number…" in sidebar annotation right-click context menu and remove it from PDF page view.

Yeah, but in that case the labels are visible, so that's fine.

> Additionally, there is no point to reset page label for single annotation. And that's another reason why this option, probably, shouldn't be in annotation context menu.

OK, if you think this will only be used for the whole PDF, we can just put it in the Tools menu. Though I still wonder whether people will waste time manually updating page numbers for iOS annotations via "Change Page Number…" without finding the option in Tools…

dstillman:
Though "Update Page Number…" could sort of apply to both actions. What about a tabbed dialog with Edit and Reset tabs? Or in some other way modifying the dialog so that you can choose to reset and it's clear that that affects all (e.g., graying out other radio options)?

You can already trigger an action that applies to all annotations from that context menu option, so I feel like it makes sense to do this all in one place. The problem here you're fixing is an incorrect label, so it doesn't really make sense to have to go to two completely different places to do it.

dstillman:
What about an "Auto-Detect" checkbox below the text box, and if you check that the text box is grayed out and shows the auto-detected label, and the radio buttons become disabled with "All pages" selected?

mrtcode:
> What about an "Auto-Detect" checkbox below the text box, and if you check that the text box is grayed out and shows the auto-detected label, and the radio buttons become disabled with "All pages" selected?

Yes, that's a brilliant idea! Simple and does exactly what we need.

mrtcode:
Here is how it looks now:

Screenshot from 2021-12-21 14-29-23

Screenshot from 2021-12-21 14-29-50

dstillman:
When clicking Auto-Detect, it shouldn't remove the other options — just disable them (including "All pages", I think) and select "All pages". Weird to have a single radio button visible.

mrtcode:
"All pages" doesn't always appear. In some cases only "This annotation" or other combinations that don't have "All pages" appear. So in theory it should select the option that covers the maximum count of annotations, even if it is "This annotation" only.

dstillman:
You mean it checks to see if other labels have been changed, and doesn't show "All pages" if not? That seems unnecessary. It's the idea here that your only option is to reset everything?

dstillman:
Or do you mean if you select multiple annotations and then use this option? Will you be able to limit the reset to just those? If so, isn't it weird to allow that but not allow you to reset a single annotation?

mrtcode:
No, I mean if there is only one annotation in PDF, it currently doesn't show two options ("This annotations" and "All annotations") that do the same.

dstillman:
Ah, OK. Then sure, showing the maximum makes sense.

mrtcode:
Ok, I think you can try this out for the final time.

PR #2270, opened 2021-12-10
Ensure the integration window appears on the same screen as Zotero

adomasven:
See https://forums.zotero.org/discussion/93398/position-of-quick-format-citation

This has actually bothered me for a while. Maybe it's an issue just on Linux, but for some reason on multi-screen setup integration windows appear all over the place, usually seems like on the second screen, regardless of where the Zotero window is. Ideally they would appear on the same screen as the user cursor (i.e. over the word processor), but I cannot find any information about gecko APIs to retrieve pointer coordinates outside of a mouse event. This centers the window on the same screen as the parent window.

This is like 99% fine, but I'd like someone to test with a macOS multi-screen setup since I don't have one in case it breaks something there.

### DIFF
--- a/chrome/content/zotero/bibliography.js
+++ b/chrome/content/zotero/bibliography.js
@@ -129,6 +129,7 @@ var Zotero_File_Interface_Bibliography = new function() {
 				Zotero.debug("No styles to select", 2);
 				return;
 			}
+			Zotero.Utilities.Internal.centerWindow(window);
 			Zotero_File_Interface_Bibliography.styleChanged();
 		}, 0);
 		
--- a/chrome/content/zotero/integration/addCitationDialog.js
+++ b/chrome/content/zotero/integration/addCitationDialog.js
@@ -73,16 +73,7 @@ var Zotero_Citation_Dialog = new function () {
 	this.load = Zotero.Promise.coroutine(function* () {
 		// make sure we are visible
 		window.setTimeout(function() {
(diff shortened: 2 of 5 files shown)


dstillman:
We don't want to center, though. The window should open at the last position, like all other Zotero windows.

I run Zotero on a multi-monitor Mac every day and I think it's always just opened at the last position, so this would be a regression. If reopening to the last position is broken on Linux and can't be fixed any other way, we should limit this to Linux.

adomasven:
> We don't want to center, though. The window should open at the last position, like all other Zotero windows.

Yeah, perhaps this is addressing the wrong problem. None of Zotero windows (except for Zotero itself) open in their last position on Linux. I'll look into it some more then.

krassowski:
Was there a follow up on this? I am using Ubuntu with multi-monitor setup (currently with one monitor disabled) and the add citation dialog does NOT show on the last position, but instead in top-left corner which gets partially occluded by Ubuntu panel on startup making it quite unusable (I do not see what I type) until dragged to a more reasonable position.

Screenshot from 2023-05-19 12-53-41

I am happy to provide a PR.

dstillman:
If the problem persists in Zotero 7, which will be out in beta very soon, let us know.

krassowski:
It works perfectly in Zotero 7.0.0-dev.19+eecd85a1c build - thank you so much!

PR #2276, opened 2021-12-15
Fix autosuggestion background color in dark mode

ccat3z:
Before:

After:


dstillman:
Thanks!

RisenL:
Awesome! it's nearly to see formal version with new feature.

Issue #2284, opened 2021-12-21
Note editor: Option to hide title/author from annotation citation


dstillman:
Our citations include the author or title (when no author) over and over, which is pretty annoying, at least for a note for a single PDF, which will be most of them. ZotFile annotations include only the page number.

We probably need a menu option to hide these, similar to the annotation-colors option. And there probably should be a different default depending on how you create the note. E.g., once we have the ability to create a note from annotations from multiple items, that should obviously default to showing the author/title, but for a note created from the annotations on the same PDF, it would make sense to hide by default.

I'm not totally sure how to handle this when you export, add to Word, etc. For now, probably just a display setting in the note that doesn't affect output, but eventually it'll probably need to be an option for Markdown output.

mrtcode:
If citations can be easily added/removed, that's probably enough. And you get the same output everywhere. 

If you want citations in Markdown or word-processor you just temporary add those to the original note and then remove after the export.

Otherwise, I guess, hidden citation nodes might accidentally appear in unexpected places, but I would need to test that. Unless that option would only somehow collapse citations but still show a few symbols of it i.e. "(...)".

If we don't want them to be visible at all, it's probably better to just remove them. Although citations that are already moved somewhere else (not after annotation) wouldn't be removed, because recreating them wouldn't be possible as well.

dstillman:
You mean since we already let you easily get back to the PDF page from the annotation, which is all the ZotFile links did anyway, you could just hide the citations altogether? And that the citations aren't useful within a single item's notes?

My concern there would be that people might still find the page numbers useful while taking notes and reviewing annotations, even for a single PDF. Say, to get a sense of where each annotation came from without jumping back to the PDF for each one?

mrtcode:
> You mean since we already let you easily get back to the PDF page from the annotation, which is all the ZotFile links did anyway, you could just hide the citations altogether? And the citations aren't useful in general within a single item's notes?

Yes for a single item they probably aren't useful, although if they aren't inserted by default, some people will never discover them as well.


> My concern there would be that people might still find the page numbers useful while taking notes and reviewing annotations, even for a single PDF. Say, to get a sense of where each annotation came from without jumping back to the PDF for each one?

Yes, maybe page number is useful, but you can add/remove citations at any time. Alternatively, the highlight could show more information (including page number) when hovered, if needed.

dstillman:
[...]

PR #2288, opened 2021-12-24
Allow to add/remove citations in notes

mrtcode:
As discussed in #2284, this is an experimental PR that allows to add/remove citations that are immediately after annotation. Citations in other places or multi-item citations are untouched.

It now also shows a title with formatted citation when hovering a highlight or image annotation.

For now the default behavior (whether to show colors or citations, or not) is not changed and it also doesn't remember the last setting yet. Firstly let's try this out and make sure actually adding/removing citations is the right way to go.

It actually feels much cleaner without citations.

### DIFF
--- a/chrome/content/zotero/xpcom/editorInstance.js
+++ b/chrome/content/zotero/xpcom/editorInstance.js
@@ -589,11 +589,15 @@ class EditorInstance {
 					if (!item) {
 						return;
 					}
-					let attachments = Zotero.Items.get(item.getAttachments()).filter(x => x.isPDFAttachment());
-					if (citationItem.locator && attachments.length === 1) {
-						let zp = Zotero.getActiveZoteroPane();
-						if (zp) {
-							zp.viewPDF(attachments[0].id, { pageLabel: citationItem.locator });
+
+					if (citationItem.locator) {
+						let attachments = await item.getBestAttachments();
+						attachments = attachments.filter(x => x.isPDFAttachment());
+						if (attachments.length) {
+							let zp = Zotero.getActiveZoteroPane();
+							if (zp) {
+								zp.viewPDF(attachments[0].id, { pageLabel: citationItem.locator });
+							}
 						}
 					}
 					else {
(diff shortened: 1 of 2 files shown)


dstillman:
So I guess my worry here would be that we make it very easy to create a version of the note without the annotations cited. It makes sense as a display option, and it makes a little more sense when using "Include Zotero Links" for Markdown export, but when you use Add Note or note export without links, it's a little dangerous (though, of course, you might have inserted a citation for the item manually). Do we want export-time options to add citations back when using those methods?

Otherwise:

I think "Remove Citations" is too scary. It's a little imprecise, but "Hide Citations" (and "Show Citations") is probably better, since it conveys that it's reversible. And maybe same for colors?

And, actually, doesn't it need to be "Hide[/Show] Annotation Citations"? Because it wouldn't affect standalone citations.

Maybe put a separator above the annotation options, since there will be 2–3 of them now?

Maybe we should have "Hide Citation" show up in the popup for an individual citation? You can delete manually, but that's tedious, and it would mirror "Add Citation" (which should probably be "Show Citation") on the annotation.

dstillman:
The colors toggle would arguably make sense as an export-option too. So maybe both of them are both display options and export options?

mrtcode:
So we have three ways to export a note: 1) export as file, 2) export to clipboard, 3) "Add Note" from processor. We can add the necessary export options to 1) and 2), but for 3) we don't have any way to do that. 

Next, an export option for colors only makes sense for "Add Note" and Rich Text or HTML export.

dstillman:
As a temporary solution until we have a proper redesign, we could probably just add the Z menu from the Quick Format dialog to the Add Note dialog? (@adomasven)

I'm not sure what the wording on these would be. It's not really "Include Annotation Colors"/"Include Citations", because that implies that if you disable them you don't get citations or colors, but that's not the option, right? It's forcing them to be included even when they're not. "Include All Annotation Colors"/"Include All Citations"?

mrtcode:
Ok, we can review this now.
Various little things from other issues are fixed here as well.

1) Removed "Insert Citation" from context menu #2308.
2) No longer shows "Go to Page" when the citation immediately follows an annotation for the same page #2306.
3) Citation button in the toolbar restores the citation if cursor is inside an annotation #2307.
4) Added a separator to the three dots menu, although the word "annotation" feels to be used excessively.
5) Opens the best PDF attachment #2306.
6) Color and citation add/remove words are now renamed to show/hide.
7) Add "Hide Citation" option to citation popup.

PR #2293, opened 2021-12-29
Overhaul Scaffold

AbeJellinek:
- New code editor using [Monaco](https://microsoft.github.io/monaco-editor/), the core of VS Code. It allows us inline ESLint errors, API autocomplete and type information, and some custom utilities:
	- Inline Run link shows above each `detect`/`do` method and Run and Update links show above tests.
	- Hovering over an item type string (e.g. `"bookSection"`) shows its fields, equivalent to the _Help -> List all fields for item type_ menu. (The editor also suggests fields as you type if it can infer the type of your item: if you declare `let item = new Zotero.Item('book')`, typing `item.` will suggest `title`, `ISBN`, etc.).
	- Hovering over a translator UUID shows the translator's name and metadata. Very useful for old translators with uncommented `setTranslator` calls.
- Support for all translator types, including export and search. Export doesn't yet support tests.
- A streamlined UI that better uses native system menus (no more _Help -> Keyboard Shortcuts_ - shortcuts just show in the menus)
	- Import has been renamed Test Input to better reflect its purpose (and because it also provides input to search translators now).
	- Tests and Testing tabs have been unified. Populate and Save are gone - the JSON is the source of truth and the GUI test list updates automatically when it changes. Buttons to create tests have been moved to the Browser and Test Input tabs.
	- You can create a new blank translator without closing and reopening Scaffold now.
- Updated the web translator template to use async methods (more work remains on that/other templates - we should discuss how we want them to look).
- Other misc UI and stability improvements.

Things that are not done yet and should be discussed:
- Tests for export translators? Has this been discussed in the past?

Known issues:
- Edit menu items are sometimes inappropriately enabled or disabled. This is true of the note editor too and I suspect that a solution for one would be pretty portable to the other.
- The toolbar is too big on Windows.

That said, I think this is ready for review and testing by others.


AbeJellinek:
@dstillman: I think this is ready. The toolbar isn't perfect on Windows, but otherwise everything is working as it should.

dstillman:
How do you have eslint set up? I have to select `node_modules/.bin/teslint` within the `translators` repo directory to get it to work — I can't just select the global eslint. Is that not the case for you? Or did you install `eslint-plugin-zotero-translator` globally?

(Also this is amazing.)

dstillman:
- For tests, I think it'd be better to move the list above the text field. Even if the JSON is the source of truth, you're not supposed to edit that directly, so the main operation in this tab is running individual tests, which is more easily done from the list.
- I still find the "Update" button in tests totally confusing. That means "Run the test and update the translator with the results", right? Can we clarify that? At least a clearer tooltip, but maybe a clearer button name as well.
- Can we fix the editor so that scrolling stops with the last line at the bottom, instead of scrolling the last line all the way to the top with everything else out of view?
- Is the triangle the only option for the inline function-running? Can we remove it and just have the text? To me it looks distractingly like a disclosure triangle rather than a run button.
- Can we show eslint status somewhere? Whether it's configured, and what path is selected?
- There should be a Cancel button in the ESLint configuration dialog. I don't think we need the installation-instructions button (particularly based on my question above).

Other than that, we should try to fix the toolbar size on Windows if it's wrong, but if you've been using this, I'd say we should just merge it and refine further later. This will surely be a huge improvement over what we had as is.

AbeJellinek:
>How do you have eslint set up?

`/opt/homebrew/bin/eslint`, version 7.31.0. Using global ESLint will prevent a few checks from working, but only ones that that rely on `teslint`'s translator caching. I didn't install the plugin globally - I think it detects it from the `eslintrc`.

Should it just try `teslint` first before even asking the user? I hadn't thought about global `eslint` vs `teslint` but we probably do want to use the latter when possible.

>(Also this is amazing.)

Thank you!

>I still find the "Update" button in tests totally confusing. That means "Run the test and update the translator with the results", right? Can we clarify that?

Yeah. I just don't know of a short, non-jargony term for it. In Rust-land it's "bless," but something about "Run and Bless" just sounds weird. (In any case, I pushed the just-sounds-weird option, so see what you think.)

>Is the triangle the only option for the inline function-running?

At the moment. Should we have a context menu option to run the test that the cursor is inside?

>Can we show eslint status somewhere? Whether it's configured, and what path is selected?

Sure. I added it in the Tools menu next to the existing "Configure ESLint..." menu item. If that's not visible enough, maybe in the Metadata tab? (Or in the menu _and_ the metadata tab.)

>we should try to fix the toolbar size on Windows if it's wrong

I would really like to, but I've messed with the CSS a bunch and haven't made any headway. It's definitely a byproduct of my piggybacking off the `#zotero-toolbar` CSS, and it's probably caused by a more specific selector somewhere that isn't matching Scaffold `toolbarbutton`s, but I just can't find it. (Would be helpful if I could get remote debugging to work on my Windows VM - are there instructions anywhere?)

dstillman:
> Should it just try teslint first before even asking the user?

I would even say we shouldn't even make it configurable — just warn that teslint isn't installed in the repo directory.

> I didn't install the plugin globally - I think it detects it from the eslintrc.

Hmm, I just get an error if I try to use homebrew eslint…

> "Run and Bless" just sounds weird

Yeah, way too weird. How about "Run and Update", and then make it clearer in the tooltip?

> Should we have a context menu option to run the test that the cursor is inside?

We can leave this for now — much more discoverable. (Maybe possible to hide the triangle via CSS? I didn't check.)

> Would be helpful if I could get remote debugging to work on my Windows VM - are there instructions anywhere?

It'd be the same process — Firefox 60 ESR (blocked from updating) devtools to a source build with the right flags. The betas also now include the devtools, so if it's easier we can wait until this is out in a beta and then you can use Firefox 60 ESR with that, rather than having to build on Windows.

AbeJellinek:
OK, all suggestions implemented.

>Maybe possible to hide the triangle via CSS?

They were just Unicode triangles I added manually in the actions' labels! All gone now.

AbeJellinek:
(You'll need to pull zotero/translators to get linting back - teslint didn't support outputting raw JSON.)

dstillman:
Merged

Issue #1413, opened 2018-01-15
Customizable renaming rules


dstillman:
À la ZotFile, though I don't know if we'd use exactly the same rules.

(I thought we had a ticket for this, but I can't find one.)

dstillman:
https://forums.zotero.org/discussion/comment/395176/#Comment_395176

tnajdek:
I'll take a stab at this. I think first step is to improve the current system to accept more wildcards for fields and offer more flexibility. We can then think about adding a new UI to preferences so that this can be configured more easily without using hidden prefs.

bwiernik:
This is the current list of wildcards Zotfile gives http://zotfile.com/

tnajdek:
If we still want this, we should probably target Zotero 7? Should I port this to `fx102` branch?

PR #2297, opened 2022-01-06
Customizable renaming rules #1413

tnajdek:
This PR includes multiple changes to extend capabilities of `getFileBaseNameFromItem` ~~but keeps it backwards-compatible~~ using template engine. See #1413.

(Removed details of previous implementation, current in comment below)

### DIFF
--- a/chrome/content/zotero/xpcom/attachments.js
+++ b/chrome/content/zotero/xpcom/attachments.js
@@ -23,7 +23,7 @@
  ***** END LICENSE BLOCK *****
 */
 
-Zotero.Attachments = new function(){
+Zotero.Attachments = new function () {
 	const { HiddenBrowser } = ChromeUtils.import("chrome://zotero/content/HiddenBrowser.jsm");
 	
 	// Keep in sync with Zotero.Schema.integrityCheck() and this.linkModeToName()
@@ -2261,35 +2261,16 @@ Zotero.Attachments = new function(){
 		if (!formatString) {
 			formatString = Zotero.Prefs.get('attachmentRenameFormatString');
 		}
-		
-		// Replaces the substitution marker with the field value,
-		// truncating based on the {[0-9]+} modifier if applicable
-		function rpl(field, str) {
-			if (!str) {
-				str = formatString;
-			}
-			
-			switch (field) {
-				case 'creator':
-					field = 'firstCreator';
-					var rpl = '%c';
-					break;
-					
-				case 'year':
-					var rpl = '%y';
-					break;
-					
-				case 'title':
-					var rpl = '%t';
-					break;
-			}
-			
+
+		const markers = {
+			c: 'firstCreator',
+			y: 'year',
+			t: 'title'
+		};
+
+		const getValue = (field) => {
 			var value;
 			switch (field) {
-				case 'title':
-					value = item.getField('title', false, true);
-					break;
-				
 				case 'year':
 					value = item.getField('date', true, true);
 					if (value) {
@@ -2298,38 +2279,35 @@ Zotero.Attachments = new function(){
 							value = '';
 						}
 					}
-				break;
+					break;
 				
 				default:
 					value = '' + item.getField(field, false, true);
 			}
-			
-			var re = new RegExp("\{?([^%\{\}]*)" + rpl + "(\{[0-9]+\})?" + "([^%\{\}]*)\}?");
-			
-			// If no value for this field, strip entire conditional block
-			// (within curly braces)
-			if (!value) {
-				if (str.match(re)) {
-					return str.replace(re, '')
-				}
-			}
-			
-			var f = function(match, p1, p2, p3) {
-				var maxChars = p2 ? p2.replace(/[^0-9]+/g, '') : false;
-				return p1 + (maxChars ? value.substr(0, maxChars) : value) + p3;
-			}
-			
-			return str.replace(re, f);
-		}
-		
-		formatString = rpl('creator');
-		formatString = rpl('year');
-		formatString = rpl('title');
-		
+
+			return value;
+		};
+
+		// Regexp contains 4 capture groups all wrapped in {}:
+		// 		* Prefix before the wildcard, can be empty string
+		// 		* Any recognized marker. % sign marks a wildcard and is required for a match but is
+		// 		  not part of the capture group. Recognized markers are specified in a `markers`
+		// 		  lookup.
+		// 		* Optionally a maximum number of characters to truncate the value to
+		// 		* Suffix after the wildcard, can be empty string
+		const re = new RegExp(`{([^%{}]*)%(${Object.keys(markers).join('|')})({[0-9]+})?([^%{}]*)}`, 'ig');
+
+		formatString = formatString.replace(re, (match, pre, marker, maxChars, post) => {
+			maxChars = maxChars ? maxChars.replace(/[^0-9]+/g, '') : false;
+			const field = markers[marker];
+			const value = getValue(field);
+			return pre + (maxChars ? value.substr(0, maxChars) : value) + post;
+		});
+
 		formatString = Zotero.Utilities.cleanTags(formatString);
 		formatString = Zotero.File.getValidFileName(formatString);
 		return formatString;
-	}
+	};
 	
 	
 	this.shouldAutoRenameFile = function (isLink) {
@@ -3029,4 +3007,4 @@ Zotero.Attachments = new function(){
 		}
 		throw new Error(`Invalid link mode name '${linkModeName}'`);
 	}
-}
+};
--- a/chrome/content/zotero/xpcom/attachments.js
+++ b/chrome/content/zotero/xpcom/attachments.js
@@ -2397,6 +2397,35 @@ Zotero.Attachments = new function () {
 		formatString = Zotero.File.getValidFileName(formatString);
 		return formatString;
 	};
+
+	this.convertLegacyFormatString = function (formatString) {
+		const markers = {
+			c: 'firstCreator',
+			y: 'year',
+			t: 'title'
+		};
+
+		// Regexp contains 4 capture groups all wrapped in {}:
+		// 		* Prefix before the wildcard, can be empty string
+		// 		* Any recognized marker. % sign marks a wildcard and is required for a match but is
+		// 		  not part of the capture group. Recognized markers are specified in a `markers`
+		// 		  lookup.
+		// 		* Optionally a maximum number of characters to truncate the value to
+		// 		* Suffix after the wildcard, can be empty string
+		const re = new RegExp(`{([^%{}]*)%(${Object.keys(markers).join('|')})({[0-9]+})?([^%{}]*)}`, 'ig');
+
+		return formatString.replace(re, (match, prefix, marker, truncate, suffix) => {
+			const field = markers[marker];
+			truncate = truncate ? truncate.replace(/[^0-9]+/g, '') : false;
+			prefix = prefix ? `prefix="${prefix}"` : null;
+			suffix = suffix ? `suffix="${suffix}"` : null;
+			truncate = truncate ? `truncate="${truncate}"` : null;
+
+			Zotero.debug({ match, field, truncate, prefix, suffix });
+
+			return `{{ ${[field, truncate, prefix, suffix].filter(f => f !== null).join(' ')} }}`;
+		});
+	};
 	
 	
 	this.shouldAutoRenameFile = function (isLink) {
--- a/chrome/content/zotero/xpcom/attachments.js
+++ b/chrome/content/zotero/xpcom/attachments.js
@@ -2376,6 +2376,7 @@ Zotero.Attachments = new function () {
 		};
 
 		const itemType = args => common(Zotero.ItemTypes.getLocalizedString(item.itemType), args);
+		const itemTypeRaw = item.itemType;
 
 		const creatorFields = ['authors', 'editors', 'creators'].reduce((obj, name) => {
 			obj[name] = (args) => {
@@ -2390,7 +2391,7 @@ Zotero.Attachments = new function () {
 			item.getField('firstCreator', false, true).replaceAll('\u2068', '').replaceAll('\u2069', ''), args
 		);
 
-		const vars = { ...fields, ...creatorFields, firstCreator, itemType, year };
+		const vars = { ...fields, ...creatorFields, firstCreator, itemType, itemTypeRaw, year };
 
 		formatString = Zotero.Utilities.Internal.generateHTMLFromTemplate(formatString, vars);
 		formatString = Zotero.Utilities.cleanTags(formatString);
--- a/chrome/content/zotero/xpcom/attachments.js
+++ b/chrome/content/zotero/xpcom/attachments.js
@@ -2327,11 +2327,11 @@ Zotero.Attachments = new function () {
 			return value;
 		};
 
-		const initializeFn = (name, shouldInitialize, initializewith) => (shouldInitialize ? name.slice(0, 1).toUpperCase() + initializewith : name);
+		const initializeFn = (name, shouldInitialize, initializeWith) => (shouldInitialize ? name.slice(0, 1).toUpperCase() + initializeWith : name);
 
-		const transformName = (creator, { name, namepartseparator, initialize, initializewith } = {}) => {
+		const transformName = (creator, { name, namePartSeparator, initialize, initializeWith } = {}) => {
 			if (creator.name) {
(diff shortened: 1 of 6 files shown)


dstillman:
If this isn't going to be a drop-in replacement anyway (and I agree that it needn't be), is there any reason to keep the one-character placeholders instead of using a different template format with actual field names? We could even reuse the [basic template engine](https://github.com/zotero/zotero/blob/14f7d3acadd7e3879288f26c59e8fe73d858941d/chrome/content/zotero/xpcom/utilities_internal.js#L2094-L2123) we use for note templates.

So instead of this for the default format:

```
{%c - }{%y - }{%t{50}}
```

we could have this:

```
{{firstCreator suffix=" - "}}{{year suffix=" - "}}{{title maxlength="50"}}
```

We could add other parameters as needed, and we'd get the conditional logic from that engine, which people might appreciate for things like item types.

Creators could be things like this:

  * `{{authors max="3"}}` — `Smith, Foo, Bar` (defaults `join=", "` and `name="family"`)
  * `{{authors max="3" name="full" initialize="given"}}` — `J. Smith, A. Foo, B. Bar` (default `initialize-with="."`)
  * `{{authors max="3" name="family-given" initialize="given" initialize-with="" name-part-separator=""}}` — `SmithJ, FooA, BarB`
  * `{{authors max="3" name="full" initialize="full" initialize-with="" name-part-separator=""}}` — `JS, AF, BB`
  * `{{authors max="3" case="upper"}}` — `SMITH, FOO, BAR`

(If we thought this was too complicated, we could just have a bunch of shortcuts in `name`, like `name="family-given-initial"` for `SmithJ`, but the composability is nice. Otherwise the shortcuts themselves get pretty complicated, and probably more limited when it comes to things like periods. And we'd just be giving a bunch of examples either way.)

For `{{itemType}}`, we could probably automatically localize it in the output while still using the unlocalized name in conditionals. (A little weird, but alternative would be something like `{{itemType localize='true'}}`, and I doubt anyone needs an unlocalized item type in the filename.)

And this would be easy to extend later, if we wanted to support more of ZotFile's functionality, like regexes.

tnajdek:
OK, I'll revisit this. This sounds better and reusing template engine sounds like a good idea.

tnajdek:
I've converted this to use the template engine which now makes all the examples you've listed possible (except dash-separated attributes are now concatenated for compatibility with template engine).

There are two minor issues left to address:

* As far as I can tell, template engine does not make it possible to handle`itemType` differently depending on whether it's used in a condition or for rendering output so we would either need to go with `{{itemType localize='true'}}` or adjust the engine slightly.

* New system isn't backward compatible so we will probably need to migrate preference? I could come up with a migration function that would convert old-style format with wildcards to a matching new-style format, I'm just not sure where to put this function, do we have a pref migration functionality?

tnajdek:
Update on the two remaining issues:

* I've introduced `itemTypeRaw` for use in logic statements. Current template engine does not support comparing to the result of a argument function so it won't be possible to use any other fields in logic statements. This can be improved but I'd need to add functionality to the template engine (not sure if that's something anyone would use?)
* I've introduced a function to migrate old syntax, just need to plug it in so it triggers after update

test/tests/attachmentsTest.js
```diff
@@ -1293,11 +1293,296 @@ describe("Zotero.Attachments", function() {
...
+			assert.equal(
+				Zotero.Attachments.getFileBaseNameFromItem(item, '{{ publicationTitle case="lower" }}'),
+				'best publications place'
+			);
+			assert.equal(
+				Zotero.Attachments.getFileBaseNameFromItem(item, '{{ publicationTitle case="title" }}'),
+				'Best Publications Place'
+			);
+			assert.equal(
+				Zotero.Attachments.getFileBaseNameFromItem(item, '{{ publicationTitle case="dash" }}'),
```
dstillman, line 1480: `dash` → `hyphen`

test/tests/attachmentsTest.js
```diff
@@ -1293,11 +1293,296 @@ describe("Zotero.Attachments", function() {
...
+			);
+		});
+
+		it('should accept itemType or any other field', function () {
+			assert.equal(
+				Zotero.Attachments.getFileBaseNameFromItem(item, '{{ itemType }}'),
+				'Journal Article'
+			);
+			assert.equal(
+				Zotero.Attachments.getFileBaseNameFromItem(item, '{{ publicationTitle }}'),
```
dstillman, line 1499: It looks like we're handling it, but we should make sure we're testing base-mapped fields (e.g., `bookTitle` in `bookSection` for `publicationTitle`).

test/tests/attachmentsTest.js
```diff
@@ -1293,11 +1293,296 @@ describe("Zotero.Attachments", function() {
...
+			assert.equal(
+				Zotero.Attachments.getFileBaseNameFromItem(item, '{{ publicationTitle case="snake" }}'),
+				'best_publications_place'
+			);
+		});
+
+		it('should accept itemType or any other field', function () {
+			assert.equal(
+				Zotero.Attachments.getFileBaseNameFromItem(item, '{{ itemType }}'),
+				'Journal Article'
```
dstillman, line 1496: `{{ itemType localize="true" }}` might make more sense. I'm not sure why anyone would be putting an item type in a filename anyway, so maybe better to optimize for use in conditionals without needing to use `itemTypeRaw`?

chrome/content/zotero/xpcom/attachments.js
```diff
@@ -2241,95 +2241,195 @@ Zotero.Attachments = new function(){
...
+		const vars = { ...fields, ...creatorFields, firstCreator, itemType, itemTypeRaw, year };
+
+		formatString = Zotero.Utilities.Internal.generateHTMLFromTemplate(formatString, vars);
 		formatString = Zotero.Utilities.cleanTags(formatString);
 		formatString = Zotero.File.getValidFileName(formatString);
 		return formatString;
-	}
+	};
+
+	this.convertLegacyFormatString = function (formatString) {
```
dstillman, line 2405: This is 100% overkill, given the extremely limited options we had, but very nice! There's a pref-migration mechanism in prefs.js.

chrome/content/zotero/xpcom/attachments.js
```diff
@@ -2241,95 +2241,195 @@ Zotero.Attachments = new function(){
...
+					break;
+				case 'lower':
+					value = value.toLowerCase();
+					break;
+				case 'sentence':
+					value = value.slice(0, 1).toUpperCase() + value.slice(1);
+					break;
 				case 'title':
-					value = item.getField('title', false, true);
+					value = value.split(' ').map(word => word.charAt(0).toUpperCase() + word.slice(1).toLowerCase()).join(' ');
```
dstillman, line 2315: Better to use `Zotero.Utilities.capitalizeTitle(str, true)`, which has some (English) skip words and also will ignore words that already contain capitalization. (The `true` for `force` is necessary until we remove the pointless `capitalizeTitles` pref.)

test/tests/attachmentsTest.js
```diff
@@ -1293,11 +1293,296 @@ describe("Zotero.Attachments", function() {
...
+			assert.equal(
+				Zotero.Attachments.getFileBaseNameFromItem(itemPatent, '{{ authors max="2" name="family" initialize="family" initializewith="" }}'),
+				'A'
+			);
+			assert.equal(
+				Zotero.Attachments.getFileBaseNameFromItem(item, '{{ authors max="1" name="full" initialize="full" namepartseparator="" initializewith="" }}'),
+				'FB'
+			);
+			assert.equal(
+				Zotero.Attachments.getFileBaseNameFromItem(item, '{{ authors max="1" order="desc" name="full" initialize="full" namepartseparator="" initializewith="" }}'),
```
dstillman, line 1409: What's the engine limitation that's preventing hyphenated attributes? Can we just change that?

test/tests/attachmentsTest.js
tnajdek: I've added explicit tests for this:

https://github.com/zotero/zotero/pull/2297/commits/7e97371a52254177248023bd22adce3b8271f9f4#diff-0870dd7dae9dd9b491e5db04d70e7d91e240a21401c7fb536b364769674f978aR1564-R1570

chrome/content/zotero/xpcom/attachments.js
tnajdek: I've moved to `prefs.js` and added a migration step 7 -> 8

test/tests/attachmentsTest.js
```diff
@@ -1293,11 +1293,296 @@ describe("Zotero.Attachments", function() {
...
+			);
+		});
+
+		it('should offer a range of options for composing creaotrs', function () {
+			assert.equal(
+				Zotero.Attachments.getFileBaseNameFromItem(item, '{{ authors max="1" }}'),
+				'Barius'
+			);
+			assert.equal(
+				Zotero.Attachments.getFileBaseNameFromItem(item, '{{ authors max="1" order="desc" }}'),
```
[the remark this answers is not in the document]
dstillman: Oh, I see. The problem with this is that, as far as I know, you might want the last author (senior author) but that's it, so using reverse order doesn't really seem appropriate — it's not like you'd ever want order=desc for anything else or with anything other than max=1. So I think we'd need some other mechanism here, if we were going to worry about it at all. But ZotFile doesn't seem to have had this, and I'm not sure how it even makes sense unless every single paper produced in your field follows this convention of senior authors, so I'd just skip it.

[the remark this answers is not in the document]
tnajdek: I think I implemented this because of [this comment](https://forums.zotero.org/discussion/comment/72254/#Comment_72254) but happy to remove it.

dstillman:
Let's call it `convertLegacyAttachmentRenameFormatString()` and clarify the comment. "legacy format string" is too vague.

test/tests/attachmentsTest.js
tnajdek: I'll tweak the template engine to allow this (currently value for comparison cannot be calculated by a function)

test/tests/attachmentsTest.js
tnajdek: I'll tweak the template engine.

test/tests/attachmentsTest.js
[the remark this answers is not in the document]
dstillman: I dunno — maybe if we had some other mechanism for last author specifically. I don't think last few authors is really a thing, and you certainly wouldn't want last few authors in reverse order. But any mechanism would also have to account for the possibility that all names had already been consumed so as not to repeat the last name.

test/tests/attachmentsTest.js
dstillman: Oh, I saw you say that, but I think I misunderstood the implication. In this case the value used for _comparison_ wouldn't be from a function — it'd be the raw item type — but I guess taking parameters means all values come from a function? (I forget exactly how this is implemented, but I guess that makes sense.)

test/tests/attachmentsTest.js
tnajdek: Updated the template engine to accept `dash-case` for variables and attributes. It converts both to `camelCase` in JS-world which I think is more elegant (and there is a precedent: DOM data attributes are converted to  `camelCase`)

test/tests/attachmentsTest.js
tnajdek: I've updated the template engine to accept functions in conditions and updated `getFileBaseNameFromItem` to accept`itemType` where we would have used `itemTypeRaw` earlier.

test/tests/attachmentsTest.js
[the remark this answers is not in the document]
tnajdek: Removed for now.

tnajdek:
@dstillman thanks for reviewing! I believe I've addressed all the comments, let me know if you can think of anything else.

test/tests/utilities_internalTest.js
```diff
@@ -569,6 +569,45 @@ describe("Zotero.Utilities.Internal", function () {
...
 
+		it("should support functions in comparison statements", function () {
+			const vars = {
+				sum: ({ a, b }) => parseInt(a) + parseInt(b),
+				fahrenheitToCelsius: ({ temp }) => Math.round((temp - 32) * 5 / 9),
+				false: 'false',
+				twoWords: 'two words',
+				onlyOne: 'actually == 1'
+			};
+			const template = `{{if sum a=1 b=2 == 3}}1 + 2 = {{sum a=1 b=2}}{{else}}no speak math{{endif}}`;
```
dstillman, line 580: This is kind of awkward and hard to read. Do we want to require some kind of delimiter when using parameters within conditions? An extra set of braces? Parens?

`{{if {{sum a=1 b=2}} == 3}}` (more consistent)

`{{if (sum a=1 b=2) == 3}}` (more concise, but introducing additional syntax)

Do we even need to support parameters in conditionals?

test/tests/utilities_internalTest.js
```diff
@@ -569,6 +569,45 @@ describe("Zotero.Utilities.Internal", function () {
...
+
+			const template2 = '{{ if false>false }}no{{elseif false <= 0}}no{{elseif false != false}}no{{elseif false == false}}yes{{else}}no{{endif}}';
+			const out2 = Zotero.Utilities.Internal.generateHTMLFromTemplate(template2, vars);
+			assert.equal(out2, 'yes');
+
+			const template3 = '{{ if twoWords == two words }}yes{{else}}no{{endif}}';
+			const out3 = Zotero.Utilities.Internal.generateHTMLFromTemplate(template3, vars);
+			assert.equal(out3, 'yes');
+
+			const template4 = '{{ if onlyOne == actually == 1 }}yes{{else}}no{{endif}}';
```
dstillman, line 592: These last two don't seem good. I think we want to require single or double quotes for strings, as we do in the [documentation](https://www.zotero.org/support/note_templates). I don't really know what should happen for something like these, but it seems like a syntax error. Throw an error that gets logged?

And honestly I'm not sure we should even support anything other than string comparison with `==` and `!=`. This is a template engine, and the variables exist primarily to be interpolated into the final string. So I think even the `3` above should probably be `"3"`, and we should throw if a function returns something other than a string.

Maybe there'll be some reason for additional complexity in the future, and you could keep the code in a branch somewhere, but for now I feel like we should keep this as simple as possible.

test/tests/utilities_internalTest.js
tnajdek: > Do we even need to support parameters in conditionals?

We don't strictly need it, for the `itemType` case for this PR we just need to be able to use a function inside logical statement. However, outside of this very specific case, functions without parameters offer very limited functionality. I'd rather go with the nested `{{ }}` but I can remove support for parameters if you think we shouldn't support that?

test/tests/utilities_internalTest.js
tnajdek: Some of these tests are to cover edge-cases and attempts at misuse the templating engine, I'd hope no one would use them like this.

> These last two don't seem good. I think we want to require single or double quotes for strings, as we do in the [documentation](https://www.zotero.org/support/note_templates).

I agree and would prefer that, however, current version does not require quotes for attributes, are we OK changing that?

> Maybe there'll be some reason for additional complexity in the future, and you could keep the code in a branch somewhere, but for now I feel like we should keep this as simple as possible.

I'll remove the numeric parsing and comparisons and throw if functions return anything else than a string.

test/tests/utilities_internalTest.js
dstillman: I guess we can go with nested `{{ }}` in case we need it later and just not document it now (since there are no functions with parameters to document).

test/tests/utilities_internalTest.js
dstillman: > I agree and would prefer that, however, current version does not require quotes for attributes, are we OK changing that?

Yes, let's require.

tnajdek: I've tweaked the template engine so that single or double quotes are required for string literals. Without quotes, value is considered to be a "variable" and if not present in `vars` object, it will evaluate as empty string.

tnajdek:
I've implemented discussed changes, added more tests and rebased. I didn't need to touch `getFileBaseNameFromItem` code in this round of updates,  only template engine has been changed.

One thing that we did not discuss is support for arrays in vars. Current engine supports [testing if arrays are not empty](https://github.com/zotero/zotero/blob/d968e8ec338088c88b7ffb9195682e1658387863/chrome/content/zotero/xpcom/utilities_internal.js#L2149C2-L2151) and there is a demo in documentation that shows something similar so I've kept it, let me know if we want throw an error instead.

PR #2299, opened 2022-01-10
Add color and tag selector to annotations sidebar

mrtcode:
The tag and color selector can be tested now. Some issues to discuss:

1) There is no a convenient way to clear the filter. A separate button would be good, but I'm not sure where to put it. Alternatively it could be in context menu as it is in the original tag selector.
2) When filtering is active, some actions like annotation creation or selection become complicated because you no longer see it in the sidebar. Maybe a better indication that filtering is active would be useful, especially in those cases when only a color is selected, because it might be difficult to notice.
3) The final element that is necessary for this workflow is "Select all annotations" button. Or should it also be in context menu?

### DIFF
--- a/chrome/content/zotero/xpcom/reader.js
+++ b/chrome/content/zotero/xpcom/reader.js
@@ -560,6 +560,28 @@ class ReaderInstance {
 		popup.openPopup(element, 'after_start', 0, 0, true);
 	}
 
+	_openSelectorPopup(data) {
+		let popup = this._window.document.createElement('menupopup');
+		this._popupset.appendChild(popup);
+		popup.addEventListener('popuphidden', function () {
+			popup.remove();
+		});
+		let menuitem;
+		// Clear Selection
+		menuitem = this._window.document.createElement('menuitem');
+		menuitem.setAttribute('label', Zotero.getString('general.clearSelection'));
+		menuitem.setAttribute('disabled', !data.enableClearSelection);
+		menuitem.addEventListener('command', () => {
+			this._postMessage({
+				action: 'popupCmd',
+				cmd: 'clearSelector',
+				ids: data.ids
+			});
+		});
+		popup.appendChild(menuitem);
+		popup.openPopupAtScreen(data.x, data.y, true);
+	}
+
 	async _postMessage(message, transfer) {
 		await this._waitForReader();
 		this._iframeWindow.postMessage({ itemID: this._itemID, message }, this._iframeWindow.origin, transfer);
@@ -654,6 +676,10 @@ class ReaderInstance {
 				}
 				case 'openColorPopup': {
 					this._openColorPopup(message.data);
+					return;
+				}
+				case 'openSelectorPopup': {
+					this._openSelectorPopup(message.data);
 					return;
 				}
 				case 'openURL': {
--- a/chrome/content/zotero/xpcom/reader.js
+++ b/chrome/content/zotero/xpcom/reader.js
@@ -366,6 +366,7 @@ class ReaderInstance {
 
 	_openTagsPopup(x, y, item) {
 		let menupopup = this._window.document.createElement('menupopup');
+		menupopup.className = 'tags-popup';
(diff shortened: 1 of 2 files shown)


dstillman:
1. Add a tag to a single annotation.
2. Click the tag.
3. Delete the tag from the annotation.

The tag will disappear from the tag selector but remain selected, so all annotations will disappear from the sidebar with no way to deselect the tag.

dstillman:
Closing and reopening the sidebar should probably clear selections? There's an argument that you might just close it temporarily to do something and want to return to what you were doing, but you could be reopening it days later, at which point it would be very confusing if a filter (or whatever) were still active.

1 and 2 could be related — if we had some sort of indication that a filter were active, that could also include the element to clear it. But I don't have any great ideas. For now, a context menu on the selector background with a single "Clear Selection" option might be best.

> The final element that is necessary for this workflow is "Select all annotations" button. Or should it also be in context menu?

Not necessary. Just https://github.com/zotero/zotero/issues/2303.

dstillman:
Follow the same steps as [above](https://github.com/zotero/zotero/pull/2299#issuecomment-1010878640) on something other than the first annotation. Now the tag entry popup stays open even after the sidebar resets and the position of the underlying annotation changes. Technically the popup should probably close if the position of the current annotation is changing, but could also just close it if it's the last instance of that tag and the selection is being reset (which might not necessitate closing if another tag is also selected or there's only one annotation, but in practice probably usually will).

dstillman:
How is the order of colors being determined? Seems like it should follow either the order of colors in the document or the same order as the color drop-down (and the latter doesn't work if you have external annotations with custom colors, so maybe has to be the former).

mrtcode:
> How is the order of colors being determined? Seems like it should follow either the order of colors in the document or the same order as the color drop-down (and the latter doesn't work if you have external annotations with custom colors, so maybe has to be the former).

Predefined colors have the same order as the color drop-down. And for other colors we don't do anything specific, but they should appear in the same order as they are in the document because annotations are always sorted by `sortIndex` in pdf-reader.

dstillman:
This is what I'm seeing.

Screen Shot 2022-01-13 at 4 15 46 AM

dstillman:
Wait, sorry, wrong PDF. Hang on.

dstillman:
I can't find the PDF I was looking at before, but I guess it puts other colors after the predefined ones?

mrtcode:
> I can't find the PDF I was looking at before, but I guess it puts other colors after the predefined ones?

Yes.

dstillman:
OK, I guess that makes sense — and will probably match whatever we do with custom colors in the colorpicker.

dstillman:
Or, well, no, because then we won't want to use the document order. But for now document order is as good as we can do.

mrtcode:
In future, it would also be possible to use some smart algorithm that sorts colors by grouping similar colors. I.e. all colors close to red would go first, and so on. We had something like that in pdf-worker to assign to assign the closest predefined color. Although salad color was somehow mistakenly categorized as our yellow color…

> Technically the popup should probably close if the position of the current annotation is changing, but could also just close it if it's the last instance of that tag and the selection is being reset (which might not necessitate closing if another tag is also selected or there's only one annotation, but in practice probably usually will).

pdf-reader has no control over popups when they are opened. So, I guess, pdf-reader will need to send a command to the client to close that popup.

dstillman:
One more thing: when any tag or color is selected, unselected tags should turn gray if they don't match the current results. This currently happens for colors but not tags.

mrtcode:
> One more thing: when any tag or color is selected, unselected tags should turn gray (as in the main tag selector with "Display All Tags in This Library") if they don't match the current results, even though they remain selectable. This currently happens for colors but not tags.

Tags that don't appear in the current annotations list already become grayish. Could you explain once again in which situation tags still don't turn gray?

Screenshot 2022-01-17 at 11 13 00

dstillman:
Sorry, my testing was bad. It's working properly.

Issue #2252, opened 2021-11-29
Bulk folder import


dstillman:
We should integrate a version of https://github.com/retorquere/zotero-folder-import:

1) Show a filepicker with mode `modeGetFolder`
2) Show options:
  a) "Place imported collections and items into new collection"
  b) Adding as stored (default) or linked files
  c) Recreating the folder structure as collections
  d) Could eventually add an option for including or excluding file types by file extension, but we can just add all non-hidden files for now
3) Add all files
4) Run metadata retrieval on all PDFs

The first two options are already in the main import wizard, so it would make some sense to reuse that and have this be another import option on the first pane. We've never referred to "importing" attachments, but in this case you're sort of importing from a different system. Alternatively, it could be a separate option (e.g., File → "Add Files from Folder…"), which avoids conflating "import" and "adding files".

If we use the import wizard, I'm not sure how we show progress for metadata retrieval. Can we integrate the progress component into a pane in the import wizard?

tnajdek:
I'll take a crack at this. It probably makes sense to build this on top of #1849 so we don't have to port it again later when we merge new wizards, does that sound good?

PR #2311, opened 2022-01-20
fx-compat: Wizards + Bulk folder import

tnajdek:
This PR adds bulk folder import #2252.

* Implemented as a new option in the Import Wizard
* Optionally recreates folder structure as collections
* Progress queue is displayed as part of the import wizard
* Avoids creating duplicate collections (same name, same parent) on subsequent imports.
* Avoids creating duplicate items (whether due to duplicate files in the folder or because of prior import)
* By default imports PDFs, attempts to auto-recognize
* Other file types can be added using glob pattern (e.g. `*.png`)
* Built on top of #1849, includes all commits from that PR as well

Potential issues:

* If there are other PDFs in the recognize queue already, this import will most likely wait until all are finished (even though imported PDFs are added at the beginning of the queue). If that's a problem we may want to extend `RecognizePDF` to allow waiting only for newly inserted items
* ~~I'm installing a listener to `rowupdated` "event" in recognize queue, however it seems current implementation of queues only supports single listener so this has potential of overriding any that already is there. This shouldn't be a problem as long as we never allow opening import wizard while `ProgressQueueDialog` is open. Alternatively we could tweak queue implementation to support multiple listeners.~~ (fixed in #2318)


dstillman:
> Do we want this to handle files other than PDF. Would that part work just like normal import?

I don't think we should do all files even as an option — I could see that easily going awry in all sorts of ways, only some of which we could predict (e.g., a Mac .app, a git repo) — but I don't think we want to privilege PDFs quite so much. We could add a text field that let you specify filename globs (e.g., "*.jpg"):

```
File Types to Import:

[X] PDFs
[ ] _____________ (with placeholder with example)
```

Text field checkbox would get checked if you started typing.

> Do we want to store file hashes to prevent duplicates on re-import?

I think we can skip that, at least for now. zotero-folder-import doesn't do it, and this is really intended to be a one-time operation. Even if you're linking files, going forward you set ZotFile to point to the same folder and save via Zotero normally.

Other things:

- "A folder with PDFs" → maybe "A folder of PDFs or other files", and move below "A file"
- There's a font size inconsistency in the settings pane between "Place imported collections…" and "Recreate folder structure…".
- The progress window should pop up automatically after the import. Most people won't notice the toolbar button.

tnajdek:
Thanks for the feedback, much appreciated! I've started working on preventing duplicates before your message so I've added that in. 

> ...We could add a text field that let you specify filename globs (e.g., "*.jpg"):

Ok, I'll add something like that

> ...and move below "A file"

Should be the last option at the bottom already, do you see something else?
Screenshot 2022-01-25 at 16 23 23

> There's a font size inconsistency...

Good spot, I'll fix it.

> The progress window should pop up automatically after the import. Most people won't notice the toolbar button.

You mean progress of PDF recognition?  I'm subscribing to the progress queue so progress is tracked by the importer and by the time that finishes there shouldn't be anything else going on, I think?

dstillman:
> You mean progress of PDF recognition? I'm subscribing to the progress queue so progress is tracked by the importer and by the time that finishes there shouldn't be anything else going on, I think?

Yes, progress of PDF recognition. I just mean pop up that little window so that it's visible on screen after the import and you don't have to click on the toolbar button to see the results of the PDF recognition. Right now that popup seems to be closed after the import (or maybe it's just popping up behind the main Zotero window — I don't have this running at the moment so can't check).

I'll take a look at the duplicate-items stuff.

tnajdek:
@dstillman I think I pretty much got this to where I wanted it, would appreciate another review.

tnajdek:
Brought this up to date with current `fx102` as we've accumulated some conflicts.

dstillman:
On opening File → Import (after rebasing):

radios

dstillman:
I pulled out the fixed error-reporting wizard so that we can make that available to developers, so this will need to be rebased to remove that commit here.

tnajdek:
Deprecated, see #2862

Issue #1986, opened 2021-03-08
Improve keyboard support in PDF reader


dstillman:
Need to be able to get to annotations and notes via keyboard, close panes, switch tools, etc.

mrtcode:
Any idea how that should work exactly? Are we talking about standard TAB, ENTER, arrow keys navigation or specific keyboard shortcuts as well?

bjohas:
I've built a few keyboard shortcuts into Zutilo - so I suppose I would advocate for making sure that as many features as possible (e.g. hide/show pane, select tool / select colour etc) can be reached via Zutilo. Then it's less critical what is developed directly in the PDF Reader vs. what can be offloaded to Zutilo.

But yes, if the shortcuts can be built into PDF Reader directly, all the better!

qqobb:
Here's a list of pdf.js viewer shortcuts:
https://github.com/mozilla/pdf.js/wiki/Frequently-Asked-Questions#faq-shortcuts

mrtcode:
Currently all toolbar buttons, page thumbnails and outline elements can be reached with tab. This is the default PDF.js behaviour.

What we still need:
- Switch between pdf-reader and right-hand side pane (including the split button? sync button?)
- Navigate annotations with tab? (that is currently possible with arrow keys when one of them is selected)
- Navigate links in PDF file (I had to disable that, doesn't seem useful, or is it?)
- Open tags and context menu popup for selected annotation
- Various other shortcuts

Further suggestions are welcome.

qqobb:
> Navigate links in PDF file (I had to disable that, doesn't seem useful, or is it?)

I think this can be very useful for academic PDFs, but only in combination with https://github.com/zotero/zotero/issues/1997. To see this, open [this PDF](https://arxiv.org/pdf/2007.07289.pdf#page=3) in Firefox; Tab, the focus will be on the first reference; Enter, go to the reference; Alt+Left, go back to the original text position (on Windows). It's very convenient for checking all the references without using the mouse.

(BTW, in Zotero 5.0.97-beta.6+ae98488d2, link navigation is not completely switched off. You can get to it with Shift+Tab: Click space to the right of the total page number; press Tab once; after the thumbnail button gets focus, keep pressing Shift+Tab.)

Incidentally, when opening a new PDF, the initial focus should probably be on the page content, such that you can immediately start using the Page Up and Page Down keys.

Let me add some papercuts for the search function. I hope it's ok to mention this here.

[...]

PR #2317, opened 2022-01-25
Implement custom keyboard navigation

mrtcode:
As discussed in #1986.

It took a lot of time and there were many difficulties to implement this. Each focus zone has its own nuances. Basically a virtual focus layer was implemented. I tried to keep things as simple and as maintainable as possible, but the way how it works is still not perfect and will need a partial rewrite when we migrate to a newer PDF.js, or new UI elements will be introduced.

So this is what was implemented:
- Sidebar search input is now fixed to the top.
- Toolbar navigation.
- Findbar navigation.
- Sidebar view buttons navigation.
- Page thumbnails navigation.
- Outline navigation (Enter to fold/unfold)
- Annotations navigation and selection that respects the current filter.
- Annotation widget navigation.
- Page label updating popup navigation.
- Tag and color selector navigation.
- Annotations in PDF view navigation, which also allows to focus the first visible annotation when sidebar is closed.
- PDF view navigation that is native PDF.js navigation.
- Text selection popup navigation to pick highlight color (when text is selected, press Tab to focus to the first color).

We can consider slightly different variations of this navigation, especial for the part where focus becomes annotation selection. Maybe instead of automatically selecting annotation we only want to focus it and show an outline around it.

There is a high chance that there are bugs.

As soon as we approve the final version of this, I’ll do more code simplifications, cleanups and add comments to save time in future when I’ll need to get back to this complex thing.

EDIT: More points added.

### DIFF
--- a/chrome/content/zotero/xpcom/reader.js
+++ b/chrome/content/zotero/xpcom/reader.js
@@ -365,7 +365,7 @@ class ReaderInstance {
 		return `data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16"><rect shape-rendering="geometricPrecision" fill="${fill}" stroke-width="2" x="2" y="2" stroke="${stroke}" width="12" height="12" rx="3"/></svg>`;
 	}
 
-	_openTagsPopup(x, y, item) {
+	_openTagsPopup(item, selector) {
 		let menupopup = this._window.document.createElement('menupopup');
 		menupopup.className = 'tags-popup';
 		menupopup.style.minWidth = '300px';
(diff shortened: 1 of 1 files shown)


dstillman:
```
TypeError: Argument 1 of Node.contains does not implement interface Node.

    editor_Content/<@resource://zotero/pdf-reader/annotator.js:2411:16

JavaScript error: resource://zotero/pdf-reader/annotator.js, line 2411: TypeError: Argument 1 of Node.contains does not implement interface Node.
```
I get a bunch of these when I press right-arrow from the Next Page button to the Page text field.

dstillman:
This is great. There are some things we might want to refine, but this is totally usable for a first version. Other than fixing that error, I think we should merge this now and refine later.

(Inability to get into the item pane isn't ideal.)

mrtcode:
The error is fixed, although it wasn't related to this PR.

PR #2324, opened 2022-02-03
Fix React icons only using 2x image once

AbeJellinek:
a1267bc fixed repeated re-appending of `'@2x'` by moving the modified URL to a local variable, but it still set `hasHiDPI` to false permanently; this meant that icons displayed repeatedly in the same window, like the tab close button, would only use their 2x versions the first time they were displayed (in the case of tabs, this is the library tab's invisible close button, so *no* tabs got visible 2x buttons).

This commit moves the HiDPI check and URL modification out of the render function and instead runs it a single time when `i()` is first called.

Before (upscaled 1x tab close button):
Screen Shot 2022-02-02 at 4 59 35 PM

After:
Screen Shot 2022-02-02 at 5 00 42 PM

### DIFF
--- a/chrome/content/zotero/components/icons.jsx
+++ b/chrome/content/zotero/components/icons.jsx
@@ -23,25 +23,22 @@ Icon.propTypes = {
 module.exports = { Icon }
 
 
-function i(name, svgOrSrc, hasHiDPI=true) {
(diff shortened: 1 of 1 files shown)


dstillman:
Good catch!

PR #2325, opened 2022-02-03
(de) translated strings for pdf reader, spelling & wording fixes

rwbru:
checked spelling & wordings in the whole doc, added translations for missing strings, added translations for 'annotations / PDF reader'


AbeJellinek:
Thank you! Zotero doesn't actually use GitHub for localization contributions, though - if you're able, please join the Zotero Transifex project and input your translations there. [See more info on the wiki.](https://www.zotero.org/support/dev/localization)

PR #2336, opened 2022-02-08
Merge attachments and update notes

### LINKED ISSUE #61: Merge identical attachments when detecting duplicates
http://forums.zotero.org/discussion/22025/merging-documents-zotero-303/

A potential pitfall here is that many websites timestamp the PDF, so the first page won't be identical, even if subsequent pages are and the content is the same. We might want to come up with a good heuristic to detect equality in these cases. One option would be to use pdftotext or pdf.js. Another option would be to assume that attachments with the same name are identical.

AbeJellinek:
We follow a different merge procedure for each attachment type:

- For PDF attachments, compare by MD5. If no match, get the top 50 words in the attachment's text and hash those (order-sensitive), then check again for a match. Update references to item keys in notes and annotations.
- For web (snapshot / link) attachments, compare by title and URL. Prefer a title + URL match but accept a title-only match.
- For other attachment types, keep all attachments from all items being merged.

I believe that the 50-word hashing approach should be unlikely to produce a hash collision in normal usage, since even a difference in the ordering of the top 50 words will produce a different hash, and it is highly effective at detecting PDFs that should be merged even when database watermarks have been added.

Closes #61.


dstillman:
> get the top 50 words in the attachment's text and hash those (order-sensitive)

Why order-sensitive? If you have two words in the list separated by one occurrence, and an intro page is added that includes one of those words twice, that shouldn't prevent the match. It seems like having the same top-50 words would be sufficient enough.

dstillman:
Did you test the 50-word threshold on some shorter PDFs? E.g., 5 pages?

AbeJellinek:
>Why order-sensitive?

I was thinking more about matching two different documents from the same database (same cover page format, different contents), but that's a good point.

>Did you test the 50-word threshold on some shorter PDFs? E.g., 5 pages?

Yup! Seems to work well on some random PDFs from JSTOR.

chrome/content/zotero/xpcom/data/notes.js
```diff
@@ -96,7 +96,36 @@ Zotero.Notes = new function() {
...
+	 *
+	 * @param {Zotero.Item} item
+	 * @param {Map<String, String>} itemKeyMap
+	 */
+	this.replaceAllItemKeys = function (item, itemKeyMap) {
+		let note = item.getNote();
+		let keys = [...itemKeyMap.keys()].join('|');
+		let re = new RegExp(`%2Fitems%2F(${keys})`, 'g');
+		note = note.replace(re, (str, key) => `%2Fitems%2F${itemKeyMap.get(key)}`);
+		re = new RegExp(`data-attachment-key="${keys}"`);
```
dstillman, line 113: Missing parentheses, so this would be matching, e.g., `data-attachment-key="foo` or `bar"`

And let's add a test with at least three keys so we would've caught this.

chrome/content/zotero/xpcom/data/items.js
```diff
@@ -1054,7 +1061,260 @@ Zotero.Items = function() {
...
+	};
+
+
+	/**
+	 * Hash an attachment by the most common words in its text.
+	 * @param {Zotero.Item} attachment
+	 * @return {Promise<String>}
+	 */
+	this._hashAttachmentText = async function (attachment) {
+		if ((await OS.File.stat(await attachment.getFilePathAsync())).size > 12000000) {
```
dstillman, line 1240: Any particular reason for this? We should keep it as big as possible, until it starts to hang the UI.

dstillman (changes requested):

chrome/content/zotero/xpcom/data/items.js
```diff
@@ -1054,7 +1061,260 @@ Zotero.Items = function() {
...
+				// URLs pointing to the old attachment.
+				remapAttachmentKeys.set(otherAttachment.key, masterAttachment.key);
+
+				// Items can only have one replaced item predicate
+				if (!masterAttachment.getRelationsByPredicate(Zotero.Relations.replacedItemPredicate)) {
+					masterAttachment.addRelation(Zotero.Relations.replacedItemPredicate,
+						Zotero.URI.getItemURI(otherAttachment));
+				}
+
+				savePromises.push(masterAttachment.save());
```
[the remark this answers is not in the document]
AbeJellinek: The whole merge operation is wrapped in a transaction, but I'll add calls to `requireTransaction()` to be sure.

chrome/content/zotero/xpcom/data/items.js
AbeJellinek: Not really. Just a holdover from an earlier hashing method I tried that was really slow, and I hadn't tested any really big files since.

dstillman:
Not that common, but we should transfer related-item relations (`relatedItemPredicate`) to the preserved attachment.

PR #2338, opened 2022-02-09
Add File Size dialog and show size in attachment box

### LINKED ISSUE #271: Show attachment file size
http://forums.zotero.org/discussion/15487/

Snapshots should show total of all files.

File size should be cached to avoid repeated filesystem hits.

AbeJellinek:
Exposed as an optional item tree column and always shown in the attachment box. Getting sorting to work reliably (particularly when switching between libraries) required refactoring the item tree's caching logic slightly.

Closes #271.


dstillman (commented):
Sorting isn't working right for me:

file size

But I think we need to reconsider this. The attachment box is fine, but I don't think we can add this to the main items list — if people leave this on, it will drastically decrease Zotero's performance after every startup. Just turning it on makes scrolling sluggish for me and makes my fans spin up, and sorting by that column would necessitate calculating the value for every column. There's also no reason to sort this by size ascending, and no great need for filtering or for viewing by collection.

I think it'd make more sense to create a new option in Tools → Manage Attachments ("Disk Usage" or whatever) with a library drop-down (`buildLibraryMenuHTML()`) that showed a window with a new tree that showed top-level items (title/creator only?) sorted by size descending. We could perhaps show attachments underneath each one, also size descending, but I don't think that's particularly necessary. Activating (e.g., double-clicking) the row would select the item in the main window, like Advanced Search.

We could later add options for deleting or purging, but that's a little tricky — sending to the trash wouldn't change anything, and we can't do local purging at least until we finish #1919 (which has a [library-level storage breakdown](https://github.com/zotero/zotero/pull/1919#issuecomment-773430096) of its own) and can know if a file is safely on the server.

For now, the most important thing would just be letting people see the largest items in their libraries.

dstillman:
In the attachment box, "0 B" is a little confusing when the file isn't present. "N/A"?

In a dedicated window, we could just not show items without files or with missing files.

chrome/content/zotero/bindings/attachmentbox.xml
```diff
@@ -302,6 +309,19 @@
...
+						this._id("fileSize-label").value = Zotero.getString('pane.item.attachments.fileSize')
+							+ Zotero.getString('punctuation.colon');
+						this._id("fileSize").value = Zotero.getString('general.loading');
+						fileSizeRow.hidden = false;
+						this.item.getFileSize().then((fileSize) => {
+							this._id("fileSize").value = Zotero.File.formatFileSize(fileSize);
+						});
+					}
+					else {
+						fileNameRow.hidden = true;
```
dstillman, line 323: `fileSizeRow`

dstillman:
I don't think it was necessary to take out the caching. This way, "Loading…" flickers every time you select an attachment. Instead, we should cache the value, display it immediately if present, and get asynchronously after that (updating the cached/display value if changed). Again, similar to Get Info on a folder — the first time it shows "Calculating", but after that it shows the cached value immediately and updates a moment later if the value has changed.

dstillman:
(And we'll want the caching anyway when we display this in a separate window, so that you don't have to wait for it to scan all files if you reopen the window.)

AbeJellinek:
Re-added caching of the raw values. There's still some flicker, though, because the `Zotero.Item` object doesn't actually seem to be the same each time the attachment box is loaded. It'll be the same with a separate disk usage window - populating the list takes a call to `Zotero.Items.getAll()`, which returns fresh `Zotero.Item` objects. Is there a non-hacky way to persist cached values between loads without touching the database?

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -416,6 +419,11 @@ pane.item.attachments.select			= Select a File
 pane.item.attachments.PDF.installTools.title	= PDF Tools Not Installed
 pane.item.attachments.PDF.installTools.text		= To use this feature, you must first install the PDF tools in the Search pane of the Zotero preferences.
 pane.item.attachments.filename			= Filename
+pane.item.attachments.fileSize			= Size
+pane.item.attachments.sizeSuffix.bytes			= B
+pane.item.attachments.sizeSuffix.kilobytes			= KB
+pane.item.attachments.sizeSuffix.megabytes			= MB
+pane.item.attachments.sizeSuffix.gigabytes			= GB
```
dstillman, line 426: No longer needed, right?

chrome/content/zotero/diskUsage.jsx
```diff
@@ -0,0 +1,172 @@
...
+
+	renderItem = (index, selection, oldDiv = null, columns) => {
+		let row = this.state.rows[index];
+		let div;
+		if (oldDiv) {
+			div = oldDiv;
+			div.innerHTML = "";
+		}
+		else {
+			div = document.createElementNS("http://www.w3.org/1999/xhtml", 'div');
```
dstillman, line 55: Move to `const HTML_NS` up top and use for all

dstillman (changes requested):
Nice!

A few things:

- We need some sort of feedback in the list while sizes are being calculated. Right now it's impossible to know anything is happening. We could just do a determinate progress bar (since we know the number of parent items) in the center, or we could update the list progressively (and maybe also show a progress bar or spinner somewhere in that case?).
- We need some sort of indication when calculations are done but there's nothing to show in a given library.
- Activating rows doesn't select items in the main window for me. It just brings the main window forward.

For the attachment box, what about this?

> In the attachment box, "0 B" is a little confusing when the file isn't present. "N/A"?

I'll look into the "Loading…"/Zotero.Item thing.

Are we currently caching file sizes as long as the app is open? That might be fine, but what happens if you delete an attachment on disk? Seems like we should at least check for that (even if we start by showing the cached value in the various places).

qqobb:
I have two comments regarding `formatFileSize`.

1. In Windows Explorer, I'm getting the file size column in KB for all files. This seems like a reasonable choice: (i) Sorting by file size will just sort numbers. (ii) Changing units could be distracting. (iii) It's easy to spot large files. Perhaps you could consider a similar approach in the item tree.

```
 6,105 KB
    12 KB
     1 KB
```

Assuming that most files are in the MB range, MB could be the common unit:

```
 12.55 MB
  1.52 MB
  0.13 MB
```

Maybe use `formatFileSize(size, format)` for that?

2. In the file properties, accessed through the context menu, Windows shows both a short and a long format, e.g., `1.82 MB (1,915,981 bytes)`. It seems they're using 1024 based units for [Megabyte](https://en.wikipedia.org/wiki/Megabyte). In case you want to support that, you could use the `formatSize` function here: https://stackoverflow.com/a/40879101. Your current approach seems to follow SI units, which should be ok, but I'm not sure how common this is.

AbeJellinek:
>In Windows Explorer, I'm getting the file size column in KB for all files. This seems like a reasonable choice: (i) Sorting by file size will just sort numbers. (ii) Changing units could be distracting. (iii) It's easy to spot large files. Perhaps you could consider a similar approach in the item tree.

(i): This doesn't really gain us anything, since we always have the size in bytes and only format before displaying to the user. Sorting uses the numerical size, not lexical order (not useful for numbers). (ii): That's fair, but sizes are now displayed in a separate "Disk Usage" dialog that sorts by size descending, so "GB" items will always be above "MB" items will always be above "KB" items. Pretty easy to parse visually. (iii): That's fair, but I think the always-sorted Disk Usage dialog makes spotting large items easy enough.

>Assuming that most files are in the MB range, MB could be the common unit

Definitely worth considering.

>In the file properties, accessed through the context menu, Windows shows both a short and a long format ... It seems they're using 1024 based units for [Megabyte](https://en.wikipedia.org/wiki/Megabyte).

I was modeling my approach off the macOS Finder, which uses SI prefixes:

image

I guess we could make the prefix definitions depend on the user's platform.

AbeJellinek:
OK, I think this is ready.

gutjuri:
When will this be merged?

LSartore:
Hi, what is the status of this pull request ?

PR #2340, opened 2022-02-09
Disable Delete/Restore menu items appropriately

### LINKED ISSUE #2301: Can't purge child items of deleted parent items from trash
https://forums.zotero.org/discussion/93881/can-not-delete-multiple-files-in-the-trash-in-a-batch-operation

1) If you delete a parent item, the child item appears in the trash (because it has to) and appears as black (because it would be weird for it to be gray), but since that item isn't actually `.deleted`, it's not possible to purge it individually from the trash.

2) Select All in the trash *does* select those child items, so you also can't just press Select All and delete the selected items.

You should be able to purge child items of items with `.deleted`, and Select All shouldn't select those child items (but they should remain black).

Tested in the beta — not sure if this has changed from Z5.

AbeJellinek:
An item can only be deleted from the trash if it or one of its ancestors is `.deleted`. An item can only be restored from the trash if it is `.deleted`.

Fixes #2301.


chrome/content/zotero/zoteroPane.js
```diff
@@ -1893,6 +1909,21 @@ var ZoteroPane = new function()
...
+	 *
+	 * @return {Boolean}
+	 */
+	this.canRestoreSelectedItems = function () {
+		let collectionTreeRow = this.getCollectionTreeRow();
+		if (!collectionTreeRow.isTrash()) {
+			return false;
+		}
+
+		return this.getSelectedItems().every(item => item.deleted);
```
dstillman, line 1925: This isn't quite right. If you select a parent item and a child item in the trash, "Restore to Library" should be enabled, even if one of them isn't technically deleted.

chrome/content/zotero/zoteroPane.js
```diff
@@ -1893,6 +1909,21 @@ var ZoteroPane = new function()
 		}
 		this.selectItem(item.id).done();
 	}
+
+
+	/**
+	 * Return whether every selected item is deleted.
```
dstillman, line 1915: Instead of having "Return" here (which would get real repetitive and is redundant with `@return`) and explaining the actual test (which could change), this can just be a more general description of what the function does. E.g., "Check whether all selected items can be restored from trash".

Also, nit: No need for periods at the end of comments

And same for the other function

dstillman (changes requested):
We should also show the appropriate item pane buttons based on these tests, so that they match the menu options. E.g., right now they're not visible when a child item of a deleted parent item is selected, even though the "Delete Permanently…" menu option is now available.

While we're at it, the menu option in the trash should really be "Delete Permanently…" (like the button) rather than "Delete Items…", which would also help explain why it's disabled if a parent item that isn't in the trash is selected.

chrome/content/zotero/zoteroPane.js
AbeJellinek: I'd like to get the logic right here, but it depends on what we're going for - should we only show Restore if it will cause everything you have selected to disappear from the trash? If so, we also need to check whether any of the selected items has a descendant that is deleted and not selected, since that would mean that part of your selection will remain in the trash after Restore. If we only care whether it'll clear _some_ of the selected items from the trash, the test can be simpler.

dstillman: - Deleted/selected parent, deleted/selected child, no other children → show Restore, undelete both
- Non-deleted/selected parent, deleted/selected child, no other children → show Restore, undelete child (because there's no reason not to)
- Non-deleted/selected parent, deleted/selected child, deleted/non-selected children → show Restore, undelete selected child, don't undelete non-selected children (because the intention is clear)
- Deleted/selected parent, deleted/non-selected child, non-deleted/non-selected child → show Restore, restore parent and deleted child (because there's no way to know the user's intention and this is safer)
- Deleted/selected parent, non-deleted/selected child, deleted/non-selected child → show Restore, restore parent only, don't undelete non-selected child (because selecting a specific child communicates intention)
- Non-deleted/selected parent, deleted/non-selected child → don't show Restore (because they didn't select anything non-gray)
- Non-deleted/non-selected parent, deleted/selected child, non-deleted/selected child → show Restore, undelete deleted child (because non-deleted items in the selection don't prevent Restore)

dstillman: (Let's create tests for all of these.)

AbeJellinek: See latest commits!

AbeJellinek:
Updated restore behavior and replaced "Delete Items..." with "Delete Permanently...". I kept the `.multiple` variant around because there are languages in which that would be translated with a direct object pronoun that could be singular or plural.

dstillman:
> I kept the .multiple variant around because there are languages in which that would be translated with a direct object pronoun that could be singular or plural.

Do you have an example? I wouldn't think this is true. We [currently](https://github.com/zotero/zotero/pull/1535) use direct objects in all our menu options, but if we didn't, I wouldn't expect to have plural variants of every option that were just based on how many items you had selected. The Mac Finder doesn't use direct objects, and if your system is in French, it doesn't change every menu option based on the number of files selected — it just uses the infinitive:

Screen Shot 2022-02-12 at 3 11 15 AM

Screen Shot 2022-02-12 at 3 11 08 AM

(In checking to see if this was universal, I found a [discussion](https://forum.wordreference.com/threads/en-edit-item-infinitive-imperative.2627307/) among French speakers about what the actual verb tense is in English menus, the answer being…¯\\\_(ツ)\_/¯.)

AbeJellinek:
>(In checking to see if this was universal, I found a [discussion](https://forum.wordreference.com/threads/en-edit-item-infinitive-imperative.2627307/) among French speakers about what the actual verb tense is in English menus, the answer being…¯\_(ツ)_/¯.)

Good to know that we can’t reuse button titles for window titles, too.

dstillman:
I'm not sure what verb forms for these are, but it doesn't look like they change:

Screen Shot ٢٠٢٢-٠٢-١٣ at ١ ٣٥ ٣٠ ص

Screen Shot 2022-02-13 at 1 35 38 AM

I suspect this is pretty universal as a UI convention, so I think can remove the `multiple` variant.

(Even practically speaking it wouldn't really work, because you have to go out of your way to even see the key on Transifex, so everyone translating these identical English strings would just translate them as the same string.)

chrome/content/zotero/zoteroPane.js
```diff
@@ -1893,26 +1909,74 @@ var ZoteroPane = new function()
...
+				if (isSelected(parent)) {
+					if (parent.deleted) {
+						parent.deleted = false;
+						savePromises.push(parent.save());
+					}
+
+					let children = [...parent.getNotes(true), ...parent.getAttachments(true)];
+					let noneSelected = !children.some(isSelected);
+					for (let child of Zotero.Items.get(children)) {
+						Zotero.debug('Child: ' + child.id, -1)
```
dstillman, line 1960: Remove

chrome/content/zotero/zoteroPane.js
```diff
@@ -1893,26 +1909,74 @@ var ZoteroPane = new function()
...
+						if (isSelected(child) && child.deleted) {
+							child.deleted = false;
+							savePromises.push(child.save());
+						}
+					}
+				}
 			}
-		}.bind(this));
-	});
+			await Promise.all(savePromises);
```
dstillman, line 1977: We don't do this for object saves. It might be fine, or it might cause weird race conditions, but we don't design or test for it anywhere, and in any case it probably doesn't speed anything up. Can just use `await item.save()` within the loop, despite the eslint warning (which we should probably remove).

This makes more sense for things like network requests where a single request could take anywhere from 0 to many seconds and we don't want other requests to wait, but that's not an issue here.

dstillman (changes requested):

dstillman:
Possible regression?

https://forums.zotero.org/discussion/94771/failed-to-restore-to-the-library

PR #2347, opened 2022-02-12
Fix item tree jumping to top on focus without selection

AbeJellinek:
This fixes a bug that caused the item tree to scroll to the very top if you focused it without selecting an item, e.g. by clicking on an item's twisty while focus is outside the tree.

### DIFF
--- a/chrome/content/zotero/components/virtualized-table.jsx
+++ b/chrome/content/zotero/components/virtualized-table.jsx
@@ -116,9 +116,10 @@ class TreeSelection {
 	 * Selects an item, updates focused item to index.
 	 * @param index {Number} The index is 0-clamped.
 	 * @param shouldDebounce {Boolean} Whether the update to the tree should be debounced
+	 * @param {Boolean} [shouldScroll=true] Whether to scroll the item into view
 	 * @returns {boolean} False if nothing to select and select handlers won't be called
 	 */
-	select(index, shouldDebounce) {
+	select(index, shouldDebounce, shouldScroll = true) {
 		if (!this._tree.props.isSelectable(index)) return;
 		index = Math.max(0, index);
 		if (this.selected.size == 1 && this._focused == index && this.pivot == index) {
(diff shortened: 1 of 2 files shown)


adomasven:
The select on focus behavior was changed/added here https://github.com/zotero/zotero/pull/2263#issuecomment-986597736

I think we should only select the first item (and tree should scroll) when navigating via keyboard for accessibility reasons. If the focus comes from a mouse click and no item was focused because of that click, then we shouldn't focus anything either. Usually you would do this by using the `focusEvent.detail` property, but since we're calling `elem.focus()` and issuing non-organic focus events, it won't do. We could use `focusEvent.relatedTarget` and only focus when the related targets are elements that you would reach with [Tab] and [Shift-Tab], but I don't know how I feel about that. Any other ideas?

AbeJellinek:
This is a bit of a hack, but since the focus event fires between `mousedown` and `mouseup`, could we just check whether the mouse is currently down and only select the first row if it isn't?

adomasven:
> Is there any focus shortcut but Tab that we should be listening for?

VoiceOver on macOS provides some custom shortcuts to navigate UIs, but I have no idea what events they fire.

AbeJellinek:
I've been testing a bit with VoiceOver, and it doesn't seem like it fires events in Zotero at all, focus or otherwise. Its navigation shortcuts move the overlay around the screen, but elements don't actually get focus unless you use tab and other standard, non-VoiceOver shortcuts to navigate. That isn't the behavior when viewing pages in Firefox - as VoiceOver navigates around the interface on this page, for instance, buttons become visually focused and can be activated with VoiceOver shortcuts, space, enter, etc.

That's all to say that we have a ways to go with VoiceOver support, but this PR doesn't cause any regression in accessibility on macOS. I can work on the focus issue and some other stuff I noticed (like the item tree always being read as "empty") in a separate PR.

dstillman:
> I can work on the focus issue and some other stuff I noticed (like the item tree always being read as "empty") in a separate PR.

(You can try, but I think VoiceOver support was just pretty buggy in this version of Firefox. I believe it's improved a bit in later versions.)

adomasven:
> That's all to say that we have a ways to go with VoiceOver support, but this PR doesn't cause any regression in accessibility on macOS.

Yeah, I think this can be merged. Good work!

dstillman:
Looks like this needs to be rebased.

AbeJellinek:
Done.

PR #2357, opened 2022-02-17
Delay querying attachment state from DB on tree scroll

adomasven:
Improves item tree fast-scrolling performance when attachment column is enabled, but adds a noticeable attachment state pop-in.

I didn't want to commit this immediately since perhaps we find the pop-in not desirable, or want to tweak the delay option.


dstillman:
How about just not triggering this while scrolling is ongoing? That would let you scroll through the whole list quickly with the scrollbar or inertial scrolling. Once scrolling stops, it seems OK to run this immediately.

adomasven:
> How about just not triggering this while scrolling is ongoing?

`Scroll is ongoing` is technically defined as "no scroll event has occurred for X time", and these functions are called (primarily) on scroll. But also we don't expose (and probably don't want to expose) scroll events so high up abstraction hierarchy, especially as they don't gain us much here.

> That would let you scroll through the whole list quickly with the scrollbar or inertial scrolling.

Unless you're saying that this doesn't work for you smoothly because of `item.getBestAttachmentState()` calls? If you're scrolling fast they shouldn't be triggering, so if it's not smooth it's probably something else. You can check by disabling the attachment column completely.

dstillman:
> If you're scrolling fast they shouldn't be triggering

Oh, right, because of the `!this.tree.rowIsVisible(index)`.

OK, this is fine. Let's go with 150 to start — that speeds things up a little bit while mostly keeping the SQL queries from appearing for me while scrolling.

adomasven:
Updated

PR #2359, opened 2022-02-17
Add support for annotation templates

mrtcode:
- Added three new prefs.
- Image annotation uses a hard-coded template that is not added to prefs.
- I'm thinking if `{{text}}` should be `{{highlight}}`, because it is actually inserting not just a text but `highlight` node as well.
- Shouldn't it be allowed to enable/disable colors over the template? Or a separate pref will be used for that?
- For the title template it sets a variable named `string.pdfReader.annotation`, does it makes sense to have the `string.` prefix?
- It seems that `noteEditor.annotationsDateLine` is no longer necessary, if we have the title template now: https://github.com/zotero/zotero/commit/f399b643b4e26e00035163ac98d2413e670df0fc#diff-0a446470c46e57216a7fcb4cbc96f96f5a17c0de214ced33df2f771bee553d89L1468
- I ended up just disabling automatic unlink when a highlight node is inside a blockquote. No additional blockquote node type was introduced.
- `note-editor` schema version number is increased, to prevent older clients unlinking highlights in blockquotes.
- If you configure a template to put `{{text}}` anywhere else, it will just become a text. I.e. `<h1>{{text}}</h1>`.
- `note-editor` now supports more quotation marks: `["'",'"', '“', '”', '‘', '’', '„','«','»']`.
- Variables available in note and highlight templates: https://github.com/zotero/zotero/commit/f399b643b4e26e00035163ac98d2413e670df0fc#diff-0a446470c46e57216a7fcb4cbc96f96f5a17c0de214ced33df2f771bee553d89R472-R481
- `punctuation.openingQMark` and `punctuation.closingQMark` are changed to typographical quotes.

An example highlight template to play with:
```
{{if color == '#ff6666'}}
	<h2>{{text}}</h2>
{{elseif color == '#2ea8e5'}}
	{{if comment}}<p>{{comment}}:</p>{{endif}}<blockquote>{{text}}</blockquote><p>{{citation}}</p>
{{else}}
	<p>{{quotestart}}{{text}}{{quotestop}} {{citation}} {{comment}} {{if tags}} #{{tags join=' #'}}{{endif}}</p>
{{endif}}
```

### DIFF
--- a/chrome/content/zotero/xpcom/editorInstance.js
+++ b/chrome/content/zotero/xpcom/editorInstance.js
@@ -412,8 +412,7 @@ class EditorInstance {
 
 			let vars = {
 				color: annotation.color,
-				highlight: highlightHTML,
-				quotedHighlight: quotedHighlightHTML,
+				highlight: (attrs) => attrs.quotes === 'true' ? quotedHighlightHTML : highlightHTML,
 				comment: commentHTML,
 				citation: citationHTML,
 				image: imageHTML,
--- a/defaults/preferences/zotero.js
+++ b/defaults/preferences/zotero.js
@@ -196,5 +196,5 @@ pref("extensions.zotero.retractions.recentItems", "[]");
 
 // Annotations
 pref("extensions.zotero.annotations.noteTemplates.title", "<h1>{{title}}<br/>({{date}})</h1>");
-pref("extensions.zotero.annotations.noteTemplates.highlight", "<p>{{quotedHighlight}} {{citation}} {{comment}}</p>");
+pref("extensions.zotero.annotations.noteTemplates.highlight", "<p>{{highlight quotes='true'}} {{citation}} {{comment}}</p>");
 pref("extensions.zotero.annotations.noteTemplates.note", "<p>{{citation}} {{comment}}</p>");


dstillman:
> I'm thinking if {{text}} should be {{highlight}}

Yeah, `{{highlight}}` seems clearer anyway.

> Shouldn't it be allowed to enable/disable colors over the template? Or a separate pref will be used for that?

I don't think that belongs in the template, since it's really more of a display preference than a content preference. I think a "Show annotation colors by default" preference in the same prefs section where these templates will go would make sense.

> For the title template it sets a variable named string.pdfReader.annotation, does it makes sense to have the string. prefix?

The idea being that `strings.` indicates that `pdfReader.annotations` is a localized string? I think that's an implementation detail and we shouldn't use any dotted or even camel-cased variable names. I think it could just be…`{title}`? As in, this template is for the title section, and the actual title ("Annotations") is the primary thing that goes in that section. (We'll also have a `highlight` within the "highlight" template.)

> It seems that noteEditor.annotationsDateLine is no longer necessary

Yeah, can remove that from `en-US`. (It'll be removed automatically from the others.)

dstillman:
Oh, and in French I don't get the popup on the highlight, presumably for the same reason.

dstillman:
Maybe `quotestart`/`quotestop` is dumb, particularly since the translations can include spaces. Could do `quotedhighlight` and `highlight`? And default to `<p>{{quotedhighlight}} {{citation}} {{comment}}</p>`?

mrtcode:
> Maybe `quotestart`/`quotestop` is dumb, particularly since the `QMark` translations can include spaces as necessary. Could do `quotedhighlight` and `highlight`? And default to `<p>{{quotedhighlight}} {{citation}} {{comment}}</p>`?

Yes, that is better than wrapping in quotes. Alternatively we could do `{{highlight quotes='true'}}` or something similar.

dstillman:
> `{{highlight quotes='true'}}`

Oh, I like that.

PR #2365, opened 2022-02-21
Allow regular parent items in zotero://open-pdf/

sizmailov:
This PR fixes `zotero://open-pdf/...` handling.

The the patch adds an extra loop over item attachments and proceeds with first found pdf.

The bug was experienced on both Windows and Linux in `Zotero 6.0-beta.1`.

### DIFF
--- a/components/zotero-protocol-handler.js
+++ b/components/zotero-protocol-handler.js
@@ -1160,47 +1160,55 @@ function ZoteroProtocolHandler() {
 				Zotero.warn(`No item found for ${uriPath}`);
 				return;
 			}
-			
-			var item = results[0];
-			
-			if (!item.isFileAttachment()) {
-				Zotero.warn(`Item for ${uriPath} is not a file attachment`);
+
+			let pdf_attachment;
+
+			// Find a pdf attachment within search results
+			for (let i=0; i < results.length; i++){
+				let result = results[i]
+				let pdfs = (
+					result.getAttachments()
+						.map(itemID => Zotero.Items.get(itemID))
+						.filter(x => x.isFileAttachment() && x.attachmentContentType == 'application/pdf')
+				)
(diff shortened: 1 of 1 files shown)


dstillman:
Thanks! If you're saying that `zotero://open-pdf` doesn't work on parent items, that's not a bug, though — it's not supposed to. It works directly on attachment items.

We could consider supporting it on parent items, though then we'd use `item.getBestAttachment()`, not iterate over the child items manually.

Can you explain why you're looking to have this?

sizmailov:
Hi! Thanks for quick reply!

I want to reference zotero pdfs from external document.

I found a zotero forum [topic](https://forums.zotero.org/discussion/73776/create-external-link-to-open-pdf-within-zotero) with partially obsolete suggestions. Probably I run into a misconception of `open-pdf` URI interpretation. My understating inferred from forum topic and comment in sources was that it should work on "parent" items.

https://github.com/zotero/zotero/blob/b71367760d6b5773e7dff08793d022c74c5eb610/components/zotero-protocol-handler.js#L1108

Is `zotero://` URI scheme documented somewhere or `zotero-protocol-handler.js` is the only source of truth?

By the way, what is canonical way to obtain zotero-URI references to item/attachment? Currently I have to use web-version to figure out itemID (as suggested on forum pages), arguably it's far from being convenient. 

>  We could consider supporting it on parent items, though then we'd use item.getBestAttachment(), not iterate over the child items manually.

Totally agreed. Could you give me directions where to put this function if this PR is not considered off the plate?

dstillman:
https://forums.zotero.org/discussion/comment/411710/#Comment_411710

sizmailov:
Thanks for the link. As far as I understand, the support for opening pdf via parent `itemID` should be added somehow orthogonal to this PR, therefore closing.

PR #2366, opened 2022-02-21
Add Short Title item tree column

### LINKED ISSUE #1729: Add Short Title column
https://forums.zotero.org/discussion/78905/display-short-titles-in-my-library

After items tree HTMLification

AbeJellinek:
Closes #1729.

### DIFF
--- a/chrome/content/zotero/itemTreeColumns.jsx
+++ b/chrome/content/zotero/itemTreeColumns.jsx
@@ -54,7 +54,7 @@ const COLUMNS = [
 		dataKey: "title",
 		primary: true,
 		defaultIn: new Set(["default", "feed"]),
-		label: "zotero.items.title_column",
+		label: "itemFields.title",
 		ignoreInColumnPicker: true,
 		flex: 4,
 		zoteroPersist: new Set(["width", "hidden", "sortDirection"])
@@ -68,7 +68,7 @@ const COLUMNS = [
 	},
 	{
 		dataKey: "itemType",
-		label: "zotero.items.type_column",
+		label: "zotero.items.itemType",
 		width: "40",
 		zoteroPersist: new Set(["width", "hidden", "sortDirection"])
 	},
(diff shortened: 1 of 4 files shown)


chrome/content/zotero/itemTreeColumns.jsx
```diff
@@ -267,6 +267,14 @@ const COLUMNS = [
 		flex: 1,
 		zoteroPersist: new Set(["width", "hidden", "sortDirection"])
 	},
+	{
+		dataKey: "shortTitle",
+		disabledIn: "feed",
+		submenu: true,
+		label: "zotero.items.shortTitle_column",
```
dstillman, line 274: @adomasven: Is there any good reason we can't use zotero.properties strings for these? Right now it [uses strings from `Zotero.Intl.strings`](https://github.com/zotero/zotero/blob/f333585d988f9c2d3b7241386d730466faf0ac36/chrome/content/zotero/components/virtualized-table.jsx#L1072-L1074), but I assume that's a vestige of the XUL tree needing strings to be in zotero.dtd. If we're getting them via JS anyway, we should just get them with `getString()` so that we can reuse zotero.properties strings. I'd rather not add another duplicate string to zotero.dtd that everyone will have to translate.

The one catch is that we're not explicitly differentiating between raw strings and string ids, and while `getString()` returns `false` if the string doesn't exist, it also logs an error. But we could just try `getString()` as long as the string looks like a key: e.g., `^[^\s]+\w\.\w[^\s]+$`

adomasven: I didn't realize we were duplicating these in both of the files. Yeah, we can switch to `getString()`.

AbeJellinek: OK, done. I kept around `creator_column` and `year_column` because they don't have equivalents in `zotero.properties` and `dateModified_column` because it's localized as "Date Modified" instead of "Modified" as in the item pane. Is that right? And should those be moved to `zotero.properties`, or are they fine staying in the DTD? (I assume that's better so we don't have to retranslate them on Transifex.)

PR #2367, opened 2022-02-22
Support opening PDFs with Zotero to import

### LINKED ISSUE #1169: Opening a PDF in Zotero should add it as a standalone attachment
Same as drag-and-drop. (And eventually, it should run Retrieve Metadata automatically.)

https://forums.zotero.org/discussion/64252/5-0-beta-open-with-zotero-does-not-add-pdf-to-library

AbeJellinek:
Fixes #1169, needs zotero/zotero-standalone-build#90. Automatically attempts to recognize imported PDFs.

### DIFF
--- a/chrome/content/zotero/selectImportTargetDialog.js
+++ b/chrome/content/zotero/selectImportTargetDialog.js
@@ -23,34 +23,110 @@
     ***** END LICENSE BLOCK *****
 */
 
+Components.utils.import("resource://gre/modules/Services.jsm");
+
 let io = window.arguments[0];
 if (io.wrappedJSObject) io = io.wrappedJSObject;
 
+let menulist;
+let selectedTarget;
+
 function load() {
-	let { filename } = io.dataIn;
+	function selectTarget(libraryOrCollection) {
+		menulist.setAttribute('label', libraryOrCollection.name);
+		menulist.setAttribute('value', libraryOrCollection.treeViewID);
+		selectedTarget = libraryOrCollection;
+		// Menu looks too cramped with an icon
+		menulist.removeAttribute('image');
+	}
+
+	async function createNewCollection(libraryID) {
+		let name = Zotero.getMainWindow().Zotero_File_Interface.nameCollectionForImportedFile(
+			file,
+			libraryID,
+			Zotero.getString("fileInterface.imported")
+		);
+		let newName = { value: name };
+		let result = Services.prompt.prompt(window,
+			Zotero.getString('pane.collections.newCollection'),
+			Zotero.getString('pane.collections.name'), newName, '', {});
+		if (result) {
+			let collection = new Zotero.Collection();
+			collection.libraryID = libraryID;
+			collection.name = newName.value || name;
+			await collection.saveTx();
+			selectTarget(collection);
+		}
+	}
+
+	let { file } = io.dataIn;
 
 	document.title = Zotero.getString('ingester.importFile.title');
-
 	document.getElementById('select-target-text').value
-		= Zotero.getString('ingester.importFile.text', filename);
+		= Zotero.getString('ingester.importFile.text', file.leafName);
 
-	let menulist = document.getElementById('select-target-menu');
-	let libraries = Zotero.Libraries.getAll().filter(lib => lib.editable);
-	Zotero.Utilities.Internal.buildLibraryMenu(
-		menulist,
-		libraries,
-		Zotero.getActiveZoteroPane().getSelectedLibraryID()
-	);
+	let isEditable = row => row.editable
+        && (!file.leafName.endsWith('.pdf') || row.filesEditable);
+	let libraries = Zotero.Libraries.getAll().filter(isEditable);
 
-	let checkbox = document.getElementById('select-target-checkbox');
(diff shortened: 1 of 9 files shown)


components/zotero-service.js
```diff
@@ -612,42 +612,49 @@ ZoteroCommandLineHandler.prototype = {
...
+					
+					if (file.leafName.substr(-4).toLowerCase() === ".csl"
+							|| file.leafName.substr(-8).toLowerCase() === ".csl.txt") {
+						// Install CSL file
+						Zotero.Styles.install({ file: file.path }, file.path);
+					}
+					else if (file.leafName.substr(-4).toLowerCase() === ".pdf") {
+						let item = await Zotero.Attachments.importFromFile({
+							file,
+							libraryID: Zotero.Libraries.userLibraryID
```
dstillman, line 628: Should we have a dialog with a library drop-down, defaulting to the currently selected library? It's sort of weird if you're in another library and open a PDF via Zotero and then the PDF goes into your personal library.

dstillman: Arguably we should use that for metadata files too. If you only have a personal library, it would skip the prompt.

AbeJellinek: Added a unified dialog. It looks like this:

image

chrome/content/zotero/selectImportTargetDialog.js
```diff
@@ -0,0 +1,56 @@
...
+function load() {
+	let { filename } = io.dataIn;
+
+	document.title = Zotero.getString('ingester.importFile.title');
+
+	document.getElementById('select-target-text').value
+		= Zotero.getString('ingester.importFile.text', filename);
+
+	let menulist = document.getElementById('select-target-menu');
+	let libraries = Zotero.Libraries.getAll().filter(lib => lib.editable);
```
dstillman, line 38: Also has to be `filesEditable` for PDFs

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -634,9 +634,9 @@ ingester.scrapeErrorDescription	= An error occurred while saving this item. Chec
 ingester.scrapeErrorDescription.linkText = Troubleshooting Translator Issues
 ingester.scrapeErrorDescription.previousError = The saving process failed due to a previous Zotero error.
 
-ingester.importFile.title	= Import File
-ingester.importFile.text	= Do you want to import the file "%S"?
-ingester.importFile.intoNewCollection = Import into new collection
+ingester.importFile.title				= Import File
+ingester.importFile.text				= Import “%S” to:
```
dstillman, line 638: `to` → `into`

components/zotero-service.js
```diff
@@ -612,42 +612,82 @@ ZoteroCommandLineHandler.prototype = {
...
-								   .getService(Components.interfaces.nsIWindowMediator);
-								var browserWindow = wm.getMostRecentWindow("navigator:browser");
-								browserWindow.Zotero_File_Interface.importFile({
+					await Zotero.uiReadyPromise;
+					var file = Zotero.File.pathToFile(param);
+
+					// Perform import in front window
+					var wm = Components.classes["@mozilla.org/appshell/window-mediator;1"]
+						.getService(Components.interfaces.nsIWindowMediator);
+					var browserWindow = wm.getMostRecentWindow("navigator:browser");
```
dstillman, line 647: → `Services.wm.getMostRecentWindow`

dstillman (changes requested):
I can't remember exactly how importing worked previously — if you didn't check `Import into new collection`, did it import into the currently selected collection or the root?

Even if the former, it's weirder with the addition of the library drop-down, because now you get different behavior based on whether you pick the current library (in which case it saves to the collection you happen to have selected behind the dialog) or another library (in which case it saves to the library root). There's no guarantee that people will have even been using Zotero recently, so I think it would be easy to accidentally import into a random collection just because you wanted something in the current library.

Maybe a better approach would be to use the full target drop-down, with the current collection selected by default. Bonus: if the current collection wasn't editable, you could then get immediate feedback on the fallback location (i.e., library root or personal library).

AbeJellinek:
The former, I think. I agree that it was/is weird. A target dropdown seems like a good approach. Should it have a “New Collection” option under each library that creates an automatically named new collection for the items, or should “New Collection…” open a dialog that prompts for a name (maybe with the automatic name as the default)?

dstillman:
> should “New Collection…” open a dialog that prompts for a name (maybe with the automatic name as the default)

I think this? The default name is pretty unlikely to be what you want, and it's just an extra Return if you don't care.

dstillman:
Oh, except this is for importing BibTeX/RIS/etc. too, in which case the auto-generated name might be fine. But still, it's just a Return. So I think show the dialog.

dstillman:
We should select the selected collection after import, and if it's a single item select the item too. Right now I believe it's just selecting the item in the current collection or library root if you select a collection other than the current one.

AbeJellinek:
OK, done.

dstillman:
https://forums.zotero.org/discussion/comment/504286/#Comment_504286

This doesn't have the option to add the attachment as a child attachment to the selected item, does it? Should it? Not sure if that would be useful or just annoying.

PR #2374, opened 2022-03-01
Automatically relink attachments from LABD, fix invalid path error

### LINKED ISSUE #2084: Automatically relocate linked files within current base directory
People frequently misuse the <a href="https://www.zotero.org/support/preferences/advanced#linked_attachment_base_directory">Linked Attachment Base Directory</a> setting and end up on a new computer with lots of missing files with absolute paths. If they can temporarily create the old folder scheme, they can fix it with the LABD setting, but otherwise they have to use Zutilo to batch-change files, which is confusing, error-prone, and I think buggy (there's something with lost first letters of filenames).

If a base directory is set and a user tries to open a linked file with an absolute path that doesn't exist, we could try to find the file within the base directory. If found, we could offer to relocate 1) just this file or 2) all files in the old location.

To test this, we'd start with just the filename and then try successive parent folders until we find a relative path or the filesystem root.

If we find a match, we could check to see how many other missing files we could fix and provide the info in the confirmation dialog.

### LINKED ISSUE #2092: File-not-found dialog and context menu not appearing on Windows for linked file with absolute macOS path
https://forums.zotero.org/discussion/comment/385310/#Comment_385310

AbeJellinek:
Closes #2084, fixes #2092.


chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -405,6 +405,13 @@ pane.item.attachments.fileNotFound.text2.stored = It may have been moved or dele
 pane.item.attachments.fileNotFound.text2.stored.notOnServer = It may have been moved or deleted outside of %1$S, or, if the file was added on another computer, it may not yet have been synced to %2$S.
 pane.item.attachments.fileNotFound.text2.linked = It may have been moved or deleted outside of %1$S, or a Linked Attachment Base Directory may be set incorrectly on one of your computers.
 pane.item.attachments.delete.confirm  = Are you sure you want to delete this attachment?
+pane.item.attachments.autoRelink.title = Located Automatically
+pane.item.attachments.autoRelink.text1 = A matching file was found in your Linked Attachment Base Directory at the following path:
```
dstillman, line 409: Probably should show the absolute path too? As written, we're not really explaining why we're showing this dialog. It doesn't particularly matter if you just want to fix it, but it'd be clearer if we showed 1) the current absolute path and 2) the location within the LABD where it was found. Maybe `The file could not be found at the specified location, but a matching file was found in your Linked Attachment Base Directory:` and add lines for "Old Location:" and "New Location:"?

chrome/content/zotero/zoteroPane.js
```diff
@@ -4440,6 +4440,76 @@ var ZoteroPane = new function()
...
+			let buttonFlags = ps.BUTTON_POS_0 * ps.BUTTON_TITLE_IS_STRING
+				+ ps.BUTTON_POS_1 * ps.BUTTON_TITLE_IS_STRING
+				+ ps.BUTTON_POS_2 * ps.BUTTON_TITLE_IS_STRING;
+			let index = ps.confirmEx(null,
+				title,
+				text,
+				buttonFlags,
+				Zotero.getString('pane.item.attachments.autoRelink.relinkAll'),
+				Zotero.getString('pane.item.attachments.autoRelink.relinkOne'),
+				Zotero.getString('pane.item.attachments.autoRelink.locateManually'),
```
dstillman, line 4479: The way the Mozilla confirmation dialog works, index 1 is what happens if you press Esc in the dialog, so that has to either be an explicit Cancel or the least-destructive option. A Cancel is usually nice to offer anyway. Can't have more than three buttons, so probably best to make `locateManually` index 1, and then you can always cancel out of the filepicker?

dstillman:
I assume this will fix https://forums.zotero.org/discussion/96979/cannot-open-associated-pdf-files-nor-have-right-click-menu (relative paths containing `/` on Windows, I think?) as well, but we should confirm that.

dstillman:
It's actually just an absolute path that somehow ended up with mixed slashes.

chrome/locale/en-US/zotero/zotero.properties
AbeJellinek: > `The file could not be found at the specified location, but a matching file was found in your Linked Attachment Base Directory:`

Maybe with an "although"? The dialog that shows before this one already tells the user that `The attached file could not be found at the following path:` - this dialog only shows after they click `Locate...`. If this dialog _again_ says that the file couldn't be found, with slightly different wording, it sort of seems like the "Locate" operation failed, or something like that, when actually it hasn't happened yet.

dstillman: Oh, I didn't realize that this appears after `Locate…`. (I haven't actually tried running this yet.)

What about just running this check automatically when a linked file can't be found and the absolute path isn't within the current LABD?

AbeJellinek:
>I assume this will fix [forums.zotero.org/discussion/96979/cannot-open-associated-pdf-files-nor-have-right-click-menu](https://forums.zotero.org/discussion/96979/cannot-open-associated-pdf-files-nor-have-right-click-menu) (relative paths containing `/` on Windows, I think?) as well, but we should confirm that.

Kind of. If the path would be valid with normalized slashes (like, `C:\Users/abe\Desktop/test.pdf` if `C:\Users\abe\Desktop\test.pdf` exists), opening in Zotero works, Show File doesn't, system reader support probably depends on how that reader handles weird paths. Because it technically exists, the Locate dialog won't be shown. Should we automatically normalize slashes after some set of attachment actions? Should we show a warning? I'm not sure.

chrome/locale/en-US/zotero/zotero.properties
AbeJellinek: Sure, that makes sense.

AbeJellinek: The only thing that worries me slightly is that finding other relinkable files in the same directory could be very slow on something like a network share. I thought giving the user a chance to back out after opening a missing file but before locating it made sense in case they decide they don't want to deal with it at the moment.

We could also fix that by breaking after one additional relinkable file is found and only performing the full search if they actually choose "Relink All." We wouldn't be able to show an exact count of other relinkable files anymore.

What do you think?

chrome/content/zotero/xpcom/data/items.js
```diff
@@ -1712,6 +1712,29 @@ Zotero.Items = function() {
...
 		return title;
 	};
+
+
+	/**
+	 * Find attachment items whose paths point to missing files and begin with
+	 * the passed `pathPrefix`.
+	 *
+	 * @param {Number} libraryID
+	 * @param {String} [pathPrefix='']
```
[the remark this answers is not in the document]
AbeJellinek: No, probably not. Made required.

chrome/locale/en-US/zotero/zotero.properties
dstillman: 1) We'd only be looking for other files if the current file matched.

2) We'd only check files using the same relative path segment as was required for the matched file, so whereas the first file might require multiple `stat`s, there'd only be one per file for the additional files. (I didn't look closely to see that this is what the code currently does, but I assume we can do this?)

Given these two things, I think it's OK to run automatically.

If we're concerned, we could have a time limit for checking other files and have an alternative version of the dialog that appears if we can't generate a count, but I doubt that's worth it or would even ever be shown

AbeJellinek: Yes to both. In that case, sounds good.

chrome/content/zotero/zoteroPane.js
```diff
@@ -4794,6 +4869,102 @@ var ZoteroPane = new function()
...
+	 * by searching successive subdirectories. Prompt the user if a match is
+	 * found and offer to relink one or all matching files in the directory.
+	 * The user can also choose to relink manually, which opens a file picker.
+	 *
+	 * If the path is 'C:\Users\user\Documents\Dissertation\Files\Paper.pdf'
+	 * and the LABD is '/Users/user/Documents', check:
+	 *
+	 * 1. /Users/user/Documents/Paper.pdf
+	 * 2. /Users/user/Documents/Dissertation/Paper.pdf
+	 * 3. /Users/user/Documents/Dissertation/Files/Paper.pdf
```
dstillman, line 4885: I specified it this way, but I wonder if we have this backward. If the LABD is ~/Documents, shouldn't it try starting with the full path?

1. /Users/user/Documents/Users/user/Documents/Dissertation/Files/Paper.pdf
2. /Users/user/Documents/user/Documents/Dissertation/Files/Paper.pdf
3. /Users/user/Documents/Documents/Dissertation/Files/Paper.pdf
4. **/Users/user/Documents/Dissertation/Files/Paper.pdf**

Or, even better, we could split normalized paths into components and remove matching components, in which case we'd get the above example in the first try:

1. Path is `['Users', 'user', 'Documents', 'Dissertation', 'Files', 'Paper.pdf']`
2. Remove LABD of `['Users', 'user', 'Documents']`
3. End up with `['Dissertation', 'Files', 'Paper.pdf']`, which is the correct relative path

Otherwise we're risking matching the wrong file if there's another Paper.pdf in ~/Documents. (Obviously linked files will generally be named more clearly, but still.) We'll need to make sure we're catching errors from paths that are too long to try (which we're probably doing already, but I can't remember).

dstillman:
https://forums.zotero.org/discussion/100232/linked-attachment-based-directory-for-google-drive

We should probably finish this.

chrome/content/zotero/zoteroPane.js
dstillman: Yeah, looking this over last night I realized that it was a contrived example and splitting would almost never help. So just the starting with the full path makes sense.

dstillman:
Maybe we should say "a file with the same name" instead of "a matching file"? I'm a little concerned we're giving the impression that we're checking the contents.

Similarly, I wonder if it's worth showing any examples of the other unlinked files in the second dialog, even just some filenames without paths. Would you feel comfortable clicking "Yes"? Maybe if it's all just one big folder and you know it's just at a new location, which is probably what it will be the vast majority of the time here, but otherwise it could be a little unsettling.

For the one-file case in the second dialog, I don't think it makes sense to say "Relink both located attachments" when you already clicked Relink in the first dialog. Maybe "Relink this attachment as well?"

And I think the buttons can be "Cancel" and "Relink"/"Relink All" in the second dialog — a little weird to have Cancel in the second dialog after you've relinked the first, but better to have "Relink"/"Relink All" than "Yes".

dstillman:
Ah, you reworded. I'd go with "Relink this attachment as well?" over "Relink the additional located attachment" — a bit less repetitive. I'm fine with "Relink all located attachments" for the multi one.

dstillman:
> Should I cherry-pick to fx102?

Yep!

PR #2375, opened 2022-03-02
HTML Tree: Add a dotted border to pivot rows

### LINKED ISSUE #2373: HTML tree: Display dotted border when selecting non-contiguous rows via keyboard
You can use Cmd-up/down and then space bar to select multiple rows via the keyboard, but unlike in Firefox and Thunderbird, it doesn't display a dotted border around the active row when moving. (I didn't even know we had this and was actually about to add it as an enhancement request.)

adomasven:
Closes #2373.

As a result of this after changing the collection the first row always has the pivot border. I don't think it's a big problem and that's how it always was on Window or Linux, but you may think it to be unusual or distracting, so your call on how/whether we address that.

### DIFF
--- a/chrome/content/zotero/components/virtualized-table.jsx
+++ b/chrome/content/zotero/components/virtualized-table.jsx
@@ -92,11 +92,13 @@ class TreeSelection {
 
 		if (this.selectEventsSuppressed) return;
 
-		if (this._tree.invalidate) {
-			this._tree.invalidateRow(index);
-		}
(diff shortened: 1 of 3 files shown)


dstillman:
Yes, I think it is unusual and distracting.

How about just not showing the pivot border when the tree doesn't have focus? That will also prevent it from showing unnecessarily around a selected-but-gray row when the tree doesn't have focus.

Issue #2289, opened 2021-12-27
PDF reader: Quick Copy drag-and-drop of annotations


dstillman:
https://forums.zotero.org/discussion/comment/396234/#Comment_396234

Any reason we couldn't do this? Seems like right now we copy as the underlying HTML (though with line breaks instead of paragraphs), so we'd have to do the same conversion we do when converting to notes and then do the regular note export?

dstillman:
(We'll need to move the templating out of editorInstance.js.)

mrtcode:
A few questions:
1) I guess we want to support not only drag & drop, but also copy & paste?
2) Do we want to support all note export translators (Markdown & Rich Text, HTML)?
3) I guess there is no better way to do that than creating a temporary note item and passing to necessary translators?

dstillman:
1. Meaning the Quick Copy shortcut key? Yes.
2. Yes.
3. Yes, that's probably easiest, as long as we can do it in-memory (which I think we can — if not, we can fix that).

mrtcode:
> 1. Meaning the Quick Copy shortcut key? Yes.

We already have "Copy Citation" and "Copy Bibliography" (and their keyboard shortcuts) in Edit menu, when a PDF reader tab is open. Should there be a "Copy Note" option as well (it has the same keyboard shortcut as "Copy Citation").

Could you explain how that Quick Copy shortcut would work for selected annotations? Isn't Cmd-C enough in this case?

dstillman:
Yeah, I guess since we're using the Quick Copy shortcuts for the parent item, Cmd-C when an annotation is selected is fine here. And maybe we change the Copy option in Edit to Copy Note when a note is selected, since it's the same shortcut?

PR #2377, opened 2022-03-03
Allow to Quick Copy annotations

mrtcode:
This can be tested now. Works with dragging and copying.

Although it doesn't change "Copy" to "Copy Note" in Edit menu because:
1) It's not only about notes. Copied annotations can be pasted into note-editor as active annotations.
2) Unclear when exactly "Copy" should be changed to "Copy Note". I guess it would need to track when focus switches between pdf-reader and note-editor. It might not be easy to do that.

#2289.


dstillman:
Awesome.

Issue #2379, opened 2022-03-04
PDF reader: Keyboard access to right side of toolbar and right-hand pane


dstillman:
https://forums.zotero.org/discussion/94838/beta-new-keyboard-navigation

We know about this, but just to have a ticket for it.

mrtcode:
Do we want to be able to tab to the right pane, or there would a special keyboard shortcut for switching? And then how to go back to pdf-reader? Because, for example, in info pane you need to press tab many times to navigate through it.

dstillman:
Maybe there could be a shortcut too, but for now, just Tab

And ideally you'd Shift-tab back out of the item pane, as in the library view, and also be able to tab out of it back to the beginning

mrtcode:
So far I was unable to implement this, but continuing to work further.

I guess that wouldn't be enough if the order of navigation would be like this: Info/Note split button -> PDF reader -> Info/Note pane? `tabindex` doesn't work if iframes are involved, I might need to manually code that, if we want a different order.

Also, after which element it should jump out from PDF reader?

Navigating into info pane was never possible right? Navigating in other sections is also unpredictable, and I basically can't reach anything (in pre-beta Zotero as well).

I guess after we reactify everything, we'll implement a proper navigation from zero. But for now let's just define the minimal goals what we want to do here. Maybe just focusing to note editor is enough for now?

PR #2395, opened 2022-03-09
Improve keyboard navigation in PDF reader tab

mrtcode:
Let's try this out.

- Toolbar navigation is now continuous.
- Improved notes list navigation.
- Toggling Item/Notes split button focuses context pane, even when doing so with mouse (which is not necessary good).

#2379

### DIFF
--- a/chrome/content/zotero/bindings/noteeditor.xml
+++ b/chrome/content/zotero/bindings/noteeditor.xml
@@ -349,6 +349,25 @@
 				</body>
 			</method>
 
+			<method name="focusFirst">
+				<body>
+					<![CDATA[
+						(async () => {
+							let n = 0;
+							while (!this._editorInstance && n++ < 100) {
+								await Zotero.Promise.delay(10);
+							}
+							await this._editorInstance._initPromise;
+							this._iframe.focus();
+							try {
+								this._editorInstance._iframeWindow.document.querySelector('.toolbar-button-return').focus();
+							} catch(e) {
+							}
+						})();
+					]]>
+				</body>
+			</method>
+
 			<method name="_id">
 				<parameter name="id"/>
 				<body>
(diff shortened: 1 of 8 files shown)


mrtcode:
We can try this out again.

Although I still need to fix RTL mode. Also there is a strange border (which was all the time?) below Item/Notes split button which is visible when context pane is closed. I'll fix it.

dstillman:
Not sure if this is unique to this PR but pressing Shift-Tab from Info is scrolling the PDF for me, seemingly to a random page. Let me know if you can't reproduce.

dstillman:
This is great, except two (related) things:

1) Shift-Tab from the toolbar focuses the PDF view, not the right-hand pane (e.g., the last row in the item pane).
2) In the Info pane, tabbing beyond Extra focuses the PDF view rather than the toolbar (so if you hold down Tab you keep cycling back through the item pane).

dstillman:
(The Shift-Tab thing would be good to fix but the other part isn't a blocker — we could open a separate ticket for that and merge this now.)

mrtcode:
> Not sure if this is unique to this PR but pressing Shift-Tab from Info is scrolling the PDF for me, seemingly to a random page. Let me know if you can't reproduce.

I can't reproduce that. Maybe a video demonstrating this would help me to guess what could cause this.

dstillman:
https://user-images.githubusercontent.com/783082/157833887-01e9f495-4f68-4c67-8900-8fca5240d85a.mov

dstillman:
PDF is http://arxiv.org/abs/2203.04293

mrtcode:
Oh, I guess it's focusing link annotations.

mrtcode:
How about tabbing into note-editor? It currently focuses editor text, but when doing Shift-Tab you have to walk through all the toolbar buttons. Do we want to focus PDF view → return button instead?

dstillman:
I think we have to? Having different tab orders in different directions seems pretty confusing in terms of accessibility.

mrtcode:
Deficiencies that still exists:
- Shift-Tab on reader toolbar focuses context pane first focusable node, not the last one (for that we need to analyze content in each pane to determine which focusable node is the last one).
- Unnecessary border.
- RTL mode issues.
- Tab in note-editor doesn't jump anywhere.

PR #2406, opened 2022-03-11
Allow marking as different and batch merging in Duplicate Items

### LINKED ISSUE #898: Enable batch merging of duplicates
As raised in the [General: Duplicates](https://forums.zotero.org/discussion/54285/duplicates/) forum thread:

> is there any possibility for killing duplicates without accepting item by item? I have more than 1000 duplicates and want to choose every time the latest entry as master. So an option to choose a general rule how to handle duplicates would be great.

### LINKED ISSUE #1122: Allow marking items as not duplicates with relations

AbeJellinek:
I'm not dead-set on "Hide in Duplicate Items" for marking items as different, but I like it because it's self-explanatory, even though it isn't really a great description of what's happening behind the scenes. I also thought about just "Mark as Different," but it's not especially self-explanatory. "Ignore Duplicates" doesn't feel quite right either, since I feel like it implies ignoring duplicates in _general_.

If you have two items that are detected as duplicates, mark them as non-duplicates, then add a third that also gets detected as a duplicate in the same set, the first two will reappear along with the third in Duplicate Items. I don't know if that's the best behavior, but it's kind of inherent to the relation-based approach - the first two are correctly detected as duplicates of the third, and there's no relation saying they aren't.

Fixes #1122, fixes #898.

### DIFF
--- a/chrome/content/zotero/xpcom/data/items.js
+++ b/chrome/content/zotero/xpcom/data/items.js
@@ -960,8 +960,7 @@ Zotero.Items = function() {
 	this.merge = function (item, otherItems) {
 		Zotero.debug("Merging items");
 
-		return Zotero.DB.executeTransaction(async function () {
-			var replPred = Zotero.Relations.replacedItemPredicate;
+		return Zotero.DB.executeTransaction(async () => {
 			var toSave = {};
 			toSave[item.id] = item;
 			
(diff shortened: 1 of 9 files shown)


dstillman:
Yeah, sure. We need something to release during the Z7 beta.

dstillman:
Yes, I mean release on Z6

Issue #2411, opened 2022-03-12
Improved RTL support


dstillman:
https://forums.zotero.org/discussion/95013/poor-bi-directional-rtl-support-in-zoteros-user-interface-ui

We can create separate tickets for individual issues, but creating this as a tracking ticket.

dstillman:
(Also not sure whether this person is using Z 5 or 6. Is 6 better or worse?)

david-von-tamar:
> (Also not sure whether this person is using Z 5 or 6. Is 6 better or worse?)

Hello, I'm using Zotero 6.0.8 (Beta), and my computer runs on Debian Sid. I came to open up an issue but then I searched and found that you opened this issue for me, so thank you.

I want to add that RTL support in Zotero's UI is borderline usable. When the RTL text in the reference list is longer than the column, you can't read it unless you edit the text and copy-paste it elsewhere. In addition, punctuation marks make it difficult to read and edit RTL text, because the cursor jumps left and right and final punctuation marks (like closing quotes) appear misplaced on the opposite side of the text.

AbeJellinek:
@david-von-tamar: Both of those issues are fixed in #2415.

david-von-tamar:
@AbeJellinek, which version of Zotero introduces these fixes? Debian's `apt` indicates that I'm using version 6.0.8 beta.

AbeJellinek:
None - that PR isn't merged yet.

PR #2415, opened 2022-03-15
Improve bidi & RTL support

### LINKED ISSUE #63: Buttons are squished or truncated in RTL localized Standalone
Tested ar and fa locales.

### LINKED ISSUE #2380: Toolbar menu buttons rendered incorrectly in RTL locale
SCR-20220303-te2

The dropdown icon next to the attachment column in the item tree is also misplaced.

AbeJellinek:
Fixed issues:

- [x] Important and visually broken UI elements
	- [x] Main toolbar
- [x] Bidirectional text display
	- [x] Item tree
	- [x] Collection tree
	- [x] Other trees
	- [x] Item box
		- [x] The main stuff
		- [x] Creator fields jumping around in size after display
		- [x] Bidi field support
- [x] Context menus appearing in the wrong location
	- [x] Item tree
	- [x] Collection tree
	- [x] Tree columns
- [x] Interactive features
	- [x] Tree column resizing
- [x] Spacing, padding, and alignment
	- [x] Item box
	- [x] Tabs
		- [x] Close button still appears on the right in RTL

For a separate PR:
- Parameterized locales for date/number differences
	- Arabic-speaking countries west of Egypt use Western Arabic numerals (012345), east of Egypt Eastern Arabic numerals (٠١٢٣٤٥), but our Arabic locale only supports the latter
	- Similarly, there's a division in the [names for Gregorian months](https://en.wikipedia.org/wiki/Arabic_names_of_Gregorian_months)
	- I have a draft ready for this but I don't think it's essential for this PR.

Addresses #2411, fixes #2380, fixes #63


AbeJellinek:
@dstillman: I think this is ready. See the bottom of the PR description for some issues that I'm going to push to another PR; parameterized locales (not sure what the best word for them is; see [Unicode docs](https://unicode.org/reports/tr35/#u_Extension)) would be good for Arabic number formats and other issues that come with having a single locale covering a large region of the world, but they're a little outside the scope of this PR.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -37,6 +37,7 @@ general.notNow                     = Not Now
 general.passed						= Passed
 general.failed						= Failed
 general.and							= and
+general.andJoiner					= \u2068%S\u2069\u2066 and \u2069\u2068%S\u2069
```
AbeJellinek, line 40: Control characters make sure an RTL name won't put the entire string in RTL mode in a `dir="auto"` element.

dstillman: Can you add a comment with the characters being used here and what they do?

AbeJellinek:
Sure, but it will probably need #2641.

AbeJellinek:
@dstillman Conflicts all resolved and this is ready again. Let me know if anything else is needed before we can merge.

chrome/content/zotero/itemTree.jsx
```diff
@@ -1389,24 +1378,12 @@ var ItemTree = class ItemTree extends LibraryTree {
...
 			if (fieldB === undefined) {
 				let firstCreator = Zotero.Items.getSortTitle(sortStringB);
-				if (sortCreatorAsString) {
-					var fieldB = firstCreator;
-				}
-				else {
-					matches = andEtAlRegExp.exec(firstCreator);
-					fieldB = matches ? matches[0] : '';
-				}
+				fieldB = firstCreator;
```
dstillman, line 1386: Does this change the behavior? (I thought I knew what this did, but I can't actually trigger a difference here.)

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -35,6 +35,11 @@ general.notNow                     = Not Now
 general.passed						= Passed
 general.failed						= Failed
 general.and							= and
+# general.andJoiner: Escape codes isolate user-provided text to prevent its
+# directionality from controlling the directionality of the whole string.
+# \u2068 FIRST STRONG ISOLATE: Isolates the directionality of characters that follow
+# \u2069 POP DIRECTIONAL ISOLATE: Pops the above isolation
+general.andJoiner					= \u2068%S\u2069 and \u2068%S\u2069
```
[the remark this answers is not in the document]
AbeJellinek: Yeah, that’s a pretty good idea… I don’t know why I didn’t do it that way. Will fix.

chrome/content/zotero/itemTree.jsx
AbeJellinek: The code that was there tries to cut off the creator’s name at the first `and` or `et al.`, but that isn’t useful - the input is a single creator’s name from the database. So I’m not sure what the idea was originally, but we don’t need it.

dstillman: `firstCreator` here isn't the `firstCreator` string as generated by `_firstCreatorSQL()` — e.g., "Smith et al."?

AbeJellinek: If `sortCreatorAsString` is true, it's `item.firstCreator`, which is the output from `_firstCreatorSQL()`. But that was the code path that would lead to the string *not* being trimmed, I guess because `sortCreatorAsString` means that you want to sort by the entire string, `et al.`s and all.

https://github.com/zotero/zotero/blob/40ae2e32462e55fd1048888c267e7ecd76838c5d/chrome/content/zotero/itemTree.jsx#L1392-L1394

If `sortCreatorAsString` is false, it's `item.sortCreator`, which is just a space-separated list of creator names with no localized delimiters. And on that path, we we *were* trying to trim the `and`s and `et al.`s.

AbeJellinek: So maybe that was a bug and the trimming was supposed to happen if `sortCreatorAsString` is true? But then I wouldn't exactly see the point of the pref. My understanding is that it makes it sort alphabetically based on exactly what you see in the Creator item tree column.

chrome/locale/en-US/zotero/zotero.properties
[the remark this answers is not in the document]
AbeJellinek: OK, fixed.

PR #2432, opened 2022-03-20
Fix bad French translation of Filing Date

pagiraud:
"Filing Date" is a field that is only relevant for patents. In Zotero, the current translation in French is "Date d’archivage", which means something like "archiving date": this is not at all relevant. That’s why I suggest to change this for "Date de dépôt", which is the correct translation.


dstillman:
Thanks — fixed on [Transifex](https://www.transifex.com/zotero/zotero) (where all translation changes need to happen)

Issue #2433, opened 2022-03-20
Note editor: Add font color option


dstillman:
Lots of demand for this.

https://forums.zotero.org/discussion/88757/new-note-editor-change-the-font-color-in-the-notes and many others
https://forums.zotero.org/discussion/95378/note-editor-text-color-options-missing

mrtcode:
So just another toolbar button? What about colors? The current highlight colors doesn't make much sense here.

If we really want the old colors palette we could add a color swatch with three dots which would expand the colors popup.

dstillman:
Yeah, I think just a toolbar button.

I don't think a small set of predefined colors makes much sense here. Can we just display the XUL colorpicker in a popup for now? Recent colors would be nice but we can do without for the moment (and build our own HTML custom element colorpicker if there's not something available in newer Firefox).

mrtcode:
You mean `customcolorpicker.xul` with default colors? We can show it in a popup, but connecting that popup to note-editor won't be faster than making a popup directly in note-editor (with the same default colors as in `customcolorpicker.xul`).

dstillman:
OK, sure.

mrtcode:
Fixed in zotero/note-editor@5548e40b

It was unnecessary to create another issue, but, anyway, the discussion regarding further improvements to color picker is here #2464.

PR #2439, opened 2022-03-21
Add note-editor text color picker

mrtcode:
Maybe we should use TinyMCE colors instead? It now uses default colors from `customcolorpicker.xul`.

#2433


dstillman:
I think these colors are better, but there are too many. We should remove the two ugliest and/or most redundant colors from each row.

Probably also makes sense to flip vertically so black is in top left.

dstillman:
A bunch of those still seem too close. I think we can remove more. Also should be horizontal, not vertical — closer to the click target.

These colors are much more pleasant than the TinyMCE ones. We could add an extra swatch that shows up if the color under the cursor isn't one of the hard-coded ones, and then it'd be active so you could use it elsewhere at the beginning of a line. (Is that possible? Does it make sense?)

dstillman:
We can also refine this later, with, e.g., a row of recently used colors.

mrtcode:
Ok, how about this:
Screenshot 2022-03-22 at 14 59 57

dstillman:
👍

mrtcode:
This was merged in 23a4a14b448ea6520bc51c01c1112e1170d2b812.

PR #2448, opened 2022-03-22
Fix next duplicates set selection after merge

### LINKED ISSUE #2430: Next selected duplicates set not selected after merging
https://forums.zotero.org/discussion/95309/merge-function-not-automatically-going-to-next-set-of-duplicate-citations

HTML tree regression? Or more recent? I didn't test if this is fixed in https://github.com/zotero/zotero/pull/2406.

AbeJellinek:
Fixes #2430

### DIFF
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -759,7 +759,10 @@ var ItemTree = class ItemTree extends LibraryTree {
 			}
 			// On removal of a selected row, select item at previous position
 			else if (savedSelection.length) {
-				if ((action == 'remove' || action == 'trash' || action == 'delete')
+				if ((action == 'remove'
+						|| action == 'trash'
+						|| action == 'delete'
+						|| action == 'removeDuplicatesMaster')
 					&& savedSelection.some(id => this.getRowIndexByID(id) === false)) {
 					// In duplicates view, select the next set on delete
 					if (collectionTreeRow.isDuplicates()) {
--- a/chrome/content/zotero/xpcom/data/items.js
+++ b/chrome/content/zotero/xpcom/data/items.js
@@ -1023,13 +1023,14 @@ Zotero.Items = function() {
 			}
 			
 			item.setField('dateAdded', earliestDateAdded);
+
+			// Hack to remove master item from duplicates view without recalculating duplicates
+			// Pass force = true so observers will be notified before this transaction is committed
+			yield Zotero.Notifier.trigger('removeDuplicatesMaster', 'item', item.id, null, true);
 			
 			for (let i in toSave) {
 				yield toSave[i].save();
 			}
-			
-			// Hack to remove master item from duplicates view without recalculating duplicates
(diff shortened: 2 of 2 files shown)


adomasven:
For the record the issue is reproduced by clicking on the last item of the set of duplicates and then clicking merge. But even when merging after selecting the first item you only still get a selection of a single item.

Either way this is not the way to address it. Item tree and collection tree UI is highly decoupled from any data-layer code and for good reason via `Zotero.Notifier`. [Here](https://github.com/zotero/zotero/blob/bca76319ce677052da058f0e422ea9ac9ba39f3e/chrome/content/zotero/xpcom/data/items.js#L1032) a `removeDuplicatesMaster` event is issued and it's handled [here](https://github.com/zotero/zotero/blob/227b6354de5908d4d36d86b7a7c60b564dedda47/chrome/content/zotero/itemTree.jsx#L494). There's also [some code here](https://github.com/zotero/zotero/blob/227b6354de5908d4d36d86b7a7c60b564dedda47/chrome/content/zotero/itemTree.jsx#L765-L772) that was supposed to already do it in the right place and in the right way. The code is kind of hairy and complex and could certainly be refactored, but just getting an efficient, feature-compatible and bug-free version of a HTML tree was a lot of work so I skipped refactoring. I can take this bug if it's too much.

AbeJellinek:
Yeah, I know this is kind of gross. Is the first commit’s approach OK? It just adds `removeDuplicatesMaster` to the list of notifier events that will cause the next duplicates set to be selected. That works, except when you’ve sorted by Date Modified.

AbeJellinek:
It might work to keep the first commit’s approach and fix `Items.merge` so it only saves the master item once at the very end and sends the `removeDuplicatesMaster` notification immediately beforehand. (Maybe it worked that way before.) I did try that - I got rid of every intermediate save I could find - but something was still updating the item’s dateModified and moving the item to the top of the list before the notifier event was sent. I’ll give it another go.

AbeJellinek:
OK, final approach, much simpler and works in all sorts. We just have to dispatch the notifier event with `force` = `true` so we can await it and _then_ save everything. This means the master item will disappear even if the transaction commit fails, but the only SQL executed after the notifier event is dispatched is a bunch of saves, so I don't think that should ever happen. All you'd have to do to get everything back to normal is to exit and reenter the Duplicate Items view.

adomasven:
This only works for me if I select the last item in the set of duplicates. If I select the first one (or any non-last one when there is more than 2 duplicates) it only selects the next item.

AbeJellinek:
Odd. You mean that you click on an item in Duplicate Items to select its entire set and then merge its set together, but it only selects the entire next set if the item that you initially clicked was the bottommost in its set?

AbeJellinek:
What are you sorting by?

AbeJellinek:
OK, I can reproduce that with one particular set of duplicates in my library, sorting by Date Modified. I'll look into it.

AbeJellinek:
OK, let me know if that fixed it for you.

adomasven:
Yep, this is good to merge

AbeJellinek:
Great, thank you!

Issue #2381, opened 2022-03-04
HTML tree: Fonts are cut off with certain OS font settings


dstillman:
https://forums.zotero.org/discussion/94852/field-size-not-displaying-correctly

Maybe we can better handle this?

adomasven:
I've looked and there's no way to detect this in the JS environment. All rendered fonts are upscaled and reported via `window.getComputedStyle()` as such, but there's no way to know what the base font was or what's the scale amount. Font scaling like this isn't really handled very well by most software, we're not an outlier. We handle proper device pixel scaling fine, so I think this isn't very urgent since this kind of font scaling isn't part of any proper standard.

dstillman:
I think this is a more general issue — we've gotten a bunch of reports of this on Windows and Linux already.

https://forums.zotero.org/discussion/95145/creator-column-misalinged-after-update
https://forums.zotero.org/discussion/95172/font-rendering

In the tag selector we get the font size by using Canvas:

https://github.com/zotero/zotero/blob/7182531ec90376a676fa6fd602ddd1e570232942/chrome/content/zotero/containers/tagSelectorContainer.jsx#L374-L390

It's a hack, but it detects the font size fairly reliably. Can we do something here (just once, on first tree init)?

adomasven:
It is an issue anywhere where the OS allows to force fonts to be rendered bigger than what the software wants to render them at. The only reason we don't see this in more places with Zotero is that other places where we render fonts we have margins where the oversized font can extend into. This is not the case for the HTML tree. Our current implementation allows us to choose a row height which nicely gets adjusted for DPI such that we are more free to display what we want in the tree. We can do multi-level rows and specify row heights in px without worrying that things will break and overlap. Except in the scenario where the user is overriding OS font rendering. If we instead base row heights on rendered font-heights we'll lose a certain amount of flexibility and will need to do additional work for multi-level rows.

This would be somewhat easier to address if we were specifying initial font-size on the top element in pixels, but right now the default rendered font size for me is 14.666666px on Linux and it's a different number on Windows and macOS. So we have no baseline known font size that we can work from when trying to figure out by what factor the font size is increased.

dstillman:
https://forums.zotero.org/discussion/95582/under-ubuntu-the-font-of-the-zotero-interface-is-only-half-displayed

PR #2452, opened 2022-03-23
Truncate and ellipsize multiline values in tree

### LINKED ISSUE #2450: Extra column shows sliver of second line
https://forums.zotero.org/discussion/95484/display-issue-multiple-lines-in-extra-column

I guess in theory this would happen with any multiline value displayed in the tree.

AbeJellinek:
Fixes #2450.

### DIFF
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -2847,7 +2847,12 @@ var ItemTree = class ItemTree extends LibraryTree {
 		else if (column.dataKey === 'hasAttachment') {
 			return this._renderHasAttachmentCell(index, data, column);
 		}
-		let cell = renderCell.apply(this, arguments);
+		if (typeof data == 'string' && data.includes('\n')) {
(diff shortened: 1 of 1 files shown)


adomasven:
Can this not be addressed with pure CSS? I think it may be related to https://github.com/zotero/zotero/issues/2381 but it's the reverse issue - the user has OS-based font downsizing, so addressing that would fix this too probably.

AbeJellinek:
The issue this PR is addressing isn’t related to OS font downsizing - for me a little sliver of the second line of Extra appears on a Mac with default font settings. It might be possible to fix it by tweaking the `line-height`, but `text-overflow` only handles horizontal overflow and the vertical equivalent (`line-clamp`) isn’t available to us.

adomasven:
I'm closing this since I'm addressing it with CSS as part of a fix for https://github.com/zotero/zotero/issues/2381. Thanks for going through the trouble of trying to fix this Abe, though!

PR #2468, opened 2022-03-24
Update zotero.properties

huaibovip:


dstillman:
Localization changes need to happen on [Transifex](https://www.transifex.com/zotero/zotero), but can you explain why this should be reverted?

PR #2488, opened 2022-03-28
Item Tree: Attempt to fix OS font scaling

### LINKED ISSUE #2381: HTML tree: Fonts are cut off with certain OS font settings

### LINKED ISSUE #2450: Extra column shows sliver of second line

adomasven:
Closes #2381, closes #2450

A slight rethinking to the way row heights are specified for the item tree. They will now be specified in multiples of lines of text we expect to render for whatever the CSS and then OS-level font-scaling expected rendered font size for the tree rows is. The default is 1 line of text per one row. This is required to deal with OS-level font-scaling. A side effect of this is that row-heights may be slightly different for different font-sizes (as chosen in the Zotero interface) when compared to before with the XUL tree. I've also set a minimum row height of 20px. Tested on all platforms and looks fine to me.

### DIFF
--- a/chrome/content/zotero/components/virtualized-table.jsx
+++ b/chrome/content/zotero/components/virtualized-table.jsx
@@ -34,7 +34,7 @@ const { injectIntl } = require('react-intl');
 const { IconDownChevron, getDOMElement } = require('components/icons');
 
 const TYPING_TIMEOUT = 1000;
-const DEFAULT_ROW_HEIGHT = 20; // px
+const MINIMUM_ROW_HEIGHT = 20; // px
 const RESIZER_WIDTH = 5; // px
 
 const noop = () => 0;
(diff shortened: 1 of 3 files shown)


dstillman:
Bumped up a couple font sizes on macOS, this is way tighter. Is this adjustable?

Screen Shot 2022-03-29 at 12 39 09 AM

Screen Shot 2022-03-29 at 12 37 36 AM

dstillman:
(What we had before might have been too much, but this is definitely too tight.)

adomasven:
I've updated it with percentage-based padding.

A11en0:
Sorry, it does not work on my computer:

System: Debian 10
Zotero Version: 6.0.4

image

dstillman:
@A11en0: The fix is in the <a href="https://www.zotero.org/support/dev_builds">Zotero beta</a>, not in 6.0.4. It will be included in 6.0.5 soon.

Please post all bug reports, feature requests, and other feedback to the <a href="https://forums.zotero.org/">Zotero Forums</a>, not GitHub.

Issue #2341, opened 2022-02-09
Update CSL validation for 1.0.2 styles


dstillman:
https://forums.zotero.org/discussion/comment/399020/#Comment_399020

dstillman:
(I'm concerned that this requires recompiling csl-validator.js…)

PR #2493, opened 2022-03-30
Update csl validator for CSL 1.0.2

adomasven:
Addresses #2341. Note that citeproc-rs doesn't seem to support CSL 1.0.2. I'll create an issue for cormac.


dstillman:
Something's wrong:

```
zotero(1)(+0000227): Style from http://www.zotero.org/styles/cell failed to validate:

style.xml:22:19: error: invalid data or text not allowed
style.xml:23:20: error: invalid data or text not allowed
```

That's https://github.com/zotero/zotero/blob/5ca8486047b546f50f8eeacb9bb0f352d75c4779/test/tests/data/cell.csl, which [validates as 1.0.2](https://validator.citationstyles.org/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fzotero%2Fzotero%2F5ca8486047b546f50f8eeacb9bb0f352d75c4779%2Ftest%2Ftests%2Fdata%2Fcell.csl&version=1.0.2).

dstillman:
Supposedly these lines:

https://github.com/zotero/zotero/blob/5ca8486047b546f50f8eeacb9bb0f352d75c4779/test/tests/data/cell.csl#L22-L23

adomasven:
Ok, turned out to be an easy fix. Embedded rnc schema file strings had single `\` instead of `\\` so some checks were failing. I ran the validator over the whole styles codebase and it only fails for some styles that have special symbols in URLs, namely `,` and some others, but RelaxNG itself also fails on those and it did prior to this update too. Citationstyles.org uses a different codebase for their validator, so theirs doesn't fail for those styles. Overall that's a relatively minor issue.

PR #2495, opened 2022-03-30
add webstorm compatibility

prototype99:
it's pretty basic, just adds some metadata for webstorm


adomasven:
This isn't necessary. Zotero developers use a variety of IDEs, including ones by IDEA and everyone sets up their development environment as they see fit. If you are annoyed with `.idea` files in your git commands add `.idea` to `~/.gitignore_global`.

PR #2500, opened 2022-03-31
Unload tabs

### LINKED ISSUE #2383: PDF Reader: Keep unused tabs out of memory
https://forums.zotero.org/discussion/94546/possible-mac-memory-issue

How hard would it be to 1) not load PDF tabs at startup until they're first accessed and 2) unload tabs after a certain amount of time of inactivity?

Particularly (1) would be good, if we're not doing that — that's always been one of my favorite features of Firefox.

mrtcode:
- Keep all tabs unloaded on Zotero opening.
- Keep loaded only the last five selected tabs.
- Keep loaded only in the last 24h selected tabs.

Fixes #2383

### DIFF
--- a/chrome/content/zotero/contextPane.js
+++ b/chrome/content/zotero/contextPane.js
@@ -181,13 +181,13 @@ var ZoteroContextPane = new function () {
 				// It seems that changing `hidden` or `collapsed` values might
 				// be related with significant slow down when there are too many
 				// DOM nodes (i.e. 10k notes)
-				if (Zotero_Tabs.selectedIndex == 0) {
+				if (Zotero_Tabs.selectedType == 'library') {
 					_contextPaneSplitter.setAttribute('hidden', true);
 					_contextPane.setAttribute('collapsed', true);
 					_tabToolbar.hidden = true;
 					_tabCover.hidden = true;
 				}
-				else {
+				else if (Zotero_Tabs.selectedType == 'reader') {
 					var reader = Zotero.Reader.getByTabID(Zotero_Tabs.selectedID);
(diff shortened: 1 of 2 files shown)


dstillman (commented):
1. I'm seeing `JavaScript error: chrome://zotero/content/xpcom/reader.js, line 954: TypeError: this._iframeWindow is null` in debug output when switching tabs.

2. It looks like we're closing a tab based on the time of selection, but that means that if you haven't used Zotero in a day and switch back to it, and start reading the front-most PDF, and then switch away from it, it's immediately unloaded. Since we never unload the front-most PDF anyway, can we instead set the time based on when the tab is switched away from rather than selected?

mrtcode:
Both issues fixed.

dstillman:
Looks good, but need to resolve a conflict with `master` now. If I do a simple merge, I get this when selecting a new tab:

> JavaScript error: chrome://zotero/content/tabs.js, line 344: TypeError: document.querySelector(...) is null

When the tab is unloaded, `select()` closes the old tab and adds a new one with a new id, but since that's rendered via React, the DOM node for the tab doesn't yet exist in the `select()` triggered by the `add()`, so a `document.querySelector()` call in `select()` fails. Maybe has to be a flag on the `open()` so it can be done by the React component itself after the component is mounted?

Alternatively, is there a way to do this without the close/open stuff?

mrtcode:
I think `setTimeout(…, 0)` is enough to allow React to render the nodes.

> Alternatively, is there a way to do this without the close/open stuff?

I was evaluating various options before, but all of them seemed more complicated or less clean.

PR #2506, opened 2022-04-04
Make tab bar scrollable

### LINKED ISSUE #2232: Improve how tabs are displayed when there are too many of them
https://forums.zotero.org/discussion/92614/pdf-reader-beta-tab-titles-not-shown
Screenshot 2021-10-29 at 09 56 45

mrtcode:
Fixes #2232


dstillman (commented):
- If I'm on the last tab and restart Zotero, after restart I'm missing the right-hand border of the last tab. I have to scroll slightly right to see it.
- Not a blocker, since this is easily an improvement over what we had before, but can we do `<` and `>` buttons on the edges to indicate that there are more tabs, the way Firefox does? Right now the only indication is truncation (which depending on positioning might just be a missing border).

mrtcode:
Both issues fixed.

Also scroll speed is now normalized on Windows. It would be good to test on Linux, but I currently can't do that.

PR #2511, opened 2022-04-06
Allow creating and editing tables in note-editor

### LINKED ISSUE #2480: Note editor: Table editing support
https://forums.zotero.org/discussion/95227/zotero-6-note-editor-html-view and others

People want tables.

mrtcode:
Currently only row/column insertion/deletion is supported. Probably no need to hurry with advanced features like merging/splitting cells, setting table header.

I took some inspiration from Google Docs for the table context menu.

You will probably see displaced table resizing handles, the first time when you open a note with table and focus the table. That's Firefox bug, and resizing handles are completely disabled from Firefox 64. That might be very time consuming to work around, therefore I'm leaving this unfixed. Maybe we could do something with prefs…

Fixes #2480

### DIFF
--- a/chrome/locale/en-US/zotero/zotero.properties
+++ b/chrome/locale/en-US/zotero/zotero.properties
@@ -1352,6 +1352,14 @@ noteEditor.applyAnnotationColors = Show Annotation Colors
 noteEditor.removeAnnotationColors = Hide Annotation Colors
 noteEditor.addCitations = Show Annotation Citations
 noteEditor.removeCitations = Hide Annotation Citations
+noteEditor.insertTable = Insert Table
+noteEditor.insertRowBefore = Insert Row Above
+noteEditor.insertRowAfter = Insert Row Below
+noteEditor.insertColumnBefore = Insert Column Left
+noteEditor.insertColumnAfter = Insert Column Right
+noteEditor.deleteRow = Delete Row
+noteEditor.deleteColumn = Delete Column
+noteEditor.deleteTable = Delete Table
 
 pdfReader.annotations = Annotations
 pdfReader.showAnnotations = Show Annotations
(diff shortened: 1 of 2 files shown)


dstillman:
I seem to be missing the left border in all tables:

Screen Shot 2022-04-11 at 4 03 03 AM

dstillman:
I don't think we need to support indenting in tables — when I press Tab, I expect to go to the next cell.

dstillman:
The drag handle for a table generally disappears before I can get my cursor over it, but if I manage (by moving the mouse very fast) and drag the table, it moves the table but leaves an empty table of the same size in the original position.

dstillman:
> when I press Tab, I expect to go to the next cell

Similarly, Tab in the last cell should add a new row.

dstillman:
Cell borders don't appear in reports, and the cells are very big:

Screen Shot 2022-04-11 at 4 12 32 AM

(Not a blocker.)

dstillman:
On the resize handles, we're hiding those somehow after the first display? Just no way to hide the first time in a session?

dstillman:
OK, that's all I have for now. This is great!

mrtcode:
> The drag handle for a table generally disappears before I can get my cursor over it

I guess I'll need to partially rewrite how drag handles work, and rely more on JS instead of CSS, because there are more nodes that doesn't work well.


> On the resize handles, we're hiding those somehow after the first display? Just no way to hide the first time in a session?

I already tried various things to determine why this doesn't work the first time in a session, but without success. If you think it's worth to dedicate more I can investigate further. But, as I said, from Firefox 64 this no longer matters.

mrtcode:
All things except Firefox resizing handles are fixed.

For now I also disabled column resizing because I noticed that changes aren't reflected on a raw HTML which means this doesn't work with TinyMCE and reports. Also column resizing doesn't work well in responsive design.

This can be tested now again.

PR #2537, opened 2022-04-12
getSortTitle: Only strip initial punctuation clusters

### LINKED ISSUE #2535: Don't strip non-leading punctuation when sorting?
https://forums.zotero.org/discussion/comment/404372/#Comment_404372

I think I'd agree with all the "expected" sorts here. Did we have a good reason for stripping non-leading punctuation?

The hardest one is "@", because (if I recall) we're just using the punctuation class, but maybe we want to explicitly add back in certain characters?

Follow-up to #2442

AbeJellinek:
Removes:

- Any punctuation of any type at the beginning of the string besides @, #, and *, including the *same character* repeated multiple times. So `-- title` would become `title` but `-_ title` would become `_ title`.
- Any [opening punctuation](https://www.fileformat.info/info/unicode/category/Ps/list.htm) or neutral quotes immediately following it.

I think this has a better balance between removing non-meaningful characters in titles, preserving meaningful ones (like the @ example in the forum thread), and letting people use title hacks if they really, really want to.

See test cases for some examples. Includes a test which sorts the examples given in the thread.

Fixes #2535.

### DIFF
--- a/test/tests/itemsTest.js
+++ b/test/tests/itemsTest.js
@@ -1192,14 +1192,17 @@ describe("Zotero.Items", function () {
 			}
 		});
 
-		it("should strip any punctuation at the beginning of the string besides the @ sign", function () {
+		it("should strip any punctuation at the beginning of the string besides @, #, and *", function () {
 			let tests = [
 				['_title', 'title'],
 				['-title', 'title'],
 				['-- longer title', 'longer title'],
 				['-_ longer title with different second character', '_ longer title with different second character'],
 				['"Quoted title', 'Quoted title'],
-				['@zotero on Twitter', '@zotero on Twitter']
+				['@zotero on Twitter', '@zotero on Twitter'],
+				['#hashtag', '#hashtag'],
+				['*special', '*special'],
+				['**repeated', '**repeated']
 			];
 
 			for (let [input, expected] of tests) {
--- a/test/tests/itemsTest.js
+++ b/test/tests/itemsTest.js
@@ -1197,20 +1197,8 @@ describe("Zotero.Items", function () {
 				['_title', 'title'],
 				['-title', 'title'],
 				['-- longer title', 'longer title'],
+				['-_ longer title with different second character', '_ longer title with different second character'],
 				['"Quoted title', 'Quoted title']
-			];
-
-			for (let [input, expected] of tests) {
-				assert.equal(Zotero.Items.getSortTitle(input), expected);
-			}
-		});
-
-		it("should strip quotes", function () {
-			let tests = [
-				['A "title"', 'A title'],
-				['A “title”', 'A title'],
-				[' xyz ”””', 'xyz'],
-				['‘Punctuation’', 'Punctuation']
 			];
 
 			for (let [input, expected] of tests) {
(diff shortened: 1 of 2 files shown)


test/tests/itemsTest.js
```diff
@@ -1192,54 +1192,67 @@ describe("Zotero.Items", function () {
...
-		it("should strip any punctuation at the beginning of the string", function () {
+		it("should strip any punctuation at the beginning of the string besides the @ sign", function () {
 			let tests = [
 				['_title', 'title'],
 				['-title', 'title'],
 				['-- longer title', 'longer title'],
-				['"Quoted title', 'Quoted title']
+				['-_ longer title with different second character', '_ longer title with different second character'],
+				['"Quoted title', 'Quoted title'],
+				['@zotero on Twitter', '@zotero on Twitter']
```
dstillman, line 1202: I see this and I think…are we sure that shouldn't sort as 'z'?

I would agree that `"@" Sign. Its accidental history.` should sort on `@`, but this seems to me like it should sort by the username.

qqobb: Saving Zotero's Twitter page results in the Title field `Zotero (@zotero) / Twitter`, so the issue might not arise in that case.

If you have several items whose Title field starts with `@`, it could be useful to group them together when the items list is sorted by the Title field. In the examples below, `@` is used to mark a username, so the items are of a similar type. In the same way you could argue that items that start with `#` should be grouped together.

Expected (case-insensitive):
```
#BLM
#hashtag
#jesuischarlie
#MeToo
#OccupyWallStreet
@BBCNews
@nytimes
@washingtonpost
@zotero
```

Ignoring `@` and `#`:
```
@BBCNews
#BLM
#hashtag
#jesuischarlie
#MeToo
@nytimes
#OccupyWallStreet
@washingtonpost
@zotero
```

I would suggest to include the characters in the list below when sorting. They could have a meaning and they aren't used as punctuation in normal writing.

https://www.unicode.org/faq/punctuation_symbols.html

> Q: Is there a list of Unicode punctuation characters that should also be treated as symbols?
> 
> There is no comprehensive list, but here are some of the most common examples of punctuation characters with conflicting uses as symbols:
> U+0023 ( # ) NUMBER SIGN
> U+0026 ( & ) AMPERSAND
> U+0040 ( @ ) COMMERCIAL AT
> U+0025 ( % ) PERCENT SIGN
> U+2030 ( ‰ ) PER MILLE SIGN
> U+2031 ( ‱ ) PER TEN THOUSAND SIGN
> U+002A ( * ) ASTERISK
> U+2020 ( † ) DAGGER
> U+2021 ( ‡ ) DOUBLE DAGGER
> U+203B ( ※ ) REFERENCE MARK

qqobb: > Q: What about the the U+002D HYPHEN-MINUS?
> 
> The character U+002D HYPHEN-MINUS is used both as a hyphen (punctuation mark) and as a minus sign (mathematical symbol), with the intended meaning only apparent from context. Because the General_Category can only take a single value for each character, the U+002D HYPHEN-MINUS is given the value Pd (dash), similar to other dash characters. At the same time, unambiguous Unicode characters also exist for each of these functions: U+2010 ( ‐ ) HYPHEN and U+2212 ( − ) MINUS SIGN. The latter has been given the General_Category of Sm (mathematical symbol).

Perhaps also include U+002D when sorting? It would be odd if U+2212 ( − ) MINUS SIGN is included since it's a mathematical symbol, but U+002D HYPHEN-MINUS is not.

AbeJellinek: >Perhaps also include U+002D when sorting? It would be odd if U+2212 ( − ) MINUS SIGN is included since it's a mathematical symbol, but U+002D HYPHEN-MINUS is not.

Maybe in principle, but U+002D is more commonly used, especially in places like filenames, and often stands in for the real minus sign.

I do think keeping the # is reasonable. If an item is intentionally saved with that or an @ at the beginning of the title, I think it's likely to be an important part of the metadata and makes sense to sort on. Asterisk also makes sense to me.

Not so convinced about the rest, though. I'm not sure why you'd ever have a meaningful ampersand at the beginning of a title, and I don't think we have too many people worried about whether their items beginning in characters like U+2031 PER TEN THOUSAND SIGN are sorted to the top.

dstillman: > I do think keeping the # is reasonable. […] Asterisk also makes sense to me.

I'm fine with sorting on those, and `@`.

PR #2538, opened 2022-04-12
Sort call numbers alphabetically, handle Dewey specially

### LINKED ISSUE #2194: Don't use natural sorting for Call Number field
https://forums.zotero.org/discussion/89408/call-number-sorting

AbeJellinek:
If a call number looks like Dewey (begins with three digits, then optionally a dot and one to four digits, then optionally a slash and three letters), we split it into parts and sort each using string comparison in order to get [this ordering](https://www.mdek12.org/sites/default/files/Offices/MDE/OA/OEER/Library%20Services/Section%206/Dewey%20Decimal%20Classification%20(1).pdf).

If it doesn't, we still sort using string comparison - locale collation will treat them like cardinal numbers, which I think isn't appropriate in many cases.

Fixes #2194.

### DIFF
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -1352,6 +1352,28 @@ var ItemTree = class ItemTree extends LibraryTree {
 				if (sortField == 'hasAttachment') {
 					return fieldB - fieldA;
 				}
+
+				if (sortField == 'callNumber') {
+					let deweyRe = /^(\d{3})(?:\.(\d{1,4}))?(?:\/([a-zA-Z]{3}))?$/;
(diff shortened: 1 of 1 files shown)


MartinRushton:
The [School Library web site](https://www.mdek12.org/sites/default/files/Offices/MDE/OA/OEER/Library%20Services/Section%206/Dewey%20Decimal%20Classification%20(1).pdf)  is a simplified version.  Dewey always starts with three numbers.  If there are only three numbers, then that's all that there is.  Adding a slash and letters is a local extension and is non-standard, though not uncommon.  If there are more than three numbers, then a decimal point is put in and the number extends indefinitely.  For instance "Diving and snorkeling guide to Scotland" is classified at 797.2309411.

AbeJellinek:
Thanks, see above commit. Should behave correctly now.

PR #2539, opened 2022-04-12
Show label of translator that causes import error

### LINKED ISSUE #1966: Identify translator being used for an import failure
https://forums.zotero.org/discussion/87984/bibtext-import-problem-not-working

This isn't our bug, and there's not a great way to address it, but maybe we can show the translator label in the error message without it being too weird. If this error said "Better BibTeX" somewhere, that could at least point people in the right direction.

AbeJellinek:
Fixes #1966


chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -774,6 +774,7 @@ fileInterface.viewSupportedFormats = View Supported Formats…
 fileInterface.untitledBibliography = Untitled Bibliography
 fileInterface.bibliographyHTMLTitle = Bibliography
 fileInterface.importError 		= An error occurred while trying to import the selected file. Please ensure that the file is valid and try again.
+fileInterface.importError.translator = An error occurred while trying to import the selected file with “%S”. Please ensure that the file is valid and try again.
```
dstillman, line 777: We can be a little less verbose:

`An error occurred importing the selected file with “%S”. Please ensure that the file is valid and try again.`

chrome/content/zotero/fileInterface.js
```diff
@@ -1038,6 +1038,22 @@ var Zotero_File_Interface = new function() {
 		// polling executed by the main window because current (wizard) window will be closed
 		Zotero.getMainWindow().setTimeout(this.authenticateMendeleyOnlinePoll.bind(this, win), 200);
 	};
+
+	/**
+	 * Generate an error string reporting a translation failure. Includes the
+	 * label of the running translator if available.
+	 *
+	 * @param {Zotero.Translate?} translate
```
dstillman, line 1046: `@param {Zotero.Translate} [translate]`

PR #2543, opened 2022-04-14
Add sorting options and column picker to View menu

### LINKED ISSUE #2273: Add items tree sorting options to View menu
In https://github.com/zotero/zotero/pull/2263#pullrequestreview-824718234, I brought up accessing the column headers via the keyboard, but that's not the way to do it, of course — it should be in the View menu.

E.g., in Finder, there's a View → Sort By menu that's equivalent to clicking on a column header, and it also has numbered key bindings for each column (e.g., Cmd-Option-Ctrl-1, though weirdly I can't find a way to sort descending via the keyboard — seems like at least triggering the same column twice should toggle the direction.

And then, separately, Finder has a Show View Options option that opens a dialog to hide/show panes, though that's in part to enable folder-specific settings. We could do it in a separate menu that just mirrored the context menu.

AbeJellinek:
ItemTree now exposes `toggleSort` (with a convenience `countVisible` parameter so ZP doesn't need to know anything about which columns are visible) and a couple menu builders.

Fixes #2273.

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -558,8 +558,7 @@ var ZoteroPane = new function()
 			key.id = 'key_sortCol' + i;
 			key.setAttribute('modifiers', Zotero.isMac ? 'accel alt control' : 'accel alt');
 			key.setAttribute('key', (i + 1) % 10);
-			// addListener('command', ...) doesn't trigger unless attached to a menuitem
-			key.setAttribute('oncommand', `ZoteroPane.itemsView.toggleSort(${i}, true)`);
+			key.addEventListener('command', () => ZoteroPane.itemsView.toggleSort(i, true));
 			sortSubmenuKeys.append(key);
 		}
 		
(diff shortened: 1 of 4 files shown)


chrome/content/zotero/zoteroPane.js
```diff
@@ -158,6 +158,17 @@ var ZoteroPane = new function()
 		
 		Zotero.Keys.windowInit(document);
 		
+		let sortSubmenuKeys = document.getElementById('sortSubmenuKeys');
+		for (let i = 0; i < 10; i++) {
+			let key = document.createElement('key');
+			key.id = 'key_sortCol' + i;
+			key.setAttribute('modifiers', Zotero.isMac ? 'accel alt control' : 'accel alt');
```
AbeJellinek, line 165: Windows Explorer doesn't have sort keyboard shortcuts to copy here, but Ctrl-Alt-[num] isn't used system-wide and seems common for numbered presets and things like that.

PR #2569, opened 2022-04-25
Sort LC call numbers better, sort integers as integers

### LINKED ISSUE #2567: Sort Library of Congress call numbers correctly
The Dewey people are happy, but now the LoC people are unhappy:

https://forums.zotero.org/discussion/96616/changes-to-sorting-by-call-number-in-zotero-6-0-5
https://forums.zotero.org/discussion/96677/loc-numbers-not-sorting-correctly

### LINKED ISSUE #2570: Sort plain integers in Call Number field
Will get more info, but this person is maybe using it for plain integers?

https://forums.zotero.org/discussion/96701/bad-signature-field-ordering-report-id-857547486

AbeJellinek:
Not strictly "correctly," due to the issues mentioned [here](https://github.com/zotero/zotero/issues/2567#issuecomment-1108819863), but in a way that handles the classification field correctly.

Fixes #2567, fixes #2570.

### DIFF
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -1355,8 +1355,10 @@ var ItemTree = class ItemTree extends LibraryTree {
 
 				if (sortField == 'callNumber') {
 					let deweyRe = /^(\d{3})(?:\.(\d+))?(?:\/([a-zA-Z]{3}))?$/;
+					let lcWithClassificationRe = /^([a-zA-Z]{1,3}\d+\s*\.)(?=[a-zA-Z])/;
 					let splitA = fieldA.toLowerCase().replace(/\s/g, '').match(deweyRe);
 					let splitB = fieldB.toLowerCase().replace(/\s/g, '').match(deweyRe);
+
 					if (splitA && splitB) {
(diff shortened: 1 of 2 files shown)


dstillman:
Can we have a test for this?

dstillman:
(Or maybe we don't have tests for Dewey either? But this seems like the sort of thing we should test, or else it's going to regress as we continue to refine it.)

AbeJellinek:
Yeah, but all the call number sorting stuff needs to be extracted from itemTree.jsx, since there's no great way to test it when it's a branch within a function within a function there. Zotero.Utilities.Item? Or Utilities.Internal?

dstillman:
`Utilities.Item` seems good. We could use this in other contexts.

AbeJellinek:
OK, opened a PR at zotero/utilities#8 and added tests here.

dstillman:
Well, or I guess if we're moving tests to the submodule we can hold off on this.

AbeJellinek:
It seems to me like it'll be easiest / give us the clearest history to merge this here and then move tests to the submodule all at once.

AbeJellinek:
(If the random tests that sometimes fail will ever cooperate...)

PR #2586, opened 2022-04-29
Fix sorting by attachment

### LINKED ISSUE #2416: HTML tree: Can't sort by attachment
https://forums.zotero.org/discussion/95017/beta-6-0-cant-sort-by-attachment

AbeJellinek:
In the process, simplified sorting and displaying attachment states a fair amount by making `getBestAttachmentState` treat the item it's called on as the "best attachment" when called on an attachment item. That removed the need for duplicate logic to determine attachment icons for standalone attachments and even more duplicate logic to figure out sort positions.

1. Is that a reasonable change?
2. And if so, should `getBestAttachment` and `getBestAttachments` also support attachment items?

Fixes #2416.

### DIFF
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -3791,7 +3791,7 @@ var ItemTree = class ItemTree extends LibraryTree {
 
 	_canGetBestAttachmentState(item) {
 		return (item.isRegularItem() && item.numAttachments())
-			|| (item.isAttachment() && item.isTopLevelItem());
+			|| (item.isFileAttachment() && item.isTopLevelItem());
 	}
 	
 	_isOnlyEmoji(str) {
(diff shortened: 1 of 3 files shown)


dstillman:
I'm getting this error:

> JavaScript error: chrome://zotero/content/xpcom/data/item.js, line 2721: Error: Zotero.Item.fileExists() cannot be called on link attachments

chrome/content/zotero/itemTree.jsx
```diff
@@ -3828,6 +3788,11 @@ var ItemTree = class ItemTree extends LibraryTree {
 		}
 		return icon;
 	}
+
+	_canCalculateAttachmentState(item) {
```
dstillman, line 3792: → `_canGetBestAttachmentState()`? `calculate` kind of makes it sound like something else.

AbeJellinek:
OK, should be fixed.

chrome/content/zotero/xpcom/data/item.js
```diff
@@ -3740,7 +3740,7 @@ Zotero.Item.prototype.getBestAttachmentState = async function () {
 	else {
 		type = 'other';
 	}
-	var exists = await item.fileExists();
+	var exists = !item.isFileAttachment() || await item.fileExists();
```
dstillman, line 3743: Wouldn't a better fix just be to change `_canGetBestAttachmentState()` to return false for non-file attachments?

AbeJellinek: Sure, makes sense.

Issue #2574, opened 2022-04-26
Don't change Year and Notes column widths when changing window size


dstillman:
Not a full fix for https://forums.zotero.org/discussion/96612/, but these two columns in particular should pretty much never change size once set, so if we can avoid distributing extra width to them during resizes, that would be a good start.

dstillman:
(While we're at it, the initial size for Year, and the min-width, would ideally be dynamically set to four numbers in the current font, if we have an easy way of doing that.)

adomasven:
> (While we're at it, the initial size for Year, and the min-width, would ideally be dynamically set to four numbers in the current font, if we have an easy way of doing that.)

We don't. Set in px. We could technically at least set something that's better than `treeWidth/numColumns`, but whatever we do, it will only affect new installs that don't have custom prefs set already.

aywi:
@dstillman @adomasven For the latest Zotero 6.0.7, there is a new problem with column width adjustment and I found someone else already posted here (I'm a new user of the forum and cannot post so I choose to go to GitHub): <https://forums.zotero.org/discussion/97077/problem-with-column-width-adjustment>. Now the column width adjustment with the mouse is highly unstable (suddenly all columns change and move to the left or right) and becomes a nightmare for me.

It would be much better if there is an option provided to go back to the previous behavior. Or, just like Windows Explorer or many other Windows applications, to provide options with: (1) adjusting the column width to the maximum width of items in the current column; (2) setting the pixel value of the column width manually. Thanks!

Platform: Windows 10 (Version 21H2)

adomasven:
See my comment on the Zotero forums. It's best to continue the conversation there, where all support and bug identification with users is done.

aywi:
@adomasven Thanks. The user [marcinkoculak](https://forums.zotero.org/profile/5276922/marcinkoculak) isn't me though, as I still cannot post any comment on the forum up to now (new user).

PR #2588, opened 2022-05-02
Improve column resizing robustness

### LINKED ISSUE #2587: HTML tree: Remember column widths when maximizing/restoring window
https://forums.zotero.org/discussion/96612/ (partly about #2574 and partly about this)
https://forums.zotero.org/discussion/96875/remember-column-widths

I'm not sure what exactly the XUL tree did, but given the number of reports I assume this is some sort of regression. Ideally you could maximize and minimize the window (or grow and shrink on non-Windows) and end up with the same column widths as before.

adomasven:
A slight redesign of the column resizing code which restores previous XPCOM tree behavior:

- Previous code calculated and stored column widths based on item tree width. This meant that if you resized the item tree (or Zotero window) the column widths would get recalculated. Making the tree wider and narrower repeatedly would cause rounding errors to accumulate and all column widths to drift towards equal width. This code instead stores all column widths normalized to item tree width of 8192 and no recalculation is ever done when resizing the tree.
- Fixed column width calculation code to be more robust on macOS and not have the column separators get out of sync between the header and cells.

Would be good to see some testing before merging in case I missed something.

Closes #2587. Addresses #2574 to a certain degree and since it restores previous behavior I think we can close.

### DIFF
--- a/chrome/content/zotero/components/virtualized-table.jsx
+++ b/chrome/content/zotero/components/virtualized-table.jsx
@@ -36,6 +36,9 @@ const { IconDownChevron, getDOMElement } = require('components/icons');
 const TYPING_TIMEOUT = 1000;
 const MINIMUM_ROW_HEIGHT = 20; // px
 const RESIZER_WIDTH = 5; // px
+const COLUMN_MIN_WIDTH = 20;
+const COLUMN_NORMALIZATION_WIDTH = 8192;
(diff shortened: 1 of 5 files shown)


dstillman:
Looks good.

But we should keep #2574 open. The previous behavior there wasn't ideal.

PR #2589, opened 2022-05-02
Pin library tab

### LINKED ISSUE #2575: Pin library tab
It should be always pinned. And we probably want to show an icon only, like browsers do. But does the icon need to change depending on which collection is selected?

https://forums.zotero.org/discussion/96684/keep-main-tab-fixed

mrtcode:
The only nuance is that it's getting the current collection icon in [a hacky way](https://github.com/zotero/zotero/compare/master...mrtcode:pin-library-tab?expand=1#diff-014a38101f2c50e35e6714dcee641532b66acde664b882a6ad87446cb62a6cc4R204-R208).

Fixes #2575


dstillman:
Re: the icon, if it doesn't need to be React we can get it from `.treeViewImage` on the row's ref (`Library`, `Collection`, etc.). @adomasven, is there an easy way to get that from the outside in the new tree?

PR #2595, opened 2022-05-05
Add support for PDF page deletion and rotation

### LINKED ISSUE #2561: PDF reader: Remove/move pages
With the removal of Store Annotations in File, it's more necessary to be able to do this within Zotero, since there's no easy way to preserve existing Zotero annotations (short of exporting to a file and overwriting the stored file).

This might reduce some of the need for opening PDFs externally.

mrtcode:
- Multiple thumbnails can be selected, although only with cursor.
- Makes pdf-reader iframe grayish while processing a PDF.
- Has 90, 180 and 270 degree options to rotate a PDF page. Although I would prefer only one option for that, but PDF reloading is too annoying to do that more than one time. Alternatively we could rename those to "Rotate Left", "Rotate Right", "Rotate Upside Down"

Fixes #2561


chrome/content/zotero/xpcom/pdfWorker/manager.js
```diff
@@ -410,6 +410,148 @@ class PDFWorker {
...
+					ids.push(id);
+					annotations.splice(i, 1);
+				}
+			}
+			if (ids.length) {
+				await Zotero.Items.erase(ids);
+			}
+
+			// Shift page index for other annotations
+			let notifierQueue = new Zotero.Notifier.Queue();
```
dstillman, line 476: Did you try doing this in an actual DB transaction? For a bunch of items, it should be faster.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -1383,6 +1383,13 @@ pdfReader.promptTransferFromPDF.text = Annotations stored in the PDF file will b
 pdfReader.promptTransferToPDF.title = Store Annotations in File
 pdfReader.promptTransferToPDF.text = Annotations will be transferred to the PDF file and will no longer be editable in %S.
 pdfReader.promptPasswordProtected = The operation is not supported for password-protected PDF files.
+pdfReader.promptDeletePages.title = Delete Pages
+pdfReader.promptDeletePages.text = Delete selected pages from the PDF file
+pdfReader.promptRotatePages.title = Rotate Pages
+pdfReader.promptRotatePages.text = Rotate selected pages in the PDF file
```
dstillman, line 1389: - I don't think there's any need to prompt before rotating. You're unlikely to do that by mistake, and if you do you can just rotate back. We're not losing any quality through these operations, are we?
- Add question mark for delete prompt
- A little alarming to say "selected pages" when only one page is selected, particularly for deletes — if I were a user I would be worried it was deleting more than just my one selected page. I think we'll want to pass the number of pages here and use plural forms. See `pane.tagSelector.deleteAutomatic.message` for an example.

mrtcode:
There is one thing that I still need to solve on pdf-worker. It seems that removing deleted page references without removing the whole outline section isn't that easy. Probably the same situation is with page labels. Simpler PDF editors are just deleting those sections when pages are deleted. Preview.app preserves the outline. We want to do the same.

chrome/content/zotero/xpcom/pdfWorker/manager.js
mrtcode: Ok, I changed that, but you have to take a look to make sure it's properly implemented and safe.

mrtcode:
Ok, this is ready now, except one thing. Page labels become incorrect after deleting a page. Preview.app just deletes the whole page label metadata section, while PDF Expert keeps incorrect labels. To solve that we would need to completely regenerate page labels section in PDF file, but that would be time consuming to implement (although I can try if we want).

Which path do we prefer?

chrome/locale/en-US/zotero/zotero.properties
mrtcode: > We're not losing any quality through these operations, are we?

No, we don't.

mrtcode:
> Why is it hard to fix? Is it different from a manual update of the next page label, set to update all following labels?

Page labels isn't just a list, it's more like a pattern for specific page index ranges. If I'm removing pages, I need to split the pattern into patterns with shorter ranges. And do everything to comply with PDF specification.

All right, I'll try to do that, but it might take a bit time.

mrtcode:
Ok, it should be good now. We can merge it.

dstillman:
Page label auto-detection doesn't seem to be working for me after deleting a page. I created some annotations across all pages of a PDF, manually adjusted the page number for pages 5 and later to be one greater, and then deleted page 3. It adjusted the numbers to account for the deleted page, but now if I try to set the adjusted pages back to use auto-detection, it's still using a number one greater than the page (e.g., page 4 is showing as "3").

mrtcode:
The issue is now fixed in pdf-reader.

PR #2606, opened 2022-05-09
Empty Trash: Visually remove deleted items, refresh icon

AbeJellinek:
- On trash empty, ItemTree was removing only the rows corresponding to items that _weren't_ deleted, not the ones that were. That led to a bunch of gray rows left in the trash after emptying.
- `Zotero.Items.emptyTrash` didn't send a refresh trash notification after the operation finished, so CollectionTree didn't know to refresh.
- Emptying the trash needs, but wasn't getting, a full CollectionTree refresh - it needs to fetch deleted items for each library and potentially add/remove the trash row if the `showTrashWhenEmpty` pref is false.

### DIFF
--- a/chrome/content/zotero/collectionTree.jsx
+++ b/chrome/content/zotero/collectionTree.jsx
@@ -640,11 +640,8 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 			this.tree.invalidate();
 			return;
 		}
-		if (action == 'refresh') {
-			// If trash is refreshed, we probably need to update the icon from full to empty
-			if (type == 'trash') {
-				this.tree.invalidate();
-			}
+		if (action == 'refresh' && type != 'trash') {
+			// Trash handled below
 			return;
 		}
 		if (type == 'feed' && (action == 'unreadCountUpdated' || action == 'statusChanged')) {
@@ -825,6 +822,11 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 						break;
 				}
 			}
+		}
+		else if (action == 'refresh' && type == 'trash') {
+			// We need to update the trash's status (full or empty), and if empty,
+			// the row might be removed
+			await this.reload();
 		}
 
 		this.forceUpdate();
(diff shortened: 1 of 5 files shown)


dstillman:
This doesn't seem fully fixed to me. If I move two child items from separate parent items to the trash and go empty the trash, I'm left with a bunch of gray items. I assume that's what the second commit reverted?

dstillman:
When this is fixed, we should restore the test disabled in https://github.com/zotero/zotero/commit/f368a533117758a2b2a054e61fbba30c827214de.

AbeJellinek:
Alright, all done! I have this ready to go for fx102 as well.

AbeJellinek:
(That is, the ghost items fix should be right this time and the test passes locally.)

chrome/content/zotero/collectionTree.jsx
```diff
@@ -826,6 +823,11 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 				}
 			}
 		}
+		else if (action == 'refresh' && type == 'trash') {
+			// We need to update the trash's status (full or empty), and if empty,
+			// the row might be removed
```
dstillman, line 828: Why would the row be removed? Do we ever not show the trash? (Possible we should hide it if it's empty rather than adding an extra row for every group, but that's another question…)

AbeJellinek: Yes, amazingly! I only learned that that’s an option while working on this PR. Pref is read here:

https://github.com/zotero/zotero/blob/6ca854a018e8bfe4251fbf42610276c441b5d943/chrome/content/zotero/collectionTree.jsx#L2311

dstillman: Huh! Who knew? (Certainly not me, who added that pref.)

dstillman:
Hmm, the test is still failing here. I can reproduce it locally with this:

```
./test/runtests.sh -b -f -s item
```

(This puts some previous items in the trash.)

AbeJellinek:
Huh, I couldn't when running the test locally. I'll take a look.

dstillman:
It won't happen if you run the test directly. You have to start with the `item` tests.

dstillman:
Oh, no, we don't need that at the moment. Can just add an `after()` to there that erases the annotation.

We'll add a separate test for annotations in the trash once we actually display them in the items list.

AbeJellinek:
Alright, should now be working!

dstillman:
OK, can merge for fx102 too.

PR #2611, opened 2022-05-11
Disable removing items only in descendent collections

### LINKED ISSUE #1769: Disable "Remove Item from Collection" for items only in descendent collections with recursiveCollections enabled
https://forums.zotero.org/discussion/comment/345713/#Comment_345713

Also disable the confirmation prompt when pressing the Delete key

AbeJellinek:
And use `canDeleteSelectedItems()` to enable/disable the option in the context menu.

Fixes #1769.

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -1792,9 +1792,11 @@ var ZoteroPane = new function()
 	 * Return whether every selected item can be deleted from the current
 	 * collection context (library, trash, collection, etc.).
 	 *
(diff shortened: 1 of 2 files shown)


dstillman:
https://forums.zotero.org/discussion/97172/suppression-impossible-dun-document-dune-collection

This isn't really going to solve the confusion here, though. Not sure how we want to address that. We could instead show a prompt that says the items will be removed from subcollections, but people won't read that and will remove the items from the subcollections accidentally…

dstillman:
Superseded by #2686

dstillman:
(Sorry for not rethinking this before you worked on it.)

PR #2631, opened 2022-05-23
fx-compat: Fix search dialogs

AbeJellinek:
This supersedes #1883: we don't really need React here and it was pretty straightforward to direct port the XBL to custom elements with all functionality intact.

~~It seems like the `flex` attribute doesn't have the right effect anymore on elements directly contained in a `<window>`. I'm dealing with that in `advancedSearch.xhtml` with a sort of hacky `style="width: 100%; height: 100vh;"`. If there's a better way, we should use it.~~


dstillman:
- We'll have to fix up the styling a bit now that some of the widgets have changed, but we can do that separately.
- So we think we have no ability to have native drop-downs anymore? Firefox doesn't seem to have them, either in web or chrome content, and neither does Chrome for that matter. (Only Safari.) But it's a shame to lose them.
- It looks like the -/+ buttons lost their HiDPI versions for some reason.
- I'm not totally sure how to reproduce it, but I'm sometimes seeing the condition menu not disappearing after I open the submenu. If you're not seeing, I can just try to keep an eye out of it.

AbeJellinek:
>So we think we have no ability to have native drop-downs anymore? Firefox doesn't seem to have them, either in web or chrome content, and neither does Chrome for that matter. (Only Safari.) But it's a shame to lose them.

`native="true"` on the menulist gets us a pretty OK imitation of the native dropdown widget's appearance, but still a nonnative menu. Top menu here:

image

Do we want that?

AbeJellinek:
> I'm not totally sure how to reproduce it, but I'm sometimes seeing the condition menu not disappearing after I open the submenu. If you're not seeing, I can just try to keep an eye out of it.

No, I'm not seeing that - let me know if you can find a way to reproduce.

AbeJellinek:
> It looks like the -/+ buttons lost their HiDPI versions for some reason.

Fixed with rebase.

AbeJellinek:
(Actually, the `native` attribute works for a lot of elements. We could consider using it... everywhere? If it generally looks good enough.)

dstillman:
Oh! Yes! If native is an option, let's use that. I feel like the only reason it wouldn't be the default is to allow web developers to style with CSS, but for our purposes, if it's still possible to use native widgets, we totally should.

dstillman:
(An exception would be, e.g., the item type menu, which we don't want a native menu for.)

chrome/content/zotero/elements/zoteroSearch.js
```diff
@@ -0,0 +1,989 @@
...
+			}
+		}
+
+		disconnectedCallback() {
+			// Empty the DOM. We will rebuild if reconnected.
+			while (this.lastChild) {
+				this.removeChild(this.lastChild);
+			}
+		}
+	}
```
dstillman, line 61: Do we want to move this to a generic component (`XULElementBase`?) that we extend whenever we want automatic `this.content` or `this.stylesheets` importing?

PR #2641, opened 2022-05-27
fx-compat: Fix toolbar

AbeJellinek:
Extending the CE just to override the dropmarker feels like a bit of a hack - there should be _some_ way to do it in CSS - but the `<dropmarker>` CE is quite opaque and I don't think there is.


AbeJellinek:
image

dstillman:
OK that's better!

PR #2645, opened 2022-06-01
Add Component DevHelper (fx102)

tnajdek:
This PR adds a new development tool that enables live, in-process updates:

https://user-images.githubusercontent.com/214628/171386152-1f9f8ab5-5f95-4517-ba60-0079f075ae1e.mp4

Changes to `.js`/`.jsx` (and possible other) files in `chrome/content/zotero/**` and changes to `.scss` files should be reflected almost-immediately in a component loaded using devHelper. For platform-specific css, current platform is always used.

**Changes to `.xhtml` require Zotero restart (or rebuild without omni updates) as previously.**

Running DevHelper
---

Firstly build system must be run with environmental variable `REWRITE_SRC` set, e.g.:
````bash
> REWRITE_SRC=1 npm start
````
Switching to and from `REWRITE_SRC=1` works without any additional clean steps.

Then Zotero binary must be started with a switch `ZoteroDevHelper` set to the path of the `build` folder in `zotero-client`, e.g.

````bash
> staging/Zotero.app/Contents/MacOS/zotero -debugger -ZoteroDevHelper "/path/to/zotero-client/build"
````

This switch only works if `zotero-client` was built with `REWRITE_SRC` enabled, otherwise it is ignored.

If the following steps were followed, Zotero starts with additional Component DevHelper window open. Any component loaded from the list will be refreshed whenever build system detects a change. Some components might misbehave if there is a dependency on window arguments.

Implementation details
---

With `REWRITE_SRC` flag enabled all `.xhtml` files are rewritten during the build so that source of `.css` files sourced via `<xml-stylesheet>`  tag and `.js` sourced via `<script>` tag are loaded using `file://` protocol from the actual path on disk (in the build folder) instead of `chrome://`. Additionally `Loader` is tweaked so that if `ZoteroDevHelper` value is provided any `js` dependency also uses `file://` protocol, value of `ZoteroDevHelper` is used to determine actual path of the file on disk.

The build process will use existing IPC channel no notify the client of changes. Upon receiving a notification, devHelper reloads selected component into an iframe.

To ensure that production build is unaffected, piece of code that recognizes `ZoteroDevHelper` command line switch is injected during the build only if `REWRITE_SRC` is set. Similarly `devHelper.xhtml` and `devHelper.js` will not be present in a build unless it was built with `REWRITE_SRC` flag set.

### DIFF
--- a/scripts/watch.js
+++ b/scripts/watch.js
@@ -1,19 +1,28 @@
+/* eslint-disable no-process-env */
 const path = require('path');
 const fs = require('fs-extra');
 const chokidar = require('chokidar');
 const multimatch = require('multimatch');
 const { exec } = require('child_process');
-const { dirs, jsFiles, scssFiles, ignoreMask, copyDirs, symlinkFiles } = require('./config');
+const { dirs, jsFiles, scssFiles, ignoreMask, copyDirs, symlinkFiles, rewriteSrcFiles } = require('./config');
 const { envCheckTrue, onSuccess, onError, getSignatures, writeSignatures, cleanUp, formatDirsForMatcher } = require('./utils');
 const getJS = require('./js');
 const getSass = require('./sass');
 const getCopy = require('./copy');
 const getSymlinks = require('./symlinks');
-
+const getRewriteSrc = require('./rewrite-src');
+const colors = require('colors/safe');
 
 const ROOT = path.resolve(__dirname, '..');
 const addOmniExecPath = path.join(ROOT, '..', 'zotero-standalone-build', 'scripts', 'add_omni_file');
 let shouldAddOmni = false;
+const REWRITE_SRC = envCheckTrue(process.env.REWRITE_SRC);
+const DEBUG = envCheckTrue(process.env.DEBUG);
+const ZOTERO_HTTP_SERVER_PORT = process.env.ZOTERO_HTTP_SERVER_PORT || 23119;
+
+const filteredIgnoreMask = REWRITE_SRC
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: scripts/watch.js [hunk 1]
```diff
@@ -78,6 +86,16 @@ async function addOmniFiles(relPaths) {
 	};
 }
 
+async function notifyIPC() {
+	const pipesDir = path.normalize("/Users/doppler/Zotero default-beta/pipes/");
```
dstillman, line 90: Hard-coded path. Seems like we'd need to prompt for the data directory in order to keep this, but @adomasven and I were talking about removing the default IPC pipe anyway, since its only remaining use, Firefox connector switching, is long since obsolete.

What about using the HTTP server (server.js) instead? Then it could just assume the first-started instance on 23119 is the one to use (or we could take a different port via an environment variable, I guess). See server_connector.js for examples of how to create an endpoint by adding an object within `Zotero.Server.Endpoints`. (It can be defined in another file.)

tnajdek: Oops, good spot - didn't mean to leave this in the PR, thanks! I'll experiment with HTTP approach.

tnajdek:
@dstillman thanks for suggesting using HTTP server to listen for notifications, much nicer this way. Also removed hardcoded path and few other unnecessary bits.

tnajdek:
I don’t think this is relevant anymore, and if we want some sort of “hot reload” for modern Zotero, it would have to be done differently.

PR #2649, opened 2022-06-09
fx-compat: long tag fixer

### LINKED ISSUE #1545: Option to split tags
https://forums.zotero.org/discussion/73146/how-to-split-a-tag

Seems like we could do this easily in the tag selector context menu.

tnajdek:
This PR fixes "Long Tag Fixer" dialog in fx-102 branch.

* Similar to #2368 but without React and preserves tabbed UI
* Manual trigger via "Split..." in tag selector. (#2368 had it in items box but I think tag selector is a better place for this? Resolves #1545)
* Minimal "split-only" UI when triggered manually, full UI when triggered during the sync (same as #2368


AbeJellinek:
Oops, sorry, clicked wrong button...

AbeJellinek:
But I was going to comment that we should reconcile this with #2368 - we absolutely don't need to use React here and this approach seems better, but adding a way to manually invoke the split option by right-clicking a tag would be good.

tnajdek:
Oh, I was looking through open PRs but must have missed #2368, thanks for letting me know.

dstillman:
I don't think we need to close this. Probably better to go with this minimally modified CE than React. We just want to be able to manually invoke it.

AbeJellinek:
Agreed.

tnajdek:
@dstillman @AbeJellinek I believe this is good to go.

chrome/content/zotero/containers/tagSelectorContainer.jsx
```diff
@@ -638,6 +638,49 @@ Zotero.TagSelector = class TagSelectorContainer extends React.PureComponent {
...
+
+		if (!dataOut.result) {
+			return;
+		}
+
+		const oldTagID = Zotero.Tags.getID(oldTag);
+
+		switch (dataOut.result.op) {
+			case 'split':
+				for (const library of Zotero.Libraries.getAll().filter(lib => lib.editable)) {
```
dstillman, line 664: This wouldn't be right. We'll only want to apply it to the current library. (If the same error is replicated elsewhere, you'll just need to apply it again — the tag selector is always library-specific.)

chrome/content/zotero/containers/tagSelectorContainer.jsx
```diff
@@ -638,6 +638,49 @@ Zotero.TagSelector = class TagSelectorContainer extends React.PureComponent {
...
+					const itemIDs = await Zotero.Tags.getTagItems(library.id, oldTagID);
+					await Zotero.DB.executeTransaction(async () => {
+						for (const itemID of itemIDs) {
+							const item = await Zotero.Items.getAsync(itemID);
+							for (const tag of dataOut.result.tags) {
+								item.addTag(tag);
```
dstillman, line 670: This should use the old tag type as the second parameter, so that if you split automatic tags you still end up with automatic tags.

dstillman (changes requested):
This is great!

Also need to update the "Long Tag Fixer" tests in syncRunnerTest.js:

- `xul` → `xhtml`
-  `dialog` → `window` in `waitForDialog` (just to reflect current arguments)
- `const Zotero_Long_Tag_Fixer` back to `var Zotero_Long_Tag_Fixer` in longTagFixer.js so that it's accessible as `window.Zotero_Long_Tag_Fixer` in the test

dstillman:
And actually, it looks like what we were doing before during sync re: libraries and tag types was wrong too (the former possibly because it dated to the old sync system, where libraries weren't synced separately). We should only apply to the library the error was for (`e.libraryID` should work, I think), and we should keep the tag type. Maybe the library part is more debatable here, since if you've dragged items across libraries there's a decent chance you'll just get the same popup a few second later when the other library syncs, but as written I think this would even apply the change to all editable libraries even when you specifically synced a given library. I guess we could consider applying it to all libraries being synced, but that's more complicated, so for now I think we should just do the library the tag was detected in.

tnajdek:
@dstillman thank you for a review! I've implemented the following changes:

* When splitting a tag (both manually and during sync):
    * only tag in current library is affected (previously: all editable libraries)
    * tag type is preserved
* Updated tests in `syncRunnerTest.js`
* Simplify calls to `createElementNS` (to match changes elsewhere in as f81b4b07)


Some caveats:
* When manually editing (not splitting) a tag during sync, new tag has type of 0, this seems to be hard-coded in `replaceTag` in `item.js`.
* Not sure if this is the most efficient way to check tag type (I'm getting all automatic tags in a library, then checking if a tag id appears on the list).

tnajdek:
Rebased onto latest `fx102`, should be good to go again.

chrome/content/zotero/xpcom/sync/syncRunner.js
```diff
@@ -1214,62 +1214,55 @@ Zotero.Sync.Runner_Module = function (options = {}) {
...
-												for (let itemID of itemIDs) {
-													let item = await Zotero.Items.getAsync(itemID);
-													for (let tag of dataOut.result.tags) {
-														item.addTag(tag);
-													}
-													item.removeTag(oldTag);
-													await item.save();
+										yield Zotero.DB.executeTransaction(async function () {
+											for (let itemID of itemIDs) {
+												let item = await Zotero.Items.getAsync(itemID);
```
dstillman, line 1242: I think you should be able to do `let tagType = item.getTagType(oldTag)` here to get the tag type of the old tag on this item, instead of the `automaticTagIDs.includes()` thing above.

PR #2654, opened 2022-06-14
fx-compat: Fix context menus not appearing in the tag selector 

tnajdek:


dstillman:
Are tag selector menus working for you? They're not for me, before or after this change.

tnajdek:
Yes, both the individual tag context menu and tag selector "actions" drop-down menu work after tweaks from this PR. I've re-tested using current `fx102`, still seems to work:

Screenshot 2022-06-21 at 11 05 04

dstillman:
Oh, weird, it's working now. Maybe was something corrupted in the build dir.

PR #2659, opened 2022-06-16
fx-compat: Add preferences with full-text/keyword search

AbeJellinek:
It works! We have full-text and keyword search with support for static keywords and .dtd/.properties keys, a Firefox-like left-sidebar layout, and instant-apply on all platforms.

Things to discuss:
- Do we want to isolate each pane inside its own shadow DOM? That would solve ID collision issues (not a problem so far but potentially painful when add-ons add new panes) but necessitate a big reworking of virtualized-table's interactions with the DOM - it's totally broken inside a shadow DOM right now. I have a draft where everything else is working, though.
- How should we load add-on panes?

Breaking changes (partial list, since lots of general Fx102 compat changes are required in pane XULs):
- The semantics of the `syncfrompreference` and `synctopreference` events have changed - event handlers can no longer return a value(!) to change what gets synced.
- Pane XUL files (`preferences_general.xhtml` and friends) are now loaded as fragments and spliced into the tree, not as their own full documents. They don't need `<?xml`s, `<!DOCTYPE`s, or `<?xml-stylesheet`s. The list of DTDs and CSS files to load has to be specified in JS.
- Inline `oncommand` handlers still work, as do handlers for standard DOM events, but handlers for custom events have to be in JS.

Issues:
- CSS needs work.
- Style Manager table sometimes only shows a few rows on initial load.
- `charsetMenu` depended on a Mozilla library that we don't get anymore, so its strings are hardcoded for now.
- Probably more


dstillman:
Incredible!

Haven't looked at the code yet, but some quick notes from just trying it out:

- We'll want to get rid of all the tabs and split those out into sections, with the former tab labels as headings.
- Hidden content is still matching the search. E.g., Data Syncing shows up for a "libraries" search because of the libraries-to-sync button, even though that's hidden when logged out.
- The Sync → Reset pane should be moved into some sort of secondary pane that's hidden by default. It looks like Mozilla has two approaches: in-pane navigation with a back button ("Settings…" for container tabs) and modal dialogs ("Advanced…" for Fonts). We're already using modals for libraries-to-sync and site-specific Quick Copy. If feasible, in-pane probably makes sense for Sync → Reset, and probably other things.
- For in-pane navigation, if the contents match a search, it should just highlight the button, rather than showing the content inline with other search results. If the button is hidden (which would be the case for "Reset Sync…" when logged out), the contents shouldn't match the search.

AbeJellinek:
OK, pushed all but in-pane Reset flow.

Now is probably the time to rename some sections so they make sense in this new context, where sections from different panes are shown side by side. Firefox names the first section under each pane after the pane itself, then names additional sections something unique. Following that approach:

- General pane
	- Add "General" heading at the top
	- "Miscellaneous" -> "Item Handling" or "Items"?
- Sync pane
	- "Settings" -> "Sync"
- Search pane
	- Add "Search" heading at the top
- Export pane
	- Add "Export" heading at the top
- Cite pane
	- "Styles" -> "Cite"
- Advanced pane
	- "General" -> "Advanced"
	- "Miscellaneous" -> "Translators", move "Language:" up to be directly under "Advanced"

How does that sound?

dstillman:
I think we'll want to rethink things more broadly, but we can do most of that separately. Some quick things:

- Language can be moved from Advanced to a section in General
- Search shouldn't be a tab — Full-Text Cache should just be in Advanced (and really I'd like to remove or hide that even more, but we can leave it there for the moment)
- OpenURL can be moved to General as "Locate". The default WorldCat resolver is (as far as we know) no longer updatable by institutions, and we now automatically populate the list from our [OpenURL Resolvers page](https://www.zotero.org/support/locate/openurl_resolvers), which institutions have been contributing too, so it makes sense as a more prominent pref now. (I'll add some additional text there.)

I'll futz with things more later.

dstillman:
I resolved a couple conflicts from 789a8a5fe9b648e4712a375deccbb8f222b62d74 and 6297d0862fce79d04e976fb6fe88ccf38c43ff8d.

AbeJellinek:
OK, all moves done.

AbeJellinek:
(Thanks for doing the merge.)

chrome/content/zotero/charsetMenu.js
```diff
@@ -23,6 +23,119 @@
...
+		{
+			"label": "Unicode",
+			"value": "UTF-8"
+		},
+		{
+			"label": "Western",
+			"value": "windows-1252"
+		}
+	],
+	"otherCharsets": [
```
dstillman, line 39: I think we can just remove the Import Character Encoding setting, the `extensions.zotero.import.charset` pref, and `otherCharsets` here. To the extent that this made sense 15 years ago, which is debatable, it's almost certainly pointless now. I'm sure approximately no one ever changes this setting, and in the extremely unlikely event that it's necessary, someone could convert the file with a text editor. We could also probably auto-detect a charset using the new JSWindowActor I added for that if our initial test reveals it's not UTF-8.

For exports, we're already only displaying UTF-8, UTF-8 w/o BOM, and Western.

For this PR, let's take this stuff out, and I'll fix up the import stuff in translate_firefox.js separately.

dstillman (changes requested):
- The selected tab should persist between loads. (Firefox doesn't do this, but apps that use separate windows generally do, and I'm not sure Firefox shouldn't either.)
- Back button for in-pane navigation can just be a `<`, as in Firefox, instead of "Back"
- Links in the Sync tab don't work
- When I first loaded the prefs, the feed-sorting setting was blank for some reason, even though the pref was set to the default of `false`. After changing it, and then resetting the pref, I can't reproduce this. Can ignore if you're not seeing, but just flagging in case it's something obvious.

AbeJellinek:
>Back button for in-pane navigation can just be a <, as in Firefox, instead of "Back"

I tried that before! If you make a native button tall enough to fit a nice-looking back arrow icon, it falls back to the old-style beveled button appearance and looks pretty bad. The Firefox prefs just use the same non-native button style that they use everywhere. We could give it a custom appearance - just a simple gray rounded rectangle that darkens when you hover/click?

dstillman:
(Rebased without temp preferences menu option commits, which caused conflicts during interactive rebase)

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -2063,3 +2063,79 @@ Zotero.JSON = new function() {
...
+	 * is loaded as a fragment, not a full document, with XUL as the default
+	 * namespace and (X)HTML tags available under `html:`.
+	 *
+	 * @param {Object} options
+	 * @param {String} options.id Must be unique
+	 * @param {String} [options.parent] ID of parent pane (if provided, pane is hidden from the sidebar)
+	 * @param {String} [options.label] A DTD/.properties key (optional for panes with parents)
+	 * @param {String} [options.rawLabel] A raw string to use as the label if options.label is not provided
+	 * @param {String} [options.image] URI of an icon (displayed in the navigation sidebar)
+	 * @param {String} options.src URI of an XHTML fragment
```
dstillman, line 2129: Why XUL? Just for the automatic .dtd parsing? Most elements we're using XUL versions of are also available as custom elements in HTML, and, e.g., Mozilla is using HTML for prefs. It doesn't really matter for our panes, but I'm concerned that we don't want to ask plugins to pass XUL here when we're definitely going to want to switch to HTML later. Seems like we could just do the .dtd entity parsing ourselves on the HTML (if there's nothing for that from Mozilla), or even just make plugins use Fluent while they're rewriting things anyway.

chrome/content/zotero/preferences/preferences.js
```diff
@@ -43,61 +43,11 @@ var Zotero_Preferences = {
...
-			src: 'chrome://zotero/content/preferences/preferences_sync_reset.xhtml',
-			onLoad() {
-				Zotero_Preferences.Sync.initResetPane();
-			}
-		});
-		this._autoSeparator = true;
+		Zotero.PreferencePanes.builtInPanes.forEach(pane => this.addPane(pane));
+		if (Zotero.PreferencePanes.pluginPanes.length) {
+			this.navigation.append(document.createElement('hr'));
+			Zotero.PreferencePanes.pluginPanes.forEach(pane => this.addPane(pane));
```
dstillman, line 49: We should sort plugin tabs alphabetically by label.

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -2063,3 +2063,79 @@ Zotero.JSON = new function() {
 		return JSON.parse(arg);
 	}
 }
+
+/**
+ * Manages preference panes.
+ */
+Zotero.PreferencePanes = {
```
dstillman, line 2070: This can be in a separate file, added to zotero-service.js. All files are in a JAR anyway, so better to keep them separate for code manageability.

chrome/content/zotero/preferences/preferences_sync_reset.xhtml
```diff
@@ -0,0 +1,40 @@
+<vbox class="main-section" id="sync-reset" onload="Zotero_Preferences.Sync.initResetPane()">
+	<html:h1>&zotero.preferences.sync.reset;</html:h1>
+
+	<!-- This doesn't wrap without an explicit width, for some reason -->
+	<description id="reset-sync-warning" width="45em">&zotero.preferences.sync.reset.warning1;<label style="margin-left: 0; margin-right: 0" class="zotero-text-link" href="http://zotero.org/support/kb/sync_reset_options">&zotero.preferences.sync.reset.warning2;</label>&zotero.preferences.sync.reset.warning3;</description>
```
dstillman, line 5: This text link doesn't work.

chrome/content/zotero/xpcom/zotero.js
dstillman: (And if we want to keep ours as default-XUL for now so that we can merge this, we could have a duplicate adding function that uses default-XHTML for panes coming from `register()`.)

chrome/content/zotero/xpcom/zotero.js
```diff
@@ -2063,3 +2063,79 @@ Zotero.JSON = new function() {
...
+	 * Register a pane to be displayed in the preferences. The pane XHTML (`src`)
+	 * is loaded as a fragment, not a full document, with XUL as the default
+	 * namespace and (X)HTML tags available under `html:`.
+	 *
+	 * @param {Object} options
+	 * @param {String} options.id Must be unique
+	 * @param {String} [options.parent] ID of parent pane (if provided, pane is hidden from the sidebar)
+	 * @param {String} [options.label] A DTD/.properties key (optional for panes with parents)
+	 * @param {String} [options.rawLabel] A raw string to use as the label if options.label is not provided
+	 * @param {String} [options.image] URI of an icon (displayed in the navigation sidebar)
```
dstillman, line 2128: Are we not doing anything for 2x images? Even for our panes? I suspect we never had 2x icons for those, and so this has no effect for our icons at the moment, but we'll definitely want to support multiple icon sizes for plugins.

For plugins we could theoretically skip both this by just having plugins pass their id, and we could get the images from the [`icons`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/icons) property in manifest.json by adding a function to plugins.js. We could also automatically use the name as the label but let plugins override with `label` in case the extension name is too long (e.g., "Zutilo" instead of "Zutilo Utility for Zotero", though honestly that's probably a holdover from Zotero for Firefox days and should just be changed).

AbeJellinek: >Are we not doing anything for 2x images? Even for our panes?

We're not, and we never did have 2x icons for those. Latest commit should do the trick (but I don't have a plugin to test with yet).

AbeJellinek: I agree that using the plugin name as label should be fine in almost all cases, but I want to leave things more open for a hypothetical complex plugin - a Jurism rework? - that might need to register multiple panes, that sort of thing.

chrome/content/zotero/xpcom/plugins.js
```diff
@@ -159,6 +159,23 @@ Zotero.Plugins = new function () {
...
+
+
+	/**
+	 * @param {String} id
+	 * @param {Number} idealSize In logical pixels (scaled automatically on hiDPI displays)
+	 * @returns {Promise<String | null>}
+	 */
+	this.getIconURI = async function (id, idealSize) {
+		var addon = await AddonManager.getAddonByID(id);
+		return AddonManager.getPreferredIconURL(addon, idealSize * (Zotero.hiDPI ? 2 : 1));
```
dstillman, line 175: I agree this is good enough for our purposes, but if we cared we could probably pass `Services.appShell.hiddenDOMWindow`, which has `devicePixelRatio`.

chrome/content/zotero/xpcom/preferencePanes.js
```diff
@@ -0,0 +1,131 @@
...
+	]),
+
+	pluginPanes: [],
+
+	/**
+	 * Register a pane to be displayed in the preferences. The pane XHTML (`src`)
+	 * is loaded as a fragment, not a full document, with XUL as the default
+	 * namespace and (X)HTML tags available under `html:`.
+	 *
+	 * Plugins must call #unregister(id) in their shutdown() method.
```
dstillman, line 88: Since we're requiring a `pluginID`, any reason we shouldn't just unregister the pane ourselves (and say here that the pane will be automatically unregistered if the plugin is disabled)? We can unregister from the `onUninstalled()` handler in `Zotero.Plugins`.

chrome/content/zotero/xpcom/preferencePanes.js
```diff
@@ -0,0 +1,131 @@
...
+		this.pluginPanes.push(addPaneOptions);
+	},
+
+	/**
+	 * Unregister a pane due to plugin shutdown.
+	 *
+	 * @param {String} id
+	 */
+	unregister: function (id) {
+		this.pluginPanes = this.pluginPanes.filter(p => p.id !== id);
```
dstillman, line 129: Can we remove the pane (or just refresh the window) if the prefs are open?

chrome/content/zotero/xpcom/plugins.js
AbeJellinek: Ah, good to know, thanks.

chrome/content/zotero/xpcom/preferencePanes.js
AbeJellinek: So that we avoid a circular dependency between `Zotero.Plugins` and `Zotero.PreferencePanes`, how about adding support for lifecycle observers to `Zotero.Plugins`? We could either allow observers to implement any of the methods on `_addonObserver`, or, if we'd like to avoid exposing the Mozilla-specific API, just `onInstall`, `onStartup`, `onShutdown`, and `onUninstall` (like the plugins themselves get).

dstillman: I dunno — what's the specific concern?

If anything I think we'd want to expose the same events the plugins get, but I don't think it's necessary for this vs. a one-line call to `Zotero.PreferencePanes.unregister()` in `onUninstalled()`. I think the main reason to do it would be if we think plugins want the ability to monitor for events from other plugins, which is not impossible — there might be some inter-plugin customizations already. They could just use `AddonManager.addAddonListener()` directly but providing our own listeners would be better. I think we could ignore this for now, but if you want to do that, that's fine.

AbeJellinek: `unregister()` wants a pane ID, not a plugin ID, so we'd need to expose a method to unregister by plugin ID, and that doesn't seem useful beyond being called from `Zotero.Plugins`. The interoperability case is also worth supporting. So why not, I guess. Let me know if what I pushed looks OK.

chrome/content/zotero/xpcom/plugins.js
```diff
@@ -135,7 +136,18 @@ Zotero.Plugins = new function () {
 				Zotero.logError(`Error running bootstrap method '${method}' on ${id}`);
 				Zotero.logError(e);
 			}
-			
+
+			for (let observer of observers) {
+				if (observer[method]) {
+					try {
+						observer[method](params, reason);
```
dstillman, line 143: These could actually return promises, so should [return `maybePromise` and `await`](https://github.com/zotero/zotero/blob/9b7b4fe134abd1431ad446c6b3649d1eb3686e91/chrome/content/zotero/xpcom/sync/syncRunner.js#L1325-L1328) so that errors get logged. (I'm actually not handling this above either, but I can fix that separately.)

chrome/content/zotero/xpcom/plugins.js
```diff
@@ -140,7 +140,10 @@ Zotero.Plugins = new function () {
 			for (let observer of observers) {
 				if (observer[method]) {
 					try {
-						observer[method](params, reason);
+						let maybePromise = observer[method](params, reason);
+						if (maybePromise.then) {
```
dstillman, line 144: Needs to be `maybePromise && maybePromise.then` — otherwise a non-async callback without a return value will throw.

AbeJellinek: Right, thanks.

PR #2660, opened 2022-06-16
fx-compat: Fix twisty color inverting on Mac

AbeJellinek:
Since we're now using a much milder background color for table rows, inverting the twisty's color looks wrong. No available highlight color option clashes with the darker twisty color.

Before:
image

After:
image

### DIFF
--- a/scss/mac/_virtualized-table.scss
+++ b/scss/mac/_virtualized-table.scss
@@ -8,10 +8,6 @@
 		}
 	}
 
-	.selected:not(.highlighted) .twisty svg {
-		fill: #fff;
-	}
-
 	.spacer-twisty {
 		min-width: 19px;
 	}


dstillman:
I'm not sure why we actually have a lighter highlight color. I don't think it was intentional?

Here's what I have Firefox 102:

Screen Shot 2022-06-18 at 3 21 56 AM

Here's what I have in Finder:

Screen Shot 2022-06-18 at 3 22 27 AM

Here's what I have in Mozotero 60:

Screen Shot 2022-06-18 at 3 26 15 AM

And here's what I have in Mozotero 102:

Screen Shot 2022-06-18 at 3 23 08 AM

So I think this is just a regression on the fx102 branch.

AbeJellinek:
I honestly liked the lighter highlight color better! But yeah, it was using the text highlight background instead of the cell highlight background. Fixed in d2dbf959b674b5bd6467ee485e6582113c2db4f1.

PR #2671, opened 2022-06-24
Implement smart notes switching in contextPane

### LINKED ISSUE #2650: Smart switching for context pane note and notes list: stick a child note to the current tab, but keep a standalone note global
- Most people are either using only child notes or only standalone notes.
- As long as you have a child note open in a tab, it will remain when you switch back to it.
- If you're on the notes list or a standalone note, the state persists between tabs.
- Switching to a child note wouldn't affect the global state. But if you stayed in the notes list or opened another standalone note, that would affect the global state.
- How about child notes not from the current parent?

https://twitter.com/katrinagroeller/status/1527373366117576709
https://forums.zotero.org/discussion/97730/can-opened-note-on-the-notes-sidebar-follow-the-opened-item-in-the-pdf-reader

mrtcode:
Implemented as discussed. We can try this out.

Fixes #2650

### DIFF
--- a/chrome/content/zotero/bindings/noteeditor.xml
+++ b/chrome/content/zotero/bindings/noteeditor.xml
@@ -353,15 +353,16 @@
 				<body>
 					<![CDATA[
 						(async () => {
-							let n = 0;
-							while (!this._editorInstance && n++ < 100) {
-								await Zotero.Promise.delay(10);
-							}
-							await this._editorInstance._initPromise;
-							this._iframe.focus();
 							try {
+								let n = 0;
+								while (!this._editorInstance && n++ < 100) {
+									await Zotero.Promise.delay(10);
+								}
+								await this._editorInstance._initPromise;
+								this._iframe.focus();
 								this._editorInstance._iframeWindow.document.querySelector('.toolbar-button-return').focus();
-							} catch(e) {
+							}
+							catch (e) {
 							}
 						})();
 					]]>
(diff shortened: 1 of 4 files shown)


dstillman:
This seems…basically fine? Like obviously I know what it's doing but I can predict what I'm going to see when I switch tabs, and I think it's pretty logical.

There's one thing I'm finding a little surprising:

1. In Tab A, open a child note.
2. In Tab B, open a standalone note.
3. Switch back to Tab A and see the child note.
4. Click Back to get to the notes list.
5. Switch to Tab B.

This shows the notes list in Tab B, because switching tabs from the notes list changed the global state, but since Tab A wasn't using the global state to begin with, I expected that to not affect Tab B.

Maybe if you're in a child note — and therefore not using global state — then you can't affect global state without actually *selecting* a global-state-affecting note? Clicking back to the notes list would clear the child item state, but that's it. So then the behavior would be:

1. In Tab A, open a child note.
2. In Tab B, open a standalone note.
3. Switch back to Tab A and see the child note.
4. Click Back to get to the notes list.
5. Switch to Tab B and see the standalone note.
6. Switch back to Tab A and see the standalone note. Since you cleared the child item state in Step 4, this now uses global state.
7. Click back to the notes list.
8. Switch to Tab B and see the notes list. Since you switched to the notes list from a global-state note in Step 7, clicking Back affected global state.

mrtcode:
Ok, this specific case is changed now.

dstillman:
OK! Let's see how it goes.

PR #2675, opened 2022-06-28
fx-compat: More progress on platform menus

AbeJellinek:
Not able to test on non-Mac yet, but this behavior should be right. Removes all the #ifdefs and moves the behavior into platformKeys.js.

### DIFF
--- a/chrome/content/zotero/zoteroPane.xhtml
+++ b/chrome/content/zotero/zoteroPane.xhtml
@@ -173,7 +173,29 @@
 					the children can be modified. All non-menu nodes are ignored
 				-->
 				<menubar id="main-menubar"
-				style="border:0px;padding:0px;margin:0px;-moz-appearance:none">
+				style="border:0px;padding:0px;margin:0px;-moz-appearance:none">		
+					<!-- nsMenuBarX hides these and uses them to build the Application menu. -->
+					<menupopup id="mac_application_menu">
+						<menuitem id="menu_preferences"
+								label="&preferencesCmdMac.label;"
+								key="key_preferencesCmdMac"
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/zoteroPane.xhtml
```diff
@@ -285,15 +303,9 @@
...
-							#ifndef XP_MACOSX
-							<menuseparator/>
-							<menuitem id="menu_preferences"
-								oncommand="Zotero.Utilities.Internal.openPreferences()" data-l10n-id="menu-preferences"/>
-							#endif
-							#endif
-							-->
+							<menuseparator id="menu_EditPreferencesSeparator"/>
+							<menuitem id="menu_EditPreferencesItem"
+								oncommand="Zotero.Utilities.Internal.openPreferences()" data-l10n-id="menu-settings"/>
```
dstillman, line 308: Ugh. Seems like Firefox, Chrome, and Edge all call it this now, but this is going to make cross-platform support annoying, both in documentation and the forums.

AbeJellinek: Especially since we call the Connector Preferences that regardless of platform.

I was just fixing the localization key, since `menu-preferences` doesn’t seem to exist. No real reason we can’t stick with what we had. I’d imagine there’s other software that uses “Preferences” on Windows, and it’s what Zotero users are used to.

dstillman: Oh, wait, Ventura actually changes it to "Settings…" in the application menu too, along with the changes from "System Preferences" to "System Settings". So a little rough until everyone upgrades, but this is fine.

Issue #2349, opened 2022-02-15
Add "Include Zotero Links" option to HTML note export


dstillman:
https://forums.zotero.org/discussion/comment/399237/#Comment_399237
https://forums.zotero.org/discussion/96293/quick-copy-rich-text-with-link-for-notes

(Later, these should all be Quick Copy options.)

dstillman:
https://forums.zotero.org/discussion/comment/408226/#Comment_408226

Probably important for Scrivener, so we should do this as an option for manual export — unchecked by default? I don't think we want to do this by default for Quick Copy, so need to wait until we have options for that.

dstillman:
https://forums.zotero.org/discussion/98002/zotfile-zotero-version-annotations-is-crippled

A lot of people seem to want links in rich-text export, so we should do this now. (The new prefs in zfx102 actually use the same files, so this might not even trigger a conflict.)

PR #2676, opened 2022-06-29
Add "Include Zotero Links" option for note export and QuickCopy

mrtcode:
For now this only implements #2349 (configurable `includeAppLinks` option for note export and QuickCopy for Markdown and HTML translators). But lies the ground for other discussed options.

Uses `markdownOptions` and `htmlOptions` (for Markdown + Rich Text translator) or `options` (for HTML translator) in `export.noteQuickCopy.setting`, to store `includeAppLinks` which is later passed to a corresponding translator over `displayOptions`.

Needs zotero/translators#2844.

We can do the other discussed options separately, because:
- If we're adding anything else to the Quick Copy prefs dialog, we really have to move Site-Specific Settings somewhere else. But maybe we should do that after fx102?
- Custom citation style is complicated enough to be a separate commit.


dstillman (commented):
(Sorry, I thought I responded to this months ago.)

This isn't properly setting the "Include App Links" checkbox to `true` for Markdown if the `export.noteQuickCopy.setting` pref was set to something other than the default value in an earlier version.

mrtcode:
> This isn't properly setting the "Include App Links" checkbox to `true` for Markdown if the `export.noteQuickCopy.setting` pref was set to something other than the default value in an earlier version.

It doesn't set the value until you select the translator in the selection box right? And the selection box is either empty or showing "Loading…"? I guess we should just hide the checkboxes in this case? But what we should show in the selection box  if `export.noteQuickCopy.setting` has an unexpected value?

mrtcode:
> Drop-down wasn't empty — it showed Markdown + Rich Text. The checkbox just wasn't set properly to match the default for that mode, in absence of that value in the stored setting.

Oh, you mean "Include Zotero Links" for Markdown should be enabled by default if you update from an older Zotero version.

Could we modify `export.noteQuickCopy.setting` and add the missing default `"markdownOptions":{"includeAppLinks":true},"htmlOptions":{"includeAppLinks":false}` options on update? Or, alternatively, if it's `Markdown + Rich Text` translator just reset the pref? Because otherwise I would need to check in various places if `markdownOptions` is missing and hard code the default value everywhere.

dstillman:
OK, that should fix it.

dstillman:
New layout:

Screen Shot 2022-10-21 at 5 20 57 AM

Issue #16, opened 2011-11-17
Default name of untitled collection reverts  back to Untitled 2 after Untitled 10


mronkko:
How to replicate:
1) Create a new collection in the root of My Library with the default name ("Untitled")
2) Repeat 1 nine times ("Untitled 2"-"Untitled 10")
1) Create a new collection

What happens:
The default name for the new collection is "Untitled 2"

What I would expect to happen:

The default name for the new collection is "Untitled 11"

PR #2702, opened 2022-07-14
Convert language to ISO, display localized

### LINKED ISSUE #2699: Map localized language names to ISO codes for CSL output
https://forums.zotero.org/discussion/comment/412045/#Comment_412045

AbeJellinek:
Closes #2699. Needs zotero/utilities#16.

### DIFF
--- a/chrome/content/zotero/xpcom/data/item.js
+++ b/chrome/content/zotero/xpcom/data/item.js
@@ -733,10 +733,6 @@ Zotero.Item.prototype.setField = function(field, value, loadIn) {
 		return true;
 	}
 
-	if (!loadIn && field == 'language') {
-		value = Zotero.Utilities.Item.languageToISO6391(value);
-	}
-	
 	// Normalize values
 	if (typeof value == 'number') {
 		value = "" + value;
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/xpcom/data/item.js
```diff
@@ -732,6 +732,10 @@ Zotero.Item.prototype.setField = function(field, value, loadIn) {
 		}
 		return true;
 	}
+
+	if (!loadIn && field == 'language') {
+		value = Zotero.Utilities.Item.languageToISO6391(value);
+	}
```
dstillman, line 738: This should go down below, within the existing `if (!loadIn) {` block on line 805. (This way, I don't think a non-trimmed value would even work.)

dstillman (changes requested):
If I'm reading https://github.com/zotero/utilities/pull/16 right, this won't do quite what we want. We want to accept 1) language names in the current locale and 2) language names in the language itself. As written, this will only accept English names and native names.

If you're using a German locale, you should be able to type either "französisch" or "français" to get "fr".

AbeJellinek:
Right, makes. But I think we’ll still want to support English names, won’t we? Since most translators, imports, and the like will give us English language names regardless of the user’s locale.

dstillman:
Yeah, that makes sense.

AbeJellinek:
OK, should be good now with [`730ed72` (#2702)](https://github.com/zotero/zotero/pull/2702/commits/730ed72a72eb775b7e98c5be34bdddc64b789180) and [zotero/utilities@`ef02147` (#16)](https://github.com/zotero/utilities/pull/16/commits/ef021472a71291b0c8639cd7abcd3b5ad6757104).

PR #2705, opened 2022-07-15
Replace Locate selection limit with confirmation dialog

### LINKED ISSUE #2704: Make item limit for view options configurable
https://forums.zotero.org/discussion/98413/is-it-possible-to-open-the-web-url-for-multiple-items-at-once

I guess why not?

AbeJellinek:
Also corrects an inconsistency between Enter key behavior, which had a max of 19, and Locate menu behavior, which had a max of 20.

Closes #2704.


dstillman:
https://forums.zotero.org/discussion/comment/412559/#Comment_412559

Just prompting above 20 actually kind of seems better. We should probably just do that. (Sorry, I was too scared to actually try this in a browser to see what they do…)

AbeJellinek:
OK, replaced with a dialog. Let me know how that looks.

PR #2728, opened 2022-07-27
Merge: Preserve embedded annotations

### LINKED ISSUE #2725: Duplicate merging: Don't merge attachments with differing embedded annotations
If you have duplicate items, and one has a pristine PDF and the other has PDF with embedded annotations, I think we're not currently taking that into account, and would happily move one of them to the trash. If we're going to keep the near-identical merging, we should probably avoid this.

This would probably require triggering pdf-worker on the attachments to check for annotations, and then performing some sort of equality check, or just skipping merging if there are any annotations. Or maybe there's some simpler way to just scan for annotations, if we don't care about more sophisticated checks. @mrtcode might be able to suggest something here.

AbeJellinek:
The PDF with embedded annotations is kept and the other is merged into it (including any Zotero annotations, following the existing behavior). If both have embedded annotations, both are kept.

Fixes #2725.


chrome/content/zotero/xpcom/data/item.js
```diff
@@ -3955,6 +3955,28 @@ Zotero.Item.prototype.getAnnotations = function (includeTrashed) {
...
+		return false;
+	}
+
+	let path = await this.getFilePathAsync();
+	if (!path) {
+		return false;
+	}
+
+	let contents = await Zotero.File.getContentsAsync(path);
+	let re = /\/Type\s+\/Annot\s/;
```
mrtcode, line 3975: I'm concerned about link annotations that are quite common in PDFs. Maybe we should test for specific annotation types instead. Or subtract link annotation count from the total annotations count.

AbeJellinek: OK, see latest commit. A slightly more complicated parsing routine but should still be pretty fast.

mrtcode: Wait, but why not just test if `Highlight`, `Underline` and other markup annotation subtypes just exist in the file, with one regex line?

AbeJellinek: Oh, that's a great question. Don't know why I didn't think of that. Can those ever be subtypes of another object type? I guess probably not.

AbeJellinek: OK, fixed. Thanks so much for pointing that out.

mrtcode: Looks good!

PR #2745, opened 2022-08-03
fx-compat: Style Editor: Validate CSL inline

AbeJellinek:
Lines that fail validation are highlighted, and hovering shows you the error:

image

Errors are also marked in red under the scroll bar so you can quickly scroll to the offending line.

### DIFF
--- a/chrome/content/zotero/tools/csledit.js
+++ b/chrome/content/zotero/tools/csledit.js
@@ -147,6 +147,10 @@ var Zotero_CSL_Editor = new function() {
 	
 	this.onStyleModified = function () {
 		let xml = editor.getValue();
+		Zotero.Styles.validate(xml).then(
+			() => this.updateMarkers(''),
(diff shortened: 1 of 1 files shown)


denismaier:
Is that a new style editor? Don't think that's still available, right?

denismaier:
Sorry. Yet available

denismaier:
I mean: that looks different than the editor in the current release version.

dstillman:
Are there any settings for this? Is it possible to indicate errors in the gutter? The red underlines are a little hard to spot.

The "View Problem" line/mode also seems a little pointless — seems like that's just showing the same error that's already shown in the popup? — but I guess that's what gets you into the mode with the arrows for moving through the errors.

dstillman:
> Errors are also marked in red under the scroll bar so you can quickly scroll to the offending line.

Oh, I missed this. I guess that works, if you notice it.

PR #2756, opened 2022-08-12
fx-compat: Fix <tab> keyboard nav, move CE imports to single file

AbeJellinek:
Having a single `customElements.js` file that we import everywhere we need it helps with organization, and it gives us a single place to put things like the `<tab>` fix.

We could switch to using `setElementCreationCallback()` like Firefox if the number of imports gets out of hand, but the overhead right now should be small.


dstillman:
Rebased

PR #2765, opened 2022-08-18
fix Chinese locale

volatile-static:
Update zotero.properties


zepinglee:
The Zotero team prefers to use Transifex (<https://www.transifex.com/zotero/zotero>) to track translations rather than GitHub PR.

I've fixed this in Transifex.

Screen Shot 2022-08-18 at 17 15 14

dstillman:
Thanks!

PR #2767, opened 2022-08-18
Create a note from annotations from multiple PDF attachments and parent items

mrtcode:


dstillman:
OK, added the UI part of this — separate "Add Note from Annotations" and "Create Note from Annotations" depending on what's selected.

There are some tricky situations, though.

If someone selects across two top-level items but only one has an attachment with annotations, we still have to show "Create Note from Annotations" in some cases because we don't know if an unprocessed attachment has external annotations. So in that case you'd end up with a standalone note for a single item, which currently doesn't include the parent item title. I guess that makes sense.

But what if someone selects, say, a regular item with a child attachment containing annotations and a child note from underneath another item? In that case we know ahead of time that there can only possibly be annotations from under the one attachment. Do we still show "Create Note from Annotations" in that case, or do we ignore the note, show "Add Note from Annotations", and add the note under the first parent item?

mrtcode:
It seems there isn't any option to extract annotations if only one top-level attachment is selected.

dstillman:
OK, fixed standalone attachments.

So we're using h2 for item titles and h3 for attachment titles regardless of whether they both exist in the given note, so for child notes with multiple attachments it will jump from h1 to h3. But maybe that's better because it lets these notes be combined without needing to adjust the headers?

PR #2784, opened 2022-08-24
fx-compat: Implement Scaffold functions, SingleFile, PDF redirects, and style installation

### LINKED ISSUE #3049: fx-compat: Clicking on a style in the style installer triggers download dialog

AbeJellinek:
Needs zotero/translate#15, fixes #3049


dstillman:
So I think the question here is what the actual requirements for a hidden-browser-using function are. We already have the ability to fetch the HTML coming over the wire and produce a DOMParser-ed Document with `HTTP.processDocuments()`, and if I recall we can even a cookie sandbox and meta redirects. So just serializing the HTML from the child process doesn't necessarily get us that much else. As written, I assume we'd get processing of inline JS, including synchronous JS redirects, which isn't nothing, but it probably wouldn't wait for other JS scripts to load and execute. I don't think we were explicitly doing that before — I'm not sure there's really a way to do it reliably — but it's possible we were getting that somewhat by accident with the 1-second `setTimeout` calls. And I assume `defer: true` in tests wouldn't actually do anything here, because I'm guessing the delay is on the non-changing serialized-and-reparsed Document, not the original document where JS might still be running.

I suspect the cookie part (`cookieSandbox.attachToBrowser`) also isn't working. [Edit: OK, looks like you saw that.]

So maybe it's enough to add a configurable delay for HiddenBrowser.jsm, but we should probably look into getting SingleFile working (maybe with @adomasven's help) and confirm that this is actually doing what we need. But there might be things where we need to actually add other JS actors to run other actions on the page.

(Also, maybe we should take this opportunity to deprecate the names `processDocuments` and `loadDocuments` and switch to function names that convey the difference.)

AbeJellinek:
> And I assume defer: true in tests wouldn't actually do anything here, because I'm guessing the delay is on the non-changing serialized-and-reparsed Document, not the original document where JS might still be running and updating the page.

Right...

> So maybe it's enough to add a configurable delay for HiddenBrowser.jsm

So maybe we should cut out `loadDocuments` and its hard-to-work-with synchronous callback model entirely and just add more utilities to `HiddenBrowser.jsm`? An option to require a successful status code and an option to add a delay, at least to start with.

dstillman:
Yeah, I think that's a good call. I think we moved the old `processDocuments` to `loadDocuments` when we switched to XHR `processDocuments` just to avoid rewriting old code, but most/all uses of `loadDocuments` at this point are for single URLs, so just using `HiddenBrowser.jsm` directly, with delay and cookie-sandbox options, makes sense.

SingleFile we might need to run as a JSWindowActor.

AbeJellinek:
OK, latest commit should get us closer. Should we just go ahead and remove `loadDocuments()` entirely now? Or keep the `HiddenBrowser`-based version but log a very severe warning every time it's used?

Didn't end up adding a delay option, since I realized that callers can just delay before calling `getDocument()`.

SingleFile will definitely need to run inside the content process to be useful.

dstillman:
> Should we just go ahead and remove `loadDocuments()` entirely now?

Let's remove. All code will need to be rewritten anyway, so now's the time for hard API breakage. (I'm also not sure anyone else is actually using it.)

chrome/content/zotero/actors/PageDataChild.jsm
dstillman, line 1: A couple problems here.

1) We got `documentIsReady()` from Firefox, but it doesn't seem to be working properly, at least in some cases:

```javascript
return await Zotero.Attachments.importFromURL({
    url: "https://forums.zotero.org/discussion/103790/word-add-on-disappeared-all-troubleshooting-steps-failed"
})
```

This creates an empty page for me, at least most of the time, because `documentIsReady()` resolves after only the `<head>` has been loaded, so serializing `document` produces an empty `<body>`.

2) If I add an artificial delay:

```javascript
case "documentHTML":
	await new Promise((resolve) => {
		var timer = Components.classes["@mozilla.org/timer;1"]
			.createInstance(Components.interfaces.nsITimer);
		timer.initWithCallback({
			notify: resolve
		}, 2000, Components.interfaces.nsITimer.TYPE_ONE_SHOT);
	});
	return new XMLSerializer().serializeToString(document);
```
then SingleFile saves text and images, but no CSS.

chrome/content/zotero/xpcom/attachments.js
```diff
@@ -539,39 +541,28 @@ Zotero.Attachments = new function(){
...
-					undefined,
-					(e) => {
-						reject(e);
-					},
-					true,
+		var nativeHandlerImport = async function () {
+			try {
+				let browser = await HiddenBrowser.create(url, {
+					requireSuccessfulStatus: true,
+					docShell: { allowImages: true },
```
dstillman, line 548: I didn't test, but is this actually doing anything if we're just getting and reparsing the document HTML? Seems like it might matter if we needed to run SingleFile in the actor, but not as is.

dstillman:
Not sure if this is related, or if I missed something, but if I load https://forums.zotero.org in the Browser tab and load Embedded Metadata (or any translator, I think) and do detect, I get this:

```
06:35:41 Running detectWeb
06:35:41 Detect using Embedded Metadata failed: 
         SyntaxError: unexpected token: '{'
         
         eval/<@chrome://zotero/content/xpcom/translate/src/translation/sandboxManager.js:73:4
         eval@chrome://zotero/content/xpcom/translate/src/translation/sandboxManager.js:74:6
         Zotero.Translate.Base.prototype._loadTranslator</parse<@chrome://zotero/content/xpcom/translate/src/translation/translate.js:1792:26
         
         url => https://forums.zotero.org/
06:35:41 detectWeb did not match
```

dstillman:
OK, yeah, so this error seems to be specifically from the combination of this PR and the updated `translate` submodule from zotero/translate#15. detectWeb works in the fx102 branch otherwise.

PR #2800, opened 2022-08-29
Make Feeds row into a selectable global view of feed items

### LINKED ISSUE #2786: Make "Feeds" a filterable global view of all feed items
https://forums.zotero.org/discussion/67124/search-in-feeds

AbeJellinek:
Feeds row appears bold when any feed has unread items, and it can be collapsed like a library (partially for users who don't want to see their feed list all the time and prefer just using the new global view, and partially as a visual clue that something has changed).

Closes #2786.

### DIFF
--- a/chrome/content/zotero/advancedSearch.js
+++ b/chrome/content/zotero/advancedSearch.js
@@ -76,7 +76,7 @@ var ZoteroAdvancedSearch = new function() {
 			isDuplicates: () => false,
 			isFeed: () => false,
 			isFeeds: () => false,
-			isFeedOrFeeds: () => false,
+			isFeedsOrFeed: () => false,
 			isShare: () => false,
 			isTrash: () => false
 		};
@@ -112,7 +112,7 @@ var ZoteroAdvancedSearch = new function() {
 			isDuplicates: () => false,
 			isFeed: () => false,
 			isFeeds: () => false,
-			isFeedOrFeeds: () => false,
+			isFeedsOrFeed: () => false,
 			isShare: () => false,
 			isTrash: () => false
 		};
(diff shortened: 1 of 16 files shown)


chrome/content/zotero/advancedSearch.js
```diff
@@ -75,6 +75,8 @@ var ZoteroAdvancedSearch = new function() {
 			isPublications: () => false,
 			isDuplicates: () => false,
 			isFeed: () => false,
+			isFeeds: () => false,
+			isFeedOrFeeds: () => false,
```
dstillman, line 79: Probably should be `isFeedsOrFeed`, since that's the order of appearance

chrome/content/zotero/xpcom/collectionTreeRow.js
```diff
@@ -65,12 +65,15 @@ Zotero.CollectionTreeRow.prototype.__defineGetter__('id', function () {
 		case 'trash':
 			return 'T' + this.ref.libraryID;
 		
+		case 'feeds':
+			// The "0" is kind of weird, but we need the ID to be a one-character prefix + a number
+			// and there's only one set of feeds
```
dstillman, line 70: Let's go with "F1" instead of "F0". We already use "L1" for My Library, which is technically the internal libraryID, but still, a little less weird-looking.

dstillman (commented):
1) The feeds are at the same level as the container (at least for me), when they should be indented.

2) I also just got a weird state where the container itself was indented, but not sure how that happened, and I haven't been able to reproduce after restarting:
  Screenshot 2022-11-12 at 6 15 28 AM

3) I think for this container to make sense we kind of need to show the feed each entry is from? We could add a column in this view, or we could just add a "Feed:" row at the top of the item pane. Column seems more useful but might take up too much space?

PR #2801, opened 2022-08-30
Add "Capitalize" text transform to creator context menu

### LINKED ISSUE #39: Text transform on authors

AbeJellinek:
Fixes #39

### DIFF
--- a/chrome/content/zotero/elements/itemBox.js
+++ b/chrome/content/zotero/elements/itemBox.js
@@ -191,10 +191,15 @@
 				}
 			});
 
+			this._id('zotero-field-transform-menu').addEventListener('popupshowing', () => {
+				this._id('creator-transform-title-case').disabled = !this.canTextTransformField(document.popupNode, 'title');
+				this._id('creator-transform-sentence-case').disabled = !this.canTextTransformField(document.popupNode, 'sentence');
+			});
+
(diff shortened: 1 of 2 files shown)


dstillman:
1. Menu doesn't currently appear at all for single-field creators (presumably because previously it only add name swapping)
2. Need to do this without the `toUpperCase()`. It currently will mess up "MacGregor" or "PhD".
3. Would be nice to disable Capitalize if it wouldn't make any changes to the current name.

dstillman:
We should probably either change "Capitalize" to "Capitalize Names" or change "Swap First/Last Names" to "Swap First/Last". Maybe the latter, just to be generally less verbose?

PR #2802, opened 2022-09-01
Clear sync error and start sync after logging in in prefs

### LINKED ISSUE #2785: Clear sync error in toolbar after setting up syncing in preferences
Otherwise when you close the prefs you might still see the sync error that took you to the Sync preferences in the first place, which might be confusing (though it would disappear as soon as you clicked the Sync button).

Some people also expect syncing to kick off automatically, but people need to be able to configure settings while the prefs are open. We could consider kicking off auto-sync once the prefs window closes if syncing was set up…

AbeJellinek:
It would be nice if `preferences.js` could just send an `unload` (or `beforeunload`) event to panes like it does with `load`, but it seems that, at least on macOS, the interpreter won't dispatch events from a [`before`]`unload` handler. So for now, at least, panes need to add their own listeners to the window.

Also fixes an existing error due to an argumentless call to `Zotero.Sync.Runner.updateIcons()` in ZoteroPane.

Fixes #2785.

### DIFF
--- a/chrome/content/zotero/preferences/preferences_sync.jsx
+++ b/chrome/content/zotero/preferences/preferences_sync.jsx
@@ -38,7 +38,6 @@ var { renderCell } = VirtualizedTable;
 Zotero_Preferences.Sync = {
 	checkmarkChar: '\u2705',
 	noChar: '\uD83D\uDEAB',
-	syncOnClose: false,
 	
 	init: async function () {
 		this.updateStorageSettingsUI();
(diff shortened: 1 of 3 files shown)


chrome/content/zotero/preferences/preferences_sync.jsx
```diff
@@ -182,6 +182,19 @@ Zotero_Preferences.Sync = {
...
+		
+		// It shouldn't be possible for a sync to be in progress if the user wasn't logged in,
+		// but check to be sure
+		if (!Zotero.Sync.Runner.syncInProgress) {
+			// Clear any displayed sync errors
+			Zotero.Sync.Runner.updateIcons([]);
+		}
+		window.addEventListener('beforeunload', () => {
+			if (!Zotero.Sync.Runner.syncInProgress) {
+				Zotero.Sync.Runner.sync();
```
dstillman, line 194: This will start a sync even if auto-sync is disabled. Could check `Zotero.Prefs.get('sync.autoSync')`, or could just use `Zotero.Sync.Runner.setSyncTimeout(1)` instead, which does the check. Maybe a little less jarring to wait a second anyway?

Issue #999, opened 2016-05-11
Creator buttons should be keyboard accessible


dstillman:
Creator type menu, field-mode switcher, and "-"/"+" buttons (though Shift-Enter does work for adding new creator rows)

dstillman:
https://forums.zotero.org/discussion/76163/keyboard-shortcuts-for-changing-author-tag

dstillman:
https://forums.zotero.org/discussion/comment/417401/#Comment_417401

dstillman:
Done for Zotero 6. Will need to be done separately for Zotero 7.

tnajdek:
Done it both branches, we can probably close this issue.

PR #2826, opened 2022-09-20
Make creator buttons accessible via keyboard

tnajdek:
This PR addresses #999:

* Creator type, switch mode, delete and add buttons accessibly via keyboard
* Slightly improve experience with text-to-speech tools (this could be further improved but probably best to do that in fx102 branch)
* Added very light outline to highlight current focus (within creator field)


dstillman:
This is great, but we should trigger the menu and buttons on space bar instead of Return. Would also be good to open the menu (and ideally select the first row, though that's not that big a deal) on up/down-arrow as well.

tnajdek:
* menu and buttons now trigger on space bar instead of Return
* menu now also triggers on down-arrow

PR #2832, opened 2022-09-22
Add split view for PDF reader

mrtcode:
We can try this out.

Split view currently doesn't persist any state, but this is a wider topic that we have to discuss.

Initially I started implementing the second view as a plain PDF.js, but that wasn't a good idea, because too many necessary features were missing (our custom selection, clicking on DOI, etc.). Therefore, now we are loading a fully working PDF reader instance (as another iframe inside the parent PDF reader iframe) and disabling all the unnecessary features. And this way I can easily enable annotations support and anything else, in future, if needed.


mrtcode:
Let's not hurry with the review. I'll finish the persistence part.

mrtcode:
Ok,  this is ready for review.

We are storing window/tab second view state in session.json now:
```json
"secondViewState": {
  "pageIndex": 7,
  "scale": "page-width",
  "rotation": 0,
  "top": 690,
  "left": -14,
  "scrollMode": 0,
  "spreadMode": 0,
  "splitType": "vertical",
  "splitSize": "50%"
}
```

dstillman:
This is great.

We probably want the split actions in the View menu as well?

We don't need all of those properties in `secondViewState`, do we? Aren't `rotation` (soon to be removed anyway),`scrollMode, and `spreadMode` just based on the parent?

mrtcode:
It should be ok now. `scrollMode` and `spreadMode` aren't saved anymore, and `rotation` will be removed in the next commit, once we merge this.

mrtcode:
Ok, I'm going to merge this.

dstillman:
👍

PR #2833, opened 2022-09-22
fx-compat: Make creator buttons accessible via keyboard

### LINKED ISSUE #999: Creator buttons should be keyboard accessible

tnajdek:
This PR fixes #999 in fx102 branch using similar approach as with #2826 but with few small improvements:

* When opening creator type popup, first option is automatically selected (or last option if `ArrowUp` was used to trigger this)
* No need for custom focus outline
* Slightly better text-to-speech support

### DIFF
--- a/chrome/content/zotero/elements/itemBox.js
+++ b/chrome/content/zotero/elements/itemBox.js
@@ -856,15 +856,19 @@
 				: this._creatorTypeMenu.childNodes[0].getAttribute('typeid');
 			
 			var rowIndex = this._creatorCount;
+			var tabindex = this._tabIndexMinCreators + ((rowIndex - 1) * 6);
 			
 			var th = document.createElement("th");
 			th.setAttribute("typeid", typeID);
 			th.setAttribute("fieldname", 'creator-' + rowIndex + '-typeID');
 			if (this.editable) {
-				th.className = 'creator-type-label zotero-clicky';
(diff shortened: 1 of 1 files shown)


dstillman:
> first option is automatically selected (or last option if ArrowUp was used to trigger this)

No need for this conditional. Standard behavior in all browsers (at least on macOS) is to still select the first option when pressing ArrowUp. Try [here](https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_select). And Return should *not* trigger it, only space bar and up/down.

In #2826 I also made it remove the focus ring when hovering over the menu, since we have our custom hover effect there. Now that the focus ring matches the other fields exactly, having both showing is maybe a little less weird, since it just seems like of course you'd have the keyboard focus indicator when it has keyboard focus? I could go either way on this.

tnajdek:
@dstillman updated:

* fixed the `ArrowUp` behaviour
* creator type selector will no longer trigger on Return
* creator buttons also won't trigger on Return
* left the focus outline as is

I think there is a nice distinction between hover (background color) and focus (outline). It's visually consistent with input fields. What we could do maybe is to match the shape of the outline on the item type selector (rounded corners)?

dstillman:
Great!

PR #2841, opened 2022-09-29
fx-compat: Restore word processor integration

adomasven:
- Word for Windows not tested since there's no build yet
- Word for Mac currently causes a sigsev, and likely we'll see this on windows too. This is happening on an API call where we pass a callback, so likely an issue in ctypes that we'll have to work around.
- All integration UIs restored.
- Added a component for richlistitems with a checkbox with corresponding interactions (spacebar/double-click to toggle).

**Requires `fx102-wordplugins` branch in `zotero-standalone-build`**


chrome/content/zotero/elements/richlistCheckbox.js
```diff
@@ -0,0 +1,85 @@
...
+"use strict";
+
+{
+	if (!customElements.get("richlistitem")) {
+		delete document.createXULElement("richlistitem");
+	}
+	/**
+	 * Extend richlistbox for checkbox inputs since we use them in multiple places
+	 */
+	class RichlistCheckbox extends customElements.get('richlistitem') {
```
dstillman, line 35: → `RichListCheckbox`

chrome/content/zotero/integration/simpleEditor.js
```diff
@@ -0,0 +1,580 @@
...
+	The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
+
+	THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT
+	LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+	IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+	LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+	WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+
+// Based on https://github.com/jaredreich/pell/blob/master/src/pell.js
```
dstillman, line 46: Let's combine all of the above into a single comment and move this line up above the MIT license with "Based on [URL], which is covered by the following copyright and permission notice:"

Source: https://softwareengineering.stackexchange.com/a/368621

dstillman (commented):
General notes:

  * I'm getting an error even trying to install the Mac Word plugin, but maybe that's just the same ctypes problem?  `Installation could not be completed because an error occurred.    [Exception... "ComponentManager::CreateInstance returned failure code:"  nsresult: "0x80570015 (NS_ERROR_XPC_CI_RETURNED_FAILURE)"  location: "JS frame :: resource://zotero-macword-integration/installer.jsm :: Plugin</this.install :: line 50"  data: no]`
  * Error trying to transfer document: `JavaScript error: chrome://zotero/content/bibliography.js, line 271: TypeError: document.documentElement.acceptDialog is not a function` (but works after manual OK)
  * Error clicking within quick format dialog: `JavaScript error: chrome://zotero/content/integration/quickFormat.js, line 1207: TypeError: node.dataset is undefined`
  * Blue bubble (and text within) is a lot smaller than before
  * Various other styling issues that we can fix separately (e.g., we really need to restore some kind of general groupbox styling)
  * Are we actually using the rich-list checkbox somewhere?

For zotero-standalone-build:

  * Spaces → tabs in build.sh
  * `integration-plugins` → `integration` for the directory? `integration-plugins` is a little verbose
  * We're not actually using most of the stuff in the manifest.json files, right? It's confusing to include a WebExtension-looking manifest.json with stuff like `description` and `strict_max_version` when it's not actually a Zotero plugin. If we need to get the version number from there, we could just have that property and remove the rest, or even just change it to version.txt to avoid the misleading manifest.json name.
 
Otherwise seems to be working great!

adomasven:
> General notes:
> 
> * I'm getting an error even trying to install the Mac Word plugin, but maybe that's just the same ctypes problem?  `Installation could not be completed because an error occurred.    [Exception... "ComponentManager::CreateInstance returned failure code:"  nsresult: "0x80570015 (NS_ERROR_XPC_CI_RETURNED_FAILURE)"  location: "JS frame :: resource://zotero-macword-integration/installer.jsm :: Plugin</this.install :: line 50"  data: no]`

Testing on a M1 Mac? Probably ctypes issue then, yes.

> * Blue bubble (and text within) is a lot smaller than before

I'm not seeing this on macOS or Linux. Some sort of DPI issue?

> * Are we actually using the rich-list checkbox somewhere?

In LibreOffice installer, but there are other places where we should use it: dictionaryManager.xhtml/js and selectItems.xhtml/js (Scaffold).

resource/word-processor-plugin-installer.js
```diff
@@ -63,52 +62,32 @@ var ZoteroPluginInstaller = function(addon, failSilently, force) {
...
-			if(this.force || (
-					(
-						Services.vc.compare(version, this._addon.LAST_INSTALLED_FILE_UPDATE) < 0
-						|| (!Zotero.isStandalone && !this.prefBranch.getBoolPref("installed"))
-					)
-					&& !this.prefBranch.getBoolPref("skipInstallation")
-				)) {
-					
-				var me = this;
+			this._version = await Zotero.File.getContentsFromURLAsync(this._addon.VERSION_FILE);
```
dstillman, line 75: Can we add a `trim()` here just to avoid some future breakage if we edit in the wrong text editor? Not sure what `setCharPref()` would do with a trailing newline.

dstillman:
Plugin installers are still looking for `integration-plugins`

dstillman:
> TypeError: Zotero.File.getContentsFromURLAsync(...).trim is not a function

Need parens

PR #2842, opened 2022-09-29
Add local implementation of Zotero web API

AbeJellinek:
This required some tweaks to other parts of Zotero infrastructure:

- Search:
  - Add 'includeDeleted' condition to match behavior of 'includeTrashed' API parameter in a single search
- Data objects:
  - Improve `toResponseJSON()` implementations so output better matches the web API
    - Add `toResponseJSON()` to `Zotero.Tags` - has to be async so it can query the database and generally works differently from other `toResponseJSON()` functions, but accomplishes the same task
  - Remove unused `getAPIData()` and `apiDataGenerator()` DataObject functions. They aren't functional and wouldn't really make implementing the local server easier, so now seemed like a decent time to remove them
- Server:
  - Support resolving routes using `pathparser.jsm`
    - Add `allowMissingParams` option to `PathParser#add()`: prevents `/route` from matching `/route/:param`
  - Replace the `query` property of the data object sent to endpoint `init()`s with `searchParams`, an instance of `URLSearchParams` - supports `#getAll()` for repeatable parameters
- URIs:
  - Make `getObjectURI()` public, add utilities for converting URIs to API endpoints and web library URLs

### DIFF
--- a/chrome/content/zotero/xpcom/connector/server_connector.js
+++ b/chrome/content/zotero/xpcom/connector/server_connector.js
@@ -1434,7 +1434,7 @@ Zotero.Server.Connector.Import.prototype = {
 		}
 		
 		try {
-			var session = Zotero.Server.Connector.SessionManager.create(requestData.query.session);
+			var session = Zotero.Server.Connector.SessionManager.create(requestData.searchParams.get('session'));
 		}
 		catch (e) {
 			return [409, "application/json", JSON.stringify({ error: "SESSION_EXISTS" })];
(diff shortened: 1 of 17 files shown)


cboulanger:
Hi, I also would be very much interested in this getting merged. Any updates on the status of this PR? Will it be part of Zotero 7?

dstillman:
Merged

ferdinandyb:
Great news! When is ths expected to make it's way into a release?

PR #2845, opened 2022-09-30
fx-compat: Preferences: Add `stylesheets` param, activate processing instrs, default to XUL

### LINKED ISSUE #2835: fx-compat: Switch back to XUL for plugin preference files
Sorry about this — I [suggested](https://github.com/zotero/zotero/pull/2659#discussion_r940893880) that we switched to default-HTML for plugin prepane fragments, but I think I was wrong there. I thought Mozilla was using HTML for prefs, but it looks like they're actually switching to the XUL namespace for the `<body>` [within an HTML document](https://searchfox.org/mozilla-central/source/browser/components/preferences/preferences.xhtml), and all the [included files](https://searchfox.org/mozilla-central/source/browser/components/preferences) are XUL, with prefixed HTML.

Playing with this for the sample plugin, I'm realizing that defaulting to HTML is quite a bit more annoying, since styling various things would require CSS that wouldn't with XUL.

(Relatedly, it seems like we don't provide a way to include a CSS file when registering the prefpane, which probably we want to have, even if we default to XUL?)

AbeJellinek:
Fixes #2835.

### DIFF
--- a/chrome/content/zotero/preferences/preferences.js
+++ b/chrome/content/zotero/preferences/preferences.js
@@ -49,25 +49,6 @@ var Zotero_Preferences = {
 
 		document.getElementById('prefs-search').focus();
 		
-		this.clearAndAddPanes();
-	},
-
-	/**
-	 * (Re)initialize the preference window, adding all panes registered with Zotero.PreferencePanes.
-	 * Current selected pane and scroll position are saved and restored after load.
-	 */
-	clearAndAddPanes: function () {
-		// Save positions and clear content in case we're reinitializing
-		// because of a plugin lifecycle event
-		let navigationValue = this.navigation.value;
-		let navigationScrollTop = this.navigation.scrollTop;
-		let navigationScrollLeft = this.navigation.scrollLeft;
-		this.navigation.replaceChildren();
-		let contentScrollTop = this.content.scrollTop;
-		let contentScrollLeft = this.content.scrollLeft;
-		// Remove all children besides the help container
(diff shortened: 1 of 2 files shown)


dstillman:
The problem with `<?xml-stylesheet`, along with it being a weird thing that no one uses, is that it requires a hard-coded URL, which means it requires chrome registration, whereas preferences.xhtml and preferences.js will just be passed as root-path files.

`Ci.amIAddonManagerStartup::registerChrome()` also [doesn't support `skin`](https://searchfox.org/mozilla-central/rev/12a34801fbaa6f14e20515a44b1fd77179b02617/toolkit/mozapps/extensions/AddonManagerStartup.cpp#818-838), so you'd have to use a `content` URL instead, which is OK but is just an annoying thing to set up when you can pass the other two files dynamically. And it would mean that you couldn't keep preferences.css next to the other two files.

Any way we can just take a CSS file along with the other two?

dstillman:
> But it does feel odd to move more and more resources out of the XHTML, where I think they naturally belong, and into the pane registration

Oh, I don't think so. It's just a fragment. I wouldn't say the style inherently belongs there at all (as evidenced by the use of a hackily-parsed-out `<?xml-stylesheet` in the first place). If it were a full document, that would be one thing, but I think it's totally appropriate to define associated JS and CSS separately for a fragment.

I guess there's no real harm in supporting `{{ROOT_URI}}` for images, though I'm not sure anyone needs that, since any image usage could (and often would need to) be done by JS anyway.

PR #2853, opened 2022-10-06
Made toolbar accessible to keyboard

jaeschrich:
I made the toolbar accessible to the keyboard. There are two sections, one before and one after the search bar. When you tab into a section you can move around with the arrow keys to select an appropriate button. The code listenes on two toolbarbuttons per section, one at the start and end, to determine from which direction the user's focus originates.

The sync button does not have a meaningful label for screen readers, but the other toolbarbuttons do.

### DIFF
--- a/chrome/content/zotero/lookup.js
+++ b/chrome/content/zotero/lookup.js
@@ -218,7 +218,9 @@ var Zotero_Lookup = new function () {
 			event.stopPropagation();
 			this._button.focus();
 			this._button = null;
-		} else this._button = null;
+		} else {
+			this._button = null;
+		}
 	}
 	
 	
--- a/chrome/content/zotero/lookup.js
+++ b/chrome/content/zotero/lookup.js
@@ -205,12 +205,20 @@ var Zotero_Lookup = new function () {
 	}
 
 	this.onFocusOut = function(event) {
+		/*
+			if the lookup popup was triggered by the lookup button,
+			we want to return there on focus out. So we check 
+			(1) that we came from a button and (2) that
+			event.relatedTarget === null, i.e. that the user hasn't used
+			the mouse or keyboard to select something, and focus is leaving
+			the popup because the popup was hidden/dismissed.
+		*/
 		if (this._button && event.relatedTarget === null) {
 			event.preventDefault();
 			event.stopPropagation();
 			this._button.focus();
 			this._button = null;
-		}
+		} else this._button = null;
 	}
 	
 	
--- a/chrome/content/zotero/lookup.js
+++ b/chrome/content/zotero/lookup.js
@@ -28,6 +28,7 @@
  * @namespace
  */
 var Zotero_Lookup = new function () {
+	this._button = null;
 	/**
 	 * Performs a lookup by DOI, PMID, or ISBN on the given textBox value
 	 * and adds any items it can.
@@ -135,6 +136,7 @@ var Zotero_Lookup = new function () {
 	this.showPanel = function (button) {
 		var panel = document.getElementById('zotero-lookup-panel');
 		panel.openPopup(button, "after_start", 16, -2, false, false);
+		this._button = button;
 	}
 	
 	
@@ -200,6 +202,15 @@ var Zotero_Lookup = new function () {
 			document.getElementById("zotero-lookup-panel").hidePopup();
 		}
 		return true;
+	}
+
+	this.onFocusOut = function(event) {
+		if (this._button && event.relatedTarget === null) {
+			event.preventDefault();
+			event.stopPropagation();
+			this._button.focus();
+			this._button = null;
+		}
 	}
 	
 	
--- a/chrome/content/zotero/lookup.js
+++ b/chrome/content/zotero/lookup.js
@@ -135,8 +135,11 @@ var Zotero_Lookup = new function () {
 
 	this.showPanel = function (button) {
 		var panel = document.getElementById('zotero-lookup-panel');
-		panel.openPopup(button, "after_start", 16, -2, false, false);
 		this._button = button;
+		if (!button) {
+			button = document.getElementById("zotero-tb-lookup");
+		}
+		panel.openPopup(button, "after_start", 16, -2, false, false);
 	}
 	
 	
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -56,7 +56,6 @@ var ZoteroPane = new function()
 	this.viewSelectedAttachment = viewSelectedAttachment;
 	this.reportErrors = reportErrors;
 
-	// from https://github.com/zotero/web-library/blob/master/src/js/hooks/use-focus-manager.js
 	const modifierIsNotShift = ev => ev.getModifierState("Meta") || ev.getModifierState("Alt") ||
 		ev.getModifierState("Control") || ev.getModifierState("OS");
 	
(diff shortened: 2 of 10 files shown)


dstillman (changes requested):
1) Be sure to always rebase/squash or reset/re-commit with just the necessary public commits before submitting a PR — no one else needs to see your intermediate commits, and certainly not ones that show massive changes to whitespace, etc. Successive changes can be pushed as is for easier review with the intention of squashing before merge, but the initial commits should always be ready to merge as is. For this, you should rebase/fixup and force-push a clean commit.

2) Overall, compare to other programs' behavior here — this diverges in all sorts of ways. Some specifics to follow.

3) The direction stuff isn't necessary. When focus enters a group, it should start at either the first element (Firefox) or the most recent element (Slack, Zotero web library), even if it was a Shift-Tab in. I'd go with first element for now. It would only focus the right-most element if every button were a tab stop, as in Chrome.

4) Up/down-arrow shouldn't move focus within a group — only left/right should do that. Up/down should be the same as Shift-Tab/Tab, unless the focused element is a menubutton, in which case it should open the menu.

5) Spacebar should be equivalent to clicking the focused element.

6) Taken together, (4) and (5) mean that…the toolbar is mostly non-functional? Only New Collection, Advanced Search, and Sync — the regular buttons — do anything at all. Unless I'm missing something, there's no way to open any of the menubuttons (which both up/down and spacebar should do) or the Add Item by Identifier panel.

7) After a group gets focus, clicking another element doesn't change the tab index, which breaks normal Tab behavior. Click search bar, press Tab to select Locate, click search bar again, and press Tab again. Instead of Locate being selected again, My Library is selected, because the tab index remained on the toolbar. Same for clicking anywhere else on the UI. This problem is specific to the toolbar groups — if you just click an item in the items list and press Tab, the Info tab is selected, and if you click back on the item and press Tab again, Info is selected again.

8) I'm guessing you tried this only on Windows, but the dashed-line focus indicator is only shown on Windows, so on macOS and Linux there's no focus indicator at all for this. We can tweak the colors later, but you should add some sort of focus highlight at least in the Linux platform CSS. We can handle macOS separately.

9) Even if we were doing the directional thing, I don't think you would need listeners on individual buttons, and those would be easy to accidentally break later if the toolbar was rearranged. I would think you would be able to have a listener on just the toolbar — or potentially even the whole window — to handle the bubbled events for a given group.

10) Along with the specific lines I noted, revert the pdf-reader change. Shouldn't be any unrelated changes.

jaeschrich:
Hey @dstillman, I made the changes you requested.
1. Put all listeners on the toolbar.
2. Focus now starts on the first element of each zone (left and right of the search bar).
3. Up/down arrows no longer move within groups and behave like Tab/Shift+Tab. For buttons with a menu or popup panel attached, up and down both activate the menu/popup.
4. Spacebar and enter both activate the selected button, including menu or popup buttons.
5. The issue with the focus being assigned incorrectly is fixed. Clicking Search, tabbing to locate, clicking search again, and hitting tab will correctly select the locate button.
6. Reverted unrelated changes and squashed/rebased the commit history.

I added a focus indicator on the Linux platform CSS, but when I tested it under Ubuntu there was an OS level orange focus indicator. What flavor of linux were you testing with? I can set up a VM for that distro to test the focus indicator out.

dstillman (commented):
OK, sorry — after trying this, I think each pane's toolbar section should in fact get its own tab stop. It feels weird to me that it jumps to New Collection instead of New Item when you Shift-Tab from the search bar. And that way it will behave consistently if the collections pane is closed.

Otherwise looking good.

Re: Linux, I was testing with stock Ubuntu. Maybe I was on a different version. I can take another look once this is merged (since we'll need to do a focus indicator on macOS too).

dstillman:
(Had a couple line comments from an earlier version. Deleted.)

jaeschrich:
Hey Dan,
I made the changes you requested and refactored things so that there are three zones rather than two. I think it made the code cleaner, since now a zone is strictly contained by an "hbox" element and I could use DOM functions for checking containing.

dstillman:
Both the cancel button and the various sync error icons should be accessible via the toolbar. To test sync error icons, unlink any syncing account and click the sync button.

Sync error icons show a panel with more info, so the button will need to be triggerable via the keyboard.

chrome/content/zotero/zoteroPane.js
```diff
@@ -136,8 +139,182 @@ var ZoteroPane = new function()
...
+		/* set up toolbar */
+
+		const zones = [
+			new TbFocusZone(document.getElementById("zotero-collections-toolbar") ),
+			new TbFocusZone(document.getElementById("zotero-items-toolbar"),
+								() => zones[0].firstButton,
+								() => document.getElementById("zotero-tb-search-menu-button")),
+			new TbFocusZone(document.getElementById("zotero-item-toolbar"),
+								() => document.getElementById("collection-tree"),
+								() => document.getElementById("zotero-tb-search"))
```
dstillman, line 198: Nit: Three parameters on separate lines, single tab for each:

```javascript
new TbFocusZone(
    document.getElementById("zotero-item-toolbar"),
    () => document.getElementById("collection-tree"),
    () => document.getElementById("zotero-tb-search")
)
```

chrome/content/zotero/lookup.js
```diff
@@ -201,6 +203,23 @@ var Zotero_Lookup = new function () {
...
+			event.relatedTarget === null, i.e. that the user hasn't used
+			the mouse or keyboard to select something, and focus is leaving
+			the popup because the popup was hidden/dismissed.
+		*/
+		if (this._button && event.relatedTarget === null) {
+			event.preventDefault();
+			event.stopPropagation();
+			this._button.focus();
+			this._button = null;
+		} else this._button = null;
```
dstillman, line 221: Nit: Always include braces unless `if` and `else` are both single-line conditionals (which is rare)

```javascript
}
else {
    this._button = null;
}
```

chrome/content/zotero/zoteroPane.js
dstillman: Also, `zotero-tb-search` and `collection-tree` are backwards, so if you press ArrowDown from the search button, you end up in the search box.

chrome/content/zotero/zoteroPane.js
```diff
@@ -136,8 +139,182 @@ var ZoteroPane = new function()
...
+
+		for (const zone of zones) {
+			zone.firstButton.setAttribute("tabindex", "0");
+		}
+
+		toolbar.addEventListener("focusin", (event) => {
+			if (!isTbButton(event.target)) {
+				return;
+			}
+			else if (zone) {
```
dstillman, line 209: No need for `else` after `return`

chrome/content/zotero/zoteroPane.js
```diff
@@ -136,8 +139,182 @@ var ZoteroPane = new function()
...
+			} 
+			else if (!Zotero.rtl && event.key === 'ArrowLeft'
+					|| Zotero.rtl && event.key === 'ArrowRight') {
+				event.preventDefault();
+				event.stopPropagation();
+				moveFocus(-1);
+				return;
+			} 
+			/* manually trigger on space and enter */
+			else if (event.key === ' ' || event.key === 'Enter') {
```
dstillman, line 255: No `else` after `return` for these

dstillman (commented):
A couple times I've gotten into a state where Shift-Tab from the search box skips the middle toolbar, and I have to delete prefs.js to fix. I'm having trouble reproducing it, but maybe you have an idea? I was playing around with Stacked mode, but not sure if that was relevant.

dstillman:
That said, I'm having second thoughts on separating the collections toolbar and items toolbar:

1) When the collections pane is collapsed, the two toolbars are combined with no space between, so having them be separate tab stops is weird.

2) When the item pane is open, there's a large gap between Locate and Sync, despite them being in the same tab stop.

So maybe it's simpler to just have two zones. Instead of passing a single node to create the zone, can we pass an array of nodes? Sorry for the back and forth!

chrome/content/zotero/zoteroPane.js
```diff
@@ -175,14 +167,60 @@ var ZoteroPane = new function()
...
-			new TbFocusZone(
-				"zotero-tb-locate", 
-				"zotero-tb-search",
-				null)
+			{
+				get start() { return document.getElementById("zotero-tb-collection-add"); },
+				focusBefore() {
+					// need to check if the itembox is open
+					const pane = document.getElementById("zotero-item-pane-content");
+					if (pane.selectedIndex === "0") {
```
dstillman, line 175: Remove comment above and add "// If no item is selected, focus items list" above this line

chrome/content/zotero/zoteroPane.js
```diff
@@ -175,14 +167,60 @@ var ZoteroPane = new function()
...
+							const notes = document.getElementById("zotero-editpane-notes");
+							const nodes = notes.querySelectorAll("button");
+							const node = nodes[nodes.length - 1];
+							node.focus();
+							// TODO: the notes are currently inaccessible to the keyboard
+						}
+						else if (tabBox.selectedIndex === 2) {
+							const tagContainer = document.getElementById("tags-box-container");
+							const tags = tagContainer.querySelectorAll("#tags-box-add-button,.zotero-clicky");
+							const next = tags[tags.length - 1];
```
dstillman, line 194: `next` → `last`

chrome/content/zotero/zoteroPane.js
```diff
@@ -175,14 +167,60 @@ var ZoteroPane = new function()
...
+							else {
+								next.click();
+							}
+						}
+						else if (tabBox.selectedIndex === 3) {
+							const related = tabBox.querySelector("relatedbox");
+							related.receiveKeyboardFocus();
+						}
+						else {
+							throw new Error("This error should be unreachable — the selectedIndex should always be between 1 and 4");
```
dstillman, line 207: Don't need the first part — an uncaught error is always something that shouldn't happen

chrome/content/zotero/zoteroPane.js
```diff
@@ -175,14 +167,60 @@ var ZoteroPane = new function()
...
+							if (next.id === "tags-box-add-button") {
+								next.focus();
+							}
+							else {
+								next.click();
+							}
+						}
+						else if (tabBox.selectedIndex === 3) {
+							const related = tabBox.querySelector("relatedbox");
+							related.receiveKeyboardFocus();
```
dstillman, line 204: If we're going to bother with a function instead of reaching in directly, we should pass in a direction, even if it's not used now.

jaeschrich:
Hey Dan,
Sorry, I didn't realize you got notified whenever I pushed to my personal repo—I would have made a comment here otherwise. I've fixed your comments. I wasn't able to reproduce the bug you found, in Stacked view, but I added some code to manually move focus from the search box to the left toolbar zone. Let me know if that fixes it!

PR #2854, opened 2022-10-06
Track Date Last Opened, add virtual collection + search condition

### LINKED ISSUE #2844: Recently Read virtual collection
Or "Recently Opened", since we probably wouldn't actually track page changes (and it will apply to external opens too), but "Recently Read" is maybe a slightly less technical way of phrasing it.

We probably want this to sync, unfortunately, because it would be pretty nice to open a PDF on your iPad and then have it show up in this collection on your desktop.

We can either add a `lastOpenTime` to `itemAttachments` or use a synced setting like `lastPageIndex`. The latter would be easier to add to sync, and we could just periodically purge those settings that are older than 2 weeks or so, but there might be a benefit to knowing the last time you opened any item in your library? Maybe you want to clean your library and delete anything you haven't opened in some number of years?

### LINKED ISSUE #3102: When removing a note, count in 'Notes' column is not updated
Seems like the items tree is not refreshed.

https://forums.zotero.org/discussion/104714/bug-number-of-notes-is-not-updated

AbeJellinek:
Adds:
- `itemAttachments.dateLastOpened` column
	- Requires a migration but doesn't mark the database as incompatible with older versions
	- I went with the database column approach because I think it's nifty/potentially useful to track last-open dates for all items for all time, and having it available as a search condition will be useful for some workflows
	- Never set on group items, and the data server should reject group library item updates containing it
- "Attachment Last Opened" search condition
- "Last Opened" item tree column
	- Parent items show the most recent open date of any of their attachments
	- Name borrowed from the Finder
- "Recently Read" virtual collection
	- Sorts by Last Opened descending, only shows items that have been read within the last two weeks

In the process:

- Made it possible to have different CollectionTreeRow visibility groups (i.e. different sets of visible columns and sorts) that show intersecting sets of items
	- Previously, items already added to the item tree would never be re-sorted when switching between collection tree rows, even if the visibility group (and hence sort columns) changed. That was fine because the only two visibility groups, `default` and `feed`, looked at mutually exclusive sets of items, but I needed to add a `recentlyRead` group that shows some of the same items as `default` but has a different sort. Now all visible items are re-sorted when switching visibility groups, not just new items.
- Fixed # of Notes and Attachment columns not updating as they should when a child was added/trashed
	- I don't think we had an issue for this, but I remember seeing reports on the forums

Will need a data server update to be able to sync, of course. This PR doesn't touch anything particularly platform-y, so it should be easy enough to port to master if we want.

Closes #2844, fixes #3102.


dstillman:
> Never set on group items, and the sync server should reject group library item updates containing it

Hmm, I didn't think of this. This we couldn't do with the column approach, but we could do it with synced settings (stored in the user library) if we wanted it — that's how we do last page position. I'm not sure how confusing it would be to see Recently Read in a group library and have it only apply to yourself, but I feel like we're going to get requests for this as soon as this rolls out… (It would mean that any features based on the column approach wouldn't work in a group library, but that's OK — the main use case I identified, clearing out old items, is less relevant in a group anyway.)

> Attachment Date Last Opened
> Date Last Opened

I think "Attachment Last Opened" (condition) and "Last Opened" (column) are fine. We say "Date Added" and "Date Modified" in general, but I think this is clear, and they're getting pretty long otherwise.

> The name has the "Items" suffix to fit in with other collection tree rows, but it is getting to be a bit of an awkward length. Should we drop it?

Definitely.

(I'll try it out later.)

dstillman:
I guess we might as well just ignore groups for now and gauge demand later. Since it would be a different implementation anyway, there's not much benefit to doing it now.

dstillman:
Oops, it just occurred to me that we have `lastAccessed` in #1919 for purging local files after a certain number of days:

https://github.com/zotero/zotero/pull/1919/files#diff-59406524bc1dcc5573ad2eb6ec47feba47db172355a78cc8b6b0613e9f120c2c

That PR got delayed in the lead-up to Z6, but I was going to try to merge it soon. But we should reconcile these two dates.

1) Let's go with `IA.lastAccessed` and `item.attachmentLastAccessed` from there. (It looks like it's a nearly identical implementation. Sorry!)

2) We don't need to limit it to group libraries, because we'll use this for file purging there.

3) In the context of that feature (and property name), I'm realizing that we also want to update this while PDFs are being interacted with, not just when they're opened. Every time you scroll around in a PDF — basically, whenever we update `lastPageIndex` — the PDF should arguably rise to the top of Recently Read. Maybe we can be a little more judicious about that and only update it when the item isn't already the most recently accessed item (so store that in an in-memory property somewhere) or it's been more than some period of time since the last update (5 minutes? only matters for the display in the items list).

4) For file purging we also need to avoid purging files in open tabs, but I think that can just be a check for open tabs at purge time, since you probably don't want background tabs to populate Recently Read.

If that all makes sense, we can switch this to `attachmentLastAccessed`, and I can later rebase that, which will just remove most of the new stuff for that property.

dstillman:
> 2. We don't need to limit it to group libraries, because we'll use this for file purging there.

Oh, except we wouldn't sync that for group libraries — in group libraries, it would just be a local property for use in file purging. Which is a little weird, but hopefully OK (and the server will enforce).

dstillman:
> Same names in the UI as this PR has now, though?

I guess it should be "Last Opened" → "Last Read" and "Attachment Last Opened" → "Attachment Last Read" anyway, just so that it's clear that it's referring to the same thing as "Recently Read".

(`accessed` is now more semantically aligned with "read", which is good.)

dstillman:
Any way to just not show that column in group libraries without creating a separate visibility group?

dstillman:
If @adomasven can't suggest an existing way to do this, maybe it'd work to pass an array of temporarily hidden columns to VirtualizedTable, which would just skip them on render without affecting the list of columns that get cached for the visibility group?

We'll also have this issue for Added By when we make that available in group libraries.

Having separate visibility groups isn't that big a deal, but I don't really see much use for it, and it does make it more annoying to manage columns — if you don't want this, which I suspect most people won't, you have to make every change twice.

adomasven:
I'm thinking of a "proper" way to do this. The constraints of having to make column stuff be fast, being able to serialize the prefs for any tree and now having partial-visibility groups make it complicated.

In `ItemTree._getColumns()` there's actually a distinction between `this.id` which is used as a `prefKey` for column preferences storage, and `this._columnsId` which is used to check if columns for the current view are cached or not. So we could set `this._columnsId = null` in `ItemTree.changeCollectionTreeRow()` when switching between group and user library views and then do some conditionals in `_getColumns()`. 

Ideally we don't do conditionals based on column `dataKey` and instead mark those columns with a property in `itemTreeColumns.js`. We have a `disabledIn` property there, but it filters based on visibility groups (currently only used for feed), and this is ... visibility subgroups? Add another key like `disabledInSubgroup`? I don't like that, especially since visibility groups and subgroups are not explicitly defined anywhere. Thoughts?

dstillman:
Isn't "visibility group" kind of a misnomer? Because it's really about user-settable column visibility, position, and sorting, not just whether a given column appears, right? It's more of a "column configuration".

So then `disabledIn` is sort of conflating column visibility in a given type of view with which views have separate column configs. And it's entirely possible that we'd eventually want to make it possible to have any number of individual column configs (as you can in Thunderbird, which also lets you apply the column/view settings to other folders). Feeds are part of a separate column config because they're a totally different conceptual thing, and that would be the case regardless of the available columns.

And it's columns being available or unavailable that's the relevant issue here — just as some columns are disabled in `feeds` now, `dateLastRead` will be disabled in `groups`/`feeds` (but not `personal`), and (eventually) `addedBy` will be disabled in `personal`/`feeds` (but not `groups`). I could also see having a `dateDeleted` (which we already track in the DB) disabled everywhere except `trash` views, and as a user you wouldn't want to have to configure columns separately for trash views. So then `disabledIn` is really about "view types" or something — or maybe it's even about "visibility groups" if we rename "visibility groups" to "column configurations".

The only catch is that, if a column only sometimes appears in a given column config, and someone sorts by that column…what do we do if they then switch to another view with the same column config but with that column disabled? (E.g., they sort by "Date Deleted" in the trash and then click on My Library.) Seems like we need to store multiple active columns if a column that's disabled in some view types is selected, so that there's always guaranteed to be a "selected" column.

dstillman:
That sounds about right.

I haven't fully thought it through, but the sort-fields stuff might get a little more complicated than that — e.g., you might select Date Last Read in My Library, Date Deleted in the trash, and Added By in a group, and that's not counting a secondary sort. So it would have to keep including columns in the sort fields as long as they're disabled somewhere, and then add the secondary sort? Which maybe means fields that are disabled somewhere don't qualify for secondary sort, because otherwise things get too weird? Something like that.

adomasven:
I like the overall idea of view types and column configurations

> Replace `sortDirection` and the `secondarySort.*` prefs with a `sortFields` array stored in the column prefs
> * If you switch to a row with a view type that doesn't display your primary sort field (`sortFields[0]`), your secondary sort (`sortFields[1]`) will be used as primary and your tertiary sort (`sortFields[2]`) as secondary, and so on (but saved preferences won't change - once you switch back to a view type that does display `sortFields[0]`, it will be used as the primary sort field again)

Yeah, the sort stuff may get a little confusing. If a user sorts by Date in My Library, then by Added By in a group, and switches back to My Library, the sort by Date will be maintained, but if a user sorts by Title in a group and switches back to My Library, the sort by Date will not be maintained. Maybe this is rare and not noticable/annoying enough that we should care about it though.

dstillman:
> if a user sorts by Title in a group and switches back to My Library, the sort by Date will not be maintained

Maybe I'm misunderstanding, but why should it be? Isn't the point that we don't actually want separate configs for personal and group libraries? Unless you're sorting by a field that's not present in the other view, the sort should persist. So it's not that this is rare but that it's expected/desired.

adomasven:
> Maybe I'm misunderstanding, but why should it be? Isn't the point that we don't actually want separate configs for personal and group libraries? Unless you're sorting by a field that's not present in the other view, the sort should persist. So it's not that this is rare but that it's expected/desired.

It's just that from a user's perspective these rules may be opaque and sorting by Added By, or Recently Viewed or whatever may not be that uncommon, or switching between those sorts. And then "sometimes" sorts will persist between My Library and group libraries, and sometimes they will not. Once again, maybe this is not a real issue in practice, but it is a maybe odd and annoying corner-case.

dstillman:
I'm always a fan of gnarly SQL.

chrome/content/zotero/xpcom/readObserver.js
```diff
@@ -0,0 +1,75 @@
...
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Affero General Public License for more details.
+    
+    You should have received a copy of the GNU Affero General Public License
+    along with Zotero.  If not, see <http://www.gnu.org/licenses/>.
+    
+    ***** END LICENSE BLOCK *****
+*/
+
+Zotero.ReadObserver = {
```
dstillman, line 26: How about `AttachmentReadObserver`? `ReadObserver` is a bit vague.

dstillman:
OK, just tried it out.

1) The collection row should probably go at the top, directly below the library? The point of this is to save time, and it doesn't really do that super well if you have to scroll down through hundreds of collections to find it. The other special collections are all special-use things, but this is something people might use constantly.

2) Is this supposed to show in group libraries? I'm not seeing it there.

3) We ended up with Recently Read just being a separate view group, with Last Read enabled by default, right? So if you want other columns there you just have to configure them separately (and then you get the same ones in all Recently Read views, assuming this shows in groups)?

4) Unfortunately, clicking this in my test library takes 3 seconds, so we're going to need to revisit the gnarliness of that SQL query, but I can look into that more later.

dstillman:
> Or we could save lastRead to itemAttachments even for group items, populating it with the value of the synced setting.

Yeah, I think that's fine. And then it can use the index.

And we're already excluding group `lastRead` values from `toJSON()`, so it won't even try to sync those as is.

dstillman:
OK, `lastRead_` should now be allowed in user libraries.

dstillman:
(Oh, but I still need to enable `lastRead` syncing for personal libraries.)

dstillman:
Sorry, the settings change didn't get deployed properly to all of the dataserver nodes, so it would've only worked intermittently. Should work now.

rmartins-net:
how can I use this feature in zotero ?

dstillman:
Rebased, plus some additional fixes.

I've also implemented the dataserver side to allow syncing of `lastRead` in user libraries, but that's not on production yet. We'll bump the schema version and hide the `lastRead` property from older clients so they don't get errors. That means a client that updates won't immediately see existing recently read items, but that's fine — it will get new items going forward.

dstillman:
We still need an icon from @yexingsha for this.

yexingsha:
Let's just use a clock icon, call it "Recent", and color it `tag-purple`:

recent

Sidebar

PR #2863, opened 2022-10-10
fx-compat: Automatically register plugin l10n sources

### LINKED ISSUE #2837: fx-compat: Accept Fluent .ftl files in `Zotero.PreferencePanes.register()`
In addition to .dtd, for developers who don't want to start with an obsolete format

(We should also document how to use localization.)

AbeJellinek:
Fixes #2837: FTL files can be included with `<link>` tags in pane XHTML fragments without manual configuration.

### DIFF
--- a/chrome/content/zotero/xpcom/intl.js
+++ b/chrome/content/zotero/xpcom/intl.js
@@ -85,6 +85,30 @@ Zotero.Intl = new function () {
 				this.strings[regexpResult[1]] = regexpResult[2];
 			}
 		}
+		
+		// Automatically register l10n sources for enabled plugins.
+		
+		// A Fluent file located at
+		//   [plugin root]/locale/en-US/make-it-red/strings.ftl
+		// could be included in an XHTML file as
+		//   <link rel="localization" href="make-it-red/strings.ftl"/>
+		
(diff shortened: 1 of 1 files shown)


chrome/content/zotero/xpcom/intl.js
```diff
@@ -85,6 +85,30 @@ Zotero.Intl = new function () {
 				this.strings[regexpResult[1]] = regexpResult[2];
 			}
 		}
+		
+		// Automatically register l10n sources for enabled plugins.
+		
+		// A Fluent file located at
+		//   [plugin root]/locale/en-US/make-it-red/strings.ftl
+		// could be included in an XHTML file as
+		//   <link rel="localization" href="make-it-red/strings.ftl"/>
```
dstillman, line 94: Zotero uses the extra "zotero" subdirectory ("make-it-red" here), but that's just a historical artifact — we wouldn't want to recommend that for plugins. So:

1) `make-it-red/strings.ftl` → `make-it-red.ftl`

2) What happens without the subdirectory? If you just use `href="make-it-red.ftl"`, will it work as long as there's no conflicting file in another `L10nFileSource`?

dstillman: Sure, that'd be great.

PR #2864, opened 2022-10-10
Fix last editpane tab being misaligned vertically on macOS

AbeJellinek:
Before:

After:

(Both upscaled.) The rightmost tab was pushed too far up before and this change corrects its position. Can be cherry-picked to master too.


dstillman:
FWIW, on a Retina MBP the tabs were aligned, but the text was too low:

tabs

On a non-Retina display, I was getting the same misalignment.

PR #2868, opened 2022-10-13
Add "Open PDF reader in new window" preference

### LINKED ISSUE #2680: Option to open PDFs in a new window by default
https://forums.zotero.org/discussion/98153/open-pdf-in-new-window-as-default-possible

Seems reasonable as an option once we have room for PDF reader options.

AbeJellinek:
When enabled:
- Double-clicking a PDF or choosing "Open PDF" opens a new window
- Shift-double-clicking opens a new tab
- "Open in New Window" locate option becomes "Open in New Tab" and has the reverse behavior

Fixes #2680

### DIFF
--- a/chrome/content/zotero/locateMenu.js
+++ b/chrome/content/zotero/locateMenu.js
@@ -123,17 +123,17 @@ var Zotero_LocateMenu = new function() {
 	});
 	
 	function _addViewOption(selectedItems, optionName, optionObject, showIcons) {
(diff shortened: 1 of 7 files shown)


dstillman:
I changed the text to "Open PDFs in new windows instead of tabs", which is a bit verbose but makes it more explicit. We could say just "Open PDFs in new windows" if we think that's unnecessary. (The Firefox pref is "Open links in tabs instead of new windows".)

PR #2869, opened 2022-10-13
Allow setting default view settings for PDF reader

### LINKED ISSUE #2648: Allow to set default PDF reader view settings
Add "View → Reset to Defaults" and "View → Set as Defaults" options. Similar to what macOS Finder does.

https://forums.zotero.org/discussion/95215/default-pdf-view-mode-and-color-preset-in-zotero-6
https://forums.zotero.org/discussion/101044/default-pdf-view-settings
https://forums.zotero.org/discussion/119956/default-epub-view-scrolled-paginated

mrtcode:
Added "Set as Defaults" option that stores view settings of the options above (scale, scroll mode, spread mode).

Screenshot 2022-10-13 at 17 49 25

- Numerical zoom values fall back to "Zoom to Page Width", because there is no point to save them.
- "Hand Tool" is not included because it's rarely needed and feels like a "bug" when using.
- Split view would probably be too much to support as well. But worth to consider in future. Although it's more like a tab state.
- We were also discussing about "Reset to Defaults" option, but it can be added separately if there would be a need. Maybe it could just be called "Reset".

Fixes #2648.


defaults/preferences/zotero.js
```diff
@@ -208,5 +208,8 @@ pref("extensions.zotero.annotations.noteTemplates.title", "<h1>{{title}}<br/>({{
 pref("extensions.zotero.annotations.noteTemplates.highlight", "<p>{{highlight}} {{citation}} {{comment}}</p>");
 pref("extensions.zotero.annotations.noteTemplates.note", "<p>{{citation}} {{comment}}</p>");
 
+// PDF Reader
+pref("extensions.zotero.pdfReader.defaultState", '{"scale":"page-width","scrollMode":0,"spreadMode":0}');
```
qqobb, line 212: Thanks for implementing default view settings! It's nice that you'll be able to change the default, but I think the initial default should be "auto". In many cases, "auto" (Ctrl/Cmd-0) is just the same as "page-width". However, in some cases "auto" is just the better default option, in my opinion. Two examples:

1. Landscape documents, see https://github.com/mozilla/pdf.js/pull/5229. If you view the PDF file mentioned there, e.g., on a 16:9 screen in Firefox, it will look much better in "auto" than in "page-width" mode:

> Landscape documents, like slide presentations, are much easier to read if "Automatic Zoom" fits the slide height so the full slide is visible. Here is an example landscape slide presentation:
> 
> https://wiki.mozilla.org/images/5/55/MobileOpportunity.pdf

2. On larger monitors, the "auto" setting limits the zoom level to something reasonable. I'd guess, especially on ultra wide monitors, the PDF could be zoomed to a ridiculously large size in "page-width" mode. Two related discussions on the forums are [here](https://forums.zotero.org/discussion/93542/feature-request-automatically-resize-option) and [here](https://forums.zotero.org/discussion/comment/394624/#Comment_394624).

In most cases, "auto" will give the best user experience, I think. It also seems that "auto" is the default in Firefox.

qqobb: I had a look at what "auto" does, see [here](https://github.com/mozilla/pdf.js/blob/8629a55215f88e406c304ccd18dd4ac54f44d9ac/web/pdf_viewer.js#L1135-L1189). If you think that the current "auto" scale is sometimes too small, you could consider changing `MAX_AUTO_SCALE`, perhaps even making it another hidden preference. The current value is 1.25, but people might prefer different maximum zoom levels in "auto" scale. (Note that this is different from a fixed numerical zoom level.)

dstillman:
Rebased for Z7

dstillman:
Wait, so the pref is only used if there's not a state file saved for the PDF, and a state file is saved as soon as the PDF is open? So existing users wouldn't be able to use these defaults for any files they'd previously opened? And even after this landed, if you changed the defaults, the new defaults wouldn't apply to any files you had opened previously?

That wouldn't work, so we'll need to do something different here. Even a "Reset to Defaults" would be really annoying if you had to run it in every file. A "Reset All to Defaults…" with a confirmation prompt that scanned for all .zotero-pdf-state files would be an option, but that wouldn't work as it is, because we can't delete the state file altogether, and the defaults aren't used if there's a state file at all.

So I think we need to 1) make it use the default state if those specific settings aren't in the state file, not just if the state file doesn't exist, 2) not save those specific settings to the state file if they haven't changed from the current defaults, and 3) add a reset-all option that cleared those settings from all state files.

And maybe there's some optimization for resetting all? Store a timestamp of the last reset-all time, and any state file with an mtime from before then ignores these settings on load (and clears them on save)?

mrtcode:
> Wait, so the pref is only used if there's not a state file saved for the PDF, and a state file is saved as soon as the PDF is open? So existing users wouldn't be able to use these defaults for any files they'd previously opened? And even after this landed, if you changed the defaults, the new defaults wouldn't apply to any files you had opened previously?
> 
> That wouldn't work, so we'll need to do something different here. Even a "Reset to Defaults" would be really annoying if you had to run it in every file. A "Reset All to Defaults…" with a confirmation prompt that scanned for all .zotero-pdf-state files would be an option, but that wouldn't work as it is, because we can't delete the state file altogether, and the defaults aren't used if there's a state file at all.
> 
> So I think we need to 1) make it use the default state if those specific settings aren't in the state file, not just if the state file doesn't exist, 2) not save those specific settings to the state file if they haven't changed from the current defaults, and 3) add a reset-all option that cleared those settings from all state files.
> 
> And maybe there's some optimization for resetting all? Store a timestamp of the last reset-all time, and any state file with an mtime from before then ignores these settings on load (and clears them on save)?

I agree with everything. Working on a fix.

mrtcode:
Things that were improved:

- Only saves settings that don't exist in the default state.
- Uses individual state properties from the defaults if they are missing.
- Adds "Reset All to Defaults…" option with a confirmation dialog. It stores a timestam in `pdfReader.stateResetTime` pref and replaces the saved state settings with the defaults when loading the PDF file. It also updates the currently loaded PDF reader instances.
- The reset currently doesn't affect the second view in split mode, but we'll fix that in the newer reader version.

andrewboldi:
Status on merging this PR?

Issue #2622, opened 2022-05-17
Link Mendeley citations in documents to imported items


dstillman:
From [Mendeley Import](https://www.zotero.org/support/kb/mendeley_import):

> When using the Zotero word processor plugins, document citations created with Mendeley won’t currently be linked to imported citations in your Zotero database. Zotero’s word processor plugins can, however, read Mendeley citations and their embedded metadata, so you can continue using the same documents with Zotero and insert additional instances of those citations by choosing from the “Cited” section of the search results in the citation dialog.

We store Mendeley item UUIDs for imported items. I haven't checked what metadata Mendeley stores in field codes, but if they're using the UUID, it would be nice to automatically use the imported item, both for metadata edits and for disambiguation (e.g., so that adding the imported item from the library itself wouldn't add a duplicate bibliography entry).

We could also consider just rewriting the item as a Zotero citation on refresh, which would probably be easier, but that might cause problems for people trying to collaborate with other Mendeley users.

adomasven:
I've downloaded Mendeley Desktop for Linux, inserted a citation into a document, imported the Mendeley library into Zotero, and checked the UUID that Mendeley Desktop puts in a LibreOffice document citation with Zotero, but I am not getting any items from ` Zotero.Relations.getByPredicateAndObject('item', 'mendeleyDB:documentUUID', UUID)`. I've also tried `remoteDocumentUUID`. @dstillman @tnajdek thoughts?

adomasven:
Ok, so if I use a local mendeley import, I can get the related items. Given that this is not the recommended method to import, and the online import has different UUIDs than what Mendeley stores in documents, I don't think this can be done easily.

We could use Zotero's duplicate detection code to search for Mendeley imported items and link them that way in citations where we detect Mendeley's URIs, but that's quite more complicated than just linking URIs.

dstillman:
Ugh. What about two synced Mendeley Desktop installs, though? How does the Mendeley Word plugin identify those items?

adomasven:
Hmm, that's a good question, actually. I originally had the latest Mendeley Desktop installed, inserted some citations, then reinstalled Mendeley Desktop 1.18, synced with it, imported into Zotero, and the UUIDs matched, so they have to be stored in the remote repo somewhere. Our import is not storing them when doing an online import then, it seems? @tnajdek ?

tnajdek:
[...]

PR #2871, opened 2022-10-17
Mendeley Import: Switch to direct login/password auth in order to obtain Mendeley Desktop item id

tnajdek:
As discussed, this approach pretends to be Mendeley Desktop and performs login directly using username and password obtained from the user via form. Signing in this way ensures that Mendeley servers include extra key for documents (`client_data`) which contains Mendeley Desktop UUID. Storing this UUID makes #2622 possible.

### DIFF
--- a/chrome/content/zotero/fileInterface.js
+++ b/chrome/content/zotero/fileInterface.js
@@ -444,10 +444,12 @@ var Zotero_File_Interface = new function() {
 		
 		var translation;
 		
-		if (options.mendeleyCode) {
+		if (options.mendeleyAuth || options.mendeleyCode) {
 			translation = yield _getMendeleyTranslation();
 			translation.createNewCollection = createNewCollection;
+			translation.mendeleyAuth = options.mendeleyAuth;
 			translation.mendeleyCode = options.mendeleyCode;
+			translation.newItemsOnly = options.newItemsOnly;
 		}
 		else {
 			// Check if the file is an SQLite database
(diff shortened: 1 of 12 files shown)


tnajdek:
@dstillman I've updated this PR, so that:

* it's now possible to switch back to OAuth flow with a `import.mendeleyUseOAuth` preference
* It's now possible to only import new items on re-import. Option only appears if at least one item with relation `mendeleyDB:documentUUID` or `mendeleyDB:remoteDocumentUUID` exists.
* importer will now update IDs on previously imported items to Mendeley Desktop IDs, this happens regardless of "Download new items only" setting

Caveats:

* There are couple of new strings so, in current shape, this only works in en_US
* When merged, it will need to be separately ported for `fx-102`. #2862

tnajdek:
I've just realized we should probably provide feedback when username/password is incorrect and let user re-enter these details. Currently it will just go ahead and instantly report 0 items imported. I'll fix that.

tnajdek:
@dstillman I've updated the PR with the following changes:

* "Download new items..." now checked by default
* Fixed number of items reported on partial re-import
* Remove collections created on re-import if 0 items imported
* Provide feedback on sign-in (moved auth logic from importer to wizard)
* Fix #2881

tnajdek:
Sorry, this needs more attention (something is not working as expected when trying with my large test library import)

dstillman:
OK, I think we're good. Can you squash these a bit and then we'll merge?

tnajdek:
OK, I've squashed everything, including your most-recent tweaks, into a couple of commits.

Also I've found one more issue where checking for previously imported items could throw an exception so I've fixed it [here](https://github.com/tnajdek/zotero/blob/0e121264360bfee7fa84c51ab12ad5fdaa1e1624/chrome/content/zotero/import/importWizard.js#L28-L33).

PR #2878, opened 2022-10-21
Adding toolbar button actions to the menubar.

jaeschrich:
I added most of the toolbarbutton actions to the menubar. Right now, the menu items are copied-and-pasted from the toolbarbutton popup menus. I can refactor this to avoid duplication in one of two ways: (1) by creating an XBL binding that encapsulates the shared menu items, or (2) by creating a JavaScript function that sets up both the menubar popup and the toolbarbutton popup identically. I'm not sure which method of encapsulating the shared functionality is more idiomatic for the codebase, or which method will be easier to port away from XUL later.

The only toolbarbutton whose functionality I didn't add to the menubar is the sync button, because I'm not sure how best to expose that functionality via a menu. Right now the information about the sync status is displayed via a popup and there's a cancel button which is dynamically added to the toolbar to cancel a sync in progress. I don't think that adapting this behavior directly to the menubar makes sense. I could create a popup window for the menubar sync functionality, but that seemed beyond the scope of an individual pull request.


dstillman:
This seems to include an outdated version of the code from the other PR? Unless this depends on that PR, in which case you should just push the new commit to that one, this PR should only include the relevant commits.

dstillman:
> I can refactor this to avoid duplication in one of two ways: (1) by creating an XBL binding that encapsulates the shared menu items, or (2) by creating a JavaScript function that sets up both the menubar popup and the toolbarbutton popup identically.

(2) for sure. XBL is discontinued in fx102.

For sync, I would add a Sync Now option to the Tools menu (like Firefox) and not worry about cancel or status for the moment.

dstillman:
Location tweaks:

- Add Attachment and Add Item by Identifier should go in File, below a separator below New*
- Locate can go in File, since it's about the selected item (i.e., the "file") — View is really about the UI, not viewing things externally
- Sync Now can go at the top of Tools, in its own section

Other things:

- At least on macOS, Locate isn't being fully populated based on the current item selection. Crossref Lookup and Google Scholar Search appear, but not Library Lookup or the View options above. If a note is selected, it says "No items selected", so it's at least partially populating it.
- In a read-only library, the New Note menuitem itself isn't disabled, like New Item. (Only the child items are disabled.) Same for Add Attachment and Add Item by Identifier. (The former has disabled child items, but the latter actually opens the panel.)

jaeschrich:
Hey Dan,
I moved the menuitems as requested and added a "Sync Now" button. I also wrote some code to keep the menu items in sync with the toolbarbuttons. I fixed the bug with the menuitems not getting disabled properly in read-only libraries. 

I can't reproduce the bug with the "locate" menu not being populated. I tried it on both windows and linux, and it works on both. It's possible that the new code to dynamically populate the menuitems will make it work on Mac; otherwise, I'm not sure what else to try.

PR #2887, opened 2022-11-01
Fix a typo in the de locale

duinomaker:
For "book selection", change "Buchteil" to "Bruchteil".


dstillman:
I don't think that's a typo. "Book" in German is "buch". According to Google Translate, "buchteil" means "book part" and "bruchteil" means "fraction".

See also https://www.dict.cc/deutsch-englisch/buchteil.html:

> Buchteil {m} "part of a book, such as a chapter, a section or a range of pages"

PR #2892, opened 2022-11-02
Fixed double-quoted string literals in SQL files

### LINKED ISSUE #2891: Double-quoted string literals in sql scripts.
Two sql scripts (system-107.sql system.sql) contain a number of double-quoted string literals.

pchemguy:
Replaced double quotes around string literals with single quotes.

Fixes #2891


dstillman:
(There are valid in SQLite, so this makes no difference in Zotero.)

PR #2893, opened 2022-11-03
Add Copy As menu with link options

### LINKED ISSUE #2889: Provide ways to copy zotero://select and zotero://open-pdf URIs for items and annotations
https://forums.zotero.org/discussion/comment/421035/#Comment_421035

Need to think about the best ways to do this

### LINKED ISSUE #2917: Add function to get web library URL for item
https://forums.zotero.org/discussion/101169/zotero-http-links-result-in-page-not-found

AbeJellinek:
Fixes #2889, fixes #2917, needs https://github.com/zotero/reader/pull/208


mrtcode:
`pdf-reader` context menus will be completely rewritten soon. Context menu data will be generated on `pdf-reader` side and then passed back to the client side to display in a native popup.

So, I guess, we should wait before implementing this.

tnajdek:
The fix for zotero/web-library#253 is now live, and linking to an item should now correctly scroll the list.

dstillman:
https://forums.zotero.org/discussion/129413/request-expose-the-existing-item-link-zotero-select-in-the-context-menu

PR #2897, opened 2022-11-07
Add "Add to Dictionary" to note editor context menu

### LINKED ISSUE #1390: Enhancement: Add words to spell checking dictionary from Notes editor dialog.
When editing a note, wavy red underlines are provided for words that seem to fail spell checking. Right-clicking the word provides a context menu. However, the context menu does not provide an option to add the word to the spell checking dictionary. It would be a useful enhancement to provide this feature.

AbeJellinek:
As well as "Undo Add to Dictionary", as in Firefox.

Fixes #1390

### DIFF
--- a/chrome/content/zotero/elements/noteEditor.js
+++ b/chrome/content/zotero/elements/noteEditor.js
@@ -60,6 +60,8 @@
 			window.addEventListener("unload", this.destroy);
 
 			var shadow = this.attachShadow({ mode: "open" });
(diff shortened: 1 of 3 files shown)


dstillman:
Nice. There's no way to undo after restarting (or maybe switching to another note)? Only option is to edit persdict.dat in your profile folder?

dstillman:
Eh, it's OK. We can always create a KB article.

dstillman:
Probably not necessary. We don't actually have the strings from Mozilla, I don't think. (Or maybe we have the English ones built in.)

PR #2902, opened 2022-11-08
Add "Any Field" advanced search condition

### LINKED ISSUE #1699: Search condition for all fields
https://forums.zotero.org/discussion/77510/search-every-field-in-advanced-saved-searches

Seems like this should be a dupe, since this is a common request, but I didn't find another issue for it.

AbeJellinek:
I'm not hugely enthused by how I ended up having to implement this - we didn't seem to have a good way to add compound conditions like "Any Field" that expand at query time instead of when initially being added, so I had to tack something onto the main loop in `_buildQuery()`. If there's a better way that I missed, I'll change it!

Fixes #1699

### DIFF
--- a/chrome/content/zotero/elements/zoteroSearch.js
+++ b/chrome/content/zotero/elements/zoteroSearch.js
@@ -354,6 +354,7 @@
 
 		isPrimaryCondition(condition) {
 			switch (condition) {
+				case 'anyField':
 				case 'collection':
 				case 'creator':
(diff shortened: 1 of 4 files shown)


dstillman:
This is great. Let's put "Any Field" at the top of the main condition menu, not in More. (It would sort first in English anyway, but we should put it first in all languages.)

dstillman:
Yep

PR #2916, opened 2022-11-16
Tag autocomplete fuzzy matches on both sides of input

hammeiam:
Problem: Tags currently only autocomplete based on a fuzzy match *after* the user's input, eg `SQL("INPUT%")`. This leads to users needing to do a lot of unnecessary typing in order to find their desired tag. This is further compounded by the fact that many users employ an information hierarchy in their tags eg `Location: USA: Ohio: Cleveland`. This means that a user wastes a lot of time typing because they can't just autocomplete "Cleveland" to the appropriate set of tags.  

Solution: Simply fuzzy match on both sides of the user input, eg `SQL("%INPUT%")`. Per dstillman's feedback, this change is behind the user-set flag `unboundedTagSearch`, which defaults to `false`. 

Screen Shot 2022-11-15 at 4 50 12 PM


dstillman:
I mean, it's not a bug — it's left-bound because that's an indexed search, and otherwise it's not.

We'll need to do some performance tests in large libraries to see how it performs non-left-bound on modern computers. We likely haven't tested this in years, but historically searches were slow in large libraries with many tags. And if that's still the case, we'd only do this either 1) as a pref, defaulting off or 2) if we added a separate word index on tags.

quinn-p-mchugh:
> as a pref, defaulting off

I would love to see this implemented as a preference, defaulting to off. Let the users decide whether the performance hit is acceptable.

jankap:
Would love to see this search happening :)

hammeiam:
@dstillman I put this change behind a user flag so people can turn it on if they choose.

quinn-p-mchugh:
Can this be merged?

hammeiam:
@quinn-p-mchugh @dstillman  rebased changes against latest `main`

hammeiam:
@dstillman Congrats on the great looking beta launch! I think it is the perfect opportunity to add this quick enhancement :)

rossant:
I agree this would be a great enhancement, I hope it will be merged soon :)

hammeiam:
Hi @dstillman - just removed conflicts, ready to merge whenever you are :)

melsophos:
Is there any news concerning this PR? Since @hammeiam seems to have implemented everything requested and since the option is not enabled by default, what prevents merging it?

hammeiam:
Hi @dstillman - I've had this PR open for like 3 years now. It's pretty much the minimum changeset that it needs to be, and is behind a user flag with the default set to disabled. I've seen a few other subsequent PRs that implement this same feature, and the comments show that users want this. 
Please let me know if I can do anything else to get this merged.

vinieich:
This feature would be really useful.. Is there any way to call attention about this issue to devs who could do the merge?

rossant:
It looks like there are conflicts now.

hammeiam:
Three and a half years later I think this feature is finally in Zotero 8!
https://www.zotero.org/blog/zotero-8/#tag-autocomplete-and-note-field-in-zotero-connector 

Edit: Just kidding, v8 still doesn't have fuzzy tag search

PR #2930, opened 2022-11-23
fx-compat: Fix ProgressWindow and feed Add To button

AbeJellinek:


chrome/content/zotero/progressWindow.xhtml
```diff
@@ -38,10 +38,9 @@
         windowtype="alert:alert">
 		
 	<vbox id="zotero-progress-text-box" flex="1" style="padding: 10px">
-		<hbox align="stretch">
+		<hbox id="zotero-progress-text-headline" align="stretch">
 			<label id="progress-label" flex="1"/>
 		</hbox>
-		<html:progress id="progress-indicator" max="100"/>
```
adomasven, line 44: Doesn't this remove the actual progress indicator? I've done a pass over this code, mainly because I was changing similar code somewhere else when working on word processor compatibility, so just wondering.

adomasven: Ah yeah, it is a mixup. Word processor integration code doesn't use `#progress-indicator` or this file for that matter.

dstillman:
Merged in 730677a918723dea873cb253dfe233f451d2fd1f and 53ace4876e90d1e2301cae3b99b31610ae093cbc

PR #2933, opened 2022-11-24
fx-compat: Make color picker CE accessible via keyboard and VoiceOver

### LINKED ISSUE #2927: Make tag colors keyboard accessible
After #2682 the colorpicker menubutton gets focus and can be opened with space bar, but the colors themselves can't be navigated to via the arrow keys.

AbeJellinek:
Replaces `<div>`s with semantically-correct `<button>`s, adds labels to the color swatches, and enables keyboard navigation inside the color picker.

Arrow key behavior approximates but doesn't quite stick to that of VoiceOver - I added support for wrapping. Left arrow on the top-left square goes to the bottom-right square, up arrow on the top row goes to the same column in the bottom row, and so on.

The color names I chose are a bit subjective.

Fixes #2927

### DIFF
--- a/chrome/content/zotero/elements/colorPicker.js
+++ b/chrome/content/zotero/elements/colorPicker.js
@@ -138,15 +138,34 @@
 
 		init() {
 			let button = this.querySelector('.button');
-			let panel = this.querySelector('.panel');
-			let grid = this.querySelector('.grid');
+
+			button.addEventListener('keydown', (event) => {
+				if (event.key == ' ' || event.key == 'Enter' || event.key == 'ArrowDown') {
+					event.preventDefault();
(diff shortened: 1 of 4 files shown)


dstillman:
(Rebased)

A couple things:

1) If I press space bar, the colorpicker opens, but I can't then use the arrow keys to get to it — I have to tab to it, in contrast to, e.g., the toolbar menubuttons, where down-arrow takes you to the first menu option.

2) The focus ring is going half into the next swatch for me:

Screenshot 2023-04-13 at 12 43 03 AM

Oh, and I guess the shape of the swatches is different between the selected color and the 9 squares, and the focus ring is the shape from the former.

Issue #1810, opened 2020-04-06
[Windows] Item type is changed when cycling through options via keyboard


dstillman:
https://forums.zotero.org/discussion/comment/352770/#Comment_352770

It should work like on a Mac.

(Didn't test on Linux)

dstillman:
Same thing happens if you use the arrow keys, so that's no good.

tnajdek:
Firstly the exact behaviour described in the linked forum post, i.e. pressing a letter after focusing item type selector (without opening it) changes the item type immediately, is consistent on all platforms including macOS.

However pressing an arrow key opens the menu on macOS, but on Windows (and Linux) it immediately changes to the next value. I looked into it and found that on Windows this is the typical behaviour. Great example is a screen resolution picker in desktop settings where it immediately applies new resolution (quite bizarre!). 

So in all places we use `<menulist>`  it fires `oncommand` when arrow keys are pressed on Windows and Linux.

So the first question is: whether we actually want to mess around with native control and change it so it behaves more like macOS?

I actually experimented with it and since we cannot depend on `oncommand` I thought we could listen to `popuphiding` and apply the value then. However I couldn't figure out how to detect whether popup was closed in a "cancelling"  (e.g. by pressing ESC) or confirming way. I didn't want to spend too much time on it but I believe this is a dead end approach. An alternative I didn't pursue yet is to to use `<menupopup>` without `<menulist`> but then we will need to render some of the visuals to make it look like a select box.

PR #2946, opened 2022-12-13
Unify behaviour of item type menu on all platforms

tnajdek:
This PR is a fix for #1810. Item type menu is tweaked to resolve the following issues:

* [All platforms] With focus on the selector, but menu collapsed, pressing a letter key would trigger immediate change to a value beginning with that letter
* [Windows and Linux] Not possible to expand the menu with keyboard (now up/down arrow and spacebar opens the popup)
* [Windows and Linux] Arrow keys immediately change value (now expand the menu)

This should make the behaviour unified across platforms.


tnajdek:
This approach won't work, see my comment in #1810

PR #2951, opened 2022-12-23
Add `/connector/request` endpoint

AbeJellinek:
This allows HTTP requests from the Connector to be routed through the client, once that functionality is implemented in `zotero/translate` and `zotero/zotero-connectors` (that's next).

Also updates `Zotero.HTTP.request()` to use the Fetch API's `Headers` class so that operations on the headers are case insensitive.

### DIFF
--- a/chrome/content/zotero/xpcom/connector/server_connector.js
+++ b/chrome/content/zotero/xpcom/connector/server_connector.js
@@ -1843,3 +1843,93 @@ Zotero.Server.Connector.IEHack.prototype = {
 			'</head><body></body></html>');
 	}
 }
+
+/**
+ * Make an HTTP request from the client. Accepts {@link Zotero.HTTP.request} options and returns a minimal response
+ * object with the same form as the one returned from {@link Zotero.Utilities.Translate#request}.
+ *
+ * Accepts:
+ *		method - The request method ('GET', 'POST', etc.)
+ *		url - The URL to make the request to. Must be an absolute HTTP(S) URL.
+ *		options - See Zotero.HTTP.request() documentation. Differences:
+ *			- responseType is always set to 'text'
+ *			- successCodes is always set to false (non-2xx status codes will not trigger an error)
+ * Returns:
+ *		Response code is always 200. Body contains:
+ *			status - The response status code, as a number
+ *			headers - An object mapping header names to values
+ *			body - The response body, as a string
+ */
+Zotero.Server.Connector.Request = function () {};
+Zotero.Server.Endpoints["/connector/request"] = Zotero.Server.Connector.Request;
+Zotero.Server.Connector.Request.prototype = {
+	supportedMethods: ["POST"],
+	supportedDataTypes: ["application/json"],
+
+	/**
+	 * For testing: allow disabling URL validation so we can make requests to the server.
+	 */
+	validateURLs: true,
+
+	init: async function (req) {
+		let { method, url, options } = req.data;
+		
+		if (typeof method !== 'string' || typeof url !== 'string') {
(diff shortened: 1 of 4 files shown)


chrome/content/zotero/xpcom/connector/server_connector.js
```diff
@@ -1843,3 +1843,93 @@ Zotero.Server.Connector.IEHack.prototype = {
...
+			return [400, 'text/plain', 'Unsupported scheme'];
+		}
+
+		// Through heuristics, attempt to prevent:
+		//   - recursive requests to this endpoint, which will deadlock the server until they time out
+		//   - requests to the local network, which shouldn't be necessary from the connector and pose a security risk
+		if (this.validateURLs
+			&& (uri.port == Zotero.Prefs.get('httpServer.port')
+				|| uri.filePath == '/connector/request'
+				|| Services.io.hostnameIsLocalIPAddress(uri))) {
```
dstillman, line 1899: Seems like all of this could just be replaced with blocking `localhost` or any IP address (which would include local addresses and loopback), but I would even say we should limit this to just the domain(s) we need for now. We've never needed this before, and hopefully won't need it for anything else, so I'd rather not unnecessarily expose an open HTTP relay. We should also check explicitly that User-Agent starts with `Mozilla/`, which will guarantee that it's a request with a connector header (since that's checked in server.js).

Basically, if Zotero is specifically given network access on an otherwise locked-down system, we don't want it to be usable for a privilege escalation attack.

dstillman: (I guess checking for the path would prevent a recursive request if the computer somehow had a non-local address, though that would be pretty rare. In any case, I think better just to allow specific domains.)

chrome/content/zotero/xpcom/connector/server_connector.js
```diff
@@ -1896,13 +1896,18 @@ Zotero.Server.Connector.Request.prototype = {
...
+		if (Zotero.Server.Connector.Request.enableValidation) {
+			if (!Zotero.Server.Connector.Request.allowedHosts.includes(uri.host)) {
+				return [
+					400,
+					'text/plain',
+					`Unsupported URL: host was ${uri.host}, expected one of [${Zotero.Server.Connector.Request.allowedHosts.join(', ')}]`
+				];
+			}
+
+			if (req.headers['User-Agent'] && !req.headers['User-Agent'].startsWith('Mozilla/')) {
```
dstillman, line 1908: Should be `!req.headers['User-Agent'] || !req.headers['User-Agent'].startsWith('Mozilla/')`

chrome/content/zotero/xpcom/connector/server_connector.js
```diff
@@ -1896,13 +1896,18 @@ Zotero.Server.Connector.Request.prototype = {
...
-				400,
-				'text/plain',
-				`Unsupported URL: host was ${uri.host}, expected one of [${Zotero.Server.Connector.Request.allowedHosts.join(', ')}]`
-			];
+		if (Zotero.Server.Connector.Request.enableValidation) {
+			if (!Zotero.Server.Connector.Request.allowedHosts.includes(uri.host)) {
+				return [
+					400,
+					'text/plain',
+					`Unsupported URL: host was ${uri.host}, expected one of [${Zotero.Server.Connector.Request.allowedHosts.join(', ')}]`
```
dstillman, line 1904: I don't think there's any reason to echo these back.

dstillman:
I'll cherry-pick to master as well. We can get this into 6.0.19.

dstillman:
Out now in the beta

PR #2961, opened 2023-01-05
LABD file relinking: Process using Unix paths

### LINKED ISSUE #2960: LABD file relinking not working when moving from macOS/Linux to Windows
https://forums.zotero.org/discussion/comment/425382/#Comment_425382

> (3)(+0000010): Does not exist: C:\Users\[user]\[…]\Bibliography\/home/[user]/[…]/Bibliography/[…].pdf
>
> (3)(+0000001): No segments left to drop; match not found in LABD
> 
> (1)(+0002099): NS_ERROR_FILE_UNRECOGNIZED_PATH Exception: Component returned failure code: 0x80520001 (NS_ERROR_FILE_UNRECOGNIZED_PATH) [nsIFile.initWithPath]

AbeJellinek:
Not adding a test for this because it would only give meaningful results when run under Windows, which I can't currently do and which CI doesn't do. Tested manually in a Windows VM, of course.

Fixes #2960

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -5456,6 +5456,11 @@ var ZoteroPane = new function()
 	 * @return {Promise<Boolean>} True if relinked successfully or canceled
 	 */
 	this.checkForLinkedFilesToRelink = async function (item) {
+		// Naive split and join implementations that split on any separator and join using forward slashes
+		// OS.Path methods have different behavior depending on the platform, and a naive approach is good enough here
+		let split = path => path.split(/[/\\]/);
+		let join = (...segments) => segments.join('/');
+		
 		Zotero.debug('Attempting to relink automatically');
 		
 		let basePath = Zotero.Prefs.get('baseAttachmentPath');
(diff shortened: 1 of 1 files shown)


chrome/content/zotero/zoteroPane.js
```diff
@@ -5490,15 +5496,20 @@ var ZoteroPane = new function()
...
 			if (!(await OS.File.exists(correctedPath))) {
 				Zotero.debug('Does not exist: ' + correctedPath);
 				continue;
 			}
 			Zotero.debug('Exists! ' + correctedPath);
+			
+			if (Zotero.isWin) {
+				correctedPath = correctedPath.replace(/\//g, '\\');
+				Zotero.debug('Converted back to Windows path: ' + correctedPath);
+			}
```
dstillman, line 5512: Why here and not above? Doesn't this mean we're now passing a path with forward slashes to `OS.File.exists()` on Windows? Does that work?

dstillman: Huh!

PR #2964, opened 2023-01-07
Add Open button to Scaffold - Browser

yfdyh000:
This avoids hand moving to keyboard to press Enter key if I use the mouse to select or paste a URL.


yfdyh000:
> I don't especially care for this UI - if we're going to add a button, I'd prefer it to be less obtrusive and placed inside the URL field, like in Firefox and other browsers.

I thought and recognized this idea, I would try to reduce it further.

chrome/content/scaffold/scaffold.js
```diff
@@ -1780,7 +1790,7 @@ var Scaffold = new function () {
 	 * the system's default browser).
 	 * @param {boolean} openExternally whether to open in the default browser
 	**/
-	this.openURL = function (openExternally) {
+	this.openURLInTests = function (openExternally) {
```
dstillman, line 1793: "openURLInTests" is a little confusing. Something like `openURLFromSelectedTest()` and `openURLFromURLField()` would be better.

chrome/content/scaffold/scaffold.xul
```diff
@@ -378,6 +378,7 @@
 								<menulist id="browser-url" editable="true" flex="1">
 									<menupopup></menupopup>
 								</menulist>
+								<button id="browser-open" label="&scaffold.browser.open;" tooltiptext="Open url in build-in browser" oncommand="Scaffold.openURLField()"/>
```
dstillman, line 381: Would be "built-in", but the tooltip can just be "Open URL" here.

chrome/content/scaffold/scaffold.js
```diff
@@ -1577,6 +1577,16 @@ var Scaffold = new function () {
 		return str + "\n}";
 	}
 	
+	// open the URL from the URL field
+	this.openURLField = async function () {
```
dstillman, line 1581: Doesn't need to be async

dstillman (commented):
This still won't do. This is what it looks like on macOS:

Screenshot 2023-01-09 at 6 42 45 PM

AbeJellinek:
Thanks, this was implemented separately.

PR #2965, opened 2023-01-07
Show fieldLocalizedName in Scaffold - List All Fields

yfdyh000:
ref: https://forums.zotero.org/discussion/101997/suggestion-show-the-localized-field-labels-in-list-all-fields-for-item-type


adam3smith:
This makes sense to me for people working on non-English locales and I don't think it's in the way at all, so +1 from me

PR #2966, opened 2023-01-07
refine itemtype changes

yfdyh000:
ref: https://forums.zotero.org/discussion/102009/ux-about-changeable-item-type

I think this is correct.

This may lack unit testing, and date & bookSection handle.

zotero


dstillman:
As [explained in the forums](https://forums.zotero.org/discussion/comment/424985/#Comment_424985), this isn't necessarily desirable.

We're hoping to add additional fields soon. That will be a better solution for this and make this behavior strictly undesirable, so I'm going to close this.

PR #2967, opened 2023-01-07
Fix build & watch

yfdyh000:
Fix scss is always build while watch.
Fix some files are not updated while build / watch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "scripts": {
     "start": "node ./scripts/build.js && node ./scripts/watch.js",
     "build": "node ./scripts/build.js",
+    "watch": "node ./scripts/watch.js",
     "sass": "node ./scripts/sass.js",
     "js": "node ./scripts/js.js",
     "clean": "node ./scripts/clean.js",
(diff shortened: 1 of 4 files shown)


dstillman:
Could you explain the exact issue here that this is fixing? What are steps to reproduce the problem?

yfdyh000:
I see the three scss files always updated (in logs) while `npm run watch`.

After successfully build, I see changed files not be copied to the `zotero-client\build` directory with `bash ../zotero-standalone-build/scripts/build_and_run -r -d`, so changes are not effective.
Maybe I missed a certain setting? in Windows 10, with Cygwin. ref: https://www.zotero.org/support/dev/client_coding/building_the_standalone_client

dstillman:
But `npm run watch` isn't a command — you added it. The command to use file-watching is `npm start`. Again, I'd just strongly encourage you to ask questions first instead of immediately submitting PRs — there's obviously file-watching code, so it stands to reason we have an existing way of running it.

But you don't actually need to use `npm start`, and I personally don't usually use it. `build_and_run` automatically does an `npm run build`, which is very quick if there are no changes. Using `npm start` is slightly faster, but it's easy for a JS parsing error to go unnoticed in a background terminal window, and then you wonder why no changes are showing up. With `build_and_run`, the build just fails.

dstillman:
Closing without further info on an existing bug

PR #2969, opened 2023-01-07
Add Collections column to item list

yfdyh000:
collections count as text, list as tooltip

ref: https://forums.zotero.org/discussion/102008/ux-suggestion-collection-column

`i('TreeitemNoteSmall', 'chrome://zotero/skin/treeitem-note-small.png');` changed to 2x, otherwise the image is 12px, but the context requires 16px, the result is fuzzy. More lift may be needed.


dstillman:
Thanks, but we don't think this really adds much value. If you want to see items that aren't in any collection, you can already do that via the special Unfiled Items collection.

dstillman:
While we greatly appreciate the contributions, in general I'd encourage you to wait for feedback in the forums before creating PRs. We read every thread, and if we think something is a good idea, we'll generally say so and create an issue here as an indication that it's something we want to do. And if we don't respond, it's usually an indication that either we're not interested or we're waiting for additional feedback before deciding whether to implement it.

If you're making changes for yourself in a custom build and you want to submit PRs as well, you can — just be aware that many/most of them may not be accepted. (If it's something we clearly did on purpose, like hiding the item-type-menu dropmarker, a PR isn't welcome without prior approval.)

If you're interested in working on things we've already committed to doing, there are hundreds of issues here that you can browse through, including some [Papercuts](https://github.com/zotero/zotero/issues?q=is%3Aopen+is%3Aissue+label%3APapercuts) issues that could have a large impact relative to their difficulty.

PR #2975, opened 2023-01-11
Refactor l10n code. Remove use of react-intl.

adomasven:
Strings from .dtd files are now accessible from Zotero.getMessage as the eventual move to Fluent would remove their distinction.

Throws if there are duplicate string ids between .dtd and .properties files

### DIFF
--- a/chrome/content/zotero/xpcom/intl.js
+++ b/chrome/content/zotero/xpcom/intl.js
@@ -82,6 +82,13 @@ Zotero.Intl = new function () {
 			let regexp = /<!ENTITY ([^\s]+)\s+"([^"]+)/g;
 			let regexpResult;
 			while (regexpResult = regexp.exec(localeXML)) {
+				let bundleString;
+				try {
+					bundleString = bundle.GetStringFromName(regexpResult[1]);
+				} catch (e) {}
+				if (bundleString) {
+					throw new Error(`Found a duplicate string ${regexpResult[1]} for both bundle and dtd strings`)
+				}
 				this.strings[regexpResult[1]] = regexpResult[2];
 			}
 		}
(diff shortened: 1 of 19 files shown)


adomasven:
I kept the `react-intl` package there for now as plugin devs that used VirtualizedTable would have needed to use the `IntlProvider` class from it. We should remove it in the future.

chrome/content/zotero/xpcom/intl.js
```diff
@@ -82,6 +82,13 @@ Zotero.Intl = new function () {
 			let regexp = /<!ENTITY ([^\s]+)\s+"([^"]+)/g;
 			let regexpResult;
 			while (regexpResult = regexp.exec(localeXML)) {
+				let bundleString;
+				try {
+					bundleString = bundle.GetStringFromName(regexpResult[1]);
+				} catch (e) {}
+				if (bundleString) {
+					throw new Error(`Found a duplicate string ${regexpResult[1]} for both bundle and dtd strings`)
+				}
```
dstillman, line 91: We should just add this to a test. It's only ~4 milliseconds for me, but no reason to do it on every startup, and milliseconds add up eventually.

chrome/content/zotero/xpcom/intl.js
```diff
@@ -105,6 +112,9 @@ Zotero.Intl = new function () {
 				}
 				l10n = bundle.formatStringFromName(name, params, params.length);
 			}
+			else if (Zotero.Intl.strings[name]) {
+				return Zotero.Intl.strings[name];
+			}
```
dstillman, line 117: `this`

chrome/content/zotero/xpcom/intl.js
adomasven: We're initializing the strings here anyway so I didn't think it egregious. If it's a test you'll end up not noticing you've added a duplicate until pushing to github and the test suite running, which is like a whole lot more compute, if we're concerned about that.

dstillman: It's totally fine, but there's a (very small) measurable difference for just this block, so we might as well avoid it.

Given that there was a single duplicate in 15 years of strings, I think it's fine for it not to be caught until CI (and that's only necessary if you mess it up in the direction where you wouldn't immediately notice it in the UI you're working on).

adomasven:
Updated.

PR #2985, opened 2023-02-01
Add a hidden pref for tab title with support for Creator-Year-Title option

mrtcode:
- Added `zotero.tabs.title` pref (just like Firefox has), but if we don't expect to have more prefs under `zotero.tabs` group, we could have a different name.
- The pref is not only for PDF tabs, but for all future tab types as well. At least for what the reader provides. Maybe except note tab, if we'll ever do it.
- The pref supports two options: `title`, `creatorYearTitle`.
- The pref will only have an effect if the current PDF file is the "best" attachment, otherwise tab title will be an attachment filename.
- The default option now is 'title' (previously we were displaying `Title - Creator - Year`).
- After changing the pref, unloaded tabs won't be updated until selected, because of technical limitations. If we think this is too annoying we can try to work around that.

### DIFF
--- a/chrome/content/zotero/xpcom/reader.js
+++ b/chrome/content/zotero/xpcom/reader.js
@@ -105,7 +105,7 @@ class ReaderInstance {
 		this.state = state;
 		this._itemID = item.id;
 		// Set `ReaderTab` title as fast as possible
-		this.updateTitle();
+		await this.updateTitle();
 		let path = await item.getFilePathAsync();
 		// Check file size, otherwise we get uncatchable error:
 		// JavaScript error: resource://gre/modules/osfile/osfile_native.jsm, line 60: RangeError: invalid array length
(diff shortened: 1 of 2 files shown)


dstillman:
> The default option now is 'title' (previously we were displaying Title - Creator - Year).

I know I said `title`, but I guess the options might as well be `titleCreatorYear` and `creatorYearTitle`? No real reason to remove creator and year from the title option.

If an item doesn't have a creator and the pref is set to `creatorYearTitle`, right now it will show, e.g., "2012 - Track C Epidemiology and Prevention Science". I think we should probably just revert to Title - Creator - Year (so "Track C Epidemiology and Prevention Science - 2012") in that case. Another way to look at it is that it's actually author-date followed by title, so if no author, author-date doesn't apply.

mrtcode:
Ok, now three title types are supported:
- `titleCreatorYear` (default).
- `creatorYearTitle` (falls back to `titleCreatorYear` if creator is missing).
- `title`.

PR #3010, opened 2023-03-02
modified standalone keyset to include the quitApplication keybinding

oaken-source:
The default keyset on unix and windows does not include the keybinding for exiting the application. On macos this works as expected because the keybinding is included in the macos keyset. This pull request adds the appropriate keybinding for the other platforms as well.

### DIFF
--- a/chrome/content/zotero-platform/unix/standalone/menuOverlay.xul
+++ b/chrome/content/zotero-platform/unix/standalone/menuOverlay.xul
@@ -40,6 +40,14 @@
 <overlay id="menuOverlay"
          xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
          xmlns="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul">
+
+	<keyset id="mainKeyset">
+		<key id="key_quitApplication"
+				key="&quitApplicationCmd.accesskey;"
+				command="cmd_quitApplication"
+				modifiers="accel"/>
+	</keyset>
+
     <menupopup id="menu_FilePopup">
 		<menuseparator/>		
 		<menuitem id="menu_FileQuitItem" 
--- a/chrome/content/zotero-platform/win/standalone/menuOverlay.xul
+++ b/chrome/content/zotero-platform/win/standalone/menuOverlay.xul
@@ -44,7 +44,7 @@
 		<menuseparator/>
 		<menuitem id="menu_FileQuitItem" 
 				label="&quitApplicationCmdWin.label;"
-				key="&quitApplicationCmdWin.accesskey;"
+				accesskey="&quitApplicationCmdWin.accesskey;"
 				command="cmd_quitApplication"/>
     </menupopup>
     <menupopup id="menu_EditPopup">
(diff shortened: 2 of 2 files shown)


dstillman:
I don't think we'd do Ctrl-Q on Windows. When in doubt we generally defer to Firefox, and Firefox uses Ctrl-Q on Linux but Ctrl-Shift-Q on Windows (which I think is more reasonable anyway given the closeness to Ctrl-W and the lack of any sort of confirmation prompt). So if we do this, I think we'd want to use `accel,shift` and either use that for Linux as well or change the attribute programmatically at startup.

(Chrome on Windows seems to be content with `Alt`-`f` + `x`.)

oaken-source:
that's a good catch, it's been a long time since I used windows. Ctrl+Shift+Q to quit seems unintuitive to me, considering that macos also uses a simple Command+Q. I think windows is the odd one out in that regard.

dstillman:
Cmd-Q has been universal on macOS for decades (though Chrome does a weird non-standard two-press thing — somewhat reasonably, given how easy it is to trigger by accident in a tabbed app and how disruptive it can be), but Ctrl-Q just isn't a thing on Windows. There's Alt-F4 or Alt-F + X, but that's it. So I think Firefox's Ctrl-Shift-Q is just a compromise to have something vaguely similar to Firefox on other platforms.

oaken-source:
understood. I'll try to find a configuration that works acceptably on all platforms, but it might take me a while to set up the build tools and environment on windows

dstillman:
Note that you don't need to build on Windows, and I wouldn't recommend it. You can build Windows (without packaging) anywhere with `scripts/dir_build -p w` (after `./fetch_xulrunner.sh -p w`) and just open the built app in `staging` on Windows.

Thanks for working on this!

qqobb:
@oaken-source: On Windows, I would recommend using Alt-F4 for closing applications, since this works universally across programs. While shortcuts like Ctrl-W work in some apps like File Explorer (Win-E), Alt-F4 also works, e.g., in the Settings app.

You could perhaps remap the Alt-F4 shortcut to Ctrl-Q using [Microsoft PowerToys](https://github.com/microsoft/PowerToys/), which provide a [Keyboard Manager utility](https://learn.microsoft.com/en-us/windows/powertoys/keyboard-manager). But I personally wouldn't recommend it for the reasons given by dstillman.

I've been using [`Zotero.Utilities.Internal.quit()`](https://github.com/zotero/zotero/blob/59ba212cf8a38da3286f019e0d1442e928a52b59/chrome/content/zotero/xpcom/utilities_internal.js#L2038-L2048) within the [Zutilo add-on](https://github.com/wshanks/Zutilo) to create shortcuts for quitting/restarting Zotero. It's not in the repo yet, but I'll create a PR soon, see https://github.com/wshanks/Zutilo/issues/249.

dstillman:
Alt-F4 already works in Zotero.

Alt-F + X isn't a shortcut — it's just selecting the File menu and then an access key for Exit. The first part already works. It looks like we're just missing the 'x' access key on Exit at the moment, so 'e' works instead now. So I think we should just add 'x' and then forget about Ctrl-Shift-Q on Windows (meaning that we'll want to assign Ctrl-Q programmatically just for Linux).

oaken-source:
I pushed a new version that (should):

a) modify the unix menu to bind the configured key (q in en_US locale) to quit and add a Ctrl+(key) keybinding
b) modify the windows menu to bind configured key (x in en_US locale) to exit

It's pretty straightforward, and Alt+f, x should now work on windows. but I haven't tested that yet.

dstillman:
Windows needs an accesskey, not a shortcut key. Fixed and merged. Thanks!

PR #3025, opened 2023-03-14
Sorts first-author last name matches first when year present.

### LINKED ISSUE #3024: Quick Format: Sort first-author last name matches first even with year present
https://forums.zotero.org/discussion/comment/430442/#Comment_430442

1. Create two items with 2014 as the date.
2. Make the author for one "Smith, John".
3. For the other, make the authors "Bar, Foo" and "Smith, John".

If you enter "smith" in the citation dialog, it will sort the Smith item first. If you enter "smith 2014", it's no longer guaranteed to sort the Smith item first. (You might need to swap the names between the items to see this.)

adomasven:
Closes #3024

I'm submitting this as a PR because I feel like maybe this might break some other search? We're only matching by author, title and date and the combinations I've tried to test all returned what I expected, so maybe it's just fine.

### DIFF
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -499,6 +499,7 @@ var Zotero_QuickFormat = new function () {
 			await Zotero.Items.loadDataTypes(items);
 			
 			searchString = searchString.toLowerCase();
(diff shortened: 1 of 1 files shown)


dstillman:
Yeah, I'm not seeing any problems. Let's see how it goes.

PR #3058, opened 2023-04-04
Add scripts to convert ftl to/from Transifex JSON

tnajdek:
This PR adds a pair of scripts to generate JSON that Transifex can understand from `chrome/locale/en-US/zotero/zotero.ftl`.

* Runing `npm run ftl-to-json` produces `tmp/tx/zotero_en_US.json` which is recognized by Transifex
* `Structured JSON` is the format in Transfiex I've used for testing. I believe `key-value JSON` also works because we're not using nested messages feature of structured JSON
* Running `npm run json-to-ftl` collects files from `tmp/tx` using regex `/(zotero_)([a-z]{2}([_-][A-Z]{2})?).json/` and creates translated `zotero.ftl` in locale-named folders in `chrome/locale/`. `en-US` is ignored because it is a source file.
* Some extra data is embedded in JSON to assist translating back to ftl. I've tested and Transifex does include it in the output file so this shouldn't be an issue
* Actual logic for translation exists in [ftl-tx](https://github.com/tnajdek/ftl-tx) and is included via dev npm depdendency.
* There are some [limitations](https://github.com/tnajdek/ftl-tx#limitations) of what is possible with the translation. Everything that is in current `zotero.ftl` works
* Requires relatively-modern Node (>= 16 should be sufficient)

I hope these scripts can be integrated with any pre-existing Transifex-related automation, if anything needs adjusting let me know.

### DIFF
--- a/chrome/locale/en-US/zotero/zotero.ftl
+++ b/chrome/locale/en-US/zotero/zotero.ftl
@@ -55,8 +55,7 @@ file-interface-items-were-relinked = { $numRelinked ->
     *[other] { $numRelinked } items were relinked
     }
 
-import-mendeley-encrypted = The selected Mendeley database cannot be read, likely because it is encrypted.
-                            See <a data-l10n-name="mendeley-import-kb">How do I import a Mendeley library into Zotero?</a> for more information.
+import-mendeley-encrypted = The selected Mendeley database cannot be read, likely because it is encrypted. See <a data-l10n-name="mendeley-import-kb">How do I import a Mendeley library into Zotero?</a> for more information.
                             
 file-interface-import-error-translator = An error occurred importing the selected file with “{ $translator }”. Please ensure that the file is valid and try again.
 
--- a/js-build/config.js
+++ b/js-build/config.js
@@ -129,8 +129,22 @@ const scssFiles = [
 	'chrome/skin/default/zotero/**/*.scss'
 ];
 
+const ftlFileBaseNames = [
+	'zotero',
+	'preferences',
+];
+
 const buildsURL = 'https://zotero-download.s3.amazonaws.com/ci/';
 
 module.exports = {
-	dirs, symlinkDirs, copyDirs, symlinkFiles, browserifyConfigs, jsFiles, scssFiles, ignoreMask, buildsURL
+	dirs,
+	symlinkDirs,
+	copyDirs,
+	symlinkFiles,
+	browserifyConfigs,
+	jsFiles,
+	scssFiles,
+	ignoreMask,
+	ftlFileBaseNames,
+	buildsURL,
 };
--- a/js-build/ftl-to-json.mjs
+++ b/js-build/ftl-to-json.mjs
@@ -5,16 +5,15 @@ import { fileURLToPath } from 'url';
 import { onError, onSuccess } from './utils.js';
 
 const ROOT = join(dirname(fileURLToPath(import.meta.url)), '..');
-const JSONDir = join(ROOT, 'tmp', 'tx');
 
 async function getJSON() {
 	const t1 = performance.now();
-	await fs.mkdirp(JSONDir);
-	const sourcefile = join(ROOT, 'chrome', 'locale', 'en-US', 'zotero', 'zotero.ftl');
-	const destFile = join(JSONDir, 'zotero_en_US.json');
-	const ftl = await fs.readFile(sourcefile, 'utf8');
+	const sourceDir = join(ROOT, 'chrome', 'locale', 'en-US', 'zotero');
+	const sourceFile = join(sourceDir, 'zotero.ftl');
+	const destFile = join(sourceDir, 'zotero.json');
+	const ftl = await fs.readFile(sourceFile, 'utf8');
 	const json = ftlToJSON(ftl, { transformTerms: false, storeTermsInJSON: false });
-	await fs.outputJSON(destFile, json);
+	await fs.outputJSON(destFile, json, { spaces: '\t' });
 	const t2 = performance.now();
 	return ({
 		action: 'ftl->json',
--- a/js-build/ftl-to-json.mjs
+++ b/js-build/ftl-to-json.mjs
@@ -2,25 +2,29 @@ import { ftlToJSON } from "ftl-tx";
 import fs from 'fs-extra';
 import { dirname, join } from 'path';
 import { fileURLToPath } from 'url';
-import { onError, onSuccess } from './utils.js';
+import { ftlFileBaseNames as sourceFileBaseNames } from './config.js';
+import { onError, onProgress, onSuccess } from './utils.js';
 
 const ROOT = join(dirname(fileURLToPath(import.meta.url)), '..');
 
 async function getJSON() {
 	const t1 = performance.now();
 	const sourceDir = join(ROOT, 'chrome', 'locale', 'en-US', 'zotero');
-	const sourceFile = join(sourceDir, 'zotero.ftl');
-	const destFile = join(sourceDir, 'zotero.json');
-	const ftl = await fs.readFile(sourceFile, 'utf8');
-	const json = ftlToJSON(ftl, { transformTerms: false, storeTermsInJSON: false });
-	await fs.outputJSON(destFile, json, { spaces: '\t' });
+	for (let sourceFileBaseName of sourceFileBaseNames) {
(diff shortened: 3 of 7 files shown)


dstillman:
So I think we could make this a bit simpler, both for us and for localizers.

With .properties and .dtd, missing strings in a given locale were hard failures, so we have a script that uses the en-US strings as a template and adds missing strings into the non-en-US files. (Transifex does that too, but this lets us add new English strings and push a beta build that isn't broken for non-en-US users without doing a whole Transifex round-trip.)

As far as I know Fluent has automatic fallback to en-US, so this isn't necessary for safety here, but I think it may still be the right approach, because it would mean we wouldn't need to do the special term handling and show weird placeholder strings to localizers. (I'm assuming localizers would currently see "FTLREF_app_name" in Transifex?)

Instead of `ftl-to-json` and `json-to-ftl`, we'd have `ftl-to-json` and (say) `localize-ftl`. `ftl-to-json` would leave the strings as is (or just convert the variables and terms to whatever format they need to be in to be protected in Transifex). `localize-ftl` would do the following:

- For each non-en-US locale in `chrome/locale`
  - Check for an .ftl file and read it into a map by key
  - Check for a .json file and convert it back into a similar map by key (combining separate strings as necessary)
  - Go line by non-empty line in the en-US .ftl file, creating a new .ftl file that uses, in decreasing priority, 1) the converted .json line, 2) the existing locale .ftl line, or 3) the en-US line. Terms would just be copied over directly, so there wouldn't be any need to fetch those from the locale .ftl/.json files.

This way, all terms from en-US will just exist by default in the locale .ftl files and localizers would just see the simple variable/term names that are in the file. (Do we even need to remove the `$` and `-`?)

dstillman:
I mean, or we just keep the `-` prefix instead of using `FTLREF_` to indicate terms, which would address my main concern regarding what localizers see, but the `terms` stuff just seems a bit weird — i.e., including a copy of the term value (e.g., "Zotero") for each string it's used in. Just seems cleaner to copy it over from the canonical source (en-US) when writing a locale .ftl file. There's no real reason for it to be in the .json file at all.

dstillman:
But `-` isn't a special character, is it? Can it not just stay `-foo-boo`?

tnajdek:
OK, this will now keep reference names as is, with nothing extra stored in JSON. I've also changed `json-to-ftl` to `localize-ftl`, which will produce a new `zotero.ftl` file for each non-en-US locale in `chrome/locale`. Before converting back to `.ftl`, it uses JSON format to combine strings from the following sources (in order of preference):

1. Transifex JSON
2. Local ftl in this locale (converted to JSON)
3. `en-US` ftl, converted to JSON already by `ftl-to-json`

Term values always come from `en-US/zotero.ftl`.

Please make sure you delete old `tmp/tx` before using this to avoid old-style, prefixed terms polluting .ftl files.

Any other suggestions, please let me know!

dstillman:
OK, rebased and tweaked to work better with the Transifex process.

Do we know how Fluent handles whitespace? E.g., this:

```
import-mendeley-encrypted = The selected Mendeley database cannot be read, likely because it is encrypted.
                            See <a data-l10n-name="mendeley-import-kb">How do I import a Mendeley library into Zotero?</a> for more information.
```

on round-trip turns into this:

```
import-mendeley-encrypted =
    The selected Mendeley database cannot be read, likely because it is encrypted.
                                See <a data-l10n-name="mendeley-import-kb">How do I import a Mendeley library into Zotero?</a> for more information.
```

dstillman:
Hmm, now it occurs to me that we might not want to include terms in the localized files at all, since they won't ever change, and for something like `-app-name` it would just be something that a fork would need to modify in every locale instead of one.

I assume Fluent will automatically fall back to en-US for terms as well, but will it do so without logging a warning?

tnajdek:
There was a bug in `ftl-tx` that would end up introducing extra whitespace when parsing multi-line, indented strings, I've fixed that, updated commit and also update that string `zotero.ftl` because it would still end up with additional whitespaces based on the rules I've quoted above.

dstillman:
Pretty sure that string was actually fine. The `=` line doesn't count, and all common indentation on subsequent lines is ignored, so since there's only one subsequent line, the whitespace gets removed. It looks like this is actually the last example in the documentation:

> Note that if a multiline pattern starts on the same line as the identifier and it only consists of one more line of text below it, then the indent common to all indented lines is equal to the indent of the second line, i.e. the only indented line. All indent will be removed in this case.

dstillman:
Hmm, so in the conversion to JSON it's adding a `\n` between multiline Fluent strings, but that's not really right. It means that it can convert it back to a localized .ftl with the same multiline representation, but there's no actual need for that for readability — only the en-US strings are being edited by hand. And it's somewhat misleading to show the newline to localizers, since it's not actually displayed that way in the app. By this logic, we should strip these the whitespace the same way Fluent will and show on Transifex and in the localized .ftl as a continuous string.

At the same time, newlines that are actually preserved in Fluent do need to be preserved. If I'm reading the examples right, this string:

```
foo = This is a first line.
          This is still part of the first line.

          This is a second line.
```

would need to be converted to `This is a first line. This is still part of the first line.\n\nThis is a second line.` in JSON so that it was preserved going back to Fluent.

So then maybe trying to replicate Fluent multiline rules in our own code is too complicated/dangerous and we should just keep preserving them all, even if the displayed string in Transifex is slightly misleading?

tnajdek:
I don't think your example is correct, I believe all new-line characters are taken into account i.e. the result would be:

````
This is a first line.\nThis is still part of the first line.\n\nThis is a second line.
````

As far as I can tell we're not doing anything different than what fluent would do - we're using the official lib for parsing and serializing. I don't think we should add any special treatment for whitespace on our end.

I think the solution here is quite simple, `import-mendeley-encrypted` should not have been a multi-line string. It's my mistake to split it into two lines and, as you've said, it can be misleading for translators, because these new-lines do not end up appearing in Zotero. We already had very long strings in `zotero.dtd`/`zotero.properties` so this isn't different in any way.

dstillman:
Well, I'm saying my example would be what we'd want to show on Transifex given Fluent's parsing — i.e., the first two lines are a continuous line, and the last line comes after two newlines. Showing `This is a first line.\nThis is still part of the first line.` would be incorrect on Transifex for the same reason showing a `\n` in the Mendeley example would be incorrect. And preserving that newline going back to the localized files isn't necessary.

But yes, I guess we can avoid this by just not relying on multiline strings for single lines in the en-US file. A little messier to edit by hand, but better than having confusing strings in Transifex.

tnajdek:
What I'm saying is that fluent doesn't parse first two lines as a continuous line in your example. Here is how it's parsed by the official lib (which we also use): [runkit](https://runkit.com/embed/rvuqaj7bw822).

In the meantime I've also force-pushed the last commit to remove the misleading, multi-line string.

dstillman:
Oh, I see. I misinterpreted the "Line breaks and blank lines are preserved as long as they are positioned inside of multiline text" example in the documentation. I meant "misleading" specifically for Transifex users, but really it was just wrong, then, since the newline shouldn't be in the rendered string regardless of whether it shows up in the app.

tnajdek:
>Hmm, now it occurs to me that we might not want to include terms in the non-English files at all, since they won't ever change, and for something like -app-name it would just be something that a fork would need to modify in every locale instead of one.

> I assume Fluent will automatically fall back to en-US for terms as well, but will it do so without logging a warning?

I agree, if there is a silent fallback, we should remove term definitions from other `.ftl` file. I think the easiest approach would be to test it, once `fx102` can be used with non-en-US locale?

dstillman:
Should be usable now. I merged the props/dtds the other day.

tnajdek:
Fluent does not seem to fall-back for term values from `en-US`:
Screenshot 2023-04-18 at 15 53 45
However I don't think this is an issue. Term values are always taken from `en-US/zotero.ftl` so if someone wants to fork Zotero and change the value, they still need to do it only in one place, then run `npm run localize-ftl` to copy it to all the other files.

dstillman:
Is it falling back to en-US for other missing strings?

tnajdek:
> Is it falling back to en-US for other missing strings?

Yes it does.

dstillman:
Are we able to support functions (for `PLATFORM()`)?

https://github.com/zotero/zotero/commit/371bcfea818ba2ce72c11f157cc46b995fa9af63

PR #3060, opened 2023-04-05
fx-compat: Item box: Use custom text field styles on Windows

### LINKED ISSUE #3050: Get rid of Win11 textbox style with blue bottom border

AbeJellinek:
Active:

Screenshot 2023-04-05 at 1 45 54 PM

Inactive:

Screenshot 2023-04-05 at 1 46 05 PM

We can't make this only affect >= 11, unfortunately - see comment in the SCSS.

Fixes #3050

### DIFF
--- a/scss/win/elements/_itemBox.scss
+++ b/scss/win/elements/_itemBox.scss
@@ -17,10 +17,22 @@ item-box {
 		margin-bottom: 0;
 	}
 	
-	input {
-		padding: 2px;
-		margin: -1px 0;
-		margin-inline-start: 1px;
+	@media (-moz-platform: windows-win10) {
+		td > input, td > textarea, .creator-name-box > input {
+			// Give text fields a consistent, native-ish look using Windows 11 colors
+			// Ideally, we only want to do this on Windows 11, but Windows 11 reports its version number as 10.0.
+			// The OS build number (the only reliable way to determine whether we're running on >=11) is only
+			// accessible from C++.
+			appearance: none;
+			padding: 2px 3px;
+			border: 1px solid #838383;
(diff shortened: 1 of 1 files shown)


dstillman:
> We can't make this only affect >= 11

Huh, and Bugzilla bug opened only a week ago: https://bugzilla.mozilla.org/show_bug.cgi?id=1825447

scss/win/elements/_itemBox.scss
```diff
@@ -17,10 +17,21 @@ item-box {
...
-		margin: -1px 0;
-		margin-inline-start: 1px;
+	@media (-moz-platform: windows-win10) {
+		td > input, td > textarea, .creator-name-box > input {
+			// Give text fields a consistent, native-ish look using Windows 11
+			// colors. Ideally, we only want to do this on Windows 11, but
+			// Windows 11 can't currently be specifically targeted.
+			// https://bugzilla.mozilla.org/show_bug.cgi?id=1825447
+			padding: 2px 3px;
+			border: 1px solid #838383;
```
adomasven, line 27: I think for colours like these we should use either mozilla's own defined `--toolbar-field-border-color` or `ThreeDShadow` css variables if they match our wanted look. That way when we add support for Dark Mode, we will have less work.

dstillman: Yeah, that gray looks fine. I guess `var(--toolbar-field-border-color)`, to be less weird Mozilla-y.

dstillman: Huh! OK, well, still, probably better to go with the CSS variable than something deprecated.

PR #3061, opened 2023-04-05
Don't add border to the top/sides of the toolbar on Windows

AbeJellinek:
Before:
Screenshot 2023-04-05 at 1 52 33 PM

After:
Screenshot 2023-04-05 at 1 52 16 PM

See gray border to the left of the New Collection button and to the right of the Sync button in the Before screenshot. It doesn't look right, and the commit message (61684d88992ea95f38f96fb971a53d201a6d3bf9) implies that it wasn't intentional.


dstillman:
Not intentional!

PR #3066, opened 2023-04-06
Tab bar: Make bottom border span the whole window

AbeJellinek:
This isn't an fx102 regression, but it keeps bothering me and now seems like an alright time to try an idea for a fix.

Before:

before-left
No dark border on the left.
before-right
Dark border until halfway through the Sync button.

After:
after-left
after-right
Dark border throughout.


mrtcode:
It's great that you're trying to fix that, but in the scrolled mode there is another border visible below the tab bar:

Before PR:
before-pr

After PR:
after-pr

PR #3113, opened 2023-05-08
Hide About window in Window menu (Mac)

### LINKED ISSUE #3103: Z7: About window doesn't have label in Window menu (macOS)

AbeJellinek:
Showing a titleless window and hiding it in the Window menu is in line with the behavior of Mail, TextEdit, and Firefox, but not Finder (which sets a title and shows it in the Window menu).

Fixes #3103.


dstillman:
Shouldn't we just add "About Zotero" as the window title?

dstillman:
Hmm, no, you're right, every Mac app other than Finder seems to do this. Weird! But I guess we can go with the flow.

PR #3114, opened 2023-05-08
fx-compat: Fix feed button appearance/overflow behavior

### LINKED ISSUE #3110: Z7: Mark as Read/Unread button in feeds overflows with longer localized strings
https://github.com/zotero/zotero/issues/3092 was fixed by disabling `overflow: hidden`, `text-overflow: ellipsis`, and `max-width: 150px`, which we had to handle longer localized strings at narrow pane widths. Without those, the button works, but long strings overflow the button. We should fix that.

AbeJellinek:
Just a whole series of weird issues, mainly with XUL, to work around here:

1. XUL button text can't wrap or be ellipsized, and overflows the button unless `overflow: hidden` is set
2. Setting `overflow: hidden` apparently breaks clicking
3. flexbox might work better, but all attempts to use flexbox inside a XUL button are doomed to fail because of native UI control rendering magic

So we replace the XUL button with an HTML button-based CE that uses flexbox, but

4. XUL menus directly within a flex container (like our HTML button-based CE) can't be shown
5. Adding a wrapper element to serve as the flex container around the icon, text, etc., creates even more layout issues

So we add support for a `popup` attribute to `showNativeElementPopup()` and move the popup outside of the button, keeping the button as the flex container.

Fixes #3110

### DIFF
--- a/chrome/content/zotero-platform/mac/itemPane.css
+++ b/chrome/content/zotero-platform/mac/itemPane.css
@@ -1,11 +1,11 @@
+#zotero-feed-item-toggleRead-button {
+	margin-top: 5px;
+	margin-bottom: 3px;
+}
+
 #zotero-feed-item-addTo-button {
 	margin: 5px 6px 3px;
 	padding-left: 8px;
-}
-
-#zotero-feed-item-addTo-button button {
-	margin: 0;
-	-moz-appearance: none;
 }
 
 /* Show duplicates date list item as selected even when not focused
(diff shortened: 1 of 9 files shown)


dstillman (changes requested):
The left button is missing the 6px padding on the left:

Screenshot 2023-05-09 at 4 00 14 AM

When the pane is widened, the target icon should stay with the text instead of separating.

Z6:

Screenshot 2023-05-09 at 3 56 17 AM

Z7:

Screenshot 2023-05-09 at 3 58 20 AM

PR #3140, opened 2023-05-26
Select an item when passed a zotero://select URI at startup

AbeJellinek:
`CollectionTree#selectItems()` calls `itemTreeView.waitForLoad()`, but `itemTreeView` is set from `ItemTree#changeCollectionTreeRow()`, which we can't listen for the completion of unless we... already have an ItemTree. Which we don't.

So we set `itemTreeView` immediately upon creating the CollectionTree and ItemTree. As far as I can tell, that doesn't break anything.

https://forums.zotero.org/discussion/comment/435588/#Comment_435588

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -1195,6 +1195,7 @@ var ZoteroPane = new function()
 			});
 			ZoteroPane.itemsView.onRefresh.addListener(() => ZoteroPane.setTagScope());
 			ZoteroPane.itemsView.waitForLoad().then(() => Zotero.uiIsReady());
+			ZoteroPane.collectionsView.itemTreeView = ZoteroPane.itemsView;
 		}
 		catch (e) {
 			Zotero.logError(e);


dstillman:
This is probably fine to do, but it doesn't fully solve the problem. At least for me, if I select a group, close Zotero, and click on a select link to My Library, on open it switches to My Library but doesn't select the item.

Basically, the collection tree hasn't yet been loaded, so the selection of the library root in zotero-protocol-handler.js fails:

```
(3)(+0000000): CollectionTree: Trying to access a row with invalid ID L1

(3)(+0000000): CollectionTree.selectLibrary(): library with ID 1 not found in collection tree
```

(This is with https://github.com/zotero/zotero/commit/77241d1c71e915160286a3ae4a4daf97d0fcbd86 applied.)

Then, as part of the startup process, the last-selected collection is selected:

```
(3)(+0000000): itemTree.changeCollectionTreeRow(): L21972
```

And right after that, collectionTree.selectItems() checks the library and sees that it's wrong, so it switches:

```
(3)(+0000000): Library ID differs; switching library

(3)(+0000007): itemTree.changeCollectionTreeRow(): L1
```

This results in two refreshes, one for the group and one for My Library:

```
(3)(+0000001): Refreshing items list for item-tree-main-default

[…]

(3)(+0000000): Refreshing items list for item-tree-main-default
```

There's then this line in collectionTree.selectItems():

```
await this.itemTreeView.waitForLoad();
```

But I'm guessing that actually only waits for the group library to load, because `this.itemTreeView.selectItems(itemIDs)` gets called and then there's this:

```
(3)(+0000000): Couldn't find row for item 84144 -- not selecting

(3)(+0000000): Item was not selected; switching to library

(3)(+0000000): CollectionTree.selectWait(): row 0 already selected

(3)(+0000000): Couldn't find row for item 84144 -- not selecting
```

So even though My Library is technically already selected and in the process of refreshing, the row map isn't yet updated and the item can't be found. This seems like a bug in the item tree when refreshed multiple times in rapid succession — `waitForLoad()` should probably wait for the most recent refresh, not the next one to finish.

But even if we fix that, this all seems like a race condition, since we're not actually waiting for the trees to be loaded before beginning to select things. So the library/collection selection in SelectExtension fails, and other actions in other protocol handler extensions might also be called too soon.

For the other extensions, we could probably just wait for `Zotero.uiReadyPromise`. For SelectExtension, though, 1) `uiReadyPromise` is too soon, since it apparently resolves when the item tree first mounts (which I'm not sure used to be the case, and we should probably talk about separately) and 2) we should probably think about this more broadly. Right now, when you restart Zotero, collection selection is persisted, but you lose item selection and scroll position. It seems like we should add support for those things, set a flag once they've been read at startup, and then have SelectExtension either set a value with selection information that would take priority over the persisted data (if they hadn't yet been read) or do the selection itself as it does now (if they had already been read). The scroll position wouldn't be relevant, so when using info from SelectExtension the tree startup code would just use `ensureRowsAreVisible()` as it does for selections now.

To store the data, I think we'd want to use session.js, which already has debouncing and can store any amount of data. (We can't store this in a pref, since item selection info could be huge.)

As a stopgap measure, I suspect we could fix this by doing what I describe above only for the `lastViewedFolder` pref to start.

PR #3143, opened 2023-05-28
Localization: "Name" to "Nachname" to clearify things.

debuglevel:
It's a german meme that we always write or firstname into "name" and then we recognize that the next field is "firstname".


dstillman:
Localization changes are made [on Transifex](https://app.transifex.com/zotero/zotero), but I've updated this there. Thanks!

debuglevel:
Thanks! I was looking for how localization is managed in Zotero but somehow missed https://www.zotero.org/support/dev/localization

Issue #3146, opened 2023-05-30
Script to convert .dtd files to .ftl files


dstillman:
To keep localizers from having to re-translate every string, we want to convert all source and target .dtd files to .ftl so that we can push them (via ftl-to-json) to new resources on Transifex.

We need to do this for Firefox 115, and ideally we want to produce a script (taking a .dtd and outputting an .ftl) that plugin developers could also use.

Unfortunately, the way the strings are used in Fluent may be somewhat different, so we might need some other approach here.

  * Keep a mapping file (that somehow encodes Fluent's more complicated key syntax) as we migrate strings?
  * Keep the values exactly the same and map the keys that way (with a `zotero.`-stripped pre-period prefix match on the key to resolve duplicates)? We only want to keep the localized strings the same if the English values are the same anyway.
  * Something else?

PR #3153, opened 2023-06-02
Convert .dtd to .ftl

tnajdek:
This is a PR to address and discuss #3146.

I've added a script `npm run dtd-to-ftl` that will process `.dtd` files and produce `.ftl` files according to the configuration in `js-build/config.js`. The config is far from complete; this is just to demo a concept.

Also, to keep things simple, I've moved the current `zotero.ftl` to `wizards.ftl`. If we don't want this (i.e., multiple `.ftl` files), we can remove that (i.e., have a single `zotero.ftl` file), but that might make importing from `.dtd` a bit more complex.

I've experimentally started replacing the uses of strings from `.ftl` in `zoteroPane.xhtml`, and I believe we will need to carefully write rules for transformations to create `.ftl` messages, depending on where the given message is used. The reason for this is that effectively, the attribute is moved from `.xhtml` to `.ftl`, e.g., `zotero.items.menu.attach` (in `.dtd`) becomes `items-menu-attach = .tooltiptext`, and `zotero.items.menu.attach.file` becomes `items-menu-attach-file = .label`.

I think the best approach is to go through all the strings currently in `zotero.dtd`, adjust the code where these are used, and write config rules for transformation at the same time. If we're happy with this approach, I can go ahead with it.

As for plugin authors, we could make the script optionally accept a custom config file that would be used instead of ours. This way, plugin authors could go through the same process.


dstillman:
> Also, to keep things simple, I've moved the current zotero.ftl to wizards.ftl. If we don't want this (i.e., multiple .ftl files), we can remove that (i.e., have a single zotero.ftl file), but that might make importing from .dtd a bit more complex.

Multiple .ftl files are fine — we might even want more. E.g., the PDF reader could have its own reader.ftl file with `reader.firstPage`/etc. strings, and that .ftl file could be included in zoteroPane.xhtml and reader.xhtml. This doesn't matter a lot, but might be nice to add a parameter to target another .ftl file instead of the one with the same basename, which it would default to.

tnajdek:
OK, I'll start a new repo for this tool.

> but might be nice to add a parameter to target another .ftl file instead of the one with the same basename, which it would default to.

If I understand this correctly, that's already possible. In the config object, key `target` specifies destination `.ftl` file. However, currently, target file is completely overridden with new content - not sure if we want to preserve messages already present and not coming from the imported `dtd`.

dstillman:
Sorry, I meant the ability to target another file for an individual string. Right now it wouldn't be possible to split specific strings off into other files, right? E.g., split all `zotero.pdfReader` strings into a reader.ftl file.

Or, better, change dtdMap into an array, where each array entry had `source`, `target`, and `mappings`. Then the PDF reader strings would just be a separate array entry that would pull those specific strings out of zotero.dtd.

tnajdek:
@dstillman I've started converting `zotero.dtd` to `.ftl` files with the help of the script I'm developing alongside. The scripts helps, but I don't think this process can be automated - every message usage has to be reviewed and sometimes adjusted (e.g. `<input>` inside the `<description>` in `errorReport.xhtml`) so it will take a bit of work. I'll keep updating this PR, if you have any comments please let me know.

windingwind:
Well, I saw a dtd2ftl script repo from the doc of Mozilla[1]: https://github.com/zbraniecki/convert_xul_to_fluent
and a properties2ftl there as well.

Might be helpful?

[1] https://firefox-source-docs.mozilla.org/l10n/migrations/legacy.html

PR #3157, opened 2023-06-06
Block remote content when indexing HTML file

### LINKED ISSUE #2721: fx-compat: Don't load remote content when indexing HTML file
In `Zotero.FullText.getPageData()`

We just want the character set and the body text, but we don't want to load remote images and such. Can we set something on the browser in HiddenBrowser.jsm to disable remote resources?

AbeJellinek:
We add an `http-on-modify-request` observer that blocks any request to a non-file: URL in browsers created with the `blockRemoteResources` option set to true. The observer is added on demand and removed when all browsers with remote content blocked have been destroyed. You'd think there'd be an easier way, but this is the best I could find.

Fixes #2721

### DIFF
--- a/chrome/content/zotero/HiddenBrowser.jsm
+++ b/chrome/content/zotero/HiddenBrowser.jsm
@@ -62,6 +62,7 @@ const HiddenBrowser = {
 	 * @param {Boolean} [options.allowJavaScript]
 	 * @param {Object} [options.docShell] Fields to set on Browser.docShell
 	 * @param {Boolean} [options.requireSuccessfulStatus]
+	 * @param {Boolean} [options.blockRemoteResources] Block all remote (non-file:) resources
 	 * @param {Zotero.CookieSandbox} [options.cookieSandbox]
 	 */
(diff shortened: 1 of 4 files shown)


chrome/content/zotero/HiddenBrowser.jsm
```diff
@@ -229,9 +234,41 @@ const HiddenBrowser = {
...
 	}
 };
+
+const RemoteResourceBlockingObserver = {
+	_observerAdded: false,
+	_browsingContextIDs: new Set(),
+	
+	observe(subject) {
+		let channel = subject.QueryInterface(Ci.nsIHttpChannel);
+		if (this._browsingContextIDs.has(channel.topBrowsingContextId)
```
dstillman, line 251: This changes to `browserId` in 115. We should add some code to try to support that.

https://searchfox.org/mozilla-central/rev/887d4b5da89a11920ed0fd96b7b7f066927a67db/netwerk/protocol/http/nsIHttpChannel.idl#467

PR #3168, opened 2023-06-15
Add support for mac word transaction initialization via http

adomasven:
Addresses zotero/zotero-word-for-mac-integration#35


adomasven:
Force pushed and added an ability to configure the target url by adding a `ZoteroUrl.txt` file in the Word startup folder, which should contain a string like `http://127.0.0.1:23119`.

dstillman:
Don't need the full URL in the file. Should just be `ZoteroPort.txt` and take a port number, with or without a trailing newline.

adomasven:
Sorry, I overwrote your commit. Done now regarding port config.

Issue #3001, opened 2023-02-22
Port keyboard-accessible toolbar to fx102


dstillman:
d866a10a2b56e96631a1563572d3f2918f1f006a

abaevbog:
I'm essentially done with the conversion, except here's a question: I can't seem to find what this commit is initially supposed to do ...

I assumed it makes it possible to navigate through all tool bar buttons (new note, new item, advanced search) by pressing Tab. That commit was part of Zotero 6.0.26, so I expected it would be possible there but not in Zotero 7. 

However, repeatedly tapping Tab in Zotero 6.0.26 goes through the same fields as Zotero 7 beta, and the toolbar buttons  never end up being reached.  [This screenshot ](https://www.dropbox.com/s/mu9k1w9uhy42434/screenshot.png?dl=0) shows the fields that get visited. 

So I am a bit confused what is the behavior that we are trying to replicate. Was I looking at the wrong Zotero version for comparison?

abaevbog:
Got it! I was looking at it on Mac and getting confused. 

I used a linux machine today, and it made much more sense. Had to tweak it a bit but with the last commit I could navigate all toolbar items with tabs and arrows

PR #3188, opened 2023-06-30
making toolbar accessible to the keyboard - initial conversion

abaevbog:
Initial work on converting [this commit](https://github.com/zotero/zotero/commit/d866a10a2b56e96631a1563572d3f2918f1f006a#diff-e82e5d43de73ef85cfc03f6402d5afb5c18a629850d689b1d764516b5cab98f3) for Zotero 7.
Addresses #3001

Here's the issue: I can't seem to find what this commit is initially supposed to do ...

As the name suggests, I assumed it makes it possible to get to `advanced search`, `new note`,` add attachment`, `new item` buttons via Tab. That commit was part of Zotero 6.0.26, so I expected it would be possible there but not in Zotero 7 without this change. 
However, the navigation via tab works for me the same exact way in Zotero 6.0.26 as it does in Zotero 7 beta - the toolbar buttons cannot be reached via Tab.  [This screenshot ](https://www.dropbox.com/s/mu9k1w9uhy42434/screenshot.png?dl=0) shows the fields that get visited by repeatedly hitting Tab for both Zotero versions. 

The buttons from the toolbar are accessible if I enable `Full Keyboard Access` on Mac but then the toolbar buttons are accessible in Zotero 7 without these changes. 

I must be missing something here.

### DIFF
--- a/chrome/content/zotero/elements/itemBox.js
+++ b/chrome/content/zotero/elements/itemBox.js
@@ -2313,6 +2313,19 @@
 		focusFirstField() {
 			this._focusNextField(1);
 		}
+
+		focusLastField() {
+			const tabbableFields=this.querySelectorAll('*[ztabindex]:not([disabled=true])');
+			const last=tabbableFields[tabbableFields.length-1];
+
+			if (last.classList.contains('zotero-focusable')) {
+				last.focus();
+			}
+			// Fields need to be clicked
+			else {
+				last.click();
+			}
+		}
 		
 		focusField(fieldName) {
 			let field = this.querySelector(`[fieldname="${fieldName}"][ztabindex]`);
--- a/chrome/content/zotero/elements/itemBox.js
+++ b/chrome/content/zotero/elements/itemBox.js
@@ -2315,8 +2315,8 @@
 		}
 
 		focusLastField() {
-			const tabbableFields=this.querySelectorAll('*[ztabindex]:not([disabled=true])');
-			const last=tabbableFields[tabbableFields.length-1];
+			const tabbableFields = this.querySelectorAll('*[ztabindex]:not([disabled=true])');
+			const last = tabbableFields[tabbableFields.length - 1];
 
 			if (last.classList.contains('zotero-focusable')) {
 				last.focus();
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -57,7 +57,10 @@ var ZoteroPane = new function()
 	this.reportErrors = reportErrors;
 	
 	this.document = document;
-	
+
+	const modifierIsNotShift = ev => ev.getModifierState("Meta") || ev.getModifierState("Alt") ||
+    ev.getModifierState("Control") || ev.getModifierState("OS");
+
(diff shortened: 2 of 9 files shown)


chrome/content/zotero/elements/itemBox.js
```diff
@@ -2313,6 +2313,19 @@
 		focusFirstField() {
 			this._focusNextField(1);
 		}
+
+		focusLastField() {
+			const tabbableFields=this.querySelectorAll('*[ztabindex]:not([disabled=true])');
+			const last=tabbableFields[tabbableFields.length-1];
```
dstillman, line 2319: Make sure you have eslint set up. You removed spaces around the `=` and `-`, which should be flagged by eslint. (Lots of existing warnings in existing code, but new code should be correct.)

chrome/content/zotero/zoteroPane.js
```diff
@@ -57,7 +57,10 @@ var ZoteroPane = new function()
 	this.reportErrors = reportErrors;
 	
 	this.document = document;
-	
+
+	const modifierIsNotShift = ev => ev.getModifierState("Meta") || ev.getModifierState("Alt") ||
+    ev.getModifierState("Control") || ev.getModifierState("OS");
```
dstillman, line 62: Fix indenting

chrome/content/zotero/zoteroPane.js
```diff
@@ -136,7 +139,260 @@ var ZoteroPane = new function()
...
+		for (const zone of zones) {
+			zone.start.setAttribute("tabindex", "0");
+		}
+
+		toolbar.addEventListener("keydown", (event) => {
+			// manually move focus when Shift+Tabbing from the search-menu-button
+			if (event.key === 'Tab' && event.shiftKey
+				&& !modifierIsNotShift(event)
+				&& event.originalTarget
+				&& event.originalTarget.id == "zotero-tb-search-menu-button") {
```
dstillman, line 299: Restore indent

chrome/content/zotero/zoteroPane.js
```diff
@@ -136,7 +139,260 @@ var ZoteroPane = new function()
...
+				zones[0].start.focus();
+				return;
+			}
+			// only handle events on a <toolbarbutton>
+			if (!isTbButton(event.target)) return;
+
+			const mapData = focusMap.get(event.target.id);
+
+			if (!Zotero.rtl && event.key === 'ArrowRight'
+				|| Zotero.rtl && event.key === 'ArrowLeft') {
```
dstillman, line 311: Restore indent, here and below

dstillman (changes requested):
This doesn't seem to be working right. From the quick search bar, the first Shift-Tab should focus the search bar's mode selector. The second Shift-Tab should focus New Collection. Here it seems to be skipping over the mode selector, but then Tab moves forward (correctly) to the mode selector.

abaevbog:
You are right. The search mode selector was moved into some kind of shadow DOM, so it wouldn't get recognized as an element that needs to be focused on. I added another special case for it to not get skipped on Shift-Tab from the search bar.

PR #3195, opened 2023-07-01
Fix indentation in bibliography

CommonLoon102:
Fix indentation in bibliography when a new item is added and therefore the recalculation of indents are needed.

https://forums.zotero.org/discussion/23689/bibliography-indentation

Before:
before

After:
after

### DIFF
--- a/chrome/content/zotero/xpcom/integration.js
+++ b/chrome/content/zotero/xpcom/integration.js
@@ -1348,18 +1348,11 @@ Zotero.Integration.Session.prototype._updateDocument = async function(forceCitat
 				} else {
 					bibliographyText = bib[0].bibstart+bib[1].join("")+bib[0].bibend;
 				}
-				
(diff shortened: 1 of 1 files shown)


adomasven:
The caching of bibliography style was added by another developer 13 years ago without explanation for its inclusion. When hitting such things we generally try to avoid changing stuff unless its broken, especially in the integration, where we have to work with 4 different word processors. Since this is fixing a bug, we need to consider it. 

I've reviewed code that this could potentially affect and it seems relatively harmless. My biggest worry is that this might negatively affect performance, but it doesn't seem likely. We'll merge it and see if we have any reports about issues in the beta.

adomasven:
Thanks for the PR!

adomasven:
We had to revert this PR as we did find a reason why the bibliography style wasn't reset on every request. See https://github.com/zotero/zotero/issues/3211

Issue #3152, opened 2023-06-01
Incorrect tab jumpback to non-library tab


dstillman:
1) Open some tabs
2) Select the first non-library tab
3) Restart Zotero
4) Select the last tab
5) Close the last tab

Expected result:

The tab to the left of the last tab is selected.

Actual result:

The first non-library tab is reselected.

Tab jumpback should only apply when opening a tab directly as a result of a user action within a tab. Selecting a tab from the tab bar should always clear tab jumpback, such that closing the selected tab just selects the tab immediately to the left.

abaevbog:
The issue arises because when we `select` a `reader-unloaded` tab ( all tab except for one are `reader-unloaded` after reopening the app ), that tab is closed and a new `reader` tab is created. That's when `add` function is called, which sets `this._prevSelectedID` to be the previous tab. Ordinarily (without closing and re-opening the app), the previous tab is `zotero-pane` (library tab), which is where we want to jump back if tab is closed. 

Unless I am missing something, the only jumpback destination is a library tab, so we should just not set `this._prevSelectedID` to anything that's not `zotero-pane`.

dstillman:
> Unless I am missing something, the only jumpback destination is a library tab

No, you can also do it if you have an annotation in a note in the Notes pane and click on it and select Show on Page. So this won't be sufficient.

abaevbog:
So we want the behavior from [the first commit](https://github.com/abaevbog/zotero/commit/ee934753770c74ece53a54d6cb589126e9cb0ec5) but we also want to keep the jumpback to PDF A if PDF B is opened from the annotation of PDF A and then closed. 

Per my last commit, we can add an extra parameter to options of `ReaderTab()`, `Zotero.Reader.open`, and `Zotero_Tabs.add()` to indicated when a jumpback should be skipped. And then when we call `Zotero.Reader.open` on `reader-unloaded` tab, we pass `preventJumpback=true`. 

That way, jumpback is prevented only when  `reader-unloaded` tab is involved.

PR #3204, opened 2023-07-06
jumpback only to library tab

abaevbog:
When a new tab is added, record _prevSelectedID only if the previous tab was zotero-pane.

Fixes: #3152

### DIFF
--- a/chrome/content/zotero/xpcom/reader.js
+++ b/chrome/content/zotero/xpcom/reader.js
@@ -1287,7 +1287,7 @@ class ReaderInstance {
 }
 
 class ReaderTab extends ReaderInstance {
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/xpcom/reader.js
```diff
@@ -1303,7 +1303,8 @@ class ReaderTab extends ReaderInstance {
 			data: {
 				itemID
 			},
-			select: !background
+			select: !background,
+			preventJumpback: preventJumpback
```
dstillman, line 1307: Can skip the second variable when they're the same

mrtcode:
From my tests, it looks ok.

Although in future I plan to rewrite tab loading/unloading because that doesn't work smooth enough, and the new Reader version should help here.

PR #3206, opened 2023-07-08
Update some zh-CN translations

lychichem:
Some translations are in zh-HK or zh-TW. This PR translate them into zh-CN.


dstillman:
Thanks, but localization changes need to be made on [Transifex](https://app.transifex.com/zotero/zotero/). You can join the group there to make changes.

PR #3208, opened 2023-07-11
Strip bidi control characters in filenames and elsewhere

### LINKED ISSUE #3207: Unicode "isolate" characters added to filenames
https://forums.zotero.org/discussion/comment/438450/#Comment_438450

I assume we're now returning these characters for `getField('firstCreator')`, and that's used in `Attachments.getFileBaseNameFromItem()`.

Another bug after https://github.com/zotero/zotero/commit/6d5ce578e430380b633c4e8a32898ab4975dfa2e and #3135

AbeJellinek:
Passing unformatted = true to `Item#getField()` now returns a bidi control character-less result, and we use that in `Reader#updateTitle()` and `getFileBaseNameFromItem()` to prevent bidi control characters from showing up in filenames and window titles (the former everywhere, the latter on Windows only).

We also strip bidi control characters in `getValidFileName()` to be extra safe.

Fixes #3207

### DIFF
--- a/chrome/content/zotero/xpcom/data/item.js
+++ b/chrome/content/zotero/xpcom/data/item.js
@@ -221,7 +221,7 @@ Zotero.Item.prototype._setParentKey = function() {
  *
  * @param {String|Integer} field fieldID or fieldName
  * @param {Boolean} [unformatted] Skip formatting of multipart date fields and
- * 		strip bidi control characters
+ * 		omit bidi control characters
  * @param {Boolean} includeBaseMapped If true and field is a base field, returns
  *   value of type-specific field instead
  *   (e.g. 'label' for 'publisher' in 'audioRecording')
--- a/chrome/content/zotero/xpcom/data/item.js
+++ b/chrome/content/zotero/xpcom/data/item.js
@@ -216,12 +216,12 @@ Zotero.Item.prototype._setParentKey = function() {
 // Public Zotero.Item methods
 //
 //////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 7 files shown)


test/tests/itemTest.js
```diff
@@ -35,6 +35,53 @@ describe("Zotero.Item", function () {
...
+					lastName: "D",
+					creatorType: "author"
+				}
+			]);
+			assert.equal(
+				item.getField('firstCreator', /* unformatted */ true),
+				Zotero.getString('general.andJoiner', ['B', 'D'])
+			);
+			
+			await item.saveTx();
```
dstillman, line 78: Add comment clarifying that we're testing both unsaved and saved, since they use different code paths

chrome/content/zotero/xpcom/data/item.js
```diff
@@ -216,12 +216,12 @@ Zotero.Item.prototype._setParentKey = function() {
...
 //////////////////////////////////////////////////////////////////////////////
-/*
+/**
  * Retrieves an itemData field value
  *
  * @param {String|Integer} field fieldID or fieldName
- * @param {Boolean} [unformatted] Skip any special processing of DB value
- *   (e.g. multipart date field)
+ * @param {Boolean} [unformatted] Skip formatting of multipart date fields and
+ * 		strip bidi control characters
```
dstillman, line 224: `strip` → `omit` (to go with the "unformatted" idea)

dstillman (changes requested):

Issue #3198, opened 2023-07-04
Disable File → Close in library tab if multiple tabs open


mrtcode:
It doesn't close the window if library tab is selected and there are more than one tab.

If this is intentional, at least the Close option should be grayed out.

dstillman:
Yes, that's [intentional](https://github.com/zotero/zotero/blob/aca86967bf3159d14b0eb4aa81b55d765b39a2f6/chrome/content/zotero/zoteroPane.js#L798-L804) (and yes, disabling Close would be good)

fcheslack:
On linux this also means there is no shortcut that quits Zotero now. Firefox and many other things (in KDE at least) uses Ctrl-Q which I think we should do.

dstillman:
There's still Alt-F + Q on Linux, but Ctrl-Q is supposed to work too. It looks like it's just broken.

PR #3210, opened 2023-07-14
disable file > close for library if 1+ tabs open

abaevbog:
When there are multiple tabs opened, menuitem File > Close is disabled for the library tab.

Fixes: #3198

### DIFF
--- a/chrome/content/zotero/tabs.js
+++ b/chrome/content/zotero/tabs.js
@@ -83,6 +83,9 @@ var Zotero_Tabs = new function () {
 			selected: tab.id == this._selectedID,
 			iconBackgroundImage: tab.iconBackgroundImage
 		})));
+		// Disable File > Close menuitem if multiple tabs are open
+		const multipleTabsOpen = this._tabs.length > 1;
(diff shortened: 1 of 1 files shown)


chrome/content/zotero/tabs.js
```diff
@@ -83,6 +83,10 @@ var Zotero_Tabs = new function () {
 			selected: tab.id == this._selectedID,
 			iconBackgroundImage: tab.iconBackgroundImage
 		})));
+		// Disable File > Close menuitem if multiple tabs are open
+		const multipleTabsOpen = this._tabs.length > 1;
+		document.getElementById('menu_close').setAttribute('disabled', multipleTabsOpen);
+		document.getElementById('menu_close').setAttribute('command', multipleTabsOpen ? false : 'cmd_close');
```
dstillman, line 89: You should be able to just disable the command, which will automatically disable associated menu items.

https://docs.huihoo.com/mozilla/xul/xultu/commands.html

abaevbog: Right! Still catching up on xulrunner

Issue #2434, opened 2022-03-21
Open Tabs menu


dstillman:
https://forums.zotero.org/discussion/95291/view-all-open-pdfs
https://forums.zotero.org/discussion/96821/feature-request-keep-master-collections-tab-visible-and-add-drop-down-vertical-tab-list

Something like the menus in Firefox and Chrome

Related to https://github.com/zotero/zotero/issues/2232

dstillman:
https://forums.zotero.org/discussion/102189/many-pdf-tabs-clutter

PR #3216, opened 2023-07-18
opened tabs menu

abaevbog:
Adds a toolbarbutton in the top right corner of the screen with a menu to view and navigate opened tabs.

Fixes: #2434


dstillman:
This is a great start, but I guess we actually want some of the other things in the Firefox menu:

  * Bold text for the current tab
  * A close button
  * The possibility of icons later to distinguish between tag types (not needed now)
  * Maybe other rich elements later

So I think we'll want to do a `panel` instead of a `menu`, with a `vbox` inside containing `display: flex` `toolbaritem`s containing pairs of `toolbarbutton`s for the tab title and close button.

See [`zotero-lookup-panel`](https://github.com/zotero/zotero/blob/d968e8ec338088c88b7ffb9195682e1658387863/chrome/content/zotero/zoteroPane.xhtml#L767-L770) for an example of a panel being opened from a button, via [this code](https://github.com/zotero/zotero/blob/d968e8ec338088c88b7ffb9195682e1658387863/chrome/content/zotero/lookup.js#L136-L143).

You can see what Firefox does for its tab menu from the [Browser Toolbox](https://firefox-source-docs.mozilla.org/devtools-user/browser_toolbox/index.html) by setting `ui.popup.disable_autohide` to true in about:config, triggering the tab menu, and searching for `#allTabsMenu-allTabsView`. Ignore the `panelmultiview`/`panelview` stuff, which we don't need.

dstillman:
And for the button itself, we can just do what Firefox does. No border except on hover, and it can just be a gray roundrect similar to Firefox.

abaevbog:
Now the content of the popup is an array of `toolbaritem`s which have two `toolbarbutton`s as children: `tabName` and `closeButton`. `closeButton` closes the tab on click, and `tabName ` opens the tab. 
There is also some gradient effect on hover of the `toolbaritem`s and the "Show opened tabs" button similar to Firefox.
This is [what it looks like now](https://www.dropbox.com/s/wsouyk0rwpppqcr/openedTabsMenu2.png?dl=0).

The cross and arrow-down icons I fetched from [bootstrap-icons](https://icons.getbootstrap.com).

I know I will need to play around with paddings and such a bit - the distances within the modal, I think, are a bit off. 
Also, Firefox has a nice search bar for tabs at the very top of the modal before the opened tabs array. 
Is that something that would be good to add as well?

dstillman:
> Firefox has a nice search bar for tabs at the very top of the modal before the opened tabs array.
Is that something that would be good to add as well?

Firefox's "Search tabs" menu item just focuses the address bar in tabs mode for me. Are you seeing something different?

But we could just have a filter bar at the top of the menu that filtered the list below, highlighting or bolding the matching text. Have to see how it looks, but it should be easy to do.

abaevbog:
Firefox search bar does the same for me. It was just an inspiration. 

In our case, yeah, a filter bar at the very top could be helpful, especially if there are a lot of tabs. I added it in the last commit with marking matching text as bold. I followed the same approach as with the `lookup` panel. It looks like [this](https://www.dropbox.com/s/zyxrs87q2ow2z70/search_bar.png?dl=0).

dstillman:
Nice!

  - No close button for the library view
  - No separate label for the search box. Try a `<search-textbox>` with a `placeholder` attribute. (That's what we use for the filter bar in preferences.xhtml.)
  - You should be able to tab from the filter bar to each title and close button.
  - I can tidy up the design.

chrome/content/zotero/tabs.js
```diff
@@ -244,6 +246,124 @@ var Zotero_Tabs = new function () {
...
+	// to add darker gradient
+	this.showOpenedTabsPanel = function (button) {
+		var panel = document.getElementById('zotero-opened-tabs-panel');
+		panel.openPopup(button, "after_start", -20, -2, false, false);
+		let toolBarButton = document.getElementById('zotero-tb-opened-tabs');
+		toolBarButton.classList.remove("inactive");
+		toolBarButton.classList.add("active");
+	};
+
+	this.onHidding = function (event) {
```
dstillman, line 343: `Hidding` → `Hiding`

chrome/skin/default/zotero/overlay.css
```diff
@@ -480,6 +480,59 @@
...
 }
 
+.img-center
+{
+	display: flex;
+	justify-content: center;
+	align-items: center;
+	background-repeat: no-repeat;
+    background-position: center; 
+    background-size: contain;
```
dstillman, line 490: Spaces → tabs

chrome/skin/default/zotero/overlay.css
```diff
@@ -480,6 +480,59 @@
...
+}
+
+.zotero-tb-open-tabs.inactive:hover
+{
+    background-image: -moz-linear-gradient(rgba(40,40,40,.3), rgba(70,70,70,.3)), url("chrome://zotero/skin/nav-arrow-down.svg"); 
+}
+
+.zotero-tb-open-tabs.active
+{
+    background-image: -moz-linear-gradient(rgba(40,40,40,.5), rgba(70,70,70,.5)), url("chrome://zotero/skin/nav-arrow-down.svg");
```
dstillman, line 533: Spaces → tabs

chrome/content/zotero/tabs.js
```diff
@@ -244,6 +246,124 @@ var Zotero_Tabs = new function () {
 		tab.iconBackgroundImage = icon.style.backgroundImage;
 	};
 
+	this._createLabel = (title, filter) => {
```
dstillman, line 249: `createTabMenuLabel`, and this can just be local scope without `this.` — not sure why the other private methods are using `this`.

chrome/content/zotero/tabs.js
```diff
@@ -244,6 +246,124 @@ var Zotero_Tabs = new function () {
...
+			}
+			// Onclick, go to selected tab + close popup
+			tabName.addEventListener("command", () => {
+				document.getElementById('zotero-opened-tabs-panel').hidePopup();
+				this.select(tab.id);
+			});
+			closeButton.addEventListener("command", () => {
+				this.close(tab.id);
+			});
+			tabName.setAttribute('id', `${tab.id}-menu`);
```
dstillman, line 325: This isn't used, is it? Can remove if not.

chrome/content/zotero/tabs.js
```diff
@@ -244,6 +246,124 @@ var Zotero_Tabs = new function () {
...
+
+	/**
+	 * Sets the data in the popup of opened tabs
+	 */
+	this._setOpenedTabsMenu = () => {
+		let openedTabsMenu = document.getElementById('zotero-opened-tabs-list');
+		// Empty existing nodes
+		while (openedTabsMenu.firstChild) {
+			openedTabsMenu.removeChild(openedTabsMenu.firstChild);
+		}
```
dstillman, line 285: Can use `openedTabsMenu.replaceChildren()` for this

chrome/content/zotero/tabs.js
```diff
@@ -244,6 +246,124 @@ var Zotero_Tabs = new function () {
...
+			
+			let tabLabel = this._createLabel(tab.title, this._openedTabsFilter);
+			tabLabel.setAttribute('crop', 'end');
+			tabLabel.setAttribute('flex', 1);
+
+			tabName.appendChild(tabLabel);
+			
+			// Selected tab is bold
+			if (tab.id == this._selectedID) {
+				tabLabel.setAttribute('class', "header");
```
dstillman, line 315: Not sure where this class is from, but better to use `tabLabel.className = 'selected'` and add a specific CSS rule for this menu.

chrome/skin/default/zotero/overlay.css
```diff
@@ -480,6 +480,59 @@
...
+{
+	background-image: -moz-linear-gradient(rgba(40,40,40,.3), rgba(70,70,70,.3)), url('chrome://zotero/skin/cross.svg'); 
+}
+
+.zotero-opened-tab-entry.title:hover
+{
+	background: -moz-linear-gradient(rgba(40,40,40,.3), rgba(70,70,70,.3)); 
+}
+
+.zotero-tb-open-tabs
```
dstillman, line 518: Move `.zotero-tb-open-tabs` rules before `.zotero-opened-tab*`

abaevbog:
Cool, so I made those changes. It now looks like [this](https://www.dropbox.com/s/7q5p8kzwnawjn0e/tab_navigation.png?dl=0).

Tab/shift-tab navigation works the same way as Firefox, except with one minor change. If you navigate to the `close` button and hit Enter, Firefox closes the tab and resets the focus as if the menu was just opened. So if you want to close multiple tabs, you have to start tab-ing from the filter bar for each of them.
I assumed it is more intuitive to move the focus to the `close` button of the next `toolbaritem` in the menu, similar to what happens with actual tabs when they are closed. For reference, [this is before](https://www.dropbox.com/s/7iv7mm20xqkaz7i/tab_delete_1.png?dl=0) hitting Enter, and [this is after](https://www.dropbox.com/s/dqgr06xjadu2hg7/tab_delete_2.png?dl=0).

dstillman:
Implemented

PR #3228, opened 2023-07-25
Add UI for editing attachment rename format pref

tnajdek:
This PR adds input box to edit `extensions.zotero.attachmentRenameFormatString` preference, including a preview for how a currently selected item would be used to derive a file name. If no item is selected, preview will use a mocked item.

`helpURL` is set to a [section](https://www.zotero.org/support/preferences/general#file_renaming) on a support page that does not yet exist. I'm happy to write a couple of paragraphs explaining how this works.

Screenshot 2023-07-25 at 18 49 02

### DIFF
--- a/chrome/content/zotero/preferences/preferences_file_renaming.js
+++ b/chrome/content/zotero/preferences/preferences_file_renaming.js
@@ -1,0 +1,66 @@
+/*
+	***** BEGIN LICENSE BLOCK *****
+    
+	Copyright © 2006–2023 Center for History and New Media
+					 George Mason University, Fairfax, Virginia, USA
+					 http://zotero.org
+    
+	This file is part of Zotero.
+    
+	Zotero is free software: you can redistribute it and/or modify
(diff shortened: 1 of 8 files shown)


dstillman:
I think in this case the help button is too subtle and we need an actual line of text with a link.

tnajdek:
* Updated to use the new `extensions.zotero.attachmentRenameTemplate` preference
* Removed help button and instead added brief instructions and a text link

Screenshot 2023-08-04 at 18 53 06

dstillman:
I've started a dedicated page for this:

https://www.zotero.org/support/file_renaming

I can add more details as necessary, but it's probably easier if you fill in the variables and parameters at the end. And I guess maybe we need some examples for conditionals, like on the [note templates](https://www.zotero.org/support/note_templates) page.

tnajdek:
Thanks! I'll edit the page and update the link in this PR.

dstillman:
Merged in d32b061 and 4e3977e

PR #3236, opened 2023-07-27
Separate attachment file renaming from title renaming

### LINKED ISSUE #3226: Allow manual file renaming by clicking on Filename field in attachment box
People should use automatic renaming, but renaming outside of Zotero and relinking via locate doesn't work (and is dangerous) if you're using "as needed" file-sync mode, since it will delete the renamed file and download from the server.

And remove "Rename associated file" from attachment title dialog, since we don't want to contribute to the idea that those always need to be the same.

AbeJellinek:
Fixed #3226


chrome/content/zotero/elements/attachmentBox.js
```diff
@@ -449,95 +459,93 @@
...
+				window,
+				'',
+				Zotero.getString('pane.item.attachments.rename.title'),
+				newTitle,
+				null,
+				{}
+			);
+			
+			// If they hit cancel or left it blank
+			if (!result || !newTitle.value) {
```
dstillman, line 473: Should add `.trim()` to make sure it's not just a space

dstillman (commented):
Now that we're using clicky for both and don't need a checkbox in the dialog, we should really switch to an editable instead of this terrible prompt, but unfortunately the Editable we use in the tags box is a React component, so I guess we can make an Editable CE and do that later (but then maybe we can do without a new localized string for now).

enzshew (approved):

dstillman:
Addressed by https://github.com/zotero/zotero-redesign/pull/148

PR #3238, opened 2023-07-29
Add main window hooks for plugins

windingwind:


windingwind:
I fixed ESLint errors and removed tabs in empty lines to make the file to be more clean. If we keep them Intentionally, I can revert these changes.

An example of `bootstrap.js` is here, which works with the build of this PR: https://github.com/windingwind/zotero-plugin-template/blob/ebddeeb3efe787aab3f17c1bf484f5f7dc746f80/addon/bootstrap.js

dstillman:
> I fixed ESLint errors and removed tabs in empty lines to make the file to be more clean. If we keep them Intentionally, I can revert these changes.

Yes, please revert all unrelated ESLint comments and whitespace changes. The `no-undef` things should be addressed properly (e.g., by defining them in the config) and the aligned tabs are intentional (we can debate, but they're intentional), and in any case unrelated changes don't belong in the same commit.

chrome/content/zotero/xpcom/plugins.js
```diff
@@ -57,22 +60,63 @@ Zotero.Plugins = new function () {
...
 			for (let addon of addons) {
 				await _callMethod(addon, 'shutdown', REASONS.APP_SHUTDOWN);
 			}
 		});
+
+		const mainWindowListener = {
+			onOpenWindow: function (aWindow) {
+				const domWindow = aWindow
+					.QueryInterface(Ci.nsIInterfaceRequestor)
+					.getInterface(Ci.nsIDOMWindowInternal || Ci.nsIDOMWindow);
```
dstillman, line 75: → `let domWindow = xulWindow.docShell.domWindow;` (and below as well)

(Also, we generally avoid `const` in Zotero code, except occasionally for explicit scalar constants.)

chrome/content/zotero/xpcom/plugins.js
```diff
@@ -57,22 +60,63 @@ Zotero.Plugins = new function () {
...
+					const { addons } = await AddonManager.getActiveAddons(["extension"]);
+					for (const addon of addons) {
+						await _callMethod(addon, 'onMainWindowLoad', REASONS.MAIN_WINDOW_LOAD, { window: domWindow });
+					}
+				}
+				domWindow.addEventListener(
+					"load",
+					onload,
+					false,
+				);
```
dstillman, line 93: Can just be on one line unless it's >100 chars (and same throughout)

dstillman (changes requested):

chrome/content/zotero/xpcom/plugins.js
windingwind: Done

chrome/content/zotero/xpcom/plugins.js
windingwind: Done

PR #3251, opened 2023-08-05
sentence-caser

retorquere:
sentence caser that groks more word-ish variants and takes markup into account.

### DIFF
--- a/chrome/content/zotero/elements/itemBox.js
+++ b/chrome/content/zotero/elements/itemBox.js
@@ -66,7 +66,7 @@
 
 			// restore protected parts from original, always last in the preserve list because
 			// it can span over other HTML
-			text.replace(/<span class="nocase">.*?<\/span>|<nc>.*?<\/nc>/gi, (match, i) => {
+			this.input.replace(/<span class="nocase">.*?<\/span>|<nc>.*?<\/nc>/gi, (match, i) => {
 				preserve.push({ start: i, end: i + match.length - 1 });
 				return match;
 			});
--- a/chrome/content/zotero/elements/itemBox.js
+++ b/chrome/content/zotero/elements/itemBox.js
@@ -26,6 +26,112 @@
 "use strict";
 
 {
+	const sentenceCaser = new class SentenceCaser {
+		constructor() {
+			this.acronym = /^(\p{Lu}[.])+(?=$|[\P{L}])/; // U.S.
+			this.quoted = /^"[^"]+"(?=$|[\P{L}])/; // "Hello There"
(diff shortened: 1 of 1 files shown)


dstillman:
Thanks! So to use this, we'd actually need to move it to utilities.js in https://github.com/zotero/utilities, around about `capitalize()`, and add tests for various input string in https://github.com/zotero/utilities/blob/master/test/tests/utilitiesTest.js. That will let us provide the same functionality across Zotero tools (the web library, ZoteroBib, etc.). Once it's in utilities, hooking it up to the item box would be trivial.

retorquere:
Do you want me to prepare a PR for utilities instead?

dstillman:
Well, I'm a bit confused by this, though. Is it not a much simpler problem? Is it not, just, split the string by whitespace, and for each substring, if anything past the first alphabetic character is a capital letter, ignore it, and otherwise lowercase it? That would leave `U.S.`, `McDonald`, `USA`, and `Q&A` alone.

What input are you handling that wouldn't be handling by that?

retorquere:
I've ran that on my test set, it would miss (a.o.)

* All (sub) sentence starts
* single-letter cases (Programming in R: A beginners guide)
* multipart words like Meta-Analysis
* words inside quotes

If attached a diff ran on my test suite as diff1.txt. But I think you're right that heuristic should be able to simplify some things.

[diff1.txt](https://github.com/zotero/zotero/files/12267410/diff1.txt)

retorquere:
Spitting on more characters gets much better results though. I have to admit I built out this sentence-caser as I found new counter examples.

[diff1.txt](https://github.com/zotero/zotero/files/12267436/diff1.txt)

northword:
https://github.com/northword/zotero-format-metadata/blob/main/src/utils/str.ts

Another implementation is available for reference. In this implementation, chemical elements and html tags (including subscripts, bold, italics, etc. are kept in case) are taken into account.
Comments contain part of the logic (Chinese), the code logic may be a little confusing, I haven't taken the time to refactor it, it would be better if the new zotero functions can support more.

retorquere:
Maybe some of these idea can carry over? But I'd suggest e.g. testing for `\p{Lu}` rather than `A-Z`, and I think `string.charAt` would have problems with multibyte characters.

retorquere:
My implementation also takes HTML markup into account, it leaves it unscathed.

retorquere:
I've tried adding chem elements detection to the sentencecaser, but it flags "No" as an element instead of as the adverb/adjective/noun.

PR #3307, opened 2023-08-10
trash functionality for collections - no date deleted

abaevbog:
Deleted collections go to trash, similar to items and show up among them. Subcollections of a deleted collection are also trashed though only the parent collection is displayed in trash itemTree.

When a parent collection is permanently deleted or restored, same applies to subcollections.

Items are not affected by trashing/restoration/permanent deletion of collections.

### DIFF
--- a/chrome/content/zotero/collectionTree.jsx
+++ b/chrome/content/zotero/collectionTree.jsx
@@ -848,114 +848,116 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 		}
 		else if (action == 'modify') {
 			let row;
-			let id = ids[0];
-			let rowID = "C" + id;
-			let selectedIndex = this.selection.focused;
 			
-			let handleFocusDuringSearch = async (type) => {
-				let object = type == 'collection' ? Zotero.Collections.get(id) : Zotero.Searches.get(id);
-				// If collections/searches are being filtered, some rows
-				// need to be (un-)greyed out or removed, so reload.
-				if (!this._isFilterEmpty()) {
-					let offset = 0;
-					if (!this._includedInTree(object, true)) {
-						offset = this._calculateOffsetForRowSelection(type[0].toUpperCase() + id);
-					}
-					await this.reload();
-					this._selectAfterRowRemoval(selectedIndex - offset);
-				}
-			};
-
-			switch (type) {
-			case 'collection':
-				let collection = Zotero.Collections.get(id);
-				row = this.getRowIndexByID(rowID);
-				// If collection is visible
-				if (row !== false) {
-					// TODO: Only move if name changed
-					let reopen = this.isContainerOpen(row);
-					if (reopen) {
-						this._closeContainer(row);
-					}
-					this._removeRow(row);
-					// Collection was moved to trash, so don't add it back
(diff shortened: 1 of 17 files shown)


dstillman:
OK, see https://github.com/zotero/zotero/commit/9dd182e9ca26f2197665a392b7634f8f96038473.

There were various bugs in the collection tree, including some bugs and/or badly resolved conflicts from the [big tree rewrite](https://github.com/zotero/zotero/commit/cbbff600a60c9e7a7407d6f2e4053309bf28b872#diff-f968f0f54d7b20fb861ace7015ce6212d44b4060184729ba4f8055f345a6bb47) a couple years ago, that caused `.deleted` not to be handled properly in various ways. (Searches in the trash also weren't being hidden from the tree, which you fixed, but is now fixed in a way that's consistent with collections.)

If you rebase on that, I think you'll find you can greatly simplify what you're doing here. For example, there shouldn't be any need for a `moveToID` parameter in `_addSortedRow()`, since that function shouldn't have selection logic at all. The easy test here was something like this:

```javascript
var c = Zotero.Collections.getByLibraryAndKey(1, '[collection key]')
c.deleted = !c.deleted;
await c.saveTx();
```

Running this from the trash would cause the selection to change, which it shouldn't have, because a collection modification when something else is selected shouldn't affect selection. So the first step was just fixing that, unrelated to any UI work to support collections in the trash. And now you should be able to restore collections and searches from the trash without worrying about selection.

abaevbog:
I went over the changes you pointed out and cleaned up a few other smaller things I ran into. I ended up squashed everything into a single commit during the cleanup because a lot of later commits undid earlier changes.

I moved  methods that collections and searches use to pretend to be items in itemTree to `Zotero.DataObjectUtilities`, since they are kinda like helper functions for our data objects. Does that make sense? 

Alternatively, they could go into the actual `Zotero.DataObject`. `Zotero.Item`, `Zotero.Collection` and `Zotero.Search` all  inherit from it, so it would be straightforward. It just seemed too specific to this trash-related feature to go there.

dstillman:
I've rebased and updated this in https://github.com/zotero/zotero/tree/trash-collections. I can't force-push it back to this PR because of Git LFS changes since this was created, though you should be able to.

The iOS app isn't properly handling `deleted` for collections and searches, so we'll need to wait for that to be fixed before we can merge this.

This is also affected by https://github.com/zotero/zotero/issues/4115 — the restore/delete buttons don't always appear properly in the `ItemMessagePane`, though they're available in the context menu.

abaevbog:
Got it! I'll rebase my initial commit onto `master` to have all conflicts resolved and push it here with https://github.com/zotero/zotero/commit/c537d1ad28b7fc725efa68efb3bdb5c8afd8222d

abaevbog:
I'm actually noticing a glitch where a deleted collection may sometimes disappear from the trash when another collection is deleted or restored. I'll look into this while we're waiting on the iOS app


https://github.com/zotero/zotero/assets/36271954/b310ab15-2fd5-49d2-b04f-d7568bb42a5d

dstillman:
Merged

abaevbog:
Awesome!

PR #3310, opened 2023-08-11
Fix type of `options.helpurl` in preferences pane

northword:
When registering the preferences pane, `options.helpurl` should be `string` instead of `string[]`.


dstillman:
Thanks!

Issue #3324, opened 2023-08-15
Use any emoji from tag in middle pane


dstillman:
You shouldn't have to make a tag a colored tag to use the emoji, and you should be able to include text as well.

Only question is whether happens if multiple emoji are separated by plain text. Are all the emoji used? Just the first contiguous span?

asan-emirsaleh:
What is the progress? Is this feature available on Zotero beta now?

asan-emirsaleh:
In case you need some help in testing this feature one possible approach would be to implement it in beta, with allowing it be optionally activated in Config editor. I would participate in testing with a pleasure.

PR #3330, opened 2023-08-16
display emojis from color-less tags in itemTreeRow

abaevbog:
Uses the first continuous span of emojis

Fixes: #3324


dstillman (changes requested):

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -411,10 +411,16 @@ Zotero.Utilities.Internal = {
...
-	isOnlyEmoji: function (str) {
-		// Remove emoji, Zero Width Joiner, and Variation Selector-16 and see if anything's left
-		const re = /\p{Extended_Pictographic}|\u200D|\uFE0F/gu;
-		return !str.replace(re, '');
+	containsEmoji: function (str) {
+		const re = /\p{Extended_Pictographic}/gu;
+		return !!str.match(re);
+	},
+
+	renderTagWithEmojis: function (str) {
```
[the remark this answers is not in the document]
abaevbog: I agree - good point.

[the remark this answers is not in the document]
dstillman: Better idea! Something like `Zotero.Tags.extractEmojiForItemsList()`? No need to clutter Utilities.Internal.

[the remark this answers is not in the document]
abaevbog: Makes sense! I think `containsEmoji` should stay with`extractEmojiForItemsList` because they are related, so in the last commit I suggest we move `containsEmoji` into `Zotero.Tags` too. 
Both of them are only used for tags, so it seems logical?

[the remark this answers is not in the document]
dstillman: No, I think `containsEmoji` should stay put. There's nothing tag-specific there, and it could easily be used in other contexts even if it's not now.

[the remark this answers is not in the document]
abaevbog: Understood! I redid and force-pushed the last commit with only emoji extraction moved to `Zotero.Tags`

chrome/content/zotero/xpcom/data/tags.js
```diff
@@ -945,7 +945,14 @@ Zotero.Tags = new function() {
...
 		}
 	}
-	
+
+	// Return the first sequence of emojis from a string
+	this.extractEmojiForItemsList = function (str) {
+		// Split by anything that is not an emoji, Zero Width Joiner, or Variation Selector-16
+		// And return first continuous span of emojis
+		const re = /[^\p{Extended_Pictographic}\u200D\uFE0F]+/gu;
+		return str.split(re).filter(Boolean)[0];
```
dstillman, line 954: Maybe a little cleaner to add ` || null` here to return `null` instead of `undefined` if there's no emoji.

chrome/content/zotero/itemTree.jsx
```diff
@@ -3821,12 +3821,12 @@ var ItemTree = class ItemTree extends LibraryTree {
...
 		span.className = 'tag-swatch';
-		// If only emoji, display directly
+		// If contains emojis, display directly
 		//
 		// TODO: Check for a maximum number of graphemes, which is hard to do
 		// https://stackoverflow.com/a/54369605
-		if (Zotero.Utilities.Internal.isOnlyEmoji(tag)) {
-			span.textContent = tag;
+		if (Zotero.Utilities.Internal.containsEmoji(tag)) {
+			span.textContent = Zotero.Tags.extractEmojiForItemsList(tag);
```
dstillman, line 3829: I guess we don't really need `containsEmoji` at all, right? It's more or less doing the same regex, so why do it twice instead of once?

chrome/content/zotero/xpcom/data/item.js
```diff
@@ -4398,7 +4398,9 @@ Zotero.Item.prototype.getColoredTags = function () {
 	let tagColors = Zotero.Tags.getColors(this.libraryID);
 	for (let tag of tags) {
 		let data = tagColors.get(tag.tag);
-		if (data) {
+		let containsEmoji = Zotero.Utilities.Internal.containsEmoji(tag.tag);
+		// Include tags that contain emojis even if they are not assigned a color
+		if (data || containsEmoji) {
```
dstillman, line 4403: Doesn't seem right that this is still called `getColoredTags()`. Since it will now return both colored tags and emoji tags, we should probably rename it to `getItemsListTags()`.

And then we're now sorting against `undefined` below for emoji tags, which means the comparison function is actually returning NaN, which is equivalent to a return value of `0` in JS, which means that emoji tags without a position defined can be sorted in the middle of the list. We need to check for `.position` and return tags with positions first. Tags without positions should return 0 against each other to remain in their sorted order. (And we should add a test for this sorting.)

dstillman (changes requested):

chrome/content/zotero/xpcom/data/item.js
abaevbog: Right! I added proper sorting in the last commit. I thought it may be more readable to aggregate and sort all colored tags (just like it is done now) first and then just `concat` tags with emojis to the end of that array instead of collecting all tags into one array and sorting them together.

abaevbog:
`getTagColors` is deprecated and seems to not be used anywhere but it calls newly renamed `getItemsListTags`. 

I tweaked it to return just the colors of tags that are assigned color in case it is used by plugins or something like that?

chrome/content/zotero/xpcom/data/item.js
```diff
@@ -4380,31 +4380,36 @@ Zotero.Item.prototype.getImageSrc = function() {
 
 
 Zotero.Item.prototype.getTagColors = function () {
-	Zotero.warn("Zotero.Item::getTagColors() is deprecated -- use Zotero.Item::getColoredTags()");
-	return this.getColoredTags().map(x => x.color);
+	Zotero.warn("Zotero.Item::getTagColors() is deprecated -- use Zotero.Item::getItemsListTags()");
+	return this.getItemsListTags().filter(x => x.color).map(x => x.color);
```
dstillman, line 4384: Let's just remove this function.

chrome/content/zotero/xpcom/data/item.js
```diff
@@ -4380,31 +4380,36 @@ Zotero.Item.prototype.getImageSrc = function() {
...
+		if (data) {
+			colorData.push({ tag: tag.tag, ...data });
+		}
+		else if (containsEmoji) {
+			tagsWithEmojis.push({ tag: tag.tag });
 		}
 	}
-	return colorData.sort((a, b) => a.position - b.position).map(x => ({ tag: x.tag, color: x.color }));
+	// Sort colored tags by their position and add tags with emojis in the end.
+	const coloredTagsWithEmojis = colorData.sort((a, b) => a.position - b.position).concat(tagsWithEmojis);
```
dstillman, line 4411: We avoid `const` in Zotero code for everything but actual scalar constants (e.g., `let MAX_PAGES = 5;`), so all of these can be changed to `var` or `let` (and always `let` in block scope — e.g., within an `if` or `for`).

dstillman (changes requested):
A couple more notes, and one very minor thing: if I add a new emoji tag to an item, it's added to the end of the tags in the items list row. If I then switch collections and switch back, it's sorted in its proper order.

abaevbog:
> If I then switch collections and switch back, it's sorted in its proper order.

Oh, right. I expected `this.getTags()` would always return tags in the same order which is not the case.
I added sorting for emoji tags to show up in the same order as tags in the `tagsBox.js`.

While we are at it, I wonder if it makes sense to tweak the sorting in `tagsBox.js` to show colored tags at the top before all the other tags, just like we do it now in `Zotero.Item.getItemsListTags()`? That way, the relative order of all tags will be 100% consistent between tags in the `content-pane` and colored/emoji tags in `itemTree`.

dstillman:
> While we are at it, I wonder if it makes sense to tweak the sorting in tagsBox.js to show colored tags at the top before all the other tags, just like we do it now in Zotero.Item.getItemsListTags()?

Yeah, I was wondering the same. Probably? If nothing else, that might just be easier to read — it's sort of distracting and confusing now with colored tags mixed in.

abaevbog:
> > While we are at it, I wonder if it makes sense to tweak the sorting in tagsBox.js to show colored tags at the top before all the other tags, just like we do it now in Zotero.Item.getItemsListTags()?
> 
>  it's sort of distracting and confusing now with colored tags mixed in.

Fully agree - I moved ordering into a separate function `Zotero.Tags.compareTagsOrder`, and it's used both by tagsBox and `Zotero.Item.getItemsListTags`, so the relative order of colored tags ordered by position first, followed by other tags alphabetically should be the same everywhere.

Meanwhile, I ran into what looks like another regression - if I have a tagsBox for an item opened, and I assign color to one of the tags of this item, `Zotero.Tags.getColors(this.item.libraryID)` in `reload()` is not aware that one of the tags was assigned color, so it doesn't show that tag as "colored" until you click away from tagsBox and back. Similar thing happens if I rename a colored tag. Should I look into it in this PR or start a new one?

PR #3337, opened 2023-08-17
Update zh-CN translations

lychichem:
Some strings are missing in transifex, I have to add them in this github repo. And I don't know why you revert some translation in my former PR (and on transifex translated by others). And I fixed some strings on transifex (and here).


dstillman:
I'm not sure what PR you're referring to — this seems to be your first contribution.

All localization happens on Transifex, and we're not able to accept changes to non-English strings here. Changes on Transifex will be incorporated into a future release. We're grateful for any contributions you want to make there.

Not all new strings become available on Transifex immediately. The renaming strings will be made available on Transifex soon.

Previously localized strings would get reverted if the original English string is changed. That's it.

("Mendeley" is misspelled in one of these strings, by the way.)

Issue #3334, opened 2023-08-17
Tag editable weirdness


dstillman:
Various strange things happen in the item Tags pane.

 - If you save a tag and then click back in it and press Shift-Enter to turn on multiline mode, press Esc, and then click Add Tag, the new tag editor opens down below where the multiline textbox was. Same if you press Shift-Enter in a tag in the middle of the list and then press Esc.
 - After that, clicking back in the editor and press Return no longer works to save the field, until you switch away from the item and back.
 - Shift-Tab doesn't work properly. For the first tag, it should focus Add Tag — not sure where focus goes now. For existing tags after the first one, it should focus the previous tag, whereas now it creates a new empty row. For a new empty row, focus doesn't change, when it should close the editor and focus the last tag.
 -  Possibly other things?

I think these are mostly regressions in Z7, though I didn't confirm.

(I'm not totally sure why Shift-Enter enables multiline mode, since you can already paste newline-separated tags into the field, which expands to accommodate, but I guess it's fine. Maybe if you're entering a lot then the constant refreshing/resorting is a little annoying?)

This is all in tagsBox.js.

abaevbog:
This PR fixes the multiline-related issues and tab-based navigation issues from above. 
I'll play around to see if there is anything else

abaevbog:
Found another strange behavior of the multiline textbox. When new tags are being added at the bottom, activating the multiline mode with shift-enter, adding a tag, hitting enter to add a new line, and then shift-enter makes the field to stick around and sometimes occupy space in weird ways.

[Screen recording without changes from this PR](https://www.dropbox.com/scl/fi/6usxfqr2jjo1u3wpgz1wv/tag_weirdness_before_pr.mov?rlkey=l7h75l8vb0fmap69saab579lh&dl=0).
[Screen recording with changes from this PR](https://www.dropbox.com/scl/fi/h0ndic77x94t72uda9q9z/tag_weirdness.mov?rlkey=fsuyk8g50p4z8itlju0q8a7vd&dl=0).

abaevbog:
Another issue: pasting a small # of lines into a textfield causes this

<img width="1112" alt="Screenshot 2023-08-18 at 3 19 59 PM" src="https://github.com/zotero/zotero/assets/36271954/a43d0f2e-5591-4b6c-92a5-5dc5b2d417c3">

And pasting a large # of lines causes this:

[...]

Issue #3250, opened 2023-08-05
Zotero 7 breaks screen reader accessibility


dstillman:
https://forums.zotero.org/discussion/106673/beta-7-on-windows-breaks-screen-reader-accessibility

Reported for Windows, but related to https://github.com/zotero/zotero/issues/2736?

Issue #3019, opened 2023-03-10
Focus items list on Return or Esc from item field


dstillman:
https://forums.zotero.org/discussion/comment/430056/#Comment_430056

Pretty sure this is a regression.

PR #3338, opened 2023-08-17
Accessibility and tab-navigation updates

abaevbog:
- Prevent multiline field from staying around and occupying space 
- Allow tab/shift-tab navigation through tags
- Added aria roles and labels to fields in itemBox to make it accessible to Wooshy and most screen readers (VoiceOver)
- CSS changes to itemBox to avoid `display:contents` and `display:grid` to make itemBox work with NVDA screen reader on windows.
- From the last filed in itemBox, move focus to `New Collection` button
- Set text direction via `style.direction` to show `...` on the correct side of the long text fields. (e.g. show `...` on the left for long text fields instead of right when `Zotero.rtl` is true.
- When ESC/Enter is pressed from items or tags pane, focus is returned to items list

Fixes: #3250
Fixes: #3334
Fixes: #3019

### DIFF
--- a/chrome/content/zotero/elements/tagsBox.js
+++ b/chrome/content/zotero/elements/tagsBox.js
@@ -368,7 +368,7 @@
 					if (event.button) {
 						return;
 					}
-					this.clickHandler(event.target, 1, valueText);
+					this.clickHandler(event.target, false, valueText);
 				}, false);
 				valueElement.className += ' zotero-clicky';
 			}
@@ -398,7 +398,7 @@
 			return valueElement;
 		}
 
-		showEditor(elem, rows, value) {
+		showEditor(elem, multiline, value) {
 
 			// Blur any active fields
 			/*
@@ -414,7 +414,7 @@
 
 			var itemID = this._item.id;
 
-			var t = document.createElement(rows > 1 ? 'textarea' : 'input', { is: 'shadow-autocomplete-input' });
+			var t = document.createElement(multiline ? 'textarea' : 'input', { is: 'shadow-autocomplete-input' });
 			t.setAttribute('class', 'editable');
 			t.setAttribute('value', value);
 			t.setAttribute('fieldname', fieldName);
@@ -422,8 +422,8 @@
 			t.setAttribute('ignoreblurwhilesearching', 'true');
 			t.setAttribute('autocompletepopup', 'PopupAutoComplete');
 			// Multi-line
-			if (rows > 1) {
-				t.setAttribute('rows', rows);
+			if (multiline) {
+				t.setAttribute('rows', 6);
 			}
 			// Add auto-complete
 			else {
@@ -526,14 +526,9 @@
 					// Reset field to original value
 					target.value = target.getAttribute('value');
 
-					var tagsbox = focused.closest('.editable');
-
 					this._lastTabIndex = false;
 					await this.blurHandler(event);
 
-					if (tagsbox) {
-						tagsbox.closePopup();
-					}
 
 					// TODO: Return focus to items pane
 					var tree = document.getElementById('zotero-items-tree');
@@ -576,13 +571,9 @@
 			}
 		};
 
-		makeMultiline(textbox, value, rows) {
+		makeMultiline(textbox, value) {
 			textbox.parentNode.classList.add('multiline');
-			// If rows not specified, use one more than lines in input
-			if (!rows) {
-				rows = value.match(/\n/g).length + 1;
-			}
-			textbox = this.showEditor(textbox, rows, textbox.getAttribute('value'));
+			textbox = this.showEditor(textbox, true, textbox.getAttribute('value'));
 			textbox.value = value;
 			// Move cursor to end
 			textbox.selectionStart = value.length;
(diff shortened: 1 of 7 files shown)


dstillman:
The accessibility improvements are great, but this is totally changing the spacing of the item box. It needs to stay the same.

Before:

Screenshot 2023-08-30 at 12 33 31 AM

After:

Screenshot 2023-08-30 at 12 34 22 AM

dstillman:
A few more bugs I noticed (probably not new):

- With the Tags tab open, Shift-Tab back from the search bar results stops at the collections button and throws an error: `JavaScript error: chrome://zotero/content/zoteroPane.js, line 194: TypeError: tagContainer is null`
- If I change items while Info is selected, and then click the Tags tab, and then press Tab, the Add button is skipped. It works on subsequent attempts.
- If I click on the first tag and press Shift-Return, there's no focus ring around Add, even though it does seem to be focused (and space bar works). If I press Shift-Tab again to move to the Tags tab and then press Tab again, Add gets the focus ring, and it works after that. I can get it to happen again by clicking in the items list and then back on the first tag.

dstillman:
- If you delete the contents of a tag in the editor and press Tab, focus skips a tag.

chrome/content/zotero/elements/tagsBox.js
```diff
@@ -414,16 +414,16 @@
...
 			t.setAttribute('value', value);
 			t.setAttribute('fieldname', fieldName);
 			t.setAttribute('ztabindex', tabindex);
 			t.setAttribute('ignoreblurwhilesearching', 'true');
 			t.setAttribute('autocompletepopup', 'PopupAutoComplete');
 			// Multi-line
-			if (rows > 1) {
-				t.setAttribute('rows', rows);
+			if (multiline) {
+				t.setAttribute('rows', 6);
```
dstillman, line 426: What's 6? Better to put magic numbers in a separate variable (and this can be `const`) for clarity.

Does this change any behavior?

abaevbog: 6 is the number of rows the multiline field has. It 100% should be a const variable for readability.

Before, the number of rows came either from the [number of new lines on paste](https://github.com/zotero/zotero/blob/794e89d307f62d5073e8bc08535fe7e85895992f/chrome/content/zotero/elements/tagsBox.js#L583) or it was [hardcoded here](https://github.com/zotero/zotero/blob/794e89d307f62d5073e8bc08535fe7e85895992f/chrome/content/zotero/elements/tagsBox.js#L479), which is the number of rows that would be used on shift-enter. 

I noticed that the fact that the number of rows of the textfield varies can break things - the multiline field can overlap with tags below of leave too much space before them. There's a few screenshots of this happening [here](https://github.com/zotero/zotero/issues/3334#issuecomment-1684343408). I thought the number of rows the multiline field occupies should always be the same to avoid this.

abaevbog:
@dstillman 
> The accessibility improvements are great, but this is totally changing the spacing of the info pane. It needs to stay the same.

Thank you for pointing it out. Spacing and other stylistic details are tricky for me. As I was working on fixing the spacing between rows and columns, I noticed I accidentally made a change to how the table is "placed".

Currently, the table is located about 10px from where the first label starts. That means that all labels are located at different spots for different item types. With the last css changes to itemBox, the table's location is fixed so the common labels (e.g. item type) ended up in the same places regardless of item types. I, personally, liked it because if you quickly click through items, the labels don't "bounce around" but stay in the same place. So I am wondering if we want to replicate the old behavior of itemBox?

Current behavior:
https://github.com/zotero/zotero/assets/36271954/01c72cdf-a8a1-42e2-a836-4762441e61a2

With these changes:
https://github.com/zotero/zotero/assets/36271954/c28d2564-ca92-48bd-9326-afdbe8ed75bc

dstillman:
Yeah, that's better, and it's what we do in the web library.

abaevbog:
Got it. In that case, there's the following question. Some locales have much longer titles than others. With longer titles (e.g. Bulgarian), the padding on the left has to be quite big to allow the entire label to fit on one line (which is how it works now). For this internet page's labels to fit on one line, it needs to be pushed away quite a bit.

long_name

Which means that other items that don't have such long labels (and other locales whose labels are not as long) have to have the same padding from the left. Does it seem like there is a lot of empty space on the left in that case?

empty_space

dstillman:
> The web library cuts the labels off if they get too long but I doubt we want to do it on desktop

No, I think we should, and those locales should just be corrected. The localizers don't necessarily know there's a length concern here, but, e.g., in English we use "Loc. in Archive" because of this.

abaevbog:
Great! I just pushed fixes to the bugs you pointed out earlier and spacing adjustments with labels that are too long being cutoff. 

There were 2 odd things to mention: 
1. when a new tab in the contextPane is selected, the focus is shifted onto the first child without the `focus-ring` (in this case, "new tag" button), which made it confusing where the focus is and make it seem like the button is skipped. It comes from `_selectNewTab chrome://global/content/elements/tabbox.js:740`, so I assumed it was mozzilla and worked around it by bouncing focus off from the "new tag" button based on `explicitOriginalTarget`.
2. Shift-tab from the first Tag actually focuses on the "new tag" button but the `focus-visible`was not set. This is why the button worked but there was not `focus-ring`. Toggling `.contentEditable` is the workaround I came up with to force `focus-ring` to show up.

PR #3341, opened 2023-08-18
ReaderSelectionAPI

windingwind:


windingwind:
this pr is based on #3290

chrome/content/zotero/xpcom/reader.js
```diff
@@ -962,6 +963,52 @@ class ReaderInstance {
...
+			}
+		}
+		else {
+			selection ??= this._internalReader._lastView._getAnnotationFromTextSelection(
+				"highlight"
+			);
+		}
+		let { sortIndex, pageLabel, position, text } = selection;
+		return { sortIndex, pageLabel, position, text };
+	}
```
mrtcode, line 1011: I think it's too fragile to use private methods and variables this way, and if we want to get something from the reader or its views, we have to implement proper public methods in all views and the main reader API in reader.js, to get specific selection data that we want.

Also `_getAnnotationFromSelectionRanges` and `_getAnnotationFromTextSelection` just create an annotation from selection data, so I'm not sure we can call this selection.

What exactly plugin developers want from the selection? Do they want to position a popup? Or do they just need text? Annotation?

chrome/content/zotero/xpcom/reader.js
```diff
@@ -1441,7 +1442,6 @@ class Reader {
 		if (!openInBackground) {
 			reader._window.focus();
 		}
-		return reader;
```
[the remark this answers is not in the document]
windingwind: sorry my bad - reverted already

chrome/content/zotero/xpcom/reader.js
```diff
@@ -599,7 +599,7 @@ class ReaderInstance {
 		// this._postMessage({ action: 'focusLastToolbarButton' });
 	}
 
-	tabToolbar(reverse) {
+	tabToolbar(_reverse) {
```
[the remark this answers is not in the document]
windingwind: reverted

chrome/content/zotero/xpcom/reader.js
windingwind: The general need can be concluded as:

1. get selected text.
2. get a selection position, mainly for jumping back to the selection position
3. creating custom annotation from the selected text

mrtcode: Aren't plugin developers also interested in listening evens i.e. when text is selected or selection changes? Are they just listening for pointerup events instead?

windingwind: It would be good if we add that. currently they use pointerup

mrtcode: I suppose pointerup has to be added to the iframes of both views, otherwise they won't receive the event. And the second view (split view) and be created/destroyed at any time, which means plugins have to constantly check that and re-attach event listeners. Is this how plugins currently work? If so, we definitely need to make this simpler.

Selection can also be modified with Shift + arrow keys, even though this is currently not functioning in PDF view.

windingwind: yes. a mutation observer on the container element of the views to dynamically watch newly added views.

mrtcode: Yes, I was also thinking about that. Allowing to hook into selection popup where plugin receives selection data and a callback that appends a custom HTML section into selection popup. This way multiple plugins could work together and can be implemented easily. @dstillman does it sound good?

I suppose getting the current selection is basically the most important thing that plugins want from the reader. What else do plugins need from the reader?

windingwind: Some plugins may also want to:
1. add a left-sidebar tab panel, alongside the annotations;
2. add buttons to the annotations in the annotation tab panel;
3. add buttons to the top-side tools bar.

chrome/content/zotero/xpcom/reader.js
```diff
@@ -962,6 +964,52 @@ class ReaderInstance {
...
+	 * let reader = Zotero.Reader._readers[0];
+	 * let selectedText = reader.getLastSelection().text || "";
+	 * ```
+	 */
+	getLastSelection() {
+		let selection = {};
+		if (this._internalReader._type === "pdf") {
+			let selectionRanges = this._internalReader._lastView._selectionRanges;
+			if (selectionRanges.length !== 0) {
+				selection ??= this._internalReader._lastView._getAnnotationFromSelectionRanges(
```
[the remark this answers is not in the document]
windingwind: Right, thanks for pointing it out. But I am unsure if this PR is still ready to go, as we might want to have a bigger plan about reader APIs for the plugins, as @mrtcode and I discussed above.

windingwind:
Since we have #3373 , close here

Issue #3340, opened 2023-08-18
Adjust filename renaming preview for selected notes and attachments


dstillman:
https://forums.zotero.org/discussion/107000/zotero-7-wrong-preview-of-filename-format-when-selecting-a-child-item

Notes, linked-URL attachments, and standalone attachments of any type shouldn't be used at all.

Child file attachments should use the parent item metadata, and we might as well use the proper file extension.

PR #3343, opened 2023-08-18
Improve file renaming preview in preferences

tnajdek:
Addresses #3340

The new logic is:

* Take selected item (or first selected item, if more than one is selected)
* If it's a regular, top-level item, use that item, assume default extension (`.pdf`)
* Otherwise, if it's an attachment that has a parent and is either imported or a linked file, use parent item and this attachment's extension
* Otherwise fall back to the mock item and default extension


dstillman:
Not sure if you intentionally weren't showing this for snapshots (`IMPORTED_URL`), but particularly now that we have snapshot annotation, it probably makes sense to treat those as potential primary attachments that would be renamed.

tnajdek:
No, not intentionally, it definitely makes sense to include that, `isFileAttachment()` works well here.

Issue #3353, opened 2023-08-20
Reopened tabs appear in different order


dstillman:
https://forums.zotero.org/discussion/107063/zotero-7-reopened-tabs-appear-in-different-order

I can reproduce this.

PR #3357, opened 2023-08-21
Set opened reader tab index when state is restored

abaevbog:
Fixes: #3353

### DIFF
--- a/chrome/content/zotero/tabs.js
+++ b/chrome/content/zotero/tabs.js
@@ -156,6 +156,7 @@ var Zotero_Tabs = new function () {
 							null,
 							{
 								title: tab.title,
+								tabIndex: i,
 								openInBackground: !tab.selected,
 								secondViewState: tab.data.secondViewState
 							}


mrtcode:
I'm not sure why just calling `Zotero.Reader.open` in specific order no longer opens the tabs in the same order, but if setting the `pageIndex` helps then great. I guess the issue is related with unloaded tab loading which is poorly working in Zotero 7 and needs rewrite.

Anyway, we can merge this.

abaevbog:
@mrtcode My guess as to why this happens is that the `open` method in [reader.js](https://github.com/zotero/zotero/blob/85cade3fb295bd2f0b63cdab5f78a327818e3c55/chrome/content/zotero/xpcom/reader.js#L1367) is asynchronous. But `this.restoreState` in [tabs.js](https://github.com/abaevbog/zotero/blob/17aedbc5affeea3e8b9a226e97b3a2431eefbc0a/chrome/content/zotero/tabs.js#L146) that calls `Zotero.Reader.open` is not, and it does not `await` for the the reader tab to be opened. 
In fact, commenting out the async line `await library.waitForDataLoad('item');` from `Zotero.Reader.open` seems to preserve the order of tabs as well.

PR #3358, opened 2023-08-22
pref to reopen tabs in Zotero_Tabs.restoreState

### LINKED ISSUE #3348: Add pref for reopening tabs
https://forums.zotero.org/discussion/107045/turn-off-automatically-reopening-pdfs-from-last-session

Browsers do it, so we might as well do it too.

### LINKED ISSUE #3376: Reader windows don't reopen on restart

abaevbog:
Fixes: #3348
Fixes: #3376

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -654,6 +654,21 @@ var ZoteroPane = new function()
 			if (state) {
 				Zotero_Tabs.restoreState(state.tabs);
 			}
+			// Reopen note windows
+			(async () => {
+				let openedNoteWindows = Zotero.Session.state.windows.filter(x => x.type == 'note');
+				for (let { itemID, collectionID, parentItemKey } of openedNoteWindows) {
+					// Wait for the item to be loaded
+					await Zotero.Items.getAsync(itemID);
+					// Wait for the collectionTree to be loaded, since this.canEdit() used in openNoteWindow relies on it.
+					let waitCount = 0;
+					while (!ZoteroPane.collectionsView.getRow(ZoteroPane.collectionsView.selection.focused) && waitCount < 100) {
+						await Zotero.Promise.delay(100);
+						waitCount += 1;
+					}
+					ZoteroPane.openNoteWindow(itemID, collectionID, parentItemKey);
+				}
+			})();
 		}
 		catch (e) {
 			Zotero.logError(e);
@@ -4454,9 +4469,9 @@ var ZoteroPane = new function()
 		var name = null;
 		
 		if (itemID) {
-			let w = this.findNoteWindow(itemID);
-			if (w) {
-				w.focus();
+			let w = Zotero.Notes.getWindows(itemID);
+			if (w.length) {
+				w[0].focus();
 				return;
 			}
 			
@@ -4471,18 +4486,6 @@ var ZoteroPane = new function()
 		window.openDialog('chrome://zotero/content/note.xhtml', name, 'chrome,resizable,centerscreen,dialog=false', io);
 	}
 	
-	
-	this.findNoteWindow = function (itemID) {
-		var name = 'zotero-note-' + itemID;
-		var wm = Services.wm;
-		var e = wm.getEnumerator('zotero:note');
-		while (e.hasMoreElements()) {
-			var w = e.getNext();
-			if (w.name == name) {
-				return w;
-			}
-		}
-	};
 	
 	
 	this.addAttachmentFromURI = Zotero.Promise.method(function (link, itemID) {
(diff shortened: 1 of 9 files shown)


defaults/preferences/zotero.js
```diff
@@ -214,6 +214,7 @@ pref("extensions.zotero.scaffold.eslint.enabled", true);
 
 // Tabs
 pref("extensions.zotero.tabs.title.reader", "titleCreatorYear");
+pref("extensions.zotero.tabs.reopenOnRestart", true);
```
dstillman, line 217: I'm realizing that this is a little more complicated. While we don't currently reopen reader or note windows (or, in the future, additional library windows) on restart, we obviously should, and the same pref should probably apply. So something like `extensions.zotero.restoreState` is probably better, moved up above `reopenPanesOnRestart`.

abaevbog:
I think that my initial approach of ignoring session data if `Zotero.Prefs.get('restoreState') is `false` was wrong.
Simply ignoring the session data meant that changing the preference to not restore tabs, closing/reopening the app, changing the preference to `true` and closing/reopening again could bring back the original opened tabs, since the session data would never be modified.

Instead, I changed the actual session.js so that if `Zotero.Prefs.get('restoreState')` is `false`, only `pane` window with `library` tab is saved. I think it's better because it keeps the session data in sync with what happens with the actual interface and would be equally applicable to reader, tabs and notes.

defaults/preferences/zotero.js
abaevbog: Makes sense. I did that, and about reopening reader windows ...
I noticed that the current codebase has everything in place for reopening reader windows just like tabs. The data is saved in session.js, and on` init()` in reader.js, we check if there are saved reader windows and try to reopen them. So it seems like the reader windows _should_ be reopened, but they are not.

The issue is that `getSecondViewState` in reader.js is throwing an exception "this._iframeWindow.wrappedJSObject.getSecondViewState is not a function", which means that `Zotero.Reader.getWindowStates()` was throwing an exception which meant that reader updates were never saved by  `Zotero.Session.save()`. I will try to see if I can get to the root of the error but to begin, I wrapped `getSecondViewState` in `try-catch`, and it makes it so that reader windows are saved and reopened just like tabs.

abaevbog: The notes windows don't seem to be saved in in Zotero.Session yet. Is that something I should try to add to this PR?

dstillman: > I noticed that the current codebase has everything in place for reopening reader windows just like tabs.

Oh, yeah, that's actually just a regression — it works in Z6. I opened https://github.com/zotero/zotero/issues/3376 for you.

> The notes windows don't seem to be saved in in Zotero.Session yet. Is that something I should try to add to this PR?

Sure! Then we could actually make a visible pref for restoring state.

chrome/content/zotero/zoteroPane.js
```diff
@@ -538,6 +538,12 @@ var ZoteroPane = new function()
 			if (state) {
 				Zotero_Tabs.restoreState(state.tabs);
 			}
+			setTimeout(function () {
```
abaevbog, line 541: Without this timeout, there is an error either that collectionTree is not loaded yet or that the item for that note is not loaded. Is there a better way? Some kind of hook that fires when everything is 100% loaded and ready?

defaults/preferences/zotero.js
abaevbog: > Then we could actually make a visible pref for restoring state.

So, now the reader windows reopening should work the same as in Z6. Plus, I have the same logic working for notes opened in windows. All of that happens based on the `restoreState` preference. 

By visible pref, you mean ones in Zotero > Settings (on Mac)? Where should we place it? Maybe under General > Miscellaneous?

abaevbog:
I made it a visible pref here in the miscellaneous section of settings
Screenshot 2023-09-05 at 10 30 03 PM

dstillman:
Can we reopen note windows too?

https://github.com/zotero/zotero/issues/3376#issuecomment-2424614068

Not really related, except that if we're adding a pref to reopen windows it's weird if that doesn't apply to note windows too.

abaevbog:
It looks like we already have a setup to save opened notes windows in `session.js` in one of the commits here: https://github.com/zotero/zotero/pull/3358/commits/f719eb8dd923194d34b3e5a4e381798da4f763ca. But this is quite a bit behind the current `main` and things have changed significantly, so I'll rebase and go over these changes to make sure everything still works.

abaevbog:
Rebased with some minor refactoring. The main additions are a `getSecondViewState` specific to `ReaderWindow` (instead of earlier `this._lastSecondaryState` variable and logic), which seemed a bit cleaner to me, as well as a slightly better approach to waiting for when the note windows can be opened in ZoteroPane (earlier, I had an arbitrary delay, but it's probably better to `await` for `Zotero.Items.getAsync` first with a small of delays until collectionTree is loaded).

PR #3359, opened 2023-08-22
Fix progress queue segfault

### LINKED ISSUE #2688: fx-compat: Occasional segfault on Find Available PDF
Relevant crash log:

```
Process:               zotero [20850]
Path:                  /Users/USER/*/Zotero.app/Contents/MacOS/zotero
Identifier:            org.zotero.zotero-source
Version:               6.0.8 (6.0.8)
Code Type:             ARM-64 (Native)
Parent Process:        Exited process [20573]
User ID:               501

Date/Time:             2022-07-06 11:02:11.8505 -0400
OS Version:            macOS 12.2.1 (21D62)
Report Version:        12
Anonymous UUID:        D96101AC-0DE9-FD07-3709-523303B3DF52

Sleep/Wake UUID:       64DDE953-75B9-4A38-928D-B95A3A38110D

Time Awake Since Boot: 1800000 seconds
Time Since Wake:       3852 seconds

System Integrity Protection: enabled

Crashed Thread:        0  MainThread  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000168
Exception Codes:       0x0000000000000001, 0x0000000000000168
Exception Note:        EXC_CORPSE_NOTIFY

Termination Reason:    Namespace SIGNAL, Code 11 Segmentation fault: 11
Terminating Process:   exc handler [20850]

VM Region Info: 0x168 is not in any region.  Bytes before following region: 4309286552
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      UNUSED SPACE AT START
--->  
      __TEXT                      100da8000-100dac000    [   16K] r-x/r-x SM=COW  .../MacOS/zotero

Thread 0 Crashed:: MainThread Dispatch queue: com.apple.main-thread
0   XUL                           	       0x11222e784 0x111750000 + 11396996
1   XUL                           	       0x115de6940 0x111750000 + 74017088
2   XUL                           	       0x115de6940 0x111750000 + 74017088
3   XUL                           	       0x115de69bc 0x111750000 + 74017212
4   XUL                           	       0x111ec8550 0x111750000 + 7832912
5   XUL                           	       0x1149de910 0x111750000 + 53012752
6   XUL                           	       0x1149dea18 0x111750000 + 53013016
7   XUL                           	       0x114be13d0 0x111750000 + 55120848
8   XUL                           	       0x1121122a4 0x111750000 + 10232484
9   XUL                           	       0x1121b5550 0x111750000 + 10900816
10  XUL                           	       0x112372c48 0x111750000 + 12725320
11  XUL                           	       0x1123c0c60 0x111750000 + 13044832
12  XUL                           	       0x112ec666c 0x111750000 + 24602220
13  XUL                           	       0x112f8b110 0x111750000 + 25407760
14  XUL                           	       0x112f8a2dc 0x111750000 + 25404124
15  XUL                           	       0x112ebbcb8 0x111750000 + 24558776
16  XUL                           	       0x112ec45e4 0x111750000 + 24593892
17  XUL                           	       0x113254184 0x111750000 + 28328324
18  ???                           	    0x18d61d058fd0 ???
19  ???                           	       0x126e3b138 ???
20  ???                           	    0x18d61d0587a8 ???
```

Could reproduce repeatedly ~5 times, but not anymore. It seemed to happen right as the PDF should have begun downloading. Unfortunately no debug symbols.

AbeJellinek:
Crazy stuff going on here. When `require()` is invoked from a non-window context, `require.js` creates a fake window global. The fake window is window-y enough to make library code work, but nodes created under it end up with a null `ownerGlobal`, because it isn't _actually_ a window. And when a node's `ownerGlobal` is null, appending children or setting its `innerHTML` causes a segfault in the XUL process. Because `ProgressQueueTable`'s `rowToTreeItem` sometimes gets called by an event that's triggered in a non-window context, it ends up creating `ownerGlobal`-less nodes and triggering the segfault.

I tried creating the nodes under another window (like `Zotero.getMainWindow()`). This does not change anything. What matters is that the `ProgressQueueTable` code is being called by something with a fake-window global. Even if it has access to a real window, the nodes it creates are cursed to forever cause segfaults.

The crash can be reproduced easily on `main` by running Retrieve Metadata on a standalone attachment that doesn't exist (causes an immediate failure), closing the progress queue window, and then running it again. Sometimes one extra run is required before you get a crash.

This is a temporary fix. The long-term fix is probably to ditch `require.js` and use ES modules, but that might need to wait until fx115.

Fixes #2688.

### DIFF
--- a/chrome/content/zotero/components/progressQueueTable.jsx
+++ b/chrome/content/zotero/components/progressQueueTable.jsx
@@ -68,6 +68,14 @@ const ProgressQueueTable = ({ onActivate = noop, progressQueue }) => {
 		for (let column of columns) {
 			if (column.dataKey === 'success') {
 				let span = document.createElement('span');
+				if (!span.ownerGlobal) {
+					// If this script was imported from a non-window context, we'll have a global object that looks like
+					// a Window and document.createElement() will succeed, but the returned Element object won't have
+					// an ownerGlobal. Trying to append a child or set its innerHTML will segfault Zotero. For now,
+					// let's just abort if we get an invalid Element.
+					// TODO: Remove once we're using ES modules
+					return div;
+				}
 				span.className = `cell icon ${column.className}`;
 				span.appendChild(getImageByStatus(row.status));
 				div.appendChild(span);


dstillman:
Wow.

PR #3360, opened 2023-08-22
Add EPUB to Attachment File Type in Advanced Search

### LINKED ISSUE #3297: Add EPUB to Attachment File Type in Advanced Search
https://forums.zotero.org/discussion/106735/zotero-7-avanced-search-for-epub-attachment-file-type

AbeJellinek:
Fixes #3297

### DIFF
--- a/chrome/content/zotero/xpcom/schema.js
+++ b/chrome/content/zotero/xpcom/schema.js
@@ -3450,7 +3450,17 @@ Zotero.Schema = new function(){
 				}
 			}
 			
-			// TEMP: When adding 122, check whether IA.authorName fix in items.js::_loadAnnotations()
+			else if (i == 122) {
+				try {
+					await Zotero.DB.queryAsync("INSERT INTO fileTypes VALUES(8, 'ebook')");
+					await Zotero.DB.queryAsync("INSERT INTO fileTypeMIMETypes VALUES(8, 'application/epub+zip')");
+					// Incorrect, for compatibility
+					await Zotero.DB.queryAsync("INSERT INTO fileTypeMIMETypes VALUES(8, 'application/epub')");
+				}
+				catch (e) {}
+			}
+			
+			// TEMP: When adding 123, check whether IA.authorName fix in items.js::_loadAnnotations()
 			// can be updated due to update steps being indempodent
 			
 			// If breaking compatibility or doing anything dangerous, clear minorUpdateFrom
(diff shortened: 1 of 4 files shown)


resource/schema/system.sql
```diff
@@ -1,4 +1,4 @@
--- 32
+-- 33
```
dstillman, line 1: We actually don't do this anymore, for reasons that I don't recall and don't want to risk relearning. The last system.sql version update was in 2015, and all changes since then have been performed by [schema.js steps with a userdata.sql bump](https://github.com/zotero/zotero/commit/3b402c564cb3bb4a41a9486f8522c8793dfdd2fb).

chrome/content/zotero/xpcom/schema.js
```diff
@@ -3450,7 +3450,17 @@ Zotero.Schema = new function(){
...
 			
-			// TEMP: When adding 122, check whether IA.authorName fix in items.js::_loadAnnotations()
+			else if (i == 122) {
+				try {
+					await Zotero.DB.queryAsync("INSERT INTO fileTypes VALUES(8, 'ebook')");
+					await Zotero.DB.queryAsync("INSERT INTO fileTypeMIMETypes VALUES(8, 'application/epub+zip')");
+					// Incorrect, for compatibility
+					await Zotero.DB.queryAsync("INSERT INTO fileTypeMIMETypes VALUES(8, 'application/epub')");
+				}
+				catch (e) {}
```
dstillman, line 3460: Oh, sorry, we still make the change to system.sql, and that's used on DB initialization. We just don't bump the system.sql version to trigger the old wipe-and-replace mechanism.

And this doesn't need to be in a try/catch — since we're setting userdata version 122 in this commit, where we'll also add the lines to system.sql, this won't run unless these rows don't exist.

dstillman: > And this doesn't need to be in a try/catch — since we're setting userdata version 122 in this commit, where we'll also add the lines to system.sql, this won't run unless these rows don't exist.

Oh, but I see why you had the try/catch — because an upgrade that does involve a system.sql bump still creates the system tables with the new system.sql, so the new rows exist. So I introduced an upgrade failure by removing this. Sorry about that.

But what we've done for past steps, and what we can do here, is just use REPLACE instead of INSERT. Fixed in 075457be4.

PR #3363, opened 2023-08-23
Make preferences less janky, preload panes on hover, allow panes to delay visibility until promise resolves

AbeJellinek:
- Prevent flashes of unlocalized labels and controls without values set
- Preload panes on hover so there's usually no delay when switching
- Allow panes to delay visibility until a promise/promises resolve, and use that in Cite and Export to prevent flashes of empty tables
- Eliminate some race conditions

I thought there was an issue for the flashes of uninitialized content but can't find it now.

### DIFF
--- a/chrome/content/zotero/preferences/preferences.js
+++ b/chrome/content/zotero/preferences/preferences.js
@@ -481,7 +481,13 @@ ${str}
 		}
 
 		for (let child of container.children) {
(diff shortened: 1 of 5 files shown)


dstillman:
The test failure here is real (as in, it definitely happens — whether or not it exposes a real issue I don't know).

PR #3387, opened 2023-08-31
Add Developer Toolbox command that uses bundled devtools

### LINKED ISSUE #3384: Open devtools internally
I think we're bundling the devtools client code now, so we should try to just open them in a window instead of requiring an external Firefox build of the same version.

AbeJellinek:
Tools -> Developer -> Browser Toolbox.

Open questions/issues:

- Now we have three very similar commands:

  1. Error Console, which shows error output and allows evaluation
  2. Browser Toolbox, which shows error output, other devtools, and allows evaluation
  3. Run JavaScript, which allows evaluation

  Maybe two would be more reasonable? Error Console is reasonably end-user-friendly, it puts evaluation and error output in the same place, it has autocomplete, and it resolves promises automatically. We could remove Run JavaScript in favor of that.
- ~Zotero isn't a browser, but "Tools -> Developer -> Developer Tools" is silly and I can't think of a better name right now.~
- No API method that I can find to activate an existing Browser Toolbox instance.

Other changes:

- The Browser Toolbox client process [is launched with a `-chrome` argument](https://searchfox.org/mozilla-esr102/source/devtools/client/framework/browser-toolbox/Launcher.jsm#295-296), which we no longer supported because of our command-line handler override. But it doesn't seem like we actually need the override! I have no idea what was going on before, but I reverted 722287b81a5db3af6c8de5c22ac72ae633e9ef18, changed `m-zotero` to `a-zotero` (gives our handler a higher priority), and it just worked.

    `-chrome` works, tests work, opening a file doesn't duplicate the main window on macOS, opening a file when Zotero isn't running works... as far as I can tell, everything just works without an override. Now we get `-chrome` again, which allows the bundled Browser Toolbox to launch on startup.

    I also tried implementing our own command-line option so we could keep our CLH override, but for some reason, it would always create two copies of the Browser Toolbox window and would occasionally launch a regular Zotero instance (on a new profile) because of the Ctrl-C safe mode issue. I get the sense, based on the odd bugs that it causes, that overriding the default CLH is a deeply terrible thing to do. Since it doesn't actually seem necessary, we just shouldn't do it.

Fixes #3384

### DIFF
--- a/chrome/content/zotero/standalone/standalone.js
+++ b/chrome/content/zotero/standalone/standalone.js
@@ -961,6 +961,15 @@ function toJavaScriptConsole() {
 	BrowserConsoleManager.openBrowserConsoleOrFocus();
 }
 
+function launchBrowserToolbox() {
+	const { BrowserToolboxLauncher } = ChromeUtils.import("resource://devtools/client/framework/browser-toolbox/Launcher.jsm");
+	// Don't launch if already open
+	// (Can we focus the existing toolbox process?)
+	if (!BrowserToolboxLauncher.getBrowserToolboxSessionState()) {
+		BrowserToolboxLauncher.init();
+	}
+}
+
 function openRunJSWindow() {
 	openWindowByType(
 		'chrome://zotero/content/runJS.html',
--- a/chrome/content/zotero/standalone/standalone.js
+++ b/chrome/content/zotero/standalone/standalone.js
@@ -962,6 +962,26 @@ function toJavaScriptConsole() {
 }
 
 function launchDeveloperToolbox() {
+	let prefs = ['devtools.chrome.enabled', 'devtools.debugger.remote-enabled'];
+	let enabled = prefs.every(pref => !!Zotero.Prefs.get(pref, true));
+	if (!enabled) {
+		let ps = Services.prompt;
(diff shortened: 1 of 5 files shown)


dstillman:
Nice!

> Zotero isn't a browser, but "Tools -> Developer -> Developer Tools" is silly and I can't think of a better name right now.

Developer Toolbox?

> Now we have three very similar commands:
>
>    i. Error Console, which shows error output and allows evaluation
>   ii. Browser Toolbox, which shows error output, other devtools, and allows evaluation
>  iii. Run JavaScript, which allows evaluation

I think that's fine. Run JavaScript is much nicer for running and tweaking code snippets than the console (where you have to keep pasting or pulling up the old entry), and it's the only one we control, so, e.g., we can add a pane that shows debug output, it's where we could add stuff related to saved code snippets (even assigning keyboard shortcuts), etc.

dstillman:
Nothing happens when I select the menu option, except this is logged in the error console:

> Could not start Browser Toolbox, you need to enable it.

crokobit:
Will this feature be developed? I am a new developer at Zotero, and I'm finding it challenging to begin development due to the limited documentation and debugging tools available. 😢

PR #3388, opened 2023-08-31
Fix segfault when hyphenating text in main process

AbeJellinek:
This will let us enable hyphenation in EPUBs again (and not disable it in snapshots).

https://forums.zotero.org/discussion/107328/report-id-1522731840-zotero-7-beta


dstillman:
Just as a reminder for ourselves, when we remove these, we may have to take extra steps to actually remove them from people's installation directories.

PR #3397, opened 2023-09-05
[TAS-37] Bugfix: mktemp error when building on macOS and pako package lost problem.

RayKr:
fix: 
```
mktemp: too few X's in template ‘tmp’
```
fix: 
```
symlink has no referent: "/Users/jzl/code/zotero/build/resource/pako.js"
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1356) [sender=3.2.7]

Traceback (most recent call last):
  File "/Users/jzl/code/zotero/app/scripts/prepare_build", line 75, in main
    subprocess.check_call([
  File "/Users/jzl/miniforge3/lib/python3.9/subprocess.py", line 373, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['rsync', '-aL', '--exclude', '.*', '--exclude', '#*', '--exclude', 'package.json', '--exclude', 'package-lock.json', './', '/Users/jzl/code/zotero/app/tmp/zotero/']' returned non-zero exit status 23.
```

### DIFF
--- a/app/config.sh
+++ b/app/config.sh
@@ -54,7 +54,7 @@ BUILD_PLATFORMS=""
 NUM_INCREMENTALS=6
 
 if [ "`uname`" = "Darwin" ]; then
-	alias mktemp='mktemp -t tmp'
+	alias mktemp='mktemp -t tmp.XXXXXXXXXX'
 	shopt -s expand_aliases
 fi
 
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "jspath": "^0.4.0",
     "mocha": "^3.5.3",
     "multimatch": "^2.1.0",
+    "pako": "^2.1.0",
     "react-virtualized": "^9.22.4",
     "sass": "^1.28.0",
     "sinon": "^7.3.2",


app/config.sh
```diff
@@ -54,7 +54,7 @@ BUILD_PLATFORMS=""
 NUM_INCREMENTALS=6
 
 if [ "`uname`" = "Darwin" ]; then
-	alias mktemp='mktemp -t tmp'
+	alias mktemp='mktemp -t tmp.XXXXXXXXXX'
```
dstillman, line 57: Where exactly are you getting an error? This code only runs on macOS, and at least on current versions of macOS `-t` is for a prefix, so `-t tmp` is correct. It's also seemingly unnecessary, though — without `-t tmp` `-d` defaults to `-t tmp`, so this line could just be removed.

dstillman:
What is `TAS-37`?

RayKr:
never mind. that is a my tip of task ：）

app/config.sh
RayKr: It really can't work only with `-t tmp` on my macOS (Apple Silicon M1, MacBook Pro, Ventura 13.5.1)
image
So this line could be removed.

dstillman: We can remove the line, but just to be clear, you've installed the GNU mktemp. This doesn't happen with the bundled BSD mktemp.

dstillman: I've removed the line in https://github.com/zotero/zotero/commit/e1def9971c18999b0e0bb127fa3f2cb1ccd59d5f. You can undo this change.

RayKr: Okey, I'm clear. Thanks:)

dstillman:
Thanks!

Issue #3410, opened 2023-09-11
Zotero 7: Issues with collection highlighting


abaevbog:
Steps to reproduce on ubuntu: select an entry in itemTree, click Alt to highlight the collection it belongs to, click on Down arrow. The file menu will open up instead of moving focus to another record.

Original forum post: https://forums.zotero.org/discussion/107663/zotero-7-file-menu-focused-after-highlighting-collections

abaevbog:
To add to that, if you hold ALT when an item in itemTree is selected and then click onto a collection, the yellow mark around the collection will get stuck and not go away until the next time alt is held on an itemTree. This behavior is not ubuntu-specific. Same happens on a Mac.



<img width="1512" alt="Screenshot 2023-09-11 at 9 19 10 AM" src="https://github.com/zotero/zotero/assets/36271954/524f1cd4-29fa-4db3-be23-f1210aef277c">

abaevbog:
To add to this list - highlighting collections when the focus is on contextPane and not on itemTree causes the highlight to get stuck even after the button is released.

<img width="1512" alt="Screenshot 2023-09-27 at 3 12 10 PM" src="https://github.com/zotero/zotero/assets/36271954/9aaaedf7-8976-48b5-a206-004c11ece684">

PR #3412, opened 2023-09-11
fix to collections highlight bugs

abaevbog:
1. Trigger ZoteroPane.handleBlur on focusout instead of blur to remove highlight when itemTree entries loose focus even if ctrl/alt is still being held.
2. Prevent default event on Alt click on linux to avoid moving focus to menubar
3. Avoid re-triggering focus multiple times when ctrl is being held on windows to prevent highlight from becoming stuck and not going away.
4. Clear the highlight in `handleKeyUp` regardless of where the event comes from - this prevents highlight from being stuck when on windows focus is moved from itemTree before releasing `Ctrl`.

Fixes: #3410

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -74,7 +74,9 @@ var ZoteroPane = new function()
 		
 		// Set key down handler
 		document.addEventListener('keydown', ZoteroPane_Local.handleKeyDown, true);
-		document.addEventListener('blur', ZoteroPane.handleBlur);
+		// focusout, unlike blur, bubbles up to document level
+		// so handleBlur gets triggered when any field, not just the document, looses focus
+		document.addEventListener('focusout', ZoteroPane.handleBlur);
 		
 		// Init toolbar buttons for all progress queues
 		let progressQueueButtons = document.getElementById('zotero-pq-buttons');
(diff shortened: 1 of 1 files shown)


chrome/content/zotero/zoteroPane.js
```diff
@@ -1044,8 +1056,7 @@ var ZoteroPane = new function()
 	}
 	
 	function handleKeyUp(event) {
-		var from = event.originalTarget.id;
-		if (ZoteroPane.itemsView && from == ZoteroPane.itemsView.id) {
+		if (ZoteroPane.itemsView) {
```
dstillman, line 1059: Probably don't need this at all, right? This was only there because we used it in the comparison.

abaevbog: Oh yeah, that's right. I just force-pushed a commit without it

abaevbog:
@adomasven, the glitch this was dealing with was not that you couldn't hold Alt and navigate items with arrows. The issues was that if you were to just click Alt to highlight collections for an item, release it, and then press ArrowDown, instead of navigating the itemTree to the next item, you would see the File menu would pop up and focus would jump to it. But, yeah, I see this certainly messes with default shortcuts.

I'm trying to see now if there's any way to not block the Alt event while still not having the menu pop up. A simpler solution, as you pointed out in #3684, would be to just use Ctrl instead of Alt. Is there any reason to not just do that and have Ctrl highlight collections on Linux, as well as windows?

dstillman:
It's totally possible that we're using Alt on Linux solely because we were using Option on macOS, and that we changed Windows to Ctrl to avoid menu interactions without thinking about Linux. So we can probably switch to Ctrl on Linux if it avoids other problems. It would mean that everyone who has learned this secret key would have to learn another secret key, though, so we should only do it if it's actually a problem.

abaevbog:
So, after some experimentation, my understanding is that after Alt is pressed, _sometimes_ we will want the menus to be activated to underline access keys (the next button click will navigate the menus), and _some other times_ we want the focus to remain on itemTree (when one highlights collections, so the next button click should navigate the itemTree). I feel like we do have to pick one or the other for everything to be consistent and robust. 

I did try to see if it's possible to not change the collection highlighting shortcut. If itemTree is focused, there's no way to tell if the user is after access keys or after collection highlights, so we could make a guess what to do with the menus/focus based on how long the key was held? E.g. more than X milliseconds - force the focus to remain on itemTree, otherwise - let it go to the menu. That is 100% janky and probably not that useful, so I'll get started on setting up Ctrl to be collection highlight trigger to free up Alt for the menu - but correct me if I'm wrong

abaevbog:
Added #3689 to try to trigger focus on linux on Control instead of Alt

Issue #3215, opened 2023-07-18
Transparent background in Run JavaScript window on Linux


dstillman:
https://forums.zotero.org/discussion/106280/zotero-7-beta-run-javascript-window-flickering-ubuntu-22-08

And for me it starts drawing all over the screen if I move the window

abaevbog:
I reproduced this too on linux both with x11 and with Wayland. The flickering only happens on Wayland - on x11, the background is just transparent or blank. 

Is hardcoding background color a viable option? The look of the the desktop app does not change for me depending on dark/light theme on my Mac, so I don't think we expect background color to be dynamic? 

If so, adding this line

`
	if (Zotero.isLinux) {
		document.body.style.backgroundColor = 'white';
	}
`

to `runJS.js` to `window.addEventListener("load", function (e)` solved this for me - it successfully set the background color to be non-transparent and fixed flickering and drawing all of the screen on drag that @dstillman pointed out.

dstillman:
Should be fixed

PR #3414, opened 2023-09-12
explicitly set white backgorund for runJS on linux

abaevbog:
It prevents background from being oddly black/transparent and stops flickering when using wayland.

Fixes: #3215


dstillman:
Is this still necessary post-redesign?

abaevbog:
I don't think so. I checked on Ubuntu 20 Xorg and Wayland, and in both cases everything looks right.

Issue #21, opened 2012-01-01
Document collections


simonster:
Make items cited in open documents appear on the left side of the Zotero pane.

aurimasv:
I'd like to put some effort into this. Could we come up with a consensus on what the UX for this should be? @fbennett's implementation is a step in the right direction, but I think it needs to be a lot more streamlined. Here's my take on it:

* Because documents can contain items from multiple libraries (multiple personal as well as group libraries), they should be displayed in an independent category (let's call it "Open Documents"), which would similar to the "Group Libraries" category. In this category, you could have a number of currently-open documents, which would appear as libraries. No collections are involved (at least at this point). Clicking on each library (or open document) would display the list of items in the document.
* These libraries are created (and removed) dynamically. They are created when a new document is opened and some Zotero operation is performed from within Word. Zotero obviously has no way of knowing when a new document is opened, hence the need to perform some Zotero-related action (e.g. Refresh, but most actions scan the whole document anyway). It may be possible to automatically trigger this from the Zotero Word Processor add-in, but given how slow these operations can be on larger documents, I don't think we would want to bog down opening documents.
* The items are populated in the following order:
  * by looking up the reference URI: if the reference is already in the user's library, the item is just a link to that item
  * by extracting embedded metadata
  * by looking up metadata on zotero.org (if the item is from a public library)
  * by randomly generating metadata. jk, I'm not really sure what to do if the metadata is not embedded in the document and only a URI is available to some private library. Maybe not display that item at all? Maybe display some sort of placeholder based on rendered citation?
* The items are updated (from document to Zotero) only when operations are performed on the document that involve fetching all items from the document. I think this pretty much means that it will stay in-sync with the document unless a reference is deleted. In the future, we may want to think how to streamline things for users that attempt to edit citations within the document.
* Since items in the "Open Document" library are links to actual documents in the user library (where available), editing those items in the "Open Document" library would change metadata of the linked item. Items that are not part of user-accessible libraries, would be read-only (perhaps with some indicator showing that the item belongs to some other library). That's sort of a stopgap measure for now, because I think that we would just want users to be able to edit these references as well, but basically save them in their library (where?) and re-insert the reference.

That's all I've got for now. I'm sure I'm missing something.

One related feature that we should consider is converting all references in a document to user or group library references. There might be an issue about that somewhere already.

fbennett:
This would be great! I'm sure it can be done more elegantly.

[...]

PR #3468, opened 2023-10-31
Citation explorer

adomasven:
image

(EDIT 2026: Updated the image above so that the initial comment here represents the current working state)

Adds a Citation Explorer dialog to the Word integration as per https://github.com/zotero/zotero/issues/21#issuecomment-1473852674 with the following functionality:

- Find all citations that cite a specific item
- List items in a selected citation
- Highlight a selected citation in the document
- Open up the citation dialog for the selected citation
- Show item in Zotero library (where you can edit item metadat)
- Highlight unlinked items and allow relinking them
- Show a diff for items with modified text
- A tab to see items only cited in the bibliography

Notes:
- A user might use this dialog to remove citations from their document, which depending on the word processor might break citation highlighting. We detect those cases and notify the user that further highlighting of citations is disabled until the dialog is reopened.
- A button has been added to support this new functionality for all word processors. Things are getting kind of busy there with the recent addition of Add Note and now this. We could consider rolling Unlink Citations into Document Preferences, or do something else, but we are quite limited in our options.
- Show in Document works better if we are able to auto-focus the document processor window. In my testing this works worst/not-at-all on Windows, and I actually added a function to activate Word (which, apparently, wasn't previously implemented). Users will have to manually switch to Word on Windows.
- Winword template will need to be signed at the end of this by Dan.


adomasven:
Merging this would allow [forum users](https://forums.zotero.org/discussion/111879/quellen-aus-bestehendem-dokument-ins-zotero-uebertragen) to rebuild their library from existing documents.

image

dstillman:
https://forums.zotero.org/discussion/comment/471834/#Comment_471834

abaevbog:
I am noticing that with `regularOnly: true`, arrowRight/left can still display/hide child rows with notes and attachments when multiple rows regular items are selected:

https://github.com/user-attachments/assets/27d24449-0bb1-4249-aa9d-6243786c5295

adomasven:
> I am noticing that with `regularOnly: true`, arrowRight/left can still display/hide child rows with notes and attachments when multiple rows regular items are selected

It may be so, but I'm not able to reproduce it in the citation dialog. The `reguralOnly` prop wasn't added by this commit, merely moved from being an attribute of the ItemTree object to a prop. Maybe some adjustments in the citation dialog tree are causing it?

abaevbog:
I can reproduce it on this branch. But it only occurs in some cases, depending on which items are selected and I think which child items are present. So it occurs in some google docs with citations I use for testing but not in others. This seems to be the most reliable way I found to reproduce it:

1. Start with a fresh document
2. Add two items by identifier from the toolbar: 10.1086/674946 and 978-1-84631-829-0
3. Make a citation with just these two items

https://github.com/user-attachments/assets/20b4bd55-5e79-404f-a835-d30f10233452

Does this happen for you?

abaevbog:
> The reguralOnly prop wasn't added by this commit, merely moved from being an attribute of the ItemTree object to a prop.

Oh, I see what you mean now. `regularOnly` has been around for a while since https://github.com/zotero/zotero/commit/cbbff600a60c9e7a7407d6f2e4053309bf28b872, it was just not used, which is why we didn't run into it  before. If I add `this.regularOnly = true` to the constructor of the `itemTree.js`, I get the same behavior in the main itemTree of ZoteroPane. So I'll make a separate issue for it, as it isn't caused by anything citations-related.

adomasven:
> If I add `this.regularOnly = true` to the constructor of the `itemTree.js`, I get the same behavior in the main itemTree of ZoteroPane. So I'll make a separate issue for it, as it isn't caused by anything citations-related.

Right, it's actually even older than that, and transferred from the old XUL item tree.

dstillman:
Merged hidden in https://github.com/zotero/zotero/commit/a3654652b21071955ae01ba8fcdc38765e1740a4 as part of #4872, but leaving this open for review.

adam3smith:
Is this accessible in the current beta (with a hidden pref or a JS call or so)? I'm very curious to try this.

adomasven:
This was broken, so updated to make it work again and add a hidden pref `integration.googleDocsCitationExplorerEnabled` to enable it in Google Docs. Enabling in other doc editors is a matter of updating UI. The dotm files and libreoffice changes are also available.

This is what it looks like now:
image

I assume @dstillman won't let this go through before @yexingsha gives her input and standardizes the look. Please have a go at it when possible.

yexingsha:
citation-explorer

[Figma link](https://www.figma.com/design/y8aZzEDwsxsgVh7Zui63to/Zotero?node-id=14125-128652&t=6REyd79Hf2j0eAWA-1) (I should be able to give you access to this file)

Aside from the obvious changes:
- Making the central divider draggable is probably overkill, but we can split the two areas 40/60 to utilize the space better.
- I think we need placeholder text in search bars; if these are wrong, feel free to correct them.
- Changed the icon of the "Is Linked" column to `link`, since the original icon means attachment.
- I think it makes more sense to put the "Is Linked" column in the same place, rather than mirrored.
- In item tree, "Year" is probably more useful than "Date".
- Changed deleted text to `accent-red`.

Other issues I found while testing:
- Colored tags are missing from titles, and the emoji tags are displayed too tight.
- When I select a citation X, then select an item Y, I don't think the previous selection of citation X and its item highlights should persist. It's busy and confusing, and also makes it difficult to tell whether item Y is cited in citation X, since X is now always colored gray no matter which item you select. The same is true in the opposite direction.

adomasven:
macOS:
image

Windows:
image

Linux:
image

abaevbog:
@adomasven, I was trying this PR and I wanted to share a few moments I noticed:

1. When a row in the left ("Citation") table is selected and selection changes in the right itemTree, I am seeing a noticeable flicker of the row that the selected citation has. In the recording below, the item "A Candid Assessment ..." does the flickering:


https://github.com/user-attachments/assets/8e3e1a27-1e6e-4e18-bdd6-190528c5eeb2

2. If i click on the "Show in Document", I always just get a warning saying "Citations in your document have been modified since Citation Explorer has been opened...", and when I close it, the "Show in Document" and "Edit Citation" buttons are both disabled, and nothing else happens. Is that expected?
3. Would it make sense to disable "Edit Citation" when multiple citations are selected? It currently opens the first selected citation but, since many are selected, it implies that somehow all selected citations will be opened for edit, I think. We could also consider to move that button as a column of the Citations table.
4. On the other hand, would we want to enable "Show in Zotero" button if multiple items are selected in the itemTree? I understand that the items are very likely to not be anywhere near each other. But it seems like people are often interested in fetching all cited items to group or export them. So this would allow one to get all cited items selected in their Library, after which one can move them into a collection for further exporting/processing/etc.
5. I think the tab order in the dialog at the moment is not quite right. It spans the entire document in a way that doesn't really correspond to the layout. I think that as you press Tab, the focus should move in the following order, similar to ZoteroPane: `#citation-search` -> `#citation-list` -> `#button-show-in-document` -> `#button-edit-citation` -> `cited/uncited tabbox` -> `#item-search` -> `itemTree` -> `#button-show-in-zotero` -> `#button-relink-item` -> `#button-addTo-library` -> `OK`
6. The View menu with options to sort or rearrange rows of the itemTree (for a11y) sometimes does not appear for me when the citation explorer opens. I think I was running into a similar issue with citation dialog and [fixed it by adding ](https://github.com/zotero/zotero/blob/e27b58176d45b2cba4f44d08248ec70d1ffc55e6/chrome/content/zotero/integration/citationDialog.xhtml#L49-L50) the `<xul:item-tree-menu-bar/>` node into the `.xhtml` file instead of waiting for itemTree to create it.

yexingsha:
Things I'm noticing from the screenshots:

- The icon in the "Add to ..." button should be colored `accent-blue`.
- The gap between "Relink Item" and "Add to ..." should be 8px.
- On macOS, the outer paddings of the two item trees look about 2px smaller than their inner paddings.
- On windows, the divider between sections look weirdly dark.

And there are still outstanding issues from my last comment:

- In item tree, "Year" is probably more useful than "Date".
- Changed deleted text to accent-red.
- Colored tags are missing from titles, and the emoji tags are displayed too tight.
- When I select a citation X, then select an item Y, I don't think the previous selection of citation X and its item highlights should persist. It's busy and confusing, and also makes it difficult to tell whether item Y is cited in citation X, since X is now always colored gray no matter which item you select. The same is true in the opposite direction.

adomasven:
MacOS:
image

adomasven:
Windows (the View menu is hidden by default, but alt-print screen made it visible):

Untitled

Thanks all for the comments!

> 1. When a row in the left ("Citation") table is selected and selection changes in the right itemTree, I am seeing a noticeable flicker of the row that the selected citation has. In the recording below, the item "A Candid Assessment ..." does the flickering:
> * I think the tab order in the dialog at the moment is not quite right. It spans the entire document in a way that doesn't really correspond to the layout. I think that as you press Tab, the focus should move in the following order, similar to ZoteroPane: `#citation-search` -> `#citation-list` -> `#button-show-in-document` -> `#button-edit-citation` -> `cited/uncited tabbox` -> `#item-search` -> `itemTree` -> `#button-show-in-zotero` -> `#button-relink-item` -> `#button-addTo-library` -> `OK`
> * The View menu with options to sort or rearrange rows of the itemTree (for a11y) sometimes does not appear for me when the citation explorer opens. I think I was running into a similar issue with citation dialog and [fixed it by adding ](https://github.com/zotero/zotero/blob/e27b58176d45b2cba4f44d08248ec70d1ffc55e6/chrome/content/zotero/integration/citationDialog.xhtml#L49-L50) the `<xul:item-tree-menu-bar/>` node into the `.xhtml` file instead of waiting for itemTree to create it.

> * The icon in the "Add to ..." button should be colored `accent-blue`.
> * The gap between "Relink Item" and "Add to ..." should be 8px.
> * On windows, the divider between sections look weirdly dark.
> * In item tree, "Year" is probably more useful than "Date".
> * Changed deleted text to accent-red.
> * Colored tags are missing from titles, and the emoji tags are displayed too tight.
> * When I select a citation X, then select an item Y, I don't think the previous selection of citation X and its item highlights should persist. It's busy and confusing, and also makes it difficult to tell whether item Y is cited in citation X, since X is now always colored gray no matter which item you select. The same is true in the opposite direction.

Fixed. Some of these were supposed to be addressed, but the code didn't quite work or regressed with other changes.

@abaevbog:

> 2. If i click on the "Show in Document", I always just get a warning saying "Citations in your document have been modified since Citation Explorer has been opened...", and when I close it, the "Show in Document" and "Edit Citation" buttons are both disabled, and nothing else happens. Is that expected?

This is due to a bug in Google Docs editor that will be fixed. It will work if you first place a cursor in a citation when opening the Citation Explorer, or after closing and reopening after it didn't work at first. I've explained why this exists in the initial PR comment.

> 3. Would it make sense to disable "Edit Citation" when multiple citations are selected? It currently opens the first selected citation but, since many are selected, it implies that somehow all selected citations will be opened for edit, I think. We could also consider to move that button as a column of the Citations table.
> 4. On the other hand, would we want to enable "Show in Zotero" button if multiple items are selected in the itemTree? I understand that the items are very likely to not be anywhere near each other. But it seems like people are often interested in fetching all cited items to group or export them. So this would allow one to get all cited items selected in their Library, after which one can move them into a collection for further exporting/processing/etc.

This is more of a design consideration. Overall, probably the most reasonable thing to do is to disable multi-select in these trees. There are two reasons for multi-select: being able to highlight multiple items in the other tree by selecting multiple in one (in case you're troubleshooting where which citation is included, or what have you), the other is multiple Show in Zotero, which you correctly observe, currently doesn't work.

Overall, I'm not sure if it's good UX to have buttons that appear or disappear based on context, so in that sense maybe buttons in the tree would be better. The other option is having buttons that are disabled and enabled based on context. I'm not a very big fan of buttons in the tree for this, since there are all quite discrete and contextual actions that are better explained with text than an icon.


@yexingsha:

> * On macOS, the outer paddings of the two item trees look about 2px smaller than their inner paddings.

Not sure what you mean here.

adomasven:
@yexingsha Do you have design notes for other OSes?

Perhaps using tabs for the switch between Cited and Uncited items is not ideal? Another thing is that I assume that only a minority of documents actually have uncited items (added via Edit Bibliography dialog), and for most users that section will be empty. As such, it might make sense to not show it by default.

yexingsha:
The fixes look good! The dialog on widows look much better now, though the OK button should have a smaller top margin (16px).

> Overall, probably the most reasonable thing to do is to disable multi-select in these trees.

I agree. As to the two potential reasons you listed, I feel like very few people would actually do that.

> On macOS, the outer paddings of the two item trees look about 2px smaller than their inner paddings.

The paddings in question:

macOS

> Perhaps using tabs for the switch between Cited and Uncited items is not ideal?

Right. Since uncited items are always added from Edit Bibliography, let's use the same label system from there:

citation-explorer-citation

citation-explorer-item

- The `Uncited` label follow the same design as in https://github.com/zotero/zotero/pull/5210. It's put at the start of item title, before tags. When the item is selected, use `#FFFFFF33` as the label's background color, and `accent-white` as its text color.
- Also made some updates to the design up top. The right item tree is titled "Items in Bibliography" to make things clearer. Both titles use `font-size: 15px; font-weight: 600;`. Gap between title and search bar is reduced to 8px.

abaevbog:
> This is more of a design consideration. Overall, probably the most reasonable thing to do is to disable multi-select in these trees. There are two reasons for multi-select: being able to highlight multiple items in the other tree by selecting multiple in one (in case you're troubleshooting where which citation is included, or what have you), the other is multiple Show in Zotero, which you correctly observe, currently doesn't work.

With these two use cases, multi-select feels quite important here. One should be able to find out which items are in a set of citations, right? And being able to find a selection of items in Zotero library is a necessary step to be able to, say, place all of your cited items into a collection before exporting them somewhere (which I believe is a frequently asked about use case). 

> Overall, I'm not sure if it's good UX to have buttons that appear or disappear based on context, so in that sense maybe buttons in the tree would be better. The other option is having buttons that are disabled and enabled based on context.

I agree that hiding/showing buttons depending on context is a bit jumpy. Disabling/enabling buttons instead could feel smoother.

Placing buttons into the columns is harder here than in other instances. For example, in citation dialog, we have `+` buttons, and it's quite self-explanatory what it would do. I think some buttons here are better candidates than others to be placed into the tables - for instance, "Edit citation" would probably make sense with a standard "Edit" icon. But others, like "Show in Document" or "Relink Item" in itemTree - not so much.

The benefit of buttons in the table is that you have only one click needed to complete the action vs two (selection + button click). But not sure if it's worth it here.

abaevbog:
> Right. Since uncited items are always added from Edit Bibliography, let's use the same label system from there:

I certainly like this look without tabs. And it's nice to draw a parallel with the Edit Bibliography dialog too. The only question I have is how would one find all uncited items? If one has many citations, it could be challenging to scroll through a long list. Is that a valid concern? If so, an alternative options could be to place the "Uncited" label into a separate column.- then one would be able to sort by it.

yexingsha:
> And being able to find a selection of items in Zotero library is a necessary step to be able to, say, place all of your cited items into a collection before exporting them somewhere (which I believe is a frequently asked about use case).

Oh, this is a good point. Then I guess we do need to allow multi-select and fix "Show in Zotero." I think disabling the "Edit Citation" button when multiple items are selected should be fine; maybe we can add a tooltip to the button in this state that says "You can only edit one citation at a time." Or something like that.

> The only question I have is how would one find all uncited items? If one has many citations, it could be challenging to scroll through a long list. Is that a valid concern?

I guess people might try to find uncited items if e.g. they added some entries to the bibliography, but then found out they need to either cite or delete them because of requirements of a journal or their professor. Since you can't actually delete a bibliography entry in the citation explorer, and I don't think it makes sense to enable it here (since we are looking at items, not really bibliography entries), this seems more like a concern for the Edit Bibliography dialog.

abaevbog:
I may have stumbled on some odd highlighting behavior between tables. I have a [test document](https://docs.google.com/document/d/1a7c5O0QSHc6Rs6Z0lUN67TbasuM9W-bWJAIVsm34tjo/edit?usp=sharing) with 2 random citations. Citation A has 2 items, and citation B - just one:

Screenshot 2025-09-24 at 5 37 21 PM

I expect that if I select either of the two items from citation A in the itemTree to the right, the citation A would be highlighted, since the items belong there. But instead, if I select the second item, the citation B gets highlighted:

Screenshot 2025-09-24 at 5 37 34 PM

And if I select the item that belongs to citation B, citation A gets highlighted then:
Screenshot 2025-09-24 at 5 37 41 PM

That feels inconsistent? Or am I just misunderstanding the workflow here?

adomasven:
> > This is more of a design consideration. Overall, probably the most reasonable thing to do is to disable multi-select in these trees. There are two reasons for multi-select: being able to highlight multiple items in the other tree by selecting multiple in one (in case you're troubleshooting where which citation is included, or what have you), the other is multiple Show in Zotero, which you correctly observe, currently doesn't work.
> 
> With these two use cases, multi-select feels quite important here. One should be able to find out which items are in a set of citations, right? And being able to find a selection of items in Zotero library is a necessary step to be able to, say, place all of your cited items into a collection before exporting them somewhere (which I believe is a frequently asked about use case).

Yeah, wanting to select all items and add to a collection in Zotero is a valid use case and users will want to do that. There's some caveats here:

1. The more obvious one: if you select multiple items with some unlinked to Zotero, then Show in Zotero won't have a target to highlight for those items. That's probably fine and not incredibly unexpected as a UI behaviour.
2. Less obvious: if the document contains items cited from multiple libraries, and you press Show in Zotero, the behaviour is undefined.

One unorthodox solution for 2 would be to add a temporary virtual collection in Zotero when Citation Explorer dialog is open, and highlight items there. Then, make that collection in Zotero contain more information, in the item pane, showing whether the item is linked, and from which library, and perhaps with some columns too, so that you can sort on them. The virtual collection disappears on citation explorer close.

abaevbog:
> The more obvious one: if you select multiple items with some unlinked to Zotero, then Show in Zotero won't have a target to highlight for those items.

Could this lead to some unexpected outcomes for the end user? If you are trying to select all cited sources to add to a collection and then export the list somewhere, I think you might end up exporting an incomplete list of cited items (no unlinked ones) without knowing it. If you have tens of sources, it's easy to miss one or two unlinked items if you don't scroll through the entire table. 

An alternative would be to re-save all selected unlinked items before selecting them all in Zotero? The drawback of that is obviously that we risk creating duplicates, which is not great either.

What if we keep the current behavior of skipping unlinked items but add a clear indication of how many unlinked items there are and add a way to save all of them to selected library in bulk? I am thinking of something like a message "Unlinked items: ${count}" somewhere below the itemTree with a button to save all of them to the selected group? That way, the user is somewhat warned that there are unlinked items (even if they don't immediately see it in the long list of sources) and allow them to fix it with just one click.

 > Less obvious: if the document contains items cited from multiple libraries, and you press Show in Zotero, the behaviour is undefined.
> One unorthodox solution for 2 would be to add a temporary virtual collection in Zotero when Citation Explorer dialog is open, and highlight items there.

That's definitely an option! We briefly discussed something similar in the context of the citation dialog. I think the concern that we had then (also applicable here) is that for one task the user would have to work with two interfaces, potentially jumping back and forth between them. If I understand correctly, that would mean removing the itemTree from the "Citation Explorer" dialog and when a citation is selected in the dialog, we highlight items in the main window? I am afraid it can lead to many questions on the forums about this workflow across two windows ... Also, it would be very challenging to make it functional for screen reader users.
If we add a pseudo collection and keep the itemTree in the citation explorer, we would have two essentially identical itemTrees across two windows, which can also be confusing. One would have to always think "Which items list should I go to in order to do X".

If we explore the option of using the main Zotero windows as the UI component for the "Citation Explorer" integration,  we could consider placing the citations list in the main Zotero window as well, and not use the dialog at all? The itemTree already exists in ZoteroPane, and citations list could appear in the itemPane. The benefit of this is that one could take advantage of all the existing methods to add items to collections and copy them between groups out of the box. The drawback is that this is not something that we do anywhere else. Also, without the explicit opening and closing of a separate window, I am not sure how the workflow would look like for the end user (e.g. how do you confirm the changes?).

As a completely different solution, we can just allow "Show in Zotero" only when selected items all belong to the same group. But for this, we would need to distinguish between items from different groups in the itemTree. We could display items grouped by their library, perhaps similar to what you suggested earlier in "Insert Annotations" pr: https://github.com/zotero/zotero/pull/5356#issuecomment-3011840655 but with library names as headers? From the UI perspective, I think it is more intuitive but it raises the question of how would one export all cited items then? We could just add buttons with all necessary functionality into the dialog, such as "Export", "Create Bibliography", etc. Then, maybe one could directly get the info that they need from the dialog without having to go to the main window. And if the use case is more niche, one can always export all items from all groups and then import them back into a collection in "My Library".

adomasven:
I've made a version which disables buttons unless there is a selection, and I think it's bad. I'd rather the buttons only appear when there's an action for a corresponding item.

image

@abaevbog:

Upon more thinking, I think we should make the `Add to Collection` button always enabled, for single and multiple selection, that way the ability to save all items into a collection is right there in the dialog.

The only remaining question is what to do with items that are not from the same library. The simplest solution is to add another column "Library" and only show it when there are items cited from multiple libraries (that way it won't impact the workflow or look in most cases), and when multiple items are selected from multiple libraries - disallow actions.

I can't think of a more elegant way to solve this. A virtual collection (solely for the purposes of dealing with multi-library cited-items, I'm not thinking of moving any citation explorer functionality into it, like I had suggested for the citation window - I don't think that's appropriate here) would be one, but it's only needed for addressing a fairly minor edge-case, and one that could be solved with an optional column. Any other ideas?

adomasven:
@yexingsha Your comments have been addressed

Screenshot from 2025-09-29 16-25-35

abaevbog:
> I've made a version which disables buttons unless there is a selection, and I think it's bad. I'd rather the buttons only appear when there's an action for a corresponding item.

Yeah, it looks a bit busy on your screenshot... I think on mac it looks better, as long as the window is not shrunk to the min width. Personally, I do prefer this version with disabled buttons because I find it helpful to always see what options are available. For instance, as soon as the dialog opens for the first time, I immediately see that I can "Edit a citation", as opposed to discovering it later and having to somewhat remember that it's one of the options. But maybe that's just my personal perference. @yexingsha, do you have any thoughts on the above?

Screenshot 2025-09-29 at 9 38 29 AM

> Upon more thinking, I think we should make the Add to Collection button always enabled, for single and multiple selection, that way the ability to save all items into a collection is right there in the dialog.

Yes, I fully agree with that. What do you think about also having the "Export" button? "Add to Collection" naturally won't work for items across different libraries but one might need to get the full list of items sometimes? With this, we may be running into the issue of having just too many buttons squashed next to one another. What do you think about moving the buttons relevant for multiple selected items into a `menupopup` as `menuitems`? Then we could replace most of the buttons below itemTree ("Add to Collection", "Export", "Show in Zotero") with a single `toolbarbutton` labelled "Options" or "Actions", or something like that?

> The only remaining question is what to do with items that are not from the same library. The simplest solution is to add another column "Library" and only show it when there are items cited from multiple libraries 

Yeah, disabling the buttons when items from different collections are selected sounds reasonable to me. The column would work fine functionally. I think it may not be immediately clear that some options get disabled because multiple selected items are in different libraries - one would have to kinda discover it. We can remedy that by adding a tooltip to relevant buttons explaining why they are disabled. Another drawback is that space is somewhat limited here, so it feels a bit wasteful to have a column that will often have repeated information. Instead of the column, we could do the grouping of items by their libraries (as we tried briefly in the [annotations pr](https://github.com/zotero/zotero/pull/5356#issuecomment-3051089977)). That comes with various challenges of reworking the itemTree though and is probably just an overkill here.

yexingsha:
The UI looks mostly fine now, except that:

- On macOS, the buttons now have too little bottom margin.
- When the "Add to..." button is disabled, the icon should use `accent-blue50`.

I still prefer disabling the buttons instead of hiding them, for the same reasons @abaevbog outlined. And disabled buttons can provide tooltips explaining why they are disabled.

As to how to handle actions on items from different libraries, I agree that adding a "Library" column is probably not that helpful. Even with this column, I assume most people will only figure out they can't show/add items from different libraries *after* they have multiple items selected and see the tooltip on a disabled button, and at that point it will still be a hassle to modify their selection. So I think grouping items by libraries, as we tried in the annotations PR, is actually pretty useful here. Since this dialog does not have a collectionTree, the issues we encountered there no longer apply. And it's less wasteful than a whole column.

If we do add "Export", then I agree it's a good idea to put these actions under the same "Actions..." button. But I wonder what's the use case for "Export" in this dialog, and how many people would use it?

abaevbog:
> If we do add "Export", then I agree it's a good idea to put these actions under the same "Actions..." button. But I wonder what's the use case for "Export" in this dialog, and how many people would use it?

Well, without it, if one has cited items across 5 different groups, in order to export a full list of citations, one would have to select all items for each of those 5 libraries, export them one library at a time in the main window, and then put those parts together, right? And that sounds like a lot of unnecessary work for the user. Here, we can help them export the entire list in one go.

> So I think grouping items by libraries, as we tried in the annotations PR, is actually pretty useful here.

@adomasven, the itemTree sections setup I put together in the annotations PR was pretty raw, since it was just an experiment (that we didn't end up going for). But if it would be helpful, I can see if any of those parts could be useful here. Let me know!

adomasven:
> Yes, I fully agree with that. What do you think about also having the "Export" button?

I think exporting shouldn't be within the scope of this dialog. Generally, users shouldn't be citing items from different libraries, although they might by accident, so items across more than 2 libraries will be exceedingly rare. Furthermore, exporting and importing items unlinks them from citations in the document, so we should not be encouraging that here, which it would seem like we do.

If a user needs to move items into a single library, they can Show in Zotero and move them manually, then re-cite.

> I think grouping items by libraries, as we tried in the annotations PR, is actually pretty useful here.

Sounds good. We can also disallow multiple-selecting across the library boundary, so the buttons don't have to be disabled.

dstillman:
I don't think "Export" is the right action here, but helping users consolidate citations from different libraries into a single library has to be within scope here — it's one of the most important reasons for this dialog. (I recently responded to two <a href="https://forums.zotero.org/discussion/127786/add-new-advanced-option-to-avoid-common-mistake-in-collaborative-documents">forum</a> <a href="https://forums.zotero.org/discussion/127784/giving-access-to-my-entire-library-to-a-book-manuscript-co-author">posts</a> complaining about this problem, and we've said [for years](https://forums.zotero.org/discussion/comment/427497/#Comment_427497) that we're going to solve it.) Currently, if you start citing in My Library (like the person in those last two threads), there's no way to start working with a collaborator in a group library without manually relinking huge numbers of items. Same if you only have an import of a library (e.g., because you made an export instead of a proper backup) — while you could add all the citations back to the library from the current dialog, you'd end up with duplicates. It's not realistic to tell people they should've started from a group library from the beginning, or that they should have made a proper backup — we need to provide a solution here.
    
We ideally would address a bunch of different situations:
    
1. Unlinked citations that don't have equivalent items in any library (e.g., a document you got from someone that you want to continue working on)
2. Unlinked citations that have equivalent items in the target library (e.g., a bad library restore via import/export)
3. Unlinked citations that have equivalent items in one library but not the target library (e.g., a bad library restore via import/export that was then dragged to a group library)
4. Linked citations that don't have equivalent items in the target library (e.g., the second forum example above — start in My Library, cite a subset of items, and then want to copy just those items to a group library)
5. Linked citations with items that also exist in the target library as linked (`linkedObjectPredicate`) items (e.g., starting in My Library and dragging items to a group library, or citing from the wrong library by mistake)
6. Linked citations with equivalent items in the target library that aren't linked (e.g., an export/import into a group library)
7. A combination of the above (e.g., unlinked or linked citations drawn from items in one or more libraries that may or may not be linked items)
8. Wanting to add all items used in a document — in various states of linkedness — to a collection in a given library
    
(1) is easy and addressed now — you select all items and use "Add to [Target Library]". None of the others currently have good solutions.

As a start, I think it could help to reconsider Relink Item. It doesn't really make sense that you can relink a citation to any random item in the library — it's error-prone and slow. You can already do that via Add/Edit Citation if you really want to, but this dialog shouldn't make substantive changes to your document. (Assuming Edit Citation pops up the citation dialog, I'm not sure we should even have that here, out of context.) So I think Relink Item should only suggest equivalent/linked items across all your libraries. But you also want to be able to do this for multiple items at once, and to choose a target library where some of the items may not have equivalent/linked items. Well, that starts to sound like "Add to [Target Library]"!

So…what if we only have a single function — maybe called "Add/Link to Library…" — that operates on multiple items and is more of a wizard (either traditional page-based or more of a vertical sectioned thing that scrolls up as you go)? You first select a target library, and then optionally a target collection (existing or new), and then it could have sections for each action with the items that will have that action: "Relink to Matching Items", "Copy from My Library", "Copy from [other library the object is linked to]", "Copy from Document", "Add to Collection", "No Change". (For matching items, where it's essentially duplicate detection, there could maybe be drop-downs to choose another source instead, and/or some way to choose between duplicates.) And then, for the final step, it could even show a diff of the bibliography with the changes that will occur if you proceed (e.g., if there are metadata differences across libraries).

Matching beyond linked-item relations would be done based on identifier or title + creator + year, similar to duplicate detection.

Copying across libraries would use our existing code to created linked-item relations and follow the "When copying items between libraries, include:" settings.

Since you probably want all items to be linked to the same place, I'm not sure it would even makes sense for this to operate on selected items rather than all items, since anything already in the specified library/collection is just "No Change". But "Show in Zotero" will operate on selected items, so we'd probably have to somehow present the button differently if it didn't depend on the selection. I guess it could just operate on all items (or all filtered items?) if nothing was selected.

So then, to review the situations outlined above:

- For (1), you choose My Library → it shows "Copy from Document" for all items.
- For (2), you choose a library → it shows "Relink to Matching Items" for all items.
- For (3), you choose a group library → it shows "Copy from My Library" for all items (based on matching the items in My Library).
- For (4), you choose a group library → it shows "Copy from My Library" for all items (based on the linked citations).
- For (5), you choose a group library → it shows "Relink to Matching Items" for all items (based on the inter-library links).
- For (6), you choose a group library → it shows "Relink to Matching Items" for all items (based on matching the items in the group library)
- For (7), a combination of the above
- For (8), you choose a library and choose an empty collection → it shows a combination of the above, or "Add to Collection" for items that are already linked to the target library

If this sounds reasonable and worth exploring, it would probably help for @yexingsha to create some mockups to see what it might look like.

abaevbog:
I meant to mention that all logic for copying items between libraries currently is in `collectionTree`, and even then it's not quite complete (e.g. if linked items exist but their metadata is different, there's no way to easily fix it).

In https://github.com/zotero/zotero/pull/5577/, among other things, I pulled the copying of items into dedicated methods, like `Zotero.Items.copyToLibrary`, which could come in handy for copying items between groups.

dstillman:
https://forums.zotero.org/discussion/128497/sharing-a-subcollection

adomasven:
Finally, some updates! After heavily refactoring item tree so that we can support other things than zotero items as item roots, we now have sectioned items:

image

Moreover, a wizard for all your item adjustment needs!

image

image

And an item relink picker with suggestions:

image

@abaevbog @yexingsha @dstillman I believe this is now in a mergable state. You can test it in Google Docs, where Citation Explorer menu entry will appear upon detection of this build.

abaevbog:
Very exciting! This works great!

The only issue I have ran into so far is the "Show in Document" button does not work for me in a fresh document. I keep getting the warning that citations have been changed in my document since when the explorer was opened.

One other thought is if there should be any additional handling at the relinking step of an unlinked item? Suppose I link a cited item to an item in my library but the item in Zotero has a somewhat different metadata. It can significantly change the citations from what they were before - should we warn about it in some way?

I'll keep playing around to see if I have anything else!

PR #3470, opened 2023-11-01
Fix editorInstance _formatCitationItemPreview

windingwind:
If `title` exists, the `authors` will be overwritten, which I suppose is unexpected.


windingwind:
Oh wait.. I didn't see the `if (!str`. So I guess this is designed to be so. Sorry I'll close the PR

PR #3526, opened 2023-12-06
Displays a browser window to clear captcha when saving attachments.

adomasven:
- Currently enabled only for ScienceDirect. Can be enabled via a whitelist
- Matches the HiddenBrowser loaded HTML page for a captcha element. If the captcha element class changes, this will break (but the alternative is potentially displaying a captcha clearing window when something else that is not a captcha guard is loaded).
- User has 60s to clear the captcha, otherwise we timeout and fail the save.
- Doesn't automatically switch focus back to the browser which intiated the item save via the Connector.
- Stores the cookies used to clear the captcha for future saves from the same domain. Discards Connector supplied User Agent, since CF bot detector checks UA header against actual UA behavior like TLS handshake and if the UA acts different to what it's supposed to, the bot challenge is not cleared.

Other changes:
- Adjusted the cookie sandbox to allow multiple cookie sandboxes to be active (and simplified some legacy code that was meant to cover a bug in old FX codebase).
- HiddenBrowser API changed to be Object oriented, translator tester in the translate repo will need to be updated after a merge (have the change ready).
- Improved Connector Server attachment progress handling

### DIFF
--- a/chrome/content/zotero/HiddenBrowser.jsm
+++ b/chrome/content/zotero/HiddenBrowser.jsm
@@ -50,22 +50,85 @@ ChromeUtils.registerWindowActor("SingleFile", {
 });
 
 const progressListeners = new Set();
-const browserFrameMap = new WeakMap();
 
 /**
  * Functions for creating and destroying hidden browser objects
  **/
-const HiddenBrowser = {
+class HiddenBrowser {
 	/**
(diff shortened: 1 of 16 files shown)


adomasven:
Report: https://forums.zotero.org/discussion/comment/448431/#Comment_448431
Test URL: https://www.sciencedirect.com/science/article/pii/S0378429019313863?via%3Dihub

PR #3544, opened 2023-12-20
Add an "Open Documents" section to the citation dialog. Closes #3332

### LINKED ISSUE #3332: "Open" section of citation dialog search results
https://forums.zotero.org/discussion/106988/could-the-opening-pdf-in-zotero-be-easily-found-in-the-word-insert-citation-interface

If you're using the built-in reader, you're probably almost always citing something you have open, so we should probably just open the search results by default if there are any open tabs. (We could consider opening it to show Cited as well.)

As you search, the Open section should remain at the top with any matches.

adomasven:
Adds an "Open Documents" section to the citation dialog that displays and searches currently opened document tabs in Zotero, which is displayed at the top, above other sections.

### DIFF
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -631,7 +631,7 @@ var Zotero_QuickFormat = new function () {
 		}
 		
 		// Clear item list
-		while (referenceBox.hasChildNodes()) referenceBox.removeChild(referenceBox.firstChild);
+		_clearEntryList();
 		
 		// Take into account items cited in this citation. This means that the sorting isn't
 		// exactly by # of items cited from each library, but maybe it's better this way.
@@ -702,6 +702,7 @@ var Zotero_QuickFormat = new function () {
 			previousLibrary = libraryID;
 		}
 
+		_resize();
 		if (!referenceBox.children.length) {
 			return;
 		}
@@ -716,7 +717,6 @@ var Zotero_QuickFormat = new function () {
 			});
 		}
 		
-		_resize();
 		referenceBox.selectedIndex = selectedIndex;
 		referenceBox.ensureIndexIsVisible(selectedIndex);
 	}
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -46,6 +46,7 @@ var Zotero_QuickFormat = new function () {
 	
 	const SEARCH_TIMEOUT = 250;
 	const SHOWN_REFERENCES = 7;
+	const ITEM_LIST_MAX_ITEMS = 50;
 	
 	/**
 	 * Pre-initialization, when the dialog has loaded but has not yet appeared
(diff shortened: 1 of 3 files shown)


dstillman:
Nice! But this seems to break something fundamental about the search — I'm now not getting results for basic terms like "test" until…some I type various other things, and eventually everything starts working. Not sure exactly what's fixing it.

What's the intended behavior? For search results to appear with Open Documents as soon as the dialog opens, if you have any open tabs?

dstillman:
This is neat!

Seems like we should sort reverse-chronologically by last open, though — just pressing Enter should select the tab you have open rather than your first tab. That might more or less eliminate typing into the citation dialog for people using the built-in reader.

We'll have a persistent last-open date after https://github.com/zotero/zotero/pull/2854, but for now we could track tab focus times in memory in tabs.js and then include tabs that haven't been opened this session in reverse tab order (which should be closer to reverse-chronological).

Could also show items selected in the items list in their own section, which would help people who don't use the reader, though there'd have to be a limit in case you had 5,000 items selected. I had figured that inserting the current selected item would be a separate button, and that might still be best, but if we're opening the search results here regardless, maybe one extra click isn't so bad. Show "Open Documents" first if a tab or reader window is selected/front-most and "Selected Items" first if the library tab is selected, so that Enter just does the thing you probably want?

adomasven:
> Could also show items selected in the items list in their own section, which would help people who don't use the reader or want to find something without opening it — and then we could get rid of the classic citation dialog. There'd have to be a limit in case you had 5,000 items selected. I had figured that inserting the current selected item would be a separate button, and that might still be best, but if we're opening the search results here regardless, maybe having to press Enter twice isn't so bad. Show "Open Documents" first if a tab or reader window is selected/front-most and "Selected Items" first if the library tab is selected, so that Enter just does the thing you probably want?

Excellent ideas. Indeed a good way to address inserting the currently selected item. The current reference list code is limited to displaying 50 items in total, starting with open documents, then cited items, then items from the library, so we will keep that, mostly for performance reasons. The user can always filter the list by typing.

stakats:
Is there any risk here of "Open Documents" appearing to users like an action rather than a state? Should it be called "Currently Open Documents" instead?

adomasven:
Comments addressed

dstillman:
This is great.

One problem: if I press Return to select the pre-selected item from Open Documents or Selected Items and then press the space bar, I can't accept the dialog without adding another item, even if I press Backspace.

Relatedly, we should also remove an item from the results once it's already been chosen. No legit reason you'd ever need to select more than once, I think?

dstillman:
So I guess the problem is the Backspace behavior — if you press Backspace so that you're back in the previous citation, then pressing Return should close the dialog.

And then removing already-chosen items from the results would make it so that you could select multiple selected/open items with Return + space + Return, etc.

adomasven:
Comments addressed.

adomasven:
> Should a note item show up for Add/Edit Citation? 

It certainly shouldn't as we are not set up to mix normal and note item insertion.

dstillman:
If you have one item selected in Zotero, choose that one, and press space, the Selected Items header still shows up, but it's empty.

Screenshot 2024-01-15 at 7 48 59 AM

adomasven:
> Can we allow inserting multiple selected/open items with Shift-up/down-arrow?

We probably could, but that requires meddling in the code that is reworked in https://github.com/zotero/zotero/pull/3464, so I would rather just merge this now and consider that for later.

dstillman:
Still seeing the empty white bar for non-matches

adomasven:
Fixed

adomasven:
> Relatedly, we should also remove an item from the results once it's already been chosen. No legit reason you'd ever need to select more than once, I think?

A legit reason has surfaced quite quickly https://forums.zotero.org/discussion/111831/multiple-citations-of-the-same-source

dstillman:
Hmph.

> The same item still won't appear multiple times in the Open Documents section, but you can look for it in your Library or Cited section

Yeah, that's good.

PR #3577, opened 2024-01-25
Fix attachment row annotation button bug

windingwind:


dstillman:
Can just be `.isFileAttachment()`

https://github.com/zotero/zotero/blob/9151fb35587f947244a21c8b5727c0c523812a2a/chrome/content/zotero/xpcom/data/item.js#L2390-L2395

PR #3586, opened 2024-01-25
Animate virtualized table containers

tnajdek:
This PR adds slide down/up animation when expanding/collapsing collections and items in a virtualized tree. 

https://github.com/zotero/zotero/assets/214628/39cfb53f-1ea9-43de-80d4-32c40013a85c

### DIFF
--- a/chrome/content/zotero/collectionTree.jsx
+++ b/chrome/content/zotero/collectionTree.jsx
@@ -404,16 +404,7 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 			}
 		}
 
-		// since row has been re-rendered, if it has been toggled open/close, we need to force twisty animation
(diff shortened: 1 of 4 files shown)


adomasven:
Presumably this imitates macOS Finder? I personally really dislike animations like these and disable them if possible. I also wonder how it works for expand/collapse all/multiple rows. If we are doing this, we should at least make the animation imitate macOS better where the time to expand/collapse is constant regardless of the number of items and ease in and out.

tnajdek:
> expand/collapse is constant regardless of the number of items and ease in and out.

The current implementation is linear 200ms (constant regardless of the number of items). All the rows are hidden behind the parent row so each child row gets a different transition duration/delay to create a coherent animation and there is no need to manipulate z-index of other rows. Adding easing will require some curve-fitting math or changing this approach but sure can be done.

PR #3589, opened 2024-01-25
Fix `editable-text` expansion with non-overlay scrollbars on macOS

AbeJellinek:
Simpler than I thought it would be! @abaevbog, see if this works for you?

(We should put together some sort of `editable-text` test suite/document with fields in various configurations, weird text that's likely to trigger wrapping edge cases, and so on.)

### DIFF
--- a/scss/elements/_editableText.scss
+++ b/scss/elements/_editableText.scss
@@ -100,10 +100,6 @@ editable-text {
 	}
 	
 	&[multiline] {
-		&::after, .input {
-			overflow-y: auto;
-		}
-		
 		.input {
 			min-height: 5em;
 		}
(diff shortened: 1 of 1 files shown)


abaevbog:
I this this already is a huge improvement! I am still running into a small edge case that still boils down to the text wrapping at different spots for `::after` and the actual text. So the actual text looks right but there's this seemingly empty line at the bottom which comes from the `::after` component ...


https://github.com/zotero/zotero/assets/36271954/8bca69e7-e986-4c52-b811-0c646db3a887

abaevbog:
That was before the second commit, yeah.

Now, the titles look right and everything is matching up! But ... that seems to affect the fields that (I am guessing) do not need a scrollbar in a similar way as before that makes the `::after` add an extra line. 

Screenshot 2024-01-25 at 3 30 18 PM

abaevbog:
I think it does it! I see now phantom lines anymore!

With book title, I think you are right. We probably want `textarea` that is not `multiline`, just like the title in the header.

Issue #3593, opened 2024-01-26
Can't open Add Attachment and New Note menus with Return or Space


dstillman:
https://forums.zotero.org/discussion/111129/zotero-7-attachment-and-note-item-dropdowns-arent-keyboard-accessible

PR #3602, opened 2024-01-26
additions to keyboard nav for toolbar buttons and reader

abaevbog:
- Space/Enter open the menus of "add note" and "add attachment" buttons
- Enter triggers the "New Collection" button for consistency
- Added locate button accessibility via tab from itemPane header or via shift-tab from the scrollable area of itemPane.

Fixes: #3593

### DIFF
--- a/chrome/content/zotero/contextPane.js
+++ b/chrome/content/zotero/contextPane.js
@@ -277,8 +277,9 @@ var ZoteroContextPane = new function () {
 
 		if (splitter.getAttribute('state') != 'collapsed') {
 			if (_panesDeck.selectedIndex == 0) {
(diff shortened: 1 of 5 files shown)


abaevbog:
This handles two separate issues but there's some overlap with the file changes, so I put it in one PR to not worry about conflicts

abaevbog:
The second commit is my suggestion to improve the keyboard navigation when a reader tab is opened. If you tab into the reader, there seems to be no good way to get the focus out to, say, focus the tabs list to go to another tab. My suggestion is to focus the selected tab on Escape, from where one can either keep tabbing to get to the context pane or hit Enter to re-focus the reader again.

dstillman:
This is great!

Issue #3603, opened 2024-01-26
Tag selector keyboard navigation


abaevbog:
There seems to still not be a good way to navigate the tag selector without using a mouse.

I am thinking we should try to do something similar to collectionTree? If there are selected tags, Shift-Tab from the tag selector input or Tab from the collectionTree will focus the first selected tag. If there are no selected tags, the first tag will be focused.

From there, ArrowRight-ArrowLeft will move between tags. Shift-Tab from the tag selector will focus the collectionTree and Tab will focus the input field.

PR #3615, opened 2024-01-28
Keyboard navigation for tag selector

abaevbog:
- Tab from the tags list or shift-tab from the tags filter field focuses the first non-disabled tag. If there are none, the tags are skipped and the focus moves directly to the input field or the tags list.
- Arrow Right/Left move focus between tags skipping over disabled tags
- Space/Enter clicks on the selected tag
- Space/Enter click on the search button when focused

Fixes: #3603


abaevbog:
The one question is what should be the first tag to be focused first on, say, Shift-tab from the input list? I started by trying to focus the first non-disabled tag node `document.querySelector('.tag-selector-item:not(.disabled)')` - is that reasonable?

dstillman:
Awesome!

Issue #3610, opened 2024-01-27
"(first)" / "(last)" written to creator fields


dstillman:
https://forums.zotero.org/discussion/111206/zotero-7-beta-last-and-first-written-in-the-author-fields-when-left-empty

PR #3618, opened 2024-01-29
fixes to itemBox creator functionality

abaevbog:
- Instead of adding/removing the default names to the fields of empty creator names, set them as placeholders - that way they appear only when the actual creator name field is empty, and it requires no additional logic to add/remove them.
- Re-do of `capitalizeCreatorName` so that it works when the creator name field is focused. https://forums.zotero.org/discussion/111304/zotero-7-beta-fix-case-on-author-field-is-not-working-from-the-right-click-menu
- Fix to get the creator name contextmenu to appear for un-saved creators. That way one can switch the `fieldMode` before saving the creator. https://forums.zotero.org/discussion/111297/impossible-to-select-single-field-before-entering-an-author-7-0-0-beta-56-9edfcba9a
- Fix to tab/Shift-tab focusing wrong element right after a new creator row is added

Fixes: #3610
Fixes: #3666

### DIFF
--- a/chrome/content/zotero/elements/editableText.js
+++ b/chrome/content/zotero/elements/editableText.js
@@ -140,7 +140,7 @@
 		sizeToContent = () => {
 			// Add a temp span, fetch its width with current paddings and set max-width based on that
 			let span = document.createElement("span");
-			span.innerText = this.value;
+			span.innerText = this.value || this.placeholder;
 			this.append(span);
 			let size = span.getBoundingClientRect();
 			this.style['max-width'] = `calc(${size.width}px)`;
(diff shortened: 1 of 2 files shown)


dstillman:
No, this isn't the right approach — it's conflating state and value. The software knows if it was showing a placeholder. It doesn't need to compare text to determine that. If I enter the field and type `(first)`, that's the value that should be saved.

abaevbog:
That makes sense. There was some explicit comparison of text from before-redesign times, so I erroneously assumed we never-ever want to have `(first)`, `(last)` and etc.

As I was working around setting and removing the default values, I realized that now we have `editable-text`, so why don't we use the proper `placeholder` attribute on the input? That removes the need to keep track of the default values altogether.

I did that and I kept the style unchanged, so `(first)` and `(last)` show up as blurry placeholders, which I think looks nice? If we want to keep the old look though, we can tweak the stylesheet to make them the same color as the real text

PR #3624, opened 2024-01-29
Item Pane API and CEify itemPane, contextPane

windingwind:


dstillman:
Let's be explicit and call this ItemPaneSection, not just PaneSection.

windingwind:
Some explanations:

The old `ZoteroItemPane` is split into two CEs:
1. `ItemPane`, which contains all sections like info, abstract, etc.
2. `LibraryPane`, which contains the `ItemMessagePane`, `ItemPane`, `NoteEditor`, and a `DuplicatesMergePane`

The old `Zotero_Duplicates_Pane` is replaced by the new CE `DuplicatesMergePane`.

The message box to display N items selected is replaced by the new CE `ItemMessagePane`.

To be discussed:

1. Do we need `PaneSectionElementBase`, which is now the ancestor class of all item pane sections and we can add some checks (e.g., must have a `render`) when we have the API for plugins?
2. I think we need to CEify the contextPane, 'cause sharing the same CE (which is the current `ItemPane`) between the library and reader would make it clearer to create an API for plugins.

chrome/content/zotero/elements/libraryPane.js
```diff
@@ -0,0 +1,505 @@
...
+
+		setItemPaneMessage(msg) {
+			this.viewType = "message";
+			this._messagePane.render(msg);
+		}
+		
+		/**
+		 * Display buttons at top of item pane depending on context
+		 */
+		updateItemPaneButtons() {
```
windingwind, line 263: This renders the item pane buttons

chrome/content/zotero/elements/itemPane.js
```diff
@@ -0,0 +1,522 @@
...
+				else {
+					box.mode = 'edit';
+				}
+				
+				box.item = item;
+				box.inTrash = inTrash;
+				// Render sync boxes immediately
+				if (!box.hidden && box.render) {
+					if (box.render instanceof AsyncFunction) {
+						pendingBoxes.push(box);
```
windingwind, line 204: During item pane rendering, only sync `render` functions are called immediately and all async `render` functions wait until the scroll to section finishes. This fixes bugs with wrong scrolling positions caused by changing box height during scrolling.

chrome/content/zotero/elements/itemPane.js
```diff
@@ -0,0 +1,522 @@
...
+
+			// Only render visible panes
+			for (let box of pendingBoxes) {
+				if (pinnedIndex > -1 && panes.indexOf(box) < pinnedIndex) {
+					continue;
+				}
+				if (!this.isPaneVisible(box.dataset.pane)) {
+					continue;
+				}
+				await box.render();
```
windingwind, line 234: After scrolling, check all boxes that have pending rendering tasks. Only those that are visible will be called. The invisible boxes are rendered when scrolling into view.

chrome/content/zotero/elements/itemPane.js
```diff
@@ -0,0 +1,522 @@
...
+				if (!box.secondaryRender) {
+					continue;
+				}
+				if (pinnedIndex > -1 && panes.indexOf(box) < pinnedIndex) {
+					continue;
+				}
+				if (this.isPaneVisible(box.dataset.pane)) {
+					continue;
+				}
+				await box.secondaryRender();
```
windingwind, line 247: We also allow a `secondaryRender`, which allows you to roughly render some initial content/set a box height in sync, and then render real contents that have to go in async later when needed. Would be useful for plugin API.

The setting data and rendering steps are separately controlled by the CE and plugins can lazy-load their panes to minimize resource usage (so does our attachment preview).

dstillman:
Ah, got it. `LibraryPane` is too broad of a name, though. Maybe that's the `ItemPane` and the thing containing Info/Abstract/etc. is `ItemDetails`? In our parlance the "item pane" is really just the third pane in the Zotero UI, no matter what's showing there.

chrome/content/zotero/elements/itemPane.js
dstillman: Neat. I guess one question is…should plugins be able to render anything synchronously, or just set a box height?

windingwind:
I've added some comments to relieve the burden of understanding this PR. It's mainly about:

1. CEify itemPane to have better control of the item pane rendering process for future APIs and some bug fixes
2. Implement the header buttons in redesign (Please check @yexingsha )
3. (Not done yet) CEify the contextPane to share the `ItemPane` CE for better plugin API integration

chrome/content/zotero/elements/itemPane.js
windingwind: If the rendering is fast, e.g. display some text, fine, I guess? We could have some warnings for sync render functions that take too long to recommend them to use async.

windingwind:
If there are no big problems, I'll proceed in this direction. Not sure if we can get the API out this week, but heading to the goal!

yexingsha:
wrt header buttons, there is this issue https://github.com/zotero/zotero-redesign/issues/250

chrome/content/zotero/elements/attachmentAnnotationsBox.js
```diff
@@ -40,7 +40,7 @@
 			this._item = item;
 			if (item?.isFileAttachment()) {
 				this.hidden = false;
-				this.render();
+				// this.render();
```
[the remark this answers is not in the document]
windingwind: To have full control of the rendering process, I decide to uncouple setting data & rendering calls and let the rendering be fully controlled by the `ItemDetails` CE. These comments will be removed

chrome/content/zotero/elements/itemMessagePane.js
```diff
@@ -0,0 +1,71 @@
...
+		renderCustomHead(callback) {
+			let customHead = this.querySelector(".custom-head");
+			customHead.replaceChildren();
+			let append = (...args) => {
+				customHead.append(...args);
+			};
+			if (callback) callback({
+				doc: document,
+				append: (...args) => {
+					append(...Components.utils.cloneInto(args, window, { wrapReflectors: true, cloneFunctions: true }));
```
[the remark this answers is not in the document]
windingwind: I am unsure whether this is necessary... the design was borrowed from `Zotero.Reader.registerEventListener`.

Would it be necessary when we have multiple main windows?

windingwind:
OK, `contextPane` CEify done now.

Changes:

The old `ZoteroContextPane` is separated into new CE `contextPane` (which handles item/notes switch and state update across tabs), new CE `notesContext` (which holds notes lists for a given libraryID and controls how the notes are displayed). The remaining part in `ZoteroContextPane` is responsible for the collapse/resize handle.


TODO:

Unify `itemPane` as a child of `contextPane` and manage all third-pane in Zotero in one place. Then I will go on with the API implementation.

windingwind:
Run this to try the API: https://gist.github.com/windingwind/afdf64046d887f5b6750b225ae2fc822

windingwind:
> In terms of the API, why don't we just let plugins extend the base class that our own sections extend

A closed set is safer and a open set is dangerous- allowing plugins to extend a CE and inject the unknown CE into our layout could be a disaster. People can do the weirdest thing you could imagine inside the CE!

On the other hand, we need to control the rendering, which you can’t expect people to do it right in the CE. With hooks we can validate them in previous.

windingwind:
> True, but they can add whatever they want inside onInit/onRender/fragment already. If they want to mess up the layout or delete your home directory, they can!

yeah, exactly. at least we now force them to follow the `collapsible-section` design and the container looks good. As for what's inside... that's not our business.

> a bit less powerful than a full CE-based API would be.

Plugins can do almost everything that we do with a CE (creating contents, add section buttons, and I will soon add the ability to create context menus), so the gap is relatively small.

> We should provide a way for plugins to style their panes with CSS

Yes. I think icons can keep using `style`, and we should validate the icon size to force them to use a `16` or `20` icon. As for the CSS, how about allowing them to set a class name to the `body` element? (I don't want to allow plugins to control the section header, though)

windingwind:
Mentioning the safety of the hook, I even want to pass a proxied value as the `body` parameter, to shadow the `parentElement`, `parenNode`, and `remove` to prevent plugins access the container and hacking strange things... Would that be too radical?

mrtcode:
The new context pane currently doesn't preserve the note pinning behavior which differs for item notes and all notes. The main idea is that item note is only pinned for the current item, and a note from "all notes" affects other tabs as well. But there are various nuances.

windingwind:
> Plugins already have access to the window and its entire DOM, and they run with chrome privileges.

yeah, privileges aren't bad, but a badly designed plugin is. the item pane sections are dynamically generated in readers, so plugins won't be able to reliably access its DOM. in that case, if we block access to the parent element, that would work.

windingwind:
The example updated: https://gist.github.com/windingwind/afdf64046d887f5b6750b225ae2fc822

One weirdness I found is that after start Z - register custom section - close and reopen main window, the `onSecondaryRender ` call stops at the await line 25, and the rest lines are not executed. What could be the cause? But it works as expected without the window reopening.

chrome/content/zotero/elements/itemDetails.js
```diff
@@ -0,0 +1,582 @@
...
+			}
+			else {
+				this._paneParent.scrollTo(0, 0);
+			}
+
+			// Only render visible panes
+			for (let box of pendingBoxes) {
+				if (pinnedIndex > -1 && panes.indexOf(box) < pinnedIndex) {
+					continue;
+				}
```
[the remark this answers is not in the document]
windingwind: Then the panes above are rendered as they are scrolled into view, then.

By `remove this optimization, and below as well`, do you mean we should always render all panes on the first load as we did before?

chrome/content/zotero/elements/itemDetails.js
```diff
@@ -0,0 +1,582 @@
...
+						box.render();
+					}
+				}
+			}
+
+			let pinnedPaneElem = this.getEnabledPane(this.pinnedPane);
+			let pinnedIndex = panes.indexOf(pinnedPaneElem);
+			
+			this._paneParent.style.paddingBottom = '';
+			await waitFrames(3);
```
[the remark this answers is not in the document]
windingwind: I guess 1 frame would be enough! We need to wait for panes to be hidden/visible here and I forget why I picked 3

chrome/content/zotero/elements/itemDetails.js
```diff
@@ -0,0 +1,582 @@
...
+			let paneElem = this.getEnabledPane(paneID);
+			if (!paneElem) return false;
+			const rect = paneElem.getBoundingClientRect();
+			// Sample per 20 px
+			const samplePeriod = 20;
+			let x = rect.left + rect.width / 2;
+			let yStart = rect.top;
+			let yEnd = rect.bottom;
+			let elAtPos;
+			// Check visibility from top/bottom to center
```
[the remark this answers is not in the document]
windingwind: That makes sense. I'll try

chrome/content/zotero/elements/itemDetails.js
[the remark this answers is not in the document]
windingwind: Well, just tested and 1 frame won't do - switching between attachment and regular item in the library with tags section pinned scrolls to the incorrect position. So at least for me wait for 3 frames works. I will add a comment.

chrome/content/zotero/elements/itemDetails.js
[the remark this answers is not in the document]
windingwind: Fixed i think

chrome/content/zotero/elements/itemDetails.js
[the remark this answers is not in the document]
windingwind: That's true but doesn't work either. I'll try to use the `hidden` as attribute instead of property in boxes to see if we can get this fixed. Probably XUL's `hidden` is the cause?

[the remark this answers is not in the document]
windingwind: Here are two different things - the waiting frame is to solve `Test case: pin tags section and switch between attachment and regular item`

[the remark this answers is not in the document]
windingwind: OK fixed in [40a56a8](https://github.com/zotero/zotero/pull/3624/commits/40a56a807176b40efa6703e9c0ecef2af9e24fd5). I forget the attachment preview has a self-hosted resize observer!

dstillman:
This seems to cause a couple extra icons to appear in the sidenav briefly at startup before some are removed:\

Screenshot 2024-02-19 at 12 16 56 AM

In the current beta, only the main item ones appear:

Screenshot 2024-02-19 at 12 18 38 AM

windingwind:
Oh, I see. That's a regression. working on it

windingwind:
> extra icons to appear in the sidenav briefly at startup before some are removed

Fixed

dstillman:
So this adds the Attachments section to the reader item pane, but it seems to be causing the attachment preview to be loaded for each open tab at Zotero startup (!). You can probably add more appropriate logging to see that, but I added

```
Zotero.debug('=-=-=-=');
Zotero.debug("URL IS " + uriPath);
```

to `zotero-protocol-handler.js::AttachmentExtension::newChannel`. If the attachment preview is enabled, that shows up for each tab. (Not sure if it's limited by `MAX_LOADED_TABS`.) And this actually happened even if the Attachments section was collapsed, which seems to point to a separate problem that predates this PR.

So, 1) we ideally wouldn't be loading anything in the context pane before a tab is loaded (not sure if we were before), but 2) certainly we shouldn't be loading the attachment preview, and 3) we shouldn't be loading attachment previews anywhere if the Attachments section is collapsed.

I'd also add 4) we shouldn't show the attachment preview in the Attachments section in the reader item pane if the primary PDF is already the current PDF in the reader. I guess it makes sense to show it if you're viewing a secondary attachment.

windingwind:
> So, 1) we ideally wouldn't be loading anything in the context pane before a tab is loaded (not sure if we were before), but 2) certainly we shouldn't be loading the attachment preview, and 3) we shouldn't be loading attachment previews anywhere if the Attachments section is collapsed. (I actually thought we fixed that before launch, but I can reproduce in the current beta.)
> I'd also add 4) we shouldn't show the attachment preview in the Attachments section in the reader item pane if the primary PDF is already the current PDF in the reader. I guess it makes sense to show the preview if you're viewing a secondary attachment.

Done.

chrome/content/zotero/elements/attachmentBox.js
```diff
@@ -33,8 +33,10 @@
 			<collapsible-section data-l10n-id="section-attachment-info" data-pane="attachment-info">
 				<html:div class="body">
 					<attachment-preview id="attachment-preview" tabindex="0"/>
-					<label id="url" crop="end" tabindex="0"
-						ondragstart="let dt = event.dataTransfer; dt.setData('text/x-moz-url', this.value); dt.setData('text/uri-list', this.value); dt.setData('text/plain', this.value);"/>
+					<html:div style="display: grid;">
```
dstillman, line 36: Why is this necessary?

chrome/content/zotero/elements/contextPane.js
```diff
@@ -55,7 +55,7 @@
 			if (!(viewType in viewTypeMap)) {
 				throw new Error(`ContextPane.viewType must be one of ["item", "notes"], but got ${viewType}`);
 			}
-			this._panesDeck.setAttribute("selectedIndex", viewTypeMap[viewType]);
+			this._panesDeck.selectedIndex = viewTypeMap[viewType];
```
dstillman, line 58: This sort of thing should be in a separate commit (with the other instance), prefixed with "fx115: ", so we can document it as a necessary change and link to it as an example.

chrome/content/zotero/elements/notesContext.js
```diff
@@ -277,8 +278,7 @@
 				vbox.style.display = 'flex';
 
 				editor = new (customElements.get('note-editor'));
-				editor.style.display = 'flex';
-				editor.style.width = '100%';
+				editor.setAttribute("flex", "1");
```
dstillman, line 281: Why is this necessary? Seems like we want to be moving away from the `flex` attribute.

chrome/content/zotero/zoteroPane.xhtml
```diff
@@ -1216,7 +1216,7 @@
 							>
 								<grippy/>
 							</splitter>
-							<context-pane id="zotero-context-pane-inner" flex="1" zotero-persist="height"></context-pane>
+							<context-pane id="zotero-context-pane-inner" flex="0" zotero-persist="height"></context-pane>
```
dstillman, line 1219: Can just remove if setting to 0, no?

chrome/content/zotero-platform/mac/overlay.css
```diff
@@ -6,15 +6,6 @@
...
 }
 
-.zotero-item-pane-content .groupbox-body {
-	-moz-appearance: none;
-	background-color: #ffffff;
-}
-
-#zotero-item-pane-message-box description {
-	color: #7f7f7f;
-}
```
dstillman, line 16: Why was this removed? I'm not sure why this was Mac-specific, but if anything it should be moved to a global style.

dstillman (changes requested):
The 115 changes look fine, but can you split them up into separate commits with commit messages that explain what's being fixed?

chrome/content/zotero/elements/attachmentBox.js
windingwind: The `lable` will overflow the parent element to the max length of text dramatically if inside non-grid layout parent element

chrome/content/zotero/elements/contextPane.js
windingwind: OK, I'll split them

chrome/content/zotero/elements/notesContext.js
windingwind: in 115 the `flex="1"` equals to `style="flex: 1"`. but if we have agreed to remove `flex`, then yes we can use the style.

chrome/content/zotero-platform/mac/overlay.css
windingwind: I didn't see this in the figma so currently this is replaced by the global style to `fill-tertiary`

chrome/content/zotero/elements/contextPane.js
windingwind: Just splitted, but haven't renamed to fx115 prefixed. Will rename commits later

chrome/content/zotero-platform/mac/overlay.css
windingwind: It's https://github.com/windingwind/zotero/blob/4dc5a2c3c280c77e15d04a24e79a5cb2bf111026/scss/elements/_itemDetails.scss#L3

chrome/content/zotero/xpcom/itemPaneManager.js
```diff
@@ -0,0 +1,358 @@
...
+ * @property {string} paneID - Registered pane id
+ * @property {Document} doc - Document of section
+ * @property {HTMLDivElement} body - Section body
+ * @typedef SectionUIHookArgs
+ * @type {object}
+ * @property {() => {item: Zotero.Item, mode: string, inTrash: boolean, tabType: string}} getData - Get section data
+ * @property {(l10nArgs: string) => void} setL10nArgs - Set l10n args for section header
+ * @property {(l10nArgs: string) => void} setEnabled - Set pane enabled state
+ * @property {(summary: string) => void} setSectionSummary - Set pane section summary
+ * @property {(buttonType: string, options: {disabled?: boolean; hidden?: boolean}) => void} setSectionButtonStatus - Set pane section button status
```
windingwind, line 51: Added new methods to hook callback to allow plugins to set summary or change section header button status.

chrome/content/zotero/xpcom/itemPaneManager.js
```diff
@@ -0,0 +1,358 @@
...
+	_customSections = {};
+
+	_lastUpdateTime = 0;
+
+	/**
+	 * Register a custom section in item pane. All registered sections must be valid, and must have a unique paneID.
+	 * @param {ItemDetailsSectionOptions | ItemDetailsSectionOptions[]} options - section data
+	 * @returns {string | string[] | false} - The paneID or false if no section were added
+	 */
+	registerSections(options) {
```
dstillman, line 140: Why should this be `registerSections` (plural)? The vast majority of the time, won't people be creating a single section? If we really thought people register multiple would be common, we could have a `registerSections` as well, but we should definitely have a `registerSection` for the common case.

chrome/content/zotero/zoteroPane.js
```diff
@@ -1346,7 +1347,7 @@ var ZoteroPane = new function()
 			}
 		}
 		
-		yield ZoteroItemPane.blurOpenField();
+		yield this.itemPane._itemDetails.blurOpenField();
```
dstillman, line 1350: We shouldn't access a private property from outside the file. The point of that convention is to know that any changes are limited to the file itself.

chrome/content/zotero/zoteroPane.js
```diff
@@ -1834,12 +1834,27 @@ var ZoteroPane = new function()
...
 			// Display buttons at top of item pane depending on context. This needs to run even if the
 			// selection hasn't changed, because the selected items might have been modified.
-			this.updateItemPaneButtons(selectedItems);
+			this.itemPane.data = selectedItems;
+			let viewMode = {
+			isFeedsOrFeed: collectionTreeRow.isFeedsOrFeed(),
+			isDuplicates: collectionTreeRow.isDuplicates(),
+			isPublications: collectionTreeRow.isPublications(),
+			isTrash: collectionTreeRow.isTrash(),
+			rowCount: this.itemsView.rowCount,
```
dstillman, line 1853: Indent these lines

chrome/content/zotero/zoteroPane.js
```diff
@@ -1834,12 +1834,27 @@ var ZoteroPane = new function()
...
+			};
+			this.itemPane.viewMode = viewMode;
```
dstillman, line 1855: `viewMode` isn't a good name for an object with properties, and it doesn't really seem like we should be reproducing these anyway. Seems like we should just be passing a reference to the `collectionTreeRow` and the `itemsView`, and the calling code can use the existing methods/properties on those interfaces.

`editable` below could *maybe* merit being set again, since that's more an intrinsic state of the pane.

chrome/content/zotero/elements/abstractBox.js
```diff
@@ -107,10 +99,9 @@
 			}
 		}
 
-		render() {
-			if (!this.item) {
-				return;
-			}
+		render(force = false) {
```
dstillman, line 102: When would `force` be necessary? If we need it, we should document somewhere what it does and why it exists.

chrome/content/zotero/elements/attachmentAnnotationsBox.js
```diff
@@ -24,49 +24,49 @@
...
 		`);
+
+		get tabType() {
+			return this._tabType;
+		}
+
+		set tabType(tabType) {
+			this._tabType = tabType;
+			this._updateHidden();
+		}
```
dstillman, line 42: The getter would just be called on `ItemPaneSectionElementBase` without this, no? And then if we need custom setters, seems like we should call `super(tabType)` and then do any extra work in these subclasses.

chrome/content/zotero/elements/contextPane.js
```diff
@@ -0,0 +1,335 @@
...
+		set viewType(viewType) {
+			let viewTypeMap = {
+				item: "0",
+				notes: "1",
+			};
+			if (!(viewType in viewTypeMap)) {
+				throw new Error(`ContextPane.viewType must be one of ["item", "notes"], but got ${viewType}`);
+			}
+			this._panesDeck.selectedIndex = viewTypeMap[viewType];
+		}
```
dstillman, line 59: If we get rid of `viewMode`, I think this can just be `mode` rather than `viewType`. It's really putting the pane into a different mode more than it's changing anything about the view. The deck is just an implementation detail.

dstillman (commented):
Still going through this, but some notes for now

chrome/content/zotero/xpcom/itemPaneManager.js
windingwind: I agree we don't need to support registering multiple sections! The behavior is inherited from item tree API, though I don't think it's necessary here. I'll update it

chrome/content/zotero/zoteroPane.js
windingwind: I try to prevent passing the `collectionTreeRow` or something that's not part of the item pane, to uncouple these modules. This leaves space for the possibility that we may want to use the item pane somewhere else in the future where there is no `collectionTreeRow`, I don't know.

My point of view here: Being able to directly access another module encourages people to bind them deeper during future maintenance (which is what happened to some parts of our code) and eventually things get mixed up.

windingwind: the `editable` is better than `viewMode`. I will update

chrome/content/zotero/elements/abstractBox.js
windingwind: The idea here is, basically the `render` is only called by the CE's container, which decides when it should be rendered, e.g. when scrolling into view. However, the container only knows when it should be reminded to render, but doesn't know if it's already rendered. So by default, the `force` is always `false`, and duplicate rendering calls are ignored, which is secured by `_isAlreadyRendered`. Say if you keep scrolling a section into and out of view, the `render` will be triggered every time but only the first one renders the section.

However, when it comes to notifier callbacks and other internal calls from this CE itself, that's where we want a force render to happen, no matter if it's already rendered.

I will update the implementation to make it more clear.

chrome/content/zotero/elements/attachmentAnnotationsBox.js
windingwind: make sense!

chrome/content/zotero/elements/abstractBox.js
windingwind: Anyway, seems like we don't have a better choice here. So let's pass them into itemPane.

chrome/content/zotero/zoteroPane.js
windingwind: Anyway, seems like we don't have a better choice here. So let's pass them into itemPane.

dstillman: Yeah, I don't think there's any reason not to pass the objects. We're already passing a whole bunch of their properties. If we used this somewhere where there was no `collectionTreeRow` or `itemsView`, we'd need to mock those properties anyway. We could define and document more limited interfaces with only the properties that are guaranteed to exist, but the ones we're already using are pretty arbitrary, and we (or a plugin) could easily want to use other properties of the `collectionRowTree` in the future.

chrome/content/zotero/elements/attachmentAnnotationsBox.js
windingwind: OK updated to use `super` as much as possible; but seems we cannot overwrite only the `setter` or `getter` and have to always inherit both.

windingwind:
Updated accordingly.

chrome/content/zotero/elements/attachmentAnnotationsBox.js
```diff
@@ -24,49 +24,53 @@
...
+
+		set tabType(tabType) {
+			super.tabType = tabType;
+			this._updateHidden();
+		}
 		
 		get item() {
-			return this._item;
+			return super.item;
```
dstillman, line 45: No need to use `super` for any of these getters, I don't think? `this._item` should work fine — it'll just use the `_item` from the prototype chain. `super.item = item` is only necessary in the setter so that it triggers the parent's setter as well. (And yes, sorry, I meant `super.tabType = tabType`, not `super(tabType)`.)

chrome/content/zotero/elements/itemDetails.js
```diff
@@ -0,0 +1,576 @@
...
+		_handlePaneStatus = (muts) => {
+			for (let mut of muts) {
+				let paneID = mut.target.dataset.pane;
+				if (paneID) {
+					this.sidenav.updatePaneStatus(paneID);
+				}
+			}
+		};
+
+		_handleIntersection = async (entries) => {
```
dstillman, line 535: I know this isn't new, but can we document what this is for?

chrome/content/zotero/elements/itemMessagePane.js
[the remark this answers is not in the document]
dstillman: You can test multiple windows on macOS with this hack:

https://forums.zotero.org/discussion/86670/feature-request-multiple-tabs

We should either test or add a comment saying it may not be necessary.

windingwind:
Updated accordingly.

chrome/content/zotero/xpcom/itemPaneManager.js
```diff
@@ -0,0 +1,335 @@
...
+ * @property {SectionL10n & SectionIcon} sidenav - Sidenav options. Icon should be 20*20 and `tooltiptext` need to be localized
+ * @property {string} [fragment] - Pane fragment as string
+ * @property {(options: SectionInitHookArgs) => void} [onInit]
+ * Lifecycle hook called when section is initialized
+ *
+ * Do:
+ * 1. Initialize data if necessary
+ * 2. Setup hooks, e.g. notifier callback
+ *
+ * NOT do:
```
dstillman, line 74: → "Don't"

chrome/content/zotero/xpcom/itemPaneManager.js
```diff
@@ -0,0 +1,335 @@
...
+ * @property {string} pluginID - Set plugin ID to auto remove section when plugin is disabled/removed
+ * @property {SectionL10n & SectionIcon} head - Header options. Icon should be 16*16 and `label` need to be localized
+ * @property {SectionL10n & SectionIcon} sidenav - Sidenav options. Icon should be 20*20 and `tooltiptext` need to be localized
+ * @property {string} [fragment] - Pane fragment as string
+ * @property {(options: SectionInitHookArgs) => void} [onInit]
+ * Lifecycle hook called when section is initialized
+ *
+ * Do:
+ * 1. Initialize data if necessary
+ * 2. Setup hooks, e.g. notifier callback
```
dstillman, line 72: → "Set up"

chrome/content/zotero/xpcom/itemPaneManager.js
```diff
@@ -0,0 +1,335 @@
...
+ * Return `false` to abort the data change, otherwise the data change will be applied
+ * (stored internally and can be accessed by `options.getData` in other hooks).
+ *
+ * Do:
+ * 1. Validate and decide whether to accept the incoming data change
+ * 2. Update the section enabled state with `options.setEnabled`. For example, if the section
+ *   is only enabled in the readers, you can use:
+ * ```js
+ * onDataChange(options) {
+ *   let { incomingData, setEnabled } = options.;
```
dstillman, line 97: - How about `newData` instead of `incomingData`?
- Remove trailing `.`

chrome/content/zotero/xpcom/itemPaneManager.js
```diff
@@ -0,0 +1,335 @@
...
+ * NOT do:
+ * 1. Render/refresh UI
+ * @property {(options: SectionHookArgs) => void | Promise<void>} onRender
+ * Lifecycle hook called when section should do primary render
+ *
+ * Create elements and append them to `options.body`.
+ * - If `onRender` is sync, it is always called when the item pane is rendered;
+ * - If `onRender` is async, it will be called when scrolling into view.
+ *
+ * If the rendering can be slow, you should either make it async, or move async part to `onSecondaryRender`.
```
dstillman, line 115: Is there a difference? How would you choose one over the other?

chrome/content/zotero/xpcom/itemPaneManager.js
```diff
@@ -0,0 +1,335 @@
...
+ * @property {(l10nArgs: string) => void} setEnabled - Set pane enabled state
+ * @property {(summary: string) => void} setSectionSummary - Set pane section summary
+ * @property {(buttonType: string, options: {disabled?: boolean; hidden?: boolean}) => void} setSectionButtonStatus - Set pane section button status
+ * @typedef SectionHookArgs
+ * @type {SectionBasicHookArgs & SectionUIHookArgs}
+ * @typedef  {SectionHookArgs & { refresh: () => Promise<void> }} SectionInitHookArgs
+ * A `refresh` is exposed to plugins to allows plugins to refresh the section when necessary,
+ * e.g. item modify notifier callback. Note that calling `refresh` during initialization
+ * have no effect.
+ * @typedef  {SectionHookArgs & { incomingData: { type: "item" | "mode" | "inTrash" | "tabType", value: any } }} SectionDataChangeHookArgs
```
dstillman, line 58: This is a pretty weird definition — it took me a while even to figure what it meant.

Do these things change independently? If the `tabType` changes, isn't the item probably also going to change? It seems like, if these are actually independent, they should be separate events (e.g., `onItemChange`), but if not, the handler should just get the new complete state of the pane, with all the necessary properties.

Pulling out `inTrash` as a separate state also seems too specific to me. In line with our other change of passing in `collectionTreeRow`, shouldn't the pane just get notified when there's a new `collectionTreeRow`, and it can check if it's a trash or duplicate or whatever else?

Also, what are the possible values of `mode`? Is that just `edit`/`view`? Something else? If just `edit`/`view`, that might be implied by `collectionTreeRow.editable`?

Seems like this could potentially just always include `tabType`, `collectionTreeRow`, and `item` as properties?

dstillman: Or I guess there isn't always a `collectionTreeRow`, like in the reader, but then maybe `collectionTreeRow` (which could be null) and `editable`? Not that we're implementing things with this API at this time (I think?), but wasn't the point of that other change that we're checking various properties on the tree row, and so plugins sections might have reasons to do the same?

chrome/content/zotero/xpcom/itemPaneManager.js
dstillman: Oh, I see a case for secondary render is given below, but I think it's still a little unclear what we expect to be done in synchronous onRender, async onRender, synchronous onSecondaryRender, and async onSecondaryRender.

chrome/content/zotero/xpcom/itemPaneManager.js
```diff
@@ -0,0 +1,335 @@
...
+ *     setEnabled(incomingData.value === "reader");
+ *   }
+ *   return true;
+ * }
+ * ```
+ * options.setEnabled(options.tabType === "reader");
+ *
+ * NOT do:
+ * 1. Render/refresh UI
+ * @property {(options: SectionHookArgs) => void | Promise<void>} onRender
```
dstillman, line 108: `options` seems a little wrong for these handlers. This is mostly a documentation issue, since people can name the argument whatever they want (or just use destructing assignment), but while `options` is totally appropriate for passing things like ids and strings and event handlers when registering a section, here things are just getting passed to the handlers regardless, so they're not options. Maybe just `props`, and encourage destructuring assignment in the documentation?

chrome/content/zotero/xpcom/itemPaneManager.js
windingwind: I think this illustrates why/how different `render` works:

image

To respond to the user's selecting item as soon as possible, we want the initial rendering to be very fast, thus not allowing any async rendering to happen there.

The differences:
- sync `onRender` is always executed during initial render;
- async `onRender` and any `onSecondaryRender ` are executed when scrolling into view (on need).

chrome/content/zotero/xpcom/itemPaneManager.js
windingwind: > Do these things change independently?

Yes. The data are set one by one by the `ItemDetails`.

> Seems like this could potentially just always include tabType, collectionTreeRow, and item as properties?

It's like *a certain property is to be updated. do you want to do something now?*

Passing the changed variable implicitly indicates what is changed. Like, we probably want to update something when `item` is changed, but other properties are not something we care about.

>  but wasn't the point of that other change that we're checking various properties on collectionTreeRow, and so plugin sections might have reasons to do the same?

Well, previously in these `box` things we do not access collectionTreeRow and use these `inTrash` or something that's set by the container.

What I have in mind with these properties:
1. Replace `mode` with `editable`
2. Remove `inTrash`, which is only used by our `attachmentsBox`. We can simply check if the `item` is in trash, right? So no need to rely on the collection tree.
3. Exclude `tabType` from this `onDataChange`, since it's never set a different value.

Do you think these changes make sense?

chrome/content/zotero/xpcom/itemPaneManager.js
dstillman: So you might use an async `onRender` if you only had a single render to do but it was going to be slow? Couldn't you just have a no-op (or maybe a height calculation) for `onRender` in that case and use `onSecondaryRender`? It seems a little confusing to have two different ways of rendering stuff at the same time. What if we just had sync `onRender` and `onSecondaryRender` (or even called them `onRender` and `onAsyncRender`)?

Alternatively, are we sure we want to allow plugins to do sync rendering at all? If these will often be below the fold, why provide a mechanism for rendering ahead of time at all and burning cycles on every item select? What if we only allowed a way to return a height, and then an async render method?

chrome/content/zotero/xpcom/itemPaneManager.js
dstillman: So then there would just be `item` and `editable`? That sounds good. But we have a counterexample in Libraries & Collections, which needs to know the currently selected library and collection, right? How is that getting notified of the selection change?

chrome/content/zotero/xpcom/itemPaneManager.js
windingwind: > What if we just had sync onRender and onSecondaryRender (or even called them onRender and onAsyncRender

Good point. That makes more sense!

> What if we only allowed a way to return a height, and then an async render method?

Might it be awkward to ask plugins to compute a height, since we have dynamic font size and compact/comfortable, which might contribute to computing the height? It might just be not practical for plugins to compute an exact height of things.

I think in most cases the `onRender` burden would not be very heavy - might be just loading some data and setting attributes to several elements. We could add a timer and throw a warning if it's slow to indicate the plugin is doing something bad here.

dstillman: > Might it be awkward to ask plugins to compute a height, since we have dynamic font size and compact/comfortable, which might contribute to computing the height? It might just be not practical for plugins to compute an exact height of things.

Well, I was thinking it could be optional, but I guess the point is that we do want heights as much as possible to avoid weird scroll behavior, and in the default, non-slow case, computing a height (in canvas? I'm not even sure how) could actually result in as much work, which might then just need to be repeated for an actual render. So then by not allowing for a sync render we'd just making the default case worse…

chrome/content/zotero/xpcom/itemPaneManager.js
windingwind: The `Libraries & Collections` currently use `ZoteroPane.collectionsView`, which is OK I think?

Actually, maybe we should skip the `isCurrent` check (make the collection row unclickable if it is the currently selected collection) in non-library (readers), but it's a separate issue.

chrome/content/zotero/xpcom/itemPaneManager.js
windingwind: Yeah. So let's try the `onRender` and `onAsyncRender`? Ideally, it would be clear enough.

dstillman: Sounds good.

chrome/content/zotero/xpcom/itemPaneManager.js
windingwind: Oh wait... I think `mode` can be `edit`/`view`/`merge`/`mergeedit`. So we actually cannot replace `mode` with `editable`

dstillman: > The Libraries & Collections currently use ZoteroPane.collectionsView, which is OK I think?

Oh right, I forgot that any time the user is switching between collections the item pane is getting re-rendered anyway. So sure, plugins can always check ZoteroPane.collectionsView or anything else in the window, and we'd only need to pass things that could change without an item re-selection.

> I think mode can be edit/view/merge/mergeedit. So we actually cannot replace mode with editable

Well, `merge` and `mergeedit` are only for a few of our own panes in the conflict-resolution dialog. I'm not sure it makes sense to have a more complicated API for plugins just because of that…

dstillman: If we did only have `item` and `editable`, would it still make sense for those to be passed separately? I'm not sure it's even possible for `editable` to change without an item reselection. So then if it's only `item` changing, or an initial `item` being set, I feel like we could just pass these together.

windingwind: > I'm not sure it's even possible for editable to change without an item reselection.

No, I think. I only know the item `editable` changes when it is trashed. But anyway, plugins can listen to item modify notifier, and thus we don't need this `editable` callback. So, let's provide an `onItemChange`.

dstillman: > I only know the item `editable` changes when it is trashed.

No, that's not right — `item.isEditable()` will return `true` for an item in the trash. The item's editable state isn't the same as the current view's editable state. So I think we do need to provide `editable` one way or another — otherwise many plugins would need to check `collectionTreeRow.editable`.

I was just suggesting that we provide both `item` and `editable` in the same event, since `editable` won't be changing on its own.

windingwind: OK, get it.

windingwind: In that case, maybe `editable` can be from `props.getData()`, instead of together with `item` in a `onDataChange`/`onItemChange` callback?

When the item changes/when rendering, plugins can get the `editable` value and do stuff accordingly.

dstillman: Is there a point to `onItemChange`? Can `onRender` be called multiple times on the same item? Or should plugins just check the item when rendering?

windingwind: > Can onRender be called multiple times on the same item

We have a mechanism to avoid duplicate `onRender` calls - only the first one works.

>  Is there a point to onItemChange? Or should plugins just check the item when rendering?

It's mainly for calling `setEnabled` before rendering, to indicate the section is not enabled and hide the corresponding icon in sidenav. For example, when the item is an attachment, the `info` section is not enabled. It would be too late to do this when it's about to render

dstillman: Ah, OK. Makes sense.

So then just getting `editable` when rendering makes sense. But while we're on the subject, is there a reason for having `getData()`, as opposed to just including those things in the passed properties?

windingwind: > as opposed to just including those things in the passed properties?

I intend to indicate they are read-only properties (don't try to change them), and also dynamic properties (don't try to persist them). Do you think the reason is strong enough or there's a better solution?

dstillman: Well, read-only in what way? `item` is the item — there's nothing really read-only about it. And I think passing them in every event handler makes clear that you should check them each time and not persist them.

But in any case, I don't think `getData()` particularly communicates those things. I'd just include them as properties.

windingwind: > read-only in what way?

Like don't try to `setData` to overwrite `editable` or `item`?

> But in any case, I don't think getData() particularly communicates those things. I'd just include them as properties.

OK, then let's do it as properties.

dstillman: Oh, well we can definitely pass a new object to each handler (`Object.assign({}, obj)`) to make sure one isn't doing something funky.

dstillman: Or I guess get fancy with freezing or whatever

dstillman: Just calling `Object.freeze()` on the props object before passing to the handlers is probably sufficient?

windingwind:
Updated accordingly. The example (https://gist.github.com/windingwind/afdf64046d887f5b6750b225ae2fc822) is also updated.

chrome/content/zotero/xpcom/itemPaneManager.js
```diff
@@ -55,7 +57,7 @@
  * A `refresh` is exposed to plugins to allows plugins to refresh the section when necessary,
  * e.g. item modify notifier callback. Note that calling `refresh` during initialization
  * have no effect.
- * @typedef  {SectionHookArgs & { newData: { type: "item" | "mode" | "inTrash" | "tabType", value: any } }} SectionDataChangeHookArgs
+ * @typedef  {SectionHookArgs & { newData: Zotero.Item }} SectionDataChangeHookArgs
```
dstillman, line 60: This shouldn't be `newData` if we're just passing an item now. Why wouldn't `onItemChange` just pass the new item as `item`?

dstillman: I see we're passing the old item as `item`:

> Zotero.debug(`Section item data changed from ${item?.id} to ${newData?.id}`);

I'm not sure if that's necessary, but if so, I feel like it should be `oldItem` and `item` or something. But also, the old item might not even exist anymore — it could've been deleted. Do we really need to pass it?

dstillman:
```javascript
onRender: (props) => {
    let { body, item, setL10nArgs, setSectionSummary, setSectionButtonStatus } = props;
```

Also, while it's not in this code, for the examples we can just use destructuring assignment right in the parameters line:

```javascript
onRender: ({ body, item, setL10nArgs, setSectionSummary, setSectionButtonStatus }) => {
```

No need to define `props` at all.

chrome/content/zotero/xpcom/itemPaneManager.js
```diff
@@ -0,0 +1,340 @@
...
+ * have no effect.
+ * @typedef  {SectionHookArgs & { newData: Zotero.Item }} SectionDataChangeHookArgs
+ * @typedef  {SectionHookArgs & { event: Event }} SectionEventHookArgs
+ * @typedef ItemDetailsSectionOptions
+ * @type {object}
+ * @property {string} paneID - Unique pane ID
+ * @property {string} pluginID - Set plugin ID to auto remove section when plugin is disabled/removed
+ * @property {SectionL10n & SectionIcon} head - Header options. Icon should be 16*16 and `label` need to be localized
+ * @property {SectionL10n & SectionIcon} sidenav - Sidenav options. Icon should be 20*20 and `tooltiptext` need to be localized
+ * @property {string} [fragment] - Pane fragment as string
```
dstillman, line 68: Can you clarify what this means?

chrome/content/zotero/xpcom/itemPaneManager.js
```diff
@@ -0,0 +1,340 @@
...
+ *
+ * Return `false` to abort the data change, otherwise the data change will be applied
+ * (stored internally and can be accessed by `props.item` in other hooks).
+ *
+ * Do:
+ * 1. Validate and decide whether to accept the ne data change
+ * 2. Update the section enabled state with `props.setEnabled`. For example, if the section
+ *   is only enabled in the readers, you can use:
+ * ```js
+ * onDataChange({ newData, setEnabled }) {
```
dstillman, line 105: Needs to be updated

chrome/content/zotero/xpcom/itemPaneManager.js
```diff
@@ -0,0 +1,340 @@
...
+ *   if (newData.type === "tabType") {
+ *     setEnabled(newData.value === "reader");
+ *   }
+ *   return true;
```
dstillman, line 109: The example document says "return false to cancel the change". What does that mean? Cancel what change?

dstillman: Oh, and above too:

> Return `false` to abort the data change, otherwise the data change will be applied
> (stored internally and can be accessed by `props.item` in other hooks).

I don't understand what this means. Isn't this just notifying the section that we've switched to a new item? How can that be aborted?

chrome/content/zotero/xpcom/itemPaneManager.js
```diff
@@ -0,0 +1,340 @@
...
+ * @property {HTMLDivElement} body - Section body
+ * @typedef SectionUIHookArgs
+ * @type {object}
+ * @property {Zotero.Item} item - Current item
+ * @property {string} tabType - Current tab type
+ * @property {boolean} editable - Whether the section is in edit mode
+ * @property {(l10nArgs: string) => void} setL10nArgs - Set l10n args for section header
+ * @property {(l10nArgs: string) => void} setEnabled - Set pane enabled state
+ * @property {(summary: string) => void} setSectionSummary - Set pane section summary
+ * @property {(buttonType: string, options: {disabled?: boolean; hidden?: boolean}) => void} setSectionButtonStatus - Set pane section button status
```
dstillman, line 53: What is `buttonType`?

chrome/content/zotero/xpcom/itemPaneManager.js
```diff
@@ -0,0 +1,340 @@
...
+ * @property {Document} doc - Document of section
+ * @property {HTMLDivElement} body - Section body
+ * @typedef SectionUIHookArgs
+ * @type {object}
+ * @property {Zotero.Item} item - Current item
+ * @property {string} tabType - Current tab type
+ * @property {boolean} editable - Whether the section is in edit mode
+ * @property {(l10nArgs: string) => void} setL10nArgs - Set l10n args for section header
+ * @property {(l10nArgs: string) => void} setEnabled - Set pane enabled state
+ * @property {(summary: string) => void} setSectionSummary - Set pane section summary
```
dstillman, line 52: What is a section summary?

chrome/content/zotero/xpcom/itemPaneManager.js
windingwind: It's the same as 
https://github.com/windingwind/zotero/blob/dad2adeb87ead19ffac363f263f1a2ee41c7e445/chrome/content/zotero/xpcom/itemPaneManager.js#L38

Clicking this button will dispatch an Event of the same name (e.g. `add` is what we use in our sections).

So if you define several different buttons, you can set the status by its `buttonType`.

chrome/content/zotero/xpcom/itemPaneManager.js
windingwind: It's the text we show on the section header when the section is collapsed. We use that in the Abstract section.

chrome/content/zotero/xpcom/itemPaneManager.js
windingwind: Oh, yes. From the very beginning, I implemented this part following the way we do in our sections. For example, in the attachment box, the `item` is only set when it's an attachment, otherwise the `this._item` is not updated. So same here, it can be aborted, which means if you return false the `this._item` will not be updated.

Looking back, I think it might not be necessary to be able to `abort` the setter, and actually, maybe we should update our sections' implementation instead. It could be confusing.

chrome/content/zotero/xpcom/itemPaneManager.js
windingwind: In case plugin developers don't want to manually `createElement` everything, they can pass a string of HTML elements. (or, do you think it's better to be a XHTML file URL?)

It's like what we allow plugins to do in the pref pane - to write a `.xhtml` and load the content from it.

dstillman: Maybe @AbeJellinek has an opinion on this. If it were a file URL, I guess we'd have to cache it ourselves so it wasn't loaded every time, unlike in preferences where loading it from the file every time is fine.

Should clarify the variable name and comment, though, and we need to define whether this is HTML or XUL by default. We're calling this the "body" elsewhere. Should this be `bodyHTML`?

chrome/content/zotero/xpcom/itemPaneManager.js
dstillman: Yeah that seems very confusing. (I actually still don't understand it.)

chrome/content/zotero/xpcom/itemPaneManager.js
windingwind: done

chrome/content/zotero/xpcom/itemPaneManager.js
windingwind: done

chrome/content/zotero/xpcom/itemPaneManager.js
windingwind: done

chrome/content/zotero/xpcom/itemPaneManager.js
dstillman: How do you define buttons?

chrome/content/zotero/xpcom/itemPaneManager.js
dstillman: OK, let's clarify that in the comment

chrome/content/zotero/xpcom/itemPaneManager.js
windingwind: > If it were a file URL, I guess we'd have to cache it ourselves so it wasn't loaded every time

Caching would introduce a lot of problems. Like when the plugin got updated, or reloaded during plugin development (where the file is changed but the plugin version is not). I would rather not face that.

`bodyHTML` instead of `fragment`? Sounds good. I think default to XUL?

chrome/content/zotero/xpcom/itemPaneManager.js
windingwind:
```js
sectionButtons: [
		{
			type: "test",
			icon: "chrome://zotero/skin/16/universal/book.svg",
			onClick: ({ item, paneID }) => {
				Zotero.debug("Section clicked!", item?.id);
				Zotero.ItemPaneManager.unregisterSection(paneID);
			},
		},
	],
```

chrome/content/zotero/xpcom/itemPaneManager.js
dstillman: Yeah, `body` defaulting to XUL sounds good.

windingwind: done

chrome/content/zotero/xpcom/itemPaneManager.js
windingwind: done

dstillman:
I think just `body` is fine. `bodyXHTML` is too finicky.

windingwind:
> I think just body is fine

Emm, since we have `props.body` which is an HTMLElement, would it be somehow confusing to also use `body` here?

abaevbog:
I have not been very closely following this but I wanted to point out a couple of issues related to `tags-box` that I just ran into that I did not see in `fx115`. If these were already covered or if it's not the right time for this, I apologize.

1. The `+` icon does not appear in the header of the collapsible section of the tagsbox for me. As I see, the `mode="edit"` attribute is never set on the `tags-box` in itemDetails.js.

Screenshot 2024-03-26 at 2 22 14 PM

2. Likely related to that, the tags popup from the annotation sidebar in the reader tab is not editable and does not include the tags from the annotation. (E.g. this annotation has 2 tags).
Screenshot 2024-03-26 at 2 23 26 PM

scss/scaffold.scss
```diff
@@ -129,6 +129,7 @@ vbox > splitter {
 
 #left-tabbox {
 	margin: 5px;
+	min-width: 500px;
```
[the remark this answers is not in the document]
windingwind: Oh, for me the 500px is approximately the content elements' non-overflow width on Windows. We can remove that, but without a minimal width, the left pane can be resized to smaller than its content...

chrome/content/scaffold/scaffold.xhtml
```diff
@@ -159,7 +159,8 @@
 			<image class="titlebar-icon" src="chrome://zotero/skin/z.svg"></image>
 		</hbox>
 		<!-- Menu -->
-		<toolbar type="menubar" id="toolbar-menubar" class="chromeclass-menubar" customizable="true"
+		<!-- TEMP: Should be <toolbar>, but only HTML -moz-window-dragging: drag elements have rclick menu on Windows -->
```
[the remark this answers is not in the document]
windingwind: Well, I added several (many) fx115 commits to this PR, to make the new API work on 115. So this is an fx115 fix, not part of the item pane things.

I don't know why the `toolbar` no longer triggers the window menu, but seems we don't have another choice here. On windows & some linux the r-click to open the window menu is something all apps do.

scss/scaffold.scss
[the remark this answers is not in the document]
windingwind: To prevent this: 
image

> the checkboxes have to wrap

That's another issue of the checkbox's container, the fx115 layout issue that we are already familiar with. Will fix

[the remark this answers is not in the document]
windingwind: Pushed a fix

scss/scaffold.scss
```diff
@@ -144,6 +145,7 @@ vbox > splitter {
 	margin: -16px -16px -16px 0px;
 	border-left: var(--material-panedivider);
 	textarea {
+		flex: 1;
```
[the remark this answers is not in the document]
windingwind: fixed

scss/elements/_collapsibleSection.scss
```diff
@@ -61,17 +61,9 @@ collapsible-section {
...
-				background: var(--fill-quinary);
-			}
-			
-			&:active {
-				background: var(--fill-quarternary);
-			}
 		}
 
-		toolbarbutton.twisty .toolbarbutton-icon {
+		toolbarbutton.twisty {
```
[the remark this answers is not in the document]
windingwind: fixed

scss/elements/_collapsibleSection.scss
```diff
@@ -84,7 +76,7 @@ collapsible-section {
 	}
 	
 	&[open]:not([empty]) > .head {
-		toolbarbutton.twisty .toolbarbutton-icon {
+		toolbarbutton.twisty {
```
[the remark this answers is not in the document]
windingwind: fixed

windingwind:
@abaevbog 

Thanks for the bug report! We decided to replace `mode` with `editable` in these sections (except for the attachment box and item box), so the `mode` attribute elsewhere should now be `readonly` attribute.

I've pushed a fix.

scss/scaffold.scss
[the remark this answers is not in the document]
windingwind: This PR was initially based on the current main, but since we want to merge fx115 ASAP, we decided to rebase this PR to fx115, which brings tons of fx115 fixes to make it work on fx115. Some parts might not be very related to the original PR.

The scaffold commits can be definitely picked into another separate PR if we want. Anyway we will need to go through these of commits and squash or rename them.

dstillman:
OK, let's squash this into reasonable commits and we'll get this in. I'd like to push out a `dev`-channel build with the new API for developers, and we can continue to tweak things after.

windingwind:
Glad to know we finally (almost) get there! It has been a very long discussion and thanks to everyone who was involved.

I squashed the major commits into 3. Many `fx115` commits are not touched, let me know if we would want to squash them into fewer commits.

PR #3630, opened 2024-01-29
Implement better editable-text focus and context menu behavior, fix regressions

### LINKED ISSUE #3588: Don't blur `editable-text` when window blurs?
While writing #3587, I noted that this is the current behavior - entering some text into an `editable-text` and switching to another window saves the field.

Is that actually what we want? Maybe not? I think we can detect whether the field itself is being blurred or the window as a whole; if it's the window, maybe we keep the field active when you switch back and don't save in the mean time?

Might fix the tags-box test, even!

### LINKED ISSUE #3619: Don't switch field to editing mode on right-click
It's wrong and also causes the editing options to appear in the context menu, pushing the custom actions down:

https://forums.zotero.org/discussion/111221/zotero-7-beta-r-click-dropdown-title-sentence-case

Screenshot 2024-01-29 at 12 28 05 AM

Screenshot 2024-01-29 at 12 28 22 AM

AbeJellinek:
- When right-clicking an unfocused field, only show Copy and Paste (as well as other actions, like View Online, Title Case, etc.) instead of a full text editing menu
	- Put Copy as URL below Copy
- Fix a redesign regression that broke Copy as URL and View Online for DOIs
- Don't trigger a save when an `editable-text` blurs because the window is becoming inactive
- Ignore base-mapped fields using `Zotero.ItemFields`

Fixes #3619, fixes #3588

### DIFF
--- a/chrome/content/zotero/editMenuOverlay.js
+++ b/chrome/content/zotero/editMenuOverlay.js
@@ -101,16 +101,16 @@ function goBuildEditContextMenu() {
             MozXULElement.parseXULToFragment(`
               <popupset>
                 <menupopup id="textbox-contextmenu" class="textbox-contextmenu">
-                  <menuitem data-l10n-id="text-action-undo" command="cmd_undo"></menuitem>
-                  <menuitem data-l10n-id="text-action-redo" command="cmd_redo"></menuitem>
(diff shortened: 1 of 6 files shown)


dstillman:
Merged

Issue #3625, opened 2024-01-29
Wrong tooltip for collection filter button


dstillman:
![Screenshot 2024-01-29 at 5 20 45 AM](https://github.com/zotero/zotero/assets/783082/fd428eba-3fa1-41a9-b725-2697339b3427)

PR #3632, opened 2024-01-29
collection search tooltiptext and placeholder cleanup

abaevbog:
- Proper fluent tooltiptext for colleciton search button
- Removed hardcoded placeholder from collection search input field
- Set placeholder dynamically to be the same as the tooltiptext of the button because data-l10n-id.placeholder does not seem to work on `<search-text>`

Fixes: #3625


abaevbog:
Oh yeah, it works! Thanks for telling me about `data-l10n-attrs` - I've had to use the JS trick a few times in the past. Good to know the right way to do it

dstillman:
We don't want to duplicate localized strings, so in a case like this we can add the string as a term and then reference that in both DOM-specific ones.

Issue #3633, opened 2024-01-30
editable-text vertical arrows on windows


abaevbog:
https://forums.zotero.org/discussion/111305/zotero-7-beta-strange-up-and-down-arrows-on-title-extra-and-abstract

@AbeJellinek, that's another aspect of the `editable-text` scrolling issue you had addressed in #3589.

The root of this goes back to the [overflow-y: scroll](https://github.com/zotero/zotero/blob/b3cf2ac67ead0296a09ff2e8bbb2fcdac9675301/scss/elements/_editableText.scss#L121) that was needed to match `::after` with the `textarea` on macOS. It's still required on windows, since text doesn't line up without it. But it to always adds arrows, even when the field is empty.

I'll start poking around to see if I run into something helpful

PR #3634, opened 2024-01-30
remove arrows from editable-text on windows

abaevbog:
Somehow `scrollbar-width: thin` seems to remove unnecessary arrows from editable-text textarea on windows with the texts still matching up...

@AbeJellinek, is this working for you?

Addresses: #3633


abaevbog:
Great! For reference, this is what resizing looks like for me (finally figured out how to screen record on my old windows laptop): https://www.dropbox.com/scl/fi/t5hblgkuft1z1rb3bmq5q/another-collection-Zotero-1_30_2024-10_48_18-AM.mp4?rlkey=umv36yg47fvgen36rvs9lyhqu&dl=0

The `::after` is red, and the white text from `textarea` is never visible, so things seem to be matching well

dstillman:
We want to do this for the main scrollbar too, no? There's an arrow showing there as well (in the top-right corner).

abaevbog:
Oh yeah, I didn't realize that it's not just `editable-text` but the `itemPane` as well... The rogue arrows do show up when there is no overflow and the actual scrollbar is hidden. I just re-pushed that commit with `scrollbar-width: thin` on `itemPane`, and it seems to work.

abaevbog:
What's interesting is that it doesn't seem to happen with other scrollable areas. For example, `collectionTree` has a default scrollbar and the random arrows don't show up. 

Is it now strange looking that the `itemPane` does not have the same scrollbar as `collectionTree`, `itemTree` and etc.? We could set `scrollbar-width: thin` globally on windows so that it all looks the same. The flip side is that the scrollbars are, apparently, very janky, so maybe it's better to not touch components that are not broken

dstillman:
Yeah, it's not ideal having different styles, and it does seem like this should be fixable, since we don't see it elsewhere, but also probably not worth spending a lot of time on unless we start getting a lot of complaints from Windows 10 users…

PR #3645, opened 2024-01-30
More attachment type icons in the item tree

tnajdek:
Summary of changes in this PR:

* Add new icons for image and video attachment types (both in attachment column and the item type column)
* Fix few existing icons off by a tiny margin between different color modes (especially visible when selecting items)
* Improve sorting by attachment type column


dstillman:
I'm actually not sure we should show non-PDF/EPUB/snapshot attachment icons for the item-type icon, or if we do, I don't think these are the right icons. It's not at all clear here that a standalone image is a standalone attachment — the icon is just *too* fancy and similar to Artwork:

Screenshot 2024-02-02 at 1 57 20 AM

I think we'd have to either stick to the blank page or do something like the blank page with something overlaid, more similar to PDF/EPUB/snapshot.

The image icon is great in the Attachments column, just not for item type. I don't know if we're OK being inconsistent with PDF/EPUB/snapshot and showing a page + overlay for item type and just the overlay icon in Attachments. Might be fine?

The video icon is a little drab. As an item-type icon, it really sticks out (but in a distracting way, not a this-is-just-a-video-file way). It's OK if only rendered small in Attachments or an overlay, though I think we can still do better — if we're going for clapboard, I think it could be more horizontal, and could also play up the clapboard-ness more.

tnajdek:
Perhaps, for item type icon, we could stick with a shape of a document, something like this (or even more minimalistic version of this):

Screenshot 2024-02-02 at 19 22 36

Or we can just revert to generic document for image and video. @yexingsha do you have thoughts on this?

tnajdek:
Updated the PR, in my opinion these icons work very well!

Screenshot 2024-02-06 at 11 53 05

dstillman:
That's using the document shape for all, though, right? It looks like there are non-document ones in the ZIP. Did we want to use those for the Attachments column?

dstillman:
Well, I said this above:

> The image icon is great in the Attachments column, just not for item type. I don't know if we're OK being inconsistent with PDF/EPUB/snapshot and showing a page + overlay for item type and just the overlay icon in Attachments. Might be fine?

Any thoughts on this? It's just a little hard to see the image and video parts within the document shape:



The logic here would be that it's just conveying the content type, so there's no need for the actual item type icon.

tnajdek:
It works for me either way. If we go this route, maybe we could make "play" in the video icon blue-ish to match the same element on item-type icon?)
Screenshot 2024-02-19 at 14 14 54

yexingsha:
Updated some icons so that the 12px image and video icons no longer have the document shape, and the elements within are larger.
[updated.zip](https://github.com/zotero/zotero/files/14337291/updated.zip)

tnajdek:
Updated:
Screenshot 2024-02-20 at 11 03 16

Screenshot 2024-02-20 at 11 01 39

Issue #3646, opened 2024-01-31
Remove max height from editable-text in item pane


dstillman:
https://forums.zotero.org/discussion/comment/453833#Comment_453833

I know a lot of work went into getting this working right, but there shouldn't actually be a max height on anything in the item pane. We always showed the full title and expanded extract before, so it's a usability regression for no real reason, and the nested scrollbars look bad.

dstillman:
Abstract is already a collapsible section, so that seems fine?

dstillman:
I guess the argument is that there's value in seeing a few lines? I'm not sure if that's actually the case.

dstillman:
What did we even show before? One truncated line?

abaevbog:
Yes, and it would expand all the way when you click on it

abaevbog:
In the screenshot from the forums, I totally see that there's no reason to have both collapsible section and the scrollbar at the same time. You would expect one or the other, right?

dstillman:
Yeah, I'm just not sure that was providing much value beyond knowing that there was an abstract. I suppose we could try to do something custom in the section header if we thought it was useful, but that can happen separately.

dstillman:
Oh, I was thinking the Abstract header already had that, but I see that it doesn't. That's easy to fix, though.

dstillman:
(And yes, more thoughts for Title changes incoming…)

abaevbog:
Something that I saw in a random excel-like app is a long text field shows a "preview" of a multiline field (like 4-5 lines or so), and if you want to see the whole thing, you would click a small icon at the bottom right of the textfield, which opens a large popup that displays all the text. Using this as an inspiration, what if we add a tiny "expand" icon to the bottom right of title, abstract, etc., and when you tap on it, it will remove the height limit? It would allow one to see the entire field if needed, and would not require collapsing anything, so the "preview" of the abstract doesn't actually go away.

dstillman:
We already have a mechanism for expanding long editable things on click for annotations in the reader sidebar if we wanted it, but I really don't think this is a problem. Scrolling is OK! We can trivially add an indication to the Abstract header — even the same truncated first line — to show that there's an abstract, which brings us back to exactly what we had before, except much more discoverable. Title we can discuss separately, but along with moving it (more on that later), I'd like to move Locate so that we get the full width, which will hopefully get us not so far off from the height we were taking up at half pane (i.e., with the Title label) in a smaller font before. So I think we can just do this and deal with those things separately.

[...]

PR #3649, opened 2024-01-31
editable-text has no max-height unless specified

abaevbog:
- min/max-lines attribute on editable-text determines how tall the field can/has to be
- if max-lines is not specified (or is below 1), the `textarea` will expand as much as needed without overflow
- removed hardcoded height and variables from css, substituted them with usage of these attributes for consistency
- No fields have any `max-lines` at this point, so all fields will expand as needed

Addresses: #3646

### DIFF
--- a/chrome/content/zotero/elements/editableText.js
+++ b/chrome/content/zotero/elements/editableText.js
@@ -41,7 +41,9 @@
 			'aria-labelledby',
 			'value',
 			'nowrap',
-			'autocomplete'
+			'autocomplete',
+			'min-lines',
+			'max-lines'
 		];
 		
 		get noWrap() {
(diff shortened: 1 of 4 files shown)


dstillman:
Sorry, I wasn't saying that we shouldn't uncap Title now — just that there will be other changes for it to help with the longer height. We should remove all the max heights now.

abaevbog:
Oh, gotcha

abaevbog:
Re-pushed. The `limited-height` is not used anywhere at this point but I thought it's better to have a special attribute than gutting all `max-height`-related css and variables, since they might be useful somewhere else in the future

abaevbog:
@AbeJellinek, here I added `max-lines` and `min-lines` attributes that set the variables on the `editable-text` component. Not adding `max-lines` (as right now) or setting it below 1 will have the fields expand as much as needed.

I assumed we should only use attributes to handle height for consistency, so I remove previously hardcoded height and variables from the stylesheet. Let me know if it makes sense!

Issue #3651, opened 2024-01-31
Cmd/Ctrl-Shift-T reopens wrong tab


dstillman:
https://forums.zotero.org/discussion/111329/z7-beta-ctrl-shift-t-doesnt-go-to-reopened-tab

PR #3653, opened 2024-01-31
fix to wrong tab being selected after close tab undo

abaevbog:
Jump to reopened tab's index only after async Zotero.Reader.open that adds the tab in Zotero_Tabs.undoClose is finished.

Fixes: #3651


dstillman:
Fixed in https://github.com/zotero/zotero/commit/ebd3d7a9889b0e6503e893c7825b7b84af1ef628 in a bit cleaner way

(I'm also not sure the logic was correct here if multiple tabs were closed together and reopened. Seems like this would be making multiple `this.jump` calls in that case.)

Issue #3643, opened 2024-01-30
Remove or fix collection highlight on Ctrl


abaevbog:
Now that the collections appear in the `itemPane`, does it make sense to remove `ZoteroPane.collectionsView.setHighlightedRows`? It seems redundant now. 

Most importantly, if you press `Ctrl-C`,`Ctrl-V` to copy-paste something on windows, the initial `Ctrl` highlights the collection rows, which is somewhat unexpected (especially for new users who may be not aware of the shortcut).

stakats:
We should definitely remove the highlighting. My only question is whether we would want to preserve ctrl as a way to open the `itemPane` collections section

adomasven:
I need highlighted virtualized table colors for the citation explorer PR, so we should fix the colors even if we remove the behavior.

abaevbog:
As a start, I removed highlighting on Cmd/Ctrl from everywhere in `ZoteroPane` though i kept the `setHighlightedRows` in collectionsView. It's not used anywhere right now but we might want to keep it in case if its's useful for something else later?

I tried a few different colors for the highlighted row of virtualized table, and `--accent-yellow` looked the best to me in both modes. I added it as a separate commit to that PR

abaevbog:
> Sorry, late to this, but what's the argument for removing, other than that we wouldn't have to fix dark mode?

I suppose the arguments are redundancy (e.g. why have 2 solutions to one problem?) and the copy-paste behavior on windows where `Ctrl` from `Ctrl-C` triggers the highlight on windows right now. The later was the reason why I brought this up because if felt odd that as I was copy-pasting text in the `itemPane`, something got highlighted on the other side of the screen. As I am typing this, I am thinking that maybe a better way to address this is to just not do the highlighting if an `input` or `textarea` if focused? That way, it would really not get in the way by highlighting things when you want to copy something, while still being available for those who know about it.

And per @adomasven, the dark mode needs to be fixed regardless if we remove highlighting of collections or not. I think from our existing colors `--accent-yellow` works quite well as a dark-mode friendly substitute.

PR #3657, opened 2024-01-31
collection highlight tweaks

abaevbog:
- Substituted previous yellow highlight color for `--accent-yellow`
- On windows, do not trigger collection highlighting when input is focused so that copy-paste Ctrl-C/Ctrl-V do not trigger highlighting

Addresses:  #3643

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -1034,7 +1034,10 @@ var ZoteroPane = new function()
 			//
 			// We use Control (17) on Windows because Alt triggers the menubar;
 			// 	otherwise we use Alt/Option (18)
-			if ((Zotero.isWin && event.keyCode == 17 && !event.altKey) ||
+			// On windows, do not highlight when an input is focused to not interfere with
+			// Ctrl-C/Ctrl-V copy-paste shortcuts
+			let inputFocused = ["input", "textarea"].includes(document.activeElement.tagName);
+			if ((Zotero.isWin && event.keyCode == 17 && !event.altKey && !inputFocused) ||
 					(!Zotero.isWin && event.keyCode == 18 && !event.ctrlKey)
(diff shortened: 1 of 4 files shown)


dstillman:
@yexingsha, this is `accent-yellow`?

Screenshot 2024-02-02 at 4 05 21 AM

At least in light mode, I think we want to keep this closer to the actual yellow it was before, not orange.

Screenshot 2024-02-02 at 4 09 03 AM

yexingsha:
`accent-yellow` is used for icons, so it needs to be orange to provide enough contrast while maintaining saturation. Let's add a new `accent-highlight` for highlights in the collection tree, and perhaps also in the item tree for future multi-line search result functionality. Light version is `#ffea0080`, dark version is `#ffd40026`.

Sidebar-light Sidebar-dark

The yellow is a bit more orange so that it doesn't look greenish in dark mode.

abaevbog:
Re-pushed. I think it looks great!

dstillman:
Screenshot 2024-02-06 at 12 54 29 PM

Do we know why this is looking different?

abaevbog:
Hm, I am honestly not sure... The HEX of the color `#ffea0080` is the same as in figma but it certainly looks a bit different. 
I took rgba of that color from figma and increased the opacity from 0.5 to 0.7. That second row looks a bit closer to what we see in figma, right? 

Screenshot 2024-02-06 at 2 04 17 PM


I am doing some googling now, and it seems this is something that has been brought up before: https://forum.figma.com/t/figma-colours-dont-match-web-export/20526/7

yexingsha:
It's a color space issue, Figma uses sRGB color space while macOS uses P3 by default, so they can look different, especially when rendering opacity overlays. I use sRGB on my device, so the colors I see in the app are exactly the same as in Figma:

Screenshot 2024-02-06 at 14 07 38

And I think the P3 color looks fine, so let's keep the `#ffea0080` as is.

Issue #3575, opened 2024-01-25
Test timeout: Item Tags Box "should update tag when pressing Enter in textbox"


dstillman:
We tried to fix this previously in a0ab7783ee2fae5551960bc1e99453f08e01229c, but possible that the `activate()` call doesn't work reliably on Linux?

Test is disabled for now

abaevbog:
I played around with this while I had my linux setup working, and `activate()` promise does seem to be the culprit. I pushed how I got that test to pass

dstillman:
Getting failures here again

https://github.com/zotero/zotero/actions/runs/9201561657/job/25309969100

PR #3665, opened 2024-02-01
Fix to tagsbox test timeout in CI

abaevbog:
Do not wait for the activate promise on linux. Passing `should update tag when pressing Enter in textbox` test during CI run.

Fixes: #3575

### DIFF
--- a/test/tests/tagsboxTest.js
+++ b/test/tests/tagsboxTest.js
@@ -16,15 +16,18 @@ describe("Item Tags Box", function () {
 	
 	
 	describe("Tag Editing", function () {
-		it.skip("should update tag when pressing Enter in textbox", async function () {
-			// editable-text behavior relies on focus, so we first need to bring the window to the front.
-			let win = Zotero.getMainWindow();
-			let activatePromise = new Promise(
-				resolve => win.addEventListener('activate', resolve, { once: true })
-			);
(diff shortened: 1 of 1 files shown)


test/tests/tagsboxTest.js
```diff
@@ -16,15 +16,18 @@ describe("Item Tags Box", function () {
...
 			let activatePromise = new Promise(
 				resolve => win.addEventListener('activate', resolve, { once: true })
 			);
 			Zotero.Utilities.Internal.activate();
 			Zotero.Utilities.Internal.activate(win);
-			await activatePromise;
+			// Required on macOS but causes a timeout on linux
+			if (!Zotero.isLinux) {
+				await activatePromise;
+			}
```
dstillman, line 30: If we're not going to await it we don't need to add the event listener above either.

And is it really that waiting for activation isn't required on Linux, or that activation isn't working, or that the Z window is automatically in front on Linux, or…? And what's the actual behavior this is relying on?

abaevbog: Ok, yeah, what I re-pushed should make more sense now. As I understand, the window is already at the front on Linux, so all activation should just be skipped. I swapped the condition from checking the platform to `doc.hasFocus()` because I think it better reflects when the activation should happen

PR #3668, opened 2024-02-01
Show title/abstract in section headers and make title sticky when expanded

### LINKED ISSUE #3660: Change the position of Title and Locate
Since we moved item Title into the new sidepane header, a lot of people have been looking for it in the Info section and missing it entirely, or not realizing that it can be edited in the header. It's also causing some trouble with manual item creation and merge.

Therefore, we are moving Title back to Info, with `position: sticky` so that it sort of becomes a header when you scroll down:

https://github.com/zotero/zotero/assets/57771795/e38c3f88-0774-4281-830c-27a5d6e24f57

Some implementation details:

1. The sticky title can be edited in place.

2. When a section below Info is pinned, switching to a different item should still have the sticky title on top.

3. When the Info section is collapsed, we no longer get this sticky title, but instead show title preview in the Info section header as in #3647:
	
	library-compact-item

4. For attachments, the same design applies, putting the title below the Attachment Info section header and above preview.

5. This isn't shown in the mock, but ideally we should give the title slightly different styling when it's scrolled to the top:
	- More `padding-block` for this div (not the text area). +2px in compact, +3px in comfortable
	- `fill/quinary` bottom border
	- Set `max-lines: 5` so super long titles don't take up too much vertical space

6. The Locate button is moved to the sidenav, below the Related section button and a divider, with no dropmarker. With plugins, the order in the sidenav will be: Sections (+ plugin sections) | (plugins that create separate views like note editor in reader) | Locate (+ plugin buttons, either regular buttons or menubuttons).

AbeJellinek:
Done:

- Titles and abstracts in section headers
- Sticky title
	- Was a bit tricky to get working because `position: sticky` stops when it hits the edge of a block element, and there's no way we could make the item box inline without pain. But with some JS to switch to `position: fixed` at the appropriate point, it works!
- Text transforms on the newly restored title field
- Locate moved to sidenav
- Locate for multiple items
- Leaving space for the sticky header when scrolling to a pane

WIP, not done yet:

- Adding extra padding within the sticky header when pinned (I experimented with this but it was causing infinite pin-unpin-pin loops when the title was less than its padding away from the top - need to investigate further)

Fixes #3660


abaevbog:
Absolutely! The general place where the keyboard navigation is setup for the itemPane is this [keypress handler](https://github.com/zotero/zotero/blob/6486c509ad8181c0a0861e5820ae445956aad2f5/chrome/content/zotero/itemPane.js#L174).

The most reliable way to handle keyboard navigation that I found is to handle a keypress based on the key and the target (manual, somewhat annoying but it does work). For example, to move from the title to the itemType of the itemBox, we can add something like this in there:

```
if (event.target.parentNode.id == "itembox-field-value-title" && event.key == "Tab" && !event.shiftKey) {
	this.querySelector("#item-type-menu").focus();
	stopEvent();
}
```

I see that this, however, does not currently work because the `Tab` keypress  is being [picked up and handled](https://github.com/zotero/zotero/blob/6486c509ad8181c0a0861e5820ae445956aad2f5/chrome/content/zotero/elements/itemBox.js#L1874) by the keyboard handler in `itemBox`, so the `Tab` event never reaches the handler in `itemPane`.

I think that we never want the title to be handled by `itemBox.handleKeyPress`. The title is always at the top, so I assume you will tab into it from `itemTree` and if there's a pinned section, we want the next tab press to be handled by `itemPane` to go to that section, as opposed to having `itemBox` send it to the next table field.

So maybe we just explicitly skip keypress events from the title in `itemBox.handleKeyPress`?

```
if (event.target.parentNode.id == "itembox-field-value-title") { return; }
```

With these two things, the focus can move from the active title to the item type menu.

I am noticing that for some reason, tab from the twisty of the itemBox section doesn't activate the `editable-text` of the title ... `document.activeElement ` is the `editable-text` itself and not it's `textarea` - not sure why that happens but it makes it so that these changes will work only when the title is already active (after mouse click or previous shift-tab from item type menu).I'll try to track down why that happens. It's probably best not to have `ztabindex` added to the title either - I'm looking into the itemBox changes now to find where it comes from in there.

windingwind:
will conflict with #3624, maybe wait for that to be merged?

windingwind:
I see this:

https://github.com/zotero/zotero/assets/33902321/08bc09ea-243f-491c-ad65-184408ef0f56

abaevbog:
> Interesting! I don’t think I was getting that. I’ll see if I can reproduce it.

I actually see the same behavior specifically with the reader opened.

About the tabbing: I looked and the a good chunk of the odd behavior with tabbing actually comes from too large `ztabindex` being assigned to the title. We can give it a `ztabindex` of one (as well as shift the index of everything else) with a few tweaks like [here](https://github.com/abaevbog/zotero/commit/b8110fb3820c256d69e54153c11d0ef13eb549f4).

Something like that will work within itemBox but we might want to handle title as a special case in whatever is the component that will contain keyboard nav logic of itemPane (after #3624) to tab to the pinned section from the title. I have #3602 with a few other keyboard-nav tweaks that I'm planning to update once this and #3624 is done (e.g. we'll also need to make the locate button somehow accessible). If it's easier, I can add take care of this issue there as well once these PRs are settled!

abaevbog:
@AbeJellinek, yeah I noticed that too and trying to look into what's going on there. The oddity is that the `document.activeElement` is the correct element: `editable-text` of the title. It just ... doesn't have it's `textarea` activated. I'm digging into it now to see if it's something small, in which case I'll just re-do that commit. If it's something trickier, I'll tackle it in a separate PR so that I don't end up making more conflicts for you. And yes, the itemBox focus setup is very special!

On a separate note, I am seeing a new glitch with `editable-text` stretching it's parent if the content has no whitespaces. Unlike earlier, it's not caused by any kind of mismatch between text in `textarea` and within the `::after` component.

Screenshot 2024-02-05 at 3 14 51 PM

abaevbog:
@AbeJellinek, I think I tracked down the issue with the focus of the title. `tabindex=0` explicitly set on `editable-text` (leftover from pre-redesign setup) was not strictly needed but didn't cause any issues before. In this case, though, it made the focus get stuck on `editable-text` without being moved to the underlying `textarea`. Does [this small change](https://github.com/abaevbog/zotero/commit/5dce17d37b8bde0e8ba3ae56246c006d8973590b) fix it for you?

yexingsha:
We definitely want to set a max height (5 lines) for sticky title, and ideally its height would be reduced at the same pace the rest of the pane scrolls up, which kinda look like this (minus the text fading part):


https://github.com/zotero/zotero/assets/57771795/b8f71ceb-871a-4ba8-a440-c04248aa4af2

dstillman:
Are we able to add an ellipsis when it's truncated?

dstillman:
Can we do some sort of fade effect? Just cutting it off isn't great. @yexingsha, thoughts?

dstillman:
Maybe a slightly smaller shadow radius? @yexingsha, any thoughts?

And this will only appear for really long titles, right?

yexingsha:
I had half formed thoughts about the fade effect but now I've broken the app so I can no longer see it. Steps to reproduce:

1. Once a super long title is stuck to the top, using sidenav to scroll to sections no longer works properly
2. I cannot scroll to the very top of the item pane to see the Info section header either
3. Switching between items also becomes buggy, with short titles also taking up the whole height. There didn't seem to be any way to get out of this, so
4. I restarted the app, and now the super long title is broken. I only recorded the screen now:

https://github.com/zotero/zotero/assets/57771795/2f939cfe-e6be-435a-aacc-faef3f75fc75

No idea if the weird things happening in the tab bar existed before the restart.

yexingsha:
I think option 2 is not that weird, but I agree that this might not be a worthwhile problem to solve. So let's just go with 3 for now?

dstillman:
I think Title needs a larger font-size — it gets sort of lost now between Info and Item Type without a label. (I actually thought this before as well, but it's more of a problem in the new location.)

Current (13px):



14px:



15px:

abaevbog:
>  tab fix works beautifully.

Great! 

About the locate button and the sidebar in general, I have a question: when reader is opened, we have not only locate button but also notes button that needs to be reachable with keyboard somehow. It has to be clicked for the `notes-pane` to show up. What about plugins? Are plugins going to be able to add buttons like that to the sidebar that must to be clicked for a pluggin-specific pane to show?

Regardless of that, I like the idea of arrows moving focus to the sidepane buttons that are not accessible otherwise. I think it's more intuitive than having a shortcut that you have to remember, especially if there are multiple buttons like that. The issue is the right/left arrows from the header are currently used to collapse/expand the section. The item/context panes' scrollable area itself is focusable - what if we use arrow-right/left from that?

windingwind:
> What about plugins? Are plugins going to be able to add buttons like that to the sidebar that must to be clicked for a pluggin-specific pane to show?

I think I can answer that. Currently, we don't support this behavior, but some/many plugins are already doing this; We will have an API to add sections under the item pane (`ItemDetails`), and after we release that API, we can wait and see how people reacts to it to decide whether to provide the ability to register panes at the same level of notes pane.

It sounds good to be able to focus the sidenav buttons with arrow key for me.

abaevbog:
@AbeJellinek, I'm looking into setting up navigation for sidenav buttons. Meanwhile, i am noticing that the title looks a bit strange during editing in sticky mode.

Normal:
Screenshot 2024-02-09 at 12 43 41 PM

Sticky: 
Screenshot 2024-02-09 at 12 43 36 PM

Part of the box-shadow seems to be gone somehow?

abaevbog:
The title looks great now!

And [this commit](https://github.com/abaevbog/zotero/commit/1e69e7dd9590a7047d3910356da3d2d30642a800) is my idea for keyboard navigation. ArrowRight/left from the scrollable area of the itemPane will focus the first relevant button, up/down from there will navigate buttons, and arrowleft/right towards the pane will re-focus the pane. What are your thoughts? If it's too many different buttons, we could just use side arrows. E.g. pressing arrow right multiple times from the scrollable area will go through all available buttons, and arrow left will go back.

One thing about this approach: arrow up from locate to notes button (when reader is opened) for some reason scrolls the contextPane down for me... I couldn't track down the root of this behavior yet - are you seeing it as well?

yexingsha:
I think 14px and `font-weight: 600` works for the title.

Right now long titles seem to make sidenav scrolling inaccurate. Also, this might be a non-issue, but switching from a non-sticky super long title to a sticky title while a section is pinned causes the title and content to overlap:

https://github.com/zotero/zotero/assets/57771795/74a49e87-3b7a-4fd4-b55f-90b2966bfff9

dstillman:
1. When the retractions banner is showing (and probably other banners), the top of the title box slightly overlaps the banner.
2. I'm worried this was a bad idea (of mine), or that we need to do something a little different. When you scroll up and lines are partially cut off, it looks…not great? Maybe it just needs a darker bottom border or some styling of the title box to more strongly set it apart? This border color is very light, and so the text kind of gets cut off into nothingness, which never looks great.

dstillman:
Sorry, I was referring to the text below getting cut off at an almost imperceptible border as you scroll up.

> and that we've had to shunt Locate to a place where it doesn't belong

It was pointed out in the forums that when Locate is in the item pane you can't run it for multiple items at once, so I think we do actually have to move that. So the comparison here would be full-width title above Info or sticky.

dstillman:
I think this never got the bigger font size we agreed on? I'm seeing 13px still.

I worry a little about Locate not having a dropmarker. It's sort of like the icon equivalent of "…" in menu options — this just seems like it's going to open something in your browser, so people may not be comfortable clicking it. Any other ideas there?

yexingsha:
I guess we can make the tooltip say "Locate options" or something similar, to convey that this button opens a menu. And we can also change the icon:

Screenshot 2024-02-23 at 15 08 09

yexingsha:
Something like binoculars, or glasses? Or is that too specific?

yexingsha:
Maybe?

Screenshot 2024-02-23 at 17 02 08

dstillman:
The signpost is cute but might be a bit too literal. I think I'd go with the second of the arrow-in-circle for now — the same icon with the dropmarker, multiple arrows notwithstanding — and we can keep thinking about it.

And then not to further complicate things, but have we discussed reordering in the context of this change? E.g., what happens if you [put Abstract at the top](https://forums.zotero.org/discussion/comment/456628/#Comment_456628)? We've had other people say they want to put the attachment preview at the top. So either there's no title at the top in that case or we have to do some annoying conditional thing. Neither seems great!

There's another suggestion immediately above that comment to [show title/creator/year at the top](https://forums.zotero.org/discussion/comment/456504/#Comment_456504), but that seems 1) pretty duplicative, 2) somewhat academic-y, and 3) potentially complicated, since people would want different things in terms of first author, last author, <i>et al.</i>, a full bibliography entry, etc. But I guess that would be an option — a read-only reference in one of a few formats, customizable with CSL, and then keep all the fields in Info.

yexingsha:
locate

re: section reordering, I agree that's an essential feature which just doesn't work with this PR. I do like the title/creator/year at the top with customizable formats: it's only duplicative when you are actively looking at or editing metadata, and you can keep the info section collapsed when you are reading papers, which is (I assume) the majority of the time when you are using zotero. Also, I think academic-y is a point in favor rather than against.

dstillman:
> you can keep the info section collapsed when you are reading papers, which is (I assume) the majority of the time when you are using zotero

That's interesting. That could actually pretty radically change how people use the item pane — choose to show a full bibliography entry at the top and only open Info when you notice something wrong. (That may not be wholly realistic, since some omissions might not be obvious from just the bib entry, but it's at least a possibility much more of the time. And less citation-focused users could probably leave Info closed almost all the time.)

OK, let's try that. This was cool, though, @AbeJellinek! I'll close this so we have this code for posterity, and you can open a separate PR with just the Locate button change. I'll open a separate issue for the header customization.

windingwind:
This is a fancy work! 

Now that seems no longer to be a blocker of #3624, could we consider dealing with that?

yexingsha:
We are keeping the Abstract snippet, and limiting titles to 1/4 of pane height sounds good to me.

PR #3678, opened 2024-02-06
Fixes for Linux appearance for various DEs

adomasven:
- Titlebar buttons now appear in the correct position based on the DE preferences
- Removed the Z in the titlebar for Linux
- Tabs aligned with the left margin on Linux and Windows
- Restored popupmenu styling to GTK widget-defined styling
- Fixed some rendering issues on Unity DE

Before:
| Gnome/KDE | Pantheon | Cinnamon | Unity |
|-------------------|--------------|---------------| --------|
| image | image | image | image |

After:
| Gnome/KDE | Pantheon | Cinnamon | Unity |
|-------------------|--------------|---------------| --------|
| image | image | image | image |

Other notes:
- Dark mode on Linux has been tested and looks good on all DEs
- Unity has dark menus in both Light and Dark mode as per normal DE settings
- Unity has cuttof titlebar buttons and weird spacing. There's no way to select just Unity with CSS and Thunderbird unfortunately has the same issue, so most likely some problem in the way they are specified in the GTK theme (or maybe even an issue in my install)
- macOS and Windows tested and are unchanged UI-wise (except for Windows alignment of tabs)

### DIFF
--- a/chrome/content/zotero-platform/win/overlay.css
+++ b/chrome/content/zotero-platform/win/overlay.css
@@ -2,11 +2,6 @@
 	--theme-border-color: #cecece;
 }
 
-/* Hide horrible blue effect for menu bar and toolbar */
-#navigator-toolbox {
-	-moz-appearance: none;
-}
-
 tree {
 	border-width: 0;
 }
--- a/chrome/content/zotero/zoteroPane.xhtml
+++ b/chrome/content/zotero/zoteroPane.xhtml
@@ -101,8 +101,53 @@
 		<!--EDIT-->
 		<command id="cmd_find"
 			oncommand="document.getElementById('zotero-tb-search-textbox').select()"/>
+			
+		<command id="cmd_zotero_reportErrors" oncommand="ZoteroPane_Local.reportErrors();"/>
+		<command id="cmd_zotero_import" oncommand="Zotero_File_Interface.showImportWizard();"/>
+		<command id="cmd_zotero_importFromClipboard" oncommand="Zotero_File_Interface.importFromClipboard();"/>
+		<command id="cmd_zotero_exportLibrary" oncommand="Zotero_File_Interface.exportFile();"/>
+		<command id="cmd_zotero_advancedSearch" oncommand="ZoteroPane_Local.openAdvancedSearchWindow();"/>
+		<command id="cmd_zotero_copyCitation"
+				 oncommand="ZoteroPane_Local.copySelectedItemsToClipboard(true);"
+				 disabled="true"/>
+		<command id="cmd_zotero_copyBibliography"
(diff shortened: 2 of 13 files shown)


adomasven:
>For Gnome/KDE, where is the menu bar background color coming from, and where was it coming from before? I guess it didn't actually match the toolbar before, but it was much closer, and I assume we actually wanted it to match. (It does on Windows.)

Previously the menubar colour came from the menubar widget colour defined by the DE theme, now it comes from the titlebar widget colour. Menu colour was set by Zotero (which caused the display issue in Unity in light theme), now it also follows the DE theme. We could make the menubar use the menubar widget colour instead, but I think that will generally look worse across all DEs. Firefox uses the titlebar widget colour for its tabbar, as does Thunderbird.

> Are the closer window controls on Gnome/KDE correct?

They are defined by Firefox/DE, I did not change them, just adjusted the positioning (to fix the Unity looks). The spacing between the items is controlled by the DE if Firefox correctly imports theme settings. We cannot control individual DE spacing for those buttons, but we can either space them out or squish them more for all DEs.

> It looks like there's some weirdness in the window controls on Unity — not evenly spaced out and cut off on the left and right.

Yes, noted in the original comment. This actually may just be some weirdness in my config (screenshots from my native Linux install, others from a VM), and they look the same (cut off and weird spacing) in Thunderbird, so this is according to the theme. Unity is not as actively maintained anymore, so may be part of the reason why it looks like that too, but I can check in a new user.

dstillman:
The problem is that this looks wrong on stock Ubuntu — it looks different from other apps and is too close to (but different from) the menu bar color.

Before:

Screenshot 2024-02-12 at 5 38 25 PM

Now:

Screenshot 2024-02-12 at 5 39 34 PM

Z icon aside, the previous styling was closer to what you get in other apps:

Files:

Screenshot 2024-02-12 at 5 40 47 PM

LibreOffice:

Screenshot 2024-02-12 at 5 44 12 PM

adomasven:
Do you see this if you run Zotero without -ZoteroDebug, i.e. without the debug output window? That just looks like the Zotero window is unable to get focus properly for some reason (looks like an unfocused Gnome window).

dstillman:
Sorry, I must've clicked the desktop when I took that. It's this:

Screenshot 2024-02-14 at 3 03 33 AM

So color is actually OK, but we need a bottom border like we had before.

adomasven:
Updated:

image

dstillman:
Looks good. Want to just separate out the async/await changes?

adomasven:
> Looks good. Want to just separate out the async/await changes in the two files?

Oh yeah that was supposed to be a separate fix.

adomasven:
Done.

PR #3680, opened 2024-02-06
allow focus to move past the note to tags section

### LINKED ISSUE #3675: Can't tab to Tags/Related from note editor
https://forums.zotero.org/discussion/111526/zotero-7-beta-switching-between-editing-note-tags-and-related-fields

I'm not sure I have a good suggestion here…

abaevbog:
- Make the actual core editor of the note focusable
- Shift-tab when the editor is active (the cursor is visible) will move focus to the core editor and add a box-shadow indicator, similar to focus-ring everywhere else.
- From the core editor, another shift-tab will focus the buttons, and tab will focus the first focusable section's header after the note (tags).
- Enter or space when the core editor is focused will move focus back inside of the editor and the cursor will appear.

Fixes: #3675

Some of this logic (`focusin` and `focusout` listeners) could be moved to the actual note editor module, and it would be more concise. It would boil down to adding a `:focus-visible` to `.editor-core` in `editor.scss` to display the box-shadow when the editor is focused, and adding `tabindex=0` to `<div className="editor-core">`. I wasn't sure if it's a good idea to mess with the submodule since the note editor is used is multiple places, so I kept everything in `noteEditor.js`

### DIFF
--- a/chrome/content/zotero/elements/collapsibleSection.js
+++ b/chrome/content/zotero/elements/collapsibleSection.js
@@ -123,8 +123,8 @@
 			this._head.setAttribute("tabindex", "0");
 			this._head.addEventListener('mousedown', this._handleMouseDown);
 			this._head.addEventListener('click', this._handleClick);
-			this._head.addEventListener('keydown', this._handleKeyDown);
 			this._head.addEventListener('contextmenu', this._handleContextMenu);
+			this.addEventListener('keydown', this._handleKeyDown);
 
 			this._title = document.createElement('span');
 			this._title.className = 'title';
@@ -257,8 +257,8 @@
 		destroy() {
 			this._head.removeEventListener('click', this._handleClick);
 			this._head.removeEventListener('mousedown', this._handleMouseDown);
-			this._head.removeEventListener('keydown', this._handleKeyDown);
 			this._head.removeEventListener('contextmenu', this._handleContextMenu);
+			this.removeEventListener('keydown', this._handleKeyDown);
 			
 			Zotero.Prefs.unregisterObserver(this._notifierID);
 		}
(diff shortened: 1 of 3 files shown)


abaevbog:
https://github.com/zotero/zotero/assets/36271954/9b7fdd86-3b2f-4aed-a2c1-0a7e813273f2

dstillman:
Not sure if this is new, but I don't think Shift-Tab should leave the note editor, since it has a function (unindent) within the note editor itself. It's weird that you can press Tab on a random line in the note to indent it, Shift-Tab to unindent it, and then press Shift-Tab again and find yourself out of the editor.

It looks like Esc already focuses the toolbar, which seems sufficient if you want to get to resume keyboard navigation outside the editor, including to Tags or Related: just Esc → Tab → Tab and you're on Tags.

And then Shift-Tab from Tags shouldn't enter the editor — it should focus the core editor, the inverse of Tab from the core editor, which takes you to Tags.

abaevbog:
Gotcha! Yeah, that's existing behavior that, as far as I can see, comes from `note-editor` itself. `shift-tab` from within the `note-editor` in the web library behaves the same way by focusing the toolbar, and I think I found a `keydown` listener that is responsible for it. I'll explore what to tweak in the `note-editor` to adjust the `shift-tab` behavior. Since the `note-editor` will need to be updated for it, we can move that `focusin`/`focusout` logic in there as well, which will be cleaner.

abaevbog:
So I made [this change](https://github.com/abaevbog/note-editor/commit/93b0c110dc6229d898944eb2fd5f1d00d3205bc4) to the `note-editor` to make the editor itself focusable and stop shift-tab from focusing the toolbar. Then, I added some logic to not place focus inside the editor on shift-tab from tags and a small refactoring to where the listener is located to handle keypresses in sections below notes.

Tabbing through the note-editor:
https://github.com/zotero/zotero/assets/36271954/cdfcbaa9-382f-4e97-b391-daf191f3d785

One question: do we want these changes (focusable editor and no focus change on shift-tab) to apply to `note-editor` everywhere or just the desktop? E.g. web-library that has no sections below the note editor?

PR #3687, opened 2024-02-07
reader Escape keypress behavior revised

abaevbog:
- don't re-focus the tab on every escape when the reader is opened (#3602). Instead, only move focus to the tab if Escape came from inside of the reader and if the focus was on `<body>` or on the `#viewerContainer`. That way, if one clicks Esc to close a popup, the focus will remain within the reader. But if one presses Esc when `<body>` is focused, most likely the intention is to leave the reader altogether, so we can focus the tab.
- don't re-focus the tab if Esc comes from outside of the reader. E.g. if a note is opened, Esc should close the modals without moving focus from the note
- force the focusring to show on the tab when focused - just so it's clear where the focus is at


abaevbog:
@mrtcode, thank you for bringing up that [side effect](https://forums.zotero.org/discussion/comment/454619/#Comment_454619) of the recent keyboard nav update when Escape will move focus to the tab even when it shouldn't (e.g. Escape meant to close the search popup). I tried to just move the same logic to a listener that doesn't capture the event but the issue was that the Escape events still get out of the reader and reach `ZoteroPane`, so the behavior was the same.

As an alternative, I thought maybe we re-focus the tab only when the Escape keypress target is `<body>` or `#viewerContainer`, which seem to be the elements where the reader places focus when Escape is pressed? That way, if a button/input/annotation/etc. is focused, the first Escape will do nothing and the second Escape will take focus out of the reader to the tab. What do you think about this approach?

warwickmm:
Feel free to dismiss this suggestion, but could a different shortcut (e.g., `Shift`+`tab`) be used to re-focus the tab bar?  I (perhaps erroneously) expect hitting `Esc` to be an idempotent operation, so `Esc` to close a popup followed by `Esc` to focus the tab bar is not intuitive to me.  But I'll gladly accept it if others find it more convenient.

abaevbog:
I do agree that it's not 100% ideal - the reason why this special treatment is needed in the first place is that once the focus gets inside of the reader, it's tricky to get out. Typical keyboard shortcuts like `tab` and `shift-tab` are handled by the reader to move focus between elements in it but it can't conveniently send the focus outside of itself back to the tab bar. 

The reason why just `escape` is proposed here is if you find yourself needing to move the focus away from the reader just by using a keyboard for navigation, hitting `escape` once or twice is probably what you'll intuitively try to do. Besides, even before these changes, `escape` was not always strictly idempotent, and focus-related jumps do happen in a few other cases (e.g. if one is editing a text field, then opens a tabs menu, and clicks `escape` twice - first escape will close the tabs menu, and the second one will blur the text field). Personally, I am reluctant to suggest shortcuts in general because they're not very intuitive and it becomes just another thing one has to remember.

That said, if this `escape` idea is an issue usability-wise, there are some things we could try. For example, we already have `Cmd`+`[1-9]`  as a shortcut to open tab at the specified index. I suppose we could think about something like `Cmd`+`0` to re-focus the currently selected tab which would sorta address this issue as well - again, the issue here is that one would have to "discover" this option.a

warwickmm:
> I suppose we could think about something like Cmd+0 to re-focus the currently selected tab which would sorta address this issue as well 

I believe `Cmd`+`0` (or `Ctrl`+`0` on other platforms) is already mapped to "Automatically Resize"?

If `Esc` is really the best option available, that's fine with me.  I don't want my preferences to get in the way of improved accessibility.  While `Esc` isn't intuitive to me (I created the separate [simplified thread](https://forums.zotero.org/discussion/111608/zotero-7-esc-focuses-tab-bar) after the [original comment](https://forums.zotero.org/discussion/comment/454619/#Comment_454619) because I thought the new `Esc` behavior was the _actual_ problem), I'll happily concede that my expectations are abnormal.

abaevbog:
> I believe Cmd+0 (or Ctrl+0 on other platforms) is already mapped to "Automatically Resize"?

Yup, you're right. I guess the reason why I don't like shortcuts is because I am the one who can never remember them.

In any case, since the shortcuts specifically for tab navigation do exist, there might be some combination out there that would be reasonable for this scenario - if this is the route we were to take.

warwickmm:
Well, if you're entertaining alternatives, I'll mention `` Ctrl+` `` (ctrl + backtick) as a possibility?  It's to the left of the `1` (on a US keyboard), which can loosely be interpreted as...something related to tabs?

dstillman:
Ctrl+` isn't an option. The Ctrl shortcuts are Cmd on macOS, and Cmd-` is a basic window-control shortcut.

I don't think it's accurate to describe Esc as idempotent, for the reasons @abaevbog mentions, but I do think it's reasonable to say that it's a key that people mash any number of times with a final destination in mind, so having it go first to the content pane (e.g., if you're in an item field) and then to the tab bar is indeed not great. (We don't currently focus the content pane when escaping out of an item field, but we should — it's the default view, equivalent to the items list in the library tab. Similarly, Chrome and Safari both focus the content pane on Esc from the address bar.) So maybe we do need another option here.

mrtcode:
> I tried to just move the same logic to a listener that doesn't capture the event but the issue was that the Escape events still get out of the reader and reach ZoteroPane, so the behavior was the same.

The reader currently doesn't stop propagation for this key. I have to update it (and note-editor as well).

I think it's a better practice to avoid hard-coding internal things like element ids from the reader or other components. And it's always better to avoid capturing phase if that isn't necessary. I'm wondering if other keyboard shortcuts are affected by this.

> As an alternative, I thought maybe we re-focus the tab only when the Escape keypress target is <body> or #viewerContainer, which seem to be the elements where the reader places focus when Escape is pressed? That way, if a button/input/annotation/etc. is focused, the first Escape will do nothing and the second Escape will take focus out of the reader to the tab. What do you think about this approach?

Reader currently uses Escape to deselect annotations or clear text selection in the view, but I'll probably eliminate this behavior, because at least on macOS this isn't common to do for Escape.

@warwickmm I see your point, but lets see how it goes. At least on macOS, for example, the first Escape press can close browser find popup, while the second press can exit full screen mode.

dstillman:
> Reader currently uses Escape to deselect annotations or clear text selection in the view, but I'll probably eliminate this behavior, because at least on macOS this isn't common to do for Escape.

I think this is good, though. I really do think there's value in being able to mash on Esc a bunch of times and then know that you can then use up/down-arrow to scroll the document.

abaevbog:
> We don't currently focus the content pane when escaping out of an item field

That was partly the behavior before these changes: a reader would always get re-focused if Escape is pressed. The issue was that sometimes we wouldn't want it: anytime one clicks escape to close a popup or a search tab in note-editor, the focus would then jump to the reader. I'll update to bring reader re-focus on escape back and try to make it smarter. If the note-editors stop propagation on Escape, that will be great!

I think the consensus is not tie focus movement out of the reader to Escape? If Escape behavior is subject to change in the reader, relying on it was definitely a bad idea.

We could indeed add a shortcut that directly or indirectly helps here. Cmd-Shift-L focuses the collectionTree in the library, Cmd-Shift-K focuses the quick search in the library, what if Cmd-Shift-J were to focus the title in the itemPane/contextPane depending on which tab is opened? I think it would be somewhat consistent with other shortcuts and would allow you to get the focus out of the reader, from where you can shift-tab to the tab bar. Not very intuitive but it might be more handy than having a dedicated shortcut for focusing the current tab.

I noticed that there actually is a way to get the focus out of the reader - `_togglePane` called by the reader focuses the context pane. As another suggestion, maybe a similar approach would work to allow one to focus the tab bar? Could we add a function that would refocus the tab bar on shift-tab from the reader's toolbar?

dstillman:
Oh, wait, but why is Sync before Title? The content pane seems equivalent to the items list — between the toolbar and then item pane. Shouldn't it be Sync → toolbar → content pane → Title?

And if the context pane was closed, Tab from the content pane would just focus the tab bar.

Going the other way, Shift-Tab from the content pane already focuses the Toggle Sidebar button. So then we'd just need Shift-Tab from that button to focus the sync button, like it does in the library view. Then it would be four Shift-Tabs from the reader to the current tab (Toggle Sidebar, Sync, open-tabs menu, current tab).

mrtcode:
> Could we add a function that would refocus the tab bar on shift-tab from the reader's toolbar?

Yes, that is possible, but then pressing Tab again should focus the PDF reader toolbar again? Also what will happen with Escape?

By the way, I think we no longer have a way to switch from reader to context pane, or do we?

Also shouldn't Escape focus the current tab when in the Library tab as well?

abaevbog:
> Oh, wait, but why is Sync before Title? The content pane seems equivalent to the items list — between the toolbar and then item pane. Shouldn't it be Sync → toolbar → content pane → Title?

The only reason for this was that once the focus gets inside of the reader, including the toolbar, some coordination between focus handling in ZoteroPane and the reader is needed to get it out. With that initial attempt, I tried to keep it simple and avoid having the reader involved at all (as it's a bit of a foreign territory for me). So I essentially "danced around" the content pane so that we still have a way to tab to the contextPane without touching the reader. Now that we're thinking that the reader has to be involved, there'll be no reason for that workaround and it really can go Sync → toolbar → content pane → Title.

abaevbog:
>  Yes, that is possible, but then pressing Tab again should focus the PDF reader toolbar again?

Right, though since we have the reader involved and there's no need to move the focus around it (like I tried here), we would probably want Shift-Tab from the toolbar to focus the sync button and the tab from the sync button to focus the toolbar.

> Also what will happen with Escape?

I'm not 100% sure what the consensus on Escape is. From earlier discussion, my impression is that Escape should be left out of this focus management process at all, and clicking Escape should not move focus out of the reader? That means I'll undo my earlier changes that do that.

> By the way, I think we no longer have a way to switch from reader to context pane, or do we?

I may have just missed something but the only way to get the focus from the reader to the context pane that I found (except for this Escape experiment) was to toggle the context pane closed and open, which shifts the focus.

> Also shouldn't Escape focus the current tab when in the Library tab as well?

If we were to stick with that Escape tab-focusing behavior, we could do that for consistency, I suppose. But, again, as I understand, we want Escape to just re-focus the reader as before and not move focus elsewhere. (Correct me if I misunderstood that point)

dstillman:
> But, again, as I understand, we want Escape to just re-focus the reader as before and not move focus elsewhere.

Yes.

abaevbog:
@mrtcode, I am trying to find a good starting point from where to suggest additions to the reader's focus handling so that the focus goes Sync → toolbar → content pane → Title. Reader is quite foreign to me, so could you point me in which direction to look for the bulk of the keyboard navigation logic? 

My goal would be to achieve 2 things here:
- a way to have reader place focus on the toolbar (something like `reader.focusFirst()` but so that the focus would land on the sidebar toggle button). This `reader.focusToolbar` would be called on tab from the sync button.
- a way to move focus from the reader to the contentPane, so that when all annotations and etc. have been tabbed through, the contextPane would be next.

abaevbog:
I think I have some idea now - to go from the sync button to the toolbar, I added `focusToolbar` function to `FocusManager` that finds the right node and just call that instead of `reader.focus`. So that seems straightforward.

To go back to sync button from the toolbar on shift-tab, I'm basically copying how `onToggleContextPane` is defined and passed to the reader, and then I'm using that new function in `FocusManager` on shift-tab from the toolbar. 

Still figuring things out but that's a start!

mrtcode:
@abaevbog It's great if you can figure out how to do the necessary changes to the reader. But don't worry about that too much, as I can later tweak the code to fit into reader architecture if necessary.

abaevbog:
So I brought back the previous `Escape` behavior that alwaysy refocuses the reader and removed that idea with moving focus to the tab bar.

Then, the bulk of the logic to enable `Sync → toolbar → reader → Title` focus path boiled down to telling the reader when to send focus to an external element. I felt like I had to at least get an idea of how reader would handle this to make sure that whatever changes I suggest here actually make sense. I have a commit [here](https://github.com/abaevbog/reader/commit/cedca6b28ff9e4ab76f625ddd8c439f77d30beb0) that adds a few focus-related functions that are used here to specify what to do we're out of tab groups to move to. Without that change, this PR doesn't work right. With those changes, this is what it looks like now

https://github.com/zotero/zotero/assets/36271954/e88c53ae-2ebe-435d-a85d-dd1a1dcc5094

If that commit in general makes sense, I can make a PR to the reader repo with it. It is also very much possible that it's not the right way to approach it - in which case, @mrtcode you could just use it for reference of what I was going for.

I didn't move that `escape` keydown handler to non-capturing listener yet, just so that it's easier to compare the changes here.

Finally, as I was experimenting with the reader, I noticed a few things with focus that may not be working exactly as intended:
- if there are no annotations, the annotations group is not focusable for me, so tab after "Search annotations" input has no effect. It's not a big deal now that Tab from the scrollable area of the reader wraps focus to the toolbar, but if we have the `Sync → toolbar → reader → Title` flow, it's something that can be more of an inconvenience
- the selected annotation is marked with `.selected` class, and it doesn't go away when the focus leaves. If an annotation is selected, you can then click on the title, and then it looks like there are 2 focused elements. It can also be seen in the video.

If this makes sense, I'd be happy to help out with these as well, now that we're looking into all sorts of focus handling behaviors.

dstillman:
This looks great!

mrtcode:
I haven't tested, but the reader related code looks good!

> * if there are no annotations, the annotations group is not focusable for me, so tab after "Search annotations" input has no effect. It's not a big deal now that Tab from the scrollable area of the reader wraps focus to the toolbar, but if we have the `Sync → toolbar → reader → Title` flow, it's something that can be more of an inconvenience
> * the selected annotation is marked with `.selected` class, and it doesn't go away when the focus leaves. If an annotation is selected, you can then click on the title, and then it looks like there are 2 focused elements. It can also be seen in the video.

Thanks for mentioning those, I'll think what we can do here.

abaevbog:
> I haven't tested, but the reader related code looks good!

Cool! I added a PR to the reader repo with that commit - it's probably more convenient to test it that way

mrtcode:
> > I haven't tested, but the reader related code looks good!
> 
> Cool! I added a PR to the reader repo with that commit - it's probably more convenient to test it that way

Ok, I tested and it works as expected. Of course there is that regression, that you already mentioned, which happens when pressing Shift-Tab and the toolbar can't be focused if there aren't any annotations. But I can fix that after the merge.

Issue #3683, opened 2024-02-07
Alt key does not show menu access keys on Linux


adomasven:
I can get it to appear if I jam press Alt and mouse-click on the menus sometimes.

Why do we need custom handling there to begin with?

windingwind:
That idea came from @dstillman in issue 52 of the redesign.

> We should deal with this in another PR, but showing the main menu access key underlines is pretty distracting. LibreOffice does something kind of clever where it only shows them if you press an invalid Alt + key combination, and then it shows them whenever you press and hold Alt until you successfully trigger a menu, after which it stops showing them on Alt. For menu options, it shows them if you press down-arrow to move focus into one of the menus. So if you know (or guess) the access keys (usually by pressing the first letter of the menu or menu option), you never have to see the underlines. And if you don't, you see them when it's appropriate, without cluttering the UI the rest of the time. Can we do that? (Holding down Alt also triggers [collection highlighting](https://www.zotero.org/support/kb/collections_containing_an_item) (or at least it's supposed to — that's not working for me on Linux in this PR, but maybe we've fixed that elsewhere), but I think that's OK if we're only showing the underlines when necessary anyway.)

adomasven:
Ok, I thought Firefox supported that kind of behavior natively, but I guess not, and checking on Zotero 6 we always showed the access keys. Anyway, currently I cannot trigger the access key at all on Linux (aside from some janky clicking with the mouse).

adomasven:
Also it seems like Dan is mistaken, at least for LibreOffice, if you press Alt, you always get access key highlighting, regardless of whether you previously knew the correct key or not. I think that is both more reasonable and easier to implement behavior.

windingwind:
> if you press Alt, you always get access key highlighting, regardless of whether you previously knew the correct key or not

Yeah, that's what the implementation meant to do. I haven't visited this feature for a while (and sorry might not be able to in the next 2 weeks, since it's the Chinese New Year holiday and I am at home with my Mac and have no access to a Linux machine. I failed to run Zotero in a Linux VM on Mac, though, that's another topic).

What I can do now is to check if it works on Windows, and hopefully we can try some fixes on Windows and see if it works on your side.

windingwind:
OK just confirmed it works as expected on Windows (press alt and before the focus leaves the menus or press alt again, the access keys are always displayed and otherwise do not show access key by default.)

Does that logic work on Linux? Could you please share a screen record?

adomasven:
[...]

PR #3689, opened 2024-02-08
linux: collection highlight on Control instead of Alt

abaevbog:
On Linux, trigger collection highlighting on Control instead of Alt to not interfere with the menubar.

Addresses: #3683


abaevbog:
@adomasven, this is how I thought we can re-do the collection highlight logic to free up Alt on Linux for the access keys. Do Alt keypresses behave as expected for you with this change?

adomasven:
This is good, thanks.

PR #3692, opened 2024-02-09
Set max-height on note-editor Tags/Related and allow to scroll

### LINKED ISSUE #3686: Adding many tags to child note scrolls main window content
https://forums.zotero.org/discussion/111588/zotero-7-tag-pane-scrolling-entire-window

AbeJellinek:
Have to set an explicit `max-height` because of some weird `-moz-box`/`flex` interactions, as usual. Maybe we'll be able to do something better (e.g. `flex-grow: 2` on the note editor and `flex-grow: 1` on the Tags/Related container) once everything is flexbox!

Fixes #3686

### DIFF
--- a/scss/elements/_noteEditor.scss
+++ b/scss/elements/_noteEditor.scss
@@ -2,6 +2,8 @@ links-box {
 	display: flex;
 	flex-direction: column;
 	padding-inline: 8px;
+	max-height: 50vh;
+	overflow-y: auto;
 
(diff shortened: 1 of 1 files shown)


dstillman:
This still pushes Related off the screen, though:

Screenshot 2024-02-19 at 5 01 06 AM

Is there something reasonable we can do that only scrolls Tags?

PR #3695, opened 2024-02-10
Fix attachment column icon flickering upon selection

tnajdek:
This PR is a fix for a visual glitch where attachments column flickers when being selected for some items:

https://github.com/zotero/zotero/assets/214628/b50a8a32-0fdb-495f-8083-439ed33b39c6

This is a regression related to the attachment preview generation which calls [`fileExists`](https://github.com/zotero/zotero/blob/872e7b24278a2f00d7558aa9055d7f641c1a768a/chrome/content/zotero/elements/attachmentPreview.js#L208) which results in `bestAttachmentState` cache being cleared for this particular item before being checked again in `itemTree` on a 150ms timeout (hence the flicker).

While this is a very small change, I'd appreciate feedback if this is the right approach to fix the issue.

### DIFF
--- a/chrome/content/zotero/xpcom/data/item.js
+++ b/chrome/content/zotero/xpcom/data/item.js
@@ -2729,7 +2729,13 @@ Zotero.Item.prototype._updateAttachmentStates = function (exists) {
 		Zotero.logError(`Attachment parent ${this.libraryID}/${parentKey} doesn't exist`);
 		return;
 	}
-	item.clearBestAttachmentState();
+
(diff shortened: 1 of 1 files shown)


dstillman:
I don't think this is right. The thing that was being cleared here is the parent item's best-attachment state, but this attachment might not be the best attachment, so we don't want to just set its `exists` state as the best-attachment state.

Seems like we'd need to cache the itemID of the best attachment from the last run of `getBestAttachment()` and only update `item._bestAttachmentState.exists` if this attachment is the best attachment?

tnajdek:
Thanks for explaining, I've tweaked the code, any other suggestions please let me know.

PR #3702, opened 2024-02-13
Fix csl editor style in dark theme

windingwind:


chrome/content/zotero/tools/csledit.js
```diff
@@ -266,11 +267,17 @@ var Zotero_CSL_Editor = new function () {
 	};
 
 	this.updateIframe = function (content, containerClass = 'preview') {
+		const isDarkMQL = Zotero.getMainWindow()?.matchMedia('(prefers-color-scheme: dark)');
```
dstillman, line 270: Should we add a (dynamically computed) `Zotero.isDarkMode` and a `Zotero.getDarkModeMQL()` so we don't have to keep repeating this stuff everywhere?

@tnajdek

tnajdek: yes, sounds like a good idea!

windingwind: Does it make more sense to be `Zotero.UIProperties.isDarkMode`?

`Zotero.getDarkModeMQL()` might be unnecessary, since 1. If you use the MQL, you generally want to listen to color scheme change; 2. What you want to do on color scheme change is to change the appearance of a certain DOM element; 3. A DOM element belongs to a window, which means you actually should use the MQL from the parent window, instead the Zotero main window, which can be closed/recreated anytime

windingwind: Anyway, just added one as `Zotero.UIProperties.isDarkMode` to avoid creating another file under `xpcom`. If that's not OK we can move it somewhere else.

dstillman: I think `Zotero.isDarkMode` is fine — easier to remember, we have various other `Zotero.is*` properties (e.g., `Zotero.isMac`), and this is a pretty fundamental property.

For `getDarkModeMQL()`, I suppose it could take a window as an argument, but I guess that's only marginally less complicated than just remembering (or copying and pasting) `window.matchMedia('(prefers-color-scheme: dark)')`.

Not sure what you mean about another file under `xpcom`, though. What would require a new file?

windingwind: > Not sure what you mean about another file under xpcom, though. What would require a new file?

Not sure where will we put `Zotero.isDarkMode` if not in a separate file?

windingwind: > For getDarkModeMQL(), I suppose it could take a window as an argument, but I guess that's only marginally less complicated than just remembering (or copying and pasting) window.matchMedia('(prefers-color-scheme: dark)').

If so, it sounds like more a Zotero.Utilities.Internal thing - not a property/method of Zotero itself?

dstillman: OK, that's fair. But then yes, let's try not to add new instances of this pattern.

> Not sure where will we put Zotero.isDarkMode if not in a separate file?

It would've just been a dynamic property in zotero.js, but we can remove this.

windingwind: > we could dispatch a UIPropertiesChanged event from Zotero.UIProperties when prefers-color-scheme changes

How should we detect prefers-color-scheme change without relying on a specific window instance?

Issue #3392, opened 2023-09-01
Use pageLabel when creating zotero://open-pdf/ link for EPUB and snapshot attachments


mrtcode:
Note Markdown and Note HTML translators are trying use `position.pageIndex` which doesn't exist for EPUB and snapshot annotations, and results to `NaN`:

zotero://open-pdf/library/items/KXNUJUHE?page=**NaN**&amp;annotation=DBFZHNVR

This can be reproduced when copying selected EPUB and snapshot annotations and links are enabled in Quick Copy settings.

mrtcode:
https://forums.zotero.org/discussion/111718/zotero-7-bug-the-export-highlight-link-generates-page-nan-for-the-epub-file

pengxy0:
@mrtcode The annotated notes I made in snapshot, when I copy and paste the notes to Obsidian, show the same problem; I can't link back (my version is Zotero+54). I tried the latest Zotero+68, but it is possible to link back. However, version 68 is not very stable at the moment, so could you let me know how to fix this in version 54?
As it happens, the links for both version 54 and version 68 show "zotero://open-pdf/library/items/DAVXN67X?page=NaN&annotation=WP3Z842R".
![image](https://github.com/zotero/zotero/assets/130196652/02155d12-5dd8-4820-95f6-7010964fb246)

PR #3705, opened 2024-02-13
Unify attachment opening code and support EPUB/snapshot note links

AbeJellinek:
Plus plumbing:

- Add `attachmentReaderType` read-only property to `Zotero.Item` so we don't have to map MIME types to `pdf`/`epub`/`snapshot` everywhere
- Fix `launchFile()` fallback cases for fx102 - was throwing when there wasn't an application associated with the file type on macOS, because `loadUrl()` has been removed

Addresses #3392


mrtcode:
That's a lot of code, and from the basic overview and testing, it looks good. Though I haven't tested the updated Note HTML translator yet, it seems like it should do what it's supposed to. I'll test again once this is in beta.

PR #3707, opened 2024-02-14
Few fixes to `ftl-to-json` and `localize-ftl` scripts

### LINKED ISSUE #3700: ftl-to-json: Omit term-only strings from Transifex JSON
```
	"sidenav-abstract.tooltiptext": {
		"string": "{ pane-abstract }"
```

If possible, it would be nice to leave these out of the JSON, which would create pointless strings on Transifex for all locales, and just write them to the .ftl in `localize-ftl`. Or maybe we can just leave them out of non-English strings altogether and it will fall back to the definition in the English file?

tnajdek:
This PR:

1. Fixes how messages that reference other messages are handled
2. Messages that only reference other messages are now omitted from produced JSON. This required a tweak to `localize-ftl` so that it now also parses en-US `.ftl` file to re-introduce these into localized `.ftl`. (I hoped that fluent would fallback for `en-US` for a missing string, find that it references something and take the reference again from a localized file, bit unfortunately it does not). Fix #3700. 
3. Removes shenanigans with maintaining list of terms/variables etc. Previously I've tried to keep track of what is a term, what a variable etc. without actually storing that information in the produced JSON. That introduced overhead and complexity. Instead now we just keep the prefix in JSON (i.e. where previously we had `{ var }` in JSON, we now have `{ $var }`. This is, as far as I can tell, not legal in `ICU Messages` (hence my earlier approach) but I don't think it actually matters for Transifex (we've discussed this before: https://github.com/zotero/zotero/pull/3058#issuecomment-1500192606)

Happy to answer any questions.

### DIFF
--- a/js-build/ftl-to-json.mjs
+++ b/js-build/ftl-to-json.mjs
@@ -6,18 +6,29 @@ import { ftlFileBaseNames as sourceFileBaseNames } from './config.js';
 import { onError, onProgress, onSuccess } from './utils.js';
 
 const ROOT = join(dirname(fileURLToPath(import.meta.url)), '..');
+const TRANSIFEX_FILE_NAME = 'zotero.json';
 
 async function getJSON() {
 	const t1 = performance.now();
 	const sourceDir = join(ROOT, 'chrome', 'locale', 'en-US', 'zotero');
+	const destFile = join(sourceDir, TRANSIFEX_FILE_NAME);
+	let messagesMap = new Map();
+	
 	for (let sourceFileBaseName of sourceFileBaseNames) {
 		const sourceFile = join(sourceDir, sourceFileBaseName + '.ftl');
-		const destFile = join(sourceDir, sourceFileBaseName + '.json');
 		const ftl = await fs.readFile(sourceFile, 'utf8');
 		const json = ftlToJSON(ftl, { transformTerms: false, storeTermsInJSON: false, skipRefOnly: true });
-		await fs.outputJSON(destFile, json, { spaces: '\t' });
-		onProgress(destFile, destFile, 'json');
+		Object.entries(json).forEach(([key, value]) => {
+			if (messagesMap.has(key)) {
+				throw new Error(`Duplicate key: ${key} found in file ${sourceFileBaseName}.ftl`);
+			}
+			messagesMap.set(key, value);
+		});
+		onProgress(`${sourceFileBaseName}.ftl`, TRANSIFEX_FILE_NAME, 'ftl->json');
 	}
+	
+	const messagesJSON = Object.fromEntries(messagesMap);
+	await fs.outputJSON(destFile, messagesJSON, { spaces: '\t' });
 	const t2 = performance.now();
 	return ({
 		action: 'ftl->json',
--- a/js-build/ftl-to-json.mjs
+++ b/js-build/ftl-to-json.mjs
@@ -14,17 +14,17 @@ async function getJSON() {
 		const sourceFile = join(sourceDir, sourceFileBaseName + '.ftl');
 		const destFile = join(sourceDir, sourceFileBaseName + '.json');
 		const ftl = await fs.readFile(sourceFile, 'utf8');
-		const json = ftlToJSON(ftl, { transformTerms: false, storeTermsInJSON: false });
+		const json = ftlToJSON(ftl, { transformTerms: false, storeTermsInJSON: false, skipRefOnly: true });
 		await fs.outputJSON(destFile, json, { spaces: '\t' });
 		onProgress(destFile, destFile, 'json');
 	}
 	const t2 = performance.now();
-		return ({
-			action: 'ftl->json',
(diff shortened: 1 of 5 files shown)


dstillman:
```
attachment-info-convert-note =
    .label = Migrate to {
        $type ->
            [standalone] Standalone
            [child] Item
            *[unknown] New
    } Note
```

becomes

```
"attachment-info-convert-note.label": {
	"string": "Migrate to {$type, select, standalone {Standalone} child {Item} unknown {New}} Note"
},
```
This is stripping the `*` default indicator. I take it we're automatically choosing the last option as the default? That's probably fine — and I'm guessing what we do everyone — but we just need to remember that.

dstillman:
If I run `npm run ftl-to-json` and then `npm run localize-json`, and then run a `git diff` on the resulting `en-GB` `zotero.ftl`, all of the strings that only used terms (e.g., `menu-print`) have been moved to the bottom of the file. Not a big deal, but maybe we have an easy way to avoid that?

dstillman:
```
attachment-info-convert-note =
    .label = Migrate to {
        $type ->
            [standalone] Standalone
            [child] Item
            *[unknown] New
    } Note
    .tooltiptext = Adding notes to attachments is no longer supported, but you can edit this note
    by migrating it to a separate note.
```

becomes

```
	"attachment-info-convert-note.tooltiptext": {
		"string": "Adding notes to attachments is no longer supported, but you can edit this note\nby migrating it to a separate note."
	},
```

Is the `\n` in the JSON a bug here or in the .ftl? The second tooltip line is indented from the `attachment-info-convert-note =` line, but maybe it's supposed to be indented from the `.tooltiptext = ` line?

dstillman:
OK, it's a bug in our .ftl:

```json
[
    "0": {
        "attributes": [
            "0": {
                "name": "label"
                "value": "Migrate to New Note"
            }
            "1": {
                "name": "tooltiptext"
                "value": "Adding notes to attachments is no longer supported, but you can edit this note\nby migrating it to a separate note."
            }
        ]
        "value": null
    }
]
```

I'll fix.

dstillman:
Actually, maybe it just needs to be on a single line? https://projectfluent.org/fluent/guide/multiline.html talks a lot about whitespace and common indents, but it doesn't talk about newlines. So I guess newlines are just preserved?

tnajdek:
> This is stripping the `*` default indicator. I take it we're automatically choosing the last option as the default? That's probably fine — and I'm guessing what we do everywhere — but we just need to remember that.

That's correct, except there is a special treatment for the word "other": https://github.com/tnajdek/ftl-tx/blob/82ff7cefb4ce512f99c47a3cdad319175ebfefb9/test/recoverable.test.js#L59-L75 - I believe I've decided that because, according to CLDR plurals, you always should have at least that (not that it's relevant for `select` statement but both use the same logic)

> If I run npm run ftl-to-json and then npm run localize-json, and then run a git diff on the resulting en-GB zotero.ftl, all of the strings that only used terms (e.g., menu-print) have been moved to the bottom of the file. Not a big deal, but maybe we have an easy way to avoid that?

Should be possible, I'll have a look

> Maybe that'd be better, and then we can freely use individual .ftl files for our own organizational purposes but still have a single new resource on Transifex?

That might be possible. I believe individual messages would need to be prefixed in `.json` file or we will eventually hit a duplicate? It would also help map messages back to individual files. I can look into it.

dstillman:
> That's correct, except there is a special treatment for the word "other": https://github.com/tnajdek/ftl-tx/blob/82ff7cefb4ce512f99c47a3cdad319175ebfefb9/test/recoverable.test.js#L59-L75 - I believe I've decided that because, according to CLDR plurals, you always should have at least that (not that it's relevant for select statement but both use the same logic)

We had a case recently where [we decided didn't need `other`], but we are using the last option there as the default, so this should be OK.

> That might be possible. I believe individual messages would need to be prefixed in `.json` file or we will eventually hit a duplicate?

Mozilla still tends to prefix based on the component by convention, and think we'll generally do the same, so I think we could also just throw on a duplicate. But if prefixing, e.g., strings from `preferences.ftl` with `preferences-` makes mapping easier, I guess that's fine too. The actual key is not very visible on Transifex.

tnajdek:
* Script will now produce a single `zotero.json` file.
* If a duplicate key appears in `ftl` files, `ftl-to-json` now throws an error.
* Order of messages in produced `ftl` files should match order in matching `en-US` files, including for messages that only reference other messages.

dstillman:
I'm seeing some term-only strings in zotero.json:

```json
	"menu-print.label": {
		"string": "{ general-print }"
	},
```

```json
	"main-window-command.label": {
		"string": "{ -app-name }"
	},
```

I thought this was fixed, so maybe a regression?

tnajdek:
@dstillman is it possible you forgot to run `npm install` after switching branch?

dstillman:
Yes!

Issue #3701, opened 2024-02-12
Trim leading/trailing space in filename format string


dstillman:
https://forums.zotero.org/discussion/111759/zotero-7-beta-space-in-filename-between-the-name-and-the-extension

Trim in the prefs or when using or both?

tnajdek:
Done, not touching the input user provides but trimming spaces before template is actually used.

@dstillman While at it, do we also want to add "trim" (or similar) as a template parameter? i.e. when a value ends with a space, most likely due to truncation, it would also remove that space. I don't think we want this as default (someone might want to have a file that has exact length) but as an option it would make sense?

dstillman:
I feel like we could just do it by default, as long as it's documented? I'm not really sure why anyone would care about filenames of an exact length, and besides, the input could always be shorter, so you can't actually guarantee a given length anyway. (And it's not like we have a right-pad option.)

PR #3711, opened 2024-02-16
Trim spaces from values in getFileBaseNameFromItem

tnajdek:
Not sure if we only want to trim only the truncated value or any value. I opted for the latter but happy to change it.

I'll update the documentation once this is merged.

Related: [#3701 ](https://github.com/zotero/zotero/issues/3701#issuecomment-1948140718)


tnajdek:
I've added a sentence explaining this in the doc: https://www.zotero.org/support/file_renaming#variables

PR #3712, opened 2024-02-16
Fix attachment column sorting not always working

tnajdek:
This PR, together with #3645 (which introduced fallback attachment icon) hopefully makes attachment column more useful. However, this commit reduces the performance of the tree view.

Sorting by the `hasAttachment` column depends on having the `bestAttachmentState` determined for every item in the current view. Normally, this is not available for rows that are not currently visible, for performance reasons. There was a piece of logic that triggered `bestAttachmentState` logic for every item in the current view when the user initiated sorting by the `hasAttachment` column. Once this logic was executed, the tree set `_cachedBestAttachmentStates` to true and from that point onwards assumed that the `bestAttachmentState` was known for every item in the view.

However, to the best of my knowledge, this assumption is not always correct. Firstly, when the user navigates to another collection or library, the assumption may no longer hold true. Additionally, this method will not work after, for example, Zotero is restarted, in which case the sorting may have been configured to the `hasAttachment` column before the restart, but the `_cachedBestAttachmentStates` logic will not be triggered.

In this commit, I have moved that logic to the `sort` method. However, for a very large library, when many items with attachments are present in the current view, this might introduce some lag on less powerful hardware. This issue could perhaps be addressed by caching `bestAttachment` at the database layer?

Any suggestions welcome.

### DIFF
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -1045,6 +1045,8 @@ var ItemTree = class ItemTree extends LibraryTree {
 
 		this.selection.clearSelection();
 		this.selection.focused = 0;
+		this._cachedBestAttachmentStates = false;
+
 		await this.refresh();
 		if (Zotero.CollectionTreeCache.error) {
 			return this.setItemsPaneMessage(Zotero.getString('pane.items.loadError'));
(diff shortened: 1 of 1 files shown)


adomasven:
> Once this logic was executed, the tree set `_cachedBestAttachmentStates` to true and from that point onwards assumed that the `bestAttachmentState` was known for every item in the view.
> 
> However, to the best of my knowledge, this assumption is not always correct. Firstly, when the user navigates to another collection or library, the assumption may no longer hold true.

Yes, the current logic is evidently wrong. E.g. switch to a collection with a few items, sort by attachment, then move to the main library. Note that items are not sorted properly. `_cachedBestAttachmentStates` should be set to false in `ItemTree.changeCollectionTree()`.

The initial load of attachment states will be slow, but since they are cached in memory, further runs will be instantanious.

> Additionally, this method will not work after, for example, Zotero is restarted, in which case the sorting may have been configured to the `hasAttachment` column before the restart, but the `_cachedBestAttachmentStates` logic will not be triggered.

Yes, moving the caching logic to the `sort()` method fixes that.

> However, for a very large library, when many items with attachments are present in the current view, this might introduce some lag on less powerful hardware. This issue could perhaps be addressed by caching `bestAttachment` at the database layer?

Given that this was always the intended behavior, I don't think we should worry about it unless we start getting excessive user complaints.

tnajdek:
Thanks for reviewing! I've moved `_cachedBestAttachmentStates` reset to `changeCollectionTreeRow`.

tnajdek:
Looks like getting best attachment state for a lot of items can get really slow:

https://forums.zotero.org/discussion/114683/zotero-7-beta-long-delay-for-sorting-by-attachments

If I understand correctly, with this PR merged, this person would need to wait more than 3 minutes after restarting Zotero. Instead, currently, they get items sorted (kind of*) incorrectly. I believe that from the two the latter is more desirable.

*kind of because we don't actually show sorting chevron next to the attachments column, so it seems Zotero "forgets" sorting after restart which is even less serious bug.

PR #3719, opened 2024-02-20
Fix window titlebar background color

windingwind:
I think it's a regression by beb7b51


adomasven:
Could you include a before/after screenshot?

windingwind:
Before:

image


After:

image

PR #3727, opened 2024-02-21
Implement Windows redesign for native elements

windingwind:


windingwind:
| Before | After |
| ---- | ---- |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |

windingwind:
The checkbox/radio is not touched, since they look less ugly. We can do the rest part in separately to not make this PR too large.

adomasven:
Why are we increasing the size of menubars in windows that have their own titlebar? It makes sense in the main window where the titlebar and menubar is merged, but otherwise the menubars shouldn't take more vertical-space than default, not least because over 70% of users are on [screens that are smaller than 1080p](https://gs.statcounter.com/screen-resolution-stats/desktop/worldwide), but even those with larger resolutions use higher DPI, and as such we should use vertical space efficiently.

scss/components/_reader.scss
```diff
@@ -2,4 +2,11 @@ menupopup.tags-popup {
 	font: inherit;
 	min-width: 300px;
 	padding-inline: 8px;
+
+	@media (-moz-platform: windows) {
```
adomasven, line 6: I am conflicted about the use of `-moz-platform`. Generally, it's probably the more convenient way to do platform-specific styling and I would prefer it in place of generic style sheets with platform specific modifications in a single file. However, we are already doing platform-specific stylesheets. We should either decide on one or the other approach and stick to it, otherwise it becomes complicated and annoying finding platform-specific styling. I understand this gets somewhat even more muddled when copying styles from Mozilla's own stylesheets.

windingwind:
> Why are we increasing the size of menubars in windows that have their own titlebar?

The figma design doesn't include menubar, so I just borrowed the style from main window. I think we need suggestion from @yexingsha that how other menu bar should be like.

Definitely doesn't need to be that much height, but also shouldn't be using the default menubar style on Windows, which is from the win xp age.

scss/components/_reader.scss
windingwind: Yeah, here is just a quick fix for the tag box popup on Windows, which is never right since the redesign. Actually we should find out what makes the padding not work on Windows and do the fix there.

Why we use the `menupopup` to contain a `tags-box` CE, instead of a `panel`?

adomasven:
> Definitely doesn't need to be that much height, but also shouldn't be using the default menubar style on Windows, which is from the win xp age.

I generally loathe the argument that just because something is old it's bad, as well as Windows 11 and its changes in general both due to how they have reduced the performance of the OS dramatically, and redesigned perfectly functional UIs to be less so, but I understand I am swimming against the tide here.

I just booted up my Windows 11 VM and saw that Notepad (any other native apps with menubars?) now has new menubars that are more spaced out and look somewhat similar to what you have done here, so maybe my complaints are unjustified.

windingwind:
Just updated the menubar change to basicViewer and other windows. I think 30px or 24px might be good (currently 36px). We have a `--tab-min-height` and can change it easily.

Yeah, old doesn't mean bad. I miss the cute buttons of the Win XP and Win 2000, but they're just gone and no new device/software supports them. We're not making them look fashion, we just need to make them look consistent in style.

northword:
I think the current menubar is just the right height **if there is no titlelebar**, otherwise it should be reduced a bit.
(Personal opinion)

yexingsha:
I went through the native apps on my windows machine and another one with menu bar is MS Paint. Its menu bar is also tall, but it contains icon buttons as well. So maybe this is not too crazy?

scaffold-metadata

Other things that need tweaking:

1. The color of the title bar should be same as the tool bar or body below it
2. Use the simple red Z in the title bar, as we do in the main window
3. In style editor, align the right end of the two rows. This can be done by increasing the width of the language dropdown
4. Scaffold sidebar has the wrong color in dark mode
5. In settings/cite/style manager, the list should have 8px padding so that the rounded corners don't collide with the border. Also "Title" should be aligned with the titles below
6. In settings/general, the spacing between lines now seems too tight with the increased element height. Let's increase that to 8px

dstillman:
> Its menu bar is also tall, but it contains icon buttons as well.

I would rather not follow Microsoft in this regard — they'll stuff literally anything anywhere, and it makes for a totally incoherent mess.

yexingsha:
Then I think we can merge the title bar and the menu bar (windows UI always does this) and get rid of the top padding of the menu bar to squeeze out some height. The menu buttons are at 24px so they don't look too out of place:

scaffold-metadata

windingwind:
> The color of the title bar should be same as the tool bar or body below it
> Use the simple red Z in the title bar, as we do in the main window

This would need the window title bar to be none-native, like the main window. In that case, we can just merge the title bar and the menubar. Do you think it's a good idea?

yexingsha:
Huh, I didn't expect that, since it's exactly what windows native apps do. I guess it's a case of the body/toolbar color need to be changed to match the native title bar? Can we do that?

windingwind:
> I guess it's a case of the body/toolbar color need to be changed to match the native title bar? Can we do that?

Yes, and the window icon. We can't detect the native toolbar color.

windingwind:
Some more fixes (firefox's default theme in dark mode is a disaster!):

Before:
image

image


After:
image

image

windingwind:
Fixed:

image

image


> In style editor, align the right end of the two rows. This can be done by increasing the width of the language dropdown

The menu/buttons don't align since they fit to content. I think that's OK?

image

chrome/content/scaffold/load.js
```diff
@@ -37,18 +41,18 @@ var Scaffold_Load = new function() {
...
-		listbox.addEventListener('keypress', (e) => {
-			if (e.key == ' ' && e.target == listbox) {
-				e.preventDefault();
-			}
-		});
+		// listbox.addEventListener('keypress', (e) => {
+		// 	if (e.key == ' ' && e.target == listbox) {
+		// 		e.preventDefault();
+		// 	}
+		// });
```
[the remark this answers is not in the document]
windingwind: Oops! by accident

yexingsha:
> The menu/buttons don't align since they fit to content. I think that's OK?

When you stretch the window wider it does seem alright.

I do think we should make title bars match the body color, even if it's not exactly native. The slight color difference is just too weird.

In the after versions of [JavaScript Application] and Open Translator, the spacing between the input area and the buttons become too tight; should be 8px.

> This seems to affect macOS in a couple ways that we may or may not want

For thick focus rings (>=2px), the lighter color `var(--color-accent) 50%` is correct; 1px focus rings like this should indeed be `var(--color-accent)`. And Scaffold titlebar and toolbar should be unified.

Also, it seems that a lot of the headers are misaligned, e.g. in Settings/Cite/Style Manager, "Title" is aligned correctly but "Updated" is not

windingwind:
> I do think we should make title bars match the body color, even if it's not exactly native. The slight color difference is just too weird.

That would need the title bar to be non-native. I'll do

windingwind:
OK, besides the title bar, I think others are fixed

windingwind:
And all window and dialog are now by default assigned the corresponding color, as @AbeJellinek suggested

windingwind:
And there're changes to the v-table and item-tree style, which fixed the misalignment and first-column padding that had been there before/since the redesign @adomasven

windingwind:
image

dstillman:
> And there're changes to the v-table and item-tree style, which fixed the misalignment and first-column padding that had been there before/since the redesign

What do you mean by this? This seems to be adding a lot of padding to the first column of the items list that's not in Figma and that we definitely can't spare.

Before:

Screenshot 2024-02-27 at 5 44 21 AM

Now:

Screenshot 2024-02-27 at 5 44 07 AM

Figma:

Screenshot 2024-02-27 at 5 45 13 AM

chrome/content/scaffold/load.xhtml
```diff
@@ -1,47 +1,47 @@
...
+	
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU Affero General Public License for more details.
+	
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero.  If not, see <http://www.gnu.org/licenses/>.
+	
+	***** END LICENSE BLOCK *****
```
dstillman, line 24: Revert

(Always use `git add -p` and `git diff --cached` / `git show`!)

windingwind: Oh my bad. Done

dstillman:
This is making some subtle changes on macOS. In Scaffold:

Before:

Screenshot 2024-03-02 at 1 08 39 AM

After (less-tall textboxes?):

Screenshot 2024-03-02 at 1 06 50 AM

Doesn't really matter, but was that intentional? (Though it looks like the textboxes now have an extra pixel or two of padding on the left, which is an improvement…)

Before:

Screenshot 2024-03-02 at 1 14 23 AM

After (misaligned (or at least worse-aligned) Status column:

Screenshot 2024-03-02 at 1 14 19 AM

(But maybe the second one is just an [unrelated tree bug](https://forums.zotero.org/discussion/112427/zotero-7-alignment-of-column-header-and-values) that this is somehow triggering — I'm actually seeing *better* column alignment in the Cite pane in the prefs.)

scss/scaffold.scss
```diff
@@ -71,10 +87,6 @@ tab {
 	grid-template-columns: 1fr max-content;
 }
 
-#tabpanel-metadata input {
```
windingwind, line 74: @dstillman

This is where the input height are changed: we used to explicitly set the height, and now that we are setting all form element e.g. button and input's height by px on Windows, I removed this style to prevent adding too much platform-specific styles.

Or if we want this back, we can treat Windows as an exception here.

windingwind:
> misaligned (or at least worse-aligned) Status column

We are using the XUL richlistbox here, which is never correctly aligned and should be replaced by our virtualized table. We can do it in a separate PR after this one.

If you find the Prefs-Cite table correctly aligned, that proves the fix in this PR works. We use v-table in the Prefs window.

scss/scaffold.scss
dstillman: I think it was vaguely nicer before, but we could just add a little bit of padding below each row instead of changing the height of the textboxes themselves.

PR #3730, opened 2024-02-21
editable-text: Make dropping text work when field is unfocused

AbeJellinek:
Previously, dropping text on an unfocused item box field would make the text appear for a second at the drop position, then the field would focus and the added text would disappear. If you dragged the text from another item box field, it would disappear from _both_ fields. Pretty broken.

Now, dropping text works like Paste:

- If the field is focused when you drag text into it, e.g. when you drag text from another application, switch to Zotero, and drop into the focused field, it'll be added at the cursor position like normal.
- If the field is unfocused when you drag text into it, e.g. when you drag text from the focused field onto an unfocused field, or another application onto a field in the unfocused Zotero window, it'll replace the entire value of the field. We set the Copy cursor to make the difference clear. After drop, pressing Escape restores the previous value; pressing Enter commits.

So dropping onto an unfocused field works like Paste on an unfocused field (overwrites whole contents), and dropping onto a focused field works like Paste on a focused field (adds at cursor).

Also includes `initialValue` fixes that were necessary to allow reverting a drop onto an unfocused window.

### DIFF
--- a/chrome/content/zotero/elements/abstractBox.js
+++ b/chrome/content/zotero/elements/abstractBox.js
@@ -114,9 +114,13 @@
 
 			let abstract = this.item.getField('abstractNote');
 			this._section.summary = abstract;
-			if (!this._abstractField.initialValue || this._abstractField.initialValue !== abstract) {
+			// If focused, update the value that will be restored on Escape;
+			// otherwise, update the displayed value
+			if (this._abstractField.focused) {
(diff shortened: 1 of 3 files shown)


dstillman:
1. Click in Title
2. Press Esc

Title clears and shows "Title" placeholder. Click away and back and the original title is back.

dstillman:
Neat!

Issue #3732, opened 2024-02-22
Drag handle appears on single creator after context menu has been opened


AbeJellinek:
1. Create an item
2. Add one creator
3. Hover over it - no drag handle
4. Right-click the name and close the context menu
5. Hover over it again - drag handle is shown now

abaevbog:
I can see it too. Thank you!

PR #3735, opened 2024-02-22
itemBox: ensure the grippy is hidden if <2 authors

abaevbog:
When contextmenu closes, it clears visibility style from all components. It is added temporarily so that the options buttons do not disappear if the mouse leaves the row. The visibility setting for the grippy for items with only one author should be in a separate class so that grippy is not displayed by accident.

Fixes: #3732


abaevbog:
That makes sense! I re-pushed to use `.meta-label > .single-creator-grippy`

PR #3742, opened 2024-02-23
Fully block all HTTP requests in the reader

AbeJellinek:
So it's still not completely clear to me what was going on here.

When using the reader in Firefox/Chrome/Safari, everything is fine. Remote requests are fully and immediately blocked due to CSP. No data goes out and no data comes in.

In the client, though, things are weird. When a snapshot includes a remote resource - `<iframe>`, ``, etc. - it's as if the CSP kicks in late. The request goes out, the server starts sending a response and the response headers get read, *then* Firefox remembers that it was supposed to be paying attention to the CSP, and it cuts off the response. Future resources added after the page is done loading (e.g. you go into the console and programmatically append an `https://example.com/` iframe) are fully blocked - no requests at all.

I'm not sure why that's happening, and why only in the client. I did a bunch of trawling through Searchfox, Firefox bugs, etc., and didn't find anything relevant. So I just adapted the remote request blocking observer that we've been using in `HiddenBrowser.jsm` to catch anything that gets through the CSP. I wish could find the root cause, but this fix does the trick for now.

(I applied the blocking observer to all reader types - couldn't think of any reason why the PDF reader would need to make an HTTP request, and couldn't find anywhere that it does. @mrtcode, let me know if that's right.)

### DIFF
--- a/chrome/content/zotero/BlockingObserver.jsm
+++ b/chrome/content/zotero/BlockingObserver.jsm
@@ -1,0 +1,75 @@
+/*
+    ***** BEGIN LICENSE BLOCK *****
+    
+    Copyright © 2024 Corporation for Digital Scholarship
+                     Vienna, Virginia, USA
+                     https://www.zotero.org
+    
+    This file is part of Zotero.
+    
+    Zotero is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Affero General Public License as published by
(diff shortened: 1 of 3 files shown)


mrtcode:
You are right, PDF view doesn't need to make any HTTP requests. Blocking everything sounds good.

PR #3747, opened 2024-02-27
Move Locate to sidenav

AbeJellinek:


dstillman:
OK, we lost the Locate menu with https://github.com/zotero/zotero/pull/3791, so we'll need to figure out what we're doing here. Current feelings? The additional B/W icon is bothering me a bit less than it was before, though I still wonder if it's necessary. The main options are available either on double-click or right-click, though obviously there's the traditional Apple school of UI design where we assume no one knows how to right-click. It's also not clear how much the lookup engines are actually used — I suspect Library Lookup and Google Scholar Search are sometimes used and Crossref is never used (and probably should be removed, since it seems to variably be either totally broken or equivalent to launching the DOI?). But we arguably could do more with this menu.

We could also buy some time by adding the lookup-engine options to a submenu in the context menu, which maybe we need to do anyway for accessibility reasons.

dstillman:
Merging for now so we can release another beta.

We technically support other lookup engines — which are OpenSearch engines — by [modifying engines.json](https://www.zotero.org/support/locate#managing_lookup_engines). (`Components.interfaces.nsISearchEngine.TYPE_OPENSEARCH` became `undefined` sometime between Firefox 31 and 45, but I think the code still works.) Arguably we should do more with this, not less.

I really don't want to do the Finder-style full-context-menu "(⋯)" menu. (To be clear, I am not actually concerned with right-click discoverability, as it is not 1985.) I do think there's a distinction from the other context-menu options, which are in-app operations that either are or should be duplicated by toolbar buttons and/or File menu options, whereas this is the primary open operation plus stuff that takes you to the web and doesn't appear elsewhere. (Show File should probably be moved out of here and only exist in the context menu and File menu.)

> "Library Catalog: XYZ University"?

I'd worry about menu width there.

Issue #3749, opened 2024-02-27
Quick Format: Backspace after bubble should delete previous bubble


dstillman:
Follow-up to #3464

If you the cursor is positioned immediately after a bubble (e.g., because you just accepted a search result to create a bubble) and you press Backspace, the previous bubble gets selected instead of deleted. You have to press Backspace a second time to delete the bubble. If there's no text to delete, I don't think there's any reason not to just delete the preceding bubble.

PR #3753, opened 2024-02-27
quick-format: post-accessibility update edits

abaevbog:
- Backspace from the very beginning of the input will delete the previous bubble, as opposed to removing the input itself. If it makes it so that two inputs are next to each other, they are merged.
- When a new input is created via a mouse click, and the current input is empty, manually delete it right away. Waiting for the blur handler to remove it makes it so that the bubbles shift back and forth for a moment when a new input is added and the old input is not removed.
- Minor tweak to make sure that even if there's a 'br' before an input, clicking before it will not make a new input and just refocus it.

Fixes: #3749

### DIFF
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -372,6 +372,10 @@ var Zotero_QuickFormat = new function () {
 				let secondInputValue = node.nextElementSibling.value;
 				node.value += ` ${secondInputValue}`;
 				node.dispatchEvent(new Event('input', { bubbles: true }));
+				// Make sure focus is not lost when two inputs are combined
+				if (node.nextElementSibling == document.activeElement) {
+					node.focus();
+				}
 				node.nextElementSibling.remove();
 			}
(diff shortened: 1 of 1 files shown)


dstillman:
I'm now seeing a weird thing if I add a bubble and then press Backspace:

Screenshot 2024-02-27 at 6 00 48 PM

dstillman:
And if I type something to get the full search results, accept that, and press Backspace:

Screenshot 2024-02-27 at 6 01 47 PM

abaevbog:
That is ... odd. The reference panel should be explicitly cleared when an input becomes a bubble. 

I'm still trying to reproduce this and no luck yet. Is the reference selected with a mouse click or with an Enter? There are no other bubbles, this is a completely fresh dialog? And this is on a mac, right?

abaevbog:
I have a guess that this could be caused by me reopening the panel when all is deleted but not re-runing the search (which also takes care of the panel resizing).

I just repushed with a small tweak to actually redo the search when all is deleted - does that fix it for you?

PR #3755, opened 2024-02-27
Locate menu: Don't show "New Tab" when external, fix bugs

### LINKED ISSUE #3741: Don't show "Open [File] in New Tab" in context menu if set to open file in external program
Just "Open PDF", etc.

AbeJellinek:
- Run label update logic once instead of repeating on every item
- Show plural attachment type when multiple attachments of same type are selected (instead of "Attachments")
- Fix incorrect icon and label showing when `openReaderInNewWindow` = true
- Fix alternate-behavior item hiding for all item types when PDFs are set to open externally (and never hiding when other types are set to open externally)

Fixes #3741


dstillman:
Now seeing this:

Screenshot 2024-02-27 at 6 04 10 PM

PR #3757, opened 2024-02-27
Use HTTPS when opening DOIs via Locate menu and double-click

AbeJellinek:
Not sure if there's a reason we stuck with http:// for this in the past (site didn't support HTTPS?), but it seems to work fine now.

### DIFF
--- a/chrome/content/zotero/locateMenu.js
+++ b/chrome/content/zotero/locateMenu.js
@@ -475,7 +475,7 @@ var Zotero_LocateMenu = new function() {
 			if(doi && typeof doi === "string") {
 				doi = Zotero.Utilities.cleanDOI(doi);
 				if(doi) {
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/locateMenu.js
```diff
@@ -475,7 +475,7 @@ var Zotero_LocateMenu = new function() {
 			if(doi && typeof doi === "string") {
 				doi = Zotero.Utilities.cleanDOI(doi);
 				if(doi) {
-					return "http://dx.doi.org/" + encodeURIComponent(doi);
+					return "https://dx.doi.org/" + encodeURIComponent(doi);
```
dstillman, line 478: Should just be `https://doi.org/` these days

(The HTTPS actually makes no difference, because dx.doi.org is in the HSTS preload list.

Issue #3764, opened 2024-02-28
Quick Format: Selected/Item headers aren't announced to screen readers


dstillman:
https://forums.zotero.org/discussion/comment/456940/#Comment_456940

PR #3772, opened 2024-02-29
quick-format dialog: edits to aria item-description

abaevbog:
- include a sentence that contains which section the item belongs to (e.g. selected/cited/my library) before the keypresses hint of the item description. The section is updated whenever an item is focused.
- if no item from the ref panel is focused, the section is empty and no item-description is announced. Since the panel is a popup, NVDA and JAWS sometimes begin reading out all items from the reference panel as soon as it appears. In that case, keypress hints do not make sense, so it's best to just skip the aria-description.

Fixes: #3764


abaevbog:
The last commit addresses another glitch brought up here: https://forums.zotero.org/discussion/112377/z7-beta-adding-page-numbers-in-new-citation-picker-after-deleting-second-citation where deleting a bubble would not clear the `locatorNode` and typing `p. X` or `: X`  would not have any effect. 

Second point from that post: the effect of typing just numbers, without "p." or ":", adds a page number to bubbles only after an input was turned into a bubble - then, `locatorNode` is cleared. It's so that one could type numbers after a bubble that are meant to be a search term.

chrome/locale/en-US/zotero/zotero.ftl
```diff
@@ -436,4 +436,7 @@ toggle-preview =
 
 quickformat-aria-bubble = Press Down Arrow to open citation properties.
 quickformat-aria-input = Type your search term. Use Down Arrow and Up Arrow to navigate the search results.
-quickformat-aria-item = Press { return-or-enter } to select this item. Press Tab to return to the search field.
+quickformat-aria-item = { $section ->
+    [none] { "" }
+    *[other] This item belongs to { $section } section. Press { return-or-enter } to select this item. Press Tab to return to the search field.
```
dstillman, line 441: I haven't been able to try this in VoiceOver yet, but do you have to press something to get this announced? Is `This item belongs to { $section } section.` a common sort of announcement? If so, how about just `This item belongs to { $section }.`? E.g., "This item belongs to Selected Items." or "This item belongs to My Library.".

Issue #3778, opened 2024-03-01
Crash clicking on child attachment in trash


dstillman:
> this._item is null   attachmentsBox.js:82

> An error has occurred. Please restart Zotero.

I deleted a child PDF attachment (with a sibling link attachment), switched to the trash, and clicked on that attachment. Let me know if you can't reproduce and I can try to debug more.

windingwind:
Can confirm it's fixed in the current beta.

PR #3780, opened 2024-03-01
Fix attachmetsBox inTrash?

windingwind:


windingwind:
Might fix #3778 (or not); I cannot reproduce it on my side, but since we have a more clean way of setting data to item pane boxes, like the `inTrash`, in #3624, I think we can do a temp fix here and revisit if it still happens later.

PR #3791, opened 2024-03-04
Item pane header customization

### LINKED ISSUE #3614: Title and Abstract aren't included in merge dialog
https://forums.zotero.org/discussion/111120/zotero-7-abstract-and-title-not-part-of-merge-dialog

### LINKED ISSUE #3746: Item pane header customization
https://github.com/zotero/zotero/pull/3668#issuecomment-1965406244

New "Item Pane Header" (?) drop-down list in Interface section of prefs with options "Title", "Title/Creator/Year", and "Citation". If you choose "Citation", a new indented drop-down appears with the list of installed non-numeric citation styles.

- Not sure if we should default to Title or Title/Creator/Year. We'll try them out.
- The style is persisted even if you switch back to a different mode.
- If the selected style doesn't exist at generation time, we default to Title/Creator/Year.
- We should cache the citation output on the item, and clear it when the item is modified

@yexingsha, want to mock up how these three modes would look?

We do have to decide a couple things:

1. Do we always show Title in Info now? Part of the impetus for #3668 was that people weren't finding Title, and just logically it does belong in Info, so probably we do want it there.
2. If you're in Title mode, does Title remain editable in the header? Is there any reason it shouldn't? Technically the title part could be editable even in Title/Creator/Year. Our editable-text component is considerably closer to a regular label now, and Title is the thing that even people unconcerned with citations (who now may want to leave Info closed) would want to edit, so I kind of like the idea of leaving it editable where we can.

AbeJellinek:
Implemented:

- None, Title, and Title/Creator/Year modes
- Citation mode
	- Item-level caching and invalidation when item or style changes
	- Support for numeric styles - numbering removed automatically
	- Links in citations open in browser
- Settings, under General -> Item Pane Header
- Putting title back in Info pane, apply text transforms to `shortTitle` too
- Limit the whole thing to 1/4 of pane height
- Fix Merge layout

---

Thoughts so far:

- I like Title/Creator/Year mode! It's nifty and the creator/year don't take up much space - it actually seems to me like it could be an equally good default.
- I was initially pretty worried because Citation mode seemed *very* slow with some citation styles, even on fast hardware, but caching the `CSL.Engine` made it much better. We should still do more testing on very resource-constrained machines, though.

Fixes #3746, fixes #3614

### DIFF
--- a/chrome/content/zotero/elements/paneHeader.js
+++ b/chrome/content/zotero/elements/paneHeader.js
@@ -240,7 +240,6 @@
 				if (this._titleFieldID) {
 					this.titleField.placeholder = Zotero.ItemFields.getLocalizedString(this._titleFieldID);
 				}
-				this.titleField.toggleAttribute('tight', headerMode === 'titleCreatorYear');
 				this.title.hidden = false;
 			}
 			
@@ -254,9 +253,19 @@
 				if (year) {
 					creatorYearString += ` (${year})`;
 				}
-				this.creatorYear.textContent = creatorYearString;
-				this.creatorYear.hidden = false;
+				
+				if (creatorYearString) {
+					this.creatorYear.textContent = creatorYearString;
+					this.creatorYear.hidden = false;
+				}
+				else {
+					this.creatorYear.hidden = true;
+				}
 			}
+
+			// Make title field padding tighter if creator/year is visible below it
+			this.titleField.toggleAttribute('tight',
+				headerMode === 'titleCreatorYear' && !this.creatorYear.hidden);
 		}
 		
 		_renderBibEntry() {


dstillman:
Haven't tried yet, but some quick thoughts:

> Right now we're letting CSL set the line-height in Citation mode, so it'll vary by citation style. Not sure if style-specific spacing is that important or if we should just set something constant.

We can definitely ignore line-height from the style. This is purely about the text (including any rich-text formatting).

> It would be seriously cool if Citation mode let you edit the citation and have your changes apply when citing the item

I'm not sure about that — when people do this, it's usually because they don't know any better and they really just need to fix the metadata. But making the *metadata* editable via the bibliography entry might actually be seriously cool. Not sure how reliably we could do that, though, unless there's a citeproc-js mode to mark up individual elements in separate identifiable HTML divs. And there'd be some obvious problems: creators, missing fields… Maybe no point in trying to replicate the info pane.

> Should we make citations copyable? (Right-click -> Copy?) That would be really convenient for people who don't use word processor plugins, but I have no idea if we want to make that workflow easier.

I suppose. You can already do it with Quick Copy, and this would let you set Quick Copy to an export format and use this for a bibliography entry.

> Should we rename Citation mode to Bibliography mode?

I was thinking about this initially but decided that while "citation" vs. "bibliography entry" or "citation" vs. "reference" is an important distinction in a citation manager (or…reference manager), "citation" could still refer to the general concept (and you are choosing a "citation style" here). But we do use "Citation" and "Bibliography" elsewhere in the software, so maybe we want to avoid confusing things. "Bibliography Entry"?

dstillman:
"View As" is good.

1. We don't need an ellipsis in "More Options" — that obviously needs to pop something up, so there's no worry that it's going to do something without warning.
2. More Options doesn't take me to General if I'm not already in that
3. After I choose a citation style, I can no longer interact with the drop-down to choose a different style
4. This is minor, but when viewing an item without a creator, there's extra padding below the title when in Title, Creator, Year mode, even though nothing is shown below the title. I think ideally that would be the same as Title mode.

AbeJellinek:
@sviatoplok: Please post on the forums with more details about what you expect to see and what you're currently seeing.

sviatoplok:
@AbeJellinek : Currently, the style preview displays the full citation:
grafik
In the styles I am using, there are also short citations. In the case of the entry chosen, I would expect the follwoing to be displayed under (or next to the) full citation:
Schmied-Kowarzik, Austria-Hungary.

AbeJellinek:
@sviatoplok: Please [post on the forums](https://forums.zotero.org/)

Issue #3782, opened 2024-03-02
Ctrl collection highlighting causing problems on Linux (and Windows?)


dstillman:
https://forums.zotero.org/discussion/112327/zotero-7-beta-conflict-with-the-ctrl-key-on-linux

Interfering with Ctrl+PgDn/PgUp is definitely a problem, but I'm confused about the other. Has this always been happening on Windows (where Ctrl has been used for collection highlighting)? Has it been highlighting and possibly scrolling the collections pane any time you used a shortcut key? For some reason collection highlighting isn't working for me at all on Windows, so I can't test this.

abaevbog:
> Interfering with Ctrl+PgDn/PgUp is definitely a problem

Looks like this is an edge case that I missed while testing if there are any issues with making the `ZoteroPane.handleKeyDown` a non-capturing listener to avoid all sorts of random issues with focus https://github.com/zotero/zotero/commit/5f0e3d550bf4fabefa8d6a1392d24018021c7f69. It looks like  Ctrl+PgDn/PgUp must be capturing because if the focus is on the itemTree in the library tab, it ends up being picked up by the scrollable area of itemTree and never reaches the handler that changes the selected tab. 


> Has it been highlighting and possibly scrolling the collections pane any time you used a shortcut key?

This is a regression from 527fa5b12f33e7b79e77e65ad12c497d293768c9 - I misunderstood the purpose of the removed conditional at the bottom (partly because of the `// Unhighlight on key up` comment). That part is actually importabt because it's responsible for stopping the highlighting from kicking in if something else was pressed.

abaevbog:
So we certainly want to bring back cancelation of collection highlighting if any other control character is pressed (just with a better comment), and I thought we can just have 2 `keydown` listeners: a bubbling and a capturing one? Sometimes we want to make sure the events are not handled by any other components (Ctrl+PgDn/PgUp) but as a general rule, we want to only handle events if no other listener took care of them (e.g. components should have the option to handle Escape if needed when reader is opened).

adomasven:
@abaevbog I haven't looked at the details of what's causing issues here, but it seems that you may be somewhat confused about JS event handlers.

[...]

PR #3793, opened 2024-03-04
Shortcuts fixes

abaevbog:
- Bring back un-highlighting of collections on keydown if any other control key besides Ctrl/Options is pressed (with better comments). It ensures that if one presses, for example, Ctrl-Shift-A to copy an item, it will stop collection highlighting and not scroll collection tree.
- To decrease the chance of collection highlighting being activated when it is not intended, only highlight collections when the itemTree is focused

- Add a keydown listener in ZoteroPane that captures events, move all tab-related shortcuts handling there. We generally want shortcuts to be handled by the non-capturing listener to avoid a conflict with lower level component that may want to handle the event (e.g. we want to handle Escape keypress only if nothing else took care of it). Tab navigation is an exception, since we want it to happen all the time. For example, Ctrl+PageUp can be handled by itemTree but we want to override it to make sure that it selects the tab even if the focus is on the itemTree.

Fixes: #3782


chrome/content/zotero/zoteroPane.js
```diff
@@ -1016,6 +993,38 @@ var ZoteroPane = new function()
 					return;
 			}
 		}
+	}
+	
+	/*
+	 * Trigger actions based on keyboard shortcuts (non-capturing). This should be the default
```
adomasven, line 999: The "capturing" event handler is really poorly named by the designers of JS, given that "we capture an event" colloquially in any non-JS context would be understood as not allowing the event to propagate further. The comment for captureKeyDown() says "A capturing listener to handle", so this should say "A bubbling listener to handle..." and explain the use case - should be used when a lower-level element may override the navigation handling by stopping event propagation.

abaevbog: That makes sense, thank you. I re-pushed with a new comment and an example to make it clearer:

	 * Bubbling listener for navigation or shortcuts keydown events that should be
	 * handled only if no lower-level element overrode it by stopping event propagation.
	 * E.g. Escape when reader is opened refocuses the scrollable area of the reader. This should
	 * not happen if Escape was pressed when a menupopup is opened - just let menupopup handle the
	 * Escape and close the popup.

PR #3805, opened 2024-03-07
Backport Lithuanian translations

embar-:
Backport Lithuanian translations from Transifex with few editing for Zotero 6


dstillman:
Thanks, but we handle merging from Transifex in an automated manner.

dstillman:
Oh, and we're not making further changes for Zotero 6.

embar-:
Transifex content is between Zotero 6 and 7 (neither 6, neither main), and few lines from Zotero 6 had nothing corresponding in Transifex - this is why I did commit directly to Github.

PR #3807, opened 2024-03-07
Show first-run banner on upgrade

### LINKED ISSUE #3698: First-run dialog for Z6 upgraders linking to Zotero 7 blog post?
https://forums.zotero.org/discussion/comment/454869/#Comment_454869

We'd have to decide exactly what this would look like.

AbeJellinek:
And move banner styles to SCSS.

Need a URL that the new features link should open!

Closes #3698

### DIFF
--- a/chrome/content/zotero/preferences/preferences.xhtml
+++ b/chrome/content/zotero/preferences/preferences.xhtml
@@ -35,7 +35,6 @@
 <?xml-stylesheet href="chrome://zotero/skin/zotero.css"?>
 <?xml-stylesheet href="chrome://zotero-platform/content/overlay.css"?>
 <?xml-stylesheet href="chrome://zotero-platform-version/content/style.css"?>
-<?xml-stylesheet href="chrome://zotero-platform/content/zotero.css"?>
 
 <!--
 	To add an observer for a preference change, add an appropriate case in
--- a/chrome/content/zotero-platform/mac/preferences.css
+++ b/chrome/content/zotero-platform/mac/preferences.css
@@ -21,5 +21,6 @@ groupbox > label > h2, groupbox > * > label > h2, caption {
 
 .help-button {
 	appearance: auto;
-	-moz-default-appearance: -moz-mac-help-button;	
+	-moz-default-appearance: -moz-mac-help-button;
+	min-width: 0;
 }
--- a/chrome/content/zotero/locateManager.xhtml
+++ b/chrome/content/zotero/locateManager.xhtml
@@ -59,7 +59,7 @@
 			
 			<separator class="thin"/>
 			
-			<hbox align="center">
+			<hbox id="button-bar" align="center">
 				<hbox pack="start" flex="1">
 					<button label="Toggle" onclick="toggleLocateEngines()" flex="1"/>
 					<button id="locateManager-restoreDefaults" label="&zotero.preferences.locate.restoreDefaults;" oncommand="restoreDefaultLocateEngines()" flex="1"/>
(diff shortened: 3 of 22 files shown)


dstillman:
I guess we can link to `https://www.zotero.org/blog/zotero-7/` and make this not appear if `Zotero.isBetaBuild`, and then we can remove that once we launch and the page exists.

dstillman:
Screenshot 2024-07-19 at 3 31 24 AM

- Too much top and bottom padding
- Too much space after the colon
- Not enough space between the radio buttons and the labels
- The radio buttons themselves are out of alignment. It seems like they're top-aligned, so that the selected one (which has a bigger diameter) goes lower?

Also, Comfortable should probably be first, since (if I recall?) we're no longer setting existing users to Compact.

CC @yexingsha

yexingsha:
The larger padding is caused by the margins of the radios and buttons. macOS native UI elements being off seems to be a recurring problem after fx115, so I wonder whether we can fix them once and for all instead of doing one-off tweaks everywhere (there have been so many radio and button issues in various dialogs). This gets radios and buttons as close to macOS native as possible, but will mess with the radios and buttons already tweaked elsewhere:

```
button {
	@media (-moz-platform: macos) {
		margin: 0 -2px -1px;
	}
}

.button-text {
	@media (-moz-platform: macos) {
		margin-block: 1px 0;
		margin-inline: 3px;
	}
}

radiogroup {
	@media (-moz-platform: macos) {
    	margin: 0;
		gap: 8px;
	}
}

radio {
	@media (-moz-platform: macos) {
		margin: 0 0 0 -2px;
		gap: 6px;
	}
}

.radio-check, .radio-icon, .radio-label {
	@media (-moz-platform: macos) {
    	margin: 0;
	}
}
```

But even with all this, the second radio button is always 1px lower than the first (not the labels, they are aligned). It doesn't seem to have anything to do with their selected state, and I can't figure out why.

Another issue is that `position: absolute` causes the buttons to overlap with the radio buttons when the window is narrow. It also makes the banner too thin when the radio and button margins are fixed.

yexingsha:
I think the centering the banner in the space left of the buttons looks fine, but when the banner wraps into multiple lines, it would make more sense for the buttons to be on a different line.

Need to get rid of `#post-upgrade-buttons {button {margin: 0;}}` since it's overriding the global button style.

As for prefs, I can still see a lot of weird things:

1. Buttons seem to have too little vertical margins in some places. The "Choose..." and "Show Data Directory" in your screenshot are good examples.
2. (I assume) checkboxes need similar global style modification as radio buttons, since there is too little spacing between the checkboxes and their labels.
3. Using inline margins for buttons does align them with most text labels, but some elements are sticking out at the left edge: radio buttons, dropdown buttons, some descriptions, links, and tables
4. The text within dropdown buttons are positioned too low.
5. The help button is not aligned to the right edge of content.
6. In Sync/Reset, the "Show Reset Options" button is using a larger font size than regular buttons.

yexingsha:
Looks good, just a couple more changes:

- The modified checkboxes seem to be 1px too high compared to their labels.

- With the removal of margin-inline-start, the buttons with their margin-inline become indented. `h1` is also indented right now.

yexingsha:
There's some missing spacing between the -/+ buttons in Export and Cite, and the three buttons in Advanced/File and Folders/Database Maintenance. Otherwise everything looks good.

yexingsha:
Everything looks fine to me — the prefs look much better now!

dstillman:
Merged

chrome/content/zotero/preferences/preferences.xhtml
```diff
@@ -35,7 +35,6 @@
 <?xml-stylesheet href="chrome://zotero/skin/zotero.css"?>
 <?xml-stylesheet href="chrome://zotero-platform/content/overlay.css"?>
 <?xml-stylesheet href="chrome://zotero-platform-version/content/style.css"?>
-<?xml-stylesheet href="chrome://zotero-platform/content/zotero.css"?>
```
windingwind, line 38: Hi @AbeJellinek , this removes the styles we apply to form elements (button, menuitem, etc.) on Windows, which I think is necessary. Do you think I should create a PR to include those again?


It's currently:

image

Issue #3811, opened 2024-03-08
Selected/Open item is added via Enter when editing existing citation


dstillman:
https://forums.zotero.org/discussion/112627/unwanted-item-is-added-via-enter-when-editing-existing-citation

This is tricky. It might be the kind of thing where the best behavior is subtle and hard to explain but ends up generally doing what you want when you use it — like, say, if you open via edit mode instead of add mode and you open a bubble and then close it, the selection stays on the bubble instead of moving to the right of it, so you can just press Enter twice after typing the page number. I suspect people in edit mode are much more likely to be adding page numbers than additional items.

But that wouldn't solve all problems, so I think that, regardless of any other changes, we need to add an "accept" button — like the right-arrow in the Firefox address/search bars — on the right side of the text field when there's any input. So you could always click on that or right-arrow over to that and press Enter. Left-arrow onto a bubble + Enter also works, but that's much less obvious and isn't a real solution.

adomasven:
The intended behavior is not displaying the reference list until anything (including a space) is typed, so this is just behaving wrong. On the other hand, we also have this https://github.com/zotero/zotero/issues/3671 - always leaving the reference list open, but not focused, which would produce the same results, but also allow to navigate into the reference list via arrow-key down.

Regardless, the citation dialog should have an accept button, but I don't think it should be navigable to via arrow keys.

dstillman:
I don't think space bar makes sense as any sort of expected trigger — that's not something anyone would know to do.

But yes, I guess #3671 is the right fix here (along with an accept button). In an empty dialog, we focus the results so that Enter works, but once there's a bubble (whether in add or edit mode), we keep the results open if there are still any to show but don't focus it, and you have to use down-arrow first to focus it.

abaevbog:
I was always a bit confused by space as a trigger to reopen the references panel, so keeping the panel opened makes sense a lot of sense to me. I can tweak some of my earlier changes in the references panel logic for it to remain opened.

As a recap of what we are after (correct me if I'm mistaken):
- empty dialog opens => reference panel is opened and the first item is selected, so Enter will add the first item as a bubble
- dialog has bubbles => the reference panel is opened as long as there are selected/open items to show but no item is selected, so Enter will accept the current dialog's state
- after some query was typed into the editor and the reference panel's results refreshed, should the first item be selected then? I think it's reasonable to expect that after you type something and see the first matching result you would want Enter to add it?



[...]

Issue #3671, opened 2024-02-02
Quick Format: Leave results window open if more selected/open items


dstillman:
https://forums.zotero.org/discussion/comment/454179/#Comment_454179

But after an item is selected, remove the automatic selection of the next item until space or down-arrow is pressed so that Enter still closes the dialog

dstillman:
This actually doesn't seem to be fixed. I have multiple tabs open, but after choosing one I still have to press space bar to see the search results again.

dstillman:
Sorry, I was testing the wrong version! It's fixed!

PR #3816, opened 2024-03-08
Quick-format: do not hide reference panel

abaevbog:
- Reference panel remains opened while the focus is on an input
- The first item from the reference panel is no longer selected by default to avoid unwanted items being added as a bubble
- The first item from the reference panel is selected only when the dialog has no bubbles or when a search for a non-empty input just ran
- Shift-Enter from input or reference panel will accept the dialog's state instead of creating a bubble
- Ensure that the reference panel reloads when a bubble is deleted

Addresses: #3811, #3671

### DIFF
--- a/chrome/content/zotero-platform/mac/integration.css
+++ b/chrome/content/zotero-platform/mac/integration.css
@@ -3,10 +3,6 @@ body {
 	font-size: 15px;
 }
 
-body[multiline="true"] {
-	line-height: 26px;
-}
-
 window.citation-dialog {
 	background: transparent;
 	-moz-appearance: none;
@@ -18,13 +14,6 @@ window.citation-dialog {
 	height: 37px;
 }
 
-.citation-dialog.search {
-	padding: 2px 2px 0 2px;
-	margin: 2.5px 3.5px;
-	border: 1px solid rgba(0, 0, 0, 0.5);
-	-moz-appearance: none;
-	border-radius: 10px;
-}
 
 
 .citation-dialog.entry {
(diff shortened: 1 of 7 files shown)


abaevbog:
I also attempted to add a button to accept the dialog - not sure about the exact icon and positioning though.

Screenshot 2024-03-08 at 2 04 19 PM

I didn't add any keyboard nav onto that button yet - I'm not totally sure it's required. Now Enter doesn't always add the first item from the ref panel and Shift-Enter always accepts the dialog, so it seems to me it will always take longer to tab your way to the accept button and to click Space/Enter than to just press Enter (or Shift-Enter in the worst case). 

If we do want that button focusable, what would be the best way to get to it? Shift-Tab from a bubble/input focuses the Z-icon, so we could plug the "accept" button into that chain. First Shift-Tab focuses the accept button, second Shift-Tab focuses the icon?

adomasven:
image

The accept button padding is wrong on linux.

If you arrow down to select an item, then arrowing back up should go back to the text field and unselect the selection, so that you can accept the dialog with Enter.

abaevbog:
> The accept button padding is wrong on linux.

Thank you, I think it's nicer now. It should align with the Z icon on all platforms.

> If you arrow down to select an item, then arrowing back up should go back to the text field and unselect the selection, so that you can accept the dialog with Enter.

Oh yeah, good point! Now the navigation between input and the reference panel feels way smoother.

adomasven:
> > The accept button padding is wrong on linux.
> 
> Thank you, I think it's nicer now. It should align with the Z icon on all platforms.


The right-padding is still too much (see screenshot above). Are you not seeing this on your Linux setup?

abaevbog:
> The right-padding is still too much (see screenshot above). Are you not seeing this on your Linux setup?

Apologies, I do. Somehow I missed it at first and just focused on the vertical alignment (that was also off). 

I tried to set equal widths for Z/spinner/accept icons, and then dynamically set the editor's width to fill out the remaining space. I think it worked out. Alternatively, we could manually set the widths for each platform individually but this felt more general and less hacky to me.  

This is what it looks like on my end. The 'accept' button should be aligned with the Z-icon.

Linux:
qf_linux
Windows:
qf_win
Mac:
qf_mac

PR #3817, opened 2024-03-08
Fix item pane expanding on each restart with note selected

### LINKED ISSUE #3813: Note editor / sidebar width keeps increasing after every restart
As I was testing https://github.com/zotero/zotero/issues/3804, I kept restarting Zotero, and clicking on a note. After a few dozen restarts, this is what my Zotero looks like on start:

image

After clicking on an item:

image

After clicking in the note editor:

image

AbeJellinek:
Persist a new attribute that tracks whether the width/height has been expanded to cover a hidden sidenav. Use that to determine whether to adjust the width/height rather than the current visibility of the sidenav.

Fixes #3813

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -1993,8 +1993,10 @@ var ZoteroPane = new function()
 
 			if (!document.querySelector("#zotero-items-splitter").collapsed) {
 				let isStackedMode = Zotero.Prefs.get("layout") === "stacked";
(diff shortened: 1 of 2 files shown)


windingwind:
We will have this automatically fixed after #3624, so... do we want to release any new beta before fx115? If not, maybe let's wait and not introduce further conflicts.

dstillman:
Fixed by https://github.com/zotero/zotero/pull/3624,  as I understand it. Reopen if not.

Issue #3821, opened 2024-03-09
Incorrect item pane layout on Windows


dstillman:
https://forums.zotero.org/discussion/112660/zotero-7-beta-vertical-shift-in-the-creator-separator

Regression from https://github.com/zotero/zotero/pull/3727, I assume

PR #3823, opened 2024-03-10
Fix editable-text height on Windows

windingwind:
fix: #3821


windingwind:
after:

comfortable:

image


compact:

image

PR #3824, opened 2024-03-10
Fix update.xhtml style on Windows

windingwind:


windingwind:
before:

image


after:

image

dstillman:
Is this better? The buttons seem nicer, but the background color just seems uglier to me? Did we do this for all dialogs in https://github.com/zotero/zotero/pull/3727? Where does this background color come from?

windingwind:
> Did we do this for all dialogs in https://github.com/zotero/zotero/pull/3727? Where does this background color come from?

yes, according to figma, we currently have two options for window background, one is `material-background` and another is `material-sidepane`. i am not sure what's the best color here but by default we use the latter.

the default background of firefox is kind of purple(35, 34, 42), which is not very fancy for me.

dstillman:
Ah, OK. At least out of context I find the purple more pleasant, but I acknowledge that it may be more of a Firefox thing than a Zotero thing. (I don't use dark mode, but the Firefox About dialog on macOS, for example, leans into the purple.)

PR #3825, opened 2024-03-10
Use system accent color on Windows

windingwind:
Firefox does not respect system accent color on Windows for buttons, links, inputs, etc., and use weird hard-coded color instead.

Since on Win10/11 users can configure accent color in settings, I think we can consider using accent color for these elements. (previously we use accent-blue everywhere)

what do you think @yexingsha?

### DIFF
--- a/scss/base/_base.scss
+++ b/scss/base/_base.scss
@@ -4,7 +4,7 @@
 	font-style: normal;
 	line-height: 1.3333333333;
 	@media (-moz-platform: windows) {
-		--color-accent: var(--accent-blue);
+		--color-accent: -moz-accent-color;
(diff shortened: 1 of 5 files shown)


windingwind:
default accent color of Win11 (blue):

| before | after |
| ---- | ---- |
| image | image |
| image | image |
| image | image |

windingwind:
> but still keep our own accent color if user has set it to "automatic", just as we do on macOS.

Well, on Windows the `automatic` means the OS infers an accent color from your wallpaper and the native components still use that inferred accent color. It would be weird if we did not respect the system accent color when `automatic` since it does not mean disabling accent color.

On the other hand, we could provide a setting to allow users to pick an accent color for Zotero, which I suppose would satisfy most users.

Issue #3827, opened 2024-03-10
Wrong background color when renaming collections on Windows


dstillman:
I assume a regression from https://github.com/zotero/zotero/pull/3727

![Screenshot 2024-03-10 at 1 09 50 AM](https://github.com/zotero/zotero/assets/783082/e552df93-9d57-41b3-b53a-d82176f698af)

windingwind:
i see. i'll make a fix

PR #3828, opened 2024-03-10
Revert editable text style on Windows

windingwind:
fix: #3826

PR #3829, opened 2024-03-10
Fix collection tree renaming background color

windingwind:
fix: #3827

after #3828


dstillman:
Not quite right:

Screenshot 2024-03-11 at 2 09 30 AM

I think there needs to be some sort of border, like on macOS:

Screenshot 2024-03-11 at 2 10 14 AM

windingwind:
does it look right now?

image

(it's rebased on #3828, which is what i mean by `after 3828`)

dstillman:
Hmm, in the screenshot that looks thicker on the top, no?

windingwind:
MacOS:

Comfortable:

image
image


Compact:

image

image


Windows:

Comfortable:

image

image


Compact:

image

image

Issue #3822, opened 2024-03-09
Scroll bar in open-tabs menu


dstillman:
https://forums.zotero.org/discussion/112661/zotero-7-beta-cannot-scroll-in-the-list-all-tabs

As in Firefox

And should scroll the current tab to the center if possible

PR #3833, opened 2024-03-11
make opened tabs menu scrollable

abaevbog:
- if there are too many tabs opened to fit into the available space on the screen, set the menu's list height so that it becomes scrollable. As a cutoff for the necessary height, use 75% of `screen.height -  window.screenTop`:  this seems to give enough space to avoid overlap with the toolbar (dock) at the bottom. 
- this cutoff will become the menu's height if there are too many entries, so the menu will be shorter if the main window is dragged down closer to the bottom of the screen and longer if the window is at the very top.
- when tabs menu is refreshed, scroll to the selected tab.
- use margins instead of padding to that the scrollbar does not overlap with the cross button.

Fixes: #3822

### DIFF
--- a/scss/components/_tabsMenu.scss
+++ b/scss/components/_tabsMenu.scss
@@ -3,10 +3,6 @@
 	margin: 0;
 }
 
-#zotero-tabs-menu-panel {
-	min-height: 0;
-}
-
 #zotero-tabs-menu-wrapper {
 	width: 350px;
 	background: var(--material-sidepane);
--- a/scss/components/_tabsMenu.scss
+++ b/scss/components/_tabsMenu.scss
@@ -3,16 +3,23 @@
 	margin: 0;
 }
 
+#zotero-tabs-menu-panel {
+	min-height: 0;
+}
+
 #zotero-tabs-menu-wrapper {
 	width: 350px;
 	background: var(--material-sidepane);
-	padding: 7px;
+	padding: 0;
 	border-radius: 5px;
 	--width-focus-border: 2px;
+	display: flex;
+	flex-direction: column;
+	min-height: 0;
 }
 
 #zotero-tabs-menu-filter {
-	margin: 0 0 7px 0;
+	margin: 7px;
 	border-radius: 5px;
 	border: 1px solid transparent;
 	padding-inline-start: 5px !important;
@@ -26,6 +33,13 @@
 #zotero-tabs-menu-list {
 	appearance: none;
 	margin: 0;
+	overflow-x: hidden;
+	overflow-y: scroll;
+	scrollbar-width: thin;
+
+	& > toolbaritem {
+		margin: 0 7px;
+	}
 
 	.row {
 		display: flex;
--- a/scss/components/_tabsMenu.scss
+++ b/scss/components/_tabsMenu.scss
@@ -19,7 +19,7 @@
 }
 
 #zotero-tabs-menu-filter {
-	margin: 7px;
+	margin: 7px 7px 2px 7px;
 	border-radius: 5px;
 	border: 1px solid transparent;
 	padding-inline-start: 5px !important;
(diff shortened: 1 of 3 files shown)


abaevbog:
This is what I mean by the varying height of the menu. The intention is to somewhat replicate the approach of firefox to make it so that the menu does not go outside of the screen

Screenshot 2024-03-11 at 1 52 31 PM


Screenshot 2024-03-11 at 1 52 50 PM

dstillman:
If I drag the window towards the bottom of the screen and the menu flips to be above the window, it gets really short:

Screenshot 2024-03-11 at 9 54 07 PM

This isn't a particularly realistic example, but is there a way to let it use the available height when it goes above?

abaevbog:
Hmm, yeah that's a bit ugly...

I tried the following tweak for the height setting logic and I think it helps to address this:
- set the height to be the available screen space below the window top as long as it's taller than a certain cutoff constant. If the available space is below the cutoff, most likely the menu will flip and appear above the window
- in that case, set the height to be the screen space that's above the window
- this utilizes a few constants to account for all the dock/toolbar/etc. heights, so this part is not great. But I don't think we have any more reliable way to tell if the menu will appear above or below the window


This is what the menus look like for me in the lower half of the screen now

Screenshot 2024-03-11 at 11 43 11 PM

Screenshot 2024-03-11 at 11 43 20 PM

dstillman:
> But I don't think we have any more reliable way to tell if the menu will appear above or below the window

Not sure if this is helpful, but `#zotero-tabs-menu-panel` does have `side="top"` or `side="bottom"` for appearing below and above the button, respectively. So could potentially check that and resize?

abaevbog:
> Not sure if this is helpful, but #zotero-tabs-menu-panel does have side="top" or side="bottom" for appearing below and above the button, respectively. So could potentially check that and resize?

The issue with `side` attribute was that it is accurate on `popupshown` but we want to handle all popup setting on `popupshowing` to avoid the `panel` jumping/resizing/scrolling once it is visible (particularly on windows). But most importantly, I think there was a somewhat simpler, mainly css-based, approach than what I had initially tried.

The first commit is mainly css tweaks to make the menu scrollable. Functionally, just that could be enough to navigate a very long list of tabs. One thing that firefox does is it enforces a gap between the end of the opened tabs menu and the edge of the screen, and this is what I put into the second commit because I wasn't sure if we actually want it or not.

For reference:

[tabs menu without restricting the height](https://www.dropbox.com/scl/fi/wkqtarhqmeey6meqeu9sr/no_gap.mov?rlkey=05kfbuuy92jgy85cmws1fdxpi&dl=0) 

[tabs menu that leaves a gap](https://www.dropbox.com/scl/fi/ykwcdnavgj3vd1r85eyz8/with_gap.mov?rlkey=p1r1ubn2fs2iq76645wi3uqrt&dl=0)

What do you think? Do we want to restrict the max height to leave a bit of space between the menu and the edge of the screen?

dstillman:
Hmm, this seems wrong:

Screenshot 2024-03-12 at 7 04 47 PM

There's much more room above, so it seems like the popup should be going up at that point. In Firefox, it seems to flip at the screen's halfway point, which makes sense. Can we do the same?

abaevbog:
I see! Yeah, I was trying to avoid going through calculating the positioning of the top/bottom of the menu in relation to the screen edge. But seems like that's really the way to go. So I re-pushed an approach where we basically check if the edge of the menu is within a margin of the edge of the available screen, and only if that's the case, we set the `max-height` to restrict the menu's height. So this does not affect at which point the menu flips and it should happen about half way through.

Windows and Linux required a bit of a special treatment. Windows `panel` sizing returned with `getBoundingClientRect` is not right until `popupshown`, so in `popupshowing` (where this is handled) I estimated the positioning based on the anchor button.  Linux `screen.availTop` and `screen.availHeight` simply didn't work correctly on my machine - elements could be moved past what should be the top of available screen from `screen.availTop`, so there are some minor workarounds to handle that too.

scss/components/_tabsMenu.scss
```diff
@@ -3,16 +3,23 @@
 	margin: 0;
 }
 
+#zotero-tabs-menu-panel {
+	min-height: 0;
```
dstillman, line 7: Can we add a comment for this?

abaevbog: I rebased and it looks like this css is no longer needed. This must've been a left-over either from pre-fx115, or from my initial attempt to setup the scrollable menu without explicitly calculating `maxHeight` with JS. So I just removed it.

Issue #3783, opened 2024-03-02
Show in Library from tab context menu doesn't focus items list


dstillman:
https://forums.zotero.org/discussion/111620/zotero-7-beta-show-in-library-not-focusing-the-item-in-the-items-list

PR #3836, opened 2024-03-11
fix to "show in library" not focusing the itemTree

abaevbog:
- handle additional zpRefocusID options value in Zotero_Tabs.select to focus a node with the specified ID, instead of what was recorded as last focused element
- use that parameter in ZoteroPane.selectItems to make sure that itemTree is focused after the tab is switched to zotero-pane
- removed redundant calls to Zotero_Tabs.select('zotero-pane') after Zotero_Pane.selectItems

Fixes: #3783

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -3110,7 +3110,7 @@ var ZoteroPane = new function()
 			document.getElementById(ZoteroPane.itemsView.id).focus();
 		}
 		
-		Zotero_Tabs.select('zotero-pane');
+		Zotero_Tabs.select('zotero-pane', false, { focusElementID: ZoteroPane.itemsView.id });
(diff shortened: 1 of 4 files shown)


chrome/content/zotero/zoteroPane.js
```diff
@@ -3110,7 +3110,7 @@ var ZoteroPane = new function()
 			document.getElementById(ZoteroPane.itemsView.id).focus();
 		}
 		
-		Zotero_Tabs.select('zotero-pane');
+		Zotero_Tabs.select('zotero-pane', false, { zpRefocusID: ZoteroPane.itemsView.id });
```
dstillman, line 3113: `zpRefocusID` seems overly concerned with the internals of `Zotero_Tabs.select()`. Shouldn't it just be, say, `focusElementID`?

PR #3841, opened 2024-03-12
Set main window min-width/height dynamically

### LINKED ISSUE #3607: Reduce minimum width on window (?)
https://forums.zotero.org/discussion/111160/zotero-7-beta-please-put-the-split-view-back

I'm not sure if the min-width is on the items list or the window, but something is indeed preventing the window from being as narrow as it could be before, even though we've made more room in the toolbar. (Previously the window could be much narrow, but I'm comparing with the point where elements started to be cut off in the old design.)

This is particularly noticeable in split view, which could easily be narrower than it is now.

AbeJellinek:
Now the window size constraints depend on the current sizes of the sidebars. So if the item pane is collapsed, you can make the window very narrow; if the item pane is very wide, the window can't be made very narrow. Adjusts automatically for Stacked mode.

Fixes #3607

### DIFF
--- a/chrome/content/zotero/standalone/standalone.js
+++ b/chrome/content/zotero/standalone/standalone.js
@@ -75,7 +75,7 @@ const ZoteroStandalone = new function() {
 								this.switchReaderSubtype(subtype);
 							}
 						}
-						setTimeout(() => ZoteroPane.updateToolbarPosition(), 0);
+						setTimeout(() => ZoteroPane.updateLayoutConstraints(), 0);
 					}
 				}
 			},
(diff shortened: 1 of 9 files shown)


dstillman:
Let's update this for 115. This seems to be a common source of frustration.

dstillman:
Hmm, so I'm not actually seeing any difference here in min-height or min-width when I have all three panes open, but maybe that's expected? I get exactly this in the current beta and with this PR:

Screenshot 2024-04-11 at 11 48 06 PM

The difference, I guess, is that if I close the item pane, I can then shrink the window further? (And if I make the item pane much wider, the min-width is also bigger.)

But I do then lose the search bar and sidenav at narrow widths:

Screenshot 2024-04-12 at 12 00 28 AM

Seems like we need to either allow the pane to shrink to an absurdly narrow width where it's totally unusable (which Thunderbird does) or try to actually keep all the visible components accessible.

PR #3846, opened 2024-03-13
SAXXMLReader: Handle non-UTF-8 encodings

### LINKED ISSUE #3844: Handle non-UTF-8 feeds
https://forums.zotero.org/discussion/112767/news-feed-formating-error-dont-show-ae-o-and-a

It seems like the Fetch API just opts out of character-set handling and treats everything as UTF-8 when you use `text()`, so we'd have to [handle this manually](https://stackoverflow.com/a/66125431).

https://github.com/zotero/zotero/blob/f8d59d14bc2f7c3a9a8ccd6ac41b01fe209aa5da/resource/feeds/SAXXMLReader.js#L65

AbeJellinek:
We don't handle encodings set in the XML declaration (`<?xml encoding="..."?>`), just whatever is set in the `Content-Type` header. Handling encodings set in the declaration would require re-parsing the feed, the `Content-Type` header has precedence, and we haven't received any reports of mis-decoded feeds without correct `Content-Type` headers. Can implement later on if it's needed.

Getting httpd to write a file as binary with the correct encoding in the test was harder than it seems like it should be, so maybe I'm missing a better approach.

Fixes #3844


dstillman:
Case-sensitive filesystem!

Issue #3763, opened 2024-02-28
Select multiple results in Quick Format via keyboard


dstillman:
https://forums.zotero.org/discussion/112343/accessibility-cannot-multi-select-in-quick-format-citation-via-spacebar

I wanted Shift+Up/Down-Arrow for Selected Items, mostly, and we're planning to change that to just having a single result that adds all, but this thread suggests it could be useful for other results too, so no reason not to support it more generally.

adomasven:
> And Cmd/Ctrl+Up/Down-Arrow and space bar as well, I guess

I don't know if we can reasonably support this. You can use arrow keys to navigate the reference list, and Enter to add an item at any point while typing to search, and spacebar is part of the typing characters. We could change the behavior of spacebar specifically once the user moves focus to the reference list, but I think it is bad to overload a single key for multiple behaviors.

PR #3851, opened 2024-03-15
QuickFormat: adding multiple items with shift-arrow

abaevbog:
- reference panel supports multiselect
- Shift/Cmd-ArrowDown/Up from the selected reference panel row will toggle the next item's selected status and focus it
- Cmd-MouseClick on an item from the reference panel will toggle the item's selected status
- Zotero_QuickFormat._bubbleizeSelected adds all selected items before sorting and clearing input

Fixes: #3763

### DIFF
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -164,6 +164,10 @@ var Zotero_QuickFormat = new function () {
 					Zotero_QuickFormat._bubbleizeSelected();
 					return;
 				}
+				// Make sure there is a selected item when shift-click is handled
+				if (referenceBox.selectedIndex < 1) {
+					_selectFirstReference();
+				}
 				// Shift-click can end up selecting disabled separator, so make sure it's removed
 				setTimeout(() => {
 					let selectedSeparators = [...document.querySelectorAll("richlistitem[disabled='true'][selected='true']")];
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -1343,7 +1343,7 @@ var Zotero_QuickFormat = new function () {
 		// to position references panel properly
 		let dialogBottom = dialog.getBoundingClientRect().bottom;
 		let panelTop = referencePanel.getBoundingClientRect().top;
-		if (Math.abs(dialogBottom - panelTop) > 5) {
+		if (Math.abs(dialogBottom - panelTop) > 5 && referencePanel.state == "open") {
 			referencePanel.hidePopup();
 			// Skip a tick, otherwise the panel may just remain open where it was
 			setTimeout(_openReferencePanel);
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -165,7 +165,7 @@ var Zotero_QuickFormat = new function () {
 					return;
 				}
 				// Make sure there is a selected item when shift-click is handled
-				if (referenceBox.selectedIndex < 1) {
+				if (e.shiftKey && referenceBox.selectedCount < 1) {
 					_selectFirstReference();
 				}
 				// Shift-click can end up selecting disabled separator, so make sure it's removed
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -1145,6 +1145,14 @@ var Zotero_QuickFormat = new function () {
 	function getAllBubbles() {
 		return [...editor.querySelectorAll(".bubble")];
 	}
+
+	// Delete the bubble and clear locator node if it pointed at this bubble
+	function _deleteBubble(bubble) {
+		if (bubble == locatorNode) {
+			locatorNode = null;
+		}
+		bubble.remove();
+	}
 	
 	/**
 	 * Clear reference box
@@ -1814,6 +1822,7 @@ var Zotero_QuickFormat = new function () {
 		else {
 			editor.prepend(newInput);
 		}
+		locatorNode = null;
 		newInput.focus();
 	}
 
@@ -1886,7 +1895,7 @@ var Zotero_QuickFormat = new function () {
 			// Backspace/Delete from the beginning of an input will delete the previous bubble.
 			// If there are two inputs next to each other as a result, they are merged
 			if (this.previousElementSibling) {
-				this.previousElementSibling.remove();
+				_deleteBubble(this.previousElementSibling);
 				_combineNeighboringInputs();
 			}
 			// Rerun search to update opened documents section if needed
@@ -1981,7 +1990,7 @@ var Zotero_QuickFormat = new function () {
 			if (!moveFocusBack(this)) {
 				moveFocusForward(this);
 			}
-			this.remove();
+			_deleteBubble(this);
 			// Removed item bubble may belong to opened documents section. Reference panel
 			// needs to be reset so that it appears among other items.
 			_clearEntryList();
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -150,15 +150,8 @@ var Zotero_QuickFormat = new function () {
 			});
 			referenceBox.addEventListener("click", (e) => {
 				let item = e.target.closest("richlistitem");
-				if (!item || e.button !== 0) return;
-				// Prevent selecting separator with CMD/Ctrl-Click
-				if (item.disabled) {
-					// Clear selection on header click
-					referenceBox.clearSelection();
-					referenceBox.currentIndex = 0;
-					e.preventDefault();
-					return;
-				}
+				if (!item || e.button !== 0 || item.disabled) return;
+
 				let mouseMultiSelect = e.shiftKey || (Zotero.isMac && e.metaKey) || (!Zotero.isMac && e.ctrlKey);
 				let multipleSelected = referenceBox.selectedCount > 1;
 				let isItemSelected = [...referenceBox.selectedItems].findIndex(node => node == item) !== -1;
@@ -1782,12 +1775,6 @@ var Zotero_QuickFormat = new function () {
 
 	function _onQuickSearchClick(event) {
 		if (qfGuidance) qfGuidance.hide();
-		if (!event.target.classList.contains("editor")) {
-			// On click of border outside of the editor, clear multi-select
-			referenceBox.currentIndex = 0;
-			referenceBox.clearSelection();
-			return;
-		}
 		let clickX = event.clientX;
 		let clickY = event.clientY;
 		let { lastBubble, startOfTheLine } = getLastBubbleBeforePoint(clickX, clickY);
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -106,7 +106,7 @@ var Zotero_QuickFormat = new function () {
 				if ((event.key == "Enter" && !event.shiftKey) || event.charCode == 59) {
 					event.preventDefault();
 					event.stopPropagation();
-					event.target.closest("richlistitem").click();
+					Zotero_QuickFormat._bubbleizeSelected();
 				}
 				// Tab will move focus back to the input field
 				else if (event.key === "Tab") {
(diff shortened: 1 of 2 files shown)


abaevbog:
Brief video demo:


https://github.com/zotero/zotero/assets/36271954/d0d12046-3d1a-46c4-ba17-41ad1df6d056

abaevbog:
There are a few minor tweaks left but I wanted to check if this is the general idea of what we are after.

I agree that it might be better to [not overload keys](https://github.com/zotero/zotero/issues/3763#issuecomment-1970556194) with multiple actions, so I did not include the space bar into the multiselect workflow.

A few questions:
- what do we want to do with the locator? If one types "some title p.6" and then selects multiple items and hits Enter, do we add "p.6" to all bubbles? Or ignore it? On one hand, the use case of adding "p.6" to all bubbles is not quite clear to me. But on the other hand, maybe there are use cases for it so why not?
- If there is a retracted item being added as one of multiple items, there will be a warning popup. One can click to see the item in the library or to cancel the insertion of the bubble. Do we just skip the retracted item and add all non-retracted items as bubbles? Or is it better to not add anything and just remove the multiselect?
- Not relevant to this PR but maybe there should be some kind of indicator in the reference panel list that an item is retracted? A red cross icon after the name, perhaps?
- I assumed that the locator node will not be attached to any specific bubble from a multiselect batch. So if you select 5 items, add them, and then type "p.5", it will be applies to the last bubble. Correct me if that's wrong.

abaevbog:
@adomasven, I rebased this and the general functionality works for me. Let me know what you think and what it's best to do with [locators and retractions](https://github.com/zotero/zotero/pull/3851#issuecomment-1998878204). For now, l kept the locator being added to all bubbles (really not sure what's the use case but also not sure if there's a reason to remove this), and retractions that are not OK'd are just skipped with other items being added.

One thing worth noting: I added `Ctrl-arrowUp/Down` as a another way to trigger multiselect because `shift-arrowUp/Down` caused issues with JAWS screen reader on windows.

adomasven:
> * what do we want to do with the locator? If one types "some title p.6" and then selects multiple items and hits Enter, do we add "p.6" to all bubbles? Or ignore it? On one hand, the use case of adding "p.6" to all bubbles is not quite clear to me. But on the other hand, maybe there are use cases for it so why not?

I'm not a fan of any option here, but I guess it would be much harder to type "query p.6" unintentionally and be surprised with all bubbles receiving the locator, as opposed to doing it intentionally and being delighted that it works. On the other hand I can't imagine how this could ever be practical, and still the logic is sound to apply it to all bubbles.

> If there is a retracted item being added as one of multiple items, there will be a warning popup. One can click to see the item in the library or to cancel the insertion of the bubble. Do we just skip the retracted item and add all non-retracted items as bubbles? Or is it better to not add anything and just remove the multiselect?

Can we keep the multiselect with removing just that single item from the selected list?

> Not relevant to this PR but maybe there should be some kind of indicator in the reference panel list that an item is retracted? A red cross icon after the name, perhaps?

I'd say better a red badge that says "Retracted". A cross is ambiguous, and you generally don't want to cite a retracted item even if you want to keep it in your library.

> I assumed that the locator node will not be attached to any specific bubble from a multiselect batch. So if you select 5 items, add them, and then type "p.5", it will be applies to the last bubble. Correct me if that's wrong.

That makes sense.

> One thing worth noting: I added Ctrl-arrowUp/Down as a another way to trigger multiselect because shift-arrowUp/Down caused issues with JAWS screen reader on windows.

Hm. Don't we have the same problem in the main library view, or is JAWS doing something different there? Ctrl-Arrow is generally reserved to moving the focused item (so that you can select it with Space) in multi-select contexts and I don't think it's good to mess with these conventions or we are going to cause all software to abandon conventions and then you will never be able to know which modifier to hold in multi-select contexts for the rest of your life.

@dstillman may want to weigh in on some of these too.

adomasven:
It's a bit weird that you can confirm-select multiple items with Shift-Click, but Ctrl-Click selected items have to be added with keyboard Enter. I don't have ideas for how to do confirm selection with the mouse though and maybe it's no big deal, but would be nice to have.

abaevbog:
> Don't we have the same problem in the main library view, or is JAWS doing something different there? 

The same happens in the `itemTree` for me, yes. On Shift-arrowUp/Down JAWS announces "selected" and nothing else. I am noticing, though, that with other applications it does work as it should. For example, shift-arrow within the file explorer does end up announcing the file name after the "selected" part. So this seems to be more of a general issue with JAWS, as opposed to something specific to the quick format dialog. I'll look into it - this might be better to treat a separate issue. 

As a side note, somehow I never had any issues with NVDA. All other tweaks and workarounds with accessibility updates were specifically to accommodate JAWS (which unfortunately happens to be the most popular screen reader)

adomasven:
> These are both great points!

I may have exaggerated a little bit on the first point.

> I am now seeing that the default richlistbox behavior (as well as, apparently, some of the convention based on what I see in Thunderbird) is that Shift-mouseclick selects all entries between the last selected entry and the target. If one want to confirm-click on multiple selected items (e.g. to open multiple emails in Thunderbird), one does Shift-doubleClick. So we should probably do the same? The only reservation with this is that we don't use double-click anywhere else in the dialog.

I don't think this makes sense for our purposes. Firstly, that kind of behavior, especially multiselect is usually used for select popups that are not dismissed on select (i.e. an item list on a page), which why even multi-select in this UI IMO is slightly confusing, but worth it for additional functionality. Secondly, would you even be able to capture a dblclick here as we are already bubbleizing on a single click, without introducing some sort of delay? Thirdly, who is ever going to discover that after Ctrl-Clicking you need to Shift-DoubleClick to confirm? 

It would need to be way more discoverable, like showing the ghosts of selected bubbles in the dialog, and being able to click on them to add, but I imagine it would be a pain to implement something like that for this fairly obscure advanced use-case where an advanced user is likely going to be able to figure out to press Enter on their own.

abaevbog:
> Can we keep the multiselect with removing just that single item from the selected list?

Makes sense. If a retracted item is not just OK'd, no bubbles will be added and the item will be removed from multiselection. I also added a red "This work has been retracted" label to the beginning of retracted item's description so that it's visible in the references panel and in the item properties popover. Not sure if this is the best spot for it but it does get the message across

 
Screenshot 2024-03-19 at 4 43 58 PM

abaevbog:
> I don't think this makes sense for our purposes. Firstly, that kind of behavior, especially multiselect is usually used for select popups that are not dismissed on select (i.e. an item list on a page), which why even multi-select in this UI IMO is slightly confusing, but worth it for additional functionality. Secondly, would you even be able to capture a dblclick here as we are already bubbleizing on a single click, without introducing some sort of delay? Thirdly, who is ever going to discover that after Ctrl-Clicking you need to Shift-DoubleClick to confirm?

That is fair. I also realize I didn't give it enough thought on what are the possible/desired use cases for the `shift-click`, so whatever it was doing before, it was probably not exactly right.

With my last change, I tried to allow `shift-click` to do it's intended thing of selecting all rows between the currently selected row and the target ("intended" as in it's handled by the `panel` itself). I think this has the following nice application: if one wants to batch-add many items from the opened or selected items, instead of holding `shift-arrowDown` or clicking through all items holding `cmd/ctrl`, one can just click the last relevant item from the batch holding `shift`. This should be fairly discoverable, since it's kinda close to the `shift-click` selection logic done everywhere else. `shift-click` doesn't trigger bubble creation, so one still needs to press `Enter`: that way if you want to, say, add all items but the very first one, you can  select them all with `shift-click` and exclude the first item with `cmd-click` before pressing `Enter`.

What do you think about this behavior? For reference, this is `shift-click` followed by `cmd-click` and `Enter` example I had in mind: 

https://github.com/zotero/zotero/assets/36271954/3b7da8f2-33a0-4713-98dc-b2ddeb4895af

adomasven:
I think the previous Shift-Click behaviour was fine. The user can manually remove the bubbles that they do not want to keep after adding with Shift-Click. My whole point was that after Ctrl-Clicking to select multiple items it is not immediately clear how to confirm the selection. Still, an advanced user able to discover Ctrl-Clicking will be able to figure out Enter too. On the other hand Shift-Click to confirm requires discovering less uncommunicated interactions, since it is obvious how to remove bubbles.

abaevbog:
> I think the previous Shift-Click behaviour was fine.

Gotcha! I re-pushed the last commit where the `shift-click` just bubbleizes the multi-selected items just like before. Prior to this, the selection was affected by the default `panel` handler, so this has some additional `event.preventDefault()`s to make sure that no extra bubbles are added.  And it doesn't matter how the items were selected - via `ctrl-click` or `shift-keydown`.

adomasven:
I think we can go with this unless @dstillman has some additional thoughts.

abaevbog:
Cool! I went back and forth with the changes a few times, so let me rebase so that there's nothing unnecessary here

dstillman:
Sorry, I didn't totally follow the above discussion, but trying this out I was immediately confused that Shift-click didn't select from the selected item to the one I Shift-clicked on, only the already selected item. I see that maybe we're using Shift-click to bubbleize the existing multi-selection, which I guess is nice to avoid the keyboard if you somehow discover that, but it's totally non-standard, whereas Shift-click to select a range is. (It also looks like right-click bubbleizes the existing multi-selection, which I'm not sure is intentional.) I would have Shift-click follow standard behavior and select a range, and if we want to find some other way to bubbleize a multi-selection with the mouse, we can.

abaevbog:
> I would have Shift-click follow standard behavior and select a range, and if we want to find some other way to bubbleize a multi-selection with the mouse, we can.

Great point! After having a break from this and trying it fresh today, our current shift-click behavior did strike me as unusual as well. So yeah, shift-click will select everything between the clicked item and the last selected item (or the very first item, if no other items are selected). 

And then, to confirm selection without having to press Enter, maybe a click on a multi-selected row could bubbleize everything that was selected? A click on non-selected row will just bubbleize the clicked row and discard multiselection as it happens now. It is not standard but maybe it could fit here?

>  (It also looks like right-click bubbleizes the existing multi-selection, which I'm not sure is intentional.)

Hm, that was not intentional at all. Thank you for pointing it out!

dstillman:
> to confirm selection without having to press Enter, maybe a click on a multi-selected row could bubbleize everything that was selected?

Good idea. Let's try that.

adomasven:
> Shift-click didn't select from the selected item to the one I Shift-clicked on

I am confused. The current state of this PR does exactly that. You can only change the currently selected item with keyboard arrow keys, but then a shift-click to a different item will bubbleize from the currently selected to the one you clicked on. Is that not what you're seeing, or do you think that is wrong? We settled on immediate bubbleization because it's easy enough to remove any unwanted items from the QF. On the other hand if a shift-click just set selection without bubbleization, then you would either need a different mouse gesture to confirm selection (which would not be obvious), or use Enter to do it.

dstillman:
> You can only change the currently selected item with keyboard arrow keys, but then a shift-click to a different item will bubbleize from the currently selected to the one you clicked on. Is that not what you're seeing, or do you think that is wrong?

It's not what I'm seeing.

https://github.com/zotero/zotero/assets/783082/65c34208-319e-4236-acef-7da96aff8007

Sorry to conflate with immediate selection. Immediate bubbleization is fine as long as it first extends the selection range.

abaevbog:
@adomasven apologies, I may have misunderstood [your earlier comment](https://github.com/zotero/zotero/pull/3851#issuecomment-2008831219) after my [initial tweak](https://github.com/zotero/zotero/pull/3851#issuecomment-2008601708) to shift-click behavior. I was under impression that shift-click should just confirm the selection without changing what the actual selection is. So I had discarded that shift-click logic and, until now, selecting the first 2 items and then shift-clicking on the 3rd one would just confirm the selection of the first 2 items. That is the behavior @dstillman was seeing.

I just incorporated that earlier shift-click logic back here. Now, the behavior for shift-click is essentially handled by `richlistbox` itself. So, shift-click on an item will select the range between the clicked item and the previously selected item if there is one or the first item otherwise. It does not confirm the selection right away, since one might want to un-select something in the middle with cmd/ctrl-click (but we can if it's better?). Then, a click without modifiers on one of selected items will confirm the entire selection. A click on an item outside of the selection will discard the selection and bubbleize just that clicked item. Also, I added a tweak to clear selection if you click on the red border - it felt intuitive to me but correct me if this is not needed.

For reference, in this test example, I Shift-click to select a range, then use cmd-click to un-select some items and then click on a selected item to confirm the selection. What do you think?

https://github.com/zotero/zotero/assets/36271954/9554b718-cbb5-4cd5-b844-916c02fc4e7a


There are also a few tiny bug fixes that I ran into - I kept them as separate commits to keep it organized.

dstillman:
This is great, but I would take out the clearing on the red border. That doesn't really make any sense to me (and that's not something we'll have in future versions anyway).

abaevbog:
> This is great, but I would take out the clearing on the red border. That doesn't really make any sense to me (and that's not something we'll have in future versions anyway).

Got it - I removed it! My idea was to give an intuitive way to discard the selection without having to use the keyboard (thinking along the lines of having an option to confirm the selection by clicking without having to hit Enter). I suppose it really  is not that much of a concern, especially that un-selecting items will likely not be a part of most people's workflow with the references box and deleting accidentally added bubbles is pretty easy.

Issue #3847, opened 2024-03-14
QuickFormat reference panel positioning on windows


abaevbog:
@adomasven, I'm afraid I just ran into the exact reason for the small change to the margins that [broke the dialog](https://github.com/zotero/zotero/commit/633bee45afce3a4f68476ccc205cdc607e126035#commitcomment-139546386) on windows with non-standard refresh rate.

This is the reference panel positioning issue I'm seeing on windows 11 VM:

<img width="896" alt="Screenshot 2024-03-14 at 12 40 58 AM" src="https://github.com/zotero/zotero/assets/36271954/7d1ad653-bdfa-4573-8f05-5a3381ee2044">

Can you reproduce it? This does not happen for me with `dialog.style.marginBottom = "-36px";` that caused refresh rate breakage, so fixing this is what must have been my intention then. 

Now that there's more context, I'll try different options of panel positioning that hopefully can fix it without breaking things elsewhere

abaevbog:
I tried [this tweak](https://github.com/abaevbog/zotero/commit/5fe9c8a24a387c7d89b86c7a840f12d3cb99ff3d) to use the editor, as opposed to `document.documentElement`, as the anchor and it did position the panel correctly. This still requires a few more tweaks to make sure that as the window expands/shrinks the panel is properly placed (especially if the ref panel is always open). 

I didn't want to go too far without making sure that it does not bring aback the refresh rate breakage, and I still have not been able to fully reproduce that on my monitor. So I wanted to check in if the dialog properly appears with this change for you?

adomasven:
> I didn't want to go too far without making sure that it does not bring aback the refresh rate breakage, and I still have not been able to fully reproduce that on my monitor. So I wanted to check in if the dialog properly appears with this change for you?

I was only able to reproduce it on Windows 10 running natively with an nVidia GPU. I do not have a setup with a VM that could run at higher refresh rates. I'll test this and let you know.

adomasven:
It opens fine with the tweak on my setup.

abaevbog:
Ok, awesome! As far as I can tell, mac and linux work work as well. I added a small tweak to make sure that after window resizing the panel is closed and re-opened so that it gets repositioned properly.

I'm a bit surprised that we didn't see any reports about this, since it makes all lines of the editor but the first one very hard to use... I guess not that many people end up having so many bubbles that they need more than a single line in the editor.

abaevbog:
[...]

PR #3853, opened 2024-03-15
QuickFormat dialog: fix wrong references panel position on windows

abaevbog:
- use dialog as the anchor, not the document element
- after the window is resized, the panel will remain where it was, so if it's far away from the dialog, if gets closed and re-opened to reposition it.

Fixes: #3847

### DIFF
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -131,7 +131,6 @@ var Zotero_QuickFormat = new function () {
 				}
 			});
 			if (Zotero.isWin) {
-				referencePanel.style.marginTop = "-34px";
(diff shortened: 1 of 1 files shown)


abaevbog:
@adomasven I'm afraid there's been another report since this change of the dialog not appearing: https://forums.zotero.org/discussion/comment/459518/#Comment_459518. They do point out that somehow there is no issue with the first annotation but the dialog won't show up once the first one is added. Is it something you can reproduce?

If this change was the culprit and for some crazy reason we must have `dialog.style.marginBottom;`, I guess we could calculate the margins dynamically after every `_resizeWindow`. It is ugly and would need to be checked with `fx115` but maybe it could help

PR #3860, opened 2024-03-18
Add auto rename functionality for attachment files. 

### LINKED ISSUE #1685: Automatically rename file when parent metadata changes
https://forums.zotero.org/discussion/77082/file-isnt-renamed-automatically-when-metadata-was-changed

tnajdek:
This uses `Zotero.Notifier.registerObserver` to observe item changes and, if auto-renaming is enabled, renames attachment files to match. It re-uses some of the logic from the manual `Rename File from Parent Metadata`, including changing attachment item title if it was set to the file name. However, contrary to the manual rename, this will only attempt to rename attachments present on `autoRenameFiles.fileTypes` pref list (i.e., PDFs and EPUBs by default).

Anything should be done differently, please let me know.

Resolve #1685


chrome/content/zotero/zoteroPane.js
```diff
@@ -2972,6 +2999,46 @@ var ZoteroPane = new function()
...
+				
+				for (let id of ids) {
+					const parentItem = await Zotero.Items.getAsync(id);
+					if (!parentItem.isTopLevelItem()) {
+						continue;
+					}
+					const attachments = await parentItem.getAttachments(true);
+					for (let attachmentID of attachments) {
+						let attachmentItem = await Zotero.Items.getAsync(attachmentID);
+						if (!Zotero.Attachments.shouldAutoRenameFile(attachmentItem.attachmentLinkMode == Zotero.Attachments.LINK_MODE_LINKED_FILE)) {
```
dstillman, line 3024: `.isLinkedFileAttachment()`

dstillman:
A bunch of things we need to consider here:

1. The current pref is "Automatically rename locally added files" (and its associated "Rename linked files" pref), but that doesn't work for this, because this applies to downloaded files too — there's no distinction after the files are already in Zotero. So if we only have one pref, it'd have to be just "Automatically rename files" or something like that, and the explanatory text would just explain that downloaded files are always automatically renamed initially. The question is whether we think people might want the current behavior, where locally added files are renamed once but aren't changed after that. I could see that being useful for linked files in particular, where you're storing the path elsewhere and you don't want the file renamed just because you changed one character in a title. But that would mean we'd need a separate pref for continuous renaming ("Keep files named based on parent metadata"), and then we'd need a separate linked-files pref for that too, which just starts to get pretty fussy. So I think if you want to use renaming but don't want filenames to change out from under you, that's what "Rename File from Parent Metadata" is for, and you just turn "Automatically rename files" off? Downloaded files still get renamed once, but you're on your own after that and for locally added files. Most people probably don't use automatic renaming for linked files anyway, and most people probably don't use paths to stored files elsewhere, so this might not affect too many people.
2. This won't affect existing files. Do we want to show some sort of prompt to rename those?
3. If you change the rename template, existing files won't be renamed. Do we want to prompt there? Does it even make sense to prompt, or should it just rename everything automatically? If you're using continuous renaming, you can't be relying on these paths externally, and it's weird for only items you rename from that point on to have their files reflect the pattern.
4. Except we probably can't automatically rename all files, because we don't currently sync the format string, and if you have different strings on different computers, you'd have a total mess on your hands. So then maybe we need to prompt, and you only do a mass-change on one computer, and you make sure others are set the same?
5. Somewhat related, and maybe depending on what we decide about prompting, we should at least disable but maybe just hide "Rename File from Parent Metadata" if the current name matches the current format string.
6. The process of mass-renaming files is unfortunately not super efficient. We have to update all attachment items, and then we actually have to do a file-sync upload for each one, because the filename is (currently, at least) part of the file-sync process. For Zotero Storage, it won't actually upload the file, but I think for WebDAV it will, which means potentially gigabytes of file uploading if all files are renamed. I don't think we want to compromise the experience too much for the sake of WebDAV, though. What I need to check is whether a Zotero Storage rename will actually trigger a download on the other side. If it does, we might want to fix that.

dstillman:
7. If we do prompt on upgrade, or when you turn the pref on, or when you change the format string, and you can say no, we probably need an option to trigger renaming on all files somewhere, either near the format-string pref or in Tools → Manage Attachments.
8. Group libraries??? The format string applies to group libraries too, so we can't really mass-change all files in groups. This is technically a problem now, but only limited to manual renames.

dstillman:
9. Secondary files? If I'm reading the code right, this is renaming all attachments of an item, but we historically have not auto-renamed secondary files, either when saving from the web (e.g., for snapshots) or when dragging files onto items that already have primary attachments. The latter might be supplementary files, where the filenames are useful for conveying exactly what they are. So I think we want to limit this to the primary attachment (a.k.a. "best attachment"). (You can't currently change the primary attachment, but that's planned, which will create a situation where a formerly renamed primary file would no longer be automatically renamed once it became secondary. But I guess there's nothing to be done about that.)

tnajdek:
1. Yes, let's change the preference name to "Automatically rename files."
2. I'd say, show a prompt.
3. & 4. I'd suggest adding a prompt here as well.
5. If we can do it without affecting performance in the item list, then we should definitely hide/disable it.
6. Maybe we show progress in a dialog similar to the one used for "retrieve metadata" for PDF? It could be minimized in the same way, with a different icon. We should check storage configuration and add extra warning if WebDAV is enabled or disable it completely in that case.
7. Yes.
8. Should this not apply in group libraries then?
9. Yes, automatic changes should only apply to the primary/"best" attachment. I'll tweak this.

dstillman:
> 8. Should this not apply in group libraries then?

Or it should, but only with the default format, at least until we add a way to set this as a synced setting in a group?

dstillman:
> 6. Maybe we show progress in a dialog similar to the one used for "retrieve metadata" for PDF? It could be minimized in the same way, with a different icon. We should check storage configuration and add extra warning if WebDAV is enabled or disable it completely in that case.

The slow part would mostly just be during the next sync, so I don't think a progress dialog like that would make sense. The actual renames would generally be very quick. For that part, we would just want a standard progress meter.

tnajdek:
@dstillman I think this is ready for another review. I've rebased on top of main but kept individual commits for now - happy to squash into a single commit before merging.

Summary of changes:

- Automatic rename will now only rename the "best" attachment.
- Reworded the "File renaming" intro and changed the label next to the checkbox that enables this functionality.
- Introduced a prompt whenever enabling this feature or when customizing the filename format to rename existing files.
- Introduced a new menu option to trigger file renaming for existing files.
- Auto-rename will now only rename files in group libraries if the filename template is kept as default.
- Removed the rename option from the item(s) context menu if any of the selected attachment files already use a name according to the template. This adds two async operations per selected item, both potentially touching the filesystem. I cannot perceive any increased latency on my system, but this might affect people on slower systems. I believe having this option always appear instead (even if it does nothing) would be safer.

dstillman:
https://forums.zotero.org/discussion/121832/retroactively-rename-previously-stored-files

chrome/content/zotero/renameFiles.xhtml
```diff
@@ -0,0 +1,25 @@
...
+	<vbox id="zotero-progress-box" style="padding:10px" flex="1">
+		<label id="status" control="progress-indicator" data-l10n-id="rename-files-is-renaming" />
+		<hbox align="center">
+			<html:progress id="progress-indicator" max="1" style="flex: 1" />
+			<button id="close-button" label="&zotero.general.close;" disabled="true" />
+		</hbox>
+	</vbox>
+    <script src="include.js" />
+    <script src="renameFiles.js" />
+</window>
```
dstillman, line 25: Instead of creating a new window just for this, we should either use `Zotero.ProgressQueue` (see `Zotero.ProgressQueues.create` in attachments.js and recognizeDocument.js) or turn chrome/content/zotero/integration/progressBar.xhtml into a generic progress-meter window.

dstillman: I guess `Zotero.ProgressQueue` is probably not right, since there's no reason to show thousands of rows for operations that should take a tiny fraction of a second.

dstillman:
@tnajdek: Sorry, can you rebase?

dstillman:
Sorry, thinking more about this, my suggestion to only apply this conditionally in groups for the default format string was a bad idea — that's too weird.

Auto-renaming and the format string really should be a per-group synced setting, configurable only by admins. We could probably add that in the group settings online fairly easily, and then consider making it configurable in the client in a future version. Communicating the settings for groups in the UI would get really complicated, though. Maybe we just add a line saying that group renaming settings are configured on zotero.org?

tnajdek:
To summarise, I’m going to do the following:

1. Convert `progressBar` into a generic, reusable component.
2. Avoid creating a new window; use the `progressBar` from step 1 instead.
3. Add a new "Rename Files Now" button in the Preferences window, rather than an option in the menu bar.
4. Add logic to disable the button from step 3 if it would have no effect.
5. Convert `attachmentRenameTemplate` from a preference to a synced setting.
6. Add a note in Preferences stating that auto-renaming applies only to My Library; group settings can be configured on zotero.org.
7. For My Library, continue using the local preferences: `autoRenameFiles`, `autoRenameFiles.fileTypes`, `autoRenameFiles.linked` (and the hidden preference `autoRenameFiles.onMetadataChange`).
8. For group libraries, check the synced settings `autoRenameFiles` and `autoRenameFilesFileTypes` before renaming.

@dstillman If I’ve missed anything, please let me know.

tnajdek:
`progressBar` seems to be tightly coupled with the citation window and includes features that don't really make sense in this context (e.g. centering on the current screen, Zotero icon, fixed width). I started working on it but then realised that `Zotero.showZoteroPaneProgressMeter` might actually be more appropriate here?

dstillman:
Oh, yeah, I guess that's fine. That blocks the window while it's running, but that's probably appropriate anyway. It probably delays syncing too.

dstillman:
Instead of having "Rename File from Current Metadata" appear conditionally in the context menu based on the pref, which seems confusing, we should probably just do what @windingwind has suggested and move that function to a button next to the filename field in the attachment's item pane. The button could then be disabled when the filename matches the current rename string. And this would hopefully help decouple that action from the attachment title in people's names and reinforce the idea that files are automatically renamed.

(We could potentially hide it entirely if the pref is on and the filename matches, but even with the pref on there could be situations (like syncing) where files appear without automatic renaming applying, so maybe better to establish that that's always where you'd be able to go to rename the file if possible.)

dstillman:
Oh, and we'll need an icon from @yexingsha for the manual rename button. I think @windingwind had mocked something up with a sync-like icon but I think we want something different from "syncing".

yexingsha:
The icon `rename` which is already being used for "Renaming File from Current Metadata" should work fine for the manual rename button.

tnajdek:
Just wanted to make sure everyone is on the same page regarding the button. For me, this icon suggests that I’m going to input the new name manually, whereas what actually happens is that the name is automatically generated based on the parent item’s metadata.

We refer to this as a “manual” rename because, in most cases, all files are renamed automatically whenever the user changes the rename template—in which case, this button appears disabled.

I can’t really think of a good way to convey this within the limits of a 16px icon (maybe 20px at most). Perhaps replacing the pen with a cog could work?

Screenshot 2025-04-24 at 00 58 58

dstillman:
Yeah, pencil isn't right. I don't think cog is either — that really implies settings. We can do a sync circle on top of the rectangle in a pinch, but we should try some other ideas here (not that I have any).

yexingsha:
Ah, sorry I misunderstood what "manual" means. In that case, I'm also thinking of some form of rectangle + sync: 

rename-auto-1
rename-auto-2
rename-auto-3
rename-auto-4
rename-auto-5
rename-auto-6

I'm leaning toward 3) or 4), since the symbol of rectangle containing a vertical line is a familiar one, used for switching between single/double creator fields. The other options seem either too simple or too complicated.

dstillman:
Yeah, let's try 3.

tnajdek:
It no longer makes much sense to trigger file renaming for group libraries when a template in preferences changes, since that only applies to the user's library. It could potentially be triggered when the synced setting is changed (for a group and then it would only rename files within that group)?

Do we want to decide on this now, or leave it until the synced setting is actually editable on the web?

Other than that, I think this is pretty much ready for another review.

dstillman:
> It could potentially be triggered when the synced setting is changed (for a group and then it would only rename files within that group)?

Hmm, no, we wouldn't want to trigger this simultaneously on every member's computer. So making this configurable on the web doesn't really make sense (unless we were going to actually implement the renaming via the web library, which would definitely cancel out any time we saved not building UI for this in the client!).

So I guess the question is whether we can get away with not implementing the UI for groups right away. It would basically be a regression in functionality, in that you'd no longer be able to use a custom format string for groups, but that's only been an option for a short while anyway, and it was conceptually muddled (in that every group member's setting would apply to the same group), so maybe it's fine. If people complain, we can say that we'll restore it properly in a future update.

tnajdek:
> unless we were going to actually implement the mass-renaming via the web library

This is doable—the web library already supports filename changes, so the missing piece is porting the templating engine.

The drawback is that the web library cannot perform this operation atomically: if the user closes the tab (despite the warning we now display), loses connection, runs out of battery, etc., the operation may end up in an undefined state (with some files renamed and others not). The same limitation applies to other long-running operations in the web library (e.g. copying items to another library). We can use a synced setting that we toggle before the mass-renaming operation and reset afterwards (similar to `autoRenameFiles.done` pref) to somewhat mitigate this problem. I think this could work.

> So I guess the question is whether we can get away with not implementing the UI for groups for now. 

If we do, we should probably hide the "rename from parent" button in the attachments pane in group libraries for now.

dstillman:
Well, I don't think my suggestion to do it in the web library really made sense — it seemed like a quick solution, but we already have the file type settings and the format previews in the client, and people will expect to configure this there. So I think we'll just need to plan to move these options together under a drop-down with library names (with `Zotero.Utilities.Internal.buildLibraryMenu()`), and they'll be disabled if you're not the admin (which I think we'll need to add sync support for). But we can do that later.

>> So I guess the question is whether we can get away with not implementing the UI for groups for now.
>
> If we do, we should probably hide the "rename from parent" button in the attachments pane in group libraries for now.

Because it won't match the personal setting? Maybe better to have it use the synced setting (if present, which it won't be for now) or the default format, so that people can at least use that to keep filenames updated after editing metadata.

tnajdek:
> Maybe better to have it use the synced setting (if present, which it won't be for now) or the default format, so that people can at least use that to keep filenames updated after editing metadata.

OK, that's the current implementation (there was a bug where the button would never be disabled, but I’ve fixed that).

Currently, the "Rename From Parent" button is also disabled if the file type is not on the `autoRenameFiles.fileTypes` list (or a matching synced setting for group libraries), which I'm not sure is correct?

@dstillman If there's anything else you can think of or anything that's not right, please let me know.

dstillman:
> Currently, the "Rename From Parent" button is also disabled if the file type is not on the autoRenameFiles.fileTypes list (or a matching synced setting for group libraries), which I'm not sure is correct?

Yeah, that's not correct. That's just for automatic renaming. You should still be able to use the manual option.

dstillman:
Trying this out, I've changed my mind — I feel like we actually should just hide the button when it's a no-op instead of making it disabled. In the default case, with auto-renaming enabled on all your computers, it should never actually be needed, so it's taking up space for no reason. It's also fairly hard to discern whether it's disabled, so I think it would be clearer to have a button that appears when it would do something and then just disappears once the file is renamed.

tnajdek:
> Trying this out, I've changed my mind — I feel like we actually should just hide the button when it's a no-op instead of making it disabled.

Sure thing, done!

dstillman:
They're both kind of weird, but maybe we want the button (when it appears) to be aligned with the top line rather than the whole field?

<p></p>

<p></p>

(Latter not precise)

Not pictured, but there's also the Reindex button a couple lines down, which is aligned with the label and single-line value.

@yexingsha, what do you think?

We do the top line for the delete button next to notes that span multiple lines:

<p></p>

dstillman:
One sort of surprising thing is that anything that triggers a modify on an item results in a rename, even if it's not actually changing metadata — including (as I just discovered while I was testing the above) adding a note to an item. But I guess that's fine? If you have this setting on, I guess you want the file stay in sync with the metadata, and if it somehow got out of sync, any opportunity to notice and fix it is fine?

(I discovered this because I forced the file to get out of sync by renaming it outside of Zotero and doing Show in Finder to trigger a relocate.)

tnajdek:
Fixed the alignment, rebased & resolved conflicts. (I'll take a look at #5268 separately)

dstillman:
Hmm, so we're not currently showing any sort of prompt on upgrade when this is enabled, right?

I think we need to show a prompt explaining that filenames will now be automatically kept in sync as you change item details, with an option to turn off automatic renaming. (I can help with the exact text.) Otherwise we're going to start renaming people's existing files without warning, including after fairly unrelated changes like tagging or filing items into collections.

I don't know that we need to offer to do a mass-renaming at upgrade time, but the "Rename Files Now" button maybe needs to also show up in the main section of the prefs — only on first upgrade and when otherwise relevant, if you've declined the prompt to do it? — rather than just the format section.

dstillman:
It seems like you also need to be able to run "Rename Files Now" if you change the file types that automatic renaming applies to. And maybe if you disable "Automatically rename files" and then re-enable it? Because you could've made changes to metadata in the meantime, right?

dstillman:
We've probably done a prompt before, but yeah, we've only done banners in recent memory. I figured this might be a little hard to explain in a banner, but maybe it's fine:

> Zotero now automatically keeps filenames in sync as you make changes to items. [Settings] [×]

Would only appear on upgrade and only if renaming was already enabled.

We could also have a separate [Learn More] link to https://www.zotero.org/support/file_renaming, but maybe OK to just link to there from the settings section.

chrome/content/zotero/zoteroPane.js
```diff
@@ -6626,6 +6596,34 @@ var ZoteroPane = new function()
...
 
+	this.showFileRenamingBanner = function () {
+		if (Zotero.Prefs.get('autoRenameFiles.bannerShown')) {
+			return;
+		}
+		if (Zotero.Prefs.get('autoRenameFiles.done')) {
+			// If user had a custom `attachmentRenameTemplate` (now a synced
+			// setting), `autoRenameFiles.done` was set to `false` during
+			// migration. Otherwise, it's `true` (the default), and no banner
+			// is needed.
```
dstillman, line 6607: Wait, why isn't it needed just because they didn't have a custom format? We want to show this for all upgrading users who have automatic renaming on, because we want them to know that filenames will now be changing automatically even after they're first added.

`bannerShown` also isn't a default pref, so even once `true` it will be saved forever in people's `prefs.js`, and clearing the prefs will cause it to appear again.

 Seems like we should make `bannerShown` a default pref, set it `true`, and then in pref migration set it to `false` if `autoRenameFiles` is set to true. We can then remove this block and just show when `bannerShown` is `false`.

dstillman:
OK, upon reflection, instead of "Open Settings", I think we should just have a "Learn More" link to https://www.zotero.org/support/file_renaming. Most people shouldn't need to do anything here, so we shouldn't encourage everyone to start messing around in the settings.

(`general.learnMore` is available in zotero.properties, but we can create a `general-learn-more` in zotero.ftl.)

chrome/content/zotero/zoteroPane.js
tnajdek: You're right, that makes much more sense, fixed!

tnajdek:
> instead of "Open Settings", I think we should just have a "Learn More" link

Absolutely, it's done!

tnajdek:
> Maybe it should only automatically rename files on an item change if the previous name matches the previous metadata?

I did a bit of experimenting, and unless I’m missing something, this would actually require a fairly fundamental change. Currently, we wait for the `modify` notification for an item and perform the rename then. At that point, the item data is already changed, and there's no access to the previous version or any kind of diff/patch of what has changed. So it’s not possible to tell whether the current filename would have matched what the metadata prior to the change.

I don’t believe we store this information anywhere, nor do we currently trigger a `pre-modify` event. Again, if I'm missing something, please let me know.

I can see a few possible solutions to this:

1. **Store a flag** per attachment item indicating that the filename has been updated based on the current metadata. A manual edit would clear that flag.
2. **Introduce a `pre-modify` event** to allow tracking what’s about to change.
3. **Track recent changes**, like we would if "undo" was supported, and query for the most recent change to compare.

Option 1 seems the most straightforward, but I’m not sure what the preferred mechanism is for storing arbitrary per-item data. 

@dstillman Any suggestions on the preferred approach—or general direction—would be much appreciated.

dstillman:
> At that point, the item data is already changed, and there's no access to the previous version or any kind of diff/patch of what has changed.

Check the `extraData` argument to the notifier event (after `ids`).

dstillman:
Because I broke it a decade ago, apparently. Try 2bc0c4861c35d08ceff25723a6f8d0e7c77842b4.

tnajdek:
I implemented the logic to rename based on metadata changes only if the previous filename matched the previous metadata, but that won't work without #5297.

dstillman:
OK, let's try rebasing.

tnajdek:
Thanks! I rebased this PR and did a quick check to confirm everything’s still working as expected.

dstillman:
Problem: If the file doesn't exist locally (e.g., because you're using on-demand downloading), it isn't renamed.

For single files (e.g., PDFs), I think we can change the filename in the DB anyway, and when you go to download the file, it will automatically be saved with that name. That should even be the case for a file from WebDAV where there's a single file with a specific name within the ZIP.

I don't think we can do that for old multi-file snapshots, because we wouldn't know what the previous primary filename was in order to change it to the new name. (Generally there should only be one .html file, but it's possible there were some weird cases where multiple .html files were saved?) And unfortunately that may mean we can't do it for snapshots at all, because I don't think we know whether a snapshot was created via SingleFile without the file(s) being present.

I'm going to study the file syncing behavior here a bit anyway — I'm seeing some weird behavior, and I suspect we can skip the unnecessary file upload attempts to begin with, even for WebDAV — so if you go ahead and have it change the attachment item's filename regardless when `!item.isSnapshotAttachment()`, I'll do further testing to make sure it's working properly.

dstillman:
Well, and we don't auto-rename snapshots anyway, right? At least unless someone has changed the file-types pref manually to include text/html?

tnajdek:
> Well, and we don't auto-rename snapshots anyway, right? At least unless someone has changed the file-types pref manually to include text/html?

Correct, it would have to be done manually using the advanced config editor.

tnajdek:
@dstillman If the file is missing on disk, it will now auto-rename the filename in the database on metadata change and during mass-renaming (e.g. after a template change). Since we disable the input for manual renaming when the file is missing, I don’t think we should display the "Rename from Parent" button either—correct?

dstillman:
We need a tooltip for this button:

Screenshot 2025-08-27 at 3 00 23 AM

tnajdek:
> We need a tooltip for this button:

I've pushed a commit that adds a tooltip "Rename File to Match Parent Item".

I'm not sure if that's the best way to phrase it, this button is more about "syncing" the name of the file to match the template but I think anything with a word "sync" could be confusing here and hopefully people who see this button already know about the template.

tnajdek:
@dstillman I’ve implemented a new dialog listing planned file renames and adjusted the labels on the buttons to reflect that it now opens a new dialog instead of making the changes immediately. Also, since we now open a dialog, I’ve used that to show progress while files are being renamed. Let me know what you think! 

https://github.com/user-attachments/assets/0e4bef18-75e5-4c25-86d4-733b4fe6c24a

dstillman:
That certainly works, but maybe better to show pairs vertically so we can show more of the full names and changes are clearer due to the vertical alignment? Not sure if that would necessitate "Old: " and "New: " before each one or we could do some sort of stylistic difference (e.g., gray for old and black for new).

@yexingsha Any quick thoughts here?

yexingsha:
Yes, I agree that vertical pairs with `fill-secondary` for old and `fill-primary` for new would be better.

tnajdek:
Thank you both for suggestions, here is how it looks like now:

Screenshot 2025-09-03 at 12 38 59

Does that work, or should there be some sort of separation between entries?

yexingsha:
I think putting 8px gap between entries should work.

tnajdek:
Sure, that's done:

Screenshot 2025-09-04 at 13 58 40

yexingsha:
Huh, why is the color sequence alternating now?

tnajdek:
> Huh, why is the color sequence alternating now?

Ha, sometimes I can’t see the wood for the trees. Thanks, fixed!
Screenshot 2025-09-05 at 01 08 02

dstillman:
This needs a bit more styling — at least some background and/or border to the list:

Screenshot 2025-09-05 at 2 43 50 AM

dstillman:
Updated prompt:

Screenshot 2025-09-05 at 2 40 27 AM

Updated banner:

Screenshot 2025-09-05 at 3 21 38 AM

tnajdek:
I’ve tweaked the colours to introduce contrast between the table and the dialog; however, I’m not sure if that’s the right approach, since other dialogs in Zotero don’t do that. I’ve also increased the padding and made the separator between items adapt to the preferred layout density (8px for compact, 12px for comfortable). Suggestions welcome.


Screenshot 2025-09-05 at 11 30 51

Screenshot 2025-09-05 at 11 30 56

yexingsha:
The colors are now correct, though the list still needs a border using `color-panedivider` (see: select citation style).

It still looks weird now because it seems that there is an extra 10px gap between every single line, both between the old and new version and between different files. Remove that and it should look okay.

The Rename Files button should use the primary style.

dstillman:
I also don't think we need to follow Compact/Comfortable for this one-off list. Density seems more important here.

tnajdek:
OK, thank you both for the input. I’ll override the behaviour of the Virtualised Table here to always use compact-style rows and remove the extra spacing introduced in “comfortable” mode so it looks the same in both cases.

tnajdek:
I’ve disabled compact/comfortable support so the table will look the same regardless of the density setting, and I’ve added the border. The Rename Files button correctly uses the accent color (the window wasn’t in focus in the previous screenshots).


Screenshot 2025-09-05 at 21 49 12
Screenshot 2025-09-05 at 21 49 37

yexingsha:
Looks to me there is still 4px extra gap between every line. But otherwise this looks good.

tnajdek:
> Looks to me there is still 4px extra gap between every line. But otherwise this looks good.

Yes, it’s a 5px gap, the default padding for rows in Virtualised Table in compact mode. I’ve removed all padding now:

Screenshot 2025-09-05 at 22 29 37

PR #3868, opened 2024-03-20
collection filter: include children of matches

### LINKED ISSUE #3838: Include collapsed subcollections of matching collections?
https://forums.zotero.org/discussion/112728/zotero-7-beta-navigate-to-sub-collections-from-filtered-collections

I guess we could consider this? A matching collection with non-matching subcollections would always be collapsed to start — or collapsed if already collapsed in the tree? — but you could expand manually.

I forget the details of how we're preserving the pre-filtered open state of the tree, and it's not totally clear how this would affect that — if you open a matching collection, does it stay open when you clear the search? — but maybe there's an easy way to do this.

abaevbog:
- subcollections of matching collections are included in collectionTree after filtering
- if a matching collection row has no matching children, it is collapsed
- that row can be expanded to reveal all of its subcollections that appear as context rows 
- parents of all matching collections are expanded to preserve behavior similar to current
- now that the collections end up being expanded and collapsed, after filtering we restore the original state of collections, regardless if the selection changed. The path to the selected row is always expanded so that it is visible

No changes to keyboard navigation (?)

Fixes: #3838

### DIFF
--- a/chrome/content/zotero/collectionTree.jsx
+++ b/chrome/content/zotero/collectionTree.jsx
@@ -105,8 +105,8 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 		this._filterResultsCache = {};
 		this._filterInitialScrollPosition = null;
 		this._filterInitialCollapsedRows = [];
+		this._filterSelectedCollectionRow = null;
 		this._treeWasFocused = false;
-		this._hiddenFocusedRow = null;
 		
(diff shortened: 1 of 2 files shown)


abaevbog:
https://github.com/zotero/zotero/assets/36271954/04eff2aa-feb5-47f4-8cd8-71ced40935b7

Issue #3852, opened 2024-03-15
Right-click behaving like a left-click in many places


dstillman:
https://forums.zotero.org/discussion/112801/zotero-7-beta-right-click-behaving-like-a-left-click-in-many-places

abaevbog:
I can reproduce it on windows only - neither macOS nor linux have the same behavior. It seems to boil down to the `click` event firing on buttons even on right-click for some strange reason. The good thing is that `event.button` tells us which button was clicked.

It would be very unfortunate to have to add `if (event.button !== 0) return;` to every `click` event handler, so I will try to come up with something better

abaevbog:
And another observation: this seems to only happen to `toolbarbutton`s that are part of a component that has a `contextmenu` listener ... Right-click on the `-` button of tags triggers `click` event but right-click on `-` of notes or related section (that do not have `contextmenu` listeners) does not.

It does go a bit farther that that. As mentioned in the forums discussion, right-click on the collectionTree and itemTree also acts just like a click by changing the selected row but I think it might be easier to focus on the `toolbarbutton` undesired click first.

@AbeJellinek, I guess a simple and naive thing to try would be to override the `click` handler of default `toolbarbutton` component, to essentially do `if (event.button !== 0) return;` before running the rest of the default handler. Do you know if something similar is done with any other component, so that I have an example to look over?

abaevbog:
> maybe we could just replace the click listeners with command listeners?

This would certainly work! My concern though is that every time a new `toolbarbutton` listener is added, one has to remember that it must be `command` and not click. Not sure about anyone else but the chance of me forgetting about it in a couple of weeks is fairly high:)

I tried to capture events like this in `ZoteroPane.js` and it seems to give us the same behavior as macOS or linux:
```
document.addEventListener("click", (e) => {
	if (e.target.tagName == "toolbarbutton" && e.button !== 0) {
		e.target.dispatchEvent(new MouseEvent('contextmenu', {
			bubbles: true,
			clientX: e.clientX,
			clientY: e.clientY
		}));
		e.stopPropagation();
		e.preventDefault();
	}
}, true);
```
I guess, in a way that is the overriding that I was thinking of at first.  And if we don't want context menu to appear on right-click on `toolbarbutton`s at all, we can remove the `dispatchEvent` part. Does that seem reasonable to you?

abaevbog:
[...]

PR #3871, opened 2024-03-21
Right-click does not behave as left-click and does not change table selection

abaevbog:
- On windows, right-click on `toolbarbutton` when its parent has `contextmenu` listener triggers `click` event on the actual `toolbarbutton`. To avoid that, capture click events and if they were caused not by left-button click, ignore them and dispatch `contextmenu` event on the event's target.
- Do not change selection of on right-click in `virtualTable`  similar to thunderbird https://forums.zotero.org/discussion/comment/458310/#Comment_458310:
  - on right key down in virtualized table, record the receiving table row as `contextMenuRow`, and mark it with a `accent-azure` color (maybe a different color or out line is better)
  - `contextMenuRow` is cleared when the main `menupopup` is hidden
  - when a selected row is fetched in `itemTree`/`collectionTree` or `ZoteroPane`, return the `contextMenuRow` if it exists. It ensures that the selected action will run on the right-clicked row and not the row currently selected
  - some actions run only after the confirmation from a dialog. By default, `contextMenuRow` wants to be cleared when the `menupopup` is hidden. To make sure it remains until the dialgo is closed, create a promise that is awaited in `popuphidden` handler, which is resolved when the dialog is confirmed/cancelled. ZoteroPane has a wrapper function through which all dialogs are opened which creates and releases the promises.
  - minor tweaks to run `ZoteroPan`e functions when the selection is not empty OR when there is a context menu row

Addresses: #3852

### DIFF
--- a/chrome/content/zotero/collectionTree.jsx
+++ b/chrome/content/zotero/collectionTree.jsx
@@ -1242,7 +1242,7 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 	 * @returns {Promise<void>}
 	 */
 	async deleteSelection(deleteItems) {
-		var treeRow = this.getRow(this.selection.focused);
+		var treeRow = this.getSelectedRow();
 		if (treeRow.isCollection() || treeRow.isFeed()) {
 			await treeRow.ref.eraseTx({ deleteItems });
 		}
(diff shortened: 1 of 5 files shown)


abaevbog:
@dstillman, the second commit is the setup to not change selection in `virtualized-table` on right click. This likely still requires some tweaks and a lot more testing because it affects almost everything is `ZoteroPane`, now that the selected row is not the row that is always used to run actions on. But I think the general idea is here, and I wanted to check if this is the right direction? My main intention is to somewhat replicate what [thunderbird does does](https://forums.zotero.org/discussion/comment/458310/#Comment_458310)

abaevbog:
For reference, this is what the right-click behavior looks like:

https://github.com/zotero/zotero/assets/36271954/85d76a3a-444e-4917-ae6b-0b0dd9cd3090

PR #3876, opened 2024-03-24
Fix Zotero_Preference#_showPane event target

windingwind:
Fix the wrong event target of `showing`. This bug was introduced with the automatic <h1> of preference panes.

### DIFF
--- a/chrome/content/zotero/preferences/preferences.js
+++ b/chrome/content/zotero/preferences/preferences.js
@@ -328,7 +328,17 @@ var Zotero_Preferences = {
 		let pane = this.panes.get(id);
 
 		pane.container.hidden = false;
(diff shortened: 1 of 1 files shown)


chrome/content/zotero/preferences/preferences.js
```diff
@@ -328,7 +328,17 @@ var Zotero_Preferences = {
 		let pane = this.panes.get(id);
 
 		pane.container.hidden = false;
-		pane.container.children[0].dispatchEvent(new Event('showing'));
+		// The automatic <h1> is introduced by commit 784f877
```
dstillman, line 331: Shouldn't reference a specific commit — can just explain what this is doing

windingwind:
Oh, I didn't notice the `load` is dispatched on all children. Make sense.

Issue #2321, opened 2022-02-02
HTML tree: Missing tooltip for long fields


dstillman:
When a column value is truncated, there should be a tooltip with the full text. See the XUL tree for reference.

adomasven:
Closed with https://github.com/zotero/zotero/commit/060d2f1c37d50eec9214875b75b9af4fa957e66b

dstillman:
This wasn't about column headers — those shouldn't have tooltips (unless, perhaps, they're also truncated). It's just about long values in tree rows. See the Firefox bookmarks manager (or Zotero 5.0) for an example.

adomasven:
Oh, sorry. The "column value" threw me off in your initial comment. What you're referring to are more precisely described as cell values.

dstillman:
Right — sorry for the confusion.

PR #3877, opened 2024-03-25
fx115: fix tooltips for tabs and tabs menu

abaevbog:
- move 'fake tooltip' so that it appears not only when the library tab is opened
- manually handle tooltip with `Zotero_Tooltip` in tabs menu


adomasven:
You should search the commit history for "tooltip" on this repo, and see commits like https://github.com/zotero/zotero/commit/060d2f1c37d50eec9214875b75b9af4fa957e66b

Basically you create a xul `<tooltip id="some-id">` element somewhere (you can even create it with JS), then set `tooltip=some-id` attribute on a xul node that hosts the html elements (or maybe can even be the top xul node), and the tooltip behavior is generally handled by Mozilla code after that. At least that's how it worked before. I'm not sure `Zotero_Tooltip` was even necessary to exist at all, we figured out how this stuff works sometime later after it was added in the early days of starting to mix XUL with HTML. It might still work.

There's probably a good reason to make this logic generic somewhere as an init function, since now you manually have to perform this tooltip setup every time you create a new XUL-HTML dialog/interface.

adomasven:
E.g. this is how VirtualizedTable, wherever it is embedded ensures tooltips for it are always displayed https://github.com/zotero/zotero/blob/90282b3739f91b7edf059e7d31d8bec2a89d0c3c/chrome/content/zotero/components/virtualized-table.jsx#L1048-L1076

abaevbog:
> Basically you create a xul <tooltip id="some-id"> element somewhere (you can even create it with JS), then set tooltip=some-id attribute on a xul node that hosts the html elements (or maybe can even be the top xul node), and the tooltip behavior is generally handled by Mozilla code after that. 

Thank you for the explanation! There were indeed a few different approaches (`Zotero-Tooltip`, `fake-tooltip`, `html-tooltip` ... ) to tooltips handling, so this really clarified things.

> I'm not sure Zotero_Tooltip was even necessary to exist at all, we figured out how this stuff works sometime later after it was added in the early days of starting to mix XUL with HTML. It might still work.

I agree! My impression is that `Zotero_Tooltip` is not needed in all of the places where it is being called now. I had to move the main `html-tooltip` so that it is visible within non-library tabs but besides that, adding the `tooltip="html-tooltip"` to the XUL parent sorta magically solved all the issues from before! I removed all calls to `Zotero_Tooltip`, added `title` attributes, and the tooltips appear everywhere (tags actions button, tabs bar, tabs menu, all buttons in the reader) as they should. So that's great and the first commit should address the tooltip issues with fx115.

At this point `Zotero_Tooltip` could even be completely removed. But when I was following the discussion from https://github.com/zotero/zotero/issues/2321, I had an idea that maybe it could be helpful there. In my second commit I tried to work around https://github.com/zotero/zotero/commit/8e2790e2d2a1d8b15efbf84935f0a80d58db4e44 that prevents events from firing on the cells by manually finding the cell that the mouse is over and displaying the tooltip with `Zotero_Tooltip`. Just changing the `title` on the row wasn't enough but with `Zotero_Tooltip` the overflowing cells show the tooltip for me.

I know this is a somewhat brute-force approach but I thought it's worth a try until a there's a more elegant solution. If this idea makes sense, we could keep the actual `Zotero_Tooltip` file around until that workaround is no longer needed. Otherwise, we could remove `Zotero_Tooltip` or at least I can add a detailed comment explaining the tooltips setup and saying that it should not be used. What do you think?

adomasven:
> At this point `Zotero_Tooltip` could even be completely removed. But when I was following the discussion from #2321, I had an idea that maybe it could be helpful there. In my second commit I tried to work around [8e2790e](https://github.com/zotero/zotero/commit/8e2790e2d2a1d8b15efbf84935f0a80d58db4e44) that prevents events from firing on the cells by manually finding the cell that the mouse is over and displaying the tooltip with `Zotero_Tooltip`. Just changing the `title` on the row wasn't enough but with `Zotero_Tooltip` the overflowing cells show the tooltip for me.

Does removing "pointer-events: none" in fx115 still make it impossible to drag unselected rows? That would be the ideal solution.

Otherwise, VirtualizedTable is used across Zotero (including ItemTree and VirtualizedTree) in various interfaces. We should come up with a solution that doesn't require the user of these components to know to add `fake-tooltip` to the xml. There is likely no avoiding having to include `tooltip.js though`, and probably that code should be responsible for adding the tooltip element.

adomasven (commented):
> the mouseover effects don't fire on the actual cells of the table, so the
tooltips do not appear on them.

They are pointer events, as opposed to the mouseover events. The former are browser handlers exactly for things like tooltips (and also changing the cursor appearance based on the type of element being hovered, like links, grippies), while the latter are JS events.

dstillman:
> They are pointer events, as opposed to the mouseover events. The former are browser handlers […], while the latter are JS events

I haven't been following this whole discussion, but just on this point, I'm not sure what "browser handlers" is supposed to mean, but pointer events are just mouse events for the touch era. They're all DOM events. (PointerEvent inherits from MouseEvent.)

abaevbog:
> Does removing "pointer-events: none" in fx115 still make it impossible to drag unselected rows? That would be the ideal solution.

Yeah, that was the first thing i tried and, unfortunately, it still prevents dragging on non-selected rows.  

>  We should come up with a solution that doesn't require the user of these components to know to add fake-tooltip to the xml. 

That is a great point. One also shouldn't have to load `Zotero_Toolbar` via `Services.scriptloader.loadSubScript("chrome://zotero/content/tooltip.js", this);` in the `.xhtml` file to properly use the `virtualized-table`. So with my initial setup, the tables in advanced search or preferences would not work without it.

I revised this to load `Zotero_Tooltip` in virtualized table as a module similar to how utils are loaded `const { Zotero_Tooltip } = require('./tooltip');`. We also create the tooltip and add it as a child of `document.documentElement` if it does not yet exist on `Zotero_Tooltip.start()`. While working on importing `tooltip.js`, I had to move it into the `components` directory, otherwise the import wouldn't go through (not sure why). I thought it's ok since `virtualized-table` is the only place where the fake tooltip is now used, so it makes sense if they're next to each other.

So with the last change, tooltips appear for me in `itemTree` and `collectionTree`, as well as in `advancedSearch` window and preferences without adding anything extra into the `.xhtml` files.

adomasven:
> I haven't been following this whole discussion, but just on this point, I'm not sure what "browser handlers" is supposed to mean, but pointer events are just mouse events for the touch era. They're all DOM events. (PointerEvent inherits from MouseEvent.)

We're talking about [CSS pointer-events](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events), but even more reason for clarified naming.

PR #3878, opened 2024-03-25
Added fulltext search cache

tisoft:
Introduced a fulltext search cache, that improves performance by storing previously searched text patterns and their corresponding matches, reducing redundant file accesses and search operations.


tisoft:
Note: I have not yet created tests for this. Will do so, if/when the general idea is accepted.

tisoft:
Discussion on zotero forums: https://forums.zotero.org/discussion/113100/fulltext-regexp-search-is-slow-for-big-pdf-collections

Wells-XW:
Can you share what you do? This is very useful for users with big collections!

PR #3880, opened 2024-03-25
Fix dates not following Zotero's locale.

### LINKED ISSUE #3559: Date format not following English locales
https://forums.zotero.org/discussion/comment/451855/#Comment_451855

E.g., English (UK) still uses the American date format, whereas it was D/M/Y in Zotero 6.

tnajdek:
Behaviour of `toLocaleString` and `toLocaleDateString` seems to have changed between Firefox 68 and 102. Below is the same code running on my system configured to `en-GB` with Zotero configured to `en-US`.

Zotero 6:
Screenshot 2024-03-25 at 13 44 18

Zotero 7:

Screenshot 2024-03-25 at 13 44 59

From what I can observe `toLocaleString` no longer respects locale configured in Zotero and instead uses system's locale. I've also checked current Firefox/Chromium and it seems to exhibit the same behavior.

To fix this, I've tweaked our code to explicitly call `toLocaleString` with `window.navigator.language` as the first argument. A similar treatment might be needed in zotero/utilities.

Resolves #3559

### DIFF
--- a/chrome/content/zotero/dateOverrides.js
+++ b/chrome/content/zotero/dateOverrides.js
@@ -1,0 +1,18 @@
+/* eslint-disable no-extend-native */
+const { Services } = ChromeUtils.import("resource://gre/modules/Services.jsm");
+
+let originalToLocaleString = Date.prototype.toLocaleString;
+Date.prototype.toLocaleString = function (locales, options) {
+	if (locales === undefined || (Array.isArray(locales) && !locales.length)) {
+		locales = Services.locale.requestedLocales;
+	}
+	return originalToLocaleString.call(this, locales, options);
+};
+
+let originalToLocaleDateString = Date.prototype.toLocaleDateString;
+Date.prototype.toLocaleDateString = function (locales, options) {
+	if (locales === undefined || (Array.isArray(locales) && !locales.length)) {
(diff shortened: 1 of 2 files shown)


tnajdek:
I slightly prefer current approach as it's more explicit way but I don't have strong feeling either way.

Couple of reasons for my approach:

* We would have to do it for `toLocaleString` and `toLocaleDateString`
* I believe that in the items tree we should be using an instance of `Intl.DateTimeFormat` (to potentially save some CPU cycles). That would still need locale passed explicitly.

Your solution obviously has an advantage of less code changes and helping future us not having to remember adding the argument explicitly every time we call one of these methods.

dstillman:
I think overriding the prototype is fine. This is really just our playground, so having it behave how we will always want it to behave seems fine.

dstillman:
(The bigger question here is actually whether there's any way that when using Automatic we can follow a configured system date format separately from the system locale setting. That's a longstanding request. But I'm not sure browsers expose that.)

tnajdek:
Updated to override Date's prototype. I've tried my best to pick the right place where this should exist, let me know if I should move the logic (or the file) somewhere else.

Curiously it doesn't affect `eval` (i.e., doesn't work in runJS).

Issue #3881, opened 2024-03-26
fx115: tagsbox autocomplete is blank in reader tab


abaevbog:
On `fx115`, the autocomplete for `tags-box` in the context pane of the reader appears blank, though it does work - the list is refreshed as you type as I see in the dev tools.

<img width="365" alt="Screenshot 2024-03-26 at 2 47 17 PM" src="https://github.com/zotero/zotero/assets/36271954/cfc6d694-e736-4a67-9676-ac78fe6fd833">

The culprit seems to be [this part](https://github.com/zotero/zotero/blob/4934e9e63d19bfd1c3d1222817d089d9ba469735/chrome/content/zotero/elements/tagsBox.js#L79-L96) that creates the popup from https://github.com/zotero/zotero/commit/49fe2b98d97523174889ec86b9925e4e795ca034. Just removing those lines seems to fix the issue for me.

@AbeJellinek I am trying to figure out why we needed to restore the autocomplete popup? I thought mozilla kinda handles it for us? At least, after I removed that part, I don't see any breakage...

abaevbog:
Oooh, I see. This thing.

<img width="307" alt="Screenshot 2024-03-26 at 3 24 24 PM" src="https://github.com/zotero/zotero/assets/36271954/21a6b640-0f61-43c7-9024-61d194ac397d">

It is interesting - with `fx115`, the behavior of elements ordering has changed a bit. I've been looking into somewhat similar issues with tooltips not appearing properly. Anyways, I'll try to play around with this to see what we can do!

abaevbog:
Ok, I think I have something. With fx115, the position of the popups in the DOM really matters. The part in the `tagsBox` adds the autocomplete `panel` into the first `popupset` in `<vbox id="zotero-pane">`, and with fx115, this causes an issue if this popup needs to appear in another pane of the `<deck id="tabs-deck">`. If we just remove that part, mozilla will add a new `popupset` at the bottom of the DOM. This fixes the "blankness" but it causes autocomplete appearing below the "Add tags" popup again. Another weird behavior there is that the auto-created autocomplete `panel` does not have an id, so mozilla would add a new `popupset` with autocomplete `panel`  every single time autocomplete is called. 

But ... if we manually add a `popupset` with a `<panel id="PopupAutoComplete"` to the bottom of the DOM, mozilla will reuse it and properly position everything.

[...]

PR #3882, opened 2024-03-26
fx115: fix tagsbox autocomplete breakage

abaevbog:
With fx115, changes from 49fe2b98d97523174889ec86b9925e4e795ca034 made the autocomplete popup appear blank when the reader tab is opened. Not re-creating the PopupAutoComplete in tagsBox makes the items from autocomplete popup visible again. But it makes the autocomplete appear below the "Add tags" popup for annotations (what 49fe2b98d97523174889ec86b9925e4e795ca034 originally fixed), and it also creates a duplicate `popupset` with a new autocomplete panel at the bottom of the DOM every time autocomplete runs. The duplicate `panel`s do not have an id.

Manually adding a `popupset` with a `panel` whose id="PopupAutoComplete" does not create any duplicates and properly positions the popup.

Fixes: #3881

Screenshot 2024-03-26 at 4 20 29 PM
Screenshot 2024-03-26 at 4 20 50 PM

### DIFF
--- a/chrome/content/zotero/elements/tagsBox.js
+++ b/chrome/content/zotero/elements/tagsBox.js
@@ -75,26 +75,6 @@
 				removeAllItemTags.disabled = !this.count;
 				this._id('tags-context-menu').openPopupAtScreen(event.screenX, event.screenY, true);
 			});
-			
-			if (!document.getElementById('PopupAutoComplete')) {
-				let popupset = document.querySelector('popupset');
-				if (!popupset) {
-					popupset = document.createXULElement('popupset');
-					document.documentElement.append(popupset);
-				}
-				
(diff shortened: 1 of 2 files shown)


abaevbog:
Awesome!

The only thing I want to point out here is that there's a somewhat cryptic error that gets logged when autocomplete is accepted ... sometimes.
`NS_ERROR_FAILURE: Component returned failure code: 0x80004005 (NS_ERROR_FAILURE) [nsIAutoCompleteInput.popup]` and it comes from `LoginManagerChild.sys.mjs (resource://gre/modules/LoginManagerChild.sys.mjs)` which does not event seem relevant to autocomplete at all.

I don't think it has anything to do with this fix because I see it in `fx115` as well. It doesn't seem to affect the actual autocompletion and it also does not always happen. I made a separate issue to keep track of it but still I thought it's worth mentioning

Issue #3885, opened 2024-03-27
fx115: quick-format dialog layout broken


abaevbog:
<img width="798" alt="Screenshot 2024-03-27 at 10 00 19 AM" src="https://github.com/zotero/zotero/assets/36271954/17e18ebe-60b2-4eb9-9b20-c072765ae73e">

abaevbog:
To add to that, on windows, dragging the dialog doesn't move the reference panel anymore ... If you move the dialog, the panel will remain as it was.

<img width="1141" alt="Screenshot 2024-03-27 at 11 00 47 AM" src="https://github.com/zotero/zotero/assets/36271954/cbd87f2a-2f67-48f8-9fba-ed477684537a">

abaevbog:
On mac, the panel generally follows the dialog on drag. However, instead of always being below it, it can do this:

https://github.com/zotero/zotero/assets/36271954/5d5cb4e4-bbc2-4bbf-9b01-7453d0224149

PR #3889, opened 2024-03-27
fx115: fixes to qf reference panel positioning

abaevbog:
- Added `flip="none"` to reference panel to stop it from opening above the dialog or sliding through it.
- Hide and show panel after editor resizing on all platforms, since now it the panel find itself in the wrong spot on mac as well
- On linux and windows, when the window is dragged, the reference panel remains where it is and does not follow the dialog. To prevent the panel from "breaking away" from the main window, hide it when the drag starts and reopen when dragging is done
- On windows, the mouse often "outruns" the window when it is being dragged. If that happens, the dialog stops moving and it can miss the 'mouseup' event. In that case, the panel won't reopen and the dialog will start being dragged on next 'mouseover' without clicking until the next 'mousedown' event. This has been an issue before fx115 as well but if panel opening/closing is a part of the window drag logic, it's best to have it addressed. There should be a better fix to make dialog dragging work better on windows overall but this is a start.

Fixes: #3885 

The only new commit is the last one - the first 4 were cherry-picked from `master` as they're not in `fx115` yet and I'll remove them when `fx115` is rebased again.

@adomasven, what do you think about hiding the reference panel for the duration of the window drag as a solution to stop the panel from separating from the dialog?


adomasven:
Holding reviewing this off while we discuss https://github.com/zotero/zotero/issues/3847, particularly removing the red border and making the citation dialog a normal dialog window with a title bar.

abaevbog:
So I revised this based on what we discussed and here I have:
- a general commit with `fx115` sizing fixes
- a commit where I replaced the old `WindowDraggingElement` with `-moz-window-drag`. As I was testing things on windows, it really used to not work that great so might as well try it now. There were some specific cases that needed handling, so there are details in the commit message. 
- as I was fixing up the positioning on linux, I wondered why don't we try to at least somewhat make the dialog look the same on linux as on other platforms. I hid the title and basically tried to recreate the appearance of the dialog on win and macOS, now that the red border seems to remain (at least for now)?

I did not include the change of `<panel>` for `<vbox>` here yet. I believe it would still worth it to avoid dealing with `<panel>` specifics and to not hide/display the `panel` when the dialog is dragged. Wasn't sure if it makes sense now?

Finally, I do not know if the refresh-rate issues is involved here. I did notice that some values of `chromemargin` lead to the dialog disappearing and the first `window.resizeTo()` also lead to the window being shrunk after I added margins around the dialog on windows. I had those tweaked and it doesn't happen now. 

Actually, do we know if the refresh-rate issue only happened when there were opened tabs/selected items, meaning the `<panel>` would auto-open? Or does it also happen if the `<panel>` does not auto-open when there are no selected items? Because in the former case, the `<panel>` opening would be most likely the culprit and we could try to wait for longer before the initial `<panel>` opening.

chrome/content/zotero/integration/quickFormat.js
```diff
@@ -1437,6 +1433,7 @@ var Zotero_QuickFormat = new function () {
 	 */
 	function _onProgress(percent) {
 		var meter = document.querySelector(".citation-dialog.progress-meter");
+		meter.style.visibility = 'visible';
```
adomasven, line 1490: Why not in CSS or HTML style attribute?

chrome/content/zotero/integration/quickFormat.js
```diff
@@ -1650,6 +1701,19 @@ var Zotero_QuickFormat = new function () {
 		return { lastBubble: lastBubble, startOfTheLine: startOfTheLine };
 	}
 	
+	// Make window draggable when mouse is over the dialog's wrapper
+	function _onDialogMouseOver(e) {
+		dialog.style["-moz-window-dragging"] = e.target == dialog ? "drag" : "no-drag";
+	}
```
adomasven, line 1710: Why do this instead of
```css
.entry {
    -moz-window-dragging: drag;
}

.deck {
    -moz-window-dragging: no-drag;
}
```
And then avoid dragtracker?

chrome/content/zotero/integration/quickFormat.js
```diff
@@ -70,8 +70,13 @@ var Zotero_QuickFormat = new function () {
...
-			} else if(Zotero.isWin) {
-				document.documentElement.setAttribute("hidechrome", true);
+			}
+	
+			// Required for dragging to work on windows.
+			// These values are important and changing them may affect how the dialog
+			// is rendered
+			if (Zotero.isWin) {
+				document.documentElement.setAttribute('chromemargin', '0,0,15,0');
 			}
```
adomasven, line 80: How do you reproduce whatever issue this is solving?

adomasven, line 80: How do you reproduce whatever issue this is solving?

adomasven:
The linux changes do not work on my Unity DE. 

image
(this is what the full dialog window looks like, not a cutoff screenshot)

I tried messing around with the CSS to make it work, but could not. I'm not sure if it's due to X11 vs Wayland, or something more specific to my DE. When originally testing the initial redesign on at least a few DEs I noticed that some of them do not actually support transparent rounded window corners, I believe at least Linux Mint was another one, making me think this is an X11 issue. Both TB and FX also have the same issue with rounded corners of the main window, although not as pronounced. You can change the original Ubuntu install in a VM to run in X11 mode via the gear icon on the login screen to test this.

adomasven:
With a few item bubbles in the dialog, when navigating through them with the keyboard, the reference panel keeps flash-reapearring. Not sure if it's related to fx115 or these changes.

chrome/content/zotero/integration/quickFormat.js
abaevbog: If I do `document.documentElement.setAttribute('chromemargin', '0,2,2,2');` (same as the main zotero pane window) then I see the dialog disappearing like [this](https://www.dropbox.com/scl/fi/9dyp9n52ml03uohz7nu36/panel_gone.mov?rlkey=wcp8csgprfg8oety9ec4yg1uh&dl=0) on windows 11 VM. 

My observation is that changing the first digit (top margin) to anything but `0` as well as decreasing the bottom margin to lower than `15` (I tried 10, 5 and 2) causes such the breakage.

chrome/content/zotero/integration/quickFormat.js
abaevbog: Oh yeah, no reason to not do that. Thank you. I was trying a bunch of different things trying to figure out why the drag wouldn't start closer to the edges of the window - that's where this came from

abaevbog:
> Is this a fx115 thing or due to some change here? If it's due to a change, can we not do the reverse and do whatever we did on macOS to make the reference box follow the window?

This is what I observed after `fx115`. Pre-`fx115`, the `<panel>` would always stick close to the dialog. After `fx115`, though, the `<panel>` would refuse to move higher when the editor is shrunk. Since this is something that was already handled on windows, I just removed the extra `if` condition to do it for `macOS` as well.

This is what it avoids:
Screenshot 2024-04-02 at 9 32 12 AM

This issue and the need to handle it will go away in the future if we `vbox`-ify the references panel.

abaevbog:
> With a few item bubbles in the dialog, when navigating through them with the keyboard, the reference panel keeps flash-reapearring. Not sure if it's related to fx115 or these changes.

Thank you for pointing it out. The infinite-loop workaround for the window drag was fighting over the state of the reference panel with other listeners. I refactored it a bit so it should be gone now.

chrome/content/zotero/integration/quickFormat.js
abaevbog: Totally fair - I moved this logic into the stylesheet

abaevbog:
> The linux changes do not work on my Unity DE.

How about now? This is what the dialog looks like for me now on ubuntu with X11.

linux_dialog

Making corners rounded is indeed a challenge - the background keeps being white instead of being transparent. I'm sure if there is a robust enough way to do it with our current setup...

One thing I tried playing around in dev tools is to wrap the entire dialog into a `<panel>` as it was earlier suggested. The actual window would be empty and just invisible. We can then set the panel's padding variable that indeed seems to make the panel properly rounded. This does lead to a whole new set of glitches and behaviors to re-consider (nested `<panels>` seems like a recipe for future problems), so I'm not sure if it makes sense at this point. But it is a possibility to explore for the future. 

rounded_dialog

dstillman:
Square corners on Linux (is it just Linux?) are fine for now if that avoids problems.

chrome/content/zotero/integration/quickFormat.js
adomasven: So this kind of sucks. There's too much hackery interfering with itself in this dialog. But let's just make this work for fx115 and I will look to make it less complicated for the redesign.

adomasven:
We'll try to make the rewrite work with the panel following the window and less hackery. I imagine we can make it simpler rewriting this from scratch. I think we can go with this for fx115.

abaevbog:
@adomasven, I think I stumbled upon the thing that caused the dialog to blink and disappear on windows with "wrong" `chromemargin` values. `background: transparent` from `integrations.css`. Removing that line allowed the dialog to be draggable without requiring any esoteric `chromemargin` values and setting `chromemargin=0,2,2,2` (or any other value as far as I can tell) does not lead to the dialog disappearing anymore. So that's some progress! I pushed a temporary commit with some code removed to make it clear what I tried. Does it work with your windows setup? And does uncommenting `background: transparent` lead to the dialog acting weird again and going away like below?

Screenshot 2024-04-03 at 1 46 43 PM

Though the issue now is that the window is resizable and no amount of `resizable=false/no/0` stops it... I'm trying to figure out why now.

abaevbog:
So I have not been able to "turn off" resizing of the window without `background: transparent`. It seems like this may be somewhat intentional based on this [long bugzilla thread](https://bugzilla.mozilla.org/show_bug.cgi?id=177838). Having a resizable window is not good at all for our current setup but I do think it's worth trying out the QF without `background: transparent` because if it caused display issues for me, maybe it is somehow at the root of the refresh-rate issue? 

I did notice that the window is resizable if `drawintitlebar` is `true` and the other way around. That seems to be the general behavior for all windows (e.g. adding `drawintitlebar=true` to "New collection" window makes it resizable). So I thought what if we add a hidden preference that will display QF without `background: transparent` and with a titlebar (to avoid resizing), so that we can ask folks affected by high-refresh rate issue to check if the dialogs render better if we do not set transparent background? 

My main reasoning is that `chromemargin` having to be a very specific value doesn't make much sense and likely does not properly address whatever issue is there. So I would not be surprised if it just didn't work with some setups/on some machines same way it worked on my VM. That way, if it works - great. If not - there is a workaround that likely helps. But most importantly, it would give us a way to check if transparent background is indeed the culprit without affecting more people than necessary. What do you think?

It is not unlikely that there is some mozilla-specific mechanism that I overlooked, so correct me if that is so. The last two commits are a bit messy but I think it's easier to refer to things I tried that way - I'll clean them all up in the end.

adomasven:
> I did notice that the window is resizable if `drawintitlebar` is `true` and the other way around. That seems to be the general behavior for all windows (e.g. adding `drawintitlebar=true` to "New collection" window makes it resizable). So I thought what if we add a hidden preference that will display QF without `background: transparent` and with a titlebar (to avoid resizing), so that we can ask folks affected by high-refresh rate issue to check if the dialogs render better if we do not set transparent background?

It doesn't seem like this is something that we want or could use in the eventual solution, even if it turned out that it would fix the issue for those users. We also know that such windows, like the library or Scaffold windows indeed display properly to users with `<panel>` use.

> My main reasoning is that `chromemargin` having to be a very specific value doesn't make much sense and likely does not properly address whatever issue is there. So I would not be surprised if it just didn't work with some setups/on some machines same way it worked on my VM.

There is that risk, but I checked on my Windows 10 and 11 VMs, as well as a native Windows 10 install and this works fine. I kind of don't like it because it's "unclean", and the only way to figure out what's actually causing the need for the chromemargin is to try to rebuild this dialog from scratch, which we will be doing anyway, but otherwise, if it works - it works. Now is the time to test in the Beta.

I was happy to say this was ready two commits ago. Unless you think we can make it work without the need to draw a native titlebar, I think there's no use in exploring that avenue.

abaevbog:
Got it, I removed the last 2 commits. It doesn't seem like there's a way to avoid resizing without the titlebar so I guess we can go with this for now.

In any case, it's good to know that transparent background on the window can have undesired effects like this for the future updates to QF.

One final thing I did was remove the `windowDraggingUtils` file - it's not used anywhere now and there's probably no need to keep it around.

adomasven:
Digging through the mozilla codebase for other stuff, I noticed this https://searchfox.org/mozilla-central/source/dom/chrome-webidl/XULPopupElement.webidl#33-75

@abaevbog could calling the openPopup function by providing the anchor element been a possible way to fix this?

PR #3893, opened 2024-03-28
Fix error when dealing with invalid certificates

### LINKED ISSUE #3892: No `errorMessage` in `nsITransportSecurityInfo`
It seems that in fx102 we can no longer depend on `errorMessage` here: https://github.com/zotero/zotero/blob/489cee6b2411b425be4cb3a6cfeda31b7c2db6c2/chrome/content/zotero/xpcom/http.js#L1273

because it's no longer present in that object. Instead, logic for that message [moved in Firefox](https://searchfox.org/mozilla-central/source/toolkit/content/aboutNetError.mjs#1146) (with messages [in ftl](https://searchfox.org/mozilla-central/source/toolkit/locales/en-US/toolkit/neterror/netError.ftl)). Ideally we would create/adapt messages for possible `errorCodeString` but I think for now we can just come up with a single message that includes the error code string.

tnajdek:
Fix #3892 

This is the simplest solution with no additional messages. I think it works well enough, given that previously error message also included the raw string error code.

Before:
Screenshot 2024-03-28 at 17 26 32

After:
Screenshot 2024-03-28 at 17 33 15

### DIFF
--- a/chrome/content/zotero/xpcom/http.js
+++ b/chrome/content/zotero/xpcom/http.js
@@ -1268,7 +1268,7 @@ Zotero.HTTP = new function() {
 			secInfo.QueryInterface(Ci.nsITransportSecurityInfo);
 			if ((secInfo.securityState & Ci.nsIWebProgressListener.STATE_IS_INSECURE)
 					== Ci.nsIWebProgressListener.STATE_IS_INSECURE) {
-				msg = Zotero.getString('networkError.connectionNotSecure', Zotero.appName) + (secInfo.errorCodeString ? ` ${secInfo.errorCodeString}` : '');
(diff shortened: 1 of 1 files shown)


dstillman:
Hmm, this doesn't look right, though — `networkError.connectionNotSecure` ("Zotero could not make a secure connection.") is already used as the header, so it doesn't need to appear again. It looks like that was already the case for the second case, but that was maybe just totally wrong and untested, since it was also missing the app name parameter.

dstillman:
Seems like we can just do that?

tnajdek:
> Seems like we can just do that?

Yes, that's better, no point in repeating the header.

PR #3895, opened 2024-03-28
editable-text: Clean up `render()`

AbeJellinek:
Things were getting a bit out of hand with the anonymous listeners, and this should be clearer and more maintainable in the future.

Event handlers extracted with no code changes. Moved one line in `render()` to make the conditions that set `focused` to true clearer.

### DIFF
--- a/chrome/content/zotero/elements/editableText.js
+++ b/chrome/content/zotero/elements/editableText.js
@@ -185,111 +185,21 @@
 				}
 				input.classList.add('input');
 				input.toggleAttribute("no-windows-native", true);
-				let handleInput = () => {
-					if (!this.multiline) {
-						this._input.value = this._input.value.replace(/\n/g, ' ');
-					}
-					this.setAttribute('value', this._input.value);
-				};
-				let handleChange = (event) => {
-					if (Services.focus.activeWindow !== window) {
-						event.stopPropagation();
-					}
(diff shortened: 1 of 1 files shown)


chrome/content/zotero/elements/editableText.js
```diff
@@ -185,111 +185,21 @@
...
+				input.addEventListener('change', this._handleChange);
+				input.addEventListener('focus', this._handleFocus);
+				input.addEventListener('blur', this._handleBlur);
+				input.addEventListener('keydown', this._handleKeyDown);
+				input.addEventListener('mousedown', this._handleMouseDown);
+				input.addEventListener('dragover', this._handleDragOver);
+				input.addEventListener('drop', this._handleDrop);
 				
-				let focused = false;
+				let focused = this._input && document.activeElement === this._input;
```
abaevbog, line 197: The old `focused` calculation was 100% unnecessarily convoluted. This looks identical to [the focused getter at the bottom](https://github.com/zotero/zotero/blob/f9e25675f0cc4060dc59e2471d08f1fa45cd1099/chrome/content/zotero/elements/editableText.js#L383) to me - could `this.focused` be used here instead? That way, we'd have one less variable floating around

abaevbog (commented):
I think this is a big improvement for readability! Everything looks good to me - just one question about the `focused` piece below

PR #3900, opened 2024-03-29
Improve error message in the importer

tnajdek:
* New messages for connection/certificate problems
* Change error message color to better fit after the redesign

Not sure about the wording of the messages, especially don't like how it looks when printed at the default wizard width (below) but couldn't come up with anything better (and this should be something people very rarely see!)

Screenshot 2024-03-29 at 18 00 53

### DIFF
--- a/chrome/content/zotero/import/importWizard.js
+++ b/chrome/content/zotero/import/importWizard.js
@@ -256,9 +256,11 @@ const Zotero_Import_Wizard = { // eslint-disable-line no-unused-vars
 			}
 			catch (e) {
 				const feedbackEl = document.getElementById('mendeley-online-login-feedback');
+				feedbackEl.textContent = '';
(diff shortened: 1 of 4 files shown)


dstillman:
There seems to be a race condition where the error message is removed:

https://github.com/zotero/zotero/assets/783082/e65ef712-95b9-4894-966e-ef2bf2c20992

Maybe only when Return is used to trigger the login? I didn't see it when clicking Next.

dstillman:
(Also I just realized that badssl.com is dead and serving expired certificates for many of the other error conditions.)

tnajdek:
@dstillman appreciate the feedback & tweaks here and in #3903.

I've fixed the problem with error message being removed too eagerly when confirming with keyboard and also adapted the code to work after c16f5d15, i.e., removed the new message and re-used one from the exception thrown by `http.js`:

Screenshot 2024-04-02 at 18 37 53

PR #3903, opened 2024-04-01
fx115: Fix detecting insecure connection no longer works

tnajdek:
It seems in fx115 this no longer works:

https://github.com/zotero/zotero/blob/751457268bd84e6c49d29044ca86d804db75cb13/chrome/content/zotero/xpcom/http.js#L1269C9-L1269C31

I'm getting generic "Error connecting to server. Check your Internet connection." for expired/revoked certificate. From what I can see `secInfo.securityState` is always 0, even though `errorCodeString` might actually be populated.

Looking for a solution I've found [`SecurityInfo` utility](https://searchfox.org/comm-central/source/mozilla/toolkit/components/extensions/webrequest/SecurityInfo.sys.mjs#27) which seems to contain the logic we need, although it doesn't seem to return value for `errorMessage` contradictory to what the comment says so we still need to use `errorCodeString`.

Any suggestions welcome.

### DIFF
--- a/chrome/content/zotero/preferences/preferences_sync.jsx
+++ b/chrome/content/zotero/preferences/preferences_sync.jsx
@@ -145,11 +145,7 @@ Zotero_Preferences.Sync = {
 		}
 		catch (e) {
 			setTimeout(function () {
-				Zotero.alert(
-					window,
-					Zotero.getString('general.error'),
-					e.message
-				);
+				Zotero.Sync.Runner.alert(e);
 			});
 			throw e;
 		}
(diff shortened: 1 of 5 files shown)


dstillman:
Oh, wait, I think this isn't right. `state == 'insecure'` happens for regular interrupted connections, so we don't want to show a security warning for those:

https://searchfox.org/mozilla-esr115/rev/c9c3c42c22a2653c522136d5109b74df1772b270/toolkit/components/extensions/webrequest/SecurityInfo.sys.mjs#73-83

dstillman:
So I'm going to change this to only throw when `state` is `broken`.  A status `0` for a regular interrupted connection will be addressed elsewhere, and we don't currently warn about weak ciphers or anything like that.

PR #3920, opened 2024-04-03
Parse dates in item tree

### LINKED ISSUE #3917: Show Date field values in uniform format if they parse cleanly
https://forums.zotero.org/discussion/113358/function-to-change-date-format-globally and many other discussions over the years, but I can't find a ticket for this.

AbeJellinek:
`strToDate()` and `Intl.DateTimeFormat` can be a little heavy, but this is cached at the row level. We could consider caching at the item level instead if we want to show parsed dates in the Info box like we do for `accessDate`.

If the date fails to parse, it's displayed unchanged.

Fixes #3917

### DIFF
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -3449,23 +3449,59 @@ var ItemTree = class ItemTree extends LibraryTree {
 			case 'dateModified':
 			case 'accessDate':
 			case 'date':
-				if (key == 'date' && !this.collectionTreeRow.isFeedsOrFeed()) {
-					break;
-				}
 				if (val) {
-					let date = Zotero.Date.sqlToDate(val, true);
-					if (date) {
-						// If no time, interpret as local, not UTC
-						if (Zotero.Date.isSQLDate(val)) {
-							date = Zotero.Date.sqlToDate(val);
-							val = date.toLocaleDateString();
+					if (key != 'date' || this.collectionTreeRow.isFeedsOrFeed()) {
+						let date = Zotero.Date.sqlToDate(val, true);
+						if (date) {
+							// If no time, interpret as local, not UTC
+							if (Zotero.Date.isSQLDate(val)) {
+								date = Zotero.Date.sqlToDate(val);
+								val = date.toLocaleDateString();
+							}
+							else {
+								val = date.toLocaleString();
+							}
 						}
 						else {
-							val = date.toLocaleString();
+							val = '';
 						}
 					}
 					else {
-						val = '';
+						// key == 'date' and we aren't in a feed
+						// Try to parse the date
+						let { year, month, day } = Zotero.Date.strToDate(val);
+						// See strToMultipart() - discard year if it contains a suffix
+						if (!/^\d{1,4}$/.test(year)) {
+							year = undefined;
+						}
+						year = parseInt(year);
+						if (isNaN(year)) {
+							year = undefined;
+						}
+						month = parseInt(month);
+						if (isNaN(month)) {
+							month = undefined;
+						}
+						day = parseInt(day);
+						if (isNaN(day)) {
+							day = undefined;
+						}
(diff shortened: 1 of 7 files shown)


chrome/content/zotero/itemTree.jsx
```diff
@@ -3002,23 +3002,59 @@ var ItemTree = class ItemTree extends LibraryTree {
...
 						else {
-							val = date.toLocaleString();
+							val = '';
 						}
 					}
 					else {
-						val = '';
+						// key == 'date' and we aren't in a feed
+						// Try to parse the date
+						let { year, month, day } = Zotero.Date.strToDate(val);
```
dstillman, line 3025: `strToDate` returns anything it doesn't recognize in `part`. I think we want to fail parsing if that has a value.

chrome/content/zotero/itemTree.jsx
```diff
@@ -3002,23 +3002,59 @@ var ItemTree = class ItemTree extends LibraryTree {
...
+							year = undefined;
+						}
+						month = parseInt(month);
+						if (isNaN(month)) {
+							month = undefined;
+						}
+						day = parseInt(day);
+						if (isNaN(day)) {
+							day = undefined;
+						}
```
dstillman, line 3041: strToDate seems to already return integers for month and day, so I don't think these last two are necessary?

chrome/content/zotero/itemTree.jsx
```diff
@@ -3002,23 +3002,59 @@ var ItemTree = class ItemTree extends LibraryTree {
...
+						if (year !== undefined || month !== undefined || day !== undefined) {
```
dstillman, line 3042: We want to use the raw value if there's no year. Currently if there's, say, a time in the Date field (`09:32:53`), it will show just a number (`9`), but when sorting by Date that row will be among rows without date fields. (This would also be taken care of by failing parsing when there's a `part` value, though — in this example, part would be `32:53`.)

dstillman:
Oh, and I think we will want to show the parsed values in the Info pane as well.

dstillman:
Hmm, maybe we actually need two levels of strictness, where known-but-unsupported things are ignored for the items list but not for the info pane. Unknown things would still fail parsing for both. I agree that `2013-12-05 03:30:01` and `2013-12-05T03:30:01Z` should be displayed as `12/5/2013` in the items list, but it doesn't seem right to hide the timestamp in the info pane. I'm not sure whether something like `Summer 2004` should show as `2004` in the items list, but it should definitely show as `Summer 2004` in the info pane. But `Dec. 5, 2013` should be displayed as `12/5/2013` (or some user-configurable format) in the info pane.

One path here would be to improve strToDate parsing to recognize and return more things (e.g., timestamps), so that we could then still fail on `part`, but probably the better goal here is to start parsing into valid EDTF strings and actually storing those instead of our SQL-style "multipart" strings… But we would then [parse the field as EDTF](https://github.com/inukshuk/edtf.js) and could then decide which of the supported components could be ignored in each context.

But this is obviously a larger discussion, since it will involve syncing issues, etc.

dstillman:
> I think we'd want to make it controllable by a flag passed to strToDate, since that's used by translators and other code that won't necessarily expect timestamps to be parsed out.

It looks like only RIS and RefWorks Tagged actually use `part`, so it's probably enough to just update those to use any new properties if they exist.

> Re UI: should we style parsed and raw dates differently in the Info pane?

I don't think that's necessary. I think the ultimate goal is to *store* dates in a uniform, ambiguous format, among other things because it will avoid the awkward bug now where ambiguous dates can be parsed differently depending on your locale (even for different people in a group!). We wouldn't mass-convert existing dates, but that would at least be the default case going forward, and the raw dates (in ISO8601 or EDTF) wouldn't matter for those.

AbeJellinek:
@dstillman: Should I rebase this and work on improving the parsing? Seems like a nice feature to follow the item field changes.

dstillman:
Sure.

Issue #3883, opened 2024-03-26
fx115: autocomplete error message


abaevbog:
Selecting an entry from the autocomplete popup logs the following error:
```
NS_ERROR_FAILURE: Component returned failure code: 0x80004005 (NS_ERROR_FAILURE) [nsIAutoCompleteInput.popup]
```
But it does not seem to affect the functionality of autocomplete in any way

dstillman:
https://github.com/zotero/zotero/pull/3882#issuecomment-2025969205

If it's that, we can just add a `return;` from fetch_xulrunner.sh.

PR #3925, opened 2024-04-04
do not trigger event causing autocomplete error

abaevbog:
DOMInputPasswordAdded event is triggered on autocomplete submission which causes an error in LoginManager.
The error gets logged in the console, though the autocomplete does work. This just removes the event from ActorManagerParent.

Fixes: #3883

### DIFF
--- a/app/scripts/fetch_xulrunner
+++ b/app/scripts/fetch_xulrunner
@@ -82,7 +82,7 @@ function remove_line {
 		egrep -v "$pattern" "$file" > "$file.tmp"
 		mv "$file.tmp" "$file"
 	else
-		echo "$pattern" not found in "$infile" -- aborting 2>&1
+		echo "$pattern" not found in "$file" -- aborting 2>&1
 		exit 1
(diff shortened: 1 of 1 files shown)


abaevbog:
That is a great question! I began by adding a `return` to the beginning of the `_maybeRecordTelemetryEvents` based on https://github.com/zotero/zotero/pull/3882#issuecomment-2025969205 but it kept logging the error. My next step was to go through all the events in `LoginManager` and see if there's anything that would stop the error from appearing and removing `DOMInputPasswordAdded` helped. Since the error wasn't actually interfering with anything and was more of an annoyance, I thought that's good enough...

I'm not sure why a login-related event is firing on autocomplete to begin with, to be honest

abaevbog:
One more observation I made is that the error seems to be logged only on `Enter` or click into the popup. Selecting an entry from autocomplete dropdown and clicking outside of it or pressing `tab` does not log the error. Not sure if this adds any more clarity though...

PR #3931, opened 2024-04-07
Fix richlistitem cut-off with double-byte characters

windingwind:


windingwind:
before:
image


after:
image

Issue #3939, opened 2024-04-09
Stop using `width`/`height` attributes when restoring `extensions.zotero.pane.persist` dimensions


dstillman:
They no longer work in fx115, so, e.g., the tag selector height isn't restored.

dstillman:
Reverted due to https://github.com/zotero/zotero/pull/3944#issuecomment-2046116709. We'll need to figure out what's going on here.

windingwind:
So, has anyone figured out what breaks the layout after #3944?

It's probably some width/height value we persisted in previous versions that conflicts/overwrites the rules we introduced/modified in fx115. If anyone can reproduce, does clearing the persist prefs fix the layout issue?

PR #3944, opened 2024-04-09
fx115: fix element width/height persist

windingwind:
fix: #3939


dstillman:
I was initially seeing some weird behavior here where the items list got pushed under the item pane when expanding the collections pane, but I can't reproduce it anymore.

dstillman:
Had to revert this, since others were seeing the same thing.

PR #3952, opened 2024-04-10
vpat: 1 add role="link" to components acting as links

abaevbog:


dstillman:
Many of these are using our `zotero-text-link` custom element. Can we not define `role` on that, and fix the others that are just setting a class to use the custom element?

abaevbog:
> Many of these are using our zotero-text-link custom element. Can we not define role on that, and fix the others that are just setting a class to use the custom element?

That makes sense! The `role="link"` is added to `zotero-text-link` and I added `is="zotero-text-link"` wherever possible. Wherever not possible (e.g. for `span`s), I added the role directly as before.

To make the `label` in `attachmentBox` a `zotero-text-link`, I set a `tooltiptext` (which i think is good for links anyways) and added `noClick` options to `zotero-text-link`. That one is not currently used anywhere but there is `this.clickableLink` in `attachmentBox`, so I assumed it's just something for the future. I also removed the `keydown` listener from `zotero-text-link` to instead use the `keyboard-clickable` class for brevity.

PR #3956, opened 2024-04-10
Render RSS description as HTML

### LINKED ISSUE #3926: Feeds: Render RSS `description` as HTML
https://forums.zotero.org/discussion/comment/460097/#Comment_460097

And presumably whatever we're using from Atom as well.

(We'd obviously want to render as untrusted content.)

AbeJellinek:
Fixes #3926


dstillman:
Nice!

Can we set the viewport width to the pane width here?

Screenshot 2024-04-17 at 5 41 22 AM

(From https://pubs.acs.org/action/showFeed?type=axatoc&feed=rss&jc=achre4)

PR #3957, opened 2024-04-10
vpat 6: aria labels for buttons in item/context panes

abaevbog:
Added aria labels and button roles for buttons in the item/Collection pane. This includes buttons in the header as well as rows from notes, related, collections, and attachments section that act as buttons. Also added missing `aria-labelledby` for mini-itembox rows below attachment preview to have filename, dateModified and etc. properly labelled.

This also covers most of vpat 16.

This does not affect how tags/header/itemBox entries are announced. There is a separate issue with NVDA and JAWS sometimes struggling with announcing text inside of `editable-text` component, so this will be handled separately.

This also does not affect annotations since those are not keyboard accessible yet.

### DIFF
--- a/chrome/locale/en-US/zotero/zotero.ftl
+++ b/chrome/locale/en-US/zotero/zotero.ftl
@@ -6,7 +6,6 @@ return-or-enter =
     }
 
 general-remove = Remove
-general-new = Add new
 general-expand = Expand
 general-collapse = Collapse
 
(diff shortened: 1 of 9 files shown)


chrome/locale/en-US/zotero/zotero.ftl
```diff
@@ -370,6 +370,25 @@ section-related =
...
     .label = { pane-attachment-info }
 
+section-button-remove =
+    .tooltiptext = Remove
+section-button-add =
+    .tooltiptext = Add new
+section-button-expand =
+    .tooltiptext = Expand
+section-button-collapse =
+    .tooltiptext = Collapse
```
dstillman, line 380: Remove, Expand, and Collapse can be `general-` strings at the top of the file and then pulled in as variables.

abaevbog: For sure - I moved them to `general-*` variables at the top of .ftl file

PR #3966, opened 2024-04-11
Prevent duplicate main window

### LINKED ISSUE #3964: `zotero://select` links open new windows
https://forums.zotero.org/discussion/113580/bug-report-hyperlink-clicks-result-in-multiple-unwanted-windows-in-latest-beta

AbeJellinek:
- Set `cmdLine.preventDefault` in all cases. Some internal handler is opening the main window when it's already open. I don't know where that's happening! It's not in `nsDefaultCLH`, because that checks for an existing window. In any case, `preventDefault` prevents it.
- Update `Zotero.openMainWindow()` to work on fx115: Import `AppConstants`, which is no longer available globally in ES modules

Fixes #3964


adomasven:
Can we not fix this in `app/assets/commandLineHandler.js`?

PR #3969, opened 2024-04-11
vpat 7,14: popups count and cursor fixes

abaevbog:
- vpat_7: when a popup is showing, mark menuseparators with role="presentation" to clear whatever semantics a separator has. It prevents the screen readers from counting it as an interactable element while announcing the index of a menuitem within the popup. It also makes the screen readers count items across all sections, instead of stopping at the first separator.
- vpat_14: when the popup is hiding, delete the popup and immediately re-insert it into the DOM. It forces voiceover to move cursor back to the focused element. Otherwise, it is not aware that the popup went away, the cursor gets stuck on the invisible element, and if it happens, no elements will be announced until the screen readers is reloaded.

### DIFF
--- a/chrome/content/zotero/customElements.js
+++ b/chrome/content/zotero/customElements.js
@@ -176,22 +176,15 @@ Services.scriptloader.loadSubScript('chrome://zotero/content/elements/itemPaneSe
 				this.addEventListener("popupshowing", clearSeparatorAriaSemantics);
 
 				// If a menu closes with voiceover cursor in it, the cursor gets stuck in no-longer-visible
-				// menu and voiceover will be quiet until it is restarted. As a workaround, remove the menu
-				// from DOM to force voiceover to move its cursor and insert it back after a small delay;
+				// menu and voiceover will be quiet until it is restarted. Marking the menu
+				// as aria-hidden for a moment forces voiceover to shift its cursor.
 				this.addEventListener("popuphidden", (e) => {
 					if (this !== e.target || this.parentNode?.closest("menupopup")) {
 						return;
 					}
-					let parent = this.parentNode;
-					let sibling = this.nextSibling;
-					this.remove();
+					this.setAttribute("aria-hidden", true);
 					setTimeout(() => {
-						if (sibling) {
-							parent.insertBefore(this, sibling);
-						}
-						else {
-							parent.appendChild(this);
-						}
+						this.removeAttribute("aria-hidden");
 					});
 				});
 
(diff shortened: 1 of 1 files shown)


dstillman:
This causes errors and failures to open the popup when quickly right-clicking. I can reproduce easily by repeatedly right-clicking the mouse while moving horizontally on an item row.

```
Uncaught (in promise) TypeError: menu is null
     ZoteroPane</this.buildItemContextMenu< chrome://zotero/content/zoteroPane.js:3533
     tryCatcher resource://zotero/bluebird/util.js:16
     module.exports/PromiseSpawn.prototype._promiseFulfilled resource://zotero/bluebird/generators.js:97
     module.exports/Promise.coroutine/< resource://zotero/bluebird/generators.js:201
     captureStackTrace resource://zotero/bluebird/debuggability.js:915
     CapturedTrace resource://zotero/bluebird/debuggability.js:807
     longStackTracesCaptureStackTrace resource://zotero/bluebird/debuggability.js:482
     module.exports/Promise.prototype._then resource://zotero/bluebird/promise.js:261
     module.exports/Promise.prototype._passThrough resource://zotero/bluebird/finally.js:94
     module.exports/Promise.prototype.finally resource://zotero/bluebird/finally.js:103
     PromiseSpawn resource://zotero/bluebird/generators.js:36
     module.exports/Promise.coroutine/< resource://zotero/bluebird/generators.js:197
     ZoteroPane</this.onItemsContextMenuOpen chrome://zotero/content/zoteroPane.js:3099
     onContextMenu chrome://zotero/content/zoteroPane.js:1649
     onItemContextMenu chrome://zotero/content/itemTree.js:1035
     _handleContextMenu chrome://zotero/content/components/virtualized-table.js:733
     _renderItem/< chrome://zotero/content/components/virtualized-table.js:1154
```

abaevbog:
> This causes errors and failures to open the popup when quickly right-clicking. I can reproduce easily by repeatedly right-clicking the mouse while moving horizontally on an item row.

Oh, yeah, thank you for pointing it out. I realized that marking the menu as `aria-hidden` for a moment also forces voiceover to move the cursor without having to remove and re-insert the node. So now, nothing should interfere with existing menus, while still getting the cursor to come back from the menus:


https://github.com/user-attachments/assets/72626e1c-6c70-4c12-8984-6896cbbd3349

abaevbog:
While going over this, I also realized that the `menuseparator` semantics tweak should apply to all platforms, while `MozMenuPopupPrototype.ensureInitialized` is only applicable to mac. So I pulled it out into a separate listener.

PR #3971, opened 2024-04-12
Add `getData` in ItemPaneManager onInit callback props

windingwind:


windingwind:
This PR adds a `getData` to the `props` of the `onInit` callback of the ItemPane custom section API. (We had some discussion before and decided to not use `getData`, but now I think it's reasonable to add it in `onInit`)

This would allow, e.g., registering a notifier observer during `onInit` and determining whether or not to take action based on the custom section's `item`. However, the `item` isn't assigned during `onInit`, and may change in the future, so we need a `getData` function to retrieve the section data (mainly `item`) at runtime.

Issue #3975, opened 2024-04-12
Test failure: "should shift-tab through the toolbar to item-tree"


dstillman:
https://github.com/zotero/zotero/actions/runs/8660839222/job/23749524125

`./test/runtests.sh -g "should shift-tab through the toolbar to item-tree"`

I assume after https://github.com/zotero/zotero/commit/68596148b03332b0cf2e7d2e693d826d29a61e63

PR #3977, opened 2024-04-12
fix to focus test breakage in zoteroPaneTest

abaevbog:
- edits to zoteroPaneTest.js focus tests to expect the updated focus sequence: selected tab -> tabs menu -> sync button -> collectionTree toolbar -> collectionTree -> tags selector -> itemTree toolbar
- updated tags selector keydown handling to explicitly handle all tab/shift-tab events using moveFocus. It is more readable and explicit focus handling for all components is required for programmic tab/shiftTab events dispatched in tests to actually move focus

Fixes: https://github.com/zotero/zotero/issues/3975

### DIFF
--- a/chrome/content/zotero/components/search.jsx
+++ b/chrome/content/zotero/components/search.jsx
@@ -92,6 +92,7 @@ class Search extends React.PureComponent {
 					onChange={this.handleChange}
 					onKeyDown={this.handleKeyDown}
 					value={this.state.immediateValue}
+					className="search-input"
 					{...pick(this.props, p => p.startsWith('data-') || p.startsWith('aria-'))}
 				/>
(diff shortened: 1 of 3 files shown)


dstillman:
Oh, so there's a problem with focused tags: they break with windowing.

First, if you try to scroll with the keyboard, once windowing kicks in, the focused tag no longer exists, so scrolling stops working. But even if you scroll down and back up with the mouse, the focused tag no longer has focus, because it was removed and recreated. And really you should be able to scroll down past the window via the keyboard and press left/right-arrow and it should still focus the previous/next tag and automatically scroll back up to put it in view. Seems like we need to tie into the windowing system to do all this right.

We should split the test fix into a separate PR so we can get CI passing again, and we can deal with the tag stuff separately.

Issue #3974, opened 2024-04-12
Remove tab stop on tag selector itself


dstillman:
Do we need that if you can scroll up/down and page up/down when an individual tag has focus?

Particularly after https://github.com/zotero/zotero/pull/3954, we really want to avoid extra tab stops.

PR #3984, opened 2024-04-15
tag selector focus edits, fix windowing breakage

abaevbog:
- no tabstop on the tag selector scrollable area
- change tag selector's role from default "grid" to "group". "grid" is not quite correct semantically and leads to voiceover suggesting irrelevant commands.
- move all keyboard handling logic to tagSelectorList.jsx. Tabbing through tag selector is now handled in ZoteroPane, so the only logic left there is arrow navigation
between tags, and there's no reason to not have it together with the tags list.
- a workaround to deal with focused tags when windowing kicks in. When a tag is focused, record its index. Each time tags are re-rendered, if the saved index is not among rendered tags, refocus it, otherwise, move focus to the tags list. So if the tag is focused, you scroll away and then back, the tag will get re-focused and it will look like it never left. In arrowUp/Down handler, if the desired tag does not exist, keep scrolling
until the tag is found and focused before handling the navigation

Fixes: #3974
Also fixes the windowing breakage brought up in https://github.com/zotero/zotero/pull/3977#issuecomment-2053562097

### DIFF
--- a/chrome/content/zotero/components/tagSelector/tagSelectorList.jsx
+++ b/chrome/content/zotero/components/tagSelector/tagSelectorList.jsx
@@ -293,17 +293,16 @@ class TagList extends React.PureComponent {
 		}
 	};
 
-	handleKeyDown(e) {
+	async handleKeyDown(e) {
 		if (!["ArrowRight", "ArrowLeft"].includes(e.key)) return;
 		// If the windowing kicks in, the node of the initially-focused tag may not
 		// exist, so first we may need to scroll to it.
 		if (!document.activeElement.classList.contains("tag-selector-item")) {
-			this.setState({ scrollToCell: this.focusedTagIndex }, async () => {
-				// Even though the state was updated, the tags are not rendered yet.
-				// So we have to wait for handleSectionRendered to run before proceeding.
-				await this.waitForSectionRender();
-				this.shiftFocus(e);
-			});
+			this.setState({ scrollToCell: this.focusedTagIndex });
+			// Even after the <Collection> re-renders, the new tag nodes may not be rendered yet.
+			// So we have to wait for handleSectionRendered to run before proceeding.
+			await this.waitForSectionRender();
+			this.shiftFocus(e);
 			return;
 		}
 		this.shiftFocus(e);
--- a/chrome/content/zotero/components/tagSelector/tagSelectorList.jsx
+++ b/chrome/content/zotero/components/tagSelector/tagSelectorList.jsx
@@ -47,6 +47,7 @@ class TagList extends React.PureComponent {
 		this.scrollToTopOnNextUpdate = false;
 		this.prevTagCount = 0;
 		this.focusedTagIndex = null;
+		this.lastFocusedTagIndex = null;
 		this.keypressToHandleOnNextUpdate = null;
 		this.state = {
 			scrollToCell: null
@@ -186,6 +187,7 @@ class TagList extends React.PureComponent {
 			onDragExit,
 			onDrop,
 			onFocus: (_) => {
+				this.lastFocusedTagIndex = this.focusedTagIndex;
 				this.focusedTagIndex = index;
 			}
 		};
@@ -245,8 +247,13 @@ class TagList extends React.PureComponent {
 		// <Collection> sets role="grid" which is not semantically correct
 		tagsList.setAttribute("role", "group");
 
-		this.setState({ scrollToCell: undefined });
 		if (this.focusedTagIndex === null) return;
+		// If the focused tag does not changed, the scrollToCell won't change
+		// either, so the <Collection> won't scroll to the desired tag if we don't reset it.
+		// E.g. second arrowLeft keypress when first tag is focused won't scroll to it.
+		if (this.lastFocusedTagIndex === this.focusedTagIndex) {
+			this.setState({ scrollToCell: null });
+		}
 		// Check if the tag that is supposed to be focused is within the rendered tags range.
 		// If it is, make sure it is focused. If it is not - focus the tags list.
 		if (indices.includes(this.focusedTagIndex)) {
@@ -271,6 +278,7 @@ class TagList extends React.PureComponent {
 		let tagsList = document.querySelector('.tag-selector-list');
 		if (!tagsList.contains(event.relatedTarget)) {
 			this.focusedTagIndex = null;
+			this.lastFocusedTagIndex = null;
 		}
 	};
 
--- a/chrome/content/zotero/components/tagSelector/tagSelectorList.jsx
+++ b/chrome/content/zotero/components/tagSelector/tagSelectorList.jsx
@@ -47,7 +47,10 @@ class TagList extends React.PureComponent {
 		this.scrollToTopOnNextUpdate = false;
 		this.prevTagCount = 0;
 		this.focusedTagIndex = null;
-		this.renderedTagRange = [];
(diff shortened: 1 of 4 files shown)


chrome/content/zotero/components/tagSelector/tagSelectorList.jsx
```diff
@@ -214,6 +218,84 @@ class TagList extends React.PureComponent {
...
+				tagsList.scrollTop -= tagsList.clientHeight;
+			}
+			else {
+				tagsList.scrollTop += tagsList.clientHeight;
+			}
+			maxCounter += 1;
+			// Wait a moment to let this.handleSectionRendered update this.renderedTagRange
+			await Zotero.Promise.delay(10);
+		}
+	}
```
adomasven, line 247: This is wildly inefficient. See [Collection.md](https://github.com/bvaughn/react-virtualized/blob/master/docs/Collection.md). We should use the prop `scrollToCell` without the need to track rendered cells, etc. If that is infeasible, we should use `this.positions[this.focusedTagIndex].y` to calculate the scroll position.

abaevbog: Oh yes, thank you for pointing me in the direction of `scrollToCell`! This is way better.

Now, when we handle keyboard navigation and the desired tag is not focused, I update `scrollToCell` state to tell `<Collection>`  where to scroll and finish handling the event in `handleSectionRendered` after the desired tag receives focus. What do you think?

chrome/content/zotero/components/tagSelector/tagSelectorList.jsx
```diff
@@ -214,6 +221,70 @@ class TagList extends React.PureComponent {
...
+			this.focusedTagIndex = null;
+		}
+	};
+
+	handleKeyDown(e) {
+		if (!["ArrowRight", "ArrowLeft"].includes(e.key)) return;
+		// If the windowing kicks in, the node of the initially-focused tag may not
+		// exist, so first we may need to scroll to it.
+		if (!document.activeElement.classList.contains("tag-selector-item")) {
+			this.setState({ scrollToCell: this.focusedTagIndex });
```
adomasven, line 282: use `setState()` second param callback to continue handling of the event here instead of in `handleSectionRendered()`

chrome/content/zotero/components/tagSelector/tagSelectorList.jsx
```diff
@@ -214,6 +221,70 @@ class TagList extends React.PureComponent {
...
+			nextOne.focus();
+		}
+	}
+
+	handleSectionRendered = ({ indices }) => {
+		let tagsList = document.querySelector('.tag-selector-list');
+		// <Collection> sets role="grid" which is not semantically correct
+		tagsList.setAttribute("role", "group");
+
+		this.setState({ scrollToCell: undefined });
```
adomasven, line 248: I imagine this might not be necessary (and might cause an unnecessary rerender/call loop?), the element will only do something if scrollToCell is different from the previously provided one.

chrome/content/zotero/components/tagSelector/tagSelectorList.jsx
adomasven: https://react.dev/reference/react/Component#setstate:~:text=optional%20callback%3A%20If%20specified%2C%20React%20will%20call%20the%20callback%20you%E2%80%99ve%20provided%20after%20the%20update%20is%20committed.

chrome/content/zotero/components/tagSelector/tagSelectorList.jsx
abaevbog: Yeah this is a precaution to take care of a situation if the focused tag didn't actually change. For example, if you have the focus on the first tag, you scroll all the way down, press `arrowLeft` to scroll back to the first tag. After this, if you scroll down again and press `arrowLeft` again, we won't scroll back up because as far as the `<Collection>` is aware the `scrollToCell` prop has not changed. I suppose there can be more cases, especially if there are many disabled tags.

But now that I typed it, maybe it makes sense to just reset `scrollToCell` if the focused tag has not changed? I pushed that tweak, and I can't think of other cases when it would be required. That way, the edge case from  above is taken care of and we shouldn't re-render more often than needed

chrome/content/zotero/components/tagSelector/tagSelectorList.jsx
adomasven: That doesn't seem right. The `setState()` callback is called after rendering the state changes. Are you saying `Collection` double-renders? Once for scroll and once for rendering of tags?

If so, can we instead add a function like `waitForTagToRender()` that returns a promise which is resolved in `handleSectionRendered()`, such that all effects of the event handler can occur in the handler function/closure?

abaevbog: > EDIT or does some delayed rendering of actual Collection nodes (tags)?

That is my impression. At least, `console.log` of all tag nodes in the callback of `setState` gives us the "stale" tags before scrolling, while logging them in `handleSectionRendered` yields current tags after scroll happen. I guess it makes sense, otherwise, why would `onSectionRendered` prop need to exist?

I pushed a change where we wait for the promise resolved in `handleSectionRendered` before shifting focus - it is indeed more readable that way

chrome/content/zotero/components/tagSelector/tagSelectorList.jsx
```diff
@@ -214,6 +223,91 @@ class TagList extends React.PureComponent {
...
+		// exist, so first we may need to scroll to it.
+		if (!document.activeElement.classList.contains("tag-selector-item")) {
+			this.setState({ scrollToCell: this.focusedTagIndex }, async () => {
+				// Even though the state was updated, the tags are not rendered yet.
+				// So we have to wait for handleSectionRendered to run before proceeding.
+				await this.waitForSectionRender();
+				this.shiftFocus(e);
+			});
+			return;
+		}
```
adomasven, line 308: So ideally we simplify this further and inline `this.shiftFocus(e)`. You can do this by replacing a callback with a promise like

```js
await new Promise(this.setState({ scrollToCell: this.focusedTagIndex }, () => resolve));
await this.waitForSectionRender();
```

abaevbog: Right, no need for a callback if it's not necessary. In fact, I think we might not need to `await` the first promise at all, since we'll be `await`-ing for the `handleSectionRendered` which happens strictly after `setState`. So what if we do just :
```
if (!document.activeElement.classList.contains("tag-selector-item")) {
	this.setState({ scrollToCell: this.focusedTagIndex });
	await this.waitForSectionRender();
	this.shiftFocus(e);
	return;
}
```
I pushed that little tweak and everything still seems to come together with the `handleKeyDown` being way more readable than before

adomasven: Sure, but just get rid of

```js
this.shiftFocus(e);
return;
```

and inline the call to the function (i.e. don't have `this.shiftFocus()` as a separate function) at the end.

abaevbog: Oh, because the `shiftFocus` in now only called in `handleKeyDown`, so there's no reason to keep it as a separate function? I moved the rest of focus-shifting logic back into `handleKeyDown` so everything is in one place. Is that what you had in mind?

PR #3986, opened 2024-04-15
vpat 52: added aria label to quick search button

abaevbog:


chrome/content/zotero/elements/quickSearchTextbox.js
```diff
@@ -71,6 +71,7 @@
 			dropmarker.id = "zotero-tb-search-menu-button";
 			dropmarker.tabIndex = -1;
 			dropmarker.setAttribute("type", "menu");
+			dropmarker.setAttribute("aria-label", Zotero.getString("quickSearch.modes"));
```
dstillman, line 74: 1. All new strings should be done via Fluent, with [`.aria-label = `](https://searchfox.org/mozilla-central/rev/fe2743c6c5c708061c7f6504b26958fcc815bb4a/browser/locales/en-US/browser/browser.ftl#59-60).
2. `modes` → `mode`

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -816,6 +816,7 @@ fileInterface.OPMLFeedFilter	= OPML Feed List
 
 quickCopy.copyAs = Copy as %S
 
+quickSearch.modes           		= Quick search modes dropdown
```
dstillman, line 819: → `Quick-search mode menu` (in zotero.ftl)

chrome/content/zotero/elements/quickSearchTextbox.js
abaevbog: Got it, right. The reason why did `Zotero.getString` at first is that the dropmarker and the input of `search-textbox` are in the shadow DOMs, so the fluent properties weren't getting set. I set `data-l10n-id` on the hosts of the shadow DOM and move the labels after with JS, which seems to work out just fine.

One thing to note: as I was experimenting with this, I tried to pull the dropmarker out of the shadow DOM and ... nothing changed? There is a comment that it had to be like that due to stylesheet conflicts but I don't see it now. So the shadom DOM might not be needed for the dropmarker anymore.

PR #3987, opened 2024-04-15
vpat 51: add aria labels to advanced search

abaevbog:
In addition, make `toolbarbutton`s explicitly focusable and make them clickable on space/Enter.


chrome/locale/en-US/zotero/zotero.ftl
```diff
@@ -509,3 +509,16 @@ item-pane-header-bibEntry =
...
+search-remove-btn =
+    .tooltiptext = { general-remove }
+search-add-btn =
+    .tooltiptext = { general-add }
+search-conditions-menu-aria-label =
+    .aria-label = Search conditions menu
+    .label = { $label }
+search-operators-menu-aria-label =
+    .aria-label = Search operator menu
+    .label = { $label }
```
dstillman, line 522: Why do we have both `aria-label` and `label`? `label` isn't actually ever set, is it?

And as mentioned elsewhere, doesn't it announce the element type already? I would think `aria-label` could be just "Search condition", "Operator", and "Value".

dstillman (changes requested):

chrome/locale/en-US/zotero/zotero.ftl
abaevbog: > Why do we have both aria-label and label? label isn't actually ever set, is it?

We have both `aria-label` and `label` because if we add just `aria-label`, it ends up clearing the `label` that is set when a menupopup selection changes (e.g. [here](https://github.com/zotero/zotero/blob/3176516b797ce5149991fe28c3d464583a11130b/chrome/content/zotero/elements/zoteroSearch.js#L346)). We don't set `label` attribute ourselves but I believe it's updated by the component itself when selected item changes. Most importantly, if `label` is cleared, the buttons appear blank (e.g. the "Title" bit of the first dropdown would be just an empty string).

So this is the cleanest workaround I found to avoid fluent clearing the labels set via JS elsewhere. It's the same approach to #4074.

> And as mentioned elsewhere, doesn't it announce the element type already? I would think aria-label could be just "Search condition", "Operator", and "Value".

It does (e.g. voiceover says "menupopup"), so that's indeed a bit too verbose. I just changed the strings for "Search condition", "Operator", and "Value".

dstillman: 1. Let's change the prefix from `search-` to `advanced-search-`.
2. I don't think we need `-aria-label` in the ids. While that's all we're actually concerned with, the id is just applied to the element itself — it's the attributes that determine what specifically is being set.

abaevbog: > Let's change the prefix from search- to advanced-search-.

Even though this would apply to components other than advanced search? `zoterosearch` custom component that we're working with here is also used in the popup to edit a saved search. This is why I thought `search-` is a good general prefix. Maybe would `zoterosearch-` be better?

> I don't think we need -aria-label in the ids.

Good point! I removed them

dstillman: > Even though this would apply to components other than advanced search? zoterosearch custom component that we're working with here is also used in the popup to edit a saved search.

A saved search is really just an advanced search that's been saved. This is the advanced-search interface.

abaevbog: Gotcha - then `advanced-search-` make perfect sense as the prefix

PR #3996, opened 2024-04-16
vpat 19: aria labels to tags and collection filter

abaevbog:
The filters only have a placeholder, so once there is some text, it goes away and nothing else gets announced, so we need a label. However, the note from the VPAT review indicates that the label needs to be VISIBLE per 3.3.2 success criteria: https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.

This is something worth discussing, since 3.3.2 seems mainly concerned with inputs for form fields, while these are search fields which is somewhat different. So I am not sure if the visibility requirement necessarily applies here. After all, https://www.w3.org/WAI/ that lists the accessibility criteria has a search field with the label not being visible ...

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -1170,6 +1170,7 @@ var ZoteroPane = new function()
 	this.initCollectionTreeSearch = function () {
 		let collectionSearchField = document.getElementById("zotero-collections-search");
 		let collectionSearchButton = document.getElementById("zotero-tb-collections-search");
+		collectionSearchField.inputField.setAttribute("aria-label", collectionSearchField.placeholder);
 		collectionSearchField.style.visibility = 'hidden';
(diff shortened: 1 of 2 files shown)


dstillman:
Yeah, we're not going to add visible labels to search fields with placeholders. Placeholders are 100% industry standard for search fields (as your w3.org example shows), and it's up to screen readers to handle them properly. And VoiceOver seems to — even with text in the field, it announces the field name based on the placeholder when I return to it. A comment on https://stackoverflow.com/a/48652774 suggests that JAWS has read placeholder text since 2019, and that at least at the time it would read both if present.

So unless other screen readers for some reason don't handle this properly, I think this can be closed. (Otherwise we'd still want to pull the localized strings out to a variable and include those in both fields, but that would be dumb. Generally speaking, ARIA properties shouldn't be necessary when things are already labeled another way — Firefox should take care of transferring those to the necessary accessibility properties.)

(And as for non–screen reader use, I agree with you that the guidance makes more sense for input fields.)

PR #4000, opened 2024-04-17
vpat 56: added tooltip for quickFormat Z-menu

abaevbog:
It could be a simple aria-label but tooltip is more functional


abaevbog:
That's fair, I didn't think about it! I renamed for "Show citation settings" and changed the `data-l10n-id` to be `quickformat-citation-settings` as well.

Issue #4002, opened 2024-04-17
VPAT 2 and 26: Quickformat unclear instructions


abaevbog:
There are multiple entries in the VPAT review that boil down to the functionality of the quick-format dialog not being effectively conveyed to the users.

VPAT 2:

Description: The search feature in the Word Plug-in provides instructions to use up/down arrows to navigate results. However, it does not clearly indicate that keywords can be typed to search the list of citations. If no search term is entered, pressing Tab shows the most recently added citation. The search functionality can be improved by clarifying what can be searched and explaining how to access the full citation list without needing to know specific keywords.

Impact Statement: Users, especially those relying on screen readers, may not understand the full functionality and options available in the search feature. This lack of clarity can hinder their ability to effectively find and use citations.

Expected result: The purpose of the search and instructions on accessing the full citation list without remembering specific keywords should be clearly communicated to screen reader users.

Actual result: The search lacks clear instructions and details on its full functionality.

----
VPAT 26:

Description: The Word Plug-in's search dialog for selecting citations lacks clear instructions for users relying on screen readers. After choosing a citation from the search results by pressing Enter, focus loops back to the search edit field, and the initial instructions to 'type your search term, use down/up arrows to navigate results' are repeated. If the user then presses the Left Arrow key, they hear a word from the selected citation (e.g., an author's surname) followed by 'press down arrow to open citation properties.' Pressing Enter on the citation at this point seems to be the way to confirm the selection and prevent looping back to the search field. While this process allows for adding multiple citations, the expected workflow is not clearly communicated to users.

Impact Statement: Unclear or inconsistent instructions can significantly hinder the usability of the plug-in for users relying on assistive technologies, particularly those unable to view the screen. Without a clear understanding of how to interact with the dialog and its controls, users may struggle to effectively add citations or may inadvertently create incomplete or incorrect citations.

Expected Result: The citation selection process should provide clear, step-by-step instructions guiding users on how to search for, select, modify, and confirm citations, as well as how to add multiple citations if desired. If looping back to the search field is intentional to allow adding more citations, this should be explicitly stated.

Actual Result: The current instructions in the Word Plug-in's citation selection dialog are unclear and do not adequately guide screen reader users through the process of adding citations, potentially leading to confusion and errors.

---

I am not exactly sure how to approach this. 
[...]

PR #4004, opened 2024-04-17
vpat 26: quickformat more detailed instructions

abaevbog:
I think this makes it a bit clearer as to what exactly is going on in the QF and what options are available.

Fixes: #4002

### DIFF
--- a/chrome/locale/en-US/zotero/zotero.ftl
+++ b/chrome/locale/en-US/zotero/zotero.ftl
@@ -468,9 +468,15 @@ toggle-preview =
             *[unknown] Toggle
     } Attachment Preview
 
-quickformat-aria-bubble = Press Down Arrow to open citation properties.
-quickformat-aria-input = Type your search term. Use Down Arrow and Up Arrow to navigate the search results.
-quickformat-aria-item = Press { return-or-enter } to select this item. Press Tab to return to the search field.
+quickformat-general-instructions = Use Left/Right arrows to navigate the entries of this citation. {
+    $dialogMenu ->
+        [active] Press Shift-Tab to focus the dialog's menu.
+        *[other] { "" }
+} Press { return-or-enter } to save edits to this citation. Press Escape to discard the changes and close the dialog.
+
+quickformat-aria-bubble = Item included in this citation. Press Down Arrow to open citation properties. { quickformat-general-instructions }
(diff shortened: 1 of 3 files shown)


chrome/locale/en-US/zotero/zotero.ftl
```diff
@@ -468,9 +468,15 @@ toggle-preview =
...
-quickformat-aria-input = Type your search term. Use Down Arrow and Up Arrow to navigate the search results.
-quickformat-aria-item = Press { return-or-enter } to select this item. Press Tab to return to the search field.
+quickformat-general-instructions = Use Left/Right arrows to navigate the entries of this citation. {
+    $dialogMenu ->
+        [active] Press Shift-Tab to focus the dialog's menu.
+        *[other] { "" }
+} Press { return-or-enter } to save edits to this citation. Press Escape to discard the changes and close the dialog.
+
+quickformat-aria-bubble = Item included in this citation. Press Down Arrow to open citation properties. { quickformat-general-instructions }
+quickformat-aria-input = Citation item search input. Type to search for an item to include in this citation. Then Press Tab to focus and navigate the list of matching items to add to this citation. { quickformat-general-instructions }
```
adomasven, line 478: Maybe "Then Press Tab to focus and navigate the list of search results to add to this citation."?

adomasven:
> The citation selection process should provide clear, step-by-step instructions guiding users on how to search for, select, modify, and confirm citations, as well as how to add multiple citations if desired.

I didn't test this with a screen reader, but is "open citation properties" clear enough regarding "instructions on how to [...] modify [...] citations"? I guess that popover just has fields that are pretty self-explanatory?

chrome/locale/en-US/zotero/zotero.ftl
abaevbog: Sure, that sounds nice

abaevbog:
> I didn't test this with a screen reader, but is "open citation properties" clear enough regarding "instructions on how to [...] modify [...] citations"? I guess that popover just has fields that are pretty self-explanatory?

I changed it for "Item included in this citation. Press Down Arrow to open a dialog with citation properties to add details. " - that should indicate as to "why" you would want to open the popup. Also, I added the aria properties to make it clear that the properties popover is a dialog.

And then, yeah, as soon as it is opened, the "Page" (or whatever locator is selected) input is focused, and it gets announced as such, so it should be fairly clear (or at least clearer) what can be done.

PR #4006, opened 2024-04-17
vpat 46: remaining aria-labels in preferences

abaevbog:
- Linked remaining found inputs/menulists to their labels.
- Made the "Choose resolver" a proper visible label, so that it does not oddly disappear if you click on the dropdown with "Custom" showing after even if the selection did not change. For the purpose of VPAT, a visible label is always good for success criteria 3.3.2 https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions
- Added explicit names to +/- buttons
- In cases when there is an input placed in the middle of a contextual sentence (e.g. "Clear trash after <input> days"), the label will announce both parts of the sentence, replacing the <input> for "_".


chrome/locale/en-US/zotero/preferences.ftl
```diff
@@ -36,3 +36,10 @@ preferences-color-scheme-dark =
 preferences-advanced-language-and-region-title = Language and Region
 preferences-advanced-enable-bidi-ui =
     .label = Enable bidirectional text editing utilities
+
+preferences-plus-btn =
+    .aria-label = Add
+    .label = +
+preferences-minus-btn =
+    .aria-label = Remove
+    .label = -
```
dstillman, line 45: - Include zotero.ftl and use `{ general-add }`/`{ general-remove }`.
- We don't want `+` and `-` to appear as localizable strings. If you omit `.label` here, does it overwrite the hard-coded `label`? (I'm not sure why it would.)

abaevbog: > Include zotero.ftl and use { general-add }/{ general-remove }

Got it - repushed.

> We don't want + and - to appear as localizable strings. If you omit .label here, does it overwrite the hard-coded label? (I'm not sure why it would.)

It does do that... Despite my expectation, adding explicit `data-l10n-attrs="aria-label"` does not help either. It's something I ran into with #3987 as well where as a clunky workaround I had to save the label, add `data-l10n-id` and re-add the label after. We try something similar here but it's not unnecessarily complex - it'll look at different fluent use cases in firefox to see if there is a solution there

abaevbog: So far, the least hacky way I discovered for here and the advanced search window (these are the only 2 instances where we need to add localization to something that already have a label) is to manually fetch the localized strings:
```
document.l10n.formatValues(['general-add', 'general-remove'])
	.then((res) => {
		let [add, remove] = res;
		document.querySelector("#quickCopy-delete").setAttribute('aria-label', remove);
		document.querySelector("#quickCopy-add").setAttribute('aria-label', add);
	});
```
While not pretty, it at least does not feel as hacky as my initial idea. Does it make sense? I'm not sure how many of such cases we are going to have. If it's just here and #3987, this is probably not an issue. But if this is something that we would encounter more often, we can move an abstraction of this into one of the JS files that's loaded for every window (customElements.js?) and expose this functionality via our own `data-l10n-` attributes.

chrome/content/zotero/preferences/preferences_general.xhtml
```diff
@@ -160,21 +160,21 @@
 		<groupbox>
 			<label><html:h2>&zotero.preferences.prefpane.locate;</html:h2></label>
 			
-			<hbox>
+			<hbox align="center">
+				<label control="openurl-menu" value="&zotero.preferences.openurl.choose;"></label>
```
dstillman, line 164: The string here is "Choose a resolver…", so we can't just stick that in front of a drop-down. We can add a new Fluent string for "Resolver" and change "Resolver:" to "Base URL" (taking the opportunity to remove the unnecessary colons).

And "Version:" I think we can just remove and hard-code, since 1.0 came out before Zotero even existed.

dstillman (changes requested):

chrome/content/zotero/preferences/preferences_export.xhtml
```diff
@@ -96,14 +96,13 @@
...
-			<button label="+"
-				onclick="Zotero_Preferences.Export.showQuickCopySiteEditor()"/>
+			<button disabled="true" id="quickCopy-delete" data-l10n-id="preferences-quickCopy-minus" data-l10n-args='{"label":"-"}' onclick="if (!this.disabled) Zotero_Preferences.Export.deleteSelectedQuickCopySite()"/>
+			<button data-l10n-id="preferences-quickCopy-plus" data-l10n-args='{"label":"+"}' id="quickCopy-add" onclick="Zotero_Preferences.Export.showQuickCopySiteEditor()"></button>
 		</hbox>
 		
 		<hbox align="center">
 			<label value="&zotero.preferences.quickCopy.dragLimit;"/>
-			<html:input type="text" preference="extensions.zotero.export.quickCopy.dragLimit" size="3"/>
+			<html:input aria-label="&zotero.preferences.quickCopy.dragLimit; _ &zotero.preferences.items;" type="text" preference="extensions.zotero.export.quickCopy.dragLimit" size="3"/>
```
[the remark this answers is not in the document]
abaevbog: Yeah, that was my "creative" idea to try to make it clear where exactly the input is in relation to other labels... Interestingly enough, it did get announced a expected with "_" being announced as "blank" for me.

But nonetheless, `aria-labelledby` is a way more reasonable and conventional way to approach this, so I added ids to all labels and set those inputs as `aria-labelledby` by them which works well.

chrome/locale/en-US/zotero/preferences.ftl
```diff
@@ -36,3 +36,11 @@ preferences-color-scheme-dark =
...
 preferences-advanced-enable-bidi-ui =
     .label = Enable bidirectional text editing utilities
+preferences-quickCopy-plus =
+    .aria-label = { general-add }
+    .label = { $label }
+preferences-quickCopy-minus =
+    .aria-label = { general-remove }
+    .label = { $label }
+preferences-general-locate-resolver = Resolver
+preferences-general-locate-base-url = Base URL
```
dstillman, line 46: Let's actually remove the `general-` here — I know we use `advanced-` in a couple above, but better to have only the section (`locate-`) so that the id doesn't get out of sync if we decide to move this to another tab.

chrome/content/zotero/preferences/preferences_general.xhtml
```diff
@@ -160,48 +160,36 @@
...
-			<hbox>
+			<hbox align="center">
+				<label control="openurl-menu" data-l10n-id="preferences-general-locate-resolver"></label>
 				<menulist id="openurl-menu" native="true">
 					<menupopup
 							id="openurl-primary-popup"
 							onpopupshowing="Zotero_Preferences.General.handleOpenURLPopupShowing(event)"
 							oncommand="Zotero_Preferences.General.handleOpenURLSelected(event)">
-						<menuitem label="&zotero.preferences.openurl.choose;"/>
+						<menuitem label="&zotero.preferences.openurl.custom;"/>
```
dstillman, line 170: Right now this shows as "Custom…" until you click on the drop-down, at which point it becomes "Custom". We should use just "Custom" (either added to Fluent as a general string and retrieved in `updateOpenURLResolversMenu` or just assigned from the existing properties string on load) and get rid of this .dtd string (English only).

dstillman (changes requested):

chrome/locale/en-US/zotero/preferences.ftl
abaevbog: That makes sense - I renamed these

chrome/content/zotero/preferences/preferences_general.xhtml
abaevbog: Gotcha. I went with using existing `general.custom` string as there is a step of setting the placeholder when the general preferences pane is initialized, so it's just a small tweak to that step.

abaevbog:
I rebased on `main` to avoid re-fetching xulrunner each time this branch is checked out. Two last commits are new to address the last set of comments, all previous commits are unchanged.

Issue #3936, opened 2024-04-08
Voiceover does not place cursor into dialog windows (VPAT #13)


abaevbog:
When voiceover is active, opening any dialog window (e.g "New collection") does not place voiceover's cursor inside of the new window. It remains within the original window even though the focus has moved into the dialog. So far, no focus tweaks or adding `role=dialog`/`aria-modal="true"` helped. In the screenshot below, the focus is inside of the "New collection" window but the small black square in the top left corner is the voiceover's cursor. Interestingly, opening a window not as a `modal` doesn't cause such issue but it does looks different and breaks the current setup of how dialogs return selected values.

<img width="1415" alt="Screenshot 2024-04-08 at 6 09 55 PM" src="https://github.com/zotero/zotero/assets/36271954/883c4bfa-964b-4cf8-91ca-bb086b0fa6ff">

A possible (though unfortunate) solution would be to not open the dialogs as a separate window but display them within the main `ZoteroPane` window with an opaque background covering the rest of the screen. Something similar to what is currently done in `Zotero.showZoteroPaneProgressMeter` with `zotero-pane-overlay`. This seems to be what [Firefox does](https://searchfox.org/mozilla-central/source/browser/base/content/main-popupset.inc.xhtml#238) as well, for example in `about:preferences#general` > "Network settings".
That way, everything happens within the same window, and voiceover just follows the focused element. 

The main drawback, of course, is that it requires revising all the dialog windows.

abaevbog:
> The main drawback, of course, is that it requires revising all the dialog windows.

To avoid redoing the logic of displaying pretty much every single `<dialog>` window, maybe we can:
1. make dialog windows non-modal
2. pass a callback function into the dialog window that will run on accept
3. cover the screen with an opaque element to make it look at least somewhat similar to what it was before.

In this case, the dialog window can be be moved around and it might open in a slightly different location than now but besides that, it should function pretty much the same way. I have a proof of concept here of what it could look like: https://github.com/abaevbog/zotero/commit/e0c421542ed497824581e8fc0f6fa9d001f8fc2b




https://github.com/zotero/zotero/assets/36271954/1d70be25-89c9-4ac8-a936-1fd5a115c374


[...]

Issue #3927, opened 2024-04-05
MacOS "Services" entry appears in context menu after certain actions


tnajdek:
This is quite bizarre, seems to be fx115 related (I wasn't able to reproduce it in pre-fx115 beta). Once added, it appears in all context menus.

https://github.com/zotero/zotero/assets/214628/4c27df42-fe70-465b-bcf4-75eaa213be4e

dstillman:
https://forums.zotero.org/discussion/113573/z7-beta-title-field-right-click-options

PR #4010, opened 2024-04-18
Use custom XUL binary on macOS

dstillman:
Backports the following Mozilla fixes:

https://bugzilla.mozilla.org/show_bug.cgi?id=1858409
(AutoFill/Services menus; fixes https://github.com/zotero/zotero/issues/3927)

https://bugzilla.mozilla.org/show_bug.cgi?id=1884631
(sheets → dialogs; fixes https://github.com/zotero/zotero/issues/3936)

https://bugzilla.mozilla.org/show_bug.cgi?id=148624
(ghost tooltips, because why not)

### DIFF
--- a/app/build.sh
+++ b/app/build.sh
@@ -498,8 +498,8 @@ if [ $BUILD_MAC == 1 ]; then
 	xz -d --stdout "$CALLDIR/mac/zotero.xz" > "$CONTENTSDIR/MacOS/zotero"
 	chmod 755 "$CONTENTSDIR/MacOS/zotero"
 
-	# TEMP: Modified versions of some Firefox components for Big Sur, placed in xulrunner/MacOS
-	#cp "$MAC_RUNTIME_PATH/../MacOS/"{libc++.1.dylib,libnss3.dylib,XUL} "$CONTENTSDIR/MacOS/"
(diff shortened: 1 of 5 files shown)


abaevbog:
This is awesome! I want to point out a somewhat odd behavior that I ran into while trying it out. If you open a dialog, make some other window active, click back onto the main zotero pane window and then close the dialog, it visually looks like the main window does not become active. The menu list at the top becomes just "File" and "Window", and all of the selected/focused elements look dimmed.

https://github.com/zotero/zotero/assets/36271954/46bce217-2d0b-4a73-a664-3b951a9374bc

In the example above, this will get fixed if I click on the vscode window and then back on the zotero pane window once again. Not sure how big of a deal that is but thought it's worth being aware of.

dstillman:
I'm actually not seeing that:

https://github.com/zotero/zotero/assets/783082/331a1b2d-1421-40ab-9d9a-d396407aea05

abaevbog:
> I'm actually not seeing that:

I realized that I had #4001 checked out to look at the final dialog's look without the titlebar but I assumed this behavior came from the custom binary so I noted it here. It does not happen for you on `master`? If not, we can blame it on my local setup

PR #4019, opened 2024-04-18
fx102: Hidden window: Bring back "Zotero" menuitem in Window menu

AbeJellinek:
Removed by mistake in 4e912b84c003d6b38e106412358b05e137e63f18.


dstillman:
Actually better to stick to the new Fluent string. I'm pretty sure we're going to have to remove all DTD strings, possibly at the next Firefox upgrade.

PR #4020, opened 2024-04-18
vpat 27: rtfscan - link labels to inputs

abaevbog:
- linked input/output file labels to the input fields so they have a descriptive name
- "Choose file" for both buttons is not context specific. Added separate .ftl entries to have "Choose input file" and "Choose output file" labels.

Note: there are readonly fields that voiceover does not announce as "readonly" - it seems to be a general firefox issue, so this is not addressed here. This is discussed in more details here: #3997

### DIFF
--- a/chrome/locale/en-US/zotero/zotero.ftl
+++ b/chrome/locale/en-US/zotero/zotero.ftl
@@ -8,6 +8,7 @@ return-or-enter =
 general-remove = Remove
 general-add = Add
 general-remind-me-later = Remind Me Later
+general-choose-file =  Choose File…
 
 menu-file-show-in-finder =
     .label = Show in Finder
@@ -221,10 +222,10 @@ rtfScan-output-file = Output File
 
 rtfScan-no-file-selected = No file selected
 rtfScan-choose-input-file =
-    .label = Choose File…
+    .label = { general-choose-file }
     .aria-label = Choose Input File
 rtfScan-choose-output-file =
-    .label = Choose File…
+    .label = { general-choose-file }
     .aria-label = Choose Output File
 
 rtfScan-intro-page = 
--- a/chrome/locale/en-US/zotero/zotero.ftl
+++ b/chrome/locale/en-US/zotero/zotero.ftl
@@ -221,9 +221,11 @@ rtfScan-output-file = Output File
 
 rtfScan-no-file-selected = No file selected
 rtfScan-choose-input-file =
-    .label = Choose Input File...
+    .label = Choose File…
+    .aria-label = Choose Input File
 rtfScan-choose-output-file =
-    .label = Choose Output File...
+    .label = Choose File…
+    .aria-label = Choose Output File
 
 rtfScan-intro-page = 
     .label = Introduction
--- a/chrome/locale/en-US/zotero/zotero.ftl
+++ b/chrome/locale/en-US/zotero/zotero.ftl
@@ -8,7 +8,7 @@ return-or-enter =
 general-remove = Remove
 general-add = Add
 general-remind-me-later = Remind Me Later
-general-choose-file =  Choose File…
+general-choose-file = Choose File…
 
 menu-file-show-in-finder =
     .label = Show in Finder
--- a/chrome/locale/en-US/zotero/zotero.ftl
+++ b/chrome/locale/en-US/zotero/zotero.ftl
@@ -222,8 +222,10 @@ rtfScan-output-file = Output File
 zotero-file-none-selected =
     .value = No file selected
 
-zotero-file-choose =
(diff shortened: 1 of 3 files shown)


chrome/locale/en-US/zotero/zotero.ftl
```diff
@@ -210,8 +210,10 @@ rtfScan-output-file = Output File
 zotero-file-none-selected =
     .value = No file selected
 
-zotero-file-choose =
-    .label = Choose File…
+rtfScan-file-choose-input =
+    .label = Choose { rtfScan-input-file }
+rtfScan-file-choose-output =
+    .label = Choose { rtfScan-output-file }
```
dstillman, line 216: I don't think we want to use variables for something like this. This is just a normal phrase, and the words might be different in the different contexts. We'd only use a variable if we were referring to, say, an actual button label that appears elsewhere.

chrome/locale/en-US/zotero/zotero.ftl
```diff
@@ -213,8 +213,10 @@ rtfScan-output-file = Output File
 zotero-file-none-selected =
     .value = No file selected
 
-zotero-file-choose =
-    .label = Choose File…
+rtfScan-file-choose-input =
+    .label = Choose input file
+rtfScan-file-choose-output =
+    .label = Choose output file
```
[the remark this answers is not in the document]
abaevbog: Done.

> (That's not showing up in the input for me, though…)

The "No file selected" string was indeed getting overridden by an empty string in `updatePath`. It should appear now.

chrome/locale/en-US/zotero/zotero.ftl
```diff
@@ -219,11 +219,11 @@ rtfScan-introPage-description2 = To get started, select an RTF input file and an
...
-zotero-file-none-selected =
-    .value = No file selected
-
-zotero-file-choose =
-    .label = Choose File…
+rtfScan-no-file-selected = No file selected
+rtfScan-choose-input-file =
+    .label = Choose Input File...
+rtfScan-choose-output-file =
+    .label = Choose Output File...
```
dstillman, line 226: 1. Real ellipses: `...` → `…`
2. Can we use `aria-label` for these and keep `label` as "Choose File…"? Visually, this is a bit overly verbose, and the buttons aren't aligned properly with labels.

chrome/locale/en-US/zotero/zotero.ftl
```diff
@@ -221,9 +221,11 @@ rtfScan-output-file = Output File
 
 rtfScan-no-file-selected = No file selected
 rtfScan-choose-input-file =
-    .label = Choose Input File...
+    .label = Choose File…
+    .aria-label = Choose Input File
 rtfScan-choose-output-file =
-    .label = Choose Output File...
+    .label = Choose File…
```
dstillman, line 227: This should use a `general-choose-file` variable so it doesn't have to be localized twice.

chrome/locale/en-US/zotero/zotero.ftl
abaevbog: > Real ellipses: ... → …

Oh, I did not actually know there is a separate character for ellipses... Good to know!

> Can we use aria-label for these and keep label as "Choose File…"? Visually, this is redundant, and the buttons aren't aligned properly with different labels.

I think so... My understanding is that in general, it's best to be consistent across the board. But the VPAT issue did specifically refer to the "accessible name" not being descriptive, which I assume is what is being announced by screen reader. So we can just go with the aria-label for now and tweak it later if they get back with more comments

dstillman: Well, the other option would be to link each button to each label ("Input File", "Output File"), I guess with `aria-describedby` (?), and then avoid an extra localized string. But I don't know how that would be read.

chrome/locale/en-US/zotero/zotero.ftl
abaevbog: Right, I moved this into `general-choose-file = Choose File…`

chrome/locale/en-US/zotero/zotero.ftl
abaevbog: `aria-describedby` would sound a bit off, I think. It's generally meant for context that is announced after the labels and main properties of the item, sometimes after a pause. So, for instance, with voiceover, it would say "Choose file, button, has focus", pause, "Input file", which sounds off.

We can consider `aria-labelledby` that would point at the button itself and the label. In that case, I am hearing "Choose file... Input file" as one announcement. It gets the message across though it does sound a bit clunky with two "file" strings. But to avoid multiple localized strings, perhaps it's worth it?

PR #4021, opened 2024-04-18
vpat 37: run JS aria edits + localization

abaevbog:
- added aria label to the code editor field
- to add the aria-label and keep everything consistent, load zotero.ftl and moved remaining hardcoded strings to .ftl file

### DIFF
--- a/chrome/content/zotero/runJS.html
+++ b/chrome/content/zotero/runJS.html
@@ -13,7 +13,7 @@
 		<main>
 			<div class="textbox-container">
 				<div class="textbox-header">
-					<label data-l10n-id="runJS-code" class="textbox-label" for="editor-code"></label>
+					<label id="code-editor-label" data-l10n-id="runJS-editor-label" class="textbox-label"></label>
 					<div>
 						<label id="run-as-async-label" data-l10n-id="runJS-runAsync" for="run-as-async"></label>
 						<input id="run-as-async" type="checkbox" onclick="update()">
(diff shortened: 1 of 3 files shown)


chrome/locale/en-US/zotero/zotero.ftl
```diff
@@ -239,6 +239,19 @@ rtfScan-complete-page =
 
 rtfScan-complete-page-description = Your document has now been scanned and processed. Please ensure that it is formatted correctly.
 
+runJS-title = Run Javascript
+runJs-code = Javascript code editor
```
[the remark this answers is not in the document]
abaevbog: Oh yeah, different capitalization was my a typo but the ftl indeed ended up being unnecessarily complex. I did some refactoring so no need for duplicate values or variables now

chrome/content/zotero/runJS.html
```diff
@@ -1,36 +1,37 @@
...
-			<button id="help-button" onclick="openHelp()">Help</button>
+			<button data-l10n-id="runJS-help" id="help-button" onclick="openHelp()"></button>
 		</nav>
 		
 		<main>
 			<div class="textbox-container">
 				<div class="textbox-header">
-					<label class="textbox-label" for="editor-code">Code:</label>
-					<label id="run-as-async-label" for="run-as-async">
+					<label id="code-editor-label" data-l10n-id="runJS-editor-label" class="textbox-label"></label>
```
dstillman, line 16: Revert. "Code:" should be localized, but it is fine as a label. This is a developer tool called "Run JavaScript". It's obvious what "Code" means.

abaevbog: Makes sense. In all VPAT issues, the degree to which the labels are descriptive is brought up very often, which is where my "Javascript code editor" came from. But as a rule of thumb, I guess it's better to not change existing labels at this point and we can always revise them later if the VPAT review indicates that it's necessary. I pushed these little tweaks

dstillman:
Merged

PR #4022, opened 2024-04-18
vpat 40/41: style editor aria fixes

abaevbog:
- linked all labels to their respective inputs so they are announced by screen readers
- added aria-labels for the actual editor and preview iframes
- locator type and locator input are assigned aria-labels
- added visible labels to citation style and language menus, since vpat 40 specifically asks for visible labels

Vpat 40 says "These elements should have descriptive, persistent, visible labels and descriptive names that include their labels." referring to all the inputs of the style editor. Per #4039, the locator type and locator input are left as is but there are aria-labels set on them so that they are announced by the screen reader.

I did add visible labels for the "Language" and the "Citation Style" because I feel like it doesn't hurt here?

### DIFF
--- a/chrome/content/zotero/tools/csledit.xhtml
+++ b/chrome/content/zotero/tools/csledit.xhtml
@@ -41,19 +41,24 @@
 	<script src="chrome://zotero/content/include.js"/>
 	<script src="csledit.js"/>
 	
+	<linkset>
+		<html:link rel="localization" href="zotero.ftl"/>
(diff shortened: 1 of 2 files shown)


dstillman:
Merged. I couldn't push to this old PR, but I moved "Citation Style:" and "Language:" to standalone strings and referenced those, since we already use them elsewhere and have additional usage in zotero.dtd that we need to migrate.

PR #4027, opened 2024-04-19
vpat 49: refactor note format preference section

abaevbog:
"Note format" section in "Export" tab is technically a table and is announced as such by screen readers, which is misleading. It also opens a menulist when a checkbox is clicked, which is not an expected behavior and can shift focus.

This refactors the "Note format" section to be a vbox container with `<checkbox>`es and labels, which is more correct semantically and does not confuse screen readers. In addition, when a chekbox is clicked, dispatch a "change" event instead of unnecessarily open up a menu.

### DIFF
--- a/chrome/content/zotero/preferences/preferences_export.jsx
+++ b/chrome/content/zotero/preferences/preferences_export.jsx
@@ -258,9 +258,9 @@ Zotero_Preferences.Export = {
 			}
 			item.value = JSON.stringify(format);
 		}
-		// After updating menulist item value we have to wait a bit before doing click(), to avoid anomalies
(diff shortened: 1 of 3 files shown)


dstillman:
Getting rid of the table is fine, but the checkboxes should remain where they were.

Before:

Screenshot 2024-06-06 at 1 22 37 AM

After:

Screenshot 2024-06-06 at 1 23 01 AM

(Not sure what that `click()` was doing there. I don't think that opened a menu when this was created.)

abaevbog:
Oh, yes! I just shifted it to the left as before:

Screenshot 2024-06-06 at 12 07 12 PM

Issue #3995, opened 2024-04-16
Update Attachment and Annotation preview sections whenever annotations are added/modified/deleted for the current attachment


mrtcode:
https://forums.zotero.org/discussion/113632/zotero-7-beta-pdf-attachment-preview-and-annotations-not-refreshed-after-adding-annotations

mrtcode:
@windingwind Now we can also render missing cached images using PDF worker to display them in the annotation preview section:

```js
await Zotero.PDFWorker.renderAttachmentAnnotations(attachmentItemID);
```

(You can test that by deleting `zotero_data_dir/cache` directory)

windingwind:
Thanks! I've updated to #4031. Works well!

Issue #3993, opened 2024-04-16
Attachment preview appears in the Item Pane of the Reader tab


mrtcode:
https://forums.zotero.org/discussion/113658/zotero-7-beta-preview-appearing-in-the-item-pane-of-the-pdf-tab

Issue #4029, opened 2024-04-21
Unnecessary rendering of context-pane sections during file sync


dstillman:
I was looking into https://forums.zotero.org/discussion/113847/bug-report-zotero-7-beta-slows-down-when-syncing, and one thing I noticed is that when I do the first file sync of a session, I see this line in the console:

> PDF de433390649f464daccee838d3e0bbae [1.5 Adobe PDF Library 6.0 / Adobe InDesign CS (3.0.1)] (PDF.js: 3.8.0 [b9073e7])

I'm seeing this in the library pane with no items selected and three tabs that I haven't yet clicked on, so there's no reason for the reader to be loaded at all. I also saw this:

> zotero(2)(+0000005): Failed to load preview for attachment 1/DLQ3KWR6: Error: Reader is already initialized

(DLQ3KWR6 is one of the attachments I have open.)

I tracked this down to this line getting called from a `redraw` notifier trigger during file sync operations:

https://github.com/zotero/zotero/blob/704d90cd83ff1831d965185636f5f75eb7c5f24a/chrome/content/zotero/elements/contextPane.js#L82

And then this line runs:

> https://github.com/zotero/zotero/blob/704d90cd83ff1831d965185636f5f75eb7c5f24a/chrome/content/zotero/elements/contextPane.js#L106

So, a few things:

1. A `redraw` probably doesn't need to affect the item pane at all — it's primarily meant to force redrawing of for the collection and item tree rows for reasons other than item modifications (which cause necessary refreshing via other events). So we can limit this to specific item events (just `modify`?).
2. Ideally nothing would be getting updated when it's not visible. If I'm in the library tab and no item is selected, we don't really want to do work in the context pane just because an item was modified. Can we mark the pane as stale and update it only when the tab is actually selected?
3. I'm not sure the `item.parentID != itemDetails.parentID` line is correct when a standalone attachment is open in a tab. I might have been seeing something weird there, related to `targetItem` being a different item for a standalone attachment.

PR #4031, opened 2024-04-22
Fix attachments & annotation box not refreshed bug and add tests

windingwind:
fix: #3993
fix: #3995

### DIFF
--- a/chrome/content/zotero/elements/attachmentPreview.js
+++ b/chrome/content/zotero/elements/attachmentPreview.js
@@ -249,7 +249,7 @@
 			if (this._isDiscarding) {
 				return;
 			}
-			if (!force && this.isVisible) {
+			if (!force && (this.isVisible || !this._reader)) {
 				return;
 			}
 			this._isDiscarding = true;
--- a/chrome/content/zotero/elements/attachmentPreview.js
+++ b/chrome/content/zotero/elements/attachmentPreview.js
@@ -217,6 +217,8 @@
 			if (!this.initialized && itemID === this._renderingItemID) {
 				return;
 			}
+			// For tests
+			this._renderDeferred = Zotero.Promise.defer();
 			this._renderingItemID = itemID;
 			let success = false;
 			if (this.isValidType && await this._item.fileExists()) {
@@ -236,6 +238,7 @@
 			if (this._renderingItemID === itemID) {
 				this._renderingItemID = null;
 			}
+			this._renderDeferred?.resolve();
 		}
 
 		async discard(force = false) {
@@ -270,6 +273,7 @@
 			this._id("preview")?.after(this.nextPreview);
 			this.setPreviewStatus("loading");
 			this._isDiscarding = false;
+			this._renderDeferred?.resolve();
 		}
 
 		async openAttachment(event) {
--- a/chrome/content/zotero/elements/attachmentPreview.js
+++ b/chrome/content/zotero/elements/attachmentPreview.js
@@ -165,6 +165,7 @@
 			this.addEventListener("focusin", this._handleFocusIn);
 			this.addEventListener("keypress", this._handleKeypress);
 			this.setAttribute("data-preview-type", "unknown");
+			this._notifierID = Zotero.Notifier.registerObserver(this, ["item"], "attachmentPreview");
 		}
 
 		destroy() {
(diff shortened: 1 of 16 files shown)


windingwind:
After this: #4029

chrome/content/zotero/xpcom/reader.js
```diff
@@ -1448,7 +1448,8 @@ class ReaderPreview extends ReaderInstance {
 	}
 
 	_isReadOnly() {
-		return true;
+		// updateAnnotations is only allowed in editable reader
+		return false;
```
dstillman, line 1452: Is there not a better way to do this? The preview is certainly read-only, and we don't want to accidentally include some other functionality later.

What are you saying is only allowed in the editable reader, and can we just change that to work properly here?

windingwind: It's https://github.com/zotero/reader/blob/224b64ef0c4c106d6f0bae1159759774d15a3d59/src/common/annotation-manager.js#L96

dstillman: And why are we running that on a preview?

dstillman: This happens if you make a change to an annotation while there's also an open preview for that file?

Seems like we need to either make that just return without throwing an error (and let some later error throw if the item is actually non-editable) or somehow skip that for previews.

windingwind: Called by this line: https://github.com/windingwind/zotero/blob/24dd040909da5e3724e85d2c463532091c0af1b0/chrome/content/zotero/elements/attachmentPreview.js#L203

Which is for updating the annotations in the preview when modified/created

windingwind: Well, I no longer see this error, so maybe this change is unnecessary. Reverted

windingwind:
The new tests have been improved to be more robust

chrome/content/zotero/xpcom/reader.js
windingwind: Well, when testing the `renderAttachmentAnnotations`, I see this error:
```
Uncaught (in promise) Error: Cannot update annotation for read-only file
    updateAnnotations resource://zotero/reader/reader.js:41502
    onUpdateAnnotations resource://zotero/reader/reader.js:43406
    _renderNext resource://zotero/reader/reader.js:25917
```

@mrtcode Is there any chance we could disable this `updateAnnotations` in the preview? I suppose this requires some reader-side changes.

mrtcode: @windingwind 

This specific error is an old reader issue and is fixed now.

Just one question. Why [this code](https://github.com/windingwind/zotero/blob/ce408a87d5ea41e5547fe4dd1ac93e1fd53fe26d/chrome/content/zotero/elements/attachmentPreview.js#L184-L213) is necessary?

windingwind: Hi @mrtcode , it's copied from xpcom/reader.js, to update the annotations in the reader, according to https://github.com/zotero/zotero/issues/3995. The `ReaderPreview` instances are not real readers and are not added to the `Zotero.Reader._readers`, and thus cannot be updated by the notifier callback of `Zotero.Reader`.

Alternatively, we could have something like `Zotero.Reader._previews` for updating preview readers. Not sure which one is better.

mrtcode: All right, we can think about that in future. For now it's ok. Thanks for explaining.

windingwind:
So, I think this PR is ready for review and merge.

It includes:
1. Fixes to #3993, #3995, and #4029
2. New tests for the attachments section and the attachment section

The PR should be able to reliably pass the new tests now.

chrome/content/zotero/elements/attachmentPreview.js
```diff
@@ -177,13 +178,47 @@
...
+			}
 		}
 
 		async render() {
 			let itemID = this._item?.id;
 			if (!this.initialized && itemID === this._renderingItemID) {
 				return;
 			}
+			// For tests
+			this._renderDeferred = Zotero.Promise.defer();
```
dstillman, line 221: If this is only for tests (and there's no way to avoid), we can put in `if (Zotero.test) {` to make sure that nothing is relying on it that shouldn't be.

dstillman: This is also technically a deprecated pattern from Bluebird that we don't really want to use. Can use native promises instead:

```javascript
let resolve;
this._renderPromise = new Promise((res) => {
    resolve = res;
}));
```

And below just:

```javascript
resolve();
```

And in tests:

```javascript
await itemDetails._renderPromise;
```

windingwind: done

chrome/content/zotero/elements/itemPane.js
```diff
@@ -371,7 +364,7 @@
 		}
 
 		setReadLabel(isRead) {
-			var elem = document.getElementById('zotero-feed-item-toggleRead-button');
+			var elem = this.getCurrentPane().querySelector('#zotero-feed-item-toggleRead-button');
```
dstillman, line 367: But we shouldn't use `id` if we're creating this multiple places in the document — we should use a class.

chrome/content/zotero/elements/itemPane.js
```diff
@@ -336,13 +336,13 @@
 			let { doc, append } = data;
 
 			let toggleReadButton = doc.createXULElement("button");
-			toggleReadButton.id = "zotero-feed-item-toggleRead-button";
+			toggleReadButton.classList.add("zotero-feed-item-toggleRead-button");
```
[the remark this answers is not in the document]
windingwind: Sorry I missed the trash and publication buttons. Updated

test/tests/itemPaneTest.js
```diff
@@ -344,7 +373,479 @@ describe("Item pane", function () {
...
+			await ZoteroPane.selectItem(attachment.id);
+			assert.isTrue(attachmentsBox.hidden);
+
+			// Standalone attachment: hide
+			let attachment1 = await importFileAttachment('test.pdf');
+			await ZoteroPane.selectItem(attachment1.id);
+			assert.isTrue(attachmentsBox.hidden);
+		});
+
+		it("should not show attachments pane preview in reader best-matched attachment item", async function () {
```
dstillman, line 415: What does this mean?

windingwind: it's `should show ... in reader (when the attachment item is the best-matched attachment item of its parent item)`

dstillman:
I squashed a bit and merged.

Issue #4008, opened 2024-04-18
Wrong tag is focused when re-entering tag selector via keyboard


dstillman:
1. Tab into the tag selector
2. Use right-arrow to move the focus to a tag on the second row
3. Use Page-Down to scroll down a bit
4. Press Tab to jump to Filter Tags
5. Shift-Tab back to tag selector

Instead of the previously focused tag, some other tag a bit above the currently visible tags will be focused.

Follow-up to https://github.com/zotero/zotero/pull/3984

PR #4035, opened 2024-04-22
properly refocus tag from tag selector on re-enter

abaevbog:
- do not erase the last recorded tag on blur of the tag selector
- on tab from collectionTree or shift-tab from tag selector input, try to refocus that last focused tag. If one does not exist, try to focus the first non-disabled tag.

Fixes: #4008


abaevbog:
Oh, yeah, thank you for pointing that out. The record of the focused tag needs to be reset at some point - doing it in `render()` of the tags list was a bad idea because it would sometimes clear the focus when it shouldn't.

I moved the focus clearing logic into `getSnapshotBeforeUpdate` of the tags container - that way the focus should be cleared only when the state of the tag selector changes and it seems to address this glitchiness

Issue #4024, opened 2024-04-19
Context pane note handling issues


mrtcode:
https://forums.zotero.org/discussion/113814/bug-add-note-can-only-be-added-to-the-first-note-opened-not-the-one-currently-open-in-the-right

PR #4036, opened 2024-04-23
Fix notes context handling issues

windingwind:
fix: #4024

### DIFF
--- a/chrome/content/zotero/elements/contextPane.js
+++ b/chrome/content/zotero/elements/contextPane.js
@@ -170,9 +170,6 @@
 				this._sidenav.hidden = true;
 			}
 			else if (Zotero_Tabs.selectedType == 'reader') {
-				let currentNoteContext = this._getCurrentNotesContext();
-				currentNoteContext?._cacheViewType();
-
 				let reader = Zotero.Reader.getByTabID(Zotero_Tabs.selectedID);
 				this._handleReaderReady(reader);
 			
@@ -218,15 +215,7 @@
 			}
 
 			let currentNoteContext = this._getCurrentNotesContext();
-			let tabNotesDeck = currentNoteContext.querySelector('.zotero-context-pane-tab-notes-deck');
-			let selectedIndex = Array.from(tabNotesDeck.children).findIndex(x => x.getAttribute('data-tab-id') == reader.tabID);
-			if (selectedIndex != -1) {
-				tabNotesDeck.setAttribute('selectedIndex', selectedIndex);
-				currentNoteContext.mode = "childNote";
-			}
-			else {
-				currentNoteContext._restoreViewType();
-			}
+			currentNoteContext.switchToTab(reader.tabID);
 		}
 
 		_getCurrentNotesContext() {
--- a/chrome/content/zotero/elements/notesContext.js
+++ b/chrome/content/zotero/elements/notesContext.js
@@ -44,12 +44,21 @@
 					</vbox>
 				</vbox>
 				<vbox class="zotero-context-note-container context-note-standalone">
-					<vbox class="zotero-context-pane-editor-parent-line"></vbox>
+					<vbox class="zotero-context-pane-editor-parent-line">
+						<html:div class="parent-title-container">
+							<toolbarbutton class="zotero-tb-note-return"></toolbarbutton>
+							<html:div class="parent-title"></html:div>
+						</html:div>
+					</vbox>
 					<html:div class="divider"></html:div>
-					<note-editor class="zotero-context-pane-pinned-note" flex="1"></note-editor>
 				</vbox>
 				<vbox class="zotero-context-note-container context-note-child">
-					<vbox class="zotero-context-pane-editor-parent-line"></vbox>
+					<vbox class="zotero-context-pane-editor-parent-line">
+						<html:div class="parent-title-container">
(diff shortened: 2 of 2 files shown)


chrome/content/zotero/elements/contextPane.js
```diff
@@ -218,15 +215,7 @@
...
-			let tabNotesDeck = currentNoteContext.querySelector('.zotero-context-pane-tab-notes-deck');
-			let selectedIndex = Array.from(tabNotesDeck.children).findIndex(x => x.getAttribute('data-tab-id') == reader.tabID);
-			if (selectedIndex != -1) {
-				tabNotesDeck.setAttribute('selectedIndex', selectedIndex);
-				currentNoteContext.mode = "childNote";
-			}
-			else {
-				currentNoteContext._restoreViewType();
-			}
+			currentNoteContext.switchToTab(reader.tabID);
```
windingwind, line 218: 1. Clean up the `_handleTabSelect` and move some dirty code to `NotesContext#switchToTab`

chrome/content/zotero/elements/notesContext.js
```diff
@@ -44,12 +44,21 @@
 					</vbox>
 				</vbox>
 				<vbox class="zotero-context-note-container context-note-standalone">
-					<vbox class="zotero-context-pane-editor-parent-line"></vbox>
+					<vbox class="zotero-context-pane-editor-parent-line">
```
windingwind, line 47: 2. Reuse note editor title bar

chrome/content/zotero/elements/notesContext.js
```diff
@@ -193,6 +201,11 @@
 			});
 		}
 
+		_initNoteEditor() {
```
windingwind, line 204: 3. Only initialize note return button callback once

chrome/content/zotero/elements/notesContext.js
```diff
@@ -247,7 +260,18 @@
 		_getCurrentEditor() {
 			let splitter = ZoteroContextPane.splitter;
 			if (splitter.getAttribute('state') == 'collapsed' || ZoteroContextPane.context.mode != "notes") return null;
-			return this.node.selectedPanel.querySelector('note-editor');
+			switch (this.mode) {
```
windingwind, line 263: 4. Fix `_getCurrentEditor` bug mentioned in #4036

chrome/content/zotero/elements/notesContext.js
```diff
@@ -270,66 +292,82 @@
...
-			container.append(returnBtn, title);
-			parentTitleContainer.replaceChildren(container);
-			ZoteroContextPane.updateAddToNote();
 		}
 
 		updateNotesListFromCache() {
 			this._updateNotesList(true);
 		}
 
+
```
windingwind, line 348: 5. Make the logic of deciding what to show in the notes context (standalone note, current item note, or notes list) clearer and some bug fixes

windingwind (commented):
The changes are marked in the comments.

windingwind:
> Item note should be individual for each tab. Instead, it currently functions like a standalone note where the last opened note is visible in all tabs. It's best to check Zotero 6 to see how that should work.

What do you mean by `be individual`?

I need to figure out what to check in Z6. The current rules are as follows, I would appreciate it if you could give some hints on which part is inconsistent with our previous Z6 behavior @mrtcode :

1. If find an opened item note editor corresponds to the current tab, show the item note;
2. Otherwise, if find an opened standalone note editor, show the standalone note;
3. Otherwise, show the list.
4. When clicking the return button, remove the current note editor.

Say we have 3 tabs of reader: A, B, and C
1. if I open an item note $N_A$ in A and an item note $N_B$ in B, switching between these tabs I always get the $N_A$ in tab A, $N_B$ in tab B, and notes list in tab C;
2. if I open an item note $N_A$ in A and a standalone note $N_S$ in B, switching between these tabs I always get the $N_A$ in tab A and $N_S$ in tab B & C;

mrtcode:
Opening an "Item Note" sets it only for the current tab.
Opening an "All Notes" note sets it for all tabs that have don't have "Item Note" set.
Pressing back from an "Item Note" shows notes list but if you don't open anything, the next time the tab is selected you will see an "All Notes" note or just the list if no "All Notes" note is set.
Pressing back from "All Notes" unsets "All Notes" note.

But again, we should use Zotero 6 as a reference here. This behavior was broken multiple times in Zotero 7 beta, therefore I think it would make sense to add a note at the top of context pane file explaining how it should work.

dstillman:
(I actually saw some breakage in the latest version of Z6 as well when testing this.)

windingwind:
> Opening an "Item Note" sets it only for the current tab.
Opening an "All Notes" note sets it for all tabs that have don't have "Item Note" set.
Pressing back from an "Item Note" shows notes list but if you don't open anything, the next time the tab is selected you will see an "All Notes" note or just the list if no "All Notes" note is set.
Pressing back from "All Notes" unsets "All Notes" note.

Yes, I think that's exactly what this PR is doing. Do you have steps to reproduce that breaks these rules?

mrtcode:
> Yes, I think that's exactly what this PR is doing. Do you have steps to reproduce that breaks these rules?

Sorry, my fault. This PR actually fixes the mentioned issues. We can merge.

It was silently failing to build because of that sha256sum issue…

PR #4037, opened 2024-04-23
vpat 57/58: classic citation editor aria fixes

abaevbog:
- vpat 57: remove tabindex values, set `tabindex=0` on all focusable fields. It allows mozilla to determine the best focus sequence for keyboard navigation and fixes wrong order where some inputs would go before the "Ok" and "Cancel" buttons and some - after
- vpat 58: added labels to all inputs to be announced when focused. Added separate labels for locator menulist and for the locator input so that the locator type has a proper label. It is also better to have visible explicit labels when possible

### DIFF
--- a/chrome/content/zotero/integration/addCitationDialog.xhtml
+++ b/chrome/content/zotero/integration/addCitationDialog.xhtml
@@ -96,14 +96,14 @@
 		<hbox align="stretch" style="margin-top: 8px">
 			<vbox flex="1">
 				<hbox align="center">
-					<label value="&zotero.citation.prefix.label;"/>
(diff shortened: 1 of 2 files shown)


adomasven:
Could you rebase PRs to the current main when you submit them? It's usually not an issue, but sometimes the main branch has a newer version of xulrunner, so checking out and testing unrebased PRs cause a redownload.

abaevbog:
Yes, absolutely! I went through the old VPAT PRs as well and rebased them too. There were a lot of xulrunner updates lately, so I'll make sure to keep it up to date

PR #4048, opened 2024-04-24
vpat 50: advanced settings - disable custom data directory radio button

abaevbog:
Selecting custom data directory radio button in advanced settings opens up a modal to select the directory. It is not a typical behavior for radio button, so it conflicts with 3.2.2 success criteria.

As a simple fix, I thought we can disable the custom radio button so that the custom data directory can be only selected via the button. That way, with a more descriptive button label, it's still clear how to select the custom directory. But if one wants to go back to the default directory, that radio button still works as before.

VPAT issue for reference:

---
Description: In the Zotero desktop app's 'Settings' modal, there are two radio buttons in the 'Data Directory Location' section of the 'Advanced' tab - 'Default [location]' and 'Custom.' When the 'Custom' radio button is selected, the directory selection modal is opened. This modal should not open unless the user activates the 'Choose...' button. This occurs with or without a screen reader active.

Impact Statement: Operable elements have certain predictable behaviors. Whenever something breaks those expectations, such as an unexpected change in focus, the user should be informed ahead of time. If they are not it can be disorienting for all users, but especially those with visual or cognitive impairments.


abaevbog:
> Maybe it would be better to replace this with a "Use custom data directory" checkbox? Or a "Move Data Directory..." button? Having an active control under a disabled and un-selected radio button seems like unusual behavior to me.

That's totally fair - I wasn't sure what's the best direction here since the radio buttons setup is indeed a bit confusing.

Do you mean to use a button as an alternative to the whole radio buttons setup? Like maybe to display the currently selected data path just as text, have a button "Move data directory" to change it (essentially the "Choose data directory" with this setup). And maybe we also add a second button "Revert to default" to go back to the default data directory. I feel like this would cover all existing use cases without any unexpected behaviors by radio buttons?

abaevbog:
Still in works but what do you think about this interface? One input field to display the current directory, a button to choose a custom directory that triggers the file explorer popup and a button to go back to the default directory (instead of the "Default" radio button option). With some minor conditional logic to disable "go back to default" button if the directory is default and to change the label of the input (default vs custom). This felt like the most understandable approach among the other few I tried

Screenshot 2024-04-24 at 11 33 14 PM

dstillman:
How about this, largely mirroring the LABD setting above?



And then when a custom location is set:



(The text field has its background image [set in JS](https://github.com/zotero/zotero/blob/fbfe4b2dfa84d8f464cc2241e7848c7010a40388/chrome/content/zotero/preferences/preferences_advanced.js#L819), and the styling comes from [here](https://github.com/zotero/zotero/blob/fbfe4b2dfa84d8f464cc2241e7848c7010a40388/scss/preferences/_advanced.scss#L42).)

The main downside vs. what we have now is that it doesn't remind you what the default location will be before you click "Revert to Default Location…", though we could add that in the confirmation dialog.

(I'll also add some explanatory text to this section, since people get these two settings confused.)

dstillman:
Oh, the other problem with this is that we store the path even when you're not using it, and before you could see what that was. But that's probably not necessary. We could still set that as the default path in the filepicker if it exists.

abaevbog:
That makes sense! This way everything looks more consistent as well. 

> The main downside vs. what we have now is that it doesn't remind you what the default location will be before you click "Revert to Default Location…", which is pretty important, since you have to move files there, but we could add that in the confirmation dialog.

We can also add a tooltip with the default location to the revert button. It would get announced by screen readers, and personally I would expect some kinda indicator related to what the default directory is when the button is hovered

abaevbog:
So I reworked this to drop the radio buttons and be more like the LABD above.

Default dir selected: 
1_default

Custom dir selected:
2_custom


As I understood, the "Choose" and "Use custom location" buttons from the [mockup](https://github.com/zotero/zotero/pull/4048#issuecomment-2076406323) would do the same thing of opening the directory explorer, so I though to make them one button? This way, it ended up being identical to LABD actually.

> The main downside vs. what we have now is that it doesn't remind you what the default location will be before you click "Revert to Default Location…", which is pretty important, since you have to move files there, but we could add that in the confirmation dialog.

I did not yet change the confirmation dialog here because once the dialog appears, it doesn't seem like you can "opt out" of the change and go back. So would showing the default directory be helpful in this case? To at least somehow address this, I added the tooltip to the button. Maybe just adding a plain text default path after the button would make sense? Or, if there are more textual instructions coming to this section, the default directory could just be incorporated into that?

I also have a question about what to do with `useDataDir` preference here. It was linked to the radio buttons and it was supposed to change depending on the radio button selection - but it remained `true` no matter how I changed the directory setting on `master`. With this setup, I did not link that reference to anything because I wasn't sure exactly what it's behavior is supposed to be. At first, it seemed like `useDataDir` should be `true` if custom dir is selected and `false` when default dir is selected. But, again, that's not what it was doing for me. So I'm wondering what `useDataDir` is in general and if it should change depending on the dir selection?

dstillman:
> As I understood, the "Choose" and "Use custom location" buttons from the https://github.com/zotero/zotero/pull/4048#issuecomment-2076406323 would do the same thing of opening the directory explorer, so I though to make them one button?

No, I made them separate on purpose. Most people shouldn't change this, and we don't want to make it seem like a normal thing to do. So the main function of this section is just to show and take you to the default location. If you really want to choose a custom location, there's a button for that, listed after everything else.

All the button text also needs to match the mockup (text, case, ellipses). (Since you can't actually cancel from the confirmation dialog for "Revert to Default Location…", that technically shouldn't have an ellipsis, but we should just fix the dialog to actually allow you to cancel. We can do that separately, though.)

> Maybe just adding a plain text default path after the button would make sense?

Yes, let's try adding plain-text after. No need for a tooltip.

[Revert to Default Location…] (Default: /Users/dan/Zotero)

> At first, it seemed like useDataDir should be true if custom dir is selected and false when default dir is selected. But, again, that's not what it was doing for me.

Yeah, despite having written this code, I noticed the same and was surprised! It seems like we currently just don't clear `useDataDir` or `dataDir` when changing back to the default — we just set `dataDir` to the default location, and when that's the same as the default, we show the Default radio as selected. It seems like could just clear both if `path == this.defaultDir` in `Zotero.DataDirectory.set()`?

abaevbog:
> No, I made them separate on purpose. Most people shouldn't change this, and we don't want to make it seem like a normal thing to do. So the main function of this section is just to show and take you to the default location. If you really want to choose a custom location, there's a button for that, listed after everything else.

Gotcha, that does makes sense! I suppose that's another example where some duplication is actually not a bad thing. I brought back the button so these are the current states.

Default directory:

Screenshot 2024-10-23 at 12 41 50 AM

Custom directory:


Screenshot 2024-10-23 at 12 43 36 AM

I added the default directory after the "Restore to Default Location" button as an input, similar to "Data directory:" above. The only reason for it (as opposed to just a text label) is that if the path is somewhat long, the text would either wrap (which does not look nice), or it would get cutoff. So to see the full path, you would have to resize the window. Using a readonly input seemed like a decent solution, since you can scroll the text instead.

> It seems like could just clear both if path == this.defaultDir in Zotero.DataDirectory.set()?

I don't think I still fully understand the role of `useDataDir` pref but I am under the impression that we somewhat expect it to be always `true`? Changing `Zotero.DataDirectory.set` to clear/reset `useDataDir` and (optionally) `dataDir` leads to a strange glitch where restoring from a custom to default directory appends " default" to the path. In the image below, the "default" part isn't supposed to be there... Which means that the path becomes wrong and items don't get found. I kept the commented out tweaks for `Zotero.DataDirectory.set` for reference but ... whatever `useDataDir` is for, it does seem to be used at some point! I'll dig into it next to see what causes this to happen.

Screenshot 2024-10-23 at 12 51 14 AM

PR #4050, opened 2024-04-25
vpat 15: more detailed labels for itemPane

abaevbog:
- added role=button to the itemPane header
- twisty button's tooltip will include the section's name, e.g. "Expand info section".

Post: #3957 (vpat 15 and vpat 6 are duplicates)
Vpat comment after the initial edits this should address:

The issue is partially resolved. The various sections of the 'Item Pane' now have names. They are missing button roles, however; they can be activated to expand/collapse the section. The dedicated expand/collapse buttons should include the section in their names. Right now, they are all 'Expand section'/'Collapse section.

Initial issue description:

There is a pane of items with what seem to be "related", "tags", Libraries and collection", "notes", and possibly more. These buttons may expand or collapse more sections, and all of these buttons, and mentioned items, are not named or described properly.

### DIFF
--- a/chrome/content/zotero/elements/collapsibleSection.js
+++ b/chrome/content/zotero/elements/collapsibleSection.js
@@ -157,6 +157,8 @@
 			let twisty = document.createXULElement('toolbarbutton');
 			twisty.className = 'twisty';
 			twisty.setAttribute("tabindex", "0");
+			twisty.setAttribute("tooltip", "dynamic-tooltip");
+			twisty.setAttribute("data-l10n-attrs", "dynamic-tooltiptext");
 			this._head.append(twisty);
 			
 			this._buildExtraButtons();
@@ -173,6 +175,12 @@
 			if (this.hasAttribute('data-l10n-id') && !this.hasAttribute('data-l10n-args')) {
 				this.setAttribute('data-l10n-args', JSON.stringify({ count: 0 }));
 			}
+			// Fetch the localized value of the current pane which is used to set aria-properties
+			document.l10n.formatValue(`pane-${this.dataset.pane}`)
+				.then((res) => {
+					this._paneName = res;
+				})
+				.catch((_) => {}); // sanity check as it should never happen
 		}
 		
 		_buildContextMenu() {
@@ -409,8 +417,9 @@
 			this._head.setAttribute("aria-label", this.label);
 			this._title.textContent = this.label;
 			this._summary.textContent = this.summary;
-			this._head.querySelector('.twisty').hidden = this._disableCollapsing;
-			this._head.querySelector('.twisty').setAttribute('data-l10n-id', `section-button-${this.open ? "collapse" : "expand"}`);
+			let twisty = this._head.querySelector('.twisty');
+			twisty.hidden = this._disableCollapsing;
+			document.l10n.setAttributes(twisty, `section-button-${this.open ? "collapse" : "expand"}`, { section: this._paneName || "" });
 		}
 	}
 	customElements.define("collapsible-section", CollapsibleSection);
--- a/chrome/content/zotero/elements/collapsibleSection.js
+++ b/chrome/content/zotero/elements/collapsibleSection.js
@@ -179,8 +179,7 @@
 			document.l10n.formatValue(`pane-${this.dataset.pane}`)
 				.then((res) => {
 					this._paneName = res;
-				})
-				.catch((_) => {}); // sanity check as it should never happen
+				});
 		}
 		
 		_buildContextMenu() {
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -58,7 +58,6 @@ var ZoteroPane = new function()
 	this.clearItemsPaneMessage = clearItemsPaneMessage;
 	this.viewSelectedAttachment = viewSelectedAttachment;
 	this.reportErrors = reportErrors;
-	this.setDynamicTooltip = setDynamicTooltip;
 	
 	this.document = document;
 
@@ -495,18 +494,6 @@ var ZoteroPane = new function()
 				node.classList.remove("hidden-focus");
 			}
 		});
-	}
-
-	// Set the label of the dynamic tooltip. Can be used when we cannot set .tooltiptext
-	// property, e.g. if we don't want the tooltip to be announced by screenreaders.
-	function setDynamicTooltip(e) {
-		let tooltip = e.target;
-		let triggerNode = tooltip.triggerNode;
-		if (!triggerNode || !triggerNode.getAttribute("dynamic-tooltiptext")) {
-			e.preventDefault();
-			return;
-		}
-		tooltip.setAttribute("label", triggerNode.getAttribute("dynamic-tooltiptext"));
 	}
 
 	/**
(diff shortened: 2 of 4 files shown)


chrome/locale/en-US/zotero/zotero.ftl
```diff
@@ -381,10 +381,12 @@ section-button-remove =
...
     .tooltiptext = { general-add }
-section-button-expand =
-    .tooltiptext = Expand section
-section-button-collapse =
-    .tooltiptext = Collapse section
+section-button-twisty =
+    .tooltiptext = { $status ->
+        [expand] Expand
+        [collapse] Collapse
+        *[other] Toggle
```
dstillman, line 388: I know we've had an issue with unspecified values in Fluent, but do we actually need to have `other`, or is the issue just not having a [default](https://mozilla-l10n.github.io/localizer-documentation/tools/fluent/basic_syntax.html#selectors-and-plurals) via `*`? Because if we're always passing a status, it seems like there shouldn't be a need for `other` — we could just set, say, "Collapse" as the default and hopefully it would never actually show up incorrectly (and if it did it would probably only be for a split second, which is not much less weird than seeing "Toggle" briefly).

abaevbog: We don't strictly need `other` - it was more of "catch all" case to have a somewhat reasonable placeholder in case if, say, this `l10n-id` is used in some other component in the future and `status` is not passed. It's very much not necessary now, so I just removed it and made `*[collapse]` the default one.

dstillman:
Just using the generated label doesn't work — in the case of the counted objects, it gets weird:

Screenshot 2024-06-06 at 1 12 01 AM

This would need to say "Collapse Attachments section".

Also, similar to https://github.com/zotero/zotero/pull/4020#discussion_r1628748682, I'd prefer to use just "Expand section" / "Collapse section" for the visual tooltip and use an `aria-label` if necessary (or just link the buttons to the section title some other way that gets announced by screen readers). I'm not concerned about consistency. When the function is clear visually, the extra text is just extra noise.

abaevbog:
Yeah, I see that the tooltip does look extremely busy and "{number} attachments section" just doesn't sound right.

I tried a few things and the constraint here is that the tooltiptext will almost always get announced by screen readers (either as a name or as a description) of the button - this is likely something that mozilla does by itself to improve accessibility. So adding an aria-label of "Expand Info section" will give us two almost identical announcements - "Expand Info section", pause, "Expand section" - which is not great. 

So I thought that maybe this is actually a good use case for the `aria-description` property - it will be announced after the tooltip which gived us some variation of "Expand section", maybe a few other properties, pause, "Section: Info". That at least seems somewhat less redundant.

And then currently, I am fetching the fluent string of the pane name (e.g. `pane-attachments` to have "Section: attachments")  via `document.l10n.formatValue` to add it into `aria-description`. Since these pane name are not exposed anywhere else besides the sidenav (which I don't think should be involved here as it's a different component), it seemed like the cleanest approach here. Alternatively, we could add a hidden textfield with `data-l10n-id=pane-info` and such, and then use `aria-describedby` to point at it - it just seemed a bit more verbose to me, correct me if that would be better.

abaevbog:
> I tried a few things and the constraint here is that the tooltiptext will almost always get announced by screen readers

I looked at this again and I think this approach needs to be revised. Combining the `tooltiptext` and `aria-description` just doesn't sound very clear and this approach may not necessarily work in the future if we ever have a similar situation with different tooltips and aria-labels.

It looks like firefox uses a specific workaround with [dynamic tooltips](dynamic-shortcut-tooltip) somewhat similar to our `fake-tooltip` used in `virtualized-table` when tooltips and labels announced by screen readers are different. Not sure if this is the main reason but, for instance, the tooltip for the downloads button is quite long but the string announced by screen readers is different and it's just "Downloads". I think it's worth to try to do something similar here so that anytime in the future we have a ready workaround for it.

abaevbog:
I revised this to use a dynamic tooltip. That way, we can have a visible tooltip that does not affect screen reader in any way, so I think it addresses the problem of the tooltip and the `aria-label` being announced together and not sounding right.

chrome/content/zotero/elements/collapsibleSection.js
```diff
@@ -173,6 +175,12 @@
 			if (this.hasAttribute('data-l10n-id') && !this.hasAttribute('data-l10n-args')) {
 				this.setAttribute('data-l10n-args', JSON.stringify({ count: 0 }));
 			}
+			// Fetch the localized value of the current pane which is used to set aria-properties
+			document.l10n.formatValue(`pane-${this.dataset.pane}`)
+				.then((res) => {
+					this._paneName = res;
+				})
+				.catch((_) => {}); // sanity check as it should never happen
```
dstillman, line 183: We wouldn't want to swallow an error here. If there's an error, it can just be thrown to the console as an unhandled rejection. (And a "sanity check" would be making sure something is behaving as expected at a very basic level before continuing, not hiding a problem.)

chrome/content/zotero/zoteroPane.js
```diff
@@ -59,6 +59,7 @@ var ZoteroPane = new function()
 	this.clearItemsPaneMessage = clearItemsPaneMessage;
 	this.viewSelectedAttachment = viewSelectedAttachment;
 	this.reportErrors = reportErrors;
+	this.setDynamicTooltip = setDynamicTooltip;
```
dstillman, line 62: We haven't used this convention in many, many years. If it's a public function, it should just be `this.functionName = function () {` below.

chrome/content/zotero/zoteroPane.js
```diff
@@ -497,6 +498,18 @@ var ZoteroPane = new function()
...
+	function setDynamicTooltip(e) {
+		let tooltip = e.target;
+		let triggerNode = tooltip.triggerNode;
+		if (!triggerNode || !triggerNode.getAttribute("dynamic-tooltiptext")) {
+			e.preventDefault();
+			return;
+		}
+		tooltip.setAttribute("label", triggerNode.getAttribute("dynamic-tooltiptext"));
+	}
+
```
dstillman, line 512: Move down to…somewhere less prominent.

chrome/content/zotero/zoteroPane.js
```diff
@@ -497,6 +498,18 @@ var ZoteroPane = new function()
 		});
 	}
 
+	// Set the label of the dynamic tooltip. Can be used when we cannot set .tooltiptext
+	// property, e.g. if we don't want the tooltip to be announced by screenreaders.
+	function setDynamicTooltip(e) {
```
dstillman, line 503: `e` → `event` (we use `e` for errors in `catch` statements)

dstillman (changes requested):

abaevbog:
Gotcha! I revised this to essentially have `this.setDynamicTooltip = function (event) {}` and to not swallow the exception. I moved `setDynamicTooltip` to the bottom where other `onpopupshowing` listeners seem to be collected.

I wondered if it would be worth to turn this dynamic tooltip setup into something like a custom element - that way we could use it outside of the main window where `ZoteroPane` would not exist. I didn't see another clear use case for it at this point, though, so it's probably best to just keep it in `ZoteroPane` for now and move it later if we have to.

PR #4051, opened 2024-04-25
vpat 11: do not skip over disabled tags during arrow navigation

abaevbog:
If one explicitly selected to "Display all tags in this library", do not jump over disabled tags during arrow navigation. Followup to #3967.

Comment after initial changes this addresses:

The issue is partially resolved. The tags now have the checkbox role, which is sufficient, and they have their state announced when focused and activated. However, when all tags are set to be viewed, the disabled tags are not able to be selected on Windows. This makes it functionally no different from having 'Display All Tags in This Library' off; reasonably, one could expect the user wants to be able to see all of the tags if that option is on. It is fine for the disabled tags to not be in the focus order (navigating by Tab key), but they should still be able to be selected by screen reader cursors (navigating by arrow keys) and have a disabled state

### DIFF
--- a/chrome/content/zotero/components/tagSelector/tagSelectorList.jsx
+++ b/chrome/content/zotero/components/tagSelector/tagSelectorList.jsx
@@ -316,8 +316,8 @@ class TagList extends React.PureComponent {
 			return node.previousElementSibling;
 		};
 		let nextOne = nextTag(document.activeElement);
-		// Skip disabled tags
-		while (nextOne && nextOne.classList.contains("disabled")) {
+		// Skip disabled tags, unless all tags are explicitly set to be displayed
+		while (!Zotero.Prefs.get('tagSelector.displayAllTags') && nextOne && nextOne.classList.contains("disabled")) {
 			nextOne = nextTag(nextOne);
 		}
 		if (nextOne) {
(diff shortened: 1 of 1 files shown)


dstillman:
I was trying to test this and found something more problematic: if I enable "Display All Tags in This Library" and switch to a collection that only has one tag that's out of the initial window, focus jumps from the collections pane to Filter Tags without hitting the tag selector at all.

abaevbog:
That was a great observation, thank you for pointing it out. I believe root of the issue could be that we were trying to focus the first _rendered_ enabled tag. So, if there was no rendered enabled tag due to windowing (or no enabled tags at all), it did behave strangely.

I revised that logic a bit, so now on tab from collectionTree or shift-tab from the input, we should:
 - refocus the previously focused tag.
 - if there is no previously focused tag, focus the first enabled tag (scroll to it if needed).
 - if there are no enabled tags at all, focus the first tag (also scroll to it if needed).

Does this fix the issue you ran into?

dstillman:
This should apply to all disabled tags, not just with `tagSelector.displayAllTags` enabled. Colored tags are visible in all collections and can be disabled (most obviously in an empty collection).

abaevbog:
Good point. I suppose my idea then was to only focus disabled tags if the user explicitly requested it. But this does run into odd edge cases when only disabled tags are displayed. Then, only the first disabled tag becomes focusable but others cannot be focused with arrow keys. So it's probably indeed best to just let all tags be focusable regardless if they're enabled or not.

PR #4053, opened 2024-04-25
vpat 3: make "remove" btn in itemPane focusable even when hidden

abaevbog:
In itemPane, do not hide the actual "-" toolbarbutton, hide the `<image>`  inside of it. Visually, it looks the same but the button can be focused via tab without having to expicitly make it visible. It means that shift-tab from the section header will not skip the `-` button.

This does not apply to the itemBox, since there is a bit more logic in it and itemBox does not cause vpat issues.

Also, make "remove" toolbarbutton from tagsbox focusable like all others for consistency.

This is a followup to #3953. For reference, this is an example of how the `-` next to the note row gets skipped on shift-tab from "Collection and libraries" section header that this addresses.

https://github.com/zotero/zotero/assets/36271954/17a1aeed-9ab9-4a57-8323-2a09016e2143

### DIFF
--- a/chrome/content/zotero/elements/tagsBox.js
+++ b/chrome/content/zotero/elements/tagsBox.js
@@ -189,9 +189,9 @@
 
 			if (this.editable) {
 				var remove = document.createXULElement("toolbarbutton");
-				remove.setAttribute('value', '-');
 				remove.setAttribute('class', 'zotero-clicky zotero-clicky-minus');
-				remove.setAttribute('tabindex', -1);
+				remove.setAttribute('tabindex', 0);
+				remove.setAttribute("data-l10n-id", 'section-button-remove');
 			}
 
 			var row = document.createElement("div");
--- a/scss/elements/_librariesCollectionsBox.scss
+++ b/scss/elements/_librariesCollectionsBox.scss
@@ -44,11 +44,13 @@ libraries-collections-box {
 
 			toolbarbutton {
 				margin-inline-start: auto;
-				visibility: hidden;
 				border-radius: 2px;
+				& > image {
+					visibility: hidden;
+				}
 			}
 
-			&:is(:hover, :focus-within) toolbarbutton {
+			&:is(:hover, :focus-within) toolbarbutton > image {
 				visibility: visible;
 			}
 		}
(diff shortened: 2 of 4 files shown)


abaevbog:
There is a question that is only somewhat relevant here because of the small tagsBox edit but is much more important for upcoming VPAT issues. 

Mac and windows handle default focus movement [a bit differently](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_general_between). By default, mac will not focus all focusable elements that windows will. For example, in this PR, just by adding the `-` button of tags rows into the tab sequence, we'll have windows focus it on tab from `editable-text` of the tag. But mac won't do it, and focus will go straight to the next tag's `editable-text`. Per that link, mac has a special setting in preferences (keyboard > keyboard navigation) that enables proper tab behavior that one would expect. With that setting on, everything works as expected on mac as well.

So the question is, for the following accessibility edits, does it make sense to assume that this setting is enabled on mac? Initially, I assumed not, however, trying to tab through anything on mac is impossible if this is not turned on. For example, the actual "System preferences" cannot be properly navigated, or the "Firefox" settings page will only focus a few components, skipping most buttons.

So I'm kinda under the impression that if one intends to use keyboard for navigation, they would have that setting on, and apple intends it to be utilized that way?

As an example, another VPAT issue this directly relates to is 21, the non-focusability of the "Merge duplicates" button on tab from the itemTree. It's not an issue on windows, and is not an issue on mac if the keyboard navigation setting is on. If not - the button will get skipped.

stakats:
We could ask QL what they think.

abaevbog:
So we got the confirmation that the "Keyboard navigation" settings in mac is expected to be on, which simplifies things quite a bit. Then this is ready for review, and with that setting on, the `-` button next to the tag's `editable-text` properly receives focus on mac.

As a side note, we have some explicit keyboard navigation handling that is actually not needed if that "keyboard navigation" is on. For example, [moving focus](https://github.com/zotero/zotero/blob/53f685ca76616f7ff7758f83e0a3b92d30689dcb/chrome/content/zotero/elements/collapsibleSection.js#L328) through the header of a collapsible section or having the `-` [button focused](https://github.com/zotero/zotero/blob/53f685ca76616f7ff7758f83e0a3b92d30689dcb/chrome/content/zotero/elements/itemDetails.js#L506) for related/notes/collections sections. The only reason why it's there to begin with is that I was not aware that such functionality would for some reason be hidden behind a preference. So would it make sense to remove those?

dstillman:
This needs to be rebased.

Issue #4066, opened 2024-04-29
Follow-up to #4040


dstillman:
See commit comments on https://github.com/zotero/zotero/commit/0077c3f07a2f0fa1fb0e2568f25aa8b598b9074a

PR #4067, opened 2024-04-29
post vpat-57: qf locator string tweak

abaevbog:
"Locator input" -> "Locator"

Fixes: #4066

Per comment from https://github.com/zotero/zotero/commit/0077c3f07a2f0fa1fb0e2568f25aa8b598b9074a:
> "input" strikes me as a fairly technical term.
Doesn't the screen reader already tell you that it's a text box? So "input" seems a bit redundant anyway. Seems like it could just be "Locator"?

Yeah, "input" is fairly technical... Screen readers do announce that it's an input field in this way or another (e.g. voiceover will say "edit text"). With `aria-labelledBy` pointing at the locator type dropdown and the input's label, screen readers announce both (e.g. "Locator type page. Locator input 6"). So the only reason why I added "input" initially is to make a clearer distinction from "Locator type" which indeed might not be helpful. I suppose we can go with "Locator" and if the VPAT review indicates there's some confusion, we can revisit the exact phrasing?

Alternatively, we could try "Locator value"? This might indicate that the locator has essentially 2 pieces - "type" from the dropdown and the value that you have to type in. It would then be announced as something "Locator type page. Locator value" when the empty input field for the locator is focused.

### DIFF
--- a/chrome/content/zotero/integration/addCitationDialog.xhtml
+++ b/chrome/content/zotero/integration/addCitationDialog.xhtml
@@ -115,8 +115,8 @@
 					<menulist onchange="Zotero_Citation_Dialog.confirmRegenerate(true)" id="label" tabindex="0" native="true" data-l10n-id="quickformat-locator-type">
 						<menupopup id="locator-type-popup"/>
 					</menulist>
-					<label id="locator_input_label" data-l10n-id="quickformat-locator-value" hidden="true"></label>
-					<html:input aria-labelledby="label locator_input_label" oninput="Zotero_Citation_Dialog.confirmRegenerate(false)"  onchange="Zotero_Citation_Dialog.confirmRegenerate(true)" id="locator" tabindex="0"/>
+					<label id="locator-input-label" data-l10n-id="quickformat-locator-value" hidden="true"></label>
+					<html:input aria-labelledby="label locator-input-label" oninput="Zotero_Citation_Dialog.confirmRegenerate(false)"  onchange="Zotero_Citation_Dialog.confirmRegenerate(true)" id="locator" tabindex="0"/>
 				</hbox>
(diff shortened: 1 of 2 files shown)


dstillman:
Yes, let's try Locator for now.

There were a couple other comment in #4066, though.

abaevbog:
Oh yeah, apologies. Somehow I missed the other two...

> This should use hyphens, not underscores.

Just pushed that change.

> Is label supposed to be here?

Yes, "label" here is the id of the locator type dropdown `<menulist onchange="Zotero_Citation_Dialog.confirmRegenerate(true)" id="label" tabindex="0" data-l10n-id="quickformat-locator-type">`. Whitespace separated id's of elements in `<html:input aria-labelledby="label locator_input_label">` [tell](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby#values) screen readers that both of these elements need to be announced on focus. So that way, we first hear "Locator type Page" followed by "Locator ...".

Granted, "label" not a very descriptive id for the locator type dropdown. Changing it for something more descriptive like "locator-type-dropdown" would mean going over `addCitationDialog.js` to make sure nothing is broken and other naming is consistent with this. Would it be worth it to do it here?

dstillman:
Ah, OK — the id was actually changed from `locatorType` to `label` [with the introduction of citeproc-js](https://github.com/zotero/zotero/commit/bd0a4a3b96d92c8ff47f3451aba34d70f5c51ccc) in 2010, because citeproc-js [refers to it as the "label"](https://citeproc-js.readthedocs.io/en/latest/csl-json/markup.html?highlight=label#cite-items) (meaning the label of the locator). That seems like a confusing leaky abstraction, given the other uses of `label` in XUL, but I guess we can leave it for now.

PR #4070, opened 2024-04-29
vpat 8: aria hide titlebar buttons

abaevbog:
Added aria-labels for titlebar buttons to minimize/maximize/close window.

### DIFF
--- a/chrome/content/scaffold/scaffold.xhtml
+++ b/chrome/content/scaffold/scaffold.xhtml
@@ -353,21 +353,25 @@
 				titlebar-btn="min"
 				oncommand="window.minimize();"
 				data-l10n-id="browser-window-minimize-button"
+				aria-hidden="true"
 				/>
 			<toolbarbutton class="titlebar-button titlebar-max"
 				titlebar-btn="max"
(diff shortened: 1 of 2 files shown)


dstillman:
I can probably just update our script to extract Mozilla langpack strings for these from browser.ftl:

https://searchfox.org/mozilla-central/source/browser/locales/en-US/browser/browser.ftl#436-443

But at least on macOS, we should keep these from being accessible altogether. They're not — at least via the arrow keys — in standard macOS app windows, nor are they in Mac Firefox. The actions are all bound to system shortcut keys. (The full-screen button is also announced ("Fill Screen") all apps if you hover over it with the mouse.)

abaevbog:
> But at least on macOS, we should keep these from being accessible altogether.

That makes sense. These are generally never accessible and it's only voiceover that sometimes places the cursor on them. It's a bit confusing and does not expose any additional functionality because, like you pointed out, system shortcuts can be used for these window commands. So what if we just mark them with `aria-hidden="true"` to prevent voiceover from placing cursor on them?

PR #4071, opened 2024-04-29
Remove a bunch of unused PNG icons

AbeJellinek:
Removed `icons.jsx` declarations that were never referenced, then ran a script to find all PNGs whose basenames have no references in Zotero code. Removed a few more manually.


windingwind:
I've mentioned this change in https://www.zotero.org/support/dev/zotero_7_for_developers#zotero_platform
Once we release a new version with this PR, let me know so I can update the descriptions in the doc. Also, we need to announce this change to the dev list.

Issue #4073, opened 2024-04-30
QuickSearch textbox placeholder is empty after opening main window


windingwind:
I've been seeing this for a while, not sure when it started. I can observe the placeholder appear for a short while and then disappear. Maybe related to https://searchfox.org/mozilla-esr115/source/toolkit/content/widgets/search-textbox.js#114 clearing the `placeholder` attribute?

<img width="470" alt="image" src="https://github.com/zotero/zotero/assets/33902321/cc314460-9366-4547-a880-37192eb3a978">

Setting a small timeout (e.g. 0 or 1) sometimes works but sometimes does not; setting a larger timeout (e.g. 100ms) works for me, but not an ideal solution.

abaevbog:
I believe this comes from the odd fluent behavior I observed in a few other places. Essentially, if you set a label or a placeholder directly, (like `node.setAttribute("placeholder","hello"`) and add a fluent id `node.setAttribute("data-l10n-id","world")`, it will actually end up removing that "hello" placeholder you just added. It doesn't seem to matter if you set "hello" before or after `data-l10n-id` is set, since it does take a bit of time for fluent to process nodes. Which is why a timeout helps, I think.

I am not exactly sure what's the "right" way to tackle this but this is the cleanest workaround I've come up with. @windingwind, does it fix it for you?

PR #4074, opened 2024-04-30
set quicksearch placeholder as fluent argument

abaevbog:
Fluent seems to clear out label/placeholders attributes set directly (not via .ftl file). So quicksearch placeholder added directly ends up being emptied out.
This is the cleanest workaround found so far: add the placeholder as the fluent argument

Fixes: #4073


windingwind:
Nice and clear! Works for me.

PR #4087, opened 2024-05-02
redone fix of focus sticking to tab after click

### LINKED ISSUE #4077: Reader iframe doesn't receive focus when switching to a tab or pressing toolbar buttons on Windows
https://forums.zotero.org/discussion/114088/zotero-7-beta-ctrl-f-in-the-pdf-after-switching-tab

abaevbog:
- revert debcb9944d19862565435fe0b8768c95c3dc633f since it breaks drag-drop reordering of tabs
- when reader is being refocused by contextPane, add a small delay so that the focus settles on the tab before focusing the reader
- also explicitly refocus the reader tab after drag, since then focus also lands on the tab and finds itself outside of the reader

This should take care of focus sticking to the tab after clicking between them or drag/drop (which I assumed is not desired as well).

There are some other instances where focus may not do exactly what we want, though. Since reader is a separate browser, if any element outside of it receives focus, clicking back into the reader won't necessarily transfer focus back into it. For example, if I click on an `editable-text` in the context pane and then click on the magnifying glass in the reader, the search bar does not get focus because the reader itself does not have focus.This does not seem as big of an issue but still worth mentioning.

This tweak is in a way a repeat of the logic used to refocus the zoteroPane, but reader is refocused in `contextPane.js`, as opposed to the `tabs.js`. So it might be nice to consolidate these things into one place at some point.

Fixes #4077

### DIFF
--- a/chrome/content/zotero/tabs.js
+++ b/chrome/content/zotero/tabs.js
@@ -161,6 +161,7 @@ var Zotero_Tabs = new function () {
 				onTabMove={this.move.bind(this)}
 				onTabClose={this.close.bind(this)}
 				onContextMenu={this._openMenu.bind(this)}
+				refocusReader={this.refocusReader.bind(this)}
 			/>,
 			document.getElementById('tab-bar-container'),
 			() => {
@@ -568,6 +569,18 @@ var Zotero_Tabs = new function () {
 	this.selectLast = function () {
 		this.select(this._tabs[this._tabs.length - 1].id);
 	};
+
+	/**
+	 * Return focus into the reader of the selected tab.
+	 * Required to move focus from the tab into the reader after drag.
+	 */
+	this.refocusReader = function () {
+		var reader = Zotero.Reader.getByTabID(this._selectedID);
+		if (!reader) return;
+		setTimeout(() => {
+			reader.focus();
+		}, 100);
+	}
 
 	/**
 	 * Moves focus to a tab in the specified direction.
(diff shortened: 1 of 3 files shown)


chrome/content/zotero/tabs.js
```diff
@@ -569,6 +570,18 @@ var Zotero_Tabs = new function () {
...
+	/**
+	 * Return focus into the reader of the selected tab.
+	 * Required to move focus from the tab into the reader after drag.
+	 */
+	this.refocusReader = function () {
+		var reader = Zotero.Reader.getByTabID(this._selectedID);
+		if (!reader) return;
+		setTimeout(() => {
+			reader.focus();
+		}, 100);
```
dstillman, line 582: We really want to avoid arbitrary delays. Can we watch for a focus event on the tab and focus the reader after that (if necessary with some maximum delay, if we think there are cases where the tab isn't focused)?

adomasven: > I assume that last time my VM was just being a bit glitchy, which is why I thought the 100 delay was required.

Generally it helps to troubleshoot this sort of thing with the environment somehow slowed down (by running something CPU intensive alongside), otherwise people with slower systems will start reporting something like that and we won't be able to reproduce.

Reading the code and thinking about it, it would make sense if running this in the next event loop may be enough (without the timeout), but it might help to make sure.

abaevbog: Good call! I'll keep that in mind for the future. 

I ran [CpuStress](https://learn.microsoft.com/en-us/sysinternals/downloads/cpustres) at ~75% CPU usage and re-focusing of the reader still worked fine without any added delay to `setTimeout`.

I also booted up my old windows 10 laptop, and it worked there as well. So it does seem like the last time was a fluke of some kind.

PR #4100, opened 2024-05-08
Z6: Fix DOI field context menu items acting on *all* DOIs

### LINKED ISSUE #4099: Z6: Right-click on DOI → View Online opens all URLs since app start
Z6 equivalent of #3221

https://github.com/zotero/zotero/issues/3221#issuecomment-2099503720

AbeJellinek:
Fixes #4099; amends f006995676fc2bc160cc6a05e3d46f8666b33c22 to use event listener properties for listeners that need to be overwritten. (I don't have a Z6 build environment set up to test, but I don't see why this wouldn't work.)


dstillman:
This actually just doesn't work at all for me. I'm not sure why, but we hopefully won't be releasing any more 6.0 builds anyway, so we can probably just not worry about it…

dstillman:
Fixed in Z7

Issue #4094, opened 2024-05-06
Long press doesn't open context menu in various parts of Zotero on Windows with a touchscreen.


mrtcode:
https://forums.zotero.org/discussion/108057/zotero-7-beta-touch

abaevbog:
I can reproduce this. There is a square appearing in the long press area and then - nothing.

If I do a long press elsewhere (e.g. file explorer), the context menu appears after the long press is released and the square goes away. I will try to take advantage of having a (albeit barely working) touchscreen windows to get some info on this.

abaevbog:
> I would guess this comes down to places where we're handling contextmenu ([1](https://github.com/zotero/zotero/blob/89320a7ee4d30b0ae35763e742d31e54cb6c615f/chrome/content/zotero/elements/tagsBox.js#L65)) vs click/mousedown ([2](https://github.com/zotero/zotero/blob/2f4a232c41aa3955bba093828b71b9aadc44a21d/chrome/content/zotero/components/virtualized-table.jsx#L725)), but I don't have a touchscreen Windows machine to test with.

It looks like that is exactly the reason why. Long press only dispatches `contextmenu` event without the `mousedown` event as with right-click. So in cases where `contextmenu` events are not handled (like the virtualized table nodes), long press does nothing.

One way to address this would be to just add a dedicated `contextmenu` listener and move the popup opening logic there from the `mousedown` handlers. Is there a reason for why some context menus have to be handled in `mousedown` listeners?

Alternatively, we can leave `mousedown` handlers as is and just call them when `contextmenu` event is picked up. For example, adding this to [_renderItem](https://github.com/zotero/zotero/blob/85f58cc97d2b73af7777c2d6e632dc7b2631531c/chrome/content/zotero/components/virtualized-table.jsx#L1146-L1152) makes the context menu appear on the long press as well.

```
node.addEventListener('contextmenu', (e) => this._handleMouseDown({ button: 2, screenX: e.screenX, screenY: e.screenY }, index), { passive: true });
```

abaevbog:
> I'm not personally a fan of creating fake events when we could just extract/move the logic instead.

Fully agree - I only brought up the alternative in case if there are instances where, maybe, `contextmenu` event wasn't firing properly, so handling in `mousedown` was necessary? Or other reasons like this that I may not be fully aware of. Because if there are no such reasons, I feel like extracting the popup-opening logic into a separate listener shouldn't be that bad...


[...]

PR #4104, opened 2024-05-09
touchscreen longpress displays context menu

abaevbog:
Fixes the issue where long press would not open up context menu in a virtualized table and tab bar because it was handled in mousedown events that longpress on a touchscreen does not dispatch.

This behavior is still observed in the scrollable area of the reader and the note editor.

Addresses: #4094


abaevbog:
https://github.com/zotero/zotero/assets/36271954/059ebbb2-fd3c-4618-8e00-bb6037e4c2bc

dstillman:
A right-click on a tab is now selecting the tab. It should only show the context menu, not select the tab.

abaevbog:
Thank you for pointing that out! I re-pushed so the tab does not get selected on right-click

PR #4105, opened 2024-05-09
Prevent empty item type menu from appearing after Cmd-Shift-N when info section is collapsed

abaevbog:
If the info section is collapsed, do not try to focus and open the itemType menu after item creation via Cmd-Shift-N as it can lead to an empty dropdown. Just focus the header title in that case.

Screenshot 2024-05-09 at 5 37 32 PM

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -1305,6 +1305,11 @@ var ZoteroPane = new function()
 						var type = mru ? mru.split(',')[0] : 'book';
 						await ZoteroPane.newItem(Zotero.ItemTypes.getID(type));
 						let itemBox = document.getElementById('zotero-editpane-item-box');
+						// If the info pane is collapsed, focus the title in the header
+						if (!itemBox._section.open) {
+							document.querySelector("#zotero-item-pane-header editable-text").focus();
+							return;
+						}
 						var menu = itemBox.itemTypeMenu;
 						// If the new item's type is changed immediately, update the MRU
 						var handleTypeChange = function () {
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -1306,7 +1306,7 @@ var ZoteroPane = new function()
 						await ZoteroPane.newItem(Zotero.ItemTypes.getID(type));
 						let itemBox = document.getElementById('zotero-editpane-item-box');
 						// If the info pane is collapsed, focus the title in the header
-						if (!itemBox._section.open) {
+						if (!itemBox.isSectionOpen) {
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/zoteroPane.js
```diff
@@ -1305,6 +1305,11 @@ var ZoteroPane = new function()
 						var type = mru ? mru.split(',')[0] : 'book';
 						await ZoteroPane.newItem(Zotero.ItemTypes.getID(type));
 						let itemBox = document.getElementById('zotero-editpane-item-box');
+						// If the info pane is collapsed, focus the title in the header
+						if (!itemBox._section.open) {
```
dstillman, line 1309: It would be better not to rely on a private property — we do that in large part so that we can feel comfortable changing things within the same file without checking for other references. If this isn't available on a public property, we should expose it.

dstillman:
Probably better to just expose `CollapsibleSection`'s `open` on `ItemPaneSectionElementBase`? The section seems like really just an implementation detail.

abaevbog:
That makes sense! I repushed a getter for `ItemPaneSectionElementBase` named `isSectionOpen` - correct me if the naming could be improved

dstillman:
Why not just `open`?

abaevbog:
No particular reason to not have `open`. I tend to be very verbose at times thinking that something might not be clear enough...

windingwind:
Looks good to me!

chrome/content/zotero/zoteroPane.js
```diff
@@ -1305,6 +1305,11 @@ var ZoteroPane = new function()
 						var type = mru ? mru.split(',')[0] : 'book';
 						await ZoteroPane.newItem(Zotero.ItemTypes.getID(type));
 						let itemBox = document.getElementById('zotero-editpane-item-box');
```
windingwind, line 1307: By the way, do we think using something like `ZoteroPane.itemPane._itemDetails.getPane('info')` to replace the `getElementById` to access the inner elements is more readable?

(here's private property, but we can expose the `_itemDetails` too)

dstillman: `ZoteroPane.itemPane.getSection('info')`?

windingwind: The `ZoteroPane.itemPane` here includes `_messagePane`, `_duplicatesPane`, `_itemDetails`, and `_noteEditor`, which corresponds to the `N item in this view` message, duplicates pane, the container of item pane sections, and the note editor.

The `ItemPane` CE controls the status of these four parts, and the `ItemDetails` controls the status of the item pane sections.

Not sure if it makes sense/is necessary to make a shortcut of the `getSection` from `ItemDetails` in `ItemPane`?

PR #4112, opened 2024-05-10
Lazy load CEs

AbeJellinek:
Since we have a lot of them now and don't need them all on startup. Should speed things up!

Notes:
- Calls of the form `new (customElements.get('ce-name'))` should just be replaced with `document.create[XUL]Element('ce-name')`.
- Calls to `customElements.get('ce-name')` for the purpose of extending one of our CEs will need to be preceded by
   ```js
   if (!customElements.get('ce-name')) {
   	delete document.createXULElement('ce-name'); // Or createElement
   }
   ```
   like we're already doing for Mozilla CEs. (We don't extend any of our own CEs in that way at the moment, though.)


dstillman:
Great!

Issue #4111, opened 2024-05-10
Focus Title in Info pane rather than item-pane header on New Item


dstillman:
After https://github.com/zotero/zotero/pull/3791, we should focus the Title field in the Info pane when creating a new item manually so that you don't have to Tab over extra elements to continue editing the item with the keyboard.

abaevbog:
Sorry, by the title field, do you mean title in the header? If I click on "New item" button in the toolbar and select the item type, the focus is already [moved to into the title](https://github.com/zotero/zotero/blob/59b1d75b98eecd2e910a18a8f77b81b44091dd93/chrome/content/zotero/zoteroPane.js#L1456) for me...

<img width="619" alt="Screenshot 2024-05-10 at 1 18 33 PM" src="https://github.com/zotero/zotero/assets/36271954/30a75667-adff-403e-8ff0-cb62afbe35b2">

abaevbog:
Something to point out here from the accessibility standpoint: my understanding is that, as a general rule, focus changes not explicitly initiated by the user is not a very good thing. Change of context after a button or tab press can be unexpected and disorientating (VPAT 44 points this out in another component). https://www.w3.org/TR/UNDERSTANDING-WCAG20/consistent-behavior-unpredictable-change.html

In this specific instance, though, it's arguably ok, since the title has an announced placeholder and it kinda makes sense that you will be editing an item after it is created. I also see the same being done in thunderbird when you create a new contact.

PR #4113, opened 2024-05-10
Focus title in itemBox after item creation

abaevbog:
When item is created manually, title field in itemBox will be focused instead of the title in the itemPane as before. If the itemBox is collapsed, it will be opened.

In addition, do not move the focus from the itemType menu onto the title in the header if the item is created via a keyboard shortcut and its type was immediately changed. It is generally better to avoid moving focus on behalf of the user, and now that the title is right below the itemType, it's likely no longer necessary.

Fixes: #4111

One observation: to focus the title in the itemBox on manual item creation, we have to open the itemBox if it is collapsed. If an item is created with a shortcut and the itemBox is closed, we can't always focus the title in the header of the itemPane as in #4105 because it may be in "Bibliography Entry" mode, which is not focusable. So for now, I changed it to just tab into the first focusable element of `itemPane`. That way if the title in the header is editable, it will receive focus. Otherwise, focus will land on the scrollable area. 

Focusing on the scrollable area of the itemPane is OK but it's not that useful, I think. So maybe we should just always force-open the itemBox on item creation from the shortcut too? Then, the itemType menu can always be focused and opened. The drawback is that it feels like a lot of things magically happening at the same time (section opening first and then item menu opening up) but it does seem like this would be more useful?

### DIFF
--- a/chrome/content/zotero/elements/itemBox.js
+++ b/chrome/content/zotero/elements/itemBox.js
@@ -755,13 +755,12 @@
 			
 			// Place, in order of preference, after title, after type,
 			// or at beginning
-			var titleFieldID = Zotero.ItemFields.getFieldIDFromTypeAndBase(this.item.itemTypeID, 'title');
-			var field = this._infoTable.querySelector(`[fieldname="${Zotero.ItemFields.getName(titleFieldID)}"]`);
+			var field = this.getTitleField();
 			if (!field) {
 				field = this._infoTable.querySelector('[fieldName="itemType"]');
 			}
 			if (field) {
-				this._beforeRow = field.parentNode.nextSibling;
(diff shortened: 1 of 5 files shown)


abaevbog:
Having tried it, I feel like there is no harm is opening the itemBox regardless if item is opened from "New item" button or via a shortcut. It does feel a bit "magical" that all these sections and menus are opening up but at least it is consistent. I added it as a separate commit so that it's easier to see what it looks like with or without it.

dstillman:
Yes, I think auto-opening the pane is right.

This messed up the placement of creators — `this._beforeRow` means the row creators will be inserted before, not the row before the title field — but I've fixed that and added a test.

dstillman:
Merged

PR #4122, opened 2024-05-14
Use transparent collection & item tree scrollbar

windingwind:


windingwind:
The only two scrollbars with different background colors (especially in collections tree) have become annoying for me every time I open Zotero.

Before:

image


After:

image

adomasven:
UI changes especially to semi-native elements have to be tested across all OS'es (or only applied to the OS it was tested with).

windingwind:
Agree. I just thought we were already applying the same `scrollbar-color` rule to most of the scrollbars on all platforms, so theoretically it's OK (e.g. item pane and tag selector)

I tried MacOS and Windows 11, look fine.

PR #4124, opened 2024-05-14
Move "Show File" out of Locate menu and into File menu

### LINKED ISSUE #4118: Move "Show File" out of Locate menu and into File menu
https://github.com/zotero/zotero/pull/3747#issuecomment-2109599702

Above "Export PDF…"?

AbeJellinek:
And improve `locateMenu.js` Fluent handling:

- `l10nKey` -> `l10nId`
- Only get `l10nArgs` once, since `ViewAttachment` calculates it with a fairly complex getter routine
- Don't make option objects stringify their l10n args themselves

Fixes #4118 (and includes "Show File" -> "Show in Finder" on macOS - will revert if we don't want that)

### DIFF
--- a/chrome/content/zotero/locateMenu.js
+++ b/chrome/content/zotero/locateMenu.js
@@ -602,15 +602,15 @@ var Zotero_LocateMenu = new function() {
 		this.className = "zotero-menuitem-show-file";
 		this.hideInToolbar = true;
 		this.l10nId = "menu-show-file";
-		this.l10nArgs = { count: 0 };
+		this.l10nArgs = { count: 1 };
 		
 		this.canHandleItem = function (item) {
 			return ZoteroPane.canShowItemInFilesystem(item);
 		};
 		
 		this.updateMenuItem = function (items) {
-			let count = items.filter(item => ZoteroPane.canShowItemInFilesystem(item))
-				.length;
+			// We only care about showing 1 or many
+			let count = Zotero.Items.numDistinctFileAttachmentsForLabel(items) > 1 ? 2 : 1;
 			this.l10nArgs = { count };
 		};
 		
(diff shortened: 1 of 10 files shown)


dstillman:
OK, a strong case could be made that this was a bad use of time, but it now says "Show File" if you select a parent item and its primary attachment.

dstillman:
And we want "Show in Finder"/"Show File" to be visible but disabled if no file attachments rather than hidden? We are hiding "Export PDF". Do we want to show "Export PDF" all the time too?

dstillman:
I also fixed a bug where Show File was showing if a parent item only had a URL attachment.

dstillman:
Merged

dstillman:
(If you have a better name than `numDistinctFileAttachmentsForLabel()`, by all means let me know.)

dstillman:
> MAX_ITEMS should be 3, or we should return a label ('zero', 'single', 'multiple'?) instead of numbers. Otherwise we're going to end up with weird localizations in languages with dual forms.

Well, are we? The English string has "one" and "other" set as default (`*`). I was assuming that as long as the localizer sets the non-"one" string as the default, that's what they'll get if we pass 2 as the count.

But in cases like this, I guess there's really no reason to rely on the confusing Fluent plural mechanism syntax at all and we should just have separate strings for `-single` and `-plural`. We used to do that when we shouldn't have (i.e., when a number was embedded in the properties string), but based on our discussions yesterday it seems like that's actually correct here.

dstillman:
(I was actually wondering whether `getBestAttachments` needed to do a DB query and be async or if it could just use `getAttachments()`, but I concluded that, regardless, we would in theory like to move in a direction where attachments could in fact be loaded asynchronously and we didn't need to load all items in the library at startup, so keep it async probably made sense.)

Issue #3611, opened 2024-01-27
Conflict-resolution window broken in redesign


dstillman:
https://forums.zotero.org/discussion/111209/zotero-7-beta-conflict-resolution-window

Much more broken than what @tnajdek saw!

dstillman:
In dark mode too: https://forums.zotero.org/discussion/112705/dark-mode-ui-bug-in-sync-popup

dstillman:
https://forums.zotero.org/discussion/113137/blank-conflict-resolution-page-on-startup-and-sync-debug-id-d1533522669-resolved-see-comment

dstillman:
https://forums.zotero.org/discussion/113840/remote-synchronization-file-selection

dstillman:
@windingwind: I can help with this (e.g., the dialog buttons), but probably easiest for you to fix things related to the item pane CEs.

To start, we can hide everything other than Info to match previous functionality. We could consider showing certain sections, like tags, but we'd have to review the actual CR behavior for those, so probably easiest to start with just Info for now. We'll also want to hide various things, like the Reindex and Migrate Item Note buttons in the attachments pane. (I assume Reindex was hidden before, but I'm not sure.)

windingwind:
OK, I'll look into it. How do we usually debug the CR window?

dstillman:
https://forums.zotero.org/discussion/113990/dark-mode-color-problems-when-there-is-a-sync-conflict

@windingwind: Sorry, missed your question. That'd be great. We have some tests that you could add delays to, but probably easier to just open two copies of Zotero (on separate computers, or in VMs, or in separate profiles), turn off auto-sync, and make conflicting changes. As long as you cancel, you can continue refining.

PR #4130, opened 2024-05-16
Fix merge window

windingwind:
fix: #3611


windingwind:
now:
image

dstillman:
There shouldn't be a horizontal scrollbar. All three panes need to fit within the window width.

windingwind:
OK. Now looks like:

image

chrome/content/zotero/elements/mergeGroup.js
```diff
@@ -68,6 +68,7 @@
 		get data() {
 			return this._data;
 		}
+
```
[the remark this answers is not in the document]
windingwind: That's intentional (to fix `Expected blank line between class members.eslint[lines-between-class-members](https://eslint.org/docs/rules/lines-between-class-members)`, which is defined by our eslint rules)

chrome/content/zotero/merge.xhtml
```diff
@@ -34,18 +34,25 @@
 <window xmlns="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul"
 		xmlns:html="http://www.w3.org/1999/xhtml"
 		onload="Zotero_Merge_Window.init()"
+		drawintitlebar-platforms="mac"
+		width="830" height="500"
```
dstillman, line 38: Don't `width` and `height` not work on `window` anymore?

windingwind: It works somehow. Otherwise, the window will only have a default small size like this:

image

windingwind: I think it's no longer working on most of the XUL elements, but maybe the window is a special case where you need to decide its size before loading any CSS, and that's why we can use `width`/`height` here.

dstillman: Hmm, maybe still works on modal dialogs somehow? Still might be better to use inline `min-width`, since this seems likely to break later?

chrome/content/zotero/elements/mergeGroup.js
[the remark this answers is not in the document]
windingwind: OK, will do a force push

chrome/content/zotero/merge.xhtml
windingwind: done

windingwind:
Disabled the collapsing and open state caching:

image

dstillman:
If I click in the fields, they look like they're editable (though they're not). They should just be read-only.

Screenshot 2024-05-17 at 8 11 43 AM

dstillman:
Oh, that's confusing, though. We should override to avoid that here.

dstillman:
Merged in 7fa7cf310e95eb5da7bfde583a4a7757855b524f. We can fix the editables separately.

Issue #4145, opened 2024-05-21
Editing new middle creator causes infinite recursion and item data inconsistencies


AbeJellinek:
1. Click the + button on a creator row other than the last one
2. Click somewhere else to blur the editable-text
3. Click back into the editable-text, type something, press Enter

Creator gets added at the end instead of in the middle, Zotero slows down a lot, and infinite recursion errors start appearing in the console.

4. Add a new creator at the end

Zotero then freezes for a second and it becomes impossible to click the +/- buttons on any of the creator rows - seems like they're being removed and recreated over and over again.

I'm getting:

```
Uncaught (in promise) InternalError: too much recursion
    stackFramesAsArray resource://zotero/bluebird/debuggability.js:645
    parseStackAndMessage resource://zotero/bluebird/debuggability.js:662
    attachExtraTrace resource://zotero/bluebird/debuggability.js:864
    longStackTracesAttachExtraTrace resource://zotero/bluebird/debuggability.js:492
    _rejectCallback resource://zotero/bluebird/promise.js:508
    _continue resource://zotero/bluebird/generators.js:138
    _promiseFulfilled resource://zotero/bluebird/generators.js:99
    coroutine resource://zotero/bluebird/generators.js:201
    saveTx chrome://zotero/content/xpcom/data/dataObject.js:978
    hideEditor chrome://zotero/content/elements/itemBox.js:1913
    createValueElement chrome://zotero/content/elements/itemBox.js:1431
    _resetStateAfterBlur chrome://zotero/content/elements/editableText.js:332
    blur chrome://zotero/content/elements/editableText.js:426
    blurOpenField chrome://zotero/content/elements/itemBox.js:2216
    moveCreator chrome://zotero/content/elements/itemBox.js:2144
    tryCatcher resource://zotero/bluebird/util.js:16
    _promiseFulfilled resource://zotero/bluebird/generators.js:97
    coroutine resource://zotero/bluebird/generators.js:201
    spawn chrome://zotero/content/xpcom/utilities_internal.js:1788
    spawn chrome://zotero/content/xpcom/zotero.js:1482
    moveCreator chrome://zotero/content/elements/itemBox.js:2143
    modifyCreator chrome://zotero/content/elements/itemBox.js:2019
    hideEditor chrome://zotero/content/elements/itemBox.js:1841
    createValueElement chrome://zotero/content/elements/itemBox.js:1431
    _resetStateAfterBlur chrome://zotero/content/elements/editableText.js:332
[...]

Issue #4143, opened 2024-05-21
Tweaks to empty creator row behavior


dstillman:
After 2071a7c4cf1d80da4afb9cb6270ba422af6f5ca6

- If focus leaves an empty creator row (without the window itself losing focus), maybe it should go away? That wasn't the behavior in Z6, but it's a bit weirder now that the empty row can be in the middle of a list of creators. Also for something like pressing Esc — that just feels like it should make the empty row go away.
- Showing a field-mode button in unsaved rows in place of `+`, as suggested in https://github.com/zotero/zotero/pull/4096#issuecomment-2121337770, is a good idea. Not sure how weird that would look if we kept the `-`, though, and not sure which order we would put them in. (Maybe less weird with the mode switcher first, but then the `-` is misaligned.) Arguably `-` isn't really necessary if we remove the row whenever it loses focus or you press Esc? That's fairly similar to how all other rows work — you don't need to "close" field rows. But possible people would feel better with a `-` to click…
- Focus goes to the wrong place if you delete an empty row via the keyboard:

https://github.com/zotero/zotero/assets/783082/315d07ca-de34-4e37-9fc8-d1036bc17183

abaevbog:
> Focus goes to the wrong place if you delete an empty row via the keyboard:

This, as well as a similar bug of a wrong component being focused on tab from a just-filled creator field, should be fixed in #4152

dstillman:
> With this setup, if you open up a dropdown, it will not delete the unsaved row but I think it's better to be on the safer side to avoid deleting row when one doesn't expect it. Correct me if this setup sounds off.

No, that sounds perfect.

Is that an existing icon for the creator-mode toggle? Regardless, I don't think it's at all clear what that is, particularly in the + position. (It's also misaligned, but I don't think fixing that would help much.)

In any case, I think I was wrong and this doesn't actually make a lot of sense. This button just isn't really more relevant for empty rows than it is for existing rows. Ideally people aren't manually entering data to begin with, names from various sources can end up in the wrong mode, and being able to see the mode at a glance is [actually useful for support and troubleshooting](https://forums.zotero.org/discussion/comment/464065/#Comment_464065).

[...]

PR #4152, opened 2024-05-22
Fixes to itembox creator row bugs

abaevbog:
- minor refactoring of blurOpenField to call blur directly on the input of `editable-text`. That is to avoid having multiple 'blur' events triggering `hideEditor` which may cause multiple re-renderings and potentially an infinite loop. It can also interfere with cases when we want to add a new creator row on shift-enter after the first refresh - second unwanted re-render would remove the newly added empty creator row.
- remove call to blurOpenField where it is not necessary and make it non-async. With `editable-text`, the fields blur and the item gets saved whenever you click outside of it, so that's no more necessary and confusing.
- on focusin of the empty unsaved creator row, use its future unsaved id that reflects its positioning instead of its real id (which technically is the very last creator row). That way, a tab from a creator row that was just filled will land to the proper component after render, as opposed to focusing the very last row.
- similar approach to removing the unsaved creator row when the focus is in it - focus the row above it, as opposed to the last row.

Fixes: #4145
Addresses part of: #4143

### DIFF
--- a/chrome/content/zotero/elements/editableText.js
+++ b/chrome/content/zotero/elements/editableText.js
@@ -325,11 +325,11 @@
 				return;
 			}
 			this._resetStateAfterBlur();
+			this.dispatchEvent(new Event('blur'));
 		};
 		
 		_resetStateAfterBlur() {
 			this._ignoredWindowInactiveBlur = false;
-			this.dispatchEvent(new Event('blur'));
 			this.classList.remove('focused');
 			this._input.scrollLeft = 0;
 			this._input.setSelectionRange(0, 0);
--- a/chrome/content/zotero/elements/itemBox.js
+++ b/chrome/content/zotero/elements/itemBox.js
@@ -115,7 +115,6 @@
 				
 				this.modifyCreator(index, fields);
 				if (this.saveOnEdit) {
-					await this.blurOpenField();
 					await this.item.saveTx();
 				}
 			});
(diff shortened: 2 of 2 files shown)


abaevbog:
@AbeJellinek, just to be sure, does this take care of that infinite recursion bug for you?

abaevbog:
Well, `editable-text` doesn't trigger multiple blur events per se. I will revise the main commit message to make it clearer. Let me explain what I found in simpler terms and then I'll try to make it more concise.

There are 2 reasons (and 2 fixes) that I see to the infinite loop problem:
1. the `blurOpenField` causes `blur` event on the `editable-text` which calls `hideEditor` which calls `modifyCreator` which calls `moveCreator` which called `blurOpenField` again, thus starting the loop. So one thing to do it to just not call `blurOpenField` in the `moveCreator` - I did just remove it because it really is redundant in there to begin with.
2. the `hideEditor` calling `modifyCreator` that locks that loop is called on `blur` event from `editable-text`. If called via `.blur()` method, `editable-text` always emits the `blur` event, regardless if the focus was inside of it or not. This is what kept calling `hideEditor` despite the focus not being present there anymore. This is what I was referring to by "multiple 'blur' events". Is that something that's worth changing as well here?

abaevbog:
I added that tweak to `editable-text` that would also prevent the infinite loop from happening. 

Correct me if this is something we don't want. In general, from what I see, calling `.blur()` on a non-focused `input` or `textarea` does nothing and does not emit the `blur` event, so perhaps this makes sense?

abaevbog:
> So maybe we move this.dispatchEvent(new Event('blur')); out of _resetStateAfterBlur() and back into _handleBlur(), but keep everything else the same?

Oh, gotcha! Yes, dispatching `blur` from `editable-text` in `_handleBlur` makes sense to me. That way, it still gets out when needed but it won't fire if there was no actual blur happening.

If I put back `blurOpenField` to `moveCreator` and run it with the last commit that calls `.blur` directly, the infinite loop does not happen as well, so I think this may be the way?

abaevbog:
Ok, awesome! Glad this fixes it - infinite loops are bad. 

At some point, I will look into adding a more comprehensive test suit for the itemBox as well

abaevbog:
I just squashed everything into one commit to make the commit message a bit clearer. 

One small tweak I did add was remove `Zotero.spawn` from `modifyCreator`, since it doesn't `yield` or `await` for anything anymore, so there's no need for it to return a promise. It's not a big deal but i think it's good to keep things straightforward to avoid confusions in the future

PR #4155, opened 2024-05-22
Allow item pane sections to declare dependencies for rerender, fix Libraries and Collections not updating

### LINKED ISSUE #4154: Libraries and Collections box doesn't update currently selected collection
When the item doesn't change.

1. Select an item in My Library
2. Add it to a collection if it isn't already in one
3. Click that collection under Libraries and Collections

My Library remains highlighted and isn't clickable.

AbeJellinek:
- By default, rerender is only necessary when tab or item ID changes
- Libraries and Collections box additionally depends on the collection tree row
	- Pass `collectionTreeRow` down to item pane sections

Fixes #4154

### DIFF
--- a/chrome/content/zotero/elements/itemPaneSection.js
+++ b/chrome/content/zotero/elements/itemPaneSection.js
@@ -151,7 +151,9 @@ class ItemPaneSectionElementBase extends XULElementBase {
 
 	_resetRenderedFlags() {
 		// Clear cached flags to allow re-rendering
+		delete this._syncRenderDependencies;
 		delete this._syncRenderItemID;
+		delete this._asyncRenderDependencies;
 		delete this._asyncRenderItemID;
 	}
 
--- a/chrome/content/zotero/elements/itemPaneSection.js
+++ b/chrome/content/zotero/elements/itemPaneSection.js
@@ -126,12 +126,11 @@ class ItemPaneSectionElementBase extends XULElementBase {
 		let key = `_${type}RenderItemID`;
 		let pendingKey = `_${type}RenderPending`;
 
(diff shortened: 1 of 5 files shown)


windingwind:
Looks good to me!

chrome/content/zotero/elements/itemPaneSection.js
```diff
@@ -126,12 +126,11 @@ class ItemPaneSectionElementBase extends XULElementBase {
 		let key = `_${type}RenderItemID`;
```
windingwind, line 126: If we are setting an array to this property, maybe better to use a different name? Renaming it may also affect some tests/components.

chrome/content/zotero/elements/itemPaneSection.js
```diff
@@ -140,10 +139,10 @@ class ItemPaneSectionElementBase extends XULElementBase {
 			return true;
 		}
 
-		if (!pendingFlag && renderFlag && newFlag === renderFlag) {
+		if (!isPending && isRendered) {
 			return true;
 		}
-		this[key] = newFlag;
+		this[key] = newDependencies;
```
windingwind, line 145: Will break some item pane tests. We used ti rely on the _renderItemID to busy wait for rendering a specific item in the tests. (Busy wait is not very elegant, though)

windingwind:
The item pane header seems to be broken for me

windingwind:
The item pane header seems to be broken for me. The relevant tests fail and changes to the header mode are not applied until the next restart.

windingwind:
Thanks for the quick fix. Seems nice!

Issue #4141, opened 2024-05-21
Improve feed item pane header


yexingsha:
The most pressing issue is at smaller pane widths, the long header buttons cause the entire pane to overflow:

<img width="342" alt="Screenshot 2024-05-20 at 23 22 52" src="https://github.com/zotero/zotero/assets/57771795/47087c8d-ae7c-4e5d-bfb9-f2c0b9baa58e">

I think we should enable truncation for the "Add to My Library" button, because it might be even wider in other locales, and the button should still be intelligible with only icon and "Add to My...". There are other things we should improve as well (mocks in Figma):

1. There is currently almost no spacing between the title area and the buttons. Should be 4px in tight, and 8px in comfortable 
2. Decrease the margins around and between the buttons to 8px
3. Improve the paddings of the two buttons. They are meant to look like system native buttons, but right now they look slightly off on both mac and windows. The padding before the icon in "Add to My Library" is especially large
4. Are the quotation marks around "My Library" necessary? I think we can remove those
5. We should use `accent-blue` for the library icon
6. The color of the divider and the chevron are both wrong in dark mode

windingwind:
> Are the quotation marks around "My Library" necessary? I think we can remove those

Maybe necessary? It can be weird without the quotation, e.g. `Add to Add to`, if the collection name is "Add to".

dstillman:
Yeah, annoying in the case of My Library, but they're there because the collection/group name can be anything.

PR #4159, opened 2024-05-23
Fix item pane header

windingwind:
fix: #4141
Move split-menu-button styles to SCSS

- [x] There is currently almost no spacing between the title area and the buttons. Should be 4px in tight, and 8px in comfortable
- [x] Decrease the margins around and between the buttons to 8px
- [x] Improve the paddings of the two buttons. They are meant to look like system native buttons, but right now they look slightly off on both mac and windows. The padding before the icon in "Add to My Library" is especially large
- [x] ~~Are the quotation marks around "My Library" necessary? I think we can remove those~~
- [x] We should use accent-blue for the library icon
- [x] The color of the divider and the chevron are both wrong in dark mode


windingwind:
fix: #4116

windingwind:
Hi @yexingsha, this is what this PR looks like, please check if I missed anything:

MacOS:

image
image


Windows:
image
image

yexingsha:
On macOS, the text and icon are positioned 1px too high within the button (it's more obvious with the library icon). And both the spacing and padding are still a bit too large. Seems like the extent of the button is about 2px wider than its optical width, so need to compensate for that:

Screenshot 2024-05-29 at 17 25 44





The spacing between the icon and text can be reduced to 4px.



The header looks much better now at narrow width, but looks weird at larger width:

Screenshot 2024-05-29 at 16 31 51

I think once the "Add to" button grows wide enough to fully display the text, both buttons should grow in width as the pane gets wider. And perhaps the split menu button should still align to center, even though it does go against convention.

windingwind:
Thanks. Done:

- [x] the text and icon are positioned 1px too high within the button
- [x] And both the spacing and padding are still a bit too large. Seems like the extent of the button is about 2px wider than its optical width, so need to compensate for that
- [x] The spacing between the icon and text can be reduced to 4px.
- [x] And perhaps the split menu button should still align to center, even though it does go against convention.
- [x] I think once the "Add to" button grows wide enough to fully display the text, both buttons should grow in width as the pane gets wider

https://github.com/zotero/zotero/assets/33902321/670b5f6a-c343-40dc-8aa6-c7a709e2edbc

yexingsha:
This looks great!

Issue #4139, opened 2024-05-19
Unreliable attachment(s) pane tests


dstillman:
Failing intermittently in CI:

https://github.com/zotero/zotero/actions/runs/9130689547/job/25108168031
https://github.com/zotero/zotero/actions/runs/9145173526/job/25143847868

windingwind:
Thanks, I'll investigate.

PR #4161, opened 2024-05-23
Refactor AttachmentPreview render and discard

windingwind:
fix: #4139

### DIFF
--- a/chrome/content/zotero/elements/attachmentBox.js
+++ b/chrome/content/zotero/elements/attachmentBox.js
@@ -291,15 +291,10 @@
 		}
 
 		notify(event, _type, ids, _extraData) {
-			if (event != 'modify' || !this.item || !this.item.id) return;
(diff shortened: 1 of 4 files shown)


windingwind:
The unreliable tests seem to be caused by the render/discard mechanism of the attachment preview. Dense calls to `render`/`discard` have the chance to mess up the iframe and reader initialization. That's also very likely the cause of random failures in the `render` we used to see when switching between items very fast.

So in this PR, I did a refactor to the `AttachmentPreview`, where any calls to `render` and `discard` updates the `_lastTask` state immediately and process the `_lastTask` until the the currently processed task finishes.

I've ran the Item pane tests locally for 20+ times and looks fine. It also seems to slightly improve the time of showing preview when switching item very fast, since many intermediate render/discard tasks are skipped and only the last queued task will be actually processed.

PR #4165, opened 2024-05-23
Item box: Save text selection along with focus

### LINKED ISSUE #4164: Item box: Modifying a field and then clicking on another field selects all text in the `editable-text`
https://forums.zotero.org/discussion/114612/zotero-7-beta-select-all-when-clicking-inside-an-editable-field-in-the-item-pane

Probably related to recreating the fields on refresh.

AbeJellinek:
@abaevbog: Let me know if this looks alright!

I didn't touch them in this PR, but some of the other places where `this._selectField` is set don't seem to be doing anything / the right thing:

- Tabbing out of an empty unsaved creator row brings me to the button bar on the next field, not [the last name](https://github.com/AbeJellinek/zotero/blob/fcaed18d0d6555e5a5e5be7fbed5071e7d508e16/chrome/content/zotero/elements/itemBox.js#L1530)
- Shift-Enter from the last creator before "More creators..." adds a new row at the end, not [in the middle](https://github.com/AbeJellinek/zotero/blob/fcaed18d0d6555e5a5e5be7fbed5071e7d508e16/chrome/content/zotero/elements/itemBox.js#L1760)
- Pasting a tab-separated creator string selects a new empty creator, not [the last name of the last new creator](https://github.com/AbeJellinek/zotero/blob/fcaed18d0d6555e5a5e5be7fbed5071e7d508e16/chrome/content/zotero/elements/itemBox.js#L1827)

Fixes #4164

### DIFF
--- a/chrome/content/zotero/elements/itemBox.js
+++ b/chrome/content/zotero/elements/itemBox.js
@@ -2233,8 +2233,7 @@
 				}
 			}
 			
-			this._selectField = null;
-			this._selectFieldSelection = null;
+			this._clearSavedFieldFocus();
 		}
 
 		getTitleField() {


abaevbog:
@AbeJellinek great idea to add this! One observation: the selection is saved on `focusin` - but what if the selection changes after that without leaving the editor? Presumably the initial selection, as opposed to a more recent one, is restored? Would it make sense to listen to `select` event and save it from there - that way the subsequent selection after focus would be saved?

Trying to demonstrate this is kinda hard, since the main idea behind `focusin` stuff was to preserve the focus after a random refresh and refreshes generally happen after a the focus leaves an `editable-text` fields. So I was triggering a refresh from the console after a timeout - correct me if there's a more appropriate way to approach this.


https://github.com/zotero/zotero/assets/36271954/2f359fe7-3f17-4b5b-8377-bea37b112693

abaevbog:
> Shift-Enter from the last creator before "More creators..." adds a new row at the end

Yup, I can reproduce thanks! 

> Pasting a tab-separated creator string selects a new empty creator

Same here. To confirm, this happens specifically after paste on the new empty creator row, right?

> Tabbing out of an empty unsaved creator row brings me to the button bar on the next field

I am actually struggling to reproduce this one. What am I missing here?

https://github.com/zotero/zotero/assets/36271954/72550a97-5bbc-4258-8a10-c5d9454cab63

abaevbog:
I like the idea of not having the `focusin` listener!

Right now, though, I think we have a conflict with a few existing behaviors. I used `this._selectField` also as a way to tell `itemBox` what to focus after refresh. So, for example, after `paste`, we want to [focus](https://github.com/zotero/zotero/blob/3176516b797ce5149991fe28c3d464583a11130b/chrome/content/zotero/elements/itemBox.js#L1808) the last creator row (There is a glitch that it doesn't pick the correct field but that's a separate thing) And after a creator row is removed, we focus the last name of the previous creator row. 

If we reset the `this._selectField` before refreshing, I'm afraid we loose this ability to tell itemBox which row to focus once `render` happens. So, right now, if I paste a number of `\n`-separated creators on a new creator row after the 1st creator, the focus will remain on the 2nd creator row after paste because the `this._selectField` that is set in `handleCreatorPaste` is cleared. And if I have a single creator row and I tab onto the `-` button, the focus will get lost because there will be no `-` button after re-render.


As an alternative, we could keep the `this. _selectField` in `focusin` and only save the selection before render. It's not pretty though. Or maybe it makes sense to split `this._selectField` into 2 fields - `this._lastFocused` (used just to make sure focus does not disappear from the `itemBox` on a random refresh)  and `this._toFocusNext` (used for cases like creator row deletion or handling paste to indicate what to focus after render)?

Also, just for understanding the context, are there any instances besides a random refresh of the `itemBox` when we would want to restore the selection of the focused input?

chrome/content/zotero/elements/itemBox.js
```diff
@@ -2188,6 +2168,75 @@
...
+			
+			if (this._selectFieldSelection) {
+				let input = refocusField.querySelector("input, textarea");
+				if (input) {
+					input.setSelectionRange(...this._selectFieldSelection);
+				}
+			}
+			
+			this._selectField = null;
+			this._selectFieldSelection = null;
```
abaevbog, line 2237: Would it be cleaner to reuse `_clearSavedFieldFocus` from above?

abaevbog (commented):
> How about that? Focus after pasting multiple creators works for me.

Oh yeah, that's a way easier way around it! Looks good to me! Just one tiny question below 

> Kind of, in that when you edit one field and then click into another, the second field gets focused before the changes from the first field are saved, so we need to save and restore the selection in the second field on either side of the refresh.

I see - in the current build, if I focus the field it will end up having all text selected vs keeping the cursor where it was clicked as right here. Got it!

PR #4172, opened 2024-05-27
Show banner to 32-bit Windows users to reinstall 64-bit version

### LINKED ISSUE #3940: Prompt 32-bit Windows users to reinstall 64-bit version
Or maybe a banner

tnajdek:
I opted for a banner that can be dismissed but will reappear on every restart since there is no reason not to reinstall. However, I expect some people won't be able to switch to a 64-bit version (e.g., because their PCs are managed by their organization). Should we add "Remind me later" and/or "Don't ask again" options?

I use `Zotero.arch` to check the build architecture and `navigator.oscpu` for the OS architecture. Everything else is adapted from other banners. Actually, we should probably tweak the other banners to use the new CSS classes to reduce duplication—I can do it once this is merged.

win11-x64-zotero32

Close #3940.

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -6031,7 +6031,7 @@ var ZoteroPane = new function()
 
 	this.showArchitectureWarning = async function () {
 		const remindInterval = 60 * 60 * 24 * 30;
-		const isWin64 = (await Services.sysinfo.processInfo).isWow64;
+		const isWow64 = (await Services.sysinfo.processInfo).isWow64;
 		const is32bitBuild = Zotero.arch === 'x86';
 		const lastDisplayed = Zotero.Prefs.get('architecture.warning.lastDisplayed') ?? 0;
 		
@@ -6039,7 +6039,7 @@ var ZoteroPane = new function()
 			return;
 		}
 
-		if (Zotero.isWin && isWin64 && is32bitBuild) {
+		if (Zotero.isWin && isWow64 && is32bitBuild) {
 			let panel = document.getElementById('architecture-warning-container');
 			let action = document.getElementById('architecture-warning-action');
 			let close = document.getElementById('architecture-warning-close');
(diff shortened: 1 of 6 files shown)


dstillman:
And yes, we probably do need "Remind Me Later" that reminds in, say, 30 days. Clicking the X would still show again on restart. I think not having "Don't Ask Again" is OK.

tnajdek:
> I don't suppose you tested this on real 32-bit hardware?

I did, the screenshot is from a 32bit Zotero build running on 64bit Windows 11. I also verified it doesn't trigger when it shouldn't, i.e., 64bit Zotero on 64bit Windows and 32bit Zotero on 32bit Windows (for that last one I used a VM, so technically not real hardware, but shouldn't make a difference?).

> do we think navigator.oscpu is still accurate in this way? 

It's [fake](https://searchfox.org/mozilla-central/source/netwerk/protocol/http/nsHttpHandler.cpp#850-855) but from what I can tell 32bit/64bit is respected: https://searchfox.org/mozilla-central/source/netwerk/protocol/http/nsHttpHandler.cpp#949-958

> A safer option might be to just check (await Services.sysinfo.processInfo).isWow64

That's better, thanks, I'll tweak the code to use that instead.

> we probably do need "Remind Me Later" that reminds in, say, 30 days

OK, I'll add that

dstillman:
> and 32bit Zotero on 32bit Windows

Ah, OK, that was the only one I was concerned about — recommending the 64-bit build erroneously on 32-bit Windows.

tnajdek:
Tweaked and rebased

chrome/content/zotero/zoteroPane.js
```diff
@@ -6027,7 +6028,44 @@ var ZoteroPane = new function()
 			this.hideRetractionBanner();
 		}
 	};
-	
+
+	this.showArchitectureWarning = async function () {
+		const remindInterval = 60 * 60 * 24 * 30;
+		const isWow64 = (await Services.sysinfo.processInfo).isWow64;
```
dstillman, line 6034: I adjusted this variable name. `isWow64` means specifically 32-bit on 64-bit, so `isWin64` is a little misleading, even if it doesn't matter for the context in which we're using it.

dstillman:
Merged

Issue #4150, opened 2024-05-22
Optimize tab loading at startup


dstillman:
Along with https://github.com/zotero/zotero/issues/4149, there are a couple other non-ideal tab behaviors during startup:

1. We don't show icons, or even leave space for them.
2. If a reader tab was selected, we still start with the library view selected and switch to the reader tab a number of seconds later. It would be better to select the reader tab immediately, even if it can't be rendered yet.

Issue #4149, opened 2024-05-22
Tabs disappear if clicked on during startup


dstillman:
https://forums.zotero.org/discussion/114548/zotero-7-beta-disappearing-tabs-during-loading

I can reproduce this.

abaevbog:
Yes - only on windows though for me. I believe this happens because when an unloaded tab is selected, it gets [closed](https://github.com/zotero/zotero/blob/a3857cb5e05cecf8a41dde86cc8dc191166bf057/chrome/content/zotero/tabs.js#L465), after which Reader [creates](https://github.com/zotero/zotero/blob/a3857cb5e05cecf8a41dde86cc8dc191166bf057/chrome/content/zotero/xpcom/reader.js#L1019) and selects a new tab. The gap between the old tab closure and new tab creation by the reader is when it looks like there's no tab.

As an idea, I am trying to avoid closing and re-opening the tab altogether, and instead reuse existing tab (if one exists) when reader is opened and just change its type from `reader-unloaded` to `reader`. That should address the disappearing tab issue and maybe allow us to simplify the overall tab-handling logic a bit - I remember adding some workaround to deal specifically with tabs closing and immediately being reopened.

Correct me if there is any specific reason I may be missing as to why unloaded tabs must be closed and re-opened instead of reused.

PR #4174, opened 2024-05-28
tabs startup loading optimization

abaevbog:
- if an unloaded tab is being opened, do not close the tab and have reader re-open a fresh tab. Instead, keep the tab as is, have reader use the tab as a container and just change the tab's type. It should fix a glitch on windows when if you click on a tab during initial loading, it will disappear until reader re-adds it. A few tweaks to allow unloaded tabs be generally "selectable".
Fixes: #4149
- during startup loading, select the yet-unloaded tab right away. That way, a tab is almost immediately selected, instead of being stuck on the library tab until the reader is ready.
- if the items are not loaded yet, use a placeholder icon for tabs. Fixes: #4150
- updated handling for reader loading message. When an unloaded tab is selected during initial load, the loading message needs to be displayed but, there is no loaded reader yet. So in `Zotero_Tabs.select` display the message, wait for the reader to load, resolve the `_initPromise` in an async function, and hide that message after. Removed loading message handling from contextPane to remain concise

### DIFF
--- a/chrome/content/zotero/elements/contextPane.js
+++ b/chrome/content/zotero/elements/contextPane.js
@@ -171,7 +171,7 @@
 				ZoteroContextPane.showLoadingMessage(false);
 				this._sidenav.hidden = true;
 			}
-			else if (Zotero_Tabs.selectedType == 'reader') {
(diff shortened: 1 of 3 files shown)


dstillman:
Works much better!

PR #4178, opened 2024-05-29
Tweaks to itembox unsaved creator row

abaevbog:
- remove unsaved creator row on escape and on blur. The blur part is a bit tricky, since we want to remove the empty unsaved creator row on blur only sometimes. For example, the focus leaves the `editable-text` when you click on the options menu or open the creator type dropdown but the row should not be deleted. So I went with the following logic: delete unsaved creator row if the focus lands outside of the itemPane (e.g. you click on the item tree) or if the focus lands on an input outside of the row. Correct me if this sounds off.
- add switch creator mode button to unsaved creator row instead of the always disabled `+` button.
- fix unsaved creator focus glitch on save. If you add a new row, type something and then click on the creator row below it, the focus will jump to the row above it after refresh. To fix it, calculate and use `unsavedIndex` for all creator rows, not
only unsaved one, when setting `this._selectField`.
- fixed bug where focus got lost from some buttons. All focusable elements require an id so that after re-render we can find and refocus the element. Added ids to options and open link icons that were previously missing.

Fixes: #4143
Fixes: #4241

### DIFF
--- a/chrome/content/zotero/elements/itemBox.js
+++ b/chrome/content/zotero/elements/itemBox.js
@@ -806,8 +806,11 @@
 					// immediately hidden
 					this._displayAllCreators = true;
 					
-					if (this._addCreatorRow) {
-						this.addCreatorRow(false, this.item.getCreator(max - 1).creatorTypeID, true);
+					if (this._addCreatorRow !== false) {
+						// Insert an empty creator row in a specified location
+						let beforeCreator = this.querySelector(`#itembox-field-value-creator-${this._addCreatorRow}-lastName`);
+						let beforeRow = beforeCreator?.closest(".meta-row") || null;
+						this.addCreatorRow(false, this.item.getCreator(max - 1).creatorTypeID, true, beforeRow);
 						this._addCreatorRow = false;
 					}
 				}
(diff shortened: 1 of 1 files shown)


abaevbog:
Having the `switch-mode` button on a new row is nice. It does run into an issue I noticed: if you have a creator that you paste and then click on the button to change its mode, it will actually do nothing because clicking on a button immediately blurs the `editable-text`. It's peculiar because it doesn't happen with all `toolbarbutton`s (e.g. ones in the sidebar have no such effect). So it's likely for that button to be useful, `editable-text` should not be blurred on click. I'll try to track down why it happens

abaevbog:
@AbeJellinek here I have a few more tweaks to itemBox. How is this looking to you?

The first commit is just a small improvement to remove the unsaved creator row on Escape if something else gets focused.

The second commit has a number of smaller fixes to bugs that I recently encountered including the issues you brought up in https://github.com/zotero/zotero/pull/4165#issue-2313280474 with pasting creators and wrong row getting focused on shift-Enter.  

The two commits don't necessarily have to go together - let me know if splitting them into separate PRs would be better.

Going over all these small bugs really made me want to have a large test suite for itemBox at some point...

dstillman:
> So I went with the following logic: delete unsaved creator row if the focus lands outside of the itemPane (e.g. you click on the item tree) or if the focus lands on an input outside of the row. Correct me if this sounds off.

This doesn't account for just tabbing to the creator type menu on the next row. That should remove the empty row. It's weird that it doesn't disappear until you get to the Last field.

(But if hiding the row reliably is too complicated, we can also just forget about it.)

abaevbog:
> This doesn't account for just tabbing to the creator type menu on the next row.

That was a great point! I tried a couple of things and I think this approach is the best: in `focusout` handler, we skip a tick to know where exactly the focus lands (instead of trying to use `event.relatedTarget`) and delete the unsaved empty creator row if the focus is outside of it. Ignore cases when `zotero-view-item` or `main-window` are focused (where focus lands when menus are being closed and opened). That way, tabbing onto the creator type dropdown after the empty one will also removed the empty creator row

abaevbog:
I also added here a fix to the focus being lost on some buttons reported on the forums: https://forums.zotero.org/discussion/comment/465664/#Comment_465664

dstillman:
Rebased

Issue #4191, opened 2024-05-31
Quick Format: Cmd/Ctrl-↓ can't be used immediately after selecting item


dstillman:
After the accessibility/focus changes, you now have to press left-arrow to refocus the bubble before you can press Cmd/Ctrl-↓ to open the citation-options panel. This is a bit of a usability regression, and it means that the guidance panel description ("After you've made your selection, click the bubble or press Ctrl-↓…") isn't even really accurate anymore.

Not sure what to do about this. It makes sense why it doesn't work, but maybe we could maybe cheat and accept Cmd/Ctrl-↓ immediately after an item selection the same way we're accepting a page number even though the bubble isn't technically selected? We'd just close the reference panel and show the options dialog instead.

abaevbog:
During the initial big accessibility-focused rewrite, it was sometimes hard to tell which behaviors were desired features and which were just dictated by the `contenteditable` setup that did not allow things to get properly focused. I must have misidentified `Ctrl-ArrowDown` behavior as the later and it was indeed removed.

Right now, when a bubble is focused, just `arrowDown` opens the options dialog (`ctrl/cmd` is not required). That was done mainly to avoid conflicts with screen readers (https://github.com/zotero/zotero/pull/3464#issuecomment-1804816874 mentions this among other things), as JAWS uses `Ctrl` as a modified for many of its own shortcuts. And, besides, when a bubble is focused, the reference panel gets hidden, so it felt like there's no more need to use a modifier key.

Since this is a behavior we want to keep, I think we could just add `ctrl/cmd-arrowDown` as a general keypress handler to all inputs that will open up the options dialog of the bubble to the left of it. It would not work in cases when most screen readers are active but I think it's not an issue here, since focusing a bubble and pressing `arrowDown` is what the `aria-description` components instruct one to do anyways.

dstillman:
Oh, I see. That might actually be fine — instead of a complicated, undiscoverable (if you don't read the guidance panel) modifier-key shortcut, you can just press left-arrow + down-arrow. That's arguably not really much different.

abaevbog:
Right, discoverability was certainly another consideration as while working through the QF I also struggled with finding out all the available options. So the only reason to bring back `ctrl-arrowDown` now in this form, as far as I can see, is just as a legacy functionality that folks may be used to. I did not see but were there any reports of people trying to use the old shortcut and being disappointed that it doesn't work anymore?

Otherwise, we could just update the guidance panel?

[...]

Issue #4187, opened 2024-05-31
Missing padding in quick-format guidance panel


dstillman:
![qf-guidance](https://github.com/zotero/zotero/assets/783082/0bf2d099-4083-421b-9841-876fe8cbf215)

Seen on Linux — not sure about others

PR #4193, opened 2024-05-31
quickFormat guidance panel fixes

abaevbog:
- added 8px padding around text on linux. Mac and win already have some padding around text through --panel-padding. Fixes: https://github.com/zotero/zotero/issues/4187
- guidancePanel.js fetches text from .ftl file instead of zotero.properties
- added first-run-guidance-quickFormat to zotero.ftl with updated text to better reflect current functionality. Fixes: https://github.com/zotero/zotero/issues/4191
- removed unnecessary special treatment for qf guidance panel on mac
- moved guidance panel text for itemBox creator menu into zotero.ftl

### DIFF
--- a/chrome/content/zotero/elements/guidancePanel.js
+++ b/chrome/content/zotero/elements/guidancePanel.js
@@ -76,7 +76,7 @@
 		 * @param {DOMElement} [options.forEl] Anchor node
 		 * @param {Boolean} [options.force] Show even if already shown
 		 */
-		show(options) {
+		async show(options) {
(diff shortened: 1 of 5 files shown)


dstillman:
> So would it be better to just move text for <guidance-panel id="zotero-author-guidance"> into zotero.ftl as well?

Yep.

abaevbog:
Cool. I repushed with `guidance-panel` only fetching text from `zotero.ftl` and moved the remaining `first-run-guidance-authorMenu` text to `zotero.ftl` as well

Issue #4188, opened 2024-05-31
Titlebar in Quick Format dialog on Wayland


dstillman:
This is stock Ubuntu 22.04:

![qf-guidance](https://github.com/zotero/zotero/assets/783082/893126e3-2101-475c-8da2-591e448cea8c)

@adomasven said he's not seeing it on Xorg.

If we can't avoid the titlebar for now, maybe we want to revert the red styling (which we maybe weren't doing before on Linux?).

abaevbog:
Hm, yes indeed. Xorg has no titlebar as expected but wayland does. Interesting, I'll play around to see why that is.

> If we can't avoid the titlebar for now, maybe we want to revert the red styling (which we maybe weren't doing before on Linux?).

Perhaps... We were indeed not doing the red styling before. It was added in #3889 to try to make linux look more like other platforms. I also thought that it did look nicer with a bit of a padding around the dialog than without it as before.

abaevbog:
I thought I had a solution but turns out if when I run `./app/script/build_and_run` from vscode on linux, it runs with default environment variable `GDK_BACKEND=x11` which ignores the wayland setup. Not helpful here but just good to know about it.

PR #4195, opened 2024-06-03
Fix QF titlebar on wayland

windingwind:
fix: #4188


windingwind (commented):
Only test on ubuntu 22.04, seems to be working

abaevbog:
For me, just setting `drawintitlebar=true` on window made the entire qf window resizable. So right now, i can drag the corner of the screen to get this. Is it just me? If others can reproduce it, we may need to tweak this a bit more. I have been trying to go this way and disable resizing but no luck there yet.

Screenshot from 2024-06-03 07-01-23

Issue #4142, opened 2024-05-21
Improve merge pane


yexingsha:
1. Both 'groupbox' and 'zotero-duplicates-merge-item-box-container' have 8px inline padding. One of those is unnecessary.

2. Right now when the item pane is collapsed, users have to drag on the edge of the sidenav or go into the View menu to open it back up. I think when people go into Duplicate Items and select items, they almost always want to do merge actions, so it makes sense here to automatically open the item pane if it was previously collapsed.

3. When #4141 causes the feed item pane to overflow, the overflow state can persist into merge panes, but not regular item panes, so there might be something going on.

windingwind:
Thanks!

> Right now when the item pane is collapsed, users have to drag on the edge of the sidenav or go into the View menu to open it back up. I think when people go into Duplicate Items and select items, they almost always want to do merge actions, so it makes sense here to automatically open the item pane if it was previously collapsed.

Maybe we should also consider auto-open item pane in trash?

> the overflow state can persist into merge panes

That's because the two panes share the same element. No worry

yexingsha:
> Maybe we should also consider auto-open item pane in trash?

I feel like that's unnecessary, because:

- Unlike merge actions which are can only be done in item pane, Restore to Library and Delete Permanently both exist in context menu. In fact context menu might even be the preferred action here, especially for multiple items, which makes accessing those buttons via the item pane less important.

- Sidenav is enabled in trash, so if users do want to reopen the item pane, they only need one click, which is less troublesome than dragging the resizer or navigating the View menu.

Combining these two factors, I think we should preserving the collapsed state of the item pane in trash.

yexingsha:
Looks good, just need to fix the merge button menu so that it doesn't hug the screen edge.

Current menu placement:

<img width="353" alt="Screenshot 2024-06-03 at 23 26 19" src="https://github.com/zotero/zotero/assets/57771795/15a2eace-b560-4608-b69b-7f988f96490a">

Compared to regular item info context menu:

<img width="347" alt="Screenshot 2024-06-03 at 23 39 38" src="https://github.com/zotero/zotero/assets/57771795/dd2d50a4-ae16-4c68-b739-68d75d2a58ff">

PR #4200, opened 2024-06-04
Improve merge pane

windingwind:
resolve: #4142

### DIFF
--- a/chrome/content/zotero/elements/duplicatesMergePane.js
+++ b/chrome/content/zotero/elements/duplicatesMergePane.js
@@ -37,12 +37,13 @@
 				</hbox>
 			</groupbox>
 			
-			<groupbox flex="1">
-				<description id="zotero-duplicates-merge-field-select">&zotero.duplicatesMerge.fieldSelect;</description>
-				<vbox id="zotero-duplicates-merge-item-box-container" flex="1">
-					<item-box id="zotero-duplicates-merge-item-box" flex="1"/>
-				</vbox>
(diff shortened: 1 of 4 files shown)


windingwind:
Now:

image

yexingsha:
Looks good, just need to fix the merge button menu so that it doesn't hug the screen edge.

Current menu placement:

Screenshot 2024-06-03 at 23 26 19

Compared to regular item info context menu:

Screenshot 2024-06-03 at 23 39 38

windingwind:
I think that's the toolbarbutton's default opening menu behavior on MacOS: the popup's top-left corner is placed at the bottom-left corner of the toolbarbutton.

image

If there is insufficient horizontal space for the menu, it will be shifted to the closest possible position.

image

yexingsha:
We can indeed follow the macOS button menu behavior, but it's always positioned 6px from the edge of the screen:

macOS toolbarbutton menu

windingwind:
I spent some time observing but still didn't find a way to control the menu popup's position when close to the screen edge. Only the context menus are positioned by the OS and have a gap between the screen edge, while the toolbarbutton's open menu popup behavior seems to be hardcoded in C++ (https://searchfox.org/mozilla-central/source/dom/xul/XULButtonElement.cpp#175) and we cannot mark the menupopup as native (context menu).

So maybe we can leave the menupopup position issue aside for now, if that's not a big problem?

windingwind:
OK, it's using the info button behavior now.

yexingsha:
Awesome!

Issue #3930, opened 2024-04-06
Accessibility regressions in JAWS?


dstillman:
https://forums.zotero.org/discussion/113442/accessibility-problems-in-zotero-7-beta

abaevbog:
> For instance, JAWS is not reading all the fields properly when I tab to them.

My observation is that JAWS struggles with reading the content of `editable-text` component when the cursor navigates it. It's best seen within the abstract section when pressing arrowUp/Down doesn't trigger any feedback from JAWS. When I tab through the `itemBox`, all text is selected, and that makes JAWS read text out for me. This is not the case with the title in the header though - it gets announced ... only sometimes.

> Using the up and down arrows to scroll through the items list is also problematic. Jaws doesn't consistently read out the items.

I can reproduce it too. To me, on arrowUp/Down it looks like sometimes it reads pieces from the `itemBox` (potentially, the last-focused component from there?), though expanding the row with right arrow makes it read the row out as it should.


A few other things to add to the list here (JAWS specific and not) :
- JAWS struggles with announcing the tags from the tag selector even though they are properly focused
- section headers in itemPane need aria labels so that they're announced properly. Same with the `+` and other buttons
- "Libraries and Collections" and "Related" section rows are not announced by JAWS
- Tab names are not always properly announced with JAWS when they're focused in the tab bar 
- The annotations from the "Annotations" section don't seem focusable at all, so JAWS and any other screen reader will skip over them. This might be a good thing to add into #3798

abaevbog:
Some observations about JAWS behavior:
- JAWS only partially works with our input fields. It will announce the selected text upon focus and it will echo characters as you type. However, if you try to navigate with arrows through the text inside of the input, it will not announce it. That is true for `editable-text` (multiline and not), as well as ordinary `<input>` and `<html:input>` fields. 
- The only exception where JAWS seems to behave as it should is, oddly enough, the input fields of the reader (e.g. search annotations), as well as the `contenteditable` area of the notes editor. So there is something that's being done there that is not being done anywhere else ... Manually adding an `<input>` component into the `<body>` of the note editor does not lead to JAWS announcing it properly. But adding it inside of the `editor-core` does. Figuring out what is different between `editor-core` of the note-editor vs input fields elsewhere seems to be the most promising direction now.
- Firefox/thunderbird do not have this issue

abaevbog:
[...]

PR #4205, opened 2024-06-05
improve JAWS in dialogs and non-main windows

abaevbog:
JAWS expects a specific executable (firefox.exe) to apply dialog-specific rules, which does not match in our case. It prevents JAWS from deciding how to handle our dialogs and, for example, it does not announce inputs correctly. (E.g. input field from new collection dialog or the inputs from quickFormat window)

As a workaround, we can add `dialog=no` as a parameter to all instances of opened dialogs or windows to force the HWND window type be MozillaWindowClass which JAWS handles better.

For more details: https://github.com/zotero/zotero/issues/3930#issuecomment-2146184884.

Addresses: #3930 

----

As far as I could see, this does not cause any breakage - but I'm still marking this as a draft for now.

@dstillman @adomasven - what do you think about this general approach? Is this something we could theoretically do?

### DIFF
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -2232,9 +2232,7 @@ var Zotero_QuickFormat = new function () {
 	 */
 	this.onClassicViewCommand = function(event) {
 		_updateCitationObject();
-		var newWindow = window.newWindow = Components.classes["@mozilla.org/embedcomp/window-watcher;1"]
-			.getService(Components.interfaces.nsIWindowWatcher)
-			.openWindow(null, 'chrome://zotero/content/integration/addCitationDialog.xhtml',
+		var newWindow = window.newWindow = Zotero.openWindow(null, 'chrome://zotero/content/integration/addCitationDialog.xhtml',
 			'', 'chrome,centerscreen,resizable', io);
 		newWindow.addEventListener("focus", function() {
 			newWindow.removeEventListener("focus", arguments.callee, true);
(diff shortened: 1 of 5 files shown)


dstillman:
This makes me nervous! Presumably `dialog=no` has to do something on some platform, no?

abaevbog:
Same here...

On one hand, I felt like this should be applied everywhere to be consistent and to make sure that this is needed for all dialogs is clear. On the other hand, I am not 100% sure to what degree this `dialog` parameter has effect besides the window type. I checked the most common dialogs and on mac and windows I see no difference. On linux, though, I do see a small change - e.g. I can drag the `newCollection` dialog as an independent window around with this PR but on `main` the `newCollection` is a modal that will make the rest of the window move with it. 

Since this is meant to help JAWS, it would probably make sense to keep this change specific to windows without affecting mac and linux. Adding a conditional, though, to potentially every single instance of `window.openDialog` or `Services.ww.openWindow` seems not right and also likely to be missed in the future. 

What if I add a separate function somewhere to open up a dialog or window same way as now, except it will add `dialog=no` parameter only when `Zotero.isWin` is true? And as a start, we could use it on just a few dialogs (maybe quickFormat and new collection, since those are marked by VPAT folks) to see how it goes. That way, there would also be one place to document why that `dialog=no` is needed.
Does this sound better? If so, is `standalone.js` a good place to for it? It already has [openScaffold](https://github.com/zotero/zotero/blob/4d7c641f7b96777089524453373852bcf9b5ed6b/chrome/content/zotero/standalone/standalone.js#L995) which seems somewhat similar.

PR #4213, opened 2024-06-06
Add sidenav#showByDefault option to ItemPaneManager#register

windingwind:


windingwind:
Make-it-red example updated: https://github.com/zotero/make-it-red/pull/9

chrome/content/zotero/xpcom/itemPaneManager.js
```diff
@@ -69,7 +69,9 @@
  * @property {string} paneID - Unique pane ID
  * @property {string} pluginID - Set plugin ID to auto remove section when plugin is disabled/removed
  * @property {SectionL10n & SectionIcon} header - Header options. Icon should be 16*16 and `label` need to be localized
- * @property {SectionL10n & SectionIcon} sidenav - Sidenav options. Icon should be 20*20 and `tooltiptext` need to be localized
+ * @property {SectionL10n & SectionIcon & { showByDefault?: boolean }} sidenav
+ * Sidenav options. Icon should be 20*20 and `tooltiptext` need to be localized
+ * `showByDefault` indicates the button should always show in sidenav's default status, e.g. no/multiple item selected.
```
dstillman, line 74: Can you specify what happens if this isn't specified or is set to false?

windingwind: Thanks! The doc comment is updated.

dstillman: So that means when set to false the button will appear and disappear as you select/deselect items? Is that what we want, as opposed to just becoming disabled? (Sorry, I think I missed the initial discussion where we laid out why this was necessary.)

dstillman: > I guess it's yexing's idea (I can't remember too) to provide an option to plugins whose custom section is always enabled (e.g. BBT) to show their icon even when no item is selected, to avoid frequent appear/disappear.

Ah, right — that was it (CC @yexingsha). Hmm, but is this the right default, then? Wouldn't most plugins that have a sidenav icon at all show it for all regular items? And if that's the case, I think we'd want them to have this enabled. So then maybe it should be the default behavior, just like it is for our own icons that appear for regular items, and we need a flag that would cause it not to show in the default set.

yexingsha: If we default to showing all the sidenav icons that appear for regular items, BBT is still different because it needs to be enabled when multiple or no items is selected, whereas our own icons and most plugin icons would be disabled. So perhaps we need another flag for that?

dstillman: Why does it need to be enabled? Isn't it used for showing the citation key (before https://github.com/zotero/zotero/pull/4232)? (I haven't actually tried loading it.)

dstillman: Is `enabled`/`disabled` not already handled by the existing custom-section API?

Issue #4206, opened 2024-06-05
Unpredictable Shift-click behavior in Quick Format


dstillman:
Minor follow-up to #3851 — if I move selection with the keyboard and then Shift-click on another item, it selects the range I expect, but if I do some Cmd-clicks first, it sometimes gets confused:

https://github.com/zotero/zotero/assets/783082/148415e0-018f-4abe-8490-15199efb1371

PR #4217, opened 2024-06-07
qf: improved multiselect range selection

abaevbog:
- `richlistbox` shift-click selects a range starting with [_selectionStart](https://searchfox.org/mozilla-central/source/toolkit/content/widgets/richlistbox.js#469). After a few cmd/ctrl clicks, `_selectionStart` ends up not being correct - it can even point at an unselected item. To help `richlistbox` properly select a range, before each click is processed, reset `_selectionStart` to be either the first or last selected item, depending on where shift-click landed.
- also, fixed a small glitch where inputs would be created if you click on the red border outside of the actual editor.

Fixes: #4206

### DIFF
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -162,6 +162,21 @@ var Zotero_QuickFormat = new function () {
 				if (e.shiftKey && referenceBox.selectedCount < 1) {
 					_selectFirstReference();
 				}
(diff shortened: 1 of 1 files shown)


abaevbog:
https://github.com/zotero/zotero/assets/36271954/4893ed99-e23f-49b9-91cf-96745e578c19

dstillman:
Looks good!

PR #4225, opened 2024-06-10
Improve Advanced Search Dialog

### LINKED ISSUE #4190: Fix Advanced Search dialog sizing issues
Screenshot 2024-05-31 at 3 49 53 AM

It should be wider, but also, the collection drop-down doesn't have a minimum width, so even at wider window sizes it's very small:

Screenshot 2024-05-31 at 3 53 12 AM

Zotero 6, for comparison:

Screenshot 2024-05-31 at 3 52 12 AM

tnajdek:
- Set the window minimum size to 700px x 500px.
- Allow the library selector to take up all the available width of the window.
- Fix a regression where non-applicable operators remain visible.
- Fix the "✔️" appearing next to multiple operators in the operator popup menu.
- Include the "Year" and "Attachments" columns by default in the results table.

Resolve #4190

### DIFF
--- a/chrome/content/zotero/advancedSearch.js
+++ b/chrome/content/zotero/advancedSearch.js
@@ -58,7 +58,7 @@ var ZoteroAdvancedSearch = new function() {
 		var elem = document.getElementById('zotero-items-tree');
 		const columns = COLUMNS.map((column) => {
 			column = Object.assign({}, column);
-			column.hidden = !['title', 'firstCreator'].includes(column.dataKey);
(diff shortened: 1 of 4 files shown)


dstillman:
Merged, except for the `flex: 1` on the library menu, which I think looks weird in the default case where it's just "My Library". It's normal for drop-downs to expand depending on the selected value.

I also fixed the width of the condition menu (I said "collection menu" by mistake — sorry) and tweaked the widths of the join-mode and operator menus.

dstillman:
Also, the Attachments menu isn't showing up for me after this, even if I clear prefs and xulstore.json, though I can add it manually. Not sure what's going on there.

Issue #4230, opened 2024-06-11
Tagsbox: tab from new empty tag input looses focus


abaevbog:
Pressing tab from `isNew` last empty tag input will have the empty input removed and loose focus. 

It's especially an issue when dealing with the tags popup opened from the sidebar of the reader using keyboard as it makes the popup get stuck.

Issue #4226, opened 2024-06-11
Can't close annotation "Add tags…" popup with Esc


dstillman:
1. Open sidebar
2. Click "Add tags…" in annotation widget
3. Press Esc once, which closes the input field but leaves the popup open
4. Press Esc again, which deselects the annotation widget while leaving the popup open

I'm not sure if we should close the popup on the first Esc, but certainly the second one should close it. And, actually, it's a problem even when not editing: if an annotation has tags and you click on them to open the popup to view them without an active input field, Esc doesn't close the popup.

dstillman:
In other words, it doesn't close it from this mode, which is where you end up with both scenarios above:

![Screenshot 2024-06-10 at 11 16 29 PM](https://github.com/zotero/zotero/assets/783082/80326776-de5a-48ba-b844-259022bf79d0)

abaevbog:
This was brought up by the vpat folks, so I had the tags popup behavior as a part of #4222. I have a few tweaks to address this among other things related to the tags popup in #4231

Issue #4222, opened 2024-06-10
VPAT Reader: focus


abaevbog:
- [x] vpat 61: Magnifying class of the "Search annotations" input should not be focusable
    - It could be focused with voiceover on arrow left click from the input.
 - [x] vpat 62: Clearing sidebar "Search annotation" resets focus
    - The review indicates "When this search is cleared with the button labeled 'x,' focus is moved to the beginning of the PDF viewer. This occurs with or without a screen reader active." Does not happen consistently for me but is sometimes reproducible on mac
 - [X] vpat 63: Sidebar unable to move focus forward if no results for search. Fixed?
 - [x] vpat 66 and 68: focus does not enter the tag selector popup for annotations
     - Once the tags popup is opened, arrows or tab keypresses will keep navigating the sidebar instead of the tags selector
 - [x] vpat 67: even if focus is places inside of the tag selector popup, Escape does not close it
     - Escape from tag selector popup should close it
  - [x] Something I noticed myself: tabbing through an annotation will move focus from the context menu button to the comment to the tags button skipping the actual text of the annotation unless it has been focused by selecting "Edit Annotation Text" in the context menu. It seems a bit inconsistent, perhaps annotation text should have a tab stop as well? Or maybe we announce the text of the annotation when the annotation box is focused? Otherwise, the only way to have annotation text announced via screen reader seems to be via "Edit Annotation Text" which is not very intuitive and has to be discovered.

mrtcode:
> Something I noticed myself: tabbing through an annotation will move focus from the context menu button to the comment to the tags button skipping the actual text of the annotation unless it has been focused by selecting "Edit Annotation Text" in the context menu. It seems a bit inconsistent, perhaps annotation text should have a tab stop as well? Or maybe we announce the text of the annotation when the annotation box is focused? Otherwise, the only way to have annotation text announced via screen reader seems to be via "Edit Annotation Text" which is not very intuitive and has to be discovered.

Yes, I guess we can add a tab stop.

And the other issues seem to be fixed already, right?

abaevbog:
> Yes, I guess we can add a tab stop.

[...]

PR #4231, opened 2024-06-12
vpat 66-68: fix focus within tagsBox popup

abaevbog:
- focus will always enter the tagsBox popup opened from the reader
- escape from within the popup with tagsBox will close the popup (vpat 67)
- one should not be able to collapse the tagsBox. Added noCollapsible getter and setters to the collapsible panel to prevent the section from changing it's open status. This may be also used in other cases, such as to prevent itemBox from being collapsed in duplicates mode.
- patched a glitch where tab from the last empty tab input would loose focus.
- changed the `menupopup` for `panel` to display the `tagsbox` because `menupopup` has implicit `role="menu"` which is not meant to contain inputs, so voiceover completely looses cursor when inputs are focused inside of the popup. Also, tweaked spacing a bit to avoid the focus-ring getting cutoff.

Addresses: #4222
Fixes: #4230
Fixes: #4226
Addresses: #4388

### DIFF
--- a/chrome/content/zotero/elements/collapsibleSection.js
+++ b/chrome/content/zotero/elements/collapsibleSection.js
@@ -43,7 +43,7 @@
 		set open(newOpen) {
 			newOpen = !!newOpen;
 			let oldOpen = this.open;
-			if (oldOpen === newOpen || this.empty || this.notCollapsible) return;
+			if (oldOpen === newOpen || this.empty || !this.collapsible) return;
 			this.render();
 			
 			// Force open before getting scrollHeight, so we get the right value
@@ -106,16 +106,16 @@
 			this.setAttribute('summary', val);
 		}
 		
-		get notCollapsible() {
-			return !!this.getAttribute("no-collapse");
+		get collapsible() {
+			return !this.getAttribute("no-collapse");
 		}
 
-		set notCollapsible(val) {
+		set collapsible(val) {
 			if (val) {
-				this.setAttribute('no-collapse', val);
+				this.removeAttribute('no-collapse');
 			}
 			else {
-				this.removeAttribute('no-collapse');
+				this.setAttribute('no-collapse', val);
 			}
 		}
 		
--- a/chrome/content/zotero/elements/collapsibleSection.js
+++ b/chrome/content/zotero/elements/collapsibleSection.js
@@ -43,7 +43,7 @@
 		set open(newOpen) {
 			newOpen = !!newOpen;
 			let oldOpen = this.open;
-			if (oldOpen === newOpen || this.empty) return;
+			if (oldOpen === newOpen || this.empty || this.notCollapsible) return;
 			this.render();
 			
 			// Force open before getting scrollHeight, so we get the right value
@@ -104,6 +104,19 @@
 		
 		set summary(val) {
 			this.setAttribute('summary', val);
+		}
+		
+		get notCollapsible() {
+			return !!this.getAttribute("no-collapse");
+		}
+
+		set notCollapsible(val) {
+			if (val) {
+				this.setAttribute('no-collapse', val);
+			}
+			else {
+				this.removeAttribute('no-collapse');
+			}
 		}
 		
 		static get observedAttributes() {
--- a/chrome/locale/en-US/zotero/zotero.ftl
+++ b/chrome/locale/en-US/zotero/zotero.ftl
@@ -11,6 +11,7 @@ general-remind-me-later = Remind Me Later
 general-choose-file = Choose File…
 general-open-settings = Open Settings
 general-help = Help
+general-tag = Tag
 
 menu-file-show-in-finder =
     .label = Show in Finder
@@ -517,7 +518,7 @@ abstract-field =
     .placeholder = Add abstract…
 
 tag-field =
-    .aria-label = Tag
+    .aria-label = { general-tag }
 
 tagselector-search =
     .placeholder = Filter Tags
--- a/chrome/locale/en-US/zotero/zotero.ftl
+++ b/chrome/locale/en-US/zotero/zotero.ftl
@@ -517,7 +517,7 @@ abstract-field =
     .placeholder = Add abstract…
 
 tag-field =
-    .placeholder = Edit tag...
+    .aria-label = Tag
 
 tagselector-search =
     .placeholder = Filter Tags
--- a/chrome/locale/en-US/zotero/zotero.ftl
+++ b/chrome/locale/en-US/zotero/zotero.ftl
@@ -516,6 +516,9 @@ expand-all-sections =
 abstract-field =
     .placeholder = Add abstract…
 
+tag-field =
+    .placeholder = Edit tag...
+
 tagselector-search =
     .placeholder = Filter Tags
 
--- a/chrome/content/zotero/elements/itemPaneSection.js
+++ b/chrome/content/zotero/elements/itemPaneSection.js
@@ -85,12 +85,12 @@ class ItemPaneSectionElementBase extends XULElementBase {
 		}
 	}
 
-	get notCollapsible() {
-		return this._section.notCollapsible;
+	get collapsible() {
+		return this._section.collapsible;
 	}
 
-	set notCollapsible(val) {
-		this._section.notCollapsible = !!val;
+	set collapsible(val) {
+		this._section.collapsible = !!val;
 	}
 	
 	connectedCallback() {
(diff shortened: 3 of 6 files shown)


dstillman (commented):
This fixes the Esc thing nicely, but I think the keyboard handling in the popup is a little weird. If you open the panel on an annotation with a tag, it focuses the first existing tag. I think it would make more sense to leave the focus on either the header or the + button. + button would be much more useful, so you could quickly add more tags via the keyboard. Maybe a little weird to start with that, but worth it? (Though, why we do allow focus on the header in the context of the reader in the first place? Just so that the count can be reader by screen readers? Can we leave that to screen reader navigation and have Tab move to +?)

I also found a weird bug where, if you click in the last existing tag and press Shift-Enter, it visually adds a newline to the last tag and, in the context of the library view, then returns focus to the items list. In the reader, it just adds the newline to the textbox and keeps focus there. We have some behavior to handle pasting in a multiline set of tags, but for manual entry there's not much reason to allow Shift-Enter. (It skips the save process after each one, but hopefully that's fast enough that it doesn't matter.) At the very least, we shouldn't return focus to the items list in the library view, but probably it should just act the same as Enter.

chrome/content/zotero/elements/collapsibleSection.js
```diff
@@ -106,6 +106,19 @@
 			this.setAttribute('summary', val);
 		}
 		
+		get notCollapsible() {
```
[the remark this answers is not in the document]
abaevbog: Yes, exactly. My original thought was that "not collapsible" is a special case, so it should have special naming. It can end up being a bit confusing, especially if we wanted to check if something is collapsible because then it had to be a double-negative.

So I changed it to be `collapsible` instead, that's definitely better

abaevbog:
> I think it would make more sense to leave the focus on either the header or the + button. + button would be much more useful, so you could quickly add more tags via the keyboard.

That is a valid point. I focused the first tag initially to somewhat mimic what happens if there are no tags. But after going over it a few times, you are right, it doesn't make much sense.

I think the section header should remain focusable and is the best candidate for the initial focus. Mainly because it does serve as a kind of top-level header informing the user that this is a tags section they are not looking at. The issue with focusing the "Add" button is that, per vpat folks, it is best to stick to focusing the first element in a modal or dialog as it ensures that nothing "behind" it will be missed. And since the add button by itself doesn't provide enough context as to what it or the modal is, it might be clearer for screen reader users to have the header focused first. Initial focus happens without the focus-ring so for those not using keyboard navigation, everything should look the same.

If there are no tags, a new empty tag input field is created and focused when the popup opens. In that case, it just announces that an editable textfield is focused and pretty much that's it. I thought maybe we can also add a placeholder to tags' `editable-text` component to ensure that when tags are focused it is clearer what exactly that field is? This is something I noticed some time ago and maybe it's a good time now to add it?

> I also found a weird bug where, if you click in the last existing tag and press Shift-Enter, it visually adds a newline to the last tag and, in the context of the library view, then returns focus to the items list.

Thank you for pointing that out! For me, this shift-enter behavior was very inconsistent as, I believe, it was a race condition where different listeners fought over who gets to handle the event first. So I removed the ability to make tag multiline on shift-Enter - it now does the same thing as Enter by saving the edits and moving focus to itemTree or reader content. In case if multiple tags are pasted, the field becomes multiline, and then I think it's ok to just let default `editable-text` multiline handling take over with Enter adding a new line (if some tags are being adjusted after pasting) and shift-Enter accepting and saving all the pasted tags. What do you think?

dstillman:
> The issue with focusing the "Add" button is that, per vpat folks, it is best to stick to focusing the first element in a modal or dialog as it ensures that nothing "behind" it will be missed.

Well, that's why I ask if we can leave header focusing to screen reader navigation. I'm still not clear on how screen reader navigation and keyboard navigation interact for most screen reader users. Because it's just a pointless extra tab stop for non-screen-reader users, right?

abaevbog:
> Well, that's why I ask if we can leave header focusing to screen reader navigation.

I am a bit confused by what is meant by "screen reader navigation" ... My understanding is that screen readers have a cursor (e.g. the black box that voiceover shows), which reads the node that its on. Depending on the mode/component a user might move the cursor to an element that is not focused. But as a general rule, when focus shifts, the cursor follows it, so the first thing that we focus in the popup is the first thing that gets announced. And then if we remove the tabstop from the header (even if it's still focusable, like with `tabindex="-1"`), it makes it less likely that one will run into it.

Granted, I may be really overthinking this and the section header is not that important. My main concern was to make it clear that one is in a new popup where they can can edit tags and some screen readers do a better job than others at conveying this. In other `panel`s, the first thing that is focused is a descriptive input field that makes it clear where one is (e.g. the "Search tabs" filter in the tabs menu) but here, if the first thing that is focused is "Add" button, it's just not as straightforward. 

However, if the placeholders are added to tags, I think there is way more context as to where one is at. The first time the popup opens, one might tab through the fields a few times, head names of tags and "Edit tag ..." message, from where it will be clear that the tags are being looked at. So perhaps, the tabstop on the header is really not needed

abaevbog:
Now there is no tabstop on the section header. If there are no tags, the first focused element in an empty tag input which says "Edit tags...", so that's very clear. And then if there are tags, the "Add" button gets focused and tabbing skips the header. But, tags' input descriptions indicate that they are tags and screen readers can still see the header. E.g. voiceover Cmd-Option-Shift-arrowUp (with default setting) moves cursor to the header to announce it.

chrome/locale/en-US/zotero/zotero.ftl
```diff
@@ -516,6 +516,9 @@ expand-all-sections =
 abstract-field =
     .placeholder = Add abstract…
 
+tag-field =
+    .placeholder = Edit tag...
```
dstillman, line 520: This can just be "Tag".

dstillman: Though it's kind of distracting visually even with just "Tag", since it appears for every single new row. Can we just use an ARIA label here?

Screenshot 2024-07-16 at 1 42 04 AM

abaevbog: For sure! There was a small part of me that always though that maybe some kind of placeholder would be good to not have an empty input field but `aria-label` (or maybe `aria-description` - let me try which is better) would also work

dstillman:
Not totally related, but it's a little annoying that you can't add a new tag while within an existing tag field without tabbing to + and pressing Return. Can we have Shift-Return from any field jump to an new row at the bottom? Similar to what we do for creators (except there Shift-Return only works from the last one, but probably that should create a new empty creator row after the current one, particularly now that we enable the + button for each row).

I also wonder if tabbing off the end should create a new empty row, which would then disappear on blur if you kept tabbing? That would be more discoverable than Shift-Return.

chrome/locale/en-US/zotero/zotero.ftl
dstillman: I moved it to a general string, so we can always change this later.

abaevbog:
> Not totally related, but it's a little annoying that you can't add a new tag while within an existing tag field without tabbing to + and pressing Return. Can we have Shift-Return from any field jump to a new row at the bottom?

That is a great point! It certainly is a bit annoying that you have to keep tabbing, so a shift-Enter would be a nice addition, especially given that we already use that shortcut. 

> Similar to what we do for creators, except there Shift-Return only works from the last one

I see what you mean. I guess it doesn't make sense to focus the next creator row on shift-Enter (which happens if the creator was not changed. If it was, shift-Enter from any creator row adds and focused an empty creator row at the end), since to navigate through fields of the itemBox you can just tab. Shift-Enter as a shortcut to always add an empty creator after the current creator row would 100% be more useful.

> I also wonder if tabbing off the last tag should create a new empty row, which would then disappear on blur if you kept tabbing? Why bother tabbing to + when it could just create a new row automatically? That would also be more discoverable than Shift-Return.

That is true. My only reservation is that it might introduce some unnecessary confusion to those who rely on screen readers for navigation. If you shift-tab from the "Add" button twice, you go through "-" button and then land on the last tag. Then, if you tab you wouldn't go back to the "-" button where you just were but you would be in another input field. My understanding is that any kind of changes of state that are not explicitly requested by the user should be used with caution.

dstillman:
Squashed a bit and merged

abaevbog:
Great! I'll make separate issues to update shift-Enter behaviors and such

Issue #4114, opened 2024-05-14
Rename `PaneHeader` to `ItemPaneHeader`?


dstillman:
Any reason this shouldn't be `ItemPaneHeader`, like `ItemPaneSection(ElementBase)` and `ItemPaneSidenav`?

windingwind:
Makes sense. Renamed in #4031

windingwind:
Mentioning this, maybe we should consider renaming some other CEs, e.g. `ItemBox` to `InfoBox`/`ItemInfoBox`. The meaning of `ItemBox` is quite vague.

dstillman:
Yeah, that's just historical from `itembox.xml`, but I was thinking the same. `InfoBox` is probably clear enough?

windingwind:
InfoBox is clear to me. We can wait for #4096 to be merged and then rename the ItemBox in a new PR.

PR #4232, opened 2024-06-12
Add item pane info section custom row api

windingwind:
resolve: #4175

## Changes
1. item-box is renamed to info-box, as discussed in #4114. This is necessary since we will expose the right API name to plugin developers. The file name is kept `itemBox.js` for now to avoid conflicts. We can rename the file later (before releasing Z7?)
2. The plugin API is unified to a base class in `chrome/zotero/modules/pluginAPIBase.mjs`. I feel like it's not necessary to attach this class to the Zotero instance, so I decided to place it in `modules` and import it when needed. Let me know if that's not proper.
3. The existing plugin APIs (item tree column and item pane section) are updated with the base class. Particularly, the item tree column API has changed from `Zotero.ItemTreeManager.registerColumns` to `Zotero.ItemTreeManager.registerColumn`, where I regret to support passing an array of options to the API, which is inconsistent with other APIs we added/will add. The old `registerColumns` is still available, but marked as deprecated and should be removed before the official Zotero 7 release. I can mention this change to the dev list once this PR is ready to merge.
4. The new item pane info section custom row API is ready for review. The rows can appear at the top of the info section, after the creator rows, and at the bottom. We may want to support a custom render hook to create any element in the row value grid, but I think this is enough as an initial PR.
5. I added `aria-labelledby`, not sure if that is enough. I would appreciate it if @abaevbog could give comments on any accessibility issues with the new API.

## Example
a minimal example:
```js
minimalExampleID = Zotero.ItemPaneManager.registerInfoRow({
	rowID: "test1",
	pluginID: "zotero@zotero.org",
	label: {
		"l10nID": "general-print",
	},
	editable: false,
	onGetData: ({item}) => {
		console.log(`Minimal example: Getting data for ${item.id}`);
		return `Minimal example: ${item.id}`;
	},
	onSetData: ({value, item}) => {
		console.log(`Minimal example: Saving ${value} to ${item.id}`);
	}
});
```

a full example:
```js
fullExampleID = Zotero.ItemPaneManager.registerInfoRow({
	rowID: "test2",
	pluginID: "zotero@zotero.org",
	label: {
		"l10nID": "general-print",
	},
	position: "afterCreators",
	multiline: true,
	nowrap: true,
	onGetData: ({item}) => {
		console.log(`Full example: Getting data for ${item.id}`);
		return `${item.id}`;
	},
	onSetData: ({value, item}) => {
		console.log(`Full example: Saving ${value} to ${item.id}`);
	},
	onItemChange: ({item, tabType, setEnabled, setEditable}) => {
		let editable = item.itemType === "book";
		let enabled = tabType === "library";
		console.log(`Full example: Item ${item.id} row is enabled: ${enabled}, is editable: ${editable}`);
		setEnabled(enabled);
		setEditable(editable);
	}
});
```

unregister:
```js
Zotero.ItemPaneManager.unregisterInfoRow(minimalExampleID);
Zotero.ItemPaneManager.unregisterInfoRow(fullExampleID);
```


abaevbog:
The accessibility side looks good to me - the rows labels are announced when the input is focused as expected which is great!

I am noticing something buggy with the how the focus is preserved between itemBox re-renderings. If an editable custom row is added, editing an ordinary field from the table and then clicking into the custom row will loose focus and somehow focus will keep getting lost when you try to click around the table after that. 

See the expected behavior before a custom row is added and lost focus after:

https://github.com/zotero/zotero/assets/36271954/03b91616-1ac9-4052-897b-d4fbed1a42ed

Something that is likely related is that when an `editable-text` from a custom row is added, it does not have an `id` like [ordinary rows](https://github.com/windingwind/zotero/blob/19d0ae98ad319e8568ef1166c86112dc2d1fea42/chrome/content/zotero/elements/itemBox.js#L1688) from `itemBox`. We want every focusable element to have an id because when we restore focus after a table refreshed, [we use the id](https://github.com/windingwind/zotero/blob/19d0ae98ad319e8568ef1166c86112dc2d1fea42/chrome/content/zotero/elements/itemBox.js#L2489) to find which node to refocus.

windingwind:
Thanks! I think the selection restore is fixed now.

dstillman (commented):
Still looking at this, but a few quick comments

windingwind:
Changes after the last review:
1. Add tests (`pluginAPITest.js`) for plugin APIs (currently only the info box custom row API). All parameters and hooks of the `option` of the API are tested.
2. Fixes according to the comments above.

chrome/content/zotero/elements/itemBox.js
```diff
@@ -866,13 +882,255 @@
...
+
+				rowElem.append(labelElem, dataElem);
+
+				this.insertCustomRow(rowElem, position);
+
+				this.updateCustomRowProperty(rowElem);
+
+				this.updateCustomRowData(rowElem);
+			}
+		}
```
dstillman, line 1018: This is duplicating a bunch of existing code, in a way that I feel is almost certain to get out of sync. Is it possible to do this such that the row is still largely rendered by `createValueElement`/`showEditor`, but with appropriate hooks or data stores for the custom parts?

chrome/content/zotero/elements/itemBox.js
```diff
@@ -866,13 +882,255 @@
...
+
+			// Attempt to make bidi things work automatically:
+			// If we have text to work off of, let the layout engine try to guess the text direction
+			if (data) {
+				valueElem.dir = 'auto';
+			}
+			// If not, assume it follows the locale's direction
+			else {
+				valueElem.dir = Zotero.dir;
+			}
```
dstillman, line 1132: More duplicated code. I'd really like to try to avoid this if possible. It seems like all rows should be rendered in only one place, and if it's a custom row the existing functions should just get the necessary attributes and values.

dstillman (changes requested):

chrome/content/zotero/elements/itemBox.js
windingwind: Agree. Back to the time I implemented this PR, I initially made some big refactors to the itemBox.js to unify all row-rendering calls, which have been a bit messy for a long time. However, these changes frequently conflict with commits in the main branch and make implementing the API based on those changes impossible. So I decided to duplicate the code a bit temporarily. Since the info box is no longer having big changes now, I think it's time to do this refactor again!

windingwind:
The create element (label/value) part in itemBox.js is now refactored and centralized.

windingwind:
rebased

windingwind:
It there's not other blockers, I think we should try to merge this, so that we can build more APIs based on the abstract basic class added in this PR.

chrome/content/zotero/modules/pluginAPIBase.mjs
```diff
@@ -0,0 +1,383 @@
...
+ *
+ * @typedef {object} PluginAPIOptDefConfig
+ * @property {PluginAPIOptDefType} [type] - The type of the value
+ * @property {boolean} [optional] - Whether the value is optional, default false
+ * @property {PluginAPIOptDefCheckHook} [checkHook]
+ *  - A function to check the value, returning true if it is valid
+ * @property {PluginAPIOptDefConfig} [children]
+ *  - The type definition of the children of the object. The children can be
+ * an object or an array of objects with the same structure
+ * @property {any} [fallbackReturn] - The value to return if the check hook fails
```
windingwind, line 54: Rebased, and updated the code to allow APIs to easily configure a fallback return value for functions as hooks when they throw errors.

The API will wrap the function, catch any possible unhandled errors, log them, and use the `fallbackReturn`. It does not promise the value to be correct, it just promises the call does not throw an error.

test/tests/pluginAPITest.js
```diff
@@ -0,0 +1,522 @@
...
+		beforeEach(async function () {
+			resetCaches();
+		});
+
+		afterEach(function () {
+			Zotero_Tabs.select("zotero-pane");
+			Zotero_Tabs.closeAll();
+		});
+
+		it("should render custom column and call dataProvider hook", async function () {
```
windingwind, line 429: A set of tests for the item tree API are added

windingwind (commented):
Made some changes since the last update as marked above. It's mainly for the auto error handling of hooks in plugin APIs.

dstillman:
@windingwind: Leaving out the dataProvider try/catch now is correct, right?

dstillman:
Also the `should set input nowrap` test is now failing. Did I do that?

windingwind:
Thanks. I will check the tests

windingwind:
Seems like we have applied `wrap` to all fields in #4802, which is why the `noWrap` parameter does not work and fails the test. Can just remove the support to `noWrap`.

dstillman:
OK, let's do it. @windingwind, can you squash these a bit? Could just do 1st commit, 2nd commit, and then all the rest together, but if there's anything else to separate to make the big diff a bit more readable, we could do that.

windingwind:
OK, squashed into 3 commits, should be fine to merge.

Issue #4211, opened 2024-06-06
Use default attachment titles for child attachments


dstillman:
Currently, if you drag a PDF onto an existing item, the filename gets set based on the parent item metadata (if it's the first attachment file) and the attachment title gets set to that filename. The same thing happens if you drag in a PDF and it's recognized.

These are inconsistent with what we do (out of necessity, since there's often no filename, or it's gibberish) when saving from the web, where the attachment title gets set by the translator to something like "ScienceDirect Full Text [sic] PDF" (which will be [localized](https://github.com/zotero/zotero/issues/3078) in the future). It also means that, if the file is again renamed via parent metadata (which will happen [automatically](https://github.com/zotero/zotero/pull/3860) soon), we change the attachment title, whereas Zotero 7 [no longer changes the title if it doesn't match the filename](https://forums.zotero.org/discussion/comment/439553/#Comment_439553).

This has confused many, many beta users.

For consistency, clarity, and cleanliness, we should set the title to just "PDF" when adding a first PDF to an existing item or when we successfully retrieve metadata and create a parent item. Same for "EPUB". "Image" for known image types? Not sure about other file types, like Word documents, since even if those are the "primary" attachment they're less likely to be a representation of the parent item, and the filename might be the best description of what they are.

What happens if you create a parent item either by identifier or manually? By identifier seems like it should work the same as retrieving metadata. For manual entry, the parent item is already set to the same title as the attachment title and filename, so probably reasonable to change the attachment title to "PDF" then too.

It will also be possible to [change the primary attachment for an item](https://github.com/zotero/zotero/pull/3333). Does that change anything?

Secondary files I guess need to keep their filenames as titles? But then if you rename based on parent metadata, the title would change. Maybe we should just never change the title to match the filename? If we're taking care of the common cases here, then for other situations like secondary files, maybe you just set the attachment title manually?

A lot to think about!

(Possible this will need to be done on top of https://github.com/zotero/zotero/pull/3860 to avoid conflicts.)

dstillman:
(We can also do some of the low-hanging fruit here before resolving all these questions.)

dstillman:
> Do you mean that we'd just remove the current attachment box behavior where we keep the title up to date with the filename when you rename the file? Or are there other cases in which this happens?

[...]

PR #4237, opened 2024-06-14
Set automatic title on first child attachment of each type

AbeJellinek:
And tweak `addAttachmentFromDialog()`:

- Accept a `files` argument for tests (`FilePicker#show()` can't be mocked)
- Return the added attachments

`setFirstAttachmentTitle()` is a kind of awkward method, but it encapsulates logic that we need in a few different places. The alternative would be to expose `_getDefaultTitleForAttachmentContentType()`, pass whatever it returns into `setField()`, and only run the `isFirstOfType` check in places where there could actually be other attachments (manually adding attachments to existing items, basically). Not sure that's that much better.

Needs #3860 (though I think any conflicts with that would be pretty minor; current conflict with `main` is from there). Addresses #4211 for Create Parent Item, Retrieve Metadata, and manually added attachments.

### DIFF
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -2496,6 +2496,13 @@ var ItemTree = class ItemTree extends LibraryTree {
 					}
 
 					if (item) {
+						try {
+							item.setFirstAttachmentTitle();
+							await item.saveTx();
+						}
+						catch (e) {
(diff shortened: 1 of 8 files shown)


dstillman:
I removed the https://github.com/zotero/zotero/pull/3860 commits, since I'm not sure yet if getting that in before 7.0.0 is realistic.

PR #4239, opened 2024-06-14
new aria strings for the reader vpat update

abaevbog:
Aria labels and descriptions used by
https://github.com/zotero/reader/commit/e0e595a09eb5c1ceb53fd6814cf2cdc91c4429ac to add a number of accessibility improvements to the reader.


chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -1380,6 +1380,12 @@ pdfReader.searchAnnotations = Search Annotations
 pdfReader.noAnnotations = Create an annotation to see it in the sidebar
 pdfReader.noExtractedText = No extracted text
 pdfReader.addComment = Add comment
+pdfReader.annotationComment = Annotation comment
+pdfReader.annotationText = Annotation text
+pdfReader.manageTags = Click to manage tags
```
dstillman, line 1385: It shouldn't be "click", presumably, if we're writing this for screen readers. "Manage tags"? "Manage tags for this annotation"?

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -1380,6 +1380,12 @@ pdfReader.searchAnnotations = Search Annotations
...
+pdfReader.openMenu = Open menu
```
dstillman, line 1386: "Open annotation menu"?

abaevbog:
These are both great points! I re-pushed with "Manage tags for this annotation" and "Open annotation menu"

Issue #4240, opened 2024-06-14
Item box: Merge version select menu button remains visible after exiting merge


AbeJellinek:
https://forums.zotero.org/discussion/115199/zotero-7-beta-duplicates-choice-icon-remain-visible-after-exiting-merging-items

@abaevbog: I'm guessing this has to do with the style overrides we apply to keep menu buttons visible while their menus are open?

abaevbog:
Hm, that is really quite interesting because the merge button will "shine through" all the other panes of the deck, even if the notes are opened.

<img width="530" alt="Screenshot 2024-06-14 at 2 15 11 PM" src="https://github.com/zotero/zotero/assets/36271954/658764a8-21b9-420c-8844-232a4173071c">

But yes, it does come from my [hack](https://github.com/zotero/zotero/blob/d2fda98bcba3f0cdfc9591117c2e565a532cb9a2/chrome/content/zotero/elements/itemBox.js#L2276) to make sure the button remains visible after `blur` if its menu is opened. I'll try to work out a better way to handle this!

PR #4243, opened 2024-06-14
Fix itembox merge button remaining visible

abaevbog:
Even after the merge mode is no longer active.

Fixes: #4240

---

The issue was arising from `document.popupNode` not being set when the popup was being opened in `handlePopupOpening`, so `popuphidden` listener wasn't clearing `display` and `visibility` style properties. I thought it might be best to not use the inline css at all and just remove/add classes in a more general `popuphidden ` listener that doesn't require  `document.popupNode` to work.

### DIFF
--- a/chrome/content/zotero/elements/itemBox.js
+++ b/chrome/content/zotero/elements/itemBox.js
@@ -2269,10 +2269,12 @@
 			event.preventDefault();
 			event.stopPropagation();
 			
+			let target = event.target;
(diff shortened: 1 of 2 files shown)


abaevbog:
Oh, right, thank you for pointing that out! We should only set `show-without-hover` if `show-on-hover` is set. 

I pushed a fix to that with a couple additional small fixes to address the merge button not properly opening the menu on space/enter.

dstillman:
Merged

PR #4250, opened 2024-06-17
Fix duplicate attachment row from attachments-box `notify()`

### LINKED ISSUE #4249: Attachments box: Newly added attachment appears twice
https://forums.zotero.org/discussion/115289/z7-beta-dragging-dropping-a-pdf-onto-entry-shows-it-as-duplicate-in-sidebar

AbeJellinek:
`refresh` event comes before `add` when an attachment is added to an item (not sure why), so we end up with a duplicate row. Fix by always removing any existing row before adding a new one.

Fixes #4249

### DIFF
--- a/chrome/content/zotero/elements/attachmentsBox.js
+++ b/chrome/content/zotero/elements/attachmentsBox.js
@@ -112,14 +112,13 @@
 			this._updateAttachmentIDs().then(() => {
 				this.updatePreview();
 
-				let attachments = Zotero.Items.get((this._attachmentIDs).filter(id => ids.includes(id)));
-				if (attachments.length === 0 && action !== "delete") {
-					return;
-				}
-				for (let attachment of attachments) {
-					this.querySelector(`attachment-row[attachment-id="${attachment.id}"]`)
+				for (let id of ids) {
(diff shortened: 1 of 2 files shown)


dstillman:
1. There's a test failure for the above, which I'm seeing as well when running `./test/runtests.sh -f -g "Attachments pane"`
2. I get a timeout if I run `./test/runtests.sh -f -g "should update attachments pane when attachments changed"` (though this was @windingwind's test, so he might want to fix that).
3. We should make sure we have a test that fails on the reported problem before this fix.

chrome/content/zotero/elements/attachmentsBox.js
```diff
@@ -113,31 +113,13 @@
 				if (attachments.length === 0 && action !== "delete") {
 					return;
 				}
-				if (action == 'add') {
-					for (let attachment of attachments) {
+				for (let attachment of attachments) {
+					this.querySelector(`attachment-row[attachment-id="${attachment.id}"]`)
+						?.remove();
+					if (action !== 'delete') {
```
dstillman, line 119: We should list `modify` and `refresh` explicitly, since others could conceivably be added in the future.

Issue #4126, opened 2024-05-15
fx115: Fix styling in Document Preferences and Bibliography dialog


dstillman:
Things got messed up in fx115, with huge headers and extra space, and some styles can be way too large now:

<img width="646" src="https://github.com/zotero/zotero/assets/783082/d46f810e-b2fa-4cf1-83e2-a0037a6c6805"/>

This causes some very weird behavior on small screens:

https://forums.zotero.org/discussion/comment/462121/#Comment_462121
https://forums.zotero.org/discussion/comment/462638/#Comment_462638

adomasven:
https://github.com/zotero/zotero/commit/d236058351af3792927581a6f5dbce0627999cbb

This is what caused this change, not upgrade to fx115. We can just remove that line.

dstillman:
It's not. That line just switched it to actually use our SCSS rules, which we definitely want — various things, like the table padding and UI font size, are wrong without them, and the headers and spacing are still wrong:

<img src="https://github.com/zotero/zotero/assets/783082/a37eded3-7828-428b-adae-61116a082f14" width="646"/>

We need to actually fix this for real.

dstillman:
We could probably move the "Footnotes" / "Endnotes" radiogroup inline with "Display Citations As" (like "Language:") to save space for note styles.

dstillman:
This is separate from fixing the basic style issues, but do we think all of this text is still necessary? Is there anything (about file formats or whatever) that's obsolete and/or that can be removed/hidden? Should Bookmarks even show up this prominently, or can we hide that in some advanced settings pane (if we even still need to support it)?

dstillman:
Like if Fields is "recommended", we should just default to that, and you should be able to click on "Advanced…" or something to switch to another format.

adomasven:
This applies to `bibliography.xhtml` used in the Create Bibliography from Items and `setDocPrefs.xhtml` used in Document Preferences from word processor plugins.

adomasven:
> Is there anything (about file formats or whatever) that's obsolete and/or that can be removed/hidden? Should Bookmarks even show up this prominently, or can we hide that in some advanced settings pane (if we even still need to support it)?

We support bookmarks, but I agree that it makes sense as a checkbox under an Advanced collapsible section. As would Medline abbreviations, and Automatic Citation updates. Although if any of the advanced options are enabled then the Advanced section should appear expanded.

The default dialog should just have styles, language and Footnotes/Endnotes when applicable.

[...]

PR #4265, opened 2024-06-21
Redesign integration doc prefs window

windingwind:
Redesign export bibliography window

resolve: #4126


windingwind:
Create bib from item dialog:

image

Doc prefs dialog:

image

image

When the doc prefs window overflows the screen available height, the style list height will dynamically shrink to 100px at minimal. You can test by setting `window.screen.availHeight` to a fixed number to constrain the window's maximum height:


https://github.com/windingwind/zotero/blob/442fbd3b1ea9c9e6fe19f8ad6a01247f9a7c0201/chrome/content/zotero/bibliography.js#L301

dstillman:
Here's what I'm seeing on first run of this:

Screenshot 2024-06-20 at 11 45 00 PM

1. The bottom buttons are too close to the bottom of the window.
2. Advanced Options is starting expanded even though no options are selected. This is probably what's causing (1). If I toggle Advanced Options closed and open, the space below the buttons is correct.
3. The selected style is no longer being scrolled into view.

Other things:

- Previously we were hiding options that weren't relevant for a style, and now we're disabling them, but in a color that's almost impossible to read:

  Screenshot 2024-06-20 at 11 58 20 PM

  I think we should keep hiding those instead — there's no reason to make a very busy dialog more busy than necessary, and as it is, I think for many people this will just register as a weird empty space.
- Similarly, if both advanced options are disabled, we could just not show Advanced Options at all:
  
Screenshot 2024-06-21 at 12 06 24 AM

- I know it was in the design, but I'm not sure why there's (a lot) more space after "Display citations as:" than after "Language:". Shouldn't those be the same?
  Screenshot 2024-06-21 at 12 05 03 AM
- It also looks like there's too much space between the radio buttons and the labels.

dstillman:
On disabling vs. hiding, hiding obviously causes more jumping around as you move between styles, there's already going to be some jumping around if people move between note and non-note styles (due to the Footnotes/Endnotes toggle), and mostly people are going to be working in a given style for a given document (or always). If you're lucky enough to be using a style with blissfully few options, you should get the benefit of that with a simpler dialog.

yexingsha:
> there's (a lot) more space after "Display citations as:" than after "Language:"

My bad, those should be the same (8px). The reason that happened was apple design template has 16px spacing when multiple radio button options are listed side by side. But that only works when there's no label or the label is at the other end of the row, so here we should just use 8px between labels and controls within the same row.

> It also looks like there's too much space between the radio buttons and the labels.

Yes, as well as between the checkboxes and the labels, which should be 6px. There's also too much space between the Cancel and OK button, which should be 8px.

Also, I see a slight indent before the "Automatically update citations" checkbox that shouldn't be there.

windingwind:
Thanks.

TODO:

- [ ] The bottom buttons are too close to the bottom of the window on the first run
- [ ] Advanced Options is starting expanded even though no options are selected.
- [ ] The selected style is no longer being scrolled into view.
- [ ] keep hiding rows
- [ ] if both advanced options are disabled, we could just not show Advanced Options at all:
- [ ] update space between label and checkbox/radio

> There's also too much space between the Cancel and OK button, which should be 8px.

That's from the firefox's `<dialog>` element style on MacOS. Should we apply this to all dialog windows?

windingwind:
Updated:

image

yexingsha:
I tested this in Google Docs and Word, and it seems like:

1. Once I expand Advanced Options, it will stay expanded the next time I open document settings. I think this open state should not persist.

2. At first render, all the advanced options available in the current word processor will show up, even the ones that aren't available for the current selected style. Switching to a different style then back can fix this issue.

3. When a style only has the "Store citations as bookmarks" advanced option, which isn't available in Gdocs, it gets an empty Advanced Options section:

Screenshot 2024-06-24 at 21 58 23

The spacing between radio buttons/checkboxes and labels are still a bit too wide.

> That's from the firefox's dialog element style on MacOS. Should we apply this to all dialog windows?

We should, since it's different than the actual MacOS style. You can see a much smaller gap between buttons in e.g. the "Save As" dialog.

Also, it looks like the citation style list items don't have text overflow set to ellipses.

windingwind:
Thanks! Updated:

image

chrome/content/zotero/elements/styleConfigurator.js
```diff
@@ -33,19 +31,26 @@
...
 				</div>
 			</div>
 		`);
 
 		set value(val) {
-			this.querySelector('#style-list').value = val;
+			if (!this.values.includes(val)) return;
+			let styleList = this.querySelector('#style-list');
+			styleList.value = val;
+			styleList.selectedItem.scrollIntoView();
```
dstillman, line 43: We should only scroll if it's not already visible. Compare to Z6:

Screenshot 2024-06-27 at 1 58 40 AM

It would also be nice to scroll the selected style a little bit further into view instead of being right against the top:

Screenshot 2024-06-27 at 1 55 54 AM

dstillman:
The dialog padding is wrong, at least on macOS:

Screenshot 2024-06-27 at 2 12 39 AM

 Compare to https://github.com/zotero/zotero/issues/4126#issuecomment-2125709474

dstillman:
Advanced Options needs to be auto-expanded if any of its options are selected. I'm checking "Store citations as bookmarks" and closing the dialog, and when I reopen it, Advanced Options is collapsed.

windingwind:
> The dialog padding is wrong, at least on macOS

Do we want to apply this padding to all dialog windows on macOS?

yexingsha:
> Do we want to apply this padding to all dialog windows on macOS?

I'm not sure how many dialogs this will apply to, and whether this could mess up some of the more complex ones. As far as I can tell, 20px padding (not 24px as in the mocks, that was a bit too large) should work for this one, RTF Scan, Zotero Style Repository, and split tags.

windingwind:
Thanks for the comments. Updated:

1. dialog padding (also in RTF Scan and split tags; Zotero Style Repository is a website, not part of the client)
2. auto-expand advanced options if any is checked
3. auto-scroll the default selected row into center of the list

image

yexingsha:
The 20px padding is for the dialog window, the spacing between rows should be 16px.

And still need to set text overflow to ellipsis, otherwise the clip looks weird:

Screenshot 2024-06-28 at 15 42 53

windingwind:
Yes, I think the padding is applied to window. Do you mean the gap between rows is too large?

About the row ellipsis, sorry I forgot to mention.  It’s always set to ellipsis, but just not working on the XUL richlist element. We could use our own virtualized table here, which would be able to correctly handle text overflow, but could be a bit too heavy for this PR. We could do it later. I guess?

yexingsha:
> Do you mean the gap between rows is too large?

Yes, specifically the gap between "Automatically update citations" and the row below it. And on a closer look, the gap between the Cancel/OK buttons and the last row above it is too small.

As to the ellipsis issue, I guess one workaround is to simply increase the width of this window to 800px, so that these cut offs will become much rarer. Also, we do have a few 800px dialogs, and this one is arguably complex enough to warrant it. @dstillman what do you think?

dstillman:
Yes, we know the content of this list — it's reasonable to increase the width of the window to show the full title for most styles. 800px seems excessive, though. 600?

dstillman:
Even 575px seems like more than enough at the default font size. If one or two user-installed styles get cut off at larger font sizes, it's fine.

windingwind:
> Yes, specifically the gap between "Automatically update citations" and the row below it

Thanks, it was the `label` element's default padding, didn't notice that. Updated.

> Even 575px seems like more than enough at the default font size.

Updated min-width to 575px.

dstillman:
Great!

chrome/content/zotero/elements/styleConfigurator.js
```diff
@@ -33,19 +31,26 @@
 				xmlns:xul="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul"
 			>
 				<div>
-					<xul:richlistbox id="style-list" tabindex="0" />
+					<xul:richlistbox id="style-list" class="theme-listbox" tabindex="0" />
```
adomasven, line 34: This change added an unwanted focusring on Linux. Is it necessary?

image

windingwind: The change is intentional, I followed our styling in the scaffold window. Since we are using it elsewhere, maybe we should consider adding rules for Linux.

adomasven: I see that you added `theme-listbox` to scaffold's `load.xhtml`. It's not used anywhere else, except for the multi-select `richlistbox` in scaffold, though not sure why. We have many richlistboxes in the codebase that do not use this, and generally for boxes that do not support multi-select there's no reason for the focusring to exist at all, and where `seltype=multi` is set, there is a css rule to add the focusring without adding the `theme-listbox` class.

All this to say that if there is no good actual styling reason for this class then it should be removed.

windingwind: It's initially from https://github.com/zotero/zotero/blob/c5027106c2cccb26e5b3092e3b87947fcce91e26/chrome/content/scaffold/scaffold.xhtml#L483 introduced by @AbeJellinek, I guess. These rules fix the looking of the listbox in darkmode on Windows after redesign.

I am not sure if applying these rules to all listbox elements would break anything. Any ideas @AbeJellinek ?

windingwind: The `theme-listbox` is from firefox: https://searchfox.org/mozilla-central/search?q=theme-listbox&path=&case=true&regexp=false , btw

adomasven: Yes, I know. XUL had a richlistbox and a listbox element, which I assume this is supposed to apply the style of listbox to the richlistbox element, which contains a superset of features of the old listbox. If there are issues with darkmode richlistbox on Windows, that should be addressed separately and directly, without breaking the look on other OSes.

windingwind: How about fixing the focus ring for linux? If the `theme-listbox` is incorrect in the doc prefs window, then it would be incorrect everywhere else.

We can address this separately for Windows, but that would be simply duplicating the `richlistbox.theme-listbox` rules for Windows.

adomasven: > How about fixing the focus ring for linux? If the theme-listbox is incorrect in the doc prefs window, then it would be incorrect everywhere else.

We do not use `theme-listbox` anywhere else, but scaffold for `richlistbox`. It would be incorrect everywhere else if used, because there should never be a focusring for `seltype=single` richlistbox items, and there's already a firefox css rule to add it for `seltype=multiple`. We don't have enough scaffold users on Linux for this to have been reported. Looking at scaffold now, it indeed looks wrong there.

> We can address this separately for Windows, but that would be simply duplicating the richlistbox.theme-listbox rules for Windows.

Are you sure it's not your own changes that fix specifically `theme-listbox` box for Windows, instead of any firefox rules (https://github.com/zotero/zotero/blob/b16bbc4a8ac40fa26b189cfe5fc76d6bd8109f51/scss/components/_richlistbox.scss#L1-L22)? Also, if `theme-listbox` is indeed required for richlistboxes to appear correctly on dark-mode on Windows, we would have had reports as they appear across the Zotero UI.

windingwind: Only `theme-listbox` uses these CSS variables. If without the class, we need to manually duplicate these rules all again.

I am thinking about removing all `theme-listbox` and maintaining a set of rules by ourselves since we expect all richlistbox elements to have the same looking.

adomasven: > Only `theme-listbox` uses these CSS variables. If without the class, we need to manually duplicate these rules all again.

Right, but what I am saying is that only `theme-listbox` richlistbox even looks wrong on darkmode windows, hence adding this class here wasn't necessary to make the richlistbox look good on darkmode windows, and also made it look bad on Linux.

windingwind: Using the `theme-listbox` has a historical reason. Tracing back to the time we tried to fix the styling on Windows after the redesign, the most important richlistitem is the one in the scaffold, which used `theme-listbox`, so the rules were designed for elements with this specific class. Later we found that other richlistitems were also not correct on Windows and borrowing the class name was the fastest solution.

More than once, I also thought about whether to remove the class from the scaffold and maintain unified rules for all richlistitem elements, during the redesign. At that time, rolling out a workable version had the highest priority. Now that we are heading to the target of being more stable and perfect, yes, I guess it's time to do this.

adomasven: > Tracing back to the time we tried to fix the styling on Windows after the redesign, the most important richlistitem is the one in the scaffold

Arguably the most important richlistbox which is seen by most users is precisely this one with the style picker.

Anyway, I just did some testing and it seems that indeed setting `theme-listbox` does make the richlistbox look better on Windows (though not correct without modifications in `_richlistbox.scss`). Still, the correct solution is to fix richlistbox appearance on Windows darkmode without using additional classes, and without breaking appearance of it on other platforms.

PR #4267, opened 2024-06-21
Enable Undo Retrieve Metadata when translator added note 

### LINKED ISSUE #2977: "Undo Retrieve Metadata" doesn't appear if child note is created during lookup
https://forums.zotero.org/discussion/comment/425977/#Comment_425977

Since we have to delete the parent item, having a child note complicates things. I guess we want to test that the note was created within a few seconds of the parent item, and only show this option if so.

AbeJellinek:
Closes #2977


dstillman:
Merged

Issue #4241, opened 2024-06-14
Item box loses focus when tabbing out of field with button after deleting contents


AbeJellinek:
e.g. DOI. Select all the text in the field, delete, press Tab. Focus moves to View Online and then is lost when that button hides.

https://forums.zotero.org/discussion/comment/465664/#Comment_465664

abaevbog:
I added a tweak to address this in https://github.com/zotero/zotero/pull/4178 as it has a few other focus-related fixes.

I ran into it before I saw the comment on the forums and I should've created an issue for it then.

dstillman:
https://forums.zotero.org/discussion/comment/466290/#Comment_466290

I can still reproduce the original issue above.

abaevbog:
Oh, I see. Apologies, I did address only part of the general issue from the forums thread with lost focus (one with focus being lost on tab after any kind of editing that is not deletion). That was due to lack of `id`s on buttons, so they could never be refocused at all.

If the content is emptied, though, the actual options button is never rendered, so there is no node to refocus (regardless of if it is mean to have an `id` or not). I'll tweak it to maybe look for the editable field from that row and simulate a tab from that to focus the next row

abaevbog:
These additions should prevent the focus from getting lost on tab after all content is erased, as well as address the issue of focus not returning to item type menu (which turned out to be a completely separate issue).

PR #4287, opened 2024-06-26
Prevent focus from being lost in itemBox

abaevbog:
- always render options and link buttons - just do not display them if their respective field is empty. That allows us to easily handle focus after refresh because otherwise, the node may does not exist.
If we try to restore focus to such hidden component, simulate a "tab" from it to focus the next possible node. This fixes the issue of focus being lost on tab after all content of editable-text is cleared.
- add tabindex=0 to itemType menulist, otherwise it was not perceived as a candidate to return focus to.
- somewhat special treatment for restoring focus to creator rows. If the desired node is not found, we'll try to focus the respective node in the last creator row. It prevents focus from being lost on tab after clearing the very last creator.

Fixes: #4241

https://github.com/zotero/zotero/assets/36271954/eded7a35-e45a-46ae-9f98-33b6942a8fbd

### DIFF
--- a/chrome/content/zotero/elements/itemBox.js
+++ b/chrome/content/zotero/elements/itemBox.js
@@ -620,7 +620,7 @@
 								.replace(/\?/g, '%3f')
 								.replace(/%/g, '%25')
 								.replace(/"/g, '%22');
-						openLinkButton = this.createOpenLinkIcon(doi);
+						openLinkButton = this.createOpenLinkIcon(doi, fieldName);
 						link = doi;
 						addLinkContextMenu = true;
 					}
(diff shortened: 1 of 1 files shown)


dstillman:
I'm still seeing the exact same bug that @AbeJellinek reported with the DOI field:

https://github.com/zotero/zotero/assets/783082/f05b71b5-4c38-4d86-9c8d-fd8030b028f6

abaevbog:
Ah, thank you. DOI button wasn't getting a proper id with the `fieldname`. It should be fixed now.


https://github.com/zotero/zotero/assets/36271954/16675403-fd37-4d7c-a771-5101e9ed20ec

Issue #4276, opened 2024-06-24
Wrong vertical position in the Item Pane after switching from a note


windingwind:
https://forums.zotero.org/discussion/115450/zotero-7-beta-wrong-vertical-position-in-the-item-pane-after-switching-from-a-note#latest

dstillman:
Reverted

PR #4288, opened 2024-06-27
Fix wrong vertical position in the Item Pane after switching from a note

windingwind:
fix: #4276

### DIFF
--- a/chrome/content/zotero/elements/itemDetails.js
+++ b/chrome/content/zotero/elements/itemDetails.js
@@ -487,7 +487,6 @@
 			// If not, scroll again
 			if (Math.abs(pane.getBoundingClientRect().top
 				- pane.parentElement.getBoundingClientRect().top) > scrollPositionThreshold) {
-				// pane.scrollIntoView({ block: 'start', behavior: 'instant' });
 				await this.scrollToPane(paneID, behavior, retryTimes + 1);
 			}
 			return scrollPromise;
--- a/chrome/content/zotero/elements/itemDetails.js
+++ b/chrome/content/zotero/elements/itemDetails.js
@@ -268,6 +268,10 @@
 			this.renderCustomSections();
 			this._restorePinnedPane();
 
+			// Unobserve intersection to prevent unwanted rendering.
+			// Checking flags in _handleIntersection does not work, since the callback may be delayed
+			this.getPanes().forEach(elem => this._intersectionOb.unobserve(elem));
+
 			let panes = this.getPanes();
 			for (let box of [this._header, ...panes]) {
 				box.editable = this.editable;
@@ -282,6 +286,10 @@
 					}
 				}
 			}
+
+			// Wait to make sure the visibility and height are updated
+			let promise = new Promise(resolve => setTimeout(resolve, 0));
+			await promise;
 
 			let pinnedPaneElem = this.getEnabledPane(this.pinnedPane);
 			let pinnedIndex = panes.indexOf(pinnedPaneElem);
@@ -315,6 +323,8 @@
 			if (Zotero.test) {
 				resolve();
 			}
+
+			this.getPanes().forEach(elem => this._intersectionOb.observe(elem));
 		}
 
 		renderCustomSections() {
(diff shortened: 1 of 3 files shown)


test/tests/itemPaneTest.js
```diff
@@ -926,6 +926,55 @@ describe("Item pane", function () {
...
+			Zotero.Prefs.set("panes.libraries-collections.open", true);
+			Zotero_Tabs.select("zotero-pane");
+		});
+
+		afterEach(function () {
+			Zotero_Tabs.select("zotero-pane");
+			Zotero_Tabs.closeAll();
+		});
+
+		it("should scroll to pinned pane after selection move from note to item", async function () {
```
dstillman, line 942: `move` → `moves`

test/tests/itemPaneTest.js
```diff
@@ -926,6 +926,55 @@ describe("Item pane", function () {
...
+			await note.saveTx();
+
+			await ZoteroPane.selectItem(item.id);
+			await waitForScrollToPane(itemDetails, paneID);
+
+			itemDetails.pinnedPane = paneID;
+
+			await ZoteroPane.selectItem(note.id);
+
+			await Zotero.Promise.delay(200);
```
dstillman, line 965: We really want to avoid random delays — they cause test failures and slow down tests.

If we really can't get an immediate trigger for something, we can poll in a tight loop (e.g., 10 ms), checking for some condition (a specific element or whatever).

chrome/content/zotero/elements/itemDetails.js
```diff
@@ -283,6 +287,10 @@
 				}
 			}
 
+			// Wait to make sure the visibility and height are updated
+			let promise = new Promise(resolve => setTimeout(resolve, 0));
+			await promise;
```
dstillman, line 292: We can use `Zotero.Promise.delay(0)` here and below if we really need to do this, but is this actually reliable?

chrome/content/zotero/elements/itemDetails.js
```diff
@@ -460,6 +470,14 @@
 			}
 			pane.scrollIntoView({ block: 'start', behavior });
 			pane.focus();
+			
+			// Check if the pane is actually scrolled to
+			let promise = new Promise(resolve => setTimeout(resolve, 0));
+			await promise;
+			if (Math.abs(pane.getBoundingClientRect().top - pane.parentElement.getBoundingClientRect().top) > 3) {
```
dstillman, line 477: What's `3`? Move to a named constant.

test/tests/itemPaneTest.js
```diff
@@ -926,6 +926,55 @@ describe("Item pane", function () {
...
+			Zotero_Tabs.select("zotero-pane");
+			Zotero_Tabs.closeAll();
+		});
+
+		it("should scroll to pinned pane after selection move from note to item", async function () {
+			let itemDetails = ZoteroPane.itemPane._itemDetails;
+			let pane = itemDetails.getPane(paneID);
+
+			let item = new Zotero.Item('book');
+			await item.saveTx();
```
dstillman, line 947: `let item = await createDataObject('item');`

test/tests/itemPaneTest.js
```diff
@@ -926,6 +926,55 @@ describe("Item pane", function () {
...
+			let file = getTestDataDirectory();
+			file.append('test.pdf');
+			await Zotero.Attachments.importFromFile({
+				file,
+				parentItemID: item.id
+			});
+
+			let note = new Zotero.Item('note');
+			await note.saveTx();
```
dstillman, line 956: `let note = await createDataObject('item', { itemType: 'note' });`

test/tests/itemPaneTest.js
```diff
@@ -926,6 +926,55 @@ describe("Item pane", function () {
...
+			let pane = itemDetails.getPane(paneID);
+
+			let item = new Zotero.Item('book');
+			await item.saveTx();
+			let file = getTestDataDirectory();
+			file.append('test.pdf');
+			await Zotero.Attachments.importFromFile({
+				file,
+				parentItemID: item.id
+			});
```
dstillman, line 953: `await importPDFAttachment(item);`

test/tests/itemPaneTest.js
```diff
@@ -926,6 +926,55 @@ describe("Item pane", function () {
...
+
+			let note = new Zotero.Item('note');
+			await note.saveTx();
+
+			await ZoteroPane.selectItem(item.id);
```
dstillman, line 958: Can now use `await select(item);` for all of these.

test/tests/itemPaneTest.js
windingwind: Thanks, changed to 10ms.

Issue #4284, opened 2024-06-26
Context notes pane is blank when opened immediately after startup


AbeJellinek:
1. Be in a reader tab
2. Make sure tab state has been saved to disk (`Zotero.Session.save()`)
3. Quit Zotero and reopen (Zotero should reopen with the reader tab selected)
4. Click the Notes button in the sidenav

Notes pane appears blank:

![image](https://github.com/zotero/zotero/assets/1770299/c9fde24c-b4a1-4372-bae3-3de00ade8b64)

It works again if you switch to the library and back.

PR #4289, opened 2024-06-27
Fix context notes pane not load

windingwind:
fix: #4284
introduce a new notifier type "load", "tab", a follow up fix to #4174

### DIFF
--- a/chrome/content/zotero/elements/contextPane.js
+++ b/chrome/content/zotero/elements/contextPane.js
@@ -176,7 +176,7 @@
 			else if (tabType == 'reader') {
 				let reader = Zotero.Reader.getByTabID(tabID);
 				this._handleReaderReady(reader);
-				this._setupNotesContext();
+				this._setupNotesContext(tabID);
 				_contextPaneSplitter.setAttribute('hidden', false);
 
 				_contextPane.setAttribute('collapsed', !(_contextPaneSplitter.getAttribute('state') != 'collapsed'));
@@ -193,16 +193,17 @@
 			ZoteroContextPane.update();
 		}
 
-		async _setupNotesContext() {
-			let selectedTab = Zotero_Tabs.getState().find(tab => tab.selected);
-			if (!selectedTab || !selectedTab.data.itemID) return;
-			let attachment = await Zotero.Items.getAsync(selectedTab.data.itemID);
+		async _setupNotesContext(tabID) {
+			let { tab } = Zotero_Tabs._getTab(tabID);
+			if (!tab || !tab.data.itemID) return;
+			let attachment = await Zotero.Items.getAsync(tab.data.itemID);
 			if (attachment) {
 				this._selectNotesContext(attachment.libraryID);
 				let notesContext = this._getNotesContext(attachment.libraryID);
 				notesContext.updateNotesListFromCache();
 			}
 			let currentNoteContext = this._getCurrentNotesContext();
+			// Always switch to the current selected tab, since the selection might have changed
 			currentNoteContext.switchToTab(Zotero_Tabs.selectedID);
 		}
 
@@ -224,7 +225,6 @@
 					});
 				}
 			}
-			
 		}
 
 		_getCurrentNotesContext() {
--- a/chrome/content/zotero/elements/contextPane.js
+++ b/chrome/content/zotero/elements/contextPane.js
@@ -174,8 +174,7 @@
 				this._sidenav.hidden = true;
 			}
 			else if (tabType == 'reader') {
-				let reader = Zotero.Reader.getByTabID(tabID);
-				this._handleReaderReady(reader);
+				this._handleReaderReady(tabID);
 				this._setupNotesContext(tabID);
 				_contextPaneSplitter.setAttribute('hidden', false);
 
@@ -207,7 +206,8 @@
 			currentNoteContext.switchToTab(Zotero_Tabs.selectedID);
 		}
 
-		async _handleReaderReady(reader) {
+		async _handleReaderReady(tabID) {
+			let reader = Zotero.Reader.getByTabID(tabID);
 			if (!reader) {
 				return;
 			}
--- a/chrome/content/zotero/elements/contextPane.js
+++ b/chrome/content/zotero/elements/contextPane.js
@@ -90,7 +90,7 @@
 				this._handleTabClose(action, type, ids, extraData);
 				return;
 			}
-			if (type == 'tab' && action == 'select') {
+			if (type == 'tab' && ["select", "load"].includes(action)) {
 				this._handleTabSelect(action, type, ids, extraData);
 			}
 		}
@@ -158,7 +158,7 @@
 			});
 		}
 
-		_handleTabSelect(action, type, ids) {
+		async _handleTabSelect(action, type, ids) {
 			// TEMP: move these variables to ZoteroContextPane
 			let _contextPaneSplitter = ZoteroContextPane.splitter;
 			let _contextPane = document.getElementById('zotero-context-pane');
@@ -270,7 +270,7 @@
 			}
 		}
 	
-		async _addItemContext(tabID, itemID, tabType = "") {
+		async _addItemContext(tabID, itemID, _tabType = "") {
 			let { libraryID } = Zotero.Items.getLibraryAndKeyFromID(itemID);
 			let library = Zotero.Libraries.get(libraryID);
 			await library.waitForDataLoad('item');
--- a/chrome/content/zotero/elements/contextPane.js
+++ b/chrome/content/zotero/elements/contextPane.js
@@ -158,21 +158,23 @@
 			});
 		}
 
-		async _handleTabSelect(action, type, ids) {
+		async _handleTabSelect(action, type, ids, extraData) {
 			// TEMP: move these variables to ZoteroContextPane
 			let _contextPaneSplitter = ZoteroContextPane.splitter;
 			let _contextPane = document.getElementById('zotero-context-pane');
+			let tabID = ids[0];
+			let tabType = extraData[tabID].type;
 			// It seems that changing `hidden` or `collapsed` values might
 			// be related with significant slow down when there are too many
 			// DOM nodes (i.e. 10k notes)
-			if (Zotero_Tabs.selectedType == 'library') {
+			if (tabType == 'library') {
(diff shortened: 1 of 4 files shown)


chrome/content/zotero/elements/contextPane.js
```diff
@@ -171,7 +171,7 @@
 				ZoteroContextPane.showLoadingMessage(false);
 				this._sidenav.hidden = true;
 			}
-			else if (Zotero_Tabs.selectedType.includes('reader')) {
+			else if (Zotero_Tabs.selectedType == 'reader') {
```
dstillman, line 174: This was presumably breaking something. Do we know what?

chrome/content/zotero/elements/contextPane.js
```diff
@@ -90,7 +90,7 @@
 				this._handleTabClose(action, type, ids, extraData);
 				return;
 			}
-			if (type == 'tab' && action == 'select') {
+			if (type == 'tab' && ["select", "load"].includes(action)) {
```
dstillman, line 93: This doesn't seem right. Won't this cause `_handleTabSelect()` to be called twice in rapid succession when you click on a tab?

chrome/content/zotero/xpcom/reader.js
```diff
@@ -1831,6 +1830,8 @@ class Reader {
 				}
 			});
 			this._readers.push(reader);
+			// Change tab's type from "unloaded-reader" to "reader" after reader loaded
+			win.Zotero_Tabs.markAsLoaded(tabID);
```
abaevbog, line 1834: On windows (especially slower ones like my VM), this may lead to a kind of race condition. When tab A is loading, if we click from it onto another tab B and then click back into the original tab A, then (if reader has not finished loading) as far as [Zotero_Tabs.select](https://github.com/zotero/zotero/blob/5a6b1a9f76dd42b09cb820907aa0cd6761cf641f/chrome/content/zotero/tabs.js#L454) is concerned, tab A is still `reader-unloaded` so it will call `Zotero.Reader.open` again. This can lead to multiple reader instances being opened in the same tab. That was the main reason behind placing the `Zotero_Tabs.markAsLoaded` at the top of the function (which 100% needed to be explained in a comment)

This is what it would look like if I start the app and, while things are loading, click between two tabs a few times:

Screenshot 2024-06-27 at 9 47 27 AM

Something I thought of earlier was to introduce another tab type, like `reader-loading`, to set this status at the top of the function and change it to just `reader` when everything is loaded. Not sure if it would lead to conflicts with anything else though

chrome/content/zotero/elements/contextPane.js
abaevbog: I initially changed it to be Zotero_Tabs.selectedType.includes('reader') as I was running into issues with context pane sometimes not properly appearing during (granted, quite aggressive) tab switching. At that point, a tab could be either reader or reader-unloaded, so this seemed appropriate. Though, I went back to https://github.com/zotero/zotero/commit/3f45def928633be3dd4f5d8fcbedb6b589b46236 and I couldn't reproduce that issue I was having with just Zotero_Tabs.selectedType == 'reader'. Perhaps the issue I saw came from a VM being slow or something like that and the includes part is really not required.

As far as I can see, the this._handleReaderReady(reader);below would be a noop for a tab whose reader is not yet loaded, and then we just set the contextPane collapsed status and sidenav visibility, so I would expect no serious breakages there.

dstillman: Oh, I didn't realize it was targeting `reader-unloaded`. I think `.startsWith('reader')` would be clearer (and nominally faster) and look less like a type mistake. But if we don't need it, that's fine too.

abaevbog:
I've been thinking about how to work around that race condition that can create multiple reader instances in the same tab if you click around fast enough, and I have a question (with a suggestion, maybe).

We seem to currently have the setting up of the notes context somewhat dependent on the reader. Unless the reader is [ready](https://github.com/zotero/zotero/blob/20e9f2058065595b257a6f2933f8a20685ed0127/chrome/content/zotero/elements/contextPane.js#L195), notes do not [load](https://github.com/zotero/zotero/blob/20e9f2058065595b257a6f2933f8a20685ed0127/chrome/content/zotero/elements/contextPane.js#L217). Is it necessary? Could we just pull out the logic of setting up the notes context from the `_handleReaderReady` into, maybe, its own function that would run regardless if the reader is ready or not. I don't think the reader itself is required to fetch the notes, we just need the actual items to be loaded.

I am thinking about something like this: https://github.com/abaevbog/zotero/commit/1c283159f88c5661f86fa7362a7890e4c2364d4e - with this, the notes appear for me after initial loading (though it does take a moment). Something like this might be easier than working around that race condition.

I also could be missing something her so correct me if some of this sounds off.

windingwind:
Thank you @abaevbog , looks neat. Do you want me to merge that into this PR or you prefer to start a new PR?

chrome/content/zotero/xpcom/reader.js
windingwind: Might it be from using `Zotero_Tabs.selectedID`? We could use the `extraData[ids[0]].type` from the notifier callback arguments, which will be called only once for each tab.

chrome/content/zotero/elements/contextPane.js
windingwind: Might be from using `Zotero_Tabs.selectedID`. Updated to use the `extraData[ids[0]].type` from the notifier callback arguments, which will be called only once for each tab.

windingwind:
Anyway, I've just merged @abaevbog 's commit and force-pushed.

chrome/content/zotero/xpcom/reader.js
windingwind: I'm setting the tab to `reader-loading` immediately. Hopefully, that gets rid of the duplication loading issue.

abaevbog:
This does work and the notes show up right away! 

I am wondering, if we do decouple the logic of setting up notes pane from the reader so it will run even if the reader is not yet loaded (which by itself fixes the issue of notes not showing after load, at least for me), what is the benefit of also adding the `load` notifier type and handling it? On the other hand, if we have a `load` notifier event that will trigger `this._handleReaderReady` when a reader is already loaded, we don't stop [here](https://github.com/zotero/zotero/blob/2dfb243924b5f33eabe00302c298d0bdbb51b071/chrome/content/zotero/elements/contextPane.js#L196), so the notes setup will run - in which case, do we need to pull out the notes logic from `this._handleReaderReady`?

My thinking was that we would do one or the other and the only reason why I suggested something like https://github.com/abaevbog/zotero/commit/1c283159f88c5661f86fa7362a7890e4c2364d4e is that it seemed a bit simpler (and not involving the tab types-related potential race conditions which are scary). There may be benefits to adding the `load` notifier though, especially for the future. With https://github.com/zotero/zotero/pull/4289/commits/519da33ad2498dd002ff2643d9474f57e16117e0, the issue of multiple reader instances in the same tab seems to be taken care of, in which case https://github.com/abaevbog/zotero/commit/1c283159f88c5661f86fa7362a7890e4c2364d4e isn't really necessary. Unless we want to do it to make it explicit that the notes pane setup is not dependent on the reader?

These are my thoughts! Correct me if I am misinterpreting something here

windingwind:
Thanks for the detailed reply!

We need a notifier event to be triggered after the reader is ready, since currently if the reader is not yet loaded by the time `ContextPane#_handleTabSelect` is called, the `_handleReaderReady` may never be executed. I was thinking of reusing existing events, e.g. tab+select, but that's not satisfying and can break something else. So in that case, we may have to add a new notifier event type to indicate the reader is loaded.

abaevbog:
I think I see another issue here but this one seems to be very easy to fix with the new `load` tab notifier.  The view menu does not update when you switch between unloaded tabs, though it does update ok once you switch to a tab that is loaded.

https://github.com/zotero/zotero/assets/36271954/fb4c340c-26ff-43c1-a8e1-3b4e725ab485

This is essentially #4308, just in a different form. When a new tab is loaded, in `standalone.js`, we [set](https://github.com/zotero/zotero/blob/390b65f2abc45e6ded6be2214df10bd90ad3109a/chrome/content/zotero/standalone/standalone.js#L75) the type and sub-type of the menus, and if the type is not "reader", or if the reader is not loaded yet, it basically doesn't work right. And we do need the reader to be loaded (unlike for the notes panel) because we need to get its [subtype](https://github.com/zotero/zotero/blob/390b65f2abc45e6ded6be2214df10bd90ad3109a/chrome/content/zotero/standalone/standalone.js#L74).

With the new `load` notifier event, though, I think we could just do `if (action == "select" || action == "load")` [here](https://github.com/zotero/zotero/blob/390b65f2abc45e6ded6be2214df10bd90ad3109a/chrome/content/zotero/standalone/standalone.js#L60), similar to what is being done here in `contextPane.js`. I tried and it seems to address it for me.

So the `load` event comes in very handy! Do we need https://github.com/abaevbog/zotero/commit/1c283159f88c5661f86fa7362a7890e4c2364d4e then? `_handleReaderReady` will run once the reader is loaded, so the notes panel setup [part](https://github.com/zotero/zotero/blob/390b65f2abc45e6ded6be2214df10bd90ad3109a/chrome/content/zotero/elements/contextPane.js#L217) should be reachable?

windingwind:
Thanks for the reply!

> Do we need https://github.com/abaevbog/zotero/commit/1c283159f88c5661f86fa7362a7890e4c2364d4e then? _handleReaderReady will run once the reader is loaded, so the notes panel setup [part](https://github.com/zotero/zotero/blob/390b65f2abc45e6ded6be2214df10bd90ad3109a/chrome/content/zotero/elements/contextPane.js#L217) should be reachable?

I guess being able to load the notes context before reader is ready is better than leaving the whole window blank until the reader is ready. Actually I was thinking of bringing a better experience of waiting for reader loading, e.g. some fake animation for the reader window. I think we have no reason to wait for the reader to load the notes context?

abaevbog:
> I guess being able to load the notes context before reader is ready is better than leaving the whole window blank until the reader is ready. Actually I was thinking of bringing a better experience of waiting for reader loading, e.g. some fake animation for the reader window. I think we have no reason to wait for the reader to load the notes context?

Gotcha, I see you mean! That makes sense to me.

Issue #4223, opened 2024-06-10
VPAT Reader: announcements


abaevbog:
- [x] vpat 72: Show annotations/thumbnails/document outline buttons should announce that the sidebar view has changed after click.
- [x] vpat 73: Nothing is announced during keyboard navigation of the thumbnails in the sidebar. Similar to Preview, perhaps we should announce the page number of the thumbnail?
- [x] vpat 75: When the annotation are navigated with arrows in the scrollable content pane, there is no audible feedback. We should announce something to indicate what annotation has been focused and for the user to to not loose orientation within the document.
- [x] vpat 78: Similar to vpat 75, while navigating search results after typing something in "Find in document" (via enter or by clicking arrow buttons), we should include some kind of indication as to what is the current search result and where it is in the document.

abaevbog:
@mrtcode so a couple of issues here that I did not touch in https://github.com/zotero/reader/pull/127 because they are a bit trickier are vpat 75 and 78. I thought I should check in with you on the best way to approach this (and if it's something I should approach to begin with).

vpat 75 and 78 are about how the selection within the reader can move between annotations or search results but it does not announce anything to screen reader.
Here is a [brief video](https://www.dropbox.com/scl/fi/840it2anqb0jiz0alzf3s/reader_announcements.mov?rlkey=sddsdr8pa74r6y7ehv3dljqbj&st=zc1m8wpf&dl=0) to display exactly what I mean. This gets interesting because, as far as I understand, the annotations or search results are never actually focused, just highlighted. So I was thinking maybe we could add a hidden [aria-live](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) region where we would insert the string to announce when the selection changes? That way things can get announced regardless of focus and such.

Do you think this makes sense for me to try? If so, where could I look in the reader to find where the highlighting of search results and keyboard selection of annotations in the actual reader happens?

mrtcode:
We can try to do this. But where that aria-live region should be placed?

abaevbog:
I don't believe there are any strict requirements as to where in the DOM they need to be. Overall, any node can be marked as `aria-live` and screen readers should be able to see and act on it.

Firefox has a similar node used exclusively for announcements and it is places just at the bottom of [browser.xhtml](https://searchfox.org/mozilla-central/source/browser/base/content/browser.xhtml#175). I think something like that would make sense?


[...]

PR #4293, opened 2024-06-27
Strings for reader aria-live announcements

abaevbog:
Used to announce which annotation is selected during keyboard navigation, as well as the index of the current search result. For reference, this is a demo of the announcements: [link](https://www.dropbox.com/scl/fi/oqe8d8s3lu31e4bohfu7f/reader_live_2.mov?rlkey=pvolawitxxmhndzyq99cf45c3&st=hixl309k&dl=0)

These may be affected by followup edits to reader's announcements logic per https://github.com/zotero/zotero/issues/4223#issuecomment-2194427797 so I'm marking this as a draft until then.

Followup to zotero/reader#130


abaevbog:
Per https://github.com/zotero/zotero/issues/4223#issuecomment-2199497002 these are good to go now

PR #4296, opened 2024-06-28
Add Gecko types and a script to update them

AbeJellinek:
Adds Mozilla's type definitions for XPCOM/privileged APIs. Gives us type checking and code completion in editors that support TypeScript type definitions (IntelliJ/WebStorm, VSCode, and others). I committed everything, but we could alternatively add the `types/gecko` directory to `.gitignore` and have people run the script on their end if they want type definitions.


windingwind:
I've been using the types/gecko for a long time and we may want to remove the `/// <reference no-default-lib="true" />`:

https://github.com/windingwind/zotero-types/blob/ab31e6c41412b6050e0f904294dfbdce0db622f2/scripts/updateGecko.mjs#L35-L41

It's overwriting the default definitions, e.g. `lib.dom.d.ts`, which causes some very fundamental methods/properties to be missing, e.g. `document.querySelector`.

Issue #4295, opened 2024-06-28
Trashed collections/searches incorrect context menu and missing header in itemPane


abaevbog:
1. Permanently delete option is disabled in context menu. And options that are not relevant (Show in library, generate report) should not appear.
2. The header with buttons to permanently delete or restore does not appear for deleted collections

https://forums.zotero.org/discussion/comment/466736/#Comment_466736

<img width="864" alt="Screenshot 2024-06-28 at 11 47 08 AM" src="https://github.com/zotero/zotero/assets/36271954/7b063a5a-cca6-4367-b90c-7f8fa99a17b5">

PR #4297, opened 2024-06-28
Show correct context menu and itemPane header for trashed collections

abaevbog:
- only have enabled "Permanently delete" and "Restore to library" options in the context menu for trashed collections and searches
- display the custom header with buttons to delete or restore in the itemPane when a trashed collection or search is selected

Fixes: #4295

https://forums.zotero.org/discussion/comment/466736/#Comment_466736

Screenshot 2024-06-28 at 11 53 34 AM

### DIFF
--- a/chrome/content/zotero/elements/itemPane.js
+++ b/chrome/content/zotero/elements/itemPane.js
@@ -494,7 +494,7 @@
 				// Guess a mode from the current data
 				// No/multiple objects are selected OR selected object is a trashed collection/search
 				if (!this.data.length || this.data.length > 1
-					||this.data.every(obj => obj instanceof Zotero.Collection || obj instanceof Zotero.Search)) {
+					|| this.data[0] instanceof Zotero.Collection || this.data[0] instanceof Zotero.Search) {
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/elements/itemPane.js
```diff
@@ -492,7 +492,9 @@
 		getCurrentPane(mode = undefined) {
 			if (!mode) {
 				// Guess a mode from the current data
-				if (!this.data.length || this.data.length > 1) {
+				// No/multiple objects are selected OR selected object is a trashed collection/search
+				if (!this.data.length || this.data.length > 1
+					||this.data.every(obj => obj instanceof Zotero.Collection || obj instanceof Zotero.Search)) {
```
dstillman, line 497: This just needs to check `this.data[0]`, right? We're already excluding n != 1.

(And space after `||`)

abaevbog: That's fair! Somehow `.every` felt more explicit to me - but it does only apply to `this.data[0]`, so I pushed that small tweak

Issue #4307, opened 2024-07-01
Trashed collections do not get immediately removed from itemPane "Libraries and Collections" section


abaevbog:
https://forums.zotero.org/discussion/115597/zotero-7-beta-deleting-or-restoring-collections-is-not-updated-in-the-item-pane-of-their-items

PR #4324, opened 2024-07-03
 Fix to stale or incorrect trashed collections data appearing in librariesCollectionsBox

abaevbog:
- listen to `collection` notifier events and re-render  the section if a relevant collection is trashed or modified. That way, a trashed collection will be removed, restored collection will be added back and renaming a collection will update the name in the librariesCollectionsBox.
  - fixed a bug where a restored collection would not be added into `item._collections` cache. Fixed false-positive test for it.
  - do not add trashed collections into `item. _collections` cache in `Zotero.Items._loadCollections`. Otherwise, trashed collections will appear in librariesCollectionsBox after the app is restarted.

  Fixes: #4307

### DIFF
--- a/chrome/content/zotero/elements/librariesCollectionsBox.js
+++ b/chrome/content/zotero/elements/librariesCollectionsBox.js
@@ -65,7 +65,7 @@ import { getCSSIcon } from 'components/icons';
 		}
 
 		init() {
-			this._notifierID = Zotero.Notifier.registerObserver(this, ['item'], 'librariesCollectionsBox');
+			this._notifierID = Zotero.Notifier.registerObserver(this, ['item', 'collection'], 'librariesCollectionsBox');
 			this._body = this.querySelector('.body');
 			this.initCollapsibleSection();
 			this._section.addEventListener('add', this._handleAdd);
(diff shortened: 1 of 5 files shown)


chrome/content/zotero/elements/librariesCollectionsBox.js
```diff
@@ -77,11 +77,19 @@ import { getCSSIcon } from 'components/icons';
...
+			if (!this._item) return;
 			if (action == 'modify'
-					&& this._item
+					&& type == "item"
 					&& (ids.includes(this._item.id) || this._linkedItems.some(item => ids.includes(item.id)))) {
 				this._forceRenderAll();
 			}
+			
+			if (["modify", "trash"].includes(action) && type == "collection") {
+				let isRelevantCollection = ids.findIndex(id => Zotero.Collections.get(id).hasItem(this._item)) !== -1;
```
dstillman, line 88: `ids.some()` would probably be clearer

dstillman (changes requested):
We should have a test for the Libraries & Collections box too, not just the cache issue.

chrome/content/zotero/elements/librariesCollectionsBox.js
abaevbog: Yes, certainly

abaevbog:
I just pushed a few tests to cover these behaviors for `librariesCollectionsBox` as well. On `main`, all of these tests would currently fail as expected

Issue #4325, opened 2024-07-03
"Assign Color…" colorpicker is blank on Windows


dstillman:
https://forums.zotero.org/discussion/115687/cant-see-color-palette-when-assigning-color-to-a-tag

From this line:

https://github.com/zotero/zotero/commit/a6076ce76c2401b4313d2905afac74836834681c#diff-addfc1f4adbaca701f44381e81086bda6976514ac5d707b9b0574a8ea57fb79cR10

windingwind:
working on it

PR #4326, opened 2024-07-03
Fix colorpicker styling on Windows

windingwind:
fix: #4325


windingwind:
Also applied our dialog styles.

Before:

https://forums.zotero.org/discussion/115687/cant-see-color-palette-when-assigning-color-to-a-tag

After:

image

PR #4328, opened 2024-07-03
Fix RTF Scan never marks citations as unmapped

tnajdek:
When porting RTF Scan, I've kept the logic unchanged. However, I've now noticed that unmapped citations are always classified as ambiguous due to an invalid check (`Zotero.Search` returns an empty array if no results are found, and that is truthy, so we need to check for length instead).

I've tracked this back to the [original code](https://github.com/zotero/zotero/blame/f274323478bbc0fb26aa876a832197add76891d8/chrome/content/zotero/rtfScan.js#L233), so perhaps `Zotero.Search` has changed in the meantime or maybe it has never worked - or maybe I'm missing something?


dstillman:
Yeah, looks like this would've been broken since db0fa3c33e9db6a793aef04d9521fea848c31e5c 10 years ago. Good catch!

PR #4334, opened 2024-07-04
Improve RTF Scan styling

### LINKED ISSUE #4322: Fix RTF Scan
1. The font size should be 13px;
2. `rtfScan-introPage-description` is mixed up;
3. The `document-accepted` and `document-missing` icons in the verify list don't look like they can be clicked. I think we need to change the list item structure a bit and put actual buttons there (see mock below);
4. Various sizing and padding issues.

rtf-scan-introduction

rtf-scan-verify

rtf-scan-format

tnajdek:
* [x] Increase font size to 13px
* [x] Tweak padding
* [x] Welcome screen: Improve layout
* [x] Welcome screen: Fix description
* [x] Verify citations screen: Convert icons to buttons (add a tooltip, `:hover` effect, keyboard accessibility)
* [x] Verify citations screen: Add stripes to the table
* [x] Verify citations screen: Change icons
* [x] All wizards: Adjust bottom buttons to align with content on macOS
* [x] Everywhere: Fix custom focus-ring color does not match native focus-ring color in dark mode on macOS

Screenshot 2024-07-04 at 16 59 18
Screenshot 2024-07-04 at 16 59 51
Screenshot 2024-07-04 at 16 59 57

There is also a separate fix for citations never being marked as "unmapped" in #4328

Resolve #4322


yexingsha:
Looks good! Regarding your questions and other adjustments needed:

1. Omitting "display citations as" was not intentional, and you are right that we should keep the wording consistent. 

2. I think we should use monochrome icons here, because that's how button icons look like everywhere else. Only item type icons have fill colors. Regarding the icons to use, I think the file + question mark icon looks too much like an indication of error, rather than an action to take. Zotero 6 uses the linked document item type icon, so I think we can just use the link icon here to better distinguish the two actions. If you have other suggestions, they are certainly welcome. Here is the accpet match icon in svg: 
accept-match

3. The window padding should be 20px (only the bottom padding needs adjustment); the spacing between rows should be 16px (the formatting window has too much spacing between rows).

4. In Introduction, there is an extra "Zotero" at the beginning of the introduction, and the focus ring on the Choose File buttons are cut off at the right edge:
Screenshot 2024-07-05 at 15 03 57

5. In the Verify window, the twisty before citations should be inset and close to the citation, so that the list hierarchy remains clear.

tnajdek:
@yexingsha Thank you for feedback, much appreciated. I've implemented 1-4.

> In the Verify window, the twisty before citations should be inset and close to the citation, so that the list hierarchy remains clear.

Do you mean like this? (this is just a mock up, I need to investigate if our implementation of the virtual table would allow this):
Screenshot 2024-07-09 at 15 24 49

yexingsha:
Yes, and the text of citations should remain aligned:

rtf-scan-verify

tnajdek:
@yexingsha thanks for explaining, tweaked. Good to go?

Screenshot 2024-07-10 at 10 53 50

yexingsha:
Yes, looks good!

Issue #4310, opened 2024-07-02
Reader: "Invalid aria prop `aria-description` on <button> tag" warning in console


dstillman:
I'm not sure if this is because `aria-description` is [still in draft](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-description) or because we're on an outdated version of React.

Is `aria-description` actually functional in Fx115? If so, we should see if upgrading React avoids this.

abaevbog:
Yeah, we had this discussed in https://github.com/zotero/reader/pull/127#issuecomment-2163588247

Essentially, this must be just this react bug https://github.com/facebook/react/issues/21035 where it throws a warning that `aria-description` is not a valid aria property even though it is and it works very much as it should. For example, if you focus the any of the tags in the sidebar of the reader with voiceover, it does say "Filter annotations by this tag", which is the `aria-description` bit.

We are currently on react 17 and upgrading react to react 18 should fix the issue.

Alternatively, @mrtcode [mentioned](https://github.com/zotero/reader/pull/127#issuecomment-2167285094) the possibility of having the warning just suppressed?

mrtcode:
I tried to suppress this error message but it was harder than I thought. I think we should just upgrade React.

abaevbog:
Gotcha! I'll look into it. I suppose we would have to switch to react 18 at some point.

abaevbog:
So I am working through [migrating to React 18](https://react.dev/blog/2022/03/08/react-18-upgrade-guide), and I wanted to check in on a couple of things to make sure I don't veer off the track.

1. Despite my initial idea, the react version in `package.json` of the reader did not affect that warning - it is the react version in the actual main repo. I guess given this is the react that is exposed in `resource://zotero/react.js` it makes sense...
[...]

PR #4338, opened 2024-07-06
Upgrade to react 18

abaevbog:
Upgrading to react 18 to not have incorrect `aria-description`-related warnings in the console. 

Changes:
- `ReactDOM.render()` -> `ReactDOM.createrRoot().render`
   - await for promise that is resolved in ref attribute of `root.render()` as an alternative for removed callback from `ReactDOM.render`
   - await-ing for promise every time when ref needs to be used after render (e.g. tag selector container), otherwise ref will be undefined
   - additional `window.sizeToContent` calls to properly size dialogs with react-rendered content (e.g. create parent), otherwise the window can cut off some of the content.
- `Icons.getCSSIcon` instead of `getDOMElement`
   - `getDOMElement` relied on `React.renderToStaticMarkup`, which is react 18 was moved to a different file than the
one exposed with `react-dom-server`. To not add another file just for that one function, replace `getDOMElement` with `getCSSIcon`.
   -  `getDOMElement` was mainly used for a few remaining png icons that were not replaced with svg. For those few icons, just record which background-url should be set when the module loads and add it in `getCSSIcon` if applicable. Alternatively, `background-image` for those pngs could be moved into a stylesheet?
   - a few hardcoded twisty svgs in icons.jsx are not used anywhere (they would be fetched via IconTwisty), so those are removed
- `ReactDOM.unmountComponentAtNode` => `root.unmount()`
   - unmountComponentAtNode is deprecated, so save or pass the root as a prop and call root.unmount when the component is destroyed

There are still a few warnings related to `ReactDOM.render` being deprecated but those come from the reader or note-editor, which should be updated as well.

Fixes: #4310

### DIFF
--- a/chrome/content/zotero/components/createParent/createParent.jsx
+++ b/chrome/content/zotero/components/createParent/createParent.jsx
@@ -80,11 +80,12 @@ CreateParent.propTypes = {
 Zotero.CreateParent = memo(CreateParent);
 
 
-Zotero.CreateParent.destroy = (domEl) => {
-	ReactDOM.unmountComponentAtNode(domEl);
+Zotero.CreateParent.destroy = () => {
+	Zotero.CreateParent.root.unmount();
 };
 
 
 Zotero.CreateParent.render = (domEl, props) => {
-	ReactDOM.createRoot(domEl).render(<CreateParent { ...props } />);
+	Zotero.CreateParent.root = ReactDOM.createRoot(domEl);
+	Zotero.CreateParent.root.render(<CreateParent { ...props } />);
 };
--- a/chrome/content/zotero/components/createParent/createParent.jsx
+++ b/chrome/content/zotero/components/createParent/createParent.jsx
@@ -25,12 +25,17 @@
 
 'use strict';
 
-import React, { memo } from 'react';
+import React, { memo, useEffect } from 'react';
 import ReactDOM from "react-dom";
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 
 function CreateParent({ loading, item, toggleAccept }) {
+	// With React 18, this is required for the window's dialog to be properly sized
+	useEffect(() => {
+		window.sizeToContent();
+	});
+
 	// When the input has/does not have characters toggle the accept button on the dialog
 	const handleInput = (e) => {
 		if (e.target.value.trim() !== '') {
@@ -81,5 +86,5 @@ Zotero.CreateParent.destroy = (domEl) => {
 
 
 Zotero.CreateParent.render = (domEl, props) => {
-	ReactDOM.render(<CreateParent { ...props } />, domEl);
+	ReactDOM.createRoot(domEl).render(<CreateParent { ...props } />);
 };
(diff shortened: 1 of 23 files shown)


abaevbog:
I think I rushed a bit, a few tests are failing. I'll mark it as a draft till that's sorted.

abaevbog:
Ok, tests are passing! I had to add a few odd tweaks to tests, potentially due to react flakiness, to avoid race conditions.

@tnajdek what do you think about this approach?

chrome/content/zotero/components/createParent/createParent.jsx
```diff
@@ -25,12 +25,17 @@
...
+import React, { memo, useEffect } from 'react';
 import ReactDOM from "react-dom";
 import PropTypes from 'prop-types';
 import cx from 'classnames';
 
 function CreateParent({ loading, item, toggleAccept }) {
+	// With React 18, this is required for the window's dialog to be properly sized
+	useEffect(() => {
+		window.sizeToContent();
+	});
```
tnajdek, line 37: Even though it doesn't matter for this component, as I don't think it can re-render (no state, props don't change), I'd recommend adding empty array as the `dependencies` argument to the `useEffect` to ensure that logic of the effect is executed once when component mounts and not on every update.

test/tests/advancedSearchTest.js
```diff
@@ -22,7 +22,9 @@ describe("Advanced Search", function () {
 		var promise = waitForWindow('chrome://zotero/content/advancedSearch.xhtml');
 		zp.openAdvancedSearchWindow();
 		var searchWin = yield promise;
-		
+		// After upgrading to react 18, at this point the onload listener has
+		// not finished running yet, so an extra tick needs to be skipped
+		yield Zotero.Promise.delay();
```
tnajdek, line 27: I'm wondering if it wouldn't be better to wait for something solid, not just the next call stack, just to make sure test won't end up being flaky (it doesn't seem to be so maybe irrelevant). E.g. we could make use of the second argument to `waitForWindow` instead?

````js
var promise = waitForWindow('chrome://zotero/content/advancedSearch.xhtml', async (win) => {
			while (!win.ZoteroAdvancedSearch.itemsView) {
				Zotero.debug("Waiting for ZoteroAdvancedSearch");
				await Zotero.Promise.delay(5);
			}
		});
````

abaevbog:
Thank you, these are both great points (and picky is always good, if you ask me)! I added `[]` to that one `useEffect` as well as proper waiting for the `win.ZoteroAdvancedSearch.itemsView` to appear. That's definitely a more reliable way to wait for everything to be ready.

I looked a bit more into the necessity to move a few `waitForTagSelector` promises and I assume it's due to somewhat [updated](https://react.dev/blog/2022/03/08/react-18-upgrade-guide#automatic-batching) batching behavior in react 18, so the tag selector gets re-rendered now at somewhat different points/frequency than before. This part feels a bit flaky to me but I re-ran the tag selector tests 10 times and they reliably pass, so I suppose it's ok.

dstillman:
Squashed a little bit and merged

Issue #4258, opened 2024-06-18
Fix Scaffold interface


yexingsha:
All following issues show up on both macOS and Windows:

1. The "Insert Template" button does not have the correct width:
	<img width="177" alt="Screenshot 2024-06-18 at 17 10 01" src="https://github.com/zotero/zotero/assets/57771795/f6b9bc97-4f77-48eb-9fc1-8c506f4b080f">

2. The active tab is misaligned with its background, which is also misaligned with the rest of the background. And separators are in the wrong spots:
	<img alt="Screenshot 2024-06-18 at 17 11 34" src="https://github.com/zotero/zotero/assets/57771795/2dca1854-a0aa-4a2e-8442-d1ba62ba64c3">
	
3. The test list columns have too little horizontal padding, their headers lack separators, and have a 2px bottom border/shadow:
	<img width="565" alt="Screenshot 2024-06-18 at 17 13 16" src="https://github.com/zotero/zotero/assets/57771795/0903f5ba-071b-4627-98b7-799a56fd39c6">
	
4. In the Browser and Test Input panels, there is not enough spacing (8px) between the top row and the box below:
	<img width="569" alt="Screenshot 2024-06-18 at 17 15 22" src="https://github.com/zotero/zotero/assets/57771795/546f5518-a481-4ea3-a520-cf1bd254935d">
	<img width="568" alt="Screenshot 2024-06-18 at 17 15 31" src="https://github.com/zotero/zotero/assets/57771795/e300c70c-6b55-4cce-89b7-5c2c7b352b70">
	
5. In general, the buttons seem to have an unnecessary horizontal margin, causing misalignment with other elements, and also too much space in between. I feel like this is a general issue after fx115, since it has also shown up in feed item pane, settings, etc.

PR #4339, opened 2024-07-06
Fix scaffold window UI

windingwind:
fix: #4258

### DIFF
--- a/chrome/content/scaffold/load.xhtml
+++ b/chrome/content/scaffold/load.xhtml
@@ -24,7 +24,6 @@
     ***** END LICENSE BLOCK *****
 -->
 <?xml-stylesheet href="chrome://global/skin/" type="text/css"?>
-<?xml-stylesheet href="chrome://zotero/skin/scaffold.css" type="text/css"?>
 <?xml-stylesheet href="chrome://zotero/skin/zotero.css" type="text/css"?>
 <?xml-stylesheet href="chrome://zotero-platform/content/zotero.css"?>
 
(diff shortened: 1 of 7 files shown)


yexingsha:
Looks good! Just a couple of small issues:

1. The text in tabs shift around slightly when different tabs are selected.

2. Buttons still have too much horizontal margin. For example, in the Tests tab, the Delete button should be aligned with the left edge of the input boxes, and the spacing between buttons should be 8px.

3. The Metadata tab seems to have a slightly larger left padding.

windingwind:
Thanks! Fixed:

image

image

yexingsha:
Much better now! Looking at the list of tests, I just realized that the columns aren't aligned with their headers. Is it possible to use our virtualized table style here?

windingwind:
Unfortunately, it is not possible to fix the misalignment unless we replace the `richlistbox` with our v-table.

yexingsha:
Ah, I see. This is not super important, so let's just keep it as is for now.

Geo123abc:
problem: [selectitems.xhtml can't  render correctly](https://groups.google.com/g/zotero-dev/c/s9xPbtjYXfk)

@zzlb0224 found that adding [width: 100%](https://github.com/zotero/zotero/blob/174bde74516f7a5a739845d69f6d4233ca0ecb60/scss/components/_richlistbox.scss#L33) will make the image fill the container. If the label still has overflow: hidden;, it will be hidden, showing as follows:

windingwind:
@Geo123abc thanks, should be fixed in b16bbc4.

dstillman:
Fixed in beta 100

PR #4356, opened 2024-07-08
RTL: Add direction-aware arrow key utilities, use in lots of places

### LINKED ISSUE #4352: RTL: Toolbar arrow key behavior should be reversed
Right arrow should move to the logical previous toolbar button, right arrow should move to the logical next.

### LINKED ISSUE #4355: RTL: Item/collection tree arrow key expand/collapse behavior should be reversed
Left arrow should expand, right arrow should collapse. This is in line with other apps like Finder.

AbeJellinek:
Fixes #4352, fixes #4355, plus various other issues (collapsible-sections, quick format, and so on). Not touching the tag selector for now because it always displays LTR.

### DIFF
--- a/chrome/content/zotero/components/virtualized-table.jsx
+++ b/chrome/content/zotero/components/virtualized-table.jsx
@@ -630,7 +630,7 @@ class VirtualizedTable extends React.Component {
 		if (shiftSelect || moveFocused) return;
 		
 		switch (e.key) {
-		case "ArrowLeft":
+		case Zotero.arrowPreviousKey:
 			const parentIndex = this.props.getParentIndex(this.selection.focused);
 			if (this.props.isContainer(this.selection.focused)
 					&& !this.props.isContainerEmpty(this.selection.focused)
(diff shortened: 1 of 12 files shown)


dstillman:
Can we just call them `Zotero.arrowPreviousKey` and `Zotero.arrowNextKey`?

abaevbog:
@AbeJellinek this is great! Making keyboard nav work for rtl layout is long overdue.

The only thing I have noticed so far is once the cursor finds itself in the quickFormat input, it gets "stuck".

This is the expected behavior that does not happen in the rtl layout - then, the cursor does not leave the edge of the input on arrowRight/left keypresses.

https://github.com/zotero/zotero/assets/36271954/5ff3531b-de75-409d-bfdc-b2a15ca117aa

That specific part comes from the conditions like [this one](https://github.com/zotero/zotero/blob/8cb93b35a4fbc827f184a28a5be4ce08019d79c8/chrome/content/zotero/integration/quickFormat.js#L1883) where we decide what counts as the start/end of the input (e.g. in rtl, `this.selectionStart` should be not 0 but `this.value.length`).

Granted, this doesn't work that well on current `main` - the cursor doesn't get stuck but it shifts focus to the bubble on the wrong key press from the wrong edge. I noticed some other janky behaviors in the rtl layout of QF (e.g. adding inputs between bubbles on click is off, backspace deletes bubbles from the wrong edge, etc.) that are not really caused by these changes, so I am thinking to go over them separately, once these this is merged. I could fix the cursor getting stuck issue then as well.

PR #4363, opened 2024-07-09
Rename subsequent attachments, don't change title when renaming from parent metadata

AbeJellinek:
Addresses https://github.com/zotero/zotero/issues/4211#issuecomment-2217514461


dstillman:
Merged

Issue #4342, opened 2024-07-06
Cache tab content type and don't show white placeholder icon at startup


dstillman:
The fix to https://github.com/zotero/zotero/issues/4150 improved things a bit, but temporarily showing a white page icon is pretty noticeable and clumsy. We should 1) cache the content type of tabs in the tab data (if we're not already) so we know what icon to show immediately and 2) show the right icon instead of the white page icon. If we can't do that for some reason, we should leave space for the icon and only show an icon once the correct one is available.

PR #4367, opened 2024-07-09
Cache tabs' icons in data field

abaevbog:
So that after reload, we can display the correct icon sooner

Fixes: #4342


abaevbog:
For comparison, this is loading of tabs before:


https://github.com/zotero/zotero/assets/36271954/2183f2c2-f8be-4837-b41b-bc466a8bb633

abaevbog:
And this is with cache, after opening the app, closing it (to have the tabs data cached for the first time), and reopening.


https://github.com/zotero/zotero/assets/36271954/5ea79394-80d8-46e0-a085-c77abd29ba34

Issue #4368, opened 2024-07-10
React 18: Tag selector mouse scrolling is laggy


AbeJellinek:
React 17:

https://github.com/zotero/zotero/assets/1770299/e46578b6-476d-429c-a44b-e9a1ff33a744

React 18:

https://github.com/zotero/zotero/assets/1770299/c269c478-8c9d-4877-ab70-417d361d9909

React 18 feels like it's rendering at about half the frame rate, and the scroll distance per click of the scroll wheel is also about halved. The frame rate change doesn't quite come through on the screen recording, but it's very noticeable when scrolling.

I tried upgrading to the latest `react-virtualized` (one minor version newer), but that didn't change anything.

abaevbog:
Interesting. For me, mouse scrolling seemingly works without issues but I am noticing the lag if I scroll with `arrowUp`/`arrowDown` keys

abaevbog:
> Are you using a mouse with a scroll wheel or something like a Magic Mouse / trackpad?

Nope! Just the usual trackpad on mac. This is the scroll using mouse at first and the focusing and holding `arrowDown`. Somewhat hard to see but I think the difference is still noticeable

https://github.com/zotero/zotero/assets/36271954/0133f997-c9a3-479b-9008-da9fbe252ef5

abaevbog:
Gotcha! 

> React 18 feels like it's rendering at about half the frame rate

I did notice that the rendering pattern has changed a bit in react 18, which made me move around some awaiting for promises in the tag selector tests. Potentially, due to [automatic batching](https://react.dev/blog/2022/03/08/react-18-upgrade-guide#automatic-batching)?

It indicates that we can opt out of that new behavior by using `flushSync` - not sure if this is really the root of this issue but perhaps worth a shot.

abaevbog:
So I tried to track down where in `react-virtualized` scrolling is handled to see if the `flushSync` wrapper would help, and I think it did... At least, as far as holding `arrowDown` key for me.

Before doing anything else, @AbeJellinek, do you mind checking if the same tweak solves the lagginess for you? 

I went to `./node_modules/react-virtualized/dist/umd/react-virtualized.js` and wrapped `_this.setState` on line 1342 with `ReactDOM.flushSync`. So the block of lines 1342-1347 became

```
    _this.state.isScrolling || isScrollingChange(!0), ReactDOM.flushSync(()=>_this.setState({
        isScrolling: !0,
        scrollLeft: scrollLeft,
[...]

PR #4370, opened 2024-07-10
Patch react-virtualized to fix react 18 scroll lag

abaevbog:
React 18 introduced automatic batching which tries to avoid unnecessary re-rendering (https://react.dev/blog/2022/03/08/react-18-upgrade-guide#automatic-batching). In some components based on `react-virtualized` (e.g. tag selector), it leads to visible lagginess on scroll via keypress or mouse wheel (not trackpad). This patch wraps setState of the scroll handler of `react-virtualized` with `ReactDOM.flushSync`, which opts out of automatic batching.

Fixes: #4368

Make sure to run `npm run clean` before `./app/scripts/build_and_run -rd` to have `babel-worker.js` rerun.

### DIFF
--- a/js-build/babel-worker.js
+++ b/js-build/babel-worker.js
@@ -46,6 +46,13 @@ async function babelWorker(ev) {
 			transformed = contents.replace('scrollDiv = document.createElement("div")', 'scrollDiv = document.createElementNS("http://www.w3.org/1999/xhtml", "div")')
 				.replace('document.body.appendChild(scrollDiv)', 'document.documentElement.appendChild(scrollDiv)')
 				.replace('document.body.removeChild(scrollDiv)', 'document.documentElement.removeChild(scrollDiv)');
+			// React 18: wrap setState in onScroll handler with ReactDOM.flushSync to avoid
+			// automatic batching https://react.dev/blog/2022/03/08/react-18-upgrade-guide#automatic-batching
+			// which causes less frequent re-rendering and lagginess on scroll of components such as tag selector
+			let onScrollSetStateChunkRegex = /(_this\.state\.isScrolling\s*\|\|\s*isScrollingChange\(!0\),\s*)(_this\.setState\s*\(\s*\{[^}]*\}\s*\))/;
+			transformed = transformed.replace(onScrollSetStateChunkRegex, (_, p1, p2) => {
+				return `${p1}ReactDOM.flushSync(() => ${p2.trim()})`;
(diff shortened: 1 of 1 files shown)


dstillman:
Aren't we fine with this in other places, though? Should we not just override this in the components where it's actually a problem?

Or if this is the best option, we should check for the pattern we're replacing first and throw if it's not present (like we do in fetch_xulrunner) — otherwise this could regress silently on a React update.

dstillman:
Oh, yeah, that's fair.

OK, then I would just have it throw if the search pattern doesn't exist.

abaevbog:
Cool, how about this? Correct me if the error sounds a bit off - I tried to be brief but informative with the exact line that we're trying to find.

js-build/babel-worker.js
```diff
@@ -46,6 +46,16 @@ async function babelWorker(ev) {
...
 				.replace('document.body.removeChild(scrollDiv)', 'document.documentElement.removeChild(scrollDiv)');
+			// React 18: wrap setState in onScroll handler with ReactDOM.flushSync to avoid
+			// automatic batching https://react.dev/blog/2022/03/08/react-18-upgrade-guide#automatic-batching
+			// which causes less frequent re-rendering and lagginess on scroll of components such as tag selector
+			let onScrollSetStateChunkRegex = /(_this\.state\.isScrolling\s*\|\|\s*isScrollingChange\(!0\),\s*)(_this\.setState\s*\(\s*\{[^}]*\}\s*\))/;
+			if (!onScrollSetStateChunkRegex.test(transformed)) {
+				throw new Error(`"_this.state.isScrolling || isScrollingChange(!0), _this.setState({" not found in react-virtualized`);
+			}
+			transformed = transformed.replace(onScrollSetStateChunkRegex, (_, p1, p2) => {
+				return `${p1}ReactDOM.flushSync(() => ${p2.trim()})`;
```
dstillman, line 57: Already merged, but why is there a `trim()` here? p2 will always start with `_` and end with `)`, no?

abaevbog: You're right. When testing this out, I got thrown off by an extra new line `console.log`-ed during the build stage, so I added `.trim()` thinking it was needed for the `;` to fall at the right place. But, yeah, apologies, `.trim()` does not actually have any effect here.

abaevbog: I added #4379 to clean it up

PR #4391, opened 2024-07-15
vpat 48: announce selection for default virtualized table

abaevbog:
- set `aria-activedescendant` in `_onSelection` handler of virtualized table. For simpler virtualized tables (e.g. style manager in Zotero_Preferences.Cite), the table does not get re-rendered when selection changes, so `aria-activedescendant` does not get updated in `render()`, which means the screen reader is just quiet.
- construct and set aria-label for rows build via `makeRowRenderer` of VirtualizedTable so that is is announced when the row is selected

### DIFF
--- a/chrome/content/zotero/components/virtualized-table.jsx
+++ b/chrome/content/zotero/components/virtualized-table.jsx
@@ -256,6 +256,7 @@ class TreeSelection {
 	_updateTree(shouldDebounce) {
 		if (!this.selectEventsSuppressed && this._tree.props.onSelectionChange) {
 			this._tree.props.onSelectionChange(this, shouldDebounce);
+			this._tree._setAriaAciveDescendant();
 		}
 	}
 
@@ -811,11 +812,6 @@ class VirtualizedTable extends React.Component {
 		else {
 			return;
 		}
-		// Set aria-activedescendant on table container in case if render() is not called after selection change
-		let selected = this._jsWindow?.getElementByIndex(this.selection.focused);
-		if (selected) {
-			selected.closest(".virtualized-table").setAttribute("aria-activedescendant", selected.id);
-		}
 
 		this.scrollToRow(index);
 	}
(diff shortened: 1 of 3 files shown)


abaevbog:
For reference, this is the current behavior on `main`:

https://github.com/user-attachments/assets/ffbd1b07-a163-4b5d-a664-5aa8db0e7f95

And this is after this update:

https://github.com/user-attachments/assets/4eba16c0-57aa-436f-869e-266328e29118

abaevbog:
@adomasven this should be a simple fix to address the lack of announcements by screen readers of selected row in some virtualized tables. Let me know what you think!

chrome/content/zotero/components/virtualized-table.jsx
```diff
@@ -811,6 +811,11 @@ class VirtualizedTable extends React.Component {
 		else {
 			return;
 		}
+		// Set aria-activedescendant on table container in case if render() is not called after selection change
+		let selected = this._jsWindow?.getElementByIndex(this.selection.focused);
+		if (selected) {
+			selected.closest(".virtualized-table").setAttribute("aria-activedescendant", selected.id);
+		}
```
adomasven, line 818: Ideally something like this would be controlled in a single place in code. Can we remove the code which sets `aria-activedescendant` in `render()` and just do it here, with a comment which says why it's done this way?

abaevbog: That makes sense. The only reason why I did not touch that `render` is because when the table is just rendered, `_onSelection` doesn't seem to be called, so that piece in `render` was responsible for setting `aria-activedescendant ` on `collectionTree` the very first time after restart.

But it looks like `_updateTree` of the tree selection is called both during the initial selection and on all further updates after. So I moved `aria-activedescendant` into a separate function and called it in `_updateTree` and it seems to behave as expected. Does that seem reasonable?

I also removed `this.forceUpdate` from selection handlers of item and collection trees because the comments made it look like updating `aria-activedescendant` is the only reason for those calls. Correct me if that's wrong

PR #4400, opened 2024-07-17
vpat 59: make quickFormat reference box separator text darker

abaevbog:
So that on windows, the separator text on white background has a high enough contrast ratio for accessibility.

Old `gray` color (`rgb(128, 128, 128)` or `#808080`) for separator text has a contrast ratio with the white background of reference box of ~3.94, which is below all minimum requirements.

This makes separator text a bit darker `rgb(87, 87, 87)` or `#575757` to make sure there is more contrast.

Tested via: https://webaim.org/resources/contrastchecker/

Before:

Screenshot 2024-07-16 at 5 41 13 PM

After:

Screenshot 2024-07-16 at 5 40 20 PM

### DIFF
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -1438,6 +1438,9 @@ var Zotero_QuickFormat = new function () {
 		// Try to make the panel appear right in the center on windows
 		let leftMargin = Zotero.isWin ? 5 : 15;
 		referencePanel.openPopup(dialog, "after_start", leftMargin, 0, false, false, null);
+		// Initially, panel has an opacity of 0.9 to not display the shadow behind it but we override
+		// it upon the first opening for the panel to not look transparent on windows
+		referencePanel.style.opacity = "1";
 	}
 	
 	/**
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -1438,8 +1438,8 @@ var Zotero_QuickFormat = new function () {
 		// Try to make the panel appear right in the center on windows
 		let leftMargin = Zotero.isWin ? 5 : 15;
 		referencePanel.openPopup(dialog, "after_start", leftMargin, 0, false, false, null);
-		// Initially, panel has an opacity of 0.9 to not display the shadow behind it but we override
-		// it upon the first opening for the panel to not look transparent on windows
+		// Initially, panel has an opacity of 0.9 to prevent a shadow from appearing behind the
+		// panel on Windows, but we override it upon the first opening of the panel
 		referencePanel.style.opacity = "1";
 	}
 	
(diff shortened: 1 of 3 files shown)


dstillman:
My main question would be: why is there translucency on that panel? I'm not seeing that on macOS.

abaevbog:
It comes from [here](https://github.com/zotero/zotero/blob/38391f356c181d4e561c9cddf443d795ce92fac0/chrome/skin/default/zotero/integration.css#L257) which was set 13 years ago, so my understanding is that it has always been that way.

It applies to macOS the same way as to windows but instead of making the entire panel transparent, macOS makes it becomes somewhat paler. At `0.9` it's almost not noticeable but, for comparison, this is what setting `opacity: 0.5` would look like: 

Screenshot 2024-07-18 at 11 04 33 PM

Would it make sense to just drop the opacity? This is what it would look like on windows:
Screenshot 2024-07-18 at 11 12 16 PM

dstillman:
The commit that added that was the initial commit of this citation dialog, which also said that it "has not been tested on any platform but OS X", so I think it's safe to say that that translucency on Windows was never intentional. We should definitely get rid of it on Windows (and possibly on Linux, depending what it does there).

yexingsha:
We should just drop the opacity on all platforms, even on macOS it's not serving any particular purpose.

I think the updated color is too dark for secondary text. Do we need to meet AAA standards?

Either way, I suggest using `var(--fill-secondary)` here to keep the styling consistent. To meet AA standards, we need to update `fill-secondary` to `#0000008c` so that it has a contrast ratio of 4.74 on white and 4.62 on `material-sidepane`. If we have to meet AAA standards, then it needs to go up to `000000AA`, which would make info look like this:

Screenshot 2024-07-19 at 13 09 04

I think this makes the secondary text too dark to maintain visual hierarchy.

abaevbog:
So I dropped the opacity and there is something to note here. On windows, I noticed that with `opacity: 0.9` removed, there is a shadow appearing below and to the right of the panel. See below:

Screenshot 2024-07-19 at 11 06 41 AM

I assumed we don't want it but I'm not exactly sure where it comes from since there is no `box-shadow` or anything similar anywhere. Perhaps it is added by OS itself? I noticed that if we keep the original opacity in the stylesheet but set `opacity: 1` when the panel opens then we end up having a panel without that shadow and that is not transparent.  Correct me if I overthought it and the shadow is not a problem, in which case it's better to tweak the stylesheet. This is what it looks like now:

Screenshot 2024-07-19 at 11 04 44 AM

On my ubuntu, the panel was transparent as well and removing the opacity in this way works too.

abaevbog:
@yexingsha AA is a minimum that we need to hit so `#0000008c` as `fill-secondary` works just fine, in my opinion. This actually seems to clear all the contrast warnings that we had before, so that's very nice!

PR #4402, opened 2024-07-17
vpat 80: focus first field in search window/dialog

abaevbog:
Instead of focusing the first condition, which ends up placing focus in the middle of the new window or modal. It is best to focus the first node, which is the general convention, per https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/, and what many screen reader users expect. When we focus a field in the middle of the window/dialog, it makes it seem for some users that this is the first node, which will make them miss all the fields above it.

Also, linked search name input to its label in searchDialog.xhtml.

For reference, comment from the vpat issue:

"Impact Statement: When focus is placed in the middle of a modal, verses being placed at the beginning of the content, it is likely that fields may be overlooked. Additionally, this may cause frustration as the user must backtrack to the beginning of the modal. Those who rely upon assistive technology may not see that they are placed in the middle unless exploring the search feature before hand."

### DIFF
--- a/chrome/content/zotero/advancedSearch.js
+++ b/chrome/content/zotero/advancedSearch.js
@@ -91,7 +91,7 @@ var ZoteroAdvancedSearch = new function() {
 
 		this.itemsView.changeCollectionTreeRow(collectionTreeRow);
 		// Focus the first field in the window
-		Services.focus.moveFocus(window, sbc, Services.focus.MOVEFOCUS_FORWARD, 0);
+		Services.focus.moveFocus(window, null, Services.focus.MOVEFOCUS_FORWARD, 0);
 	}
 	
 	this.onUnload = function () {
--- a/chrome/content/zotero/advancedSearch.js
+++ b/chrome/content/zotero/advancedSearch.js
@@ -90,6 +90,8 @@ var ZoteroAdvancedSearch = new function() {
 		};
 
 		this.itemsView.changeCollectionTreeRow(collectionTreeRow);
+		// Focus the first field in the window
+		Services.focus.moveFocus(window, sbc, Services.focus.MOVEFOCUS_FORWARD, 0);
 	}
 	
 	this.onUnload = function () {
--- a/chrome/content/zotero/searchDialog.js
+++ b/chrome/content/zotero/searchDialog.js
@@ -43,6 +43,8 @@ function doLoad()
 	document.getElementById('search-name').value = io.dataIn.name;
 
 	document.addEventListener('dialogaccept', doAccept);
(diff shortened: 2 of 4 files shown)


chrome/content/zotero/searchDialog.js
```diff
@@ -43,6 +43,8 @@ function doLoad()
 	document.getElementById('search-name').value = io.dataIn.name;
 
 	document.addEventListener('dialogaccept', doAccept);
+	// Focus the first focusable node in the dialog
+	Services.focus.moveFocus(window, sbc, Services.focus.MOVEFOCUS_FORWARD, 0);
```
dstillman, line 47: This line doesn't seem necessary? The Name input field gets focused for me automatically when editing a saved search, at least on macOS.

It would be better for the text to get selected, though.

dstillman (changes requested):

chrome/content/zotero/searchDialog.js
abaevbog: Good point! It does look like the first focusable field in dialogs gets selected (though I can't easily identify where exactly that happens). In any case, I swapped explicit focusing of the name field for selecting all text

chrome/content/zotero/advancedSearch.js
```diff
@@ -90,6 +90,8 @@ var ZoteroAdvancedSearch = new function() {
 		};
 
 		this.itemsView.changeCollectionTreeRow(collectionTreeRow);
+		// Focus the first field in the window
+		Services.focus.moveFocus(window, sbc, Services.focus.MOVEFOCUS_FORWARD, 0);
```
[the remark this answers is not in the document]
abaevbog: You're right - I changed it for `Services.focus.moveFocus(window, null, Services.focus.MOVEFOCUS_FORWARD, 0);`

chrome/content/zotero/searchDialog.js
```diff
@@ -40,7 +40,9 @@ function doLoad()
 	var searchBox = document.getElementById('search-box');
 	searchBox.groups = io.dataIn.groups;
 	searchBox.search = io.dataIn.search;
-	document.getElementById('search-name').value = io.dataIn.name;
+	let searchName = document.getElementById('search-name');
+	searchName.value = io.dataIn.name;
+	searchName.setSelectionRange(0, searchName.value.length, "backward");
```
dstillman, line 45: `searchName.select()`?

abaevbog: Oh cool, I didn't realize that `.select` works just as well as `.setSelectionRange`. Good to know!

Issue #4382, opened 2024-07-12
Scroll selected tab into view on startup


dstillman:
https://forums.zotero.org/discussion/115900/z7-beta-tab-list-doesnt-scroll-to-active-tab-on-startup

Or on "Move to End", as in the other linked thread

abaevbog:
This looks like a post-react 18 bug to me. There are other reports that I could reproduce of the tab not being scrolled to, not only on startup: https://forums.zotero.org/discussion/116025/zotero-7-beta-autofocus-on-the-active-tab-when-opening-a-new-tab-after-many-open-tabs

It looks like it is due to somewhat different rendering mechanics (batching, maybe?) in react-18 which causes the [node](https://github.com/zotero/zotero/blob/a46f1bcbbb25b66745e3d8a391b529055621efbc/chrome/content/zotero/tabs.js#L484) of the tab not exist in `select` when we expect it.

I suppose we should either properly wait for the tab node to appear (just skipping a tick before finding the node is not reliable) or maybe move the logic of scrolling to the right node into `useEffect` of the tab bar.

PR #4404, opened 2024-07-18
Fix tab not scrolling into view

abaevbog:
After react 18 update, the tab node may not yet be rendered by tabBar.jsx when we try to scroll it into view in Zotero_Tabs.select.

- To make sure scrolling happens when rendering is done, move scroll-related logic into a `useEffect` of tabBar.jsx. It also makes sure that we'll scroll to a selected tab if it is moved via context menu to the very beginning or the end.

- Added a small scroll-padding to tabs' container to make sure the border does not get cutoff after the tab is scrolled into view. This is instead of JS code accounting for the border.

- Fixed a glitch where the pinned library tab would not get selected on shift-tab from opened tabs menu.

Fixes: #4382

### DIFF
--- a/chrome/content/zotero/tabs.js
+++ b/chrome/content/zotero/tabs.js
@@ -493,24 +493,6 @@ var Zotero_Tabs = new function () {
 				tabNode.focus();
 			}
 		}
-		// Allow React to create a tab node
-		setTimeout(() => {
-			tabNode.scrollIntoView({ behavior: 'smooth' });
-		});
-		// Border is not included when scrolling element node into view, therefore we do it manually.
-		// TODO: `scroll-padding` since Firefox 68 can probably be used instead
-		setTimeout(() => {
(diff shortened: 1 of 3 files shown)


chrome/content/zotero/tabs.js
```diff
@@ -633,14 +615,12 @@ var Zotero_Tabs = new function () {
...
+			// Go through all candidates and try to focus the visible one
 			let candidates = document.querySelectorAll(`[data-id="${nextTab.id}"]`);
 			for (let node of candidates) {
-				if (node.offsetParent) {
-					node.focus();
-					return;
-				}
+				node.focus();
+				// Visible tab was found and focused
+				if (document.activeElement == node) return;
```
dstillman, line 623: Can you explain this? The idea here is that if a tab is off-screen, it doesn't become the active element?

PR #4405, opened 2024-07-18
Restore autoscroll support in reader

### LINKED ISSUE #4384: Reader: Auto-scroll with middle mouse button not working
https://forums.zotero.org/discussion/107529/7-0-beta-middle-click-not-working

I haven't tried, but a lot of reports. In theory @AbeJellinek fixed this in a7f1235fd18905f6b44a3ca1e8c1b4521d8be62d.

AbeJellinek:
- Mozilla's `ActorManagerParent` was only being imported when devtools were initialized (either by opening the Error Console or by connecting via remote debugging). Import it in our `ActorManager` so Mozilla actors, including `AutoScroll`, are always registered.
   - This may have other effects! I'm assuming most of us test with the error console visible, though, so we would've had Mozilla actors, while users running without devtools wouldn't have. We definitely want those actors. (Most of them, at least.)
- Don't override autoscroller styles with our panel customizations.

Fixes #4384

### DIFF
--- a/chrome/content/zotero/actors/ActorManager.jsm
+++ b/chrome/content/zotero/actors/ActorManager.jsm
@@ -1,5 +1,8 @@
 var EXPORTED_SYMBOLS = [];
 
+// Register Mozilla actors
+ChromeUtils.importESModule("resource://gre/modules/ActorManagerParent.sys.mjs");
+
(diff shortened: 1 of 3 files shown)


dstillman:
Good find!

Issue #4407, opened 2024-07-19
vpat 45: edit bibliography dialog and voiceover


abaevbog:
VPAT 45 indicates that there is not enough clarity how to use the "Edit bibliography" dialog, suggesting to change focus order or to add instructions. I think the focus order is not an issue (as the issue description indicates a misunderstanding of what itemTree and collectionTree are for) but it's probably fair to say that we need some additional instructions on how to use the dialog.

Ideally, we would have a general way to announce arbitrary messages to screen readers to provide users with such information. It's trivial for NVDA and JAWS by using `aria-live` regions but voiceover doesn't handle `aria-live` from, what looks like, XUL windows. It means that such announcements works just fine in the reader but not in the library tab, edit bibliography dialog, or elsewhere.

Having tried to find a workaround for voiceover, I ran into the [bug report](https://bugzilla.mozilla.org/show_bug.cgi?id=1733266) that seems to boil down to exactly the same issue.

So my thinking is that until this is fixed, we can do a similar thing as with QuickFormat dialog and just add `aria-description` properties for all components to indicate what is their purpose. It's not as neat and it will be trickier to put together sentences that will make sense on each component separately but it might be the best solution for now.

VPAT 45 for the record: 

---

Description: In the Zotero Word Plugin's 'Edit Bibliography' modal, screen reader users must explore the dialog to discover that they need to press the Tab key approximately three times to reach the list of editable bibliographies. Upon opening the modal, focus is initially placed on an edit field called "All Fields & Tags." This edit field only filters results in the first list of citations and does not allow for editing. Its purpose in this context is unclear, as it presents citations not yet included in the document. To access the list of cited materials that can be edited, users must press the Tab key three times. After selecting a bibliography to edit using the arrow keys, users need to press the Tab key a few more times to reach the edit field where changes can be made. The focus order in this modal should be revised to initially place focus on the list of editable bibliographies, bypassing the extraneous collection tree view and item list. 

Impact Statement: Users relying on assistive technologies would benefit from clearer instructions on how to operate dialogs like this; the current focus order and lack of guidance leads to confusion and hinders the completion of tasks. 

Expected Result: When choosing to edit a bibliography, focus should be placed on the list of editable bibliographies upon opening the modal, allowing for immediate interaction using arrow keys to select a bibliography for editing. If controlling the focus order or eliminating the extra fields is not feasible, providing textual instructions to guide users to the editable bibliographies using the Tab key would help reduce confusion. 

[...]

PR #4413, opened 2024-07-19
vpat 45: bibliography dialog instructions and aria

abaevbog:
- made the wrapper of the dialog focusable and adde aria-description so that when the dialog opens it announces how to use the dialog
- added aria-labels to buttons, references list and editor
- make sure that the references list always selects an item when focused without clicking
- added a border to focused buttons from simple editor to make it visually clear when they received focus

Addresses #4407 

--- 

The vpat issue indicated that it is not clear how to use the dialog and suggested to focus the references list when the dialog opens or to add instructions. See the full text of the vpat comment in linked issue. It seems like there was a misunderstanding how to use the dialog (since they refer to collection and itemTree as "extraneous") so I though adding instructions like this is more appropriate. Besides, there is vpat 80 addressed in #4402 that states that it's bad to focus a node in the middle of a window since then users can miss out on the components before it.

Finally, I had an idea to setup a generalized way to announce messages arbitrarily from JS using `aria-live` (instead of making the container focusable and attaching description to it) but it seems to not properly work on macOS outside of web html (in firefox as well), so that probably has to wait.

### DIFF
--- a/chrome/content/zotero/integration/editBibliographyDialog.js
+++ b/chrome/content/zotero/integration/editBibliographyDialog.js
@@ -118,6 +118,13 @@ var Zotero_Bibliography_Dialog = new function () {
 		_updatePreview();
 	}
 	
+	// When reference list is focused and nothing is selected, select the first item
+	this.listItemFocused = function () {
+		if (!_itemList.selectedItems.length) {
+			_itemList.selectedIndex = 0;
+		}
+	};
+
 	/**
 	 * Called when an item in the reference list is clicked
 	 */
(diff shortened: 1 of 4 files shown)


abaevbog:
This approach of having a focusable wrapper with `aria-description` works fine but it is unfortunate that in that case we have an extra tabstop which makes it inconsistent with other similar dialogs.

One thing I tried while working on JAWS not reading the preferences page (linked commit) is to load the whole `editBibliographyDialog` into a `<browser>` of a wrapper window, in which case we can announce the instructions on load (and maybe on a special key-press) regardless of where the focus is with an `aria-live` component. It might be a nicer alternative - not sure if it's necessarily worth it though

dstillman:
Merged in c0c3d93019fdfb6cb96f3a11e6edbdfd6aa5addf

I couldn't push changes to this old PR due to Git LFS, but I changed `btn` → `button` (no need to abbreviate) and tweaked the text.

PR #4414, opened 2024-07-19
vpat 25: idea to make prefs visible to jaws cursor?

abaevbog:
VPAT 25 issue boils down to important text in the preferences (e.g. Export > > QuickCopy notes at the top) not being visible for some screen readers' (mainly nvda/jaws) cursors, which means they are not currently readable. Similarly, we have vpat 46 which mentions that sections do not get announced by JAWS.

The issue is solved easily for NVDA by adding `role="document"` to the container of preferences content, after which it can switch between browse and focus mode to, if needed, read the text. This is an [example](https://www.dropbox.com/scl/fi/ed7n4otsndlacacm9u8fr/1_nvda_working.mov?rlkey=kinyohqb6bd8b71hdisbounsq&st=jzfh0npk&dl=0).

Unfortunately, it does not help JAWS (this is JAWS with the first commit only with unchanged behavior: [recording](https://www.dropbox.com/scl/fi/2ieqepljdnmo67x9jhmmq/1_jaws_not_working.mov?rlkey=frum5dsxsvmti894n5som9cdt&st=9j5b9i1l&dl=0)). Because in some cases JAWS is configured to not use the virtual cursor, I am guessing the root of the issue is similar to what we had with the window class in #3930, where we have to fit into what JAWS is set up to do. This issue only applies to XUL windows - JAWS has no problems, for instance, in reader tabs loaded into `<browser>`. So as a workaround for this (and potentially other issues) I thought what if we load the preferences.xhtml into a browser of a new window, similar to what we do with reader tabs? I believe this is what firefox and thunderbird do as well.

This is just a draft in which JAWS works ([recording](https://www.dropbox.com/scl/fi/u1941qhyuqogtbp1gqy48/2_jaws_working.mov?rlkey=jii35f5ahstquiuy6o5hl04go&st=umaf0909&dl=0)) as expected thought there obviously is some breakage with how the window looks

Before continuing down this road, I wanted to check if there is any reason to not do this that I may not know? Is there anything specific to keep in mind before going further?

There are alternative workarounds to getting JAWS to read the text in the settings (as well as other places where this issue may be relevant) but none of the ones I thought of seem as good. We could link all text nodes as `aria-describedby` to some of the focusable nodes but there may not be a great candidate to link the text to (e.g. quick format instructions). For section headers, we could make the entire section focusable and add them as `aria-labelledby` to the actual section (which would add a bunch of extra  tab stops).

@dstillman what do you think about all this? It could be that I am really overthinking it and a adding a few extra tabstops by making each section focusable, adding section headers as `aria-label`s (for jaws), and linking text to relevant nodes (or section if no checkboxes or inputs apply) with `aria-describedby` is not as bad as it seemed to me at first and we should do just that (edit: though this does leave it up to JAWS to read all `aria-describedby` which, as I just tried, is not very reliable). But I was also thinking that since JAWS still struggles with inputs particularly in dialogs (#4205) maybe loading window content through the `<browser>` is not a bad idea to explore (again, since firefox and thunderbird do it almost everywhere) in case we want to apply it elsewhere.

### DIFF
--- a/chrome/content/zotero/browser_wrapper.xhtml
+++ b/chrome/content/zotero/browser_wrapper.xhtml
@@ -8,32 +8,39 @@
 		width="800"
 		height="600"
 		style="display: flex;">
-	
-	<browser style="width:100%;height:100%"/>
-
 	<script>
-		let browser = document.querySelector("browser");
-		
-		browser.addEventListener("DOMContentLoaded", (e) => {
-			if (!browser.getAttribute("src")) return;
-			// When the content of browser is loaded, pass over arguments and copy its title
-			browser.contentWindow.arguments = window.arguments;
-			document.title = browser.contentDocument.title;
-			// Set the same min-width/min-height as the browser's content
-			let dialogStyle = browser.contentWindow.getComputedStyle(browser.contentDocument.querySelector("window"));
-			if (dialogStyle.getPropertyValue("min-width")) {
-				document.querySelector("window").style.minWidth = dialogStyle.getPropertyValue("min-width");
+		var init = (url, options) => {
+			let browser = document.createXULElement("browser");
+			if (!options.noType) {
+				browser.setAttribute("type", "content");
 			}
-			if (dialogStyle.getPropertyValue("min-height")) {
-				document.querySelector("window").style.minHeight = dialogStyle.getPropertyValue("min-height");
-			}
-		}, true);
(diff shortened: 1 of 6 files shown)


abaevbog:
Closing this as it seems easier and safer to just make all relevant components in the preferences window focusable 
per https://github.com/zotero/zotero/issues/3930#issuecomment-2251045406

Issue #3970, opened 2024-04-12
vpat 16-17: mouse actions alternative for virtualized table


abaevbog:
We want all functionality to be accessible via a keyboard and these things currently cannot be done in `itemTree` or `collectionTree` without using a mouse:
1. one can't set the primary sorting/resize/reorder/add/remove columns without right-clicking the header or dragging the rows
2. one can't move a collection into another collection or move a child note/attachment to be a child of another parent item without dragging

Similar problems are already solved with our context menus. For example, as an alternative to dragging an item to a collection, there is a "Add to collection" menu option. I think it would make sense to expand on this approach to make everything above available through the context menus.

In this case, the context menu of `itemTree` would need to be updated to allow one to not only add/remove but also resize/reorder/sort. Maybe each column in the context menu could become a submenu, so that one could, for example, do `contextMenu > creator > make wider/shorter/move right/move left`. This would be somewhat close to what happens with reordering of creators in `itemBox`.

In a similar fashion, we can add a "Move to collection" option to the context menu in `collectionTree` to allow collections to be moved into another collection/group without dragging. The question of avoiding drag-drop for to move a row as a child into another row in `itemTree` is a bit trickier, since we would need to list all available target items as `menuitem`s, and there theoretically could be thousands ... Maybe we could filter the items by some criteria (e.g. only use currently rendered items by virtualized table)?

One final problem to note is that there is no way to invoke the context menu to manipulate columns in `itemTree` without right-clicking on it, since the header is not focusable or selectable. Maybe we add a shortcut for this? As much as I personally dislike shortcuts, I can't think of any better way to expose this functionality without making the header selectable via keyboard which, I think, would make the interface of `itemTree` unnecessarily complex. Alternatively, we could add a small icon/button somewhere in the corner of the header that would be focusable but I'm not sure if it would look nice. 

These are some of my initial thoughts on how to address this. @dstillman @adomasven - what do you think?

adomasven:
> one can't set the primary sorting/resize/reorder/add/remove columns without right-clicking the header or dragging the rows

Column sorting and visibility we can and should support via the View menu. Although we don't have a menubar in other views where the item tree is displayed (advanced search, citation dialog, select items). Not as big an issue on macOS with global menubars, but on Windows and Linux we might need to explore having a hiding menubar only visible on pressing Alt.

[...]

Issue #1730, opened 2019-09-06
Duplicate Collection feature


dstillman:
https://forums.zotero.org/discussion/78890/advanced-collection-structure-management-duplicate-structure

dstillman:
https://forums.zotero.org/discussion/comment/395692#Comment_395692

PR #4420, opened 2024-07-23
vpat 16: context menu as a drag-drop alternative to move collections

abaevbog:
- added context menu to move a collection into another collection. Since one can drag a collection into another library, the top-level entries of the contextmenu are libraries and not collections as with itemTree (shouldn't the same be true for itemTree in fact? You can drag an item to a collection in another group but I don't think it's possible to do it with the context menu).
- refactoring of collectionTree to pull out `copyCollections()` and `copyItems()` from out of the scope of `CollectionTree.onDrop`. If a collection is moved into another library, the functionality of copying over a collection needed to be exposed to ZoteroPane without the actual drag-drop happening. `copyCollection` and `copyItems` are unchanged except for the added parameters. We could avoid the refactoring by maybe faking a `drop` event from `ZoteroPane.moveToCollection` but I thought this would be cleaner and potentially useful for in the future.

Screenshot 2024-07-22 at 11 19 14 PM

Addresses part of #3970

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -4091,13 +4091,38 @@ var ZoteroPane = new function()
 
 	this.buildCopyCollectionMenu = function (event) {
 		if (event.target !== event.currentTarget) return;
-		let popup = event.target;
+		let popup = document.getElementById("zotero-copy-collection-popup");
 		popup.replaceChildren();
 		let selected = this.getSelectedCollection();
 
 		// Fetch all libraries
 		let topLevelEntries = Zotero.Libraries.getAll().filter(lib => !(lib instanceof Zotero.Feed));
 
+		// Check which libraries have collections linked to the selected collection
+		// and disable their menuitems. Same logic as in CollectionTree.canDropCheckAsync.
+		let linkedCollectionsExist = {};
+		(async () => {
+			for (let library of Zotero.Libraries.getAll()) {
+				if (library.libraryID == selected.libraryID || library instanceof Zotero.Feed) continue;
+				// Check which library has a collection linked to the selected collection
+				let linkedCollection = await selected.getLinkedCollection(library.libraryID, true);
+				linkedCollectionsExist[library.libraryID] = linkedCollection;
+				// Also check which library has collections linked to a subcollection of the selected collection
+				for (let descendent of selected.getDescendents(false, 'collection')) {
+					let subcollection = Zotero.Collections.get(descendent.id);
+					let linkedSubcollection = await subcollection.getLinkedCollection(library.libraryID, true);
+					linkedCollectionsExist[library.libraryID] = linkedSubcollection || linkedCollectionsExist[library.libraryID];
+				}
+			}
+			// Libraries that have linked collections have their menus disabled
+			for (let libraryMenuItem of [...popup.childNodes]) {
+				let menuItemLibID = libraryMenuItem.getAttribute("value").substring(1);
+				if (linkedCollectionsExist[parseInt(menuItemLibID)]) {
+					libraryMenuItem.disabled = true;
+				}
+			}
+		})();
+		
 		// If there is only one library, display its collections as top-level menuitems
 		if (topLevelEntries.length == 1) {
 			// Manually add My Library menuitem at the top, so one can still copy into it
@@ -4135,7 +4160,8 @@ var ZoteroPane = new function()
 				(target) => {
 					// can't copy collection into itself or into non-editable groups
 					return selected == target
-						|| (target instanceof Zotero.Group && !target.editable);
+						|| (target instanceof Zotero.Group && !target.editable)
+						|| linkedCollectionsExist[target.libraryID];
 				}
 			);
 			popup.append(menuItem);
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -2566,42 +2566,45 @@ var ZoteroPane = new function()
 	});
 
 	// Move selected collection to specified target collection.
-	// If target is in a different library, this collections is copied over.
-	// Essentially, a replication of drag-drop mechanism of collections from CollectionTree.onDrop.
-	this.moveToCollection = async (target) => {
+	this.moveCollection = async (target) => {
 		let selected = this.getSelectedCollection();
 		if (!selected) return;
-		
-		if (target.libraryID == selected.libraryID) {
-			// Target is in the same library - just update parent of the collection
-			selected.parentID = target.id;
-			await selected.saveTx();
+
+		if (target.libraryID !== selected.libraryID) {
+			throw new Error("Moving collections is only possible within the same library.");
 		}
-		else {
-			// Target is in another library - copy the collection over
-			let targetTreeRowID = `L${target.libraryID}`;
-			if (target instanceof Zotero.Collection) {
-				targetTreeRowID = `C${target.id}`;
-				// Make sure the row is actually visible
-				await ZoteroPane.collectionsView.expandToCollection(target.id);
-			}
-			let targetTreeRowIndex = ZoteroPane.collectionsView.getRowIndexByID(targetTreeRowID);
-			let targetTreeRow = ZoteroPane.collectionsView.getRow(targetTreeRowIndex);
-			let copyOptions = {
-				tags: Zotero.Prefs.get('groups.copyTags'),
-				childNotes: Zotero.Prefs.get('groups.copyChildNotes'),
-				childLinks: Zotero.Prefs.get('groups.copyChildLinks'),
-				childFileAttachments: Zotero.Prefs.get('groups.copyChildFileAttachments'),
-				annotations: Zotero.Prefs.get('groups.copyAnnotations'),
-			};
-			ZoteroPane.collectionsView.executeCollectionCopy({
-				collection: selected,
-				targetCollectionID: target.id,
-				targetLibraryID: target.libraryID,
-				targetTreeRow,
-				copyOptions
-			});
+		selected.parentID = target.id;
+		await selected.saveTx();
+	};
+
+	// Copy selected collection into another collection or library.
+	// Partially, a replication of drag-drop mechanism from CollectionTree.onDrop.
+	this.copyCollection = async (target) => {
+		let selected = this.getSelectedCollection();
+		if (!selected) return;
+
+		let targetTreeRowID = `L${target.libraryID}`;
+		if (target instanceof Zotero.Collection) {
+			targetTreeRowID = `C${target.id}`;
+			// Make sure the row is actually visible
+			await ZoteroPane.collectionsView.expandToCollection(target.id);
 		}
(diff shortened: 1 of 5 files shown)


abaevbog:
> Maybe “Change Parent”? We’d want to title-case it like other context menu items in any case.

That is a good point! I'll have the case fixed. One question on "Change Parent": if you move the collection to a different library, it doesn't technically "change" the parent but rather creates a copy in another group. Would "Change Parent" not be a bit confusing because it implies that the collection should be gone from the old library? "Move to another collection" is actually no better (maybe even more misleading) in this sense. Maybe I'm really overthinking this though

abaevbog:
> This could get complicated, but I guess we could put all the other libraries under a submenu labeled “Other Libraries (As Copy)”. Maybe that’s weirder.

Thanks to `Zotero.Utilities.Internal.createMenuForTarget`, it seems be be quite straightforward actually! I think this looks nicer and more informative

Screenshot 2024-07-26 at 5 06 17 PM

abaevbog:
Right, sorry! Seems like I need to take a break now that I am confusing things

dstillman:
1. This doesn't technically do anything for accessibility on macOS until we do https://github.com/zotero/zotero/issues/4724 for the collection tree.
2. Instead of "Change Parent" and a super-awkward "Other Collections (As Copy)", how about Move To and Copy To? Move To would be limited to the current library, and Copy To could show libraries at the top level if there were any other than My Library.

(Copy To would also technically get us https://github.com/zotero/zotero/issues/1730. That might still be worth its own menu option, since navigating through the hierarchy could be annoying, but I don't know how common an operation duplicating a collection would be.)

abaevbog:
> Instead of "Change Parent" and a super-awkward "Other Collections (As Copy)", how about Move To and Copy To? Move To would be limited to the current library, and Copy To could show libraries at the top level if there were any other than My Library.

That sounds way better! I also agree that "Copy to" deserves its own independent menu option: I think it would look strange if "Copy to" was a submenu of "Move to".

How does this behavior look?

https://github.com/user-attachments/assets/850e400d-2875-4e2f-9b7e-fbaee05bb7b8

abaevbog:
Hm, I can't seem to reproduce it yet...

This is the setup, right?

https://github.com/user-attachments/assets/42d60575-6eef-4f38-a838-a29b8056d4d4

dstillman:
Yeah, that's the setup. It's listing the same key each time, and it's the collection with collectionID 1 in my DB. Something being cast incorrectly?

dstillman:
Or difference between libraryID 1 and collectionID 1?

abaevbog:
I believe this is what was wrong: `target.id` instead of something like `targetCollectionID: target instanceof Zotero.Collection ? target.id : null,` in the call to `ZoteroPane.collectionsView.executeCollectionCopy`.  I don't quite understand why it would not error on me because it does look like it should have... So I never could reproduce the issue, though the added test did fail without changing that line and it is passing now. Does it fix it for you?

dstillman:
That's working now, yes. I'd guess you weren't seeing it because you don't have a collection with collectionID 1.

abaevbog:
You are right, I just checked and I have no collection with id == 1.

abaevbog:
I rebased to clean up the tests merge conflicts

dstillman:
For "Move To", for a top-level collection, the current library should probably also be disabled, the same way that the parent collection is disabled for a subcollection.

dstillman:
This isn't checking destination editability (from `canDropCheck()`), nor does it look like it's checking things from `canDropCheckAsync()`.

We can just duplicate those things in ZoteroPane in order to get this merged, but later, we should probably refactor all of this copying code to get it out of the collection tree — maybe `Zotero.Collection::canCopy()`, `Zotero.Collection::canCopyAsync()`, and `Zotero.Collection::copy()`. (We already have `Zotero.Item::moveToLibrary()`, with fewer options.) And then for drag and drop, we'd just call those things from `canDropCheck()`/`canDropCheckAsync()` for `dataType == 'zotero/collection'`.

abaevbog:
> This isn't checking destination editability (from canDropCheck()), nor does it look like it's checking things from canDropCheckAsync().

Oh, right. Thank you for pointing it out. As far as I can see, there are three things that collectionTree checks for in `canDropCheck` and `canDropCheckAsync`:
1. Check if the group is editable: https://github.com/zotero/zotero/blob/c14896a64089ed3aed0d8bf04d1e9b3d24e4e3e4/chrome/content/zotero/collectionTree.jsx#L1569-L1572 I just added a small tweak for that, so targets that are non-editable groups will be disabled, so that should be taken care of.
3. Do not allow dropping a collection into itself or its children in canDropCheck: https://github.com/zotero/zotero/blob/c14896a64089ed3aed0d8bf04d1e9b3d24e4e3e4/chrome/content/zotero/collectionTree.jsx#L1687-L1697 We already disable those menuitems, so this should be good too.
4. Do not allow copying a collection into a library if a linked collection already exists: https://github.com/zotero/zotero/blob/c14896a64089ed3aed0d8bf04d1e9b3d24e4e3e4/chrome/content/zotero/collectionTree.jsx#L1776-L1793 That means that once you copy a collection into another library once, you cannot do it again. I just added an extra step to disable menus of groups containing subcollections. It happens after the "Copy to" menu is opened, since `ZoteroPane.buildCopyCollectionMenu` is not async, to avoid popup appearing after delay. Let me know if this makes sense!

I do have a few of questions about the last point:
- What is the role of linked collections? I see the relation is added but what is it used for? 
- I think it feels a bit off now that menus of libraries are disabled because I think it's not immediately clear why. One could copy a collection to a group, rename the copied collection, then come back some time later, and not be able to copy again, without a clear indication why. 
- I think "Copy to" now feels a tiny bit misleading, since now for all libraries outside of the current one it is "Copy (only once)". If it were "Create linked collection", or something like that, it might have clarified these concerns a bit but it doesn't work well for copying within the current library...
- Finally, if is important to maintain this one-to-one relations between linked collections, should we also add an extra check into `CollectionTree._copyCollections()` to see if a linked collection already exists and throw an error, if so? That is to avoid situations

chrome/content/zotero/zoteroPane.js
```diff
@@ -4091,13 +4091,38 @@ var ZoteroPane = new function()
...
 		let selected = this.getSelectedCollection();
 
 		// Fetch all libraries
 		let topLevelEntries = Zotero.Libraries.getAll().filter(lib => !(lib instanceof Zotero.Feed));
 
+		// Check which libraries have collections linked to the selected collection
+		// and disable their menuitems. Same logic as in CollectionTree.canDropCheckAsync.
+		let linkedCollectionsExist = {};
+		(async () => {
+			for (let library of Zotero.Libraries.getAll()) {
```
dstillman, line 4105: Start with `topLevelEntries` and remove the `instanceof Zotero.Feed` check below?

chrome/content/zotero/zoteroPane.js
```diff
@@ -4091,13 +4091,38 @@ var ZoteroPane = new function()
...
+				if (library.libraryID == selected.libraryID || library instanceof Zotero.Feed) continue;
+				// Check which library has a collection linked to the selected collection
+				let linkedCollection = await selected.getLinkedCollection(library.libraryID, true);
+				linkedCollectionsExist[library.libraryID] = linkedCollection;
+				// Also check which library has collections linked to a subcollection of the selected collection
+				for (let descendent of selected.getDescendents(false, 'collection')) {
+					let subcollection = Zotero.Collections.get(descendent.id);
+					let linkedSubcollection = await subcollection.getLinkedCollection(library.libraryID, true);
+					linkedCollectionsExist[library.libraryID] = linkedSubcollection || linkedCollectionsExist[library.libraryID];
```
dstillman, line 4114: Clearer to just use `if (linkedSubcollection) {` instead of short-circuiting, I think

chrome/content/zotero/zoteroPane.js
```diff
@@ -4091,13 +4091,38 @@ var ZoteroPane = new function()
...
+				}
+			}
+			// Libraries that have linked collections have their menus disabled
+			for (let libraryMenuItem of [...popup.childNodes]) {
+				let menuItemLibID = libraryMenuItem.getAttribute("value").substring(1);
+				if (linkedCollectionsExist[parseInt(menuItemLibID)]) {
```
dstillman, line 4120: No real reason to use `parseInt` here as is, since the object key is always a string anyway. (Or you could use `Map` instead of an array for `linkedCollectionsExist`, and then they'd be actual numbers.)

chrome/content/zotero/zoteroPane.js
```diff
@@ -4135,7 +4160,8 @@ var ZoteroPane = new function()
 				(target) => {
 					// can't copy collection into itself or into non-editable groups
 					return selected == target
-						|| (target instanceof Zotero.Group && !target.editable);
+						|| (target instanceof Zotero.Group && !target.editable)
+						|| linkedCollectionsExist[target.libraryID];
```
dstillman, line 4164: Is this supposed to be here? Is this not taken care of by the block above, which runs asynchronously?

dstillman:
The link is currently used only to prevent a re-drag, but it'll be used in the future to support some sort of intelligent merging/overwriting of data between libraries, with some appropriate dialog (so these won't stay disabled). This is one of the biggest current limitations of group-based workflows, and while just blocking and forcing a deletion in the target library isn't a great solution, it lets us be prevent lots of duplicate copies for that eventual future.

dstillman:
> Finally, if is important to maintain this one-to-one relations between linked collections, should we also add an extra check into CollectionTree._copyCollections() to see if a linked collection already exists and throw an error, if so?

Not necessary. The front-end code shouldn't allow it in most cases, and I think you can end up with duplicates just by moving an item in the target library to the trash that and then undeleting it, so we'll need to handle the possibility of duplicate relations when we implement this properly in any case.

chrome/content/zotero/zoteroPane.js
abaevbog: Oh yeah, for sure! I had a glitch with the `async` execution, so I isolated `topLevelEntries` for troubleshooting. That had nothing to do with `topLevelEntries`, so I changed it back to `for (let library of topLevelEntries)`.

chrome/content/zotero/zoteroPane.js
abaevbog: No, sorry about that. This was accidentally leftover from when I was trying to not disable menus asynchronously.

abaevbog:
> The link is currently used only to prevent a re-drag, but it'll be used in the future to support some sort of intelligent merging/overwriting of data between libraries, with some appropriate dialog (so these won't stay disabled). 

Gotcha! That makes sense, thank you for for the explanation.

While testing this out, I ran into the issue of finding it hard to connect which collection in another group is linked to the collection I am copying, especially if a subcollection is linked or if a collection got renamed after initial copying/linking. So, until the merging data logic is added, would we want to add some way to find out which collections are linked? As a quick fix, instead of disabling the group menu, we could set it to have a single disabled `menuitem` saying something like "Linked as {collection name}".  

Screenshot 2024-10-14 at 3 15 43 PM

At the same time, it doesn't seem like this drag-drop mechanism has caused much confusion in the past, so it very much might be redundant.

PR #4423, opened 2024-07-23
vpat 16: context menu as a drag-drop alternative to change parent item of attachment/note

abaevbog:
- added "Change Parent Item" context menu option to itemTree as a non-drag-drop alternative to changing the parent of attachments or notes
- context menu option appears only when all selected items are non-regular items
- upon activation, a dialog to select the new parent will appear. Added a tweak to the dialog to disable the "accept" button if a selected item is not top level
- minor edit to itemTree to set the multiselect property based on a prop, since we do not want to have multiselect enabled in this instance
- when the new top-level item is picked, set that new item as a parent of the selected items

Addresses part of https://github.com/zotero/zotero/issues/3970

### DIFF
--- a/chrome/content/zotero/selectItemsDialog.js
+++ b/chrome/content/zotero/selectItemsDialog.js
@@ -172,8 +172,8 @@ function onItemSelected()
 	if (io.onlyRegularItems) {
 		// Disable "accept" button if not a top level item is selected
 		let selected = itemsView.getSelectedItems();
-		let acceptBtn = document.querySelector("dialog").getButton("accept");
-		acceptBtn.disabled = (selected && !selected.every(item => item.isRegularItem()));
+		let disableAccept = (selected && !selected.every(item => item.isRegularItem()));
+		document.querySelector("dialog").setAttribute("buttondisabledaccept", disableAccept);
 	}
 }
 
--- a/chrome/content/zotero/selectItemsDialog.js
+++ b/chrome/content/zotero/selectItemsDialog.js
@@ -178,10 +178,10 @@ function onItemSelected()
 {
 	itemsView.runListeners('select');
 	if (io.onlyRegularItems) {
-		// Disable "accept" button if not a top level item is selected
+		// Disable "accept" button if a top-level item isn't selected
 		let selected = itemsView.getSelectedItems();
 		let disableAccept = (selected && !selected.every(item => item.isRegularItem()));
-		// Temp. Disable the button directly only as long as we move the button box in doLoad().
+		// TEMP: Disable the button directly only as long as we move the button box in doLoad().
 		// Then, we should set buttondisabledaccept attribute on the dialog
 		document.querySelector("dialog button[dlgtype='accept']").setAttribute("disabled", disableAccept);
 	}
--- a/chrome/content/zotero/selectItemsDialog.js
+++ b/chrome/content/zotero/selectItemsDialog.js
@@ -89,10 +89,11 @@ var doLoad = async function () {
 	itemsView.setItemsPaneMessage(Zotero.getString('pane.items.loading'));
 
 	const filterLibraryIDs = false || io.filterLibraryIDs;
+	const hideSources = io.hideCollections || ['duplicates', 'trash', 'feeds'];
 	collectionsView = await CollectionTree.init(document.getElementById('zotero-collections-tree'), {
 		onSelectionChange: Zotero.Utilities.debounce(() => onCollectionSelected(), 100),
 		filterLibraryIDs,
-		hideSources: ['duplicates', 'trash', 'feeds']
+		hideSources
 	});
 
 	await collectionsView.makeVisible();
--- a/chrome/content/zotero/selectItemsDialog.js
+++ b/chrome/content/zotero/selectItemsDialog.js
@@ -118,6 +118,13 @@ var doLoad = async function () {
 			button.setAttribute("tabindex", nextButtonTabindex++);
 		}
 	}
+	// Handle any custom button config that can be passed
+	for (let buttonConfig of io.extraButtons || []) {
+		let button = document.querySelector(`dialog button[dlgtype='${buttonConfig.type}']`);
+		button.hidden = false;
+		document.l10n.setAttributes(button, buttonConfig.l10nLabel);
+		button.addEventListener("click", event => buttonConfig.onclick(event));
+	}
 	
 	// Used in tests
 	loaded = true;
--- a/chrome/content/zotero/selectItemsDialog.js
+++ b/chrome/content/zotero/selectItemsDialog.js
@@ -84,7 +84,7 @@ var doLoad = async function () {
 		persistColumns: true,
 		columnPicker: true,
 		emptyMessage: Zotero.getString('pane.items.loading'),
-		multiselect: !io.singleSelection
+		multiSelect: !io.singleSelection
 	});
 	itemsView.setItemsPaneMessage(Zotero.getString('pane.items.loading'));
 
--- a/chrome/content/zotero/selectItemsDialog.js
+++ b/chrome/content/zotero/selectItemsDialog.js
@@ -121,8 +121,8 @@ var doLoad = async function () {
 	// Handle any custom button config that can be passed
 	for (let buttonConfig of io.extraButtons || []) {
 		let button = document.querySelector(`dialog button[dlgtype='${buttonConfig.type}']`);
-		button.hidden = false;
-		document.l10n.setAttributes(button, buttonConfig.l10nLabel);
+		button.hidden = buttonConfig.isHidden(document);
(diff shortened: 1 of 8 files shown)


chrome/content/zotero/selectItemsDialog.js
```diff
@@ -136,6 +137,12 @@ function onSearch()
 function onItemSelected()
 {
 	itemsView.runListeners('select');
+	if (io.onlyRegularItems) {
+		// Disable "accept" button if not a top level item is selected
+		let selected = itemsView.getSelectedItems();
+		let acceptBtn = document.querySelector("dialog").getButton("accept");
+		acceptBtn.disabled = (selected && !selected.every(item => item.isRegularItem()));
```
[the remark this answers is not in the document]
abaevbog: Thank you for pointing that out! I didn't know about `buttondisabledaccept` attribute of the dialog - this is way nicer

chrome/content/zotero/zoteroPane.js
```diff
@@ -3938,6 +3939,11 @@ var ZoteroPane = new function()
 			}
 		}
 		
+		// Update parent item of notes/attachments
+		if (items.every(item => !item.isRegularItem())) {
```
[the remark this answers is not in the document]
abaevbog: > Maybe future-proof this by specifically checking for `isNote`/`isAttachment`

Good point! Since annotation is not a regular item, this option would appear when annotations are selected, which would not be right.

> Maybe call this “Move to Parent”?

"Move to Parent" sounds good to me!

abaevbog:
@AbeJellinek I added the fixes to your comments above. One extra thing I had to change is I had to go back to disabling the "accept" button directly `document.querySelector("dialog button[dlgtype='accept']").setAttribute("disabled", disableAccept)` instead of via `buttondisabledaccept` on the dialog because the later stopped working for me after [moving](https://github.com/zotero/zotero/blob/7c7d8c143341b3ad6b96030e902500d0608756d7/chrome/content/zotero/selectItemsDialog.js#L51) the buttonBox in `onLoad`.

dstillman:
1. Need an ellipsis on the menu option
2. I think it should be "Move to Parent Item…"  rather than "Move to Parent…". People get confused with "parent" in terms of collections and items.
3. We unfortunately also need a way to make [certain files](https://github.com/zotero/zotero/blob/b00b23306d1e9c19fe509e770e37ca7daa5ba0db/chrome/content/zotero/itemTree.jsx#L2191-L2198) standalone attachments, since it allows you to run Retrieve Metadata. On iOS we call this "Move to Standalone Attachment", which doesn't make a ton of grammatical sense but is probably clear enough. Instead of wasting a menu option for this, maybe we can just add a "Move to Standalone Attachment" button in the select-items dialog, over on the left side in line with Cancel/Select?
4. Shouldn't have My Publications, Retracted Items, or Trash in the dialog.

abaevbog:
That makes sense! I tweaked the menu name and added a few customizations to the `selectItems.xhtml` dialog:
- set which collections are hidden via `io.hideCollections` so we can hide all virtual collections here without interfering with other instances where this dialog is used. I thought we might want some of the virtual collections to appear in some other situations (e.g. it doesn't hurt for "My publications" to appear if you are looking to add a related item). Correct me if that's redundant though and we if want to just always hide all virtual collections.
- added `io.extraButtons` parameters that, if set, will add the extra buttons to the dialog. Since `selectItems` dialog is used in a variety of instances, I thought it could also be used in the future to add just one or two buttons without dealing with xhtml. Currently, it's used to display a "Move to standalone attachment" button if all selected items can be turned into top-level standalone attachments. On click of that, attachments are unlinked from their parent items and added to their old parents' collections, so they don't get lost.

dstillman:
OK, I think this has to be "Change Parent Item…", even if "Change Parent Item…" and "Create Parent Item…" are similar. "Move to Parent Item…" is confusing, since it suggests that the item doesn't already have a parent item, and it doesn't really make sense when we also offer the option to convert it to a standalone attachment.

dstillman:
The "Convert to Standalone Attachment" button shouldn't appear when an attachment is already a standalone attachment.

chrome/content/zotero/zoteroPane.js
```diff
@@ -4944,6 +4950,81 @@ var ZoteroPane = new function()
...
+	this.changeParentItem = async function () {
+		let selectedItems = this.getSelectedItems();
+		// Only applies when selected items are not top level items
+		if (selectedItems.some(item => item.isRegularItem())) return;
+
+		let libraryID = this.getSelectedLibraryID();
+		let shouldConvertToStandaloneAttachment = false;
+		let extraButtons = [];
+		// Keep in sync with Zotero.RecognizeDocument.canRecognize()
+		let canBeMovedOutOfparent = !selectedItems.some(item => item.isWebAttachment() && !item.isPDFAttachment() && !item.isEPUBAttachment());
```
dstillman, line 4967: → `OfParent`

chrome/content/zotero/zoteroPane.js
```diff
@@ -4944,6 +4950,81 @@ var ZoteroPane = new function()
...
+					// Unlink parent item
+					item.parentID = null;
+					await item.save({ skipSelect: true });
+				}
+			});
+			return;
+		}
+		if (!io.dataOut?.length) return;
+
+		let newParentItem = await Zotero.Items.getAsync(io.dataOut);
```
dstillman, line 5017: This should be able to just be `get()` — the items will already be loaded.

dstillman:
Does this still work if the dialog is a modal? Because it should be modal — it doesn't really make sense that this context-less dialog can disappear behind the main window. (Same for adding related items, though, which makes me wonder if this it's non-modal for a reason.)

abaevbog:
True, it's not clear that the button will just disregard whatever selection you make from the items list.

I thought about adding an extra step of hiding/disabling the "Convert to Standalone Attachment" button if something is selected. But that runs into issues that if you do click onto something (even accidentally), you can't easily un-select the row in itemTree. There is a tweak for the itemTree that will need to be added for the new citation dialog per design in figma, where a `+` and `-` would be displayed as a last column to include/remove item from the citation. If this feature is added later on, we would then be able hide the "Convert to standalone attachment" button if something is selected, since undoing itemTree selection would get easier. 

I do agree though that this is a nicer approach in general, since context menu space is quite valuable, so it's good to not clutter it if possible.

abaevbog:
I played around with this and ran into no issues if we open it as a modal on mac or windows

dstillman:
Oh, I have one: on macOS, if you try to drag the scrollbar in the collections pane, it drags the whole dialog. (Not sure if this is new with the modal.)

dstillman:
But it's fine in the items pane, for some reason.

In the macOS open/save dialog, the left pane won't actually drag the window. That might just be because the sidebar items are themselves draggable, but I guess an easy fix here would be to just change this to be the same if we can.

abaevbog:
Is that specific to opening the dialog modal? I am trying it now on related items dialog, and dragging the scrollbar will sometimes actually drag it, and sometimes drag the window. Just like with the new dialog here. Let's see if explicitly disabling `-moz-window-drag` on the scrollbar would help

abaevbog:
This css tweak seems to do it for me. I think it boiled down to the container of the collection tree being draggable (so that you can drag below the collection rows), which ended up overlapping with the scrollbar, so they would conflict with one another.

One more thing that `modal` change did is break the tests for some reason...

abaevbog:
Ok, tests should be good now, and this is the scroll behavior after the css spacing tweaks:

https://github.com/user-attachments/assets/f2dd5667-2b37-44e3-b620-c05d8ea2e9af

Earlier, the scrollbar would only properly scroll if you click on its left-most edge (where click would land on the non-draggable part of the collectionTree containers). Now, it should scroll wherever you click.

scss/components/_selectItemsDialog.scss
```diff
@@ -73,6 +73,19 @@
...
+
+	// Move padding into rows so that draggable collection container
+	// does not expand all the way past the scrollbar and cause draggin scrollbar
+	// to drag the entire window.
+	#zotero-collections-tree-container {
+		.virtualized-table-body {
+			padding-inline-end: 0;
+			.row {
+				padding-inline-end: 8px;
+				width: 183px;
```
dstillman, line 85: Why 183px?

abaevbog: After moving padding around, I noticed that the collection rows would stretch all the way to the end, so to leave some space between them and the scrollbar, I set the width that the rows initially had, prior to this css tweak, which was 183px.

Screenshot 2024-10-10 at 1 56 35 PM

I realize it is a somewhat hacky. My intention was to still have the edges and bottom of the left sidebar allow window drag. It is quite possible that it's not worth the hack, in which case we can just make the entire left pane `-moz-window-drag: no-drag`.

PR #4424, opened 2024-07-24
Add `attachmentTitle` and extend file rename templating with regexp

### LINKED ISSUE #3252: `attachmentTitle` variable in filename template string?
https://forums.zotero.org/discussion/comment/440331/#Comment_440331

People will want a way to include an arbitrary string in the filename, and if you have multiple child attachments, that can't be a parent item field.

The problem is that you wouldn't want any attachment title to be used — you'd probably want the ability to test the value (`startsWith`, `includes`, or `regexp`) and extract a substring (index-based `slice` or `match` with a regexp). You could then either exclude anything with "PDF", "Accepted Version", or "Submitted Version" (to exclude all the default titles) or use a prefix like `fn: ` to specify a string to include in the filename. Just thinking out loud here:

Borrowing from jq/shell:

```
{{ if attachmentTitle | startsWith("fn:") }}{{ attachmentTitle | substring(4) }}{{ endif }}
```

Sticking closer to JS:

```
{{ if attachmentTitle.startsWith("fn:") }}{{ attachmentTitle.substring(4) }}{{ endif }}
```

Granted, at some point the question becomes why we're not just using an existing template language…

For this specific problem, we could also make the attachment note accessible, but we've been trying to deprecate those for years.

tnajdek:
This PR adds an option to test and use the title of the best attachment (as `attachmentTitle`) in the file renaming template. For example, the following template will conditionally use the best attachment's title or fall back to Zotero's default renaming template for items that do not have an attachment or for items where the attachment has one of the generic names ("Full Text", "Submitted Version" etc.).

````
{{ if {{ attachmentTitle match="^(full.*|submitted.*|accepted.*)$" }} }}
{{ firstCreator suffix=" - " }}{{ year suffix=" - " }}{{ title truncate="100" }}
{{ else }}
{{ if attachmentTitle }}
{{ attachmentTitle replaceFrom=".pdf|.epub" }}
{{ else }}
{{ firstCreator suffix=" - " }}{{ year suffix=" - " }}{{ title truncate="100" }}
{{ endif }}
````

This PR includes the commit from #3562. However, I've tweaked the functionality so that regex is case-insensitive by default because I expect that's what most people would prefer (also, our condition checking is already case-insensitive). This can still be configured.

Because the new `attachmentTitle` field requires a DB query to find the best attachment, `getFileBaseNameFromItem()` has become an async function. All calls throughout the codebase have been updated to reflect that. One noteworthy change is in `FileDragDataProvider` where previously `getFileBaseNameFromItem()` was used, and now the title of the attachment is used instead.

I've also introduced `start` function to enable truncating from left (should we call it `truncateLeft` instead?) and few small changes to make editing templates slightly easier and to avoid repeated hyphens/underscores when using `case="snake"` or `case="hyphen"`.

Fix #3252

### DIFF
--- a/chrome/content/zotero/preferences/preferences_file_renaming.js
+++ b/chrome/content/zotero/preferences/preferences_file_renaming.js
@@ -43,26 +43,27 @@ Zotero_Preferences.FileRenaming = {
 		this._itemsView.onSelect.removeListener(this._updatePreview);
 	},
 
-	getActiveTopLevelItem() {
-		const selectedItem = Zotero.getActiveZoteroPane()?.getSelectedItems()?.[0];
+	getActiveItem() {
+		let selectedItem = Zotero.getActiveZoteroPane()?.getSelectedItems()?.[0];
 		if (selectedItem) {
 			if (selectedItem.isRegularItem() && !selectedItem.parentKey) {
-				return [selectedItem, this.defaultExt];
+				return [selectedItem, this.defaultExt, ''];
 			}
-			if (selectedItem.isFileAttachment() && selectedItem.parentKey) {
-				const path = selectedItem.getFilePath();
-				const ext = Zotero.File.getExtension(Zotero.File.pathToFile(path));
-				return [Zotero.Items.getByLibraryAndKey(selectedItem.libraryID, selectedItem.parentKey), ext ?? this.defaultExt];
+			if (selectedItem.isFileAttachment()) {
+				let path = selectedItem.getFilePath();
+				let ext = Zotero.File.getExtension(Zotero.File.pathToFile(path));
+				let parentItem = Zotero.Items.getByLibraryAndKey(selectedItem.libraryID, selectedItem.parentKey);
+				return [parentItem, ext ?? this.defaultExt, selectedItem.getField('title')];
 			}
 		}
 
 		return null;
 	},
 
-	updatePreview() {
-		const [item, ext] = this.getActiveTopLevelItem() ?? [this.mockItem ?? this.makeMockItem(), this.defaultExt];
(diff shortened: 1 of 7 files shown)


dstillman:
I might be misunderstanding something, but why are we using the best attachment here? The point of `attachmentTitle` is that it lets you use the title of the attachment being renamed. One purpose is to let you use renaming for secondary files (where a translator or you have set a more descriptive title, like a chapter name, or, to use an example from the original thread, some suffix like " - Annotated") and the filename would incorporate that rather than just being set based on the parent item's metadata alone. Sorry if that wasn't clear.

dstillman:
I realize this probably stems from `getFileBaseNameFromItem()` and `getRenamedFileBaseNameIfAllowedType()` taking the parent item, since the child attachment doesn't necessarily exist yet, but I think we'll have to rework things to be able to pass an attachment title — either the current one or what we're planning to set — to those. This might be somewhat backwards from what we did previously, where the title sometimes depended on the filename rather than vice versa.

This might interact with the changes we need to make for https://github.com/zotero/zotero/issues/4445.

tnajdek:
Thanks for clarifying. Sorry, I believe I got confused about how this is supposed to work.

I've reverted the changes related to using `bestAttachment` (including making `getFileBaseNameFromItem()` asynchronous) and tweaked the function's signature and calls so that it’s possible to include the `attachmentTitle` of the current attachment (or the one being created). I've also attempted to update all the various calls to `getFileBaseNameFromItem()` so that the title of the attachment is included, but I'm not 100% sure I got that right. Any feedback, please let me know!

dstillman:
Merged

PR #4430, opened 2024-07-25
Fix attachment conflict resolve

windingwind:


windingwind:
I'm not sure how to create a real attachment conflict. I'm using the test `should show the conflict resolution window on attachment conflicts` and everything looks fine:

image

dstillman:
1. There should be more padding between the panes so the buttons aren't touching.
2. There shouldn't be a horizontal scrollbar.
3. What's that horizontal line below in the third pane?

windingwind:
Fixed:

image


(FYR, code to make an attachment conflict: https://gist.github.com/windingwind/0dc2250db5d85e754a7604ecb28f5ffe)

chrome/content/zotero/xpcom/data/item.js
```diff
@@ -2652,12 +2652,22 @@ Zotero.Item.prototype.getFilePathAsync = Zotero.Promise.coroutine(function* () {
...
+
+		let storageDir;
+		// Catch errors thrown by getStorageDirectory, otherwise the this.fileExists fails silently
+		try {
+			storageDir = Zotero.Attachments.getStorageDirectory(this);
+		}
+		catch (e) {
+			Zotero.warn(e);
+			return false;
+		}
```
dstillman, line 2664: Can you move this to a separate commit and explain when there would be an error from `getStorageDirectory()`?

dstillman: And since the `coroutine` → `async` changes aren't actually involved here, those should be in their own commit as well. (If a line being changed or added uses `yield`, it's fine to change it to `await` and update the rest of the function, but otherwise it should be its own commit.)

windingwind: Actually the yield -> await plus the try-catch together fix the empty attachment box bug here. The reason of these two changes will still be “fixing cr bug” if they are moved to a separate commit, I suppose?

dstillman: That doesn't make sense. `coroutine` and `async` work exactly the same way, including with try/catch.

chrome/content/zotero/elements/mergeGroup.js
```diff
@@ -142,7 +142,7 @@
 			this._mergePane.libraryID = this.libraryID;
 			this._leftPane.data = this._data.left;
 			this._rightPane.data = this._data.right;
-			this._mergePane.data = this._data.merge;
+			// Fix render error: Do not set data to _mergePane, since choosePane will update it
```
dstillman, line 145: If this isn't necessary it can just be removed — no need to leave an explanation.

Is the idea that we were always running choosePane anyway and this wasn't necessary, and now it causes some extra problem due to the asynchronous render?

chrome/content/zotero/xpcom/data/item.js
windingwind: Yes it doesn’t make sense, but it does happen - the coroutine silently fails after try catch at some bluebird internal code. I am not sure why but it just happened

chrome/content/zotero/elements/mergeGroup.js
windingwind: Yes. Not related to the async render, calling choosePane after setting data to the pane immediately would discard and create a new CE instance, which is unnecessary and throws a error from the CE initialization (because it’s discarded halfway of the initialization)

chrome/content/zotero/xpcom/data/item.js
dstillman: I mean, it doesn't really matter, but there's really no reason it should be any different. To be clear, this works fine:

```javascript
await Zotero.Promise.coroutine(function* () {
    Zotero.debug('1');
    try {
        throw new Error('foo');
    }
    catch (e) {
        Zotero.debug('error')
    }
    Zotero.debug('2');
})();
```

This will log '1', 'error', '2'.

Anyway, any change to `getFilePathAsync()` should be in a separate commit and should be very clear about what exactly it's fixing. `getStorageDirectory()` shouldn't throw an error unless an item isn't passed or a key isn't set. If we're fixing the latter, we should handle that explicitly in this function, not wrap getStorageDirectory in a try/catch for some unspecified potential error.

windingwind: Seriously, async v.s. bluebird-ed coroutine does make a difference, though I'm not sure why. I suspect it's a bug with Bluebird, possibly indicating some more serious issue with Bluebird. We should consider gradually stopping using the `coroutine`.


Here are two different git diff patches, one using coroutine and one using async, besides that, everything is the same. The former failed to render the attachment CR and the latter rendered the attachment title.

1. Apply the patch (`git apply`), build and start Zotero
2.  select an attachment in library
3. Run the code below to show a CR window:
```js
item = ZoteroPane.getSelectedItems()[0];
j = item.toJSON();
j.mtime = new Date().getTime() + 100000;
j.md5 = 'f4ce1167f3a854896c257a0cc1ac387f';
await Zotero.Sync.Data.Local.saveCacheObjects('item', 1, [j]);
item.attachmentSyncState = "in_conflict";
await item.saveTx({ skipAll: true });
await Zotero.Sync.Storage.Local.resolveConflicts(1); 
```

git patch (using coroutine):
image


<details>
<summary>Show</summary>

```
diff --git a/chrome/content/zotero/xpcom/data/item.js b/chrome/content/zotero/xpcom/data/item.js
index 650e5c911..9a24e4934 100644
--- a/chrome/content/zotero/xpcom/data/item.js
+++ b/chrome/content/zotero/xpcom/data/item.js
@@ -2653,9 +2653,15 @@ Zotero.Item.prototype.getFilePathAsync = Zotero.Promise.coroutine(function* () {
 			return false;
 		}
 		
-		path = OS.Path.join(
-			OS.Path.normalize(Zotero.Attachments.getStorageDirectory(this).path), path
-		);
+		try {
+			path = OS.Path.join(
+				OS.Path.normalize(Zotero.Attachments.getStorageDirectory(this).path), path
+			);
+		}
+		catch (e) {
+			Zotero.logError(e);
+			return false;
+		}
 		
 		if (!(yield OS.File.exists(path))) {
 			Zotero.debug("Attachment file '" + path + "' not found", 2);
```

</details>

git patch (using async):

image


<details>
<summary>Show</summary>

```
diff --git a/chrome/content/zotero/xpcom/data/item.js b/chrome/content/zotero/xpcom/data/item.js
index 650e5c911..cf85c2cf7 100644
--- a/chrome/content/zotero/xpcom/data/item.js
+++ b/chrome/content/zotero/xpcom/data/item.js
@@ -2615,7 +2615,7 @@ Zotero.Item.prototype.getFilePath = function () {
  * @return {Promise<String|false>} - A promise for either the absolute path of the attachment
  *                                   or false for invalid paths or if the file doesn't exist
  */
-Zotero.Item.prototype.getFilePathAsync = Zotero.Promise.coroutine(function* () {
+Zotero.Item.prototype.getFilePathAsync = async function () {
 	if (!this.isAttachment()) {
 		throw new Error("getFilePathAsync() can only be called on attachment items");
 	}
@@ -2653,11 +2653,17 @@ Zotero.Item.prototype.getFilePathAsync = Zotero.Promise.coroutine(function* () {
 			return false;
 		}
 		
-		path = OS.Path.join(
-			OS.Path.normalize(Zotero.Attachments.getStorageDirectory(this).path), path
-		);
+		try {
+			path = OS.Path.join(
+				OS.Path.normalize(Zotero.Attachments.getStorageDirectory(this).path), path
+			);
+		}
+		catch (e) {
+			Zotero.logError(e);
+			return false;
+		}
 		
-		if (!(yield OS.File.exists(path))) {
+		if (!(await OS.File.exists(path))) {
 			Zotero.debug("Attachment file '" + path + "' not found", 2);
 			this._updateAttachmentStates(false);
 			return false;
@@ -2675,7 +2681,7 @@ Zotero.Item.prototype.getFilePathAsync = Zotero.Promise.coroutine(function* () {
 			this._updateAttachmentStates(false);
 			return false;
 		}
-		if (!(yield OS.File.exists(path))) {
+		if (!(await OS.File.exists(path))) {
 			Zotero.debug("Attachment file '" + path + "' not found", 2);
 			this._updateAttachmentStates(false);
 			return false;
@@ -2701,12 +2707,12 @@ Zotero.Item.prototype.getFilePathAsync = Zotero.Promise.coroutine(function* () {
 		}
 		
 		// If valid, convert this to a regular string
-		yield Zotero.DB.queryAsync(
+		await Zotero.DB.queryAsync(
 			"UPDATE itemAttachments SET path=? WHERE itemID=?",
 			[file.leafName, this._id]
 		);
 		
-		if (!(yield OS.File.exists(file.path))) {
+		if (!(await OS.File.exists(file.path))) {
 			Zotero.debug("Attachment file '" + file.path + "' not found", 2);
 			this._updateAttachmentStates(false);
 			return false;
@@ -2718,7 +2724,7 @@ Zotero.Item.prototype.getFilePathAsync = Zotero.Promise.coroutine(function* () {
 	}
 	
 	// NOTE: Test for platform slashes before changing to IOUtils.exists()
-	if (!(yield OS.File.exists(path))) {
+	if (!(await OS.File.exists(path))) {
 		Zotero.debug("Attachment file '" + path + "' not found", 2);
 		this._updateAttachmentStates(false);
 		return false;
@@ -2727,7 +2733,7 @@ Zotero.Item.prototype.getFilePathAsync = Zotero.Promise.coroutine(function* () {
 	this._updateAttachmentStates(true);
 	
 	return path;
-});
+};
 
 
 /**
@@ -2786,7 +2792,7 @@ Zotero.Item.prototype.getFilename = function () {
 /**
  * Asynchronous check for file existence
  */
-Zotero.Item.prototype.fileExists = Zotero.Promise.coroutine(function* () {
+Zotero.Item.prototype.fileExists = async function () {
 	if (!this.isAttachment()) {
 		throw new Error("Zotero.Item.fileExists() can only be called on attachment items");
 	}
@@ -2795,8 +2801,8 @@ Zotero.Item.prototype.fileExists = Zotero.Promise.coroutine(function* () {
 		throw new Error("Zotero.Item.fileExists() cannot be called on link attachments");
 	}
 	
-	return !!(yield this.getFilePathAsync());
-});
+	return !!(await this.getFilePathAsync());
+};
 
 
 /**
```

</details>

windingwind: Did more investigation and here are the minimal steps to reproduce the difference between async v.s coroutine:

1. Apply patches mentioned in the comment above;
2. Select an attachment item in library and execute this code:


```js
item = ZoteroPane.getSelectedItems()[0];
unsaved = new Zotero.Item(item.itemType);
unsaved.libraryID = item.libraryID;
unsaved.fromJSON(item.toJSON());
console.log(unsaved.getFilePathAsync());
```

The result of the coroutine:

```
11:24:14.089
Object { _bitField: 33554432, _fulfillmentHandler0: undefined, _rejectionHandler0: false, _promise0: undefined, _receiver0: undefined, _cancellationParent: {…}, _trace: {…} }
​
_bitField: 33554432
​
_cancellationParent: Object { _bitField: 1, _fulfillmentHandler0: finallyHandler(reasonOrValue), _branchesRemainingToCancel: 1, … }
​
_fulfillmentHandler0: undefined
​
_promise0: undefined
​
_receiver0: undefined
​
_rejectionHandler0: false
​
_trace: Object { _promisesCreated: 0, _length: 1, stack: "captureStackTrace@resource://zotero/bluebird/debuggability.js:915:23\nCapturedTrace@resource://zotero/bluebird/debuggability.js:807:22\nlongStackTracesCaptureStackTrace@resource://zotero/bluebird/debuggability.js:482:19\nmodule.exports/Promise.prototype._then@resource://zotero/bluebird/promise.js:261:17\nmodule.exports/Promise.prototype._passThrough@resource://zotero/bluebird/finally.js:94:17\nmodule.exports/Promise.prototype.finally@resource://zotero/bluebird/finally.js:103:17\nPromiseSpawn@resource://zotero/bluebird/generators.js:36:34\nmodule.exports/Promise.coroutine/<@resource://zotero/bluebird/generators.js:197:21\n@debugger eval code:5:21\ngetEvalResult@resource://devtools/server/actors/webconsole/eval-with-debugger.js:238:24\nexports.evalWithDebugger@resource://devtools/server/actors/webconsole/eval-with-debugger.js:166:14\nevaluateJS@resource://devtools/server/actors/webconsole.js:940:38\nevaluateJSAsync/<@resource://devtools/server/actors/webconsole.js:846:35\nexports.makeInfallible/<@resource://devtools/shared/ThreadSafeDevToolsUtils.js:103:22\n", … }
​
<prototype>: Object { _promiseCreated: _promiseCreated(), _pushContext: _pushContext(), _popContext: _popContext(), … }
debugger eval code:5:9
```

The result of the async:
```
11:26:13.366
Promise { <state>: "fulfilled", <value>: false }
​
<state>: "fulfilled"
​
<prototype>: Promise.prototype { … }
debugger eval code:5:9
```

windingwind: > getStorageDirectory() shouldn't throw an error unless an item isn't passed or a key isn't set

I think the behavior is correct - we use unsaved item objects in the CR window and these items do not have keys. We would need to anyway use a try-catch

dstillman: I'm not sure which behavior you mean. I'm just saying that any change to a core Zotero.Item function is its own thing separate from CR and needs to be fixed in terms of what inputs we want the function to handle. If we want fileExists() or getFilePathAsync() to be able to be run on an item without a key, those functions should explicitly check for the presence of a key and return false. They shouldn't just swallow any thrown exception from another function.

windingwind: > I'm not sure which behavior you mean.

To be clear, here we are discussing two different things: separating commit to the Zotero.Item and how/why `coroutine and async work exactly the same way, including with try/catch.` is not correct.

> any change to a core Zotero.Item function is its own thing separate from CR and needs to be fixed in terms of what inputs/contexts we want the function to handle

Yes, I will.

Anyway, apparently Bluebird fails to handle the try-catch properly when the calling stack is complex, at least in the case above (I believe you can 100% reproduce). I was trying to state that "there's really no reason coroutine and async should be any different" does not align with the facts and we should be aware that Bluebird's coroutine can fail in corner cases and consider replacing it in our code.

windingwind: The Zotero.Item related changes are separated to #4468

dstillman: Yeah, I can reproduce the UI difference, though I'm not actually sure what's happening. What are you suggesting the actual difference is in behavior?

The coroutine version does appear to handle the try/catch properly, at least in the basic sense. I'm not clear what you were showing in the second example, but if you change `console.log(unsaved.getFilePathAsync());` to `return await unsaved.getFilePathAsync()`, you get the logged error and `false` either way, so it's definitely working in that sense.

It's possible there are some subtle differences in the timing of yielding/returning between Bluebird and async/await that affect the rendering.

Note that, though the UI is correct, I get a thrown error when running your first example with the async version (and only the async version):

```
JavaScript error: chrome://zotero/content/elements/attachmentBox.js, line 342: TypeError: this._id(...) is null
```

dstillman: (To be clear, we don't actually need to spend time figuring out what the difference is, beyond academic interest. Changing `coroutine` to `async` is always fine for its own sake, and if it fixes some weird problem, all the better. I just want to avoid claims that exceed our actual understanding of the problem.)

dstillman:
Merged

chrome/content/zotero/xpcom/data/item.js
windingwind: > Note that, though the UI is correct, I get a thrown error when running your first example with the async version (and only the async version):

Yes, that's what I meant to say in https://github.com/zotero/zotero/pull/4430#discussion_r1696839734 . Should be fixed after merging the PR

windingwind: > (To be clear, we don't actually need to spend time figuring out what the difference is, beyond academic interest. Changing coroutine to async is always fine for its own sake, and if it fixes some weird problem, all the better. I just want to avoid claims that exceed our actual understanding of the problem.)

Yeah totally agree.

> though I'm not actually sure what's happening. What are you suggesting the actual difference is in behavior?

I don't know why either - its unexpected behavior made the debugging extremely difficult. Either from logging or break point failed to catch why the execution stops at a certain point of the coroutine function until I tried to use async instead of coroutine. I just want to leave a remark about this weird behavior of coroutine for future reference.

PR #4431, opened 2024-07-25
File renaming: Match content type prefixes, add UI

### LINKED ISSUE #4429: `autoRenameFiles.fileTypes` should take file extensions rather than MIME types
If you drag something without an extension, it can detect and then get the possible extensions we know about for that MIME type and check those (or maybe we already add the extension anyway when we sniff the file type).

We could also support files with alternative extensions for specified ones. E.g., if you set `.jpg` but not `.jpeg`, and then you drag a `.jpeg`, we would ideally map that to `image/jpeg` and then see that `jpg` is a possible extension and then rename the incoming file based on `.jpg` being in the pref. Same for, say, `.aif` and `.aiff`. (It looks like we'll need some additional mapping tables and/or entries for this, but this doesn't need to be super comprehensive — just some common types.)

We can do a pref migration to convert modified prefs that contain known types (though I doubt anyone has changed this).

AbeJellinek:
Closes #4429, except I kept matching on MIME types only and just expanded it to accept prefixes. We don't always have a reliable file extension at the point where we decide whether to rename (`importFromURL()` - we just guess from the MIME type a lot of the time there).

The UI looks awful right now:

image

Suggestions would be appreciated! (Changing the label to just "Rename:" looked worse, removing the indent looked way worse...)

### DIFF
--- a/chrome/content/zotero/elements/zoteroSearch.js
+++ b/chrome/content/zotero/elements/zoteroSearch.js
@@ -476,7 +476,7 @@
 				
 				case 'fileTypeID':
 					var rows = Zotero.FileTypes.getTypes().map(type => ({
-						name: Zotero.getString('fileTypes.' + type.name),
+						name: Zotero.getString('file-type-' + type.name),
 						value: type.id
 					}));
 					
--- a/chrome/content/zotero/preferences/preferences_general.js
+++ b/chrome/content/zotero/preferences/preferences_general.js
@@ -218,7 +218,7 @@ Zotero_Preferences.General = {
 	}),
 	
 	setAutoRenameFileTypes: function () {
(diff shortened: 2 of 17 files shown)


dstillman (changes requested):
1. Re: the UI, I think we have to indent the label so that it starts at the same level as "Rename linked files", and then give it a smidge of top margin. This all only applies to automatic renaming and should be disabled if "Automatically rename locally added files" is disabled, right?
2. And then for now I think we can just put the checkboxes on the same line. If we add more types, we could make a nicer component with checkmarks, but with just four it's probably fine to go horizontal.
3. If I enable Audio and add an MP3, the file is renamed, but the attachment title is also changed to the filename. It should be set to…"Audio"? Same for "Image".
4. Oh, but we should have "Video" too. Not sure if that gets beyond what we can reasonably do horizontally.

PR #4434, opened 2024-07-26
vpat 25: make all relevant text and section headers in preferences announceable

abaevbog:
- explicitly add aria-labels to groupbox and vbox sections of the settings. It's required for voiceover to announce that label when focus enters the group.  It's not needed for NVDA and JAWS as they recognize most h2 labels as labels of grouppings.
- add all relevant additionl text for the section as aria-description/aria-describedby so it is announced after the section's name when focus enters it
- add role="group" to vbox sections, since vbox does not have any semantics
- skip tabstop on scrollable "#prefs-content" container since it is not required

Essentially a different (likely better per https://github.com/zotero/zotero/issues/3930#issuecomment-2251045406) approach to #4414. The main problem here is JAWS as it does not want to activate the virtual cursor inside of dialog. An alternative from the linked PR was to wrap the window in a `<browser>` (which is done by firefox and thunderbird) but it might be even messier.

For reference, vpat 25 issue description: 

---

Description: When navigating through the general settings page of the Zotero desktop app, it was discovered that the JAWS and NVDA screen readers are unable to read or focus on important text elements. This issue can lead to confusion, as some interactive elements rely on the associated text labels to provide context and guidance to users. It is essential to ensure that all relevant information required for understanding and using the application can be accessed and announced by screen readers.

### DIFF
--- a/chrome/content/zotero/preferences/preferences.xhtml
+++ b/chrome/content/zotero/preferences/preferences.xhtml
@@ -92,7 +92,7 @@
 					hidden="true"/>
 				<search-textbox id="prefs-search" placeholder="&zotero.lookup.button.search;" timeout="1"/>
 			</html:div>
-			<html:div id="prefs-content">
+			<html:div id="prefs-content" tabindex="-1">
 				<html:div id="prefs-help-container">
 					<button
 						oncommand="Zotero_Preferences.openHelpLink()"
--- a/chrome/content/zotero/preferences/preferences_advanced.xhtml
+++ b/chrome/content/zotero/preferences/preferences_advanced.xhtml
@@ -25,7 +25,7 @@
 
 <vbox id="zotero-prefpane-advanced" onload="Zotero_Preferences.Advanced.init()">
 	<vbox class="main-section" id="zotero-prefpane-advanced-general-tab">
-		<groupbox id="zotero-prefpane-advanced-miscellaneous">
+		<groupbox id="zotero-prefpane-advanced-miscellaneous" aria-label="&zotero.preferences.miscellaneous;">
 			<label><html:h2>&zotero.preferences.miscellaneous;</html:h2></label>
 			
 			<html:div class="pref-row">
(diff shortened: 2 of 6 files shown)


abaevbog:
Making all sections focusable like this now is an overkill since the headers of sections get announced by JAWS or NVDA when the focus enters them. That was not the case before. Not sure if it's improved due to recent gecko update or maybe the actual JAWS/NVDA updates.

In any case, now it looks like all we need to get headers and other text announced is to explicitly add them to the groupbox as `aria-label` or `aria-description`. `aria-label` is needed only for voiceover as NVDA and JAWS pick up the `h2` as the label automatically. Voiceover still won't read additional text added via `aria-description` when focus enters the group but it's not as big of an issue because it's virtual cursor can go through the text of the window (unlike JAWS).

abaevbog:
I think this gets us what we want: sections get announced when focus enters them with all screen readers. Linked descriptions will be announced with JAWS and NVDA. Voiceover does not announce them but with quicknav enabled it can still move the cursor onto the text to read it.

For reference, this is before:

https://github.com/user-attachments/assets/15df1ce1-84a9-4498-bcca-fe93c453f86c

https://github.com/user-attachments/assets/f20d7c90-cb90-49d1-a56f-444ebed59940

abaevbog:
And this is after:

https://github.com/user-attachments/assets/ddbdf439-6610-47d8-981d-0eacff44cd61



https://github.com/user-attachments/assets/f5b65f5f-311e-472c-ad0b-2d6729f7dc25

dstillman:
So in NVDA that's going to read that whole intro text before it announcing "Automatically rename locally added files"? As a user, can you skip that? And that only happens when you enter the group, not just when you move between checkboxes within the group?

abaevbog:
> So in NVDA that's going to read that whole intro text before it announcing "Automatically rename locally added files"?

Correct, yes, that is the behavior for all screen readers (I checked on JAWS, NVDA and Voiceover). I feel like it's in general not a bad thing as these comments are indeed quite important. When possible, I added remaining comments as `aria-describedby`, in which case they get announced after the main input label after a delay (e.g. Cite > Citation Options > Include URLs of paper articles in references). But for general comments that need to be available without being 100% specific to any input, this felt like the best options.

> As a user, can you skip that?

When a message is being announced, you can always stop the speech with a shortcut. For example, with NVDA, it's Control. Then, there is another shortcut to read the currently focused element. For NVDA, it's Insert(by default)+Tab. For JAWS, it's the same actually. So if you wanted to skip the description text, you could press Ctrl to stop the speech and Insert+Tab to announce the currently focused element, which will just read the checkbox label.

> And that only happens when you enter the group, not just when you move between checkboxes within the group?

Right

PR #4436, opened 2024-07-26
Fix note editor tooltip

windingwind:
fix: #4435


windingwind:
I have been confused by the randomly disappearing of tooltip text in the note editor for a while. It turns out that the `tooltip="html-tooltip"` in the note editor CE actually points to the `<tooltip>` element created inside the tags box CE, which isn't visible when the tags box is collapsed.

Seems like we should always create the `<tooltip>` element for each top-level window and use that tooltip element in all CEs and iframes, rather than create the tooltip element inside CEs.

windingwind:
I think this PR is safe to launch. I confirm the tooltip text now works fine in the item pane, note editor, and reader, in the main window or standalone window.

PR #4438, opened 2024-07-26
Add plugin blocking list

windingwind:


dstillman:
Nice!

One thing: it looks like if once a plugin is blocked, if I change the hard-coded version range, it doesn't re-enable the plugin. That would probably almost never be an issue, but if we set the wrong version range by mistake, it would be bad if we couldn't undo it.

chrome/content/zotero/xpcom/plugins.js
```diff
@@ -384,7 +397,64 @@ Zotero.Plugins = new function () {
...
 	
-	
+
+	// TODO: Get blocking list from server
+	const BLOCKED_PLUGINS = {
+		"zoterostyle@polygon.org": {
+			versionRanges: [{
+				maxVersion: "4.5.99"
+			}],
+			reason: "This plugin, up to version 4.6.0, violates Zotero's plugin policies by severely break user interface.",
```
dstillman, line 407: I think we can just say `Versions of this plugin prior to version 4.6.0 break the Zotero user interface.`

windingwind: updated

windingwind:
Thanks! Updated.

PR #4446, opened 2024-07-27
Prompt to import supported file types on drag-and-drop?

### LINKED ISSUE #4409: Prompt to import supported file types on drag?
https://forums.zotero.org/discussion/comment/468408/#Comment_468408

tnajdek:
Checks if all dropped files can be imported (by examining extensions) and asks whether to import rather than create attachments (or maybe we should say items?). It's still possible to opt-out of importing which will create attachments/items as normal. Alternatively, if it's not possible to import any of the dragged files, it will proceed to create attachments as normal without prompting.

This needs to check translators which can add a noticeable lag first time files are dropped onto items tree.

https://github.com/user-attachments/assets/bc06a255-b025-4da0-a268-efa8dc70d7aa

Resolve #4409

### DIFF
--- a/chrome/content/zotero/collectionTree.jsx
+++ b/chrome/content/zotero/collectionTree.jsx
@@ -2409,6 +2409,15 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 			else {
 				var parentCollectionID = false;
 			}
+
+
+			if (!targetTreeRow.isDuplicates() && !targetTreeRow.isUnfiled() && !targetTreeRow.isRetracted() && !targetTreeRow.isTrash() && !targetTreeRow.isPublications()) {
+				let { promptToImport } = ChromeUtils.importESModule('chrome://zotero/content/modules/promptToImport.mjs');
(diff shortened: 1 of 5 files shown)


tnajdek:
Being able to drag-and-drop `.bib` files would probably help here:

https://forums.zotero.org/discussion/comment/492843#Comment_492843

dstillman:
It probably shouldn't prompt to import if you drop on an item? Only top-level in a collection or on a library/collection itself? Makes it a little more annoying to drag in, but it's not really logical to import if you drag onto an item.

There should probably also be a Cancel button, with adding as an attachment as a secondary action. It's nice to be able to just bail on a potentially confusing prompt.

tnajdek:
I’ve introduced the following changes:

* Do not prompt for importing if files are dropped onto an item row
* Add a Cancel button to abort importing/adding attachments
* Make localisation aware of the number of items for a nicer prompt
* Also prompt to import if compatible files are dropped onto a collection or library node
  * To make this work without duplicating code, I’ve moved the implementation to `modules/promptToImport.mjs`
* Rebased onto the current `main`

tnajdek:
https://forums.zotero.org/discussion/comment/515976

PR #4459, opened 2024-07-29
Don't set default attachment title if not renaming file

### LINKED ISSUE #4445: Don't set default attachment title if not renaming file
https://forums.zotero.org/discussion/comment/469170/#Comment_469170

I noticed this while testing #4431 and it seemed wrong to me. If we're not renaming a file (because auto-renaming is disabled in general or for a given file type), we shouldn't set a default title either, because the original filename may be descriptive. That's already our behavior for subsequent files.

If the person then manually uses Rename File from Parent Metadata, we can set the default title at that point.

AbeJellinek:
Except from Rename File from Parent Metadata.

Fixes #4445

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -5667,6 +5667,7 @@ var ZoteroPane = new function()
 			
 			let parentItemID = item.parentItemID;
 			let parentItem = await Zotero.Items.getAsync(parentItemID);
+			var oldBaseName = item.attachmentFilename.replace(/\.[^.]+$/, '');
 			var newName = Zotero.Attachments.getFileBaseNameFromItem(parentItem, { attachmentTitle: item.getField('title') });
 			
 			let extRE = /\.[^\.]+$/;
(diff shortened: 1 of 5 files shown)


chrome/content/zotero/zoteroPane.js
```diff
@@ -5682,6 +5683,11 @@ var ZoteroPane = new function()
 				continue;
 			}
 			
+			if (item.getField('title') === oldBaseName) {
```
dstillman, line 5686: Can you add a comment explaining this?

I'm confused by this, because if I set a PDF title to "Foo" and then run Rename File from Parent Metadata, the file still gets named as "PDF". Is that not supposed to happen? If I change the test below to set the title to "Foo", it stays that way.

dstillman: Merged, but would still be good to clarify this and make sure it's doing what we want.

dstillman: Sorry, I may have been confusing "Rename File from Parent Metadata" and "Retrieve Metadata". I'm seeing what you're seeing for Rename File, and the behavior seems right.

PR #4464, opened 2024-07-30
Make macOS control normalization into a mixin, fix more components

### LINKED ISSUE #4461: Incorrect button margins in Retrieve Metadata window
Before:

Screenshot 2024-07-30 at 6 47 21 AM

After:

Screenshot 2024-07-30 at 6 46 12 AM

Regression from 75cb9d44bd11650100edfdb3493c7cb9b5025312

AbeJellinek:
With the idea that we'll include this in more and more places over time, and eventually just make it global again.

Have to make sure not to include this mixin in the global scope in stylesheets that are themselves imported elsewhere (i.e., filename begins with an underscore).

Fixes #4461


dstillman:
Merged

PR #4468, opened 2024-07-31
Fix Zotero.Item.fileExists calling on unsaved item

windingwind:
Separated from #4430


dstillman:
I removed the `coroutine` → async `changes.` We can make those separately, no commentary needed — it's fine to do for its own sake. But 1) they're not related to the `fileExists` change here and 2) it's still not clear to me what the actual behavioral difference was, and I don't want to make a potentially misleading claim in a commit message about coroutine and try/catch, which definitely works in general.

PR #4474, opened 2024-07-31
Auto-rename file and set automatic title after Create Parent Item

### LINKED ISSUE #4470: Auto-rename file after Create Parent Item
https://forums.zotero.org/discussion/comment/469395/#Comment_469395

AbeJellinek:
Fixes #4470

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -5342,12 +5342,24 @@ var ZoteroPane = new function()
 			}
 		}
 
-		await Zotero.DB.executeTransaction(async () => {
-			for (let item of items) {
-				item.setAutoAttachmentTitle();
-				await item.save();
+		for (let item of items) {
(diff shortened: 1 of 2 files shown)


dstillman:
Wait, isn't this only affecting the title? The thing we're fixing here is renaming the file:

> The filename should also be changed automatically as long as you've enabled renaming for linked files in the settings. It looks like that's not working properly for Create Parent Item in beta 108 the way that it works for Retrieve Metadata, but we'll fix that for an upcoming version. 

In the current beta, the attachment title already gets set to "PDF", so I'm not sure what this is fixing.

(For future clarity, let's say that "renaming" is always referring to the file, not the attachment title. The attachment title gets set or changed. Only the file gets renamed.)

dstillman:
> I can't reproduce that with Create Parent Item on main. If I add a linked PDF with Rename linked files disabled, then run Create Parent Item -> Manual Entry, the title doesn't change.

I was seeing it in beta 108, but I'm not in 109, so maybe this was affected by other changes in the interim.

PR #4479, opened 2024-07-31
Fix to test breakage

abaevbog:
- fix annotation test breakage. `importOptions` always has a title field, set in `importPDFAttachment`,
so the title was never getting set.
- fix document recognition test. if `autoRenameFiles.linked` pref is false, the file's title should not change

### DIFF
--- a/test/content/support.js
+++ b/test/content/support.js
@@ -1007,9 +1007,7 @@ function importFileAttachment(filename, options = {}) {
 	Object.assign(importOptions, options);
 	// Override default behavior - don't set title based on type,
 	// just use extension-less leafName
-	if (!('title' in importOptions)) {
-		importOptions.title = file.leafName.replace(/\.[^.]+$/, '');
-	}
+	importOptions.title ??= file.leafName.replace(/\.[^.]+$/, '');
 	return Zotero.Attachments.importFromFile(importOptions);
 }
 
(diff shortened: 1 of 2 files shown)


abaevbog:
@AbeJellinek these are a few tweaks that got the tests to pass for me. They are followups to https://github.com/zotero/zotero/commit/dd1601793c2580b6430264e7f2884b5f2e5ec6f3.

For annotations "should create a note from multiple PDF files containing single annotation" test, the  `if (!('title' in importOptions))` condition of `importFileAttachment` was never passing because the title is key is always added in [importPDFAttachment](https://github.com/zotero/zotero/blob/374dab2cb4f2322690538f6ed35ba0fafd61dd95/test/content/support.js#L1033), so the file's name was not what the test expects.

And for the recognition test, since `autoRenameFiles.linked` is set to false in the beginning of the [test](https://github.com/zotero/zotero/blob/374dab2cb4f2322690538f6ed35ba0fafd61dd95/test/tests/recognizeDocumentTest.js#L284), [Zotero.Attachments.shouldAutoRenameFile](https://github.com/zotero/zotero/blob/dd1601793c2580b6430264e7f2884b5f2e5ec6f3/chrome/content/zotero/xpcom/data/item.js#L3911) ends up being false and the item is not renamed. Based on the comments, I assumed we expect the item's name to remain unchanged, so I changed the expected condition.

Correct me if I misunderstood something there

PR #4489, opened 2024-08-03
Setting - Feeds fix misaligned unread/read cleanup option

kassiansun:
The settings was writing to the wrong preference key, code tested on macOS


dstillman:
Thanks!

Issue #4492, opened 2024-08-03
Citation-details line not wrapping in Quick Format citation popup


dstillman:
https://forums.zotero.org/discussion/116467/z7-beta-very-long-fields-in-citation-picker

(And it looks like there is a comma, but a space is missing.)

dstillman:
Fixed by https://github.com/zotero/zotero/commit/d446fc3c6f6374105a349868b7f804e3384ce733

PR #4496, opened 2024-08-04
QF: fix citation properties popup getting stretched

abaevbog:
- set `max-width` on the entire popup so that a long description of the item does not stretch the panel
- make sure all whitespaces in the description are properly rendered
- add tooltiptext to the description of the citation so it can still be visible on hover if it does not fully fit into the panel

Fixes: #4492

Screenshot 2024-08-04 at 4 42 05 PM

### DIFF
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -1009,7 +1009,7 @@ var Zotero_QuickFormat = new function () {
 				}
 			}
 		}
-		// if there is no comma in the end, insert it
+		// if there is no period in the end, insert it
 		if (infoHbox.textContent.length && infoHbox.textContent[infoHbox.textContent.length - 1] !== ".") {
 			let period = document.createElement("span");
 			period.textContent = ".";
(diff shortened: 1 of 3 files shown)


abaevbog:
That does make sense! To wrap description text in citation details popup, I think we have to refactor `_buildItemDescription` a bit to construct description not as a collection of `<label>`s inside of a `<hbox>` but as a collection of `<spans>` inside of a `<description>`. Otherwise, the text would not break in the right place or carry over to the next line too early when I tried it.

How does this look?

For comparison, this is before on the left and after on the right:

Screenshot 2024-08-05 at 5 01 42 PM


`_buildItemDescription` is also used to construct the description of items list. This is the items list before and after. As far as I can tell, this is unchanged as expected.

Screenshot 2024-08-05 at 5 04 21 PM

dstillman:
Merged in https://github.com/zotero/zotero/commit/d446fc3c6f6374105a349868b7f804e3384ce733 and https://github.com/zotero/zotero/commit/802210a6bc00bb6e2d49807e9bf898ee1b801fec

(Sorry, I should've used a clearer description for the max-width change.)

Issue #4498, opened 2024-08-05
Missing indentation in quick search mode menu on Windows


dstillman:
![Screenshot 2024-08-05 at 1 00 51 AM](https://github.com/user-attachments/assets/cbc1a41a-d8d7-456a-b4b8-5a1bd03f6bc7)

Issue #4497, opened 2024-08-05
Quick search mode dropmarker overlaps icon on Windows


dstillman:
https://forums.zotero.org/discussion/116513/zotero-beta-windows-icon-overlaps-in-the-search-box

PR #4501, opened 2024-08-05
Fix quick search styling

windingwind:
fix: #4497
fix: #4498


windingwind:
MacOS:

image

Windows:

image

windingwind:
Menupopup:

image

Issue #4491, opened 2024-08-03
Accessibility of OpenURL resolver settings


dstillman:
https://forums.zotero.org/discussion/116468/zotero-beta-7-accessibility-regression-selecting-a-resolver

I didn't test.

PR #4506, opened 2024-08-05
Fix preferences resolver dropdown glitchy keyboard navigation

abaevbog:
- minor refactoring to avoid deleting the first ("Custom") menuitem of the resolver menulist. It is the first item that is selected when popup opens and deleting it confuses keyboard navigation, so that arrowDown/Up won't navigate the menu (unless the menu is hovered over with a mouse)
- explicitly re-select the first item when the menulist closes. Otherwise, in case of having navigated the menus with arrows without changing selection and closing the popup, next time resolver selector appears, arrowUp/Down will not navigate the list. Only occurs on windows.
- clear the resolver menus (except for the 1st item) when the popup closes so that arrowUp/Down on focused dropdown don't select invalid top-level menus (e.g. "North America").
- make sure that if the url has been editted, the resolver dropdown's value will switch to "Custom" even if the resolvers were not loaded

fixes: #4491


abaevbog:
For reference, some parts of the glitchy behavior before:

https://github.com/user-attachments/assets/42917703-a975-448d-bd94-d7c08f4d2b27

abaevbog:
And this is after when I try to perform similar actions with arrow keys:

https://github.com/user-attachments/assets/f5f2919d-da62-4a6a-8176-44b35bdc64a4

Issue #4510, opened 2024-08-06
Single-select richlistitem has unwanted focus ring


windingwind:
This change added an unwanted focusring on Linux. Is it necessary?

![image](https://github.com/user-attachments/assets/4c2eda95-d860-4ace-a88f-ae23c3c754ff)

_Originally posted by @adomasven in https://github.com/zotero/zotero/pull/4265#discussion_r1703683095_

PR #4511, opened 2024-08-06
Unify richlistitem styling rules

windingwind:
Fix richlistitem unwanted focus ring on Linux
fix: #4510

### DIFF
--- a/scss/components/_richlistbox.scss
+++ b/scss/components/_richlistbox.scss
@@ -4,22 +4,7 @@ richlistbox {
 	// Prevent horizontal scrollbar on Windows
 	overflow-x: hidden;
 
-	// Following rules from https://searchfox.org/mozilla-central/source/toolkit/themes/windows/global/richlistbox.css
 	richlistitem {
-		--listitem-selectedBackground: var(--color-quarternary-on-sidepane);;
-		--listitem-selectedFocusCurrentBackground: var(--color-accent);
-		--listitem-selectedTextColor: var(--color-accent-text);
-		--listitem-selectedBorder: transparent;
-		--listitem-selectedBottomBorder: transparent;
-		--listitem-selectedCurrentBorder: transparent;
-		--listitem-selectedFocusBorder: transparent;
-		--listitem-selectedFocusBottomBorder: transparent;
-		--listitem-selectedFocusCurrentBorder: transparent;
-		--listitem-selectedFocusBackground: var(--color-accent);
-		--listitem-selectedFocusColor: var(--accent-white);
-		--listitem-selectedColor: var(--accent-white);
-		--listitem-selectedFocusCurrentBottomBorder: transparent;
-
 		padding: 0.2em 0.4em;
 		align-items: center;
 
@@ -41,38 +26,19 @@ richlistbox {
 			width: 100%;
 		}
 
+		// Selected but not focused
 		&[selected="true"] {
-			border-top-color: var(--listitem-selectedBorder);
-			border-inline-color: var(--listitem-selectedBorder);
-			border-bottom-color: var(--listitem-selectedBottomBorder);
-			background-image: var(--listitem-selectedImage);
-			background-color: var(--listitem-selectedBackground);
-			color: var(--listitem-selectedTextColor);
+			background-color: var(--color-quarternary-on-sidepane);
+			color: var(--color-accent-text);
 		}
 	}
 
 	&:focus {
 		& > richlistitem {
+			// Selected and focused
 			&[selected="true"] {
-				border-top-color: var(--listitem-selectedFocusBorder);
-				border-inline-color: var(--listitem-selectedFocusBorder);
-				border-bottom-color: var(--listitem-selectedFocusBottomBorder);
-				background-image: var(--listitem-selectedFocusImage);
-				background-color: var(--listitem-selectedFocusBackground);
-				color: var(--listitem-selectedTextColor);
-			}
-
-			&[current="true"] {
-				border-color: var(--listitem-selectedCurrentBorder);
-				outline: none;
-			}
-
-			&[selected="true"][current="true"] {
-				border-top-color: var(--listitem-selectedFocusCurrentBorder);
-				border-inline-color: var(--listitem-selectedFocusCurrentBorder);
-				border-bottom-color: var(--listitem-selectedFocusCurrentBottomBorder);
-				background-color: var(--listitem-selectedFocusCurrentBackground);
-				outline: none;
+				background-color: var(--color-accent);
+				color: var(--color-accent-text);
 			}
 		}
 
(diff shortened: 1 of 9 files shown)


windingwind:
We no longer use the `theme-listbox` class, as we maintain a set of rules ourselves now.

Linux:

image

image


Use background color from redesign when no focus:

image

image

seltype=multiple has focus ring (scaffold tests)

image

image

Windows:

image

image

image


image

image

image


MacOS:

image

image

image

image

image

image

scss/components/_richlistbox.scss
```diff
@@ -18,24 +19,70 @@ richlistbox.theme-listbox {
...
+				outline: none;
+			}
+		}
+
+		// Add focus ring to richlistbox if the richlistbox is multi-select
+		&[seltype="multiple"] > richlistitem {
+			&[selected="true"][current="true"] {
+				outline: var(--default-focusring);
+				outline-color: currentColor;
+				outline-offset: calc(-1 * var(--default-focusring-width));
```
adomasven, line 84: As far as I can tell this is the only rule we actually need in terms of multiselect outline. Do we need all those other rules relating to borders, or setting `outline: none`? Isn't `outline: none` the default there anyway? Disabling the border rules in the debugger doesn't seem to change the look of these elements.

Basically, can we simplify this scss file and cut it down to just the rules that actually do something?

adomasven:
The focusring width is different in richlistbox and virtualized table (item tree, style selector in preferences). We should probably just set `--default-focusring-width: 1px`, and also change virtualized table scss outline to use `--default-focusring`.

adomasven:
Vertical alignment seems wrong in scaffold for me.

image

scss/components/_richlistbox.scss
windingwind: Most of the rules are inherited from Mozilla code. We can simplify, of course, as their CSS isn't actively maintained and can have duplication parts. But might it be easier to maintain by keeping them as they are?

adomasven: If they are from Mozilla code, then we're getting them by including `global/skin` anyway, so we're needlessly duplicating here anyway, right?

These rules definitely make this code harder to maintain. They don't actually do any relevant styling, they are just noise in this code, and when we need to change something about the focusring in 2 years and everyone's forgotten what's going on here, that person will most likely have to retest these border rules and see if they are actually relevant to the focusring or not.

windingwind:
> Vertical alignment seems wrong in scaffold for me.

Fixed, thanks

scss/components/_richlistbox.scss
windingwind: They are from Mozilla code, the `richlistbox.theme-listbox`. Their css variables should be the simplest way to customize most of the rules.

> These rules definitely make this code harder to maintain. They don't actually do any relevant styling, they are just noise in this code, and when we need to change something about the focusring in 2 years and everyone's forgotten what's going on here, that person will most likely have to retest these border rules and see if they are actually relevant to the focusring or not.

Hmm, sorry I'm not sure if I understand correctly. Most of the variables are doing what their name indicates they should do, and only the focus ring is added as an extra patch. We could add some comments there, if necessary. Since we are just copying the Mozilla code, I don't expect there would be such a difficulty.

windingwind: To clarify, Mozilla has two sets of rules for this richlistbox thing: one is the default one, which does not provide css variables and cannot be customized easily, and is definitely not what we want; one is the `theme-listbox` class, which provides css variables but we do not want to use that class name.

So from my view, the most reliable and easiest way to do it is copying rules with css variables from the `theme-listbox` rule sets and replace the variables with ours. Since that does not handle the focus ring correctly, I added lines you quoted above.

adomasven: > Their css variables should be the simplest way to customize most of the rules.

I don't think use or lack of css variables themselves is relevant to whether we want to use something, or whether it can be consider good or bad. We want code that is readable, understandable and clear in what it does. Neither overly verbose, nor underdocumented or unclear.

I can see that the `_richlistbox.scss` has 70% of style declarations and variables that do not actually do any relevant styling on any platform. That is bad. There is superflouous use of variables, where they are only used once, so instead of reading the style declaration directly, you need to go reference some variable declared somewhere else, hurting readability. Further there are undeclared variables being used, and rules being applied that make no sense, like `background-image` for list items (I imagine this could be used for a gradient, but we're not doing that).

Also, I am looking at fx115 richlistbox.css, and it doesn't have the declarations with variables at all, so they're some addition between fx115 and the current firefox (where you found them on searchfox) added along with who-knows-what other changes to richlistbox or other fx UI.

Unless you have compelling reason for why the above is good, or can be beneficial in the future that outweighs the increased complexity of this stylesheet, we should simplify it.

windingwind: OK, just updated the pr.

Some replies:

> Also, I am looking at fx115 richlistbox.css, and it doesn't have the declarations with variables at all, so they're some addition between fx115 and the current firefox (where you found them on searchfox) added along with who-knows-what other changes to richlistbox or other fx UI.

Well, it's in fx115: https://searchfox.org/mozilla-esr115/source/toolkit/themes/windows/global/richlistbox.css#63-77. It's added after fx60, at least starting from fx68.

> Unless you have compelling reason for why the above is good, or can be beneficial in the future that outweighs the increased complexity of this stylesheet, we should simplify it.

I don't have a preference between following Mozilla but with duplications, or simpler but maintaining by ourselves. I thought if we follow the implementation of the upstream, we'd better keep the rules untouched if possible. The benefit of following their implementation is we can follow their future changes easily.

Since we have decided not to follow the Mozilla implementation, that's fine. We have already used our custom rules everywhere.

windingwind:
> and also change virtualized table scss outline to use --default-focusring.

Are we using the focus ring for our v-table? I didn't see them in the tables in the prefs window - cite - style manager.

scss/components/_richlistbox.scss
adomasven: > Well, it's in fx115: https://searchfox.org/mozilla-esr115/source/toolkit/themes/windows/global/richlistbox.css#63-77. It's added after fx60, at least starting from fx68.

Sorry. I have these files set up locally in my IDE and I had updated them for FX115, but maybe not entirely correctly. 

> The benefit of following their implementation is we can follow their future changes easily.

Since Firefox presumably uses `<richlistbox>` without `theme-listbox` themselves, we're going to get updates via `skins/global` anyway, whereas copying those styles into our stylesheets makes them permanent until we manually change them.

adomasven:
> Are we using the focus ring for our v-table? I didn't see them in the tables in the prefs window - cite - style manager.

Yes. It's only available for multi-select v-table, and only visible via ctrl/cmd-arrow navigation.

windingwind:
We are using the `--default-focusring-width` in v-table now.

PR #4516, opened 2024-08-07
Fix window background overflow when no native titlebar

windingwind:


windingwind:
Before:

image

After:

image

PR #4534, opened 2024-08-08
Bidi improvements and fixes

### LINKED ISSUE #4517: RTL word displayed at end of title in items list
https://forums.zotero.org/discussion/116579/zotero-7-beta-114-display-issue-with-rtl-text-in-library-view

AbeJellinek:
- Guess item field direction based on the field itself, the item's language field, and the app locale
- Fix `editable-text` `dir` attribute and `text-align` style (apparently these haven't actually done anything since the redesign - oops)
- Slightly improve `editable-text` mouse selection behavior when value or alignment changes on focus
- Don't bother calculating anything if bidi utilities aren't enabled

Fixes #4517

### DIFF
--- a/chrome/content/zotero/elements/editableText.js
+++ b/chrome/content/zotero/elements/editableText.js
@@ -43,6 +43,8 @@
 		
 		_ignoredWindowInactiveBlur = false;
 		
+		_focusMousedownEvent = false;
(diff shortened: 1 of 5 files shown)


dstillman:
Merged

Issue #3959, opened 2024-04-11
Restore creator-field-mode button


dstillman:
https://forums.zotero.org/discussion/113471/z7-user-interface

A bunch of complaints like this, so I think we probably need to do this.

And then we should probably increase the pane min-width to 320px.

abaevbog:
Right, the main question is what would this look like? We could just add another switch button to the right. My concern with this is that 4 buttons looks a bit busy and it feels odd that about a 4th of the visible row is taken up by buttons (this is with future min-width of about 320px). Also buttons appear on hover and shift the creator names, so if the first name is long and you hover over any part of it's right half, your mouse ends up over one of the 4 buttons. This is the behavior now but I feel like it is not as pronounced with only 3 buttons.

(Duplicated `-` button just to see what spacing would look like)
<img width="387" alt="Screenshot 2024-04-30 at 10 43 51 PM" src="https://github.com/zotero/zotero/assets/36271954/9c289501-2a3e-46d3-8312-28f926106e16">

Every row but the last one has a disabled `+` button. I wonder if there's a way to rearrange buttons a bit to use up that space for the new button?

It also could very much be that I'm overthinking it after all the time spent with `itemBox` and 4 buttons are not an issue at all.

dstillman:
> Every row but the last one has a disabled `+` button. I wonder if there's a way to rearrange buttons a bit to use up that space for the new button?

Unfortunately that's a fairly common complaint, and, [as you noted](https://github.com/zotero/zotero/issues/3674), we should just make those all functional.

I did suggest that we could consider hiding the button at narrow pane widths, but I believe @yexingsha thought that was too weird. I do think it would be fine if the ⋯ menu were actually an overflow menu with a », but at least on macOS it's fairly non-standard to conditionally hide something within a ⋯ menu. I'm not sure if that icon is used that way more on other platforms. And, regardless, it's obviously a little weird to do it for a single icon, and I don't think we have any intention of adding additional icons for the other options in the menu even at much wider widths.

So I guess we could just ignore this for now, and see if there's more reaction after Z7 is out. (The other thing is that you can also right-click anywhere in the row and select "Switch to Single Field", so it's not like you actually have to click on the little button.)

abaevbog:
Gotcha! So we'll try to add it just as a 4th button for now and maybe rework it later if needed.

[...]

PR #4535, opened 2024-08-08
Itembox: bring back switch creator mode button

abaevbog:
- added button to switch mode to creator rows
- removed that option from options menu
- extra tweak to avoid cursor flickering during re-render after clicking on the button
- increased itemPane min-width to 320px

Fixes: #3959

A while back, there was a discussion of dynamically hiding this button when the itemPane is narrow and displaying this option in the menu. I wasn't sure if we still want it now since we do make the itemPane a bit wider to accommodate the extra added button (and I when I tried, it felt a bit strange because this is not done anywhere else). But correct me if I'm wrong and we should hide/display the button and options menu depending on the width.


abaevbog:
For reference, the extra tweak that adds and removes the `.show-on-hover` from buttons is to avoid the cursor blinking that would occur when the itemPane is narrow:


https://github.com/user-attachments/assets/2eda66cd-25ef-4096-9e63-46e99ba1eed9

abaevbog:
Also, I am noticing that visually the switch mode icon looks a bit not aligned with the rest of them.  

My impression is that it is caused by the actual `+` and `-` icons not being centered inside of their viewbox and being pushed a bit down.

Screenshot 2024-08-08 at 10 36 32 AM

Screenshot 2024-08-08 at 10 37 56 AM

We could just add a little margin to push the switch mode icon lower. Or is that something that's worth tweaking within actual svg files?

dstillman:
Merged in 7450622047b619232d2e21cf9ee10a5d1a9ff955 and 9dd3807683d06c2c010cb80a6f2f06580a5d9498

PR #4541, opened 2024-08-09
Item box: Remove hidden buttons from flex calculations

### LINKED ISSUE #4539: Item pane shows horizontal scrollbar
image

No matter how wide the window and pane are.

I think this is due to the creator button changes. The scrollbar goes away if you hover over a creator row.

AbeJellinek:
Somehow, even with 0 width, height, margin (tried adding that), padding, and border, the hidden buttons were still causing the flex layout algorithm to make the row overflow horizontally. `position: absolute` removes child elements from their parents' flex calculations, and that seems to fix the issue (whatever it is) here.

@abaevbog: Could you make sure this doesn't break anything with screen readers?

Fixes #4539

### DIFF
--- a/scss/abstracts/_mixins.scss
+++ b/scss/abstracts/_mixins.scss
@@ -139,6 +139,7 @@
 				width: 0;
 				height: 0;
 				padding: 0;
+				position: absolute;
 			}
 		}
 		.meta-data {


abaevbog:
It started happening after 794dc6bf014739be1f33a09e98b5076b35d0c852 when some svg icons became a bit wider. Odd!

In any case, it doesn't seem to affect the tab order, so I think it's all good!

The main thing with `.show-on-hover` was to have hidden buttons still be focusable by mozilla on shift-tab from the next field. And that looks unaffected here.

PR #4560, opened 2024-08-12
vpat 31 and 33 - aria tweaks for Tools > Plugins

abaevbog:
- vpat 31: mark which menuitems' button is checked, otherwise selected options (e.g. Update plugins automatically) are not announced as such
- vpat 33: when plugin details pane appears, remove a misleading role="tabpanel" from its section because it is not structured as a collection of tabs

---

Vpat issues for reference:

- 31: Description: In the Zotero desktop app's 'Add-ons Manager' modal, activating a link for an add-on opens more details about that add-on. The content following the initial section containing the add-on name is incorrectly announced as a tab panel by screen readers. However, the content is not structured as a tab panel and should not have this role.
- 33: Description: Throughout the Zotero desktop app, some menu items can be activated or deactivated, such as the 'Update Add-ons Automatically' option in the 'Add-ons Manager' modal. While these menu items display a checkmark to visually indicate their state, there are no non-visual indicators for screen readers. The menu items are described by screen readers in the same way regardless of their current state, without any "checked" or "pressed" indication. The active state should be communicated non-visually.

### DIFF
--- a/chrome/content/zotero/standalone/standalone.js
+++ b/chrome/content/zotero/standalone/standalone.js
@@ -724,7 +724,8 @@ const ZoteroStandalone = new function() {
 		}, true);
 
 		// A11y: after a click, check if the panel with plugin details appeared.
-		// If so, delete a misleading role=tabpanel
+		// If so, delete a misleading role=tabpanel because default firefox tabs
+		// ("Details" and "Permissions") are explicitly hidden in fetch_xulrunner
 		doc.addEventListener("click", (_) => {
 			setTimeout(() => {
 				let details = doc.querySelector("#details-deck section[role='tabpanel']");
--- a/chrome/content/zotero/standalone/standalone.js
+++ b/chrome/content/zotero/standalone/standalone.js
@@ -715,6 +715,23 @@ const ZoteroStandalone = new function() {
 			);
 			return { remove: result === 0, report: null };
 		};
+		// A11y: when a popup appears, mark which buttons are checked for screen readers
+		doc.addEventListener("shown", (event) => {
+			for (let item of [...event.target.querySelectorAll("panel-item")]) {
+				item.button.setAttribute("role", "menuitemcheckbox");
+				item.button.setAttribute("aria-checked", item.checked);
+			}
+		}, true);
+
+		// A11y: after a click, check if the panel with plugin details appeared.
+		// If so, delete a misleading role=tabpanel
+		doc.addEventListener("click", (_) => {
+			setTimeout(() => {
+				let details = doc.querySelector("#details-deck section[role='tabpanel']");
+				if (!details) return;
+				details.removeAttribute("role");
+			});
+		}, true);
 	}
 	
 	/**


abaevbog:
For comparison, this is before. Notice that the checked `menuitem` is not announced as "checked" and that the details section is incorrectly announced as "tabpanel" .


https://github.com/user-attachments/assets/14226133-406c-4371-88e9-bdaf627afffa

abaevbog:
And this is after:

https://github.com/user-attachments/assets/c48d06da-ef9d-49bd-a7a1-03891d5cdf26

chrome/content/zotero/standalone/standalone.js
```diff
@@ -715,6 +715,23 @@ const ZoteroStandalone = new function() {
...
+		// A11y: when a popup appears, mark which buttons are checked for screen readers
+		doc.addEventListener("shown", (event) => {
+			for (let item of [...event.target.querySelectorAll("panel-item")]) {
+				item.button.setAttribute("role", "menuitemcheckbox");
+				item.button.setAttribute("aria-checked", item.checked);
+			}
+		}, true);
+
+		// A11y: after a click, check if the panel with plugin details appeared.
+		// If so, delete a misleading role=tabpanel
```
dstillman, line 727: Should clarify that we're hiding the tab bar in fetch_xulrunner — that's why it's not a tab.

chrome/content/zotero/standalone/standalone.js
```diff
@@ -715,6 +715,23 @@ const ZoteroStandalone = new function() {
 			);
 			return { remove: result === 0, report: null };
 		};
+		// A11y: when a popup appears, mark which buttons are checked for screen readers
+		doc.addEventListener("shown", (event) => {
+			for (let item of [...event.target.querySelectorAll("panel-item")]) {
+				item.button.setAttribute("role", "menuitemcheckbox");
+				item.button.setAttribute("aria-checked", item.checked);
+			}
+		}, true);
```
dstillman, line 724: Does 33 happen in Firefox (115 or current)? We shouldn't have to set this manually every time we use a checked menu item.

chrome/content/zotero/standalone/standalone.js
abaevbog: Oh, yeah. I actually didn't realize that it's explicitly hidden [here](https://github.com/zotero/zotero/blob/d879c507a55bd7060699a943a26554f633fd24a4/app/scripts/fetch_xulrunner#L290). Good to know, I'll add a comment.

chrome/content/zotero/standalone/standalone.js
abaevbog: Yes, I'm on firefox 129.0 and if I go to `about:addons`, click on the "Tools for all add-ons" and tab to a checked "Update Add-ons Automatically", it still does not say that it's checked.

This has not been an issue for all menu items. All other regular menuitems within menus that I tried behave fine. For instance, top-level menus `View > *`, tag selector "Actions" menu, quickSearch mode selector - their checked  menuitems are announced as checked. So I think it's just this menu, and I suppose because it's a `panel-item`, and not a `menuitem` component that we use.

I am not sure if the "Throughout the Zotero desktop app" part of the vpat 33 comment implies that there are other instances of this happening. But I assume that if this were to happen elsewhere, it would have been brought up in other issues.

Issue #4531, opened 2024-08-08
Two item panes occasionally appear at startup


dstillman:
https://forums.zotero.org/discussion/116606/z7-beta-two-item-panes-after-startup

@windingwind, I think you had an idea for what might be causing this?

windingwind:
What's their Zotero version? It should have been fixed. I may need their debug output

dstillman:
7.0.0-beta.115+4c32317ba

windingwind:
OK, managed to reproduce. Will fix that.

PR #4561, opened 2024-08-13
Fix two item panes occasionally appear at same time

windingwind:
fix: #4531

### DIFF
--- a/chrome/content/zotero/elements/contextPane.js
+++ b/chrome/content/zotero/elements/contextPane.js
@@ -181,11 +181,6 @@
 				_contextPaneSplitter.setAttribute('hidden', false);
 
 				_contextPane.setAttribute('collapsed', !(_contextPaneSplitter.getAttribute('state') != 'collapsed'));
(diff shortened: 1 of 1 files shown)


chrome/content/zotero/elements/contextPane.js
```diff
@@ -181,11 +181,6 @@
 				_contextPaneSplitter.setAttribute('hidden', false);
 
 				_contextPane.setAttribute('collapsed', !(_contextPaneSplitter.getAttribute('state') != 'collapsed'));
-				// It seems that on heavy load (i.e. syncing) the line below doesn't set the correct value,
-				// therefore we repeat the same operation at the end of JS message queue
-				setTimeout(() => {
-					_contextPane.setAttribute('collapsed', !(_contextPaneSplitter.getAttribute('state') != 'collapsed'));
```
windingwind, line 187: I believe this line is causing all the weirdness of the context pane/item pane. After #4289, we always use the correct `tabType` to decide which item pane to display and thus no longer need this trick.

PR #4563, opened 2024-08-13
Show a different error message for WinWord when no document is found

adomasven:
image

Links to https://www.zotero.org/support/kb/could_not_find_a_running_word_instance


adomasven:
Closes https://github.com/zotero/zotero-word-for-windows-integration/issues/75

adomasven:
Also running Zotero as Administrator no longer seems to cause issues, and running Word as Administrator throws this error (previous behaviour was buttons unresponsive).

Issue #4393, opened 2024-07-16
Shift-Enter from itemBox creator row should add an empty creator row after current row


abaevbog:
As opposed to focus the next row (if no changes happened) and add an empty creator row in the very end otherwise.
This is true now that empty creator rows can be added not only in the very end.
Per: https://github.com/zotero/zotero/pull/4231#issuecomment-2230079254

PR #4567, opened 2024-08-13
itembox: shift-enter adds new creator + bug fixes

abaevbog:
- shift enter on a creator row will add a new empty creator row after focused row. If shift-Enter is on the last creator before "More creators", all creators will be rendered.
- fixed encountered glitch where switching the mode of creator would not always update the pref, so next time a new creator row is added, it would not be of correct mode.
- fixed encountered bug where a newly added creator row could receive an index of an existing creator that is not rendered (when "More creators" label is displayed). In that case, saving such creator would override an existing creator and erasing that creator would remove invisble creators.
- added a few tests for these edge cases

Fixes: #4393
Fixes: #4710

### DIFF
--- a/chrome/content/zotero/elements/itemBox.js
+++ b/chrome/content/zotero/elements/itemBox.js
@@ -815,14 +815,6 @@
 					// this item, so that added creators aren't
 					// immediately hidden
 					this._displayAllCreators = true;
-					
-					if (this._addCreatorRow !== false) {
-						// Insert an empty creator row in a specified location
-						let beforeCreator = this.querySelector(`#itembox-field-value-creator-${this._addCreatorRow}-lastName`);
-						let beforeRow = beforeCreator?.closest(".meta-row") || null;
-						this.addCreatorRow(false, this.item.getCreator(max - 1).creatorTypeID, true, beforeRow);
-						this._addCreatorRow = false;
-					}
 				}
 			}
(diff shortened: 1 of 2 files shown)


abaevbog:
> The empty creator row should go away (as it does when you focus another field instead of clicking the pane background), but it stays.

This was a somewhat intentional workaround https://github.com/zotero/zotero/pull/4178#issue-2323915256 to the issue of the `editable-text` loosing focus when you click on the options button to open up a menu. Removing creator on every blur would mean that if you press `+` to add a new empty creator and then click on the options button, the creator row would blur and immediately go away. So removing it if focus goes into another field or out of `itemBox` completely felt like a reasonable compromise. 

Having looked at it a bit more, I believe this behavior can be attributed to to an extra [tabindex](https://github.com/zotero/zotero/blob/c59c8cbffe22d0945a52b2474c1acf8dc5e1fe06/chrome/content/zotero/elements/itemDetails.js#L52) on the `zotero-view-item`. Without it, if an `editable-text` is focused, opening up a menu would not blur it. Would that be better? On one hand, it would certainly make that `blur` situation more straightforward and not require any workarounds like above. On the other hand, I feel like it would feel a bit strange if you have a cursor in one `editable-text` field and you can then open up any menu, even for a different field. I suppose, we could explicitly blur `editable-text` in that instance.

PR #4579, opened 2024-08-16
Throw an error when an update to word processor plugin fails to install

### LINKED ISSUE #4566: Show error if word processor plugin can't be automatically upgraded
A bunch of people have non-working plugins after the Z7 update with no explanation why. We should add a warning and then push an update with a new version number so everyone gets an upgrade.

(Doing this for Word is the priority.)

adomasven:
- This commit will retrigger the installation for all word processor
  plugins. For the majority of users this will succeed silently.
  - It will fail with an error for Windows users that have Word open.
  - It will succeed with a warning for macOS users that have Word open saying they should restart Word.
- The updated error prompt links to the [new documentation page](https://www.zotero.org/support/kb/word_processor_plugin_installation_error) for failed plugin installs

Closes #4566

In testing this I also discovered that automatic winword plugin updating was just completely broken in Zotero 7, hence, most likely, the reports.

### DIFF
--- a/chrome/content/zotero/xpcom/integration.js
+++ b/chrome/content/zotero/xpcom/integration.js
@@ -95,7 +95,7 @@ Zotero.Integration = new function() {
 		if (Zotero.isMac) {
 			entryPoints.push(PLUGIN_PATHS.MacWord);
 		}
-		else if (Zotero.isWindows) {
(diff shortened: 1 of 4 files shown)


adomasven:
Updated prompt:

image

dstillman:
Merged

Issue #4583, opened 2024-08-18
localize-ftl doesn't handling conversion from attribute to raw value


dstillman:
I think localize-ftl is broken because of these changes in 4f1e44978ebae977b13fc56644667daab4563c3d:

```diff
-rtfScan-intro-page = 
-   .label = Introduction
+rtfScan-intro-page = Introduction
```

I'm now getting this:

```
Error: TypeError: Failed to localize "af-ZA": Cannot read properties of undefined (reading 'type')
```

And if I ignore the error, it leaves the `.label` lines in place.

tnajdek:
I'm on it, I'll push a fix shortly.

tnajdek:
@dstillman I've fixed `localize-ftl` crashing and now it will show a warning for any messages that are in the JSON but not in the base FTL. However, I don't think we can automatically migrate these because, technically, they could be entirely different messages (even though in this case, they're not).

tnajdek:
Though it should be relatively simple to write a quick one-off script to migrate these few strings in all localized .ftl files, let me try to come up with something.

tnajdek:
@dstillman Please merge the above PR before running the localization scripts to preserve translated messages in cases where we've moved from an attribute to a raw value

PR #4584, opened 2024-08-18
Fix localized labels after 4f1e449

tnajdek:
In #4583, I've fixed the issue with `localize-ftl` crashing after 4f1e449. However, the translated messages would have been lost when running `localize-ftl`. This PR preserves the translated messages.


tnajdek:
For reference, here is a one-off script I've used to generate this. If expanded a little to take a lookup and not a set, this could be useful in a future as more generic tool to migrate translated messages.

````js
import fs from 'fs-extra';
import { parse, serialize } from "@fluent/syntax";
import { dirname, join } from 'path';
import { fileURLToPath } from 'url';
import { ftlFileBaseNames as sourceFileBaseNames } from './config.js';
import { onError, onProgress, onSuccess } from './utils.js';

const ROOT = join(dirname(fileURLToPath(import.meta.url)), '..');
const localesDir = join(ROOT, 'chrome', 'locale');

const labelsToMigrate = new Set([
	'rtfScan-intro-page.label',
	'rtfScan-scan-page.label',
	'rtfScan-citations-page.label',
	'rtfScan-style-page.label',
	'rtfScan-format-page.label',
	'rtfScan-complete-page.label',
	'publications-intro-page.label',
	'publications-sharing-page.label',
	'publications-license-page.label',
]);

function getLocaleDir(locale) {
	return join(localesDir, locale, 'zotero');
}

async function getFixLabels() {
	const t1 = performance.now();

	const foundLocales = (await fs.readdir(localesDir, { withFileTypes: true }))
		.filter(dirent => dirent.isDirectory())
		.map(dirent => dirent.name)
		// Valid locale codes only
		.filter(name => /^[a-z]{2}(-[A-Z]{2})?$/.test(name));

	let count = 0;
	for (let locale of foundLocales) {
		// Skip source locale
		if (locale == 'en-US') {
			continue;
		}
		for (let sourceFileBaseName of sourceFileBaseNames) {
			const ftlFilePath = join(getLocaleDir(locale), sourceFileBaseName + '.ftl');
			const ftlText = await fs.readFile(ftlFilePath, 'utf8');
			const res = parse(ftlText);
			for (let entry of res.body) {
				if (entry.type !== 'Message') {
					continue;
				}
				if (entry.attributes.length !== 1) {
					continue;
				}
				if (entry.value !== null) {
					continue;
				}

				const JSONKey = `${entry.id.name}.${entry.attributes[0].id.name}`;
				if (!labelsToMigrate.has(JSONKey)) {
					continue;
				}

				entry.value = entry.attributes[0].value;
				entry.attributes = [];
			}
			let migratedFTL = serialize(res);
			await fs.writeFile(ftlFilePath, migratedFTL);
			count++;
			onProgress(`${locale}/${sourceFileBaseName}.ftl`, null, 'fix-labels');
		}
	}

	const t2 = performance.now();
	return ({
		action: 'fix-labels',
		count,
		totalCount: count,
		processingTime: t2 - t1
	});
}


if (process.argv[1] === fileURLToPath(import.meta.url)) {
	(async () => {
		try {
			onSuccess(await getFixLabels());
		}
		catch (err) {
			process.exitCode = 1;
			global.isError = true;
			onError(err);
		}
	})();
}
````

dstillman:
Thanks for writing this!

PR #4601, opened 2024-08-20
Add option to show filenames in the item tree and "Normalize Attachment Titles..." tool

### LINKED ISSUE #4587: Add menu option to reset attachment titles
https://www.zotero.org/support/file_renaming#attachment_title_handling_in_zotero_7

AbeJellinek:
And move item tree pref observer into the item tree itself.

Closes #4587

### DIFF
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -2974,9 +2974,6 @@ var ItemTree = class ItemTree extends LibraryTree {
 		}
 		let textSpanAriaLabel = [textWithFullStop, itemTypeAriaLabel, tagAriaLabel, retractedAriaLabel].join(' ');
 		textSpan.className = "cell-text";
-		if (this._rowCache[item.id].titleIsFilename) {
-			textSpan.classList.add('filename');
-		}
 		if (lazy.BIDI_BROWSER_UI) {
 			textSpan.dir = Zotero.ItemFields.getDirection(
 				item.itemTypeID, column.dataKey, item.getField('language')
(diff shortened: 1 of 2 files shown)


AbeJellinek:
@dstillman: Rebased

dstillman:
Something like this for the dialog?

Screenshot 2025-12-16 at 3 59 18 PM

("More Information" opens https://www.zotero.org/support/kb/attachment_title_vs_filename)

AbeJellinek:
Looking good. Maybe "titles that have been changed" or "modified titles" instead of "custom titles"? "Custom titles" sounds like some kind of customization feature to me.

And "More Information…", since it opens an external link?

dstillman:
OK, how about "Attachment titles that have been manually changed will not be affected."?

Screenshot 2025-12-16 at 11 18 51 PM

(Don't need an ellipsis just to open a browser window — only to convey that there will be an intermediate step before some action occurs. E.g., in the Firefox Help menu, "Share Ideas and Feedback…" but not "More Troubleshooting Information".)

dstillman:
"Normalize Attachment Titles" is a bit technical, but I don't have a better suggestion. And realistically, people are probably mostly going to learn about this option after first ending up on the wiki page.

AbeJellinek:
Yeah, that looks good.

I guess we could do “Update Automatic Attachment Titles” or something like that, but it’s wordy and not really clearer.

dstillman:
A little clearer (to the extent that people know what "primary attachment" means):

dstillman:
Merged in 2d79f1f9596c36074ff57dfec707d017e8ae2dbb and aec284b0b8cc08753f4a08c7cb9636796735b131

We can tweak the wording further if we want.

Issue #4210, opened 2024-06-06
Quick Format dialog is resizable on Wayland


dstillman:
https://github.com/zotero/zotero/pull/4195#issuecomment-2145033795

PR #4603, opened 2024-08-20
Qf: potential fix to the dialog becoming invisible

abaevbog:
I could not reproduce this issue https://forums.zotero.org/discussion/117023/disappearing-add-citation-bar, however in the past I ran into `background: transparent` causing problems on windows, so that is likely the culprit https://github.com/zotero/zotero/pull/3889#issuecomment-2035228247. Last time, we went with very specific chromemargin values that I [stumbled upon ](https://github.com/zotero/zotero/pull/3889#discussion_r1547869959) which seemed to have helped. It looks like it helped in some but not all cases, so maybe we try to get rid of `background: transparent` now?

This gets rid of  `background: transparent` and `chromemargins` on windows. We also have to remove `border-radius` on windows because otherwise the corners of the actual window are now visible around the QF dialog. But the window itself has rounded corners so I think it doesn't look that different.

Unfortunately, the window without transparent background and with `drawintitlebar` becomes resizable, disregarding `resizable=false` option. I believe, by [design](https://bugzilla.mozilla.org/show_bug.cgi?id=177838). The best solution I could come up with is to set  max and min `height` and `width` so that the dialog cannot be stretched. This also fixes the dialog being resizable on linux: #4210. If you hover over an edge of the window, though, there will appear a resizing cursor (that does nothing), and I am not sure if we can remove it.

In addition, I ran into the issue of accept button and spinner not being visible with the dark theme, as well as the `richlistitem` having it's color overriden by another stylesheet. So I added a few hard-coded values to work around it for now.

### DIFF
--- a/chrome/content/zotero-platform/win/integration.css
+++ b/chrome/content/zotero-platform/win/integration.css
@@ -1,5 +1,4 @@
 window.citation-dialog {
-	background: transparent;
 	padding: 0;
 	-moz-appearance: none;
 }
@@ -8,16 +7,8 @@ window.citation-dialog {
 .citation-dialog.entry {
 	background: -moz-linear-gradient(-90deg, rgb(243,123,119) 0, rgb(180,47,38) 50%, rgb(156,36,27) 50%);
 	padding: 10px;
-	/* mouse events on win don't fire right at the edge of the window. The dialog has a gap
-	before the window edges to ensure the window can be dragged from any point of the dialog */
-	margin:10px; 
 }
 
 .note-dialog .citation-dialog.entry {
 	background: -moz-linear-gradient(-90deg, rgb(249, 231, 179) 0, rgb(228, 193, 94) 50%, rgb(221, 184, 81) 50%);
-}
-
-.citation-dialog.entry:not([square="true"]) {
-	-moz-border-radius: 15px;
-	border-radius: 15px;
 }
--- a/chrome/content/zotero/integration/insertNoteDialog.xhtml
+++ b/chrome/content/zotero/integration/insertNoteDialog.xhtml
@@ -41,9 +41,12 @@
 	xmlns="http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul"
 	persist="screenX screenY"
 	onkeypress="Zotero_QuickFormat.onWindowKeyPress(event)"
-	onunload="Zotero_QuickFormat.onUnload()">
+	onunload="Zotero_QuickFormat.onUnload()"
+	drawintitlebar-platforms="linux,win"
+	no-titlebar-icon="true">
 	
 	<script src="../include.js"/>
+	<script src="../titlebar.js" type="text/javascript"/>
 	<script src="quickFormat.js" type="text/javascript"/>
 	<script src="insertNoteDialog.js" type="text/javascript"/>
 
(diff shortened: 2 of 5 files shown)


adomasven:
I went back to #3889 and temporarily reverted it on the current master to see if we can fix the panel positioning without all the complicated code there. It seems that the panel will follow its anchor if the attribute `type="arrow"` is set. I couldn't find documentation for why that works. Other panels in Zotero that follow its anchor do set `type="arrow"` though. Given the above, can we remove some of the complications introduced in #3889? Perhaps we can set the `chromemargin` value that we use on the main window and fix this in a more fundamental level? I'm not sure if the `chromemargin` change there was related to the dragging behaviour or just a fix for fx102/115.

abaevbog:
>  It seems that the panel will follow its anchor if the attribute type="arrow" is set.

Oh yeah! Not sure which part of the `type="arrow"` behavior (or a number of additional attributes it [sets](https://searchfox.org/mozilla-central/source/toolkit/content/widgets/panel.js#36)) is what helps here but with this we can drop the part where we hide the reference panel whenever the window is being dragged. So that would reduce some complexity.

> Perhaps we can set the chromemargin value that we use on the main window and fix this in a more fundamental way? I'm not sure if the chromemargin change there was related to the dragging behaviour or just a fix for fx102/115.

The `chromemargin` hack was required to get `-moz-window-dragging: drag` to handle dragging of the window for us instead of using the less-reliable `WindowDraggingElement`.  `-moz-window-dragging` though is a bit picky as well. My understanding (just via trial and error) is that it either requires `drawintitlebar="true"` OR a `chromemargin` to work. If we set `chromemargin="0,2,2,2"`(same as on the main window), it ends up [shrinking](https://www.dropbox.com/scl/fi/9dyp9n52ml03uohz7nu36/panel_gone.mov?rlkey=wcp8csgprfg8oety9ec4yg1uh&dl=0) the window, which is where the hacky `chromemargin="0,0,15,0"` came from. If, with default `chromemargin="0,2,2,2"` set, we remove `background: transparent` from the window, that strange window disappearance stops. That made me think that this issue boils down to `background: transparent` and removing it is the fundamental way to address this (as opposed to hacky `chromemargin="0,0,15,0"` that worked in, apparently, only some cases). This is why here, I dropped the transparency, which (unfortunately) makes the window resizable, so I added height and width limits to somewhat counteract that.

 All of the above is not affected by the `<panel>` setup: say, completely deleting the ` reference-panel` would still have the window disappear with `chromemargin="0,2,2,2"`.

abaevbog:
`type="arrow"` allows us to drop quite a few pieces where we hide and display the panel to keep it next to the dialog. So that's great! I checked that it acts as expected on a mac and windows. Will need check on linux later today when I get to the other laptop.

abaevbog:
The panel is behaving well on linux as well, so that's nice!

abaevbog:
Another report: https://forums.zotero.org/discussion/117253/disappearing-citation-bar-in-word

adomasven:
Let's separate the `type="arrow"` changes into a separate PR and merge this.

abaevbog:
I pulled that last commit into a separate PR. Really hoping that this will help with the disappearing window...

abaevbog:
I never heard from those who initially reported this issue if these changes actually fixed it. But per https://forums.zotero.org/discussion/118114/word-plugin-citation-dialog-slowly-disappears-with-keystrokes, it looks like this really did solve the problem.

PR #4609, opened 2024-08-23
Checkbox-and-radio-redesign-on-Windows

windingwind:
resolve: #4570


yexingsha:
I can't seem to build the windows app on WSL, it gets stuck on `rm: cannot remove '/home/yexing/zotero-client/app/staging/Zotero_win32/integration/word-for-windows/libzoteroWinWordIntegration_ARM64.dll': No such file or directory`. Can you drop some screenshots so I can take a look?

windingwind:
That might be a bug with the build script, I'm not sure. Will share some screenshots :)

dstillman:
`git submodule update --init --recursive` might fix that error. I don't think that should happen unless you have an empty or outdated submodule. (But also I've never tried to build on WSL — I just build on macOS and run it via a shared folder. We have a script to automate that via Parallels.)

windingwind:
https://github.com/user-attachments/assets/ff70334d-12e7-4ab7-8292-32fead8ea4cc


https://github.com/user-attachments/assets/6bdde9d9-c1bb-4670-871a-ee66a8309411

windingwind:
If the `label` is empty, the focus ring is only for the icon:


https://github.com/user-attachments/assets/f46980ee-69d6-4d67-bf20-31d752621c55

yexingsha:
The appearances of the checkbox and radio buttons look fine to me, just a few small issues:

1. The gap between the checkbox/radio and the label should be 8px. It's too large right now, looks like 12px?

2. The entire checkbox/radio + label is positioned too high. You can see it clearly in the "Color Scheme" line, where they are 2px higher than the label at the start.

3. The checkbox/radio + label should be visually aligned to the left and right edges of the headers and text descriptions. So we need some negative horizontal margins to offset the padding and border of the focus ring.

windingwind:
Thanks :)

windingwind:
image

image

yexingsha:
Looks great!

This is somewhat related, but the buttons have a similar problem where they don't left align with the start of a line. Also, their labels are sometimes misaligned with other labels in the line (e.g. the first line under "Miscellaneous").

dstillman:
Just out of curiosity, where are we getting that radio-button style from? I'm seeing various different styles elsewhere.

Chrome:

Screenshot 2024-08-29 at 1 36 27 AM

Firefox:

Screenshot 2024-08-29 at 1 35 59 AM

Edge:

Screenshot 2024-08-29 at 1 37 02 AM

I guess it's Word?

Screenshot 2024-08-29 at 1 38 06 AM

Any particular reason to not use the filled style from Chrome/Firefox? Particularly since we're using a larger widget size, this way just seems slightly more confusing for no particular reason. (Edge is the worst by far — unfilled and grayed out)

windingwind:
Fixed:

image

image

yexingsha:
Looks great!

@dstillman this style is [windows UI 3](https://learn.microsoft.com/en-us/windows/apps/design/controls/radio-button), which I guess is still very loosely enforced even in native windows apps. One of the apps that apply this style consistently is Settings, where you can see radio buttons in e.g. Personalization > Start.

PR #4615, opened 2024-08-23
Add new status info below local API checkbox

### LINKED ISSUE #4612: Show current server port under "Allow other applications on this computer to communicate with Zotero" when enabled?
And disable the checkbox when `httpServer.enabled` is false? I don't know if anyone turns that off, but it would break the local API.

Re this thread: https://forums.zotero.org/discussion/117119/is-allow-other-applications-on-this-computer-to-communicate-with-zotero-working-post-7-0 (though I don't know if any of this is causing the issue there)

AbeJellinek:
image
image

And tweak `synctopreference` event behavior.

Closes #4612

### DIFF
--- a/chrome/locale/en-US/zotero/preferences.ftl
+++ b/chrome/locale/en-US/zotero/preferences.ftl
@@ -71,6 +71,10 @@ preferences-styleManager-add-button =
 
 preferences-advanced-enable-local-api =
     .label = Allow other applications on this computer to communicate with { -app-name }
+preferences-advanced-local-api-running = Running on <code data-l10n-name="url">{ $url }</span>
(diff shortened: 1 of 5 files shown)


chrome/locale/en-US/zotero/preferences.ftl
```diff
@@ -71,6 +71,10 @@ preferences-styleManager-add-button =
 
 preferences-advanced-enable-local-api =
     .label = Allow other applications on this computer to communicate with { -app-name }
+preferences-advanced-local-api-running = Running on <code data-l10n-name="url">{ $url }</span>
```
dstillman, line 74: "Available at"? "Running at"? "Running on" makes sense for the host/port, but maybe not with a full URL?

dstillman:
Merged

Issue #4614, opened 2024-08-23
macOS menulist: Don't show selected item background when not hovered/keyboard-selected?


AbeJellinek:
I have no idea if this is possible without breaking keyboard selection. But one thing that makes non-native menulist popups feel slightly clunkier than native menus on macOS is that the selected item background persists when the mouse pointer leaves the menu.

Native menus don't do that - the selected item background goes away when you move the mouse away from the menu, and it shows again if you then use the keyboard to move through the menu.

Would be nice to make our non-native menus do that too if possible. Very minor, but the current behavior makes the UI feel less responsive.

PR #4619, opened 2024-08-24
No menu background when not hovered on MacOS

windingwind:
fix: #4614

### DIFF
--- a/scss/mac/_menupopup.scss
+++ b/scss/mac/_menupopup.scss
@@ -34,6 +34,15 @@ menupopup {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


windingwind:
https://github.com/user-attachments/assets/54ef773b-d1c4-45f6-9b31-764615ad83f1

dstillman:
Reverted. (Sorry, didn't test this one.)

windingwind:
Sorry didn’t test that. I think it will require some CE hack to make it work with keyboard navigation, could be too complicated

PR #4626, opened 2024-08-26
Tweak the uninstaller message

tnajdek:
I've reverted the message displayed by the helper/uninstaller to what we had before ca83e43, so that only the installer suggests that something will happen when the user selects 'No'.


tnajdek:
@dstillman actually, I think it should have said "upgrade" instead of "uninstall", sorry about that - I got confused. I'll open another PR.

PR #4633, opened 2024-08-27
Add an option to disable alternating row colors

### LINKED ISSUE #4628: Add option to disable alternating rows
https://forums.zotero.org/discussion/116993/zotero-7-ui-style-can-i-resolve-my-idiosyncratic-issues-with-it-in-css
https://forums.zotero.org/discussion/117312/the-white-and-gray-background-colors-on-the-items-make-the-whole-interface-look-messy

Windows people always [wanted the macOS behavior](https://forums.zotero.org/discussion/98500/feature-request-alternating-background-color-for-library-entries-in-windows), and never the reverse, but a few people have asked to turn this off, so we might as well make it a pref.

tnajdek:
The preference name is `extensions.zotero.ui.noAlternatingRows` and can be toggled using a checkbox in the General preferences labeled "**Disable alternating row colors**".

It’s a bit unusual to have a "negative" preference (i.e., check to disable), but in this case, it makes sense, as the default is much nicer and likely preferred by the vast majority of users.

Resolves #4628.

### DIFF
--- a/chrome/content/zotero/components/virtualized-table.jsx
+++ b/chrome/content/zotero/components/virtualized-table.jsx
@@ -1258,7 +1258,8 @@ class VirtualizedTable extends React.Component {
 			onMouseLeave: e => this._handleMouseLeave(e),
 			className: cx(["virtualized-table", {
 				resizing: this.state.resizing,
-				'multi-select': this.props.multiSelect
+				'multi-select': this.props.multiSelect,
+				'no-alternating-rows': Zotero.Prefs.get('ui.noAlternatingRows'),
 			}]),
 			id: this.props.id,
 			ref: ref => this._topDiv = ref,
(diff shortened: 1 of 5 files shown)


dstillman:
1. We should add a pref observer to invalidate the tree immediately, so you can see this take effect behind the settings window.
2. Let's hide and ignore this on macOS. We've never had it there before, so no reason to start now. We can even add this as a default pref only on Windows/Linux [in build.sh](https://github.com/zotero/zotero/blob/e0aaa7e36655fdedac6e03777759bc57b56e51f9/app/build.sh#L290).

dstillman:
Oh, we have a pref observer for the item tree that's being moved in https://github.com/zotero/zotero/pull/4601, so should come after that.

dstillman:
https://forums.zotero.org/discussion/119532/disable-striped-view

PR #4638, opened 2024-08-27
Attachment box: Wrap Filename field

### LINKED ISSUE #4637: Allow attachment box Filename field to wrap, or ellipsize in the middle?
Finder does the latter. The current Filename field is definitely not that useful, since auto-renamed filenames are almost always going to be cut off.

https://forums.zotero.org/discussion/comment/472148/#Comment_472148

AbeJellinek:
And replace use of deprecated method.

Fixes #4637 - ellipsizing in the middle is trickier and I think this looks good:

image


dstillman:
Merged

PR #4645, opened 2024-08-29
build_and_run: Support -q option for quick build

AbeJellinek:


dstillman:
build_and_run already uses `-q`. (That was what it was for.)

https://github.com/zotero/zotero/blob/051d3384c52b5b1bea05d1bc98b5c06789f5f7d8/app/scripts/build_and_run#L56

PR #4649, opened 2024-08-29
Use extension-less filenames when adding multiple attachments at once, remove URL dropping / `temporaryPDFHack`

### LINKED ISSUE #1425: Dragging a URL from Chrome isn't working
This works from Firefox and Safari, and I'm pretty sure this used to work from Chrome. We should look at the drop data to see what's coming in.

### LINKED ISSUE #4648: Don't set automatic attachment title when adding multiple attachments at once?
I don't know if we discussed this, but the current behavior (one file gets "PDF", the rest get their extension-less filenames) is weird. Use extension-less filenames for all files when adding multiple at a time?

https://forums.zotero.org/discussion/117429/bug-adding-multiple-files-at-once-via-add-attachment-file-results-in-one-the-files-misnamed

AbeJellinek:
- Use extension-less filename as title when adding multiple attachments of the same type at once
	- By delaying the `setAutoAttachmentTitle()` call until all attachments have been added
	- Use a new notifier queue in `addAttachmentFromDialog()` (like we already did in the item tree) to prevent notifier events from firing in the middle.
- Don't load and translate URLs dropped from Firefox
	- This made sense when Zotero was a Firefox extension, but now it works badly on many sites (no cookies), only supports one browser (#1425), and is inferior to using the Connector in basically all cases. I'm not sure if anyone still uses it. If there's demand, we should reimplement it via the Connector.
	- `getDataFromDataTransfer()` now converts `file` URIs into files and discards everything else
	- Remove `temporaryPDFHack`, which would have celebrated its 15th birthday in January
- Add tests for both

I did these together because of conflicts, but the two commits can be separated if needed.

Closes #4648
Closes #1425 (won't fix?)

### DIFF
--- a/chrome/content/zotero/collectionTree.jsx
+++ b/chrome/content/zotero/collectionTree.jsx
@@ -1749,19 +1749,17 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 				}
 				return true;
 			}
-			else if (dataType == 'text/x-moz-url' || dataType == 'application/x-moz-file') {
+			else if (dataType == 'application/x-moz-file') {
 				if (treeRow.isSearch() || treeRow.isPublications()) {
 					return false;
 				}
-				if (dataType == 'application/x-moz-file') {
-					// Don't allow folder drag
-					if (data[0].isDirectory()) {
-						return false;
-					}
-					// Don't allow drop if no permissions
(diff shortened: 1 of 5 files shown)


dstillman:
https://forums.zotero.org/discussion/comment/499977/#Comment_499977

Issue #4644, opened 2024-08-29
Word for Mac integration installer runs on new profile, without Word installed


AbeJellinek:
I may have had Word installed at some point, but not now.

1. Start Zotero with `-P`
2. Create a new profile
3. This shows:

![image](https://github.com/user-attachments/assets/a12ea641-5f4a-4b93-a300-687275e07012)

PR #4660, opened 2024-09-03
Show a banner before installing the Word plugin on macOS Sequoia+

### LINKED ISSUE #4526: Show banner before trying to install Mac Word plugin on Sequoia
Sequoia locks down writes to Group Containers, so we need to show a dialog before actually doing the write, warning that there'll be a system prompt (with Continue and Cancel). We need this mostly for automatic installs, but we might as well show for manual installs too, just since the dialog is vague and kind of scary.

We can limit this to Sequoia and up (`Zotero.isMac && (await Zotero.getOSVersion()).split(' ')[1].split('.')[0] >= 15`).

I can write the text.

Screenshot 2024-08-08 at 2 03 06 AM

adomasven:
Shows a warning banner on macOS Sequoia+ before running the word template installer. Will display on initial Zotero install (if Word is installed) or if we need to install a template update. No pre-prompt when installing via Preferences -> Cite.

Look:
Screenshot 2024-09-03 at 10 28 14

Compared to other banners:
Screenshot 2024-09-03 at 10 28 56

@dstillman will want to adjust wording.

To test you will need to reset `extensions.zoteroMacWordIntegration` prefs. There's also a [branch for testing on Sonoma](https://github.com/adomasven/zotero-word-for-mac-integration/tree/test-installation-warning-banner) which installs to `~/Library/Containers/com.microsoft.Word/Data`. As far as I know the OS warning dialog on Sequoia is different. I don't get an option in Sonoma to approve access for a single session, and get the OS prompt multiple times.

Closes #4526


adomasven:
Also closes #4644

adomasven:
Updated to include a "Don't ask again option". It's adding a duplicate string into the zotero.ftl file, but otherwise we would have to handle that single string differently from the rest in the banner.

Screenshot 2024-09-13 at 16 24 18

As a reminder, you can test the behaviour with Sonoma using the [branch for testing on Sonoma](https://github.com/adomasven/zotero-word-for-mac-integration/tree/test-installation-warning-banner) which installs to ~/Library/Containers/com.microsoft.Word/Data. I am getting 4 separate dialogs this way (since the installer performs 4 operations in that location), so that's not great. @dstillman you said that Sequoia displays a prompt that allows the user to give permission for the entire session, so we need to make sure that actually occurs.

Useful testing commands:
```js
// Display the prompt
ZoteroPane.showMacWordPluginInstallWarning()

// Reset the prefs
Zotero.Prefs.resetBranch([], "extensions.zoteroMacWordIntegration.")
// Run the command that runs on startup which checks prefs (e.g. for 7 day delay if asked to remind later)
// and shows or doesnt show the banner based on previous interactions
var macWordInstaller = ChromeUtils.import("resource://zotero-macword-integration/installer.jsm"); var i = new macWordInstaller.Installer();

@yexingsha Would be great if you could make this look less alarming. Yellow is too intense here.

yexingsha:
If we don't want to treat this banner as a warning, then I think blue is the best choice. Some other thoughts:

1. With both "Remind Me Later" and "Don't Ask Again" options, the close button is redundant and too vague.
2. Let's style the three buttons in this banner to draw attention to the primary call-to-action button.
3. I don't think we need to bold the text here, the contrast is fine and the colored banner is sufficiently attention grabbing.

This will make the banner more in line with the first-run banner:

word-plugin-banner

dstillman:
Merged

PR #4664, opened 2024-09-03
Load DTD strings lazily

### LINKED ISSUE #4661: Startup error: "module record has unexpected status: Evaluating"
https://forums.zotero.org/discussion/114867/open-error-module-record-has-unexpected-status-evaluating
https://forums.zotero.org/discussion/115121/
https://forums.zotero.org/discussion/117304/software-update-screen-froze
https://forums.zotero.org/discussion/117551/zotero-software-crashing-after-installed

Race condition triggered by synchronous XHR:

https://github.com/zotero/zotero/blob/1f8b480e7007e06928950615c9b4deb84aacdc10/chrome/content/zotero/xpcom/intl.js#L85

Supposedly fixed in Firefox 106, but apparently not for us:

https://bugzilla.mozilla.org/show_bug.cgi?id=1788532

AbeJellinek:
And resolve XML entities in the loaded strings. Doesn't handle references to other DTD strings (e.g. `aboutProduct.label` = `About &brandShortName;`) but might fix display bugs in locales that use lots of `&quot;` in their strings.

Should fix #4661 by delaying the synchronous XHR until the first `getString()` call (which will be well after module loading is completed, I hope).

### DIFF
--- a/chrome/content/zotero/xpcom/intl.js
+++ b/chrome/content/zotero/xpcom/intl.js
@@ -94,12 +94,19 @@ Zotero.Intl = new function () {
 		const intlFiles = ['zotero.dtd', 'preferences.dtd', 'mozilla/editMenuOverlay.dtd'];
 
 		let strings = [];
+		let { documentElement: elem } = new DOMParser().parseFromString('<root></root>', 'application/xml');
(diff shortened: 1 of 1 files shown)


dstillman:
Merged

Issue #4669, opened 2024-09-05
Automatically switch to Stacked mode when resizing window in reader tab


dstillman:
We're getting lots of posts complaining about the minimum width of reader tabs. [One suggestion](https://forums.zotero.org/discussion/comment/473243/#Comment_473243):

> Would it perhaps be an intuitive solution for the library tab to automatically collapse the item pane when the window is shrunk beyond the usable minimum and to expand it again automatically if the window is made big enough? 

My first thought was that this would be very confusing, but it might actually make sense? If you did switch back to the library tab while the window was too narrow, the item pane would be gone, but the sidenav is right there, and hopefully most people would figure out they could click one of the buttons. And we already expands the window to fit the item pane if you click a sidenav button while the pane is closed and the window was made too narrow.

I'm not sure whether we'd do this in the library tab too. It's possible that would cause more confusion, but it's also sort of clearer than the item pane only disappearing while you're in another pane. As long as it auto-expands as you enlarge the window, maybe it's fine, and better to be consistent?

adomasven:
Could also switch to stacked layout in this case.

dstillman:
Hmm. If there's ever a choice between thinking about stacked layout and not thinking about stacked layout, I prefer the latter, but I guess as long as we have it, it's hard to argue that that's not the friendlier, less-alarming option. And we also already automatically expand the window as necessary when moving from stacked to standard. So maybe that makes the most sense. We'd still want to automatically switch out of it on a resize-out if we switched into it due to a resize-in in the same session. (If people panic and restart Zotero to try to fix the problem…I guess they need to find the menu option.)

windingwind:
I was thinking that if the window is already in its min-width with the item pane open, you just cannot resize the window anymore, and dragging the window border does not trigger a resize event. How can we know people are resizing the window in that case and take actions to collapse the item pane or switch to stack mode or something?

dstillman:
Can we avoid setting an actual min-width when the item pane is showing, and just adjust the UI?

windingwind:
Could we try to set a max-width for the item pane based on the window width, so that it won't get ridiculously large and thus won't feel like that it prevents the window from being resized?

PR #4676, opened 2024-09-09
Fix main window layout and window resizing

windingwind:
fix: #4669
Replace flex attribute with flex-shrink and flex-grow style
to allow elements to be resized by splitter and window resizing.
Set the max-width on collections pane
to prevent it from pushing other panes outside window border
Set the min-width on window according to
the size of both library and reader tabs

### DIFF
--- a/chrome/content/zotero/contextPane.js
+++ b/chrome/content/zotero/contextPane.js
@@ -48,6 +48,17 @@ var ZoteroContextPane = new function () {
 			: _contextPaneSplitter)
 	});
 
+	Object.defineProperty(this, 'collapsed', {
+		get: () => {
+			return this.splitter.getAttribute('state') === 'collapsed';
+		},
+		set: (collapsed) => {
+			_contextPane.setAttribute('collapsed', collapsed);
+			this.splitter.setAttribute('state', collapsed ? 'collapsed' : 'open');
+			_update();
+		}
+	});
+
 	this.update = _update;
 
 	this.focus = () => {
@@ -176,6 +187,8 @@ var ZoteroContextPane = new function () {
 		
 		_updatePaneWidth();
 		_updateAddToNote();
+
+		ZoteroPane.updateLayoutConstraints();
 	}
 	
 	function _isLibraryReadOnly(libraryID) {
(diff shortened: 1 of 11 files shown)


dstillman:
This doesn't seem to be doing anything for me. I drag the right edge of the window to the left and it just stops in the normal place, without switching into Stacked.

I'm also getting logged warnings in the console, which may or may not be from this:

```
Error in parsing value for ‘width’.  Declaration dropped.
```

windingwind:
No it's not. From my understanding, we want to collapse the item pane in the library to stack layout because it cannot be shrunk by dragging the window to a smaller size, which makes the window too wide if the item pane is very wide (e.g. when you leave the maximized window - the item pane is never resized unless you manually drag the splitter).

Now that this PR allows all layout components to be resized to their min-width while resizing the window, the window itself will never be ridiculously wide, and thus I think it resolves the original issue and there is no need to introduce the auto-stack.

windingwind:
We can set a smaller min-width to the left/right pane to make the window fit the half-screen width on a regular laptop screen, e.g. 1080/2 px. I guess we discussed before, about setting the min-width of the collections pane to 100px or so.

But if that's not enough, we can do the auto-stack thing. I didn't add it here, as the changes in this PR seem to be more fundamental. It would be easy to add the auto-stack based on this PR, and I guess it would be better to be a separate PR/commit.

dstillman:
This is great. One glitch: in a reader tab, if the context pane is closed and I make the window narrow, the context pane reappears on the bottom.

dstillman:
But yes, we can make the collections pane min-width smaller. I don't think we can really make the item pane smaller, due to the creator row buttons.

dstillman:
(The width of the field labels also varies by locale, and I think some are wider than English.)

windingwind:
How about decreasing the min-width of the items tree from 370px to 200px or 250px?

As for the item pane, we could use some icons for field labels when the item pane is narrow so that it can free 100 or 150 more px.

dstillman:
Let's start with just decreasing the collections min-width and the auto-stack.

dstillman:
(We're definitely not going to start swapping out field labels for icons, and the auto-stack makes reducing the items list well past the point of usability unnecessary.)

windingwind:
OK, the collections pane is now 100px min-width.

I noticed that actually many JS layout tricks we used before aren't necessary. Replacing the `flex=1` attribute with `flex-grow` and `flex-shrink` along with `min-width: 0` can perfectly layout the elements, e.g. collection search textbox.

dstillman:
Hmm, this is actually much weirder. As I narrow the window, the collections pane narrows to be totally useless — just the icon plus one character — and then snaps back to a much larger size when stacked mode turns on.

If we're using auto-stack, I don't think super narrow, useless min-widths make sense. Narrow window widths are a genuinely good reason to use stacked mode, so let's just embrace it and keep the min-width we had before.

PR #4680, opened 2024-09-10
VPAT: minor settings tweaks 

abaevbog:
A number of small tweaks after the last review.

- it was marked that `+/-` buttons below the cite manager table should have "Add" and "Remove" aria labels because "+" and "-" are not descriptive enough. There is a tooltip that is also announced so I don't think there is much room for confusion here. But `+/-` below export settings do have "Add" and "Remove" so it's worth adding them even just for consistency.
- the "Color scheme" radiogroup had its radio buttons linked to the label via `aria-describedby`. It was indicated that it's not a good thing because semantically it's not quite correct and  it also requires a setting to be checked in JAWS to be heard. Wrapping it into a groupbox will have screen readers announce the "Color scheme" when focus lands on a radiobuton. It feels more appropriate semantically though functionally doesn't change much.
- linked a few missed itemPane header preference labels to their menus

### DIFF
--- a/chrome/content/zotero/preferences/preferences_general.xhtml
+++ b/chrome/content/zotero/preferences/preferences_general.xhtml
@@ -27,15 +27,16 @@
 		<groupbox aria-labelledby="preferences-appearance-title">
 			<label><html:h2 id="preferences-appearance-title" data-l10n-id="preferences-appearance-title"/></label>
 		
-			<hbox align="center">
-				<label id="color-scheme-label" data-l10n-id="preferences-color-scheme"/>
-				<!-- aria-describedby set on individual radio buttons because voiceover won't announce it if set on radiogroup-->
-				<radiogroup orient="horizontal" id="color-scheme" preference="browser.theme.toolbar-theme">
-					<radio data-l10n-id="preferences-color-scheme-auto" value="2" aria-describedby="color-scheme-label"/>
-					<radio data-l10n-id="preferences-color-scheme-light" value="1" aria-describedby="color-scheme-label"/>
-					<radio data-l10n-id="preferences-color-scheme-dark" value="0" aria-describedby="color-scheme-label"/>
-				</radiogroup>
-			</hbox>
+			<groupbox aria-labelledby="color-scheme-label">
+				<hbox align="center">
+					<label id="color-scheme-label" data-l10n-id="preferences-color-scheme"/>
+					<radiogroup orient="horizontal" id="color-scheme" preference="browser.theme.toolbar-theme">
+						<radio data-l10n-id="preferences-color-scheme-auto" value="2"/>
+						<radio data-l10n-id="preferences-color-scheme-light" value="1"/>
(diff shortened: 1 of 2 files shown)


abaevbog:
One significant thing that was brought up during the last review is that, apparently, JAWS on windows 10 actually refuses to announce headers and descriptions within settings labels, though it works fine on windows 11. It's the first instance of different JAWS behavior on win10 vs win11, and my VM is win11, so I missed that. 

It's quite unfortunate because solutions to that are either hacky or a bit tricky. 

- My initial idea was to announce headers and other text via `role="alert"` component when focus enters the group only on windows 10 https://github.com/abaevbog/zotero/commit/edb131f14bc1918bdb6f84aab188e1033fe533e4. It does work decently well, except the VPAT folks mentioned that braille display behavior would need to be verified as well, and it looks like some of the alerts don't make it into the braille display and end up being skipped. I'm not exactly sure on the standards and expectations for braille viewer, so it may or may not be a good fix.
- A less nice but more reliable approach could be to make actual sections focusable. Then, braille display would show it as just a region with a label. It would be quite clunky though as the number of tabstops would increase.
- Finally, maybe given all these issues, it is worth to revisit the idea of wrapping panes into `<browser>` components https://github.com/zotero/zotero/pull/4414 which allows the virtual cursor of JAWS to turn on. This is the approach that Thunderbird and safari take already, and I feel like it would be a "proper" way to solve this. The main issue here is just taking care of all the breakages that happen when panes are not within the same `document`. I think I'll have another look at how bad it would be, and if it leads to a ton of issues, then I suppose we can try to have VPAT folks check the `role="alert"` tweak and see what they say about the braille display.

dstillman:
If it works in Win11 and not Win10, that just seems like a bug in JAWS that they should fix. Or I guess potentially a bug in Firefox where it doesn't work with older Windows versions. But it seems a little dumb to spend a huge amount of time fixing this and totally messing up how our settings are built for an OS that Microsoft will mostly stop supporting in 2025.

abaevbog:
That is a very fair point. Perhaps then the best course of action is to ask QL folks to run the testing on win11 (which was actually brought up) and see how that goes? If there are no additional issues (specifically, I am not exactly sure what the expectation for the braille display that was mentioned during last check), then maybe it's fine. If not, we can try that alert tweak that does communicate messages ok-ish as a temp addition until windows10 is not supported or JAWS somehow fixes this.

abaevbog:
There are so many different glitches that I ran into between firefox and screen readers (not exclusively JAWS to be fair) that I don't even know whom to blame and when. The fact that we have chrome windows without a `<browser>` or `<iframe>` to wrap elements in a web context generally complicates things because it's a more niche setup. So yeah, it could be either JAWS or Firefox (or it could be something else altogether ...) but I agree that putting a lot of effort into hacking around a glitch for OS that may be no longer supported in a bit doesn't feel right, so we should avoid it if we can.

PR #4682, opened 2024-09-12
Fix prefs window style on MacOS

windingwind:
Use MacOS menupopup style to prefs window
Move style mac/menupopup into mac/components


windingwind:
Fix regression after https://github.com/zotero/zotero/commit/3d0014f258058797c92af689fd5b9e2a0e2c9773

Current: 
image

After:
image

PR #4694, opened 2024-09-18
add server connector endpoint to undo save session

abaevbog:
- Added `/connector/cancel` endpoint to connector server that the connector can call to cancel (or rather undo) the saving process if it was started by mistake.
- If there is no session (meaning, saving did not start yet), it will do nothing as the client will just need to not to hit the `/connector/save*` endpoint.
- If there is a session and all saving is done, added items will be erased right away.
- If there is a session and saving is not done, send all added top-level items so far to the trash so they immediately go away from the current view. Then, start a loop to check periodically if the saving is done, and erase the items as soon as it is. Letting all the saving process finish and handling this in one place seems like the most reliable way to achieve this without the risk of all sorts of possible glitches. While experimenting with other approaches, I ended up corrupting my database a few times, so letting everything settle first in the trash seemed like not a bad idea as it's least likely to go wrong ... Though, correct me if there is a more elegant solution that I missed!

A few additional fixes:
- fixed session not getting removed or garbage-collected
- fixed `isSavingDone` not returning true is any of the attachments failed to get saved

Addresses part of: https://github.com/zotero/zotero-connectors/issues/301

### DIFF
--- a/test/tests/server_connectorTest.js
+++ b/test/tests/server_connectorTest.js
@@ -1928,4 +1928,153 @@ describe("Connector Server", function () {
 			assert.equal(JSON.parse(req.response).status, 200);
 		});
 	});
+
+	describe('/connector/cancelSession', function () {
+		var pdfArrayBuffer;
+		var pdfPath = OS.Path.join(getTestDataDirectory().path, 'test.pdf');
+		before(async () => {
+			await selectLibrary(win, Zotero.Libraries.userLibraryID);
+			pdfArrayBuffer = (await OS.File.read(pdfPath)).buffer;
+		});
+
+		it('should delete saved items if called after everything is saved', async function () {
+			var collection = await createDataObject('collection');
+			let sessionID = Zotero.Utilities.randomString();
+			let bookItemID = Zotero.Utilities.randomString();
+			await select(win, collection);
+
+			let body = {
+				sessionID,
+				items: [
+					{
+						id: bookItemID,
+						itemType: "book",
+						title: "Book Title",
+					}
+				]
+			};
+			let itemAddPromise = waitForItemEvent('add');
+			// Save the item as if by the connector
+			await Zotero.HTTP.request(
+				'POST',
+				connectorServerPath + "/connector/saveItems",
+				{
+					headers: {
+						"Content-Type": "application/json"
+					},
+					body: JSON.stringify(body)
+				}
+			);
+			let itemIDs = await itemAddPromise;
+
+			// Add the attachment to the item
+			let attachmentAddPromise = waitForItemEvent('add');
+			let attachmentReq = await httpRequest(
+				'POST',
+				connectorServerPath + "/connector/saveAttachment",
+				{
+					headers: {
+						"Content-Type": "application/pdf",
+						"X-Metadata": JSON.stringify({
+							sessionID,
+							title: "Book Attachment",
+							parentItemID: bookItemID,
+							url: `${testServerPath}/attachment1.pdf`,
+						})
+					},
+					body: pdfArrayBuffer
+				}
+			);
+			assert.equal(attachmentReq.status, 201);
+			let attachmentIds = await attachmentAddPromise;
+			assert.lengthOf(attachmentIds, 1);
+			let attachment1 = Zotero.Items.get(attachmentIds[0]);
+			let bookItem = Zotero.Items.get(itemIDs[0]);
+			assert.equal(attachment1.parentItemID, bookItem.id);
+
+			// Cancel the session
+			let cancelReq = await Zotero.HTTP.request(
+				'POST',
+				connectorServerPath + '/connector/cancelSession',
+				{
+					headers: { 'content-type': 'application/json' },
+					body: JSON.stringify({ sessionID })
+				}
+			);
+			assert.equal(cancelReq.status, 200);
+
+			// The item and attachment should be deleted
+			let deletedItem = await Zotero.Items.getAsync(attachment1.parentItemID);
+			let deletedAttachment = await Zotero.Items.getAsync(attachment1.id);
+			assert.isFalse(deletedItem);
+			assert.isFalse(deletedAttachment);
+		});
+
+		it('should not save attachment items if session is cancelled half-way through', async function () {
+			var collection = await createDataObject('collection');
+			let sessionID = Zotero.Utilities.randomString();
+			let bookItemID = Zotero.Utilities.randomString();
+			await select(win, collection);
+
+			let body = {
+				sessionID,
+				items: [
+					{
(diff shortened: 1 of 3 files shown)


chrome/content/zotero/xpcom/connector/server_connector.js
```diff
@@ -1947,3 +1971,50 @@ Zotero.Server.Connector.Request.prototype = {
...
+			if (session.isSavingDone()) {
+				session.eraseItems();
+				clearInterval(intervalID);
+			}
+			// Ensure that even if something goes wrong, the loop breaks after 3 minutes
+			checkCount += 1;
+			if (checkCount > maxMinutes * 60 * 1000 / intervalDelay) {
+				clearInterval(intervalID);
+			}
+		}, intervalDelay);
```
adomasven, line 2016: Move this into the session and make it return a promise, then in the test spy the function and await the returned promise so that we don't have to add more arbitrary delays than necessary.

abaevbog: I moved it all into `session.eraseItemsWhenDone`, which the second test can `await` for. It's definitely cleaner that way!

I'm not sure if we actually need two different tests to check for item deletion now. They are handled differently but ultimately, the call to `/connector/cancel` when the save process is finished is like a subset of what is being done if `/connector/cancel` is called when the save process is ongoing. So maybe I should drop the `should delete saved items if called after everything is saved` test?

adomasven: > So maybe I should drop the `should delete saved items if called after everything is saved` test?

No harm in having it. Let's keep it in case something changes in the future.

test/tests/server_connectorTest.js
```diff
@@ -2945,42 +2945,11 @@ describe("Connector Server", function () {
...
-					}
-				);
-				let data = JSON.parse(checkProgress.response);
-				isDone = data.done;
-			}
-			while (!isDone);
+			// When /connector/cancel is called and save session is not done,
+			// it will call eraseItemsWhenDone to wait for items to load before deleting them
+			let session = Zotero.Server.Connector.SessionManager.get(sessionID);
+			await session.eraseItemsWhenDone();
```
adomasven, line 2951: You should still request `/connector/cancel` here, but do something like

```js
let eraseItemsWhenDone = sinon.spy(session, 'eraseItemsWhenDone')
// ... do stuff
await eraseItemsWhenDoen.firstCall.returnValue;
```

abaevbog: Oh, I actually didn't know about `sinon.spy` and didn't realize that's what you meant. Apologies! I'll keep it in mind for the future.

But it's super handy! I brought back the `/connector/cancel` call followed by `await eraseItemsWhenDoneSpy.returnValues[0]`. In addition, I thought it's worth to also spy for `session.trashItems` so we can `await` for when we move the item to the trash to make sure it does go away from the current view.

abaevbog:
At some point during testing this (either while working on tests or saving and cancelling a bunch of items while trying out different design approaches), I ended up corrupting my db again. The corrupted db notification is followed by `database disk image is malformed [QUERY: PRAGMA foreign_key_check] ` error. So there must be still something off somewhere. The foreign key error makes me think that it's something to do with attachments and their parent items.

Whatever the issue is, it needs to be sorted as it's the 3rd time I get a corrupted db and that's not good. I did hope that allowing everything to save before erasing items would prevent it but perhaps the problem was elsewhere. I don't suppose there is any better way to troubleshoot this than to try to somewhat reliably reproduce first?

abaevbog:
The output of `PRAGMA integrity_check;` is many lines as below:
```
On tree page 2303 cell 265: Rowid 2825495 out of order
On tree page 36 cell 219: 2nd reference to page 2348
row 470 missing from index sqlite_autoindex_syncCache_1
NULL value in syncCache.syncObjectTypeID
NULL value in syncCache.version
```

So I wonder if that was just caused by repeatedly saving a bunch of items after which the sync process kicks in, which would sometimes be interrupted if I kill the main process?

adomasven:
Would be simple enough to test this by disabling sync if you can reproduce this at least semi-consistently, but we also need to make sure closing Zotero normally doesn't also cause issues. Although I guess we would have users constantly corrupting the DB if closing Zotero manually left DB transactions unfinished @dstillman ?

abaevbog:
> Although I guess we would have users constantly corrupting the DB if closing Zotero manually left DB transactions unfinished

The only reason why I assumed this might have been the case (besides the `PRAGMA integrity_check;` output) is because if you don't close Zotero window normally (by clicking the red X button in the titlebar) but instead kill the process (e.g. run `./app/scripts/build_and_run -rd` from the terminal and then stop it with `Ctrl-C`), wrap-up operations, such as `ZoteroPane.destroy`, don't get to run. And while testing thing back and forth, I tend to do the latter. So I don't think that anyone who doesn't build Zotero themselves would have encountered this (assuming this is the reason). But I am also not exactly sure about the details of the sync engine and how things are handled.

In any case, I still haven't been able to reproduce it again... So i'll keep trying!

abaevbog:
Actually, maybe this is related: https://forums.zotero.org/discussion/118115/zotero-7-zotero-database-zotero-sqlite-appears-to-have-become-corrupted

abaevbog:
Please disregard the corrupted DB comments... I am pretty sure the reason was that I missed that an instance of Zotero was running in my windows VM, while this on mac with another instance. Both instances of Zotero talked to the same directory, which is more than likely to mess up the database.

adomasven:
> Please disregard the corrupted DB comments... I am pretty sure the reason was that I missed that an instance of Zotero was running in my windows VM, while this on mac with another instance. Both instances of Zotero talked to the same directory, which is more than likely to mess up the database.

In that case can this be simplified to not need to trash items first?

abaevbog:
> In that case can this be simplified to not need to trash items first?

The reason why I added the trashing step is to provide a somewhat immediate UI feedback to the cancelation. When you press the "cancel" button, we still wait for everything to finish saving before deleting it so there can be quite a long time before something visually changes for the user. I thought it might look like the cancelation didn't work and something broke because items would just stay as they were (and new items might even get added). In the example below without trashing, it takes about 4-5 seconds after the popup disappears for the item to go away and I think it feels off? If one is saving many items, that gap could get even longer.

https://github.com/user-attachments/assets/c46269c7-31a1-4bc0-a753-7b5549ad08d7


For comparison, this is the current behavior with trashing where the item goes away pretty much immediately. I think it's better? Obviously, one would still be able to find the items in the trash until they're erased but I think it does make it clearer that the process is working as expected:

https://github.com/user-attachments/assets/40f7083d-11bb-409f-b3ee-82bbbdba2212

Issue #4688, opened 2024-09-14
Inconsistent label in the item details pane when trash contains collections


tnajdek:
When all rows are selected the labels says "2 **objects** selected" but when nothing is selected it says "2 **items** in this view".

<img width="1276" alt="Screenshot 2024-09-14 at 17 02 20" src="https://github.com/user-attachments/assets/4fe8531c-962c-4979-9dfb-cdac4b3de40e">

<img width="1276" alt="Screenshot 2024-09-14 at 17 02 23" src="https://github.com/user-attachments/assets/fba62ed0-37a9-47a1-a174-b91f4ab99507">

PR #4696, opened 2024-09-18
Unify message in the item details pane when trash contains collections or searches

windingwind:
fix: #4688


windingwind:
It now shows `xx objects in this view` if any searches/collections exist.

I don't think it's necessary to distinguish if these rows in the trash are all collections or searches, we only need to keep them unified with what shows after some are selected (objects or items).

PR #4697, opened 2024-09-19
Fix reboot loop on limited accounts on Windows

tnajdek:
When a user without admin privileges rejects a UAC dialog (which might not even appear in heavily restricted environments), the Zotero installer falls back to installing Zotero in a local directory. However, this is the same folder where Zotero keeps [profile directories](https://www.zotero.org/support/kb/profile_directory).

In the installer, [we instruct NSIS to remove the >**empty**< old Zotero installation directory](https://github.com/zotero/zotero/blob/6e130a86495aa1f81326783fb5d32b5877c9fc5c/app/win/installer/installer.nsi#L239) (no `/r` flag). Since we're using `/REBOOTOK`, NSIS uses the `MoveFileEx` WinAPI call with the `MOVEFILE_DELAY_UNTIL_REBOOT` flag and internally triggers `RebootFlag` if removal is not possible. There are multiple issues here:

1. The directory is not empty (it contains the user's profile), so `RmDir` will always fail, thus triggering the `RebootFlag` for all installations executed in this environment.
2. `MOVEFILE_DELAY_UNTIL_REBOOT` can only be used if the process is running under a user who belongs to the administrators group or the LocalSystem account (as per the [MoveFileExA function](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-movefileexa)).

Because the reboot flag is triggered due to issue 1 above, the installer replaces `zotero.exe` with `helper.exe` and sets up a post-reboot instruction to [revert this change](https://github.com/zotero/zotero/blob/6e130a86495aa1f81326783fb5d32b5877c9fc5c/app/win/installer/installer.nsi#L532). However, due to issue 2, this reversion can never occur, rendering Zotero unusable because the executable is actually `helper.exe`, which detects the `.moz-upgrade` file, assuming the user has not yet rebooted.

Here it is practice:

https://github.com/user-attachments/assets/df18f8a0-bad1-4eb2-80ba-96c409e212a7

Removing `REBOOTOK` from the instruction that removes the installation directory should fix most common cases. This is what this PR does.

https://github.com/user-attachments/assets/4ea6b473-8733-4e44-9bdf-12fa1070852e

However, this issue can manifests in few different places (e.g. in the same environment when uninstalling while Zotero is running). To fix it properly, we should check user type in both installer and uninstaller and avoid using `REBOOTOK` in latter.

In this PR I've also removed previous attempt to fix this issue (remove entire installation directory) as it might have other unexpected consequences.


dstillman:
Merged

dstillman:
(Technically that "Zotero" folder doesn't contain profile directories — it contains the cache folders associated with each profile directory.)

Issue #4692, opened 2024-09-17
Add Title to attachment box


dstillman:
https://forums.zotero.org/discussion/118002/rename-attachment-file-new-file-name-not-shown-error-id-864572657

Might help avoid some of this confusion

PR #4700, opened 2024-09-21
Update attachment box fields

windingwind:
Add Title to attachment box
Add Rename button to attachment box
resolve: #4692


windingwind:
image

windingwind:
https://github.com/user-attachments/assets/3ea341c1-b7b1-489d-a2a4-673f61f3dc28

dstillman:
Let's separate out the Title field and get that merged first.

windingwind:
The PR for title field is #4702

windingwind:
Rebased.

I think a rename filename button is safe to go.
Adding a tooltip for title & filename field input is arguable, but it could filter out most users before they complain in the forums. We can decide if we want it.

dstillman:
We no longer need this after the button added in https://github.com/zotero/zotero/pull/3860, right?

windingwind:
> We no longer need this after the button added in #3860, right?

Yes I think so. I'll close this PR.

Issue #4707, opened 2024-09-24
Manage styles label in bibliography.xhtml not in tab order


abaevbog:
https://forums.zotero.org/discussion/118215/zotero-7-05-document-pref-dialog-cant-access-manage-styles-link-with-keyboard

PR #4709, opened 2024-09-24
Make manage styles link in style configurator keyboard accessible

abaevbog:
Fixes: #4707


dstillman:
This shows a much larger focus ring than it should:

Screenshot 2024-09-26 at 7 58 59 PM

PR #4712, opened 2024-09-27
Trim openURL.resolver pref before opening the link

### LINKED ISSUE #4711: Trim whitespace from OpenURL Base URL
https://forums.zotero.org/discussion/comment/475262/#Comment_475262

abaevbog:
If a custom resolver url has trailing whitespaces, Library Lookup will still work.

Fixes: #4711

### DIFF
--- a/chrome/content/zotero/preferences/preferences_general.js
+++ b/chrome/content/zotero/preferences/preferences_general.js
@@ -345,7 +345,7 @@ Zotero_Preferences.General = {
 		}
 		var openURLMenu = document.getElementById('openurl-menu');
 		let openURLMenuFirstItem = openURLMenu.menupopup.firstChild;
-		if (!this._openURLResolvers) {
+		if (!this._openURLResolvers?.length) {
 			openURLMenuFirstItem.setAttribute('label', Zotero.getString('general.loading'));
 			try {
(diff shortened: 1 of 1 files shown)


dstillman:
I think it'd be better to trim when writing the pref, though, so that it matches the directory. In the case of the report, it would've still said "Custom" even though they were pasting from the directory page.

abaevbog:
Good point! I'll add it. 

I think we might want to trim before opening URL though as well, for a rare situation if someone else already has a saved resolver url with a whitespace.

abaevbog:
How about this approach? I added the trimming of the openURL value `oninput` in prefs, as well a matching of the input against resolvers. A small overhead is that resolvers need to be loaded dynamically, so we can start loading them as soon as the user types, and once resolvers are loaded, we check to see if the user typed something that matches any of them.

For reference, this is what it would look like, with me pasting "    https://search.library.dartmouth.edu/openurl/01DCL_INST/01DCL_INST:01DCL?":

https://github.com/user-attachments/assets/1c68cde3-43ab-4322-bf66-c9164776ea1c

abaevbog:
Also, I ran into George Mason University resolver's url not working. `http://sfx.wrlc.org/gm` doesn't resolve to anything in my browser. Not sure if the link is right but it's just temporarily down or if the url changed, so I thought to mention it.

Issue #4722, opened 2024-10-02
Collections highlight with Ctrl is laggy


abaevbog:
A few reports about collection highlighting on Ctrl keydown not properly working on windows and linux:

https://forums.zotero.org/discussion/118422/highlighting-of-item-in-multiple-collection-not-working

https://forums.zotero.org/discussion/118423/lagging-after-pressing-ctrl-to-highlight-the-position-in-zotero-7-0-5

https://forums.zotero.org/discussion/118391/highlighted-collections-by-pressing-ctrl-with-item-selected-changed-behaviour-same-problem

Mac seems to work as expected. On windows, holding Control initially does nothing but collections will get highlighted if I scroll with a mouse anywhere on the screen, after which the collections do get highlighted. However, the collectionTree either can't be scrolled at all or can be but feels very laggy.

abaevbog:
The linked PR should take care of https://forums.zotero.org/discussion/118391/highlighted-collections-by-pressing-ctrl-with-item-selected-changed-behaviour-same-problem. But I still can't quite reproduce https://forums.zotero.org/discussion/118422/highlighting-of-item-in-multiple-collection-not-working.

dstillman:
Original thread for fixed behavior was https://forums.zotero.org/discussion/118007/highlighted-collections-by-pressing-ctrl-with-item-selected-changed-behaviour

abaevbog:
https://forums.zotero.org/discussion/126844/displaying-locations-of-notes

I am seeing that on windows holding Ctrl no longer highlights collections. For me `keydown` event does not fire until Ctrl is released OR something else happens (e.g. another key is pressed). I assume it's part of mozilla-specific handling of modifier keys - something similar happens with Alt keypresses.

One way I do get collection highlighting is by scrolling while holding Ctrl - once scrolling happens, `keydown` events are going through and the collections are highlighted as expected. Interestingly, scrolling doesn't even need to happen within Zotero. In the recording below, I hold Ctrl, nothing happens, I move the mouse from the Zotero window, scroll, and then highlighting appears.

https://github.com/user-attachments/assets/2892ece8-42ce-419b-b58d-291271def84a

abaevbog:
Holding Ctrl with any of the digits (e.g Ctrl+1) also seems to work

PR #4723, opened 2024-10-02
Allow to scroll collectionTree during highlighting via control

abaevbog:
Holding Ctrl and scrolling does not dispatch a scroll event as it is used for zooming in/out. To allow one to scroll through collectionTree while holding Ctrl, listen to `wheel` event and scroll manually based on that. I'm using `e.deltaY * 10` to determine how far scroll, and it works OK though a bit different from proper scrolling. On some platforms, it's just a bit faster and on others - just a bit slower. I assume there is a lot of variation here, so not sure if there is a way to get a perfect value by how much to scroll.

Addresses: #4722


abaevbog:
Oh yeah, thank you! That's a way better approach. Collection tree can scroll just fine now, and the reader is not affected as far as I can tell.

PR #4731, opened 2024-10-04
Implement macOS Sequoia Ctrl-Enter context menu shortcut

### LINKED ISSUE #4724: Ctrl-Return should open context menu on Sequoia
https://lifehacker.com/tech/right-click-with-your-keyboard-in-mac-os-sequoia

This is now a native macOS feature, but it's broken everywhere in Zotero. I assume we'll need to add custom support for this to various components (editable-text, the trees, etc.).

We can create separate issues for individual components if that helps to track this.

AbeJellinek:
This implements Sequoia's native behavior pretty exactly (I think!). The context menu is positioned:

- Below active list items, even if the list itself is focused (using `aria-activedescendant`)
- Below selected text in input fields
- To the right of the cursor in input fields if no text is selected
- Below all other focused controls

Closes #4724


dstillman:
Awesome! Merged.

PR #4734, opened 2024-10-07
Implement Ctrl-Enter handler in child frames

### LINKED ISSUE #4724: Ctrl-Return should open context menu on Sequoia

AbeJellinek:
Tested and working in Scaffold, note-editor, and basicViewer. EPUB and snapshot readers work as well. PDF needs https://github.com/zotero/reader/pull/141.

Fixes #4724

### DIFF
--- a/chrome/content/zotero/actors/ActorManager.jsm
+++ b/chrome/content/zotero/actors/ActorManager.jsm
@@ -63,3 +63,17 @@ ChromeUtils.registerWindowActor("ExternalLinkHandler", {
 	messageManagerGroups: ["feedAbstract", "basicViewer"]
 });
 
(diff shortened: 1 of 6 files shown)


dstillman:
Not a huge deal, but in notes, when no text is selected, I'm seeing the menu over on the right:

Screenshot 2024-10-17 at 1 11 02 AM

PR #4747, opened 2024-10-11
Fix select items window overflow

windingwind:


windingwind:
When the list is very long, it overflows the window.

Test with:

```js
arr = []

for (let i=0;i<100;i++) arr.push('1111111111111111111111')

window.openDialog("chrome://zotero/content/ingester/selectitems.xhtml",
			"_blank", "chrome,modal,centerscreen,resizable=yes", {
dataIn: arr});
```


Before:

image

After:

image

dstillman:
We have _selectItemsDialog.scss for this dialog.

windingwind:
That's for the `chrome://zotero/content/selectItemsDialog.xhtml` window. We have two *select items* dialog, which is indeed a bit confusing.

We can add a new SCSS file for this dialog, but it would have only one line. Maybe a small fix with an inline style would be neater here.

adomasven:
Why not something like this? https://github.com/zotero/zotero/blob/7c7d8c143341b3ad6b96030e902500d0608756d7/scss/components/_selectItemsDialog.scss#L77-L83

windingwind:
> contain-richlistbox

I didn't know that before, but it does not work for this dialog actually

adomasven:
> I didn't know that before, but it does not work for this dialog actually

Adding the class `.contain-richlistbox` to the window and the scss rule doesn't help?

windingwind:
> > I didn't know that before, but it does not work for this dialog actually
> 
> Adding the class `.contain-richlistbox` to the window and the scss rule doesn't help?

Yes, it doesn't work

adomasven:
Manually adding `min-height: 0` to all elements above the richlistbox works for me on linux to contain it.

windingwind:
Hmm, that does not work for me on MacOS

windingwind:
I don't know why but this patch still does not fix the issue for me.

The CSS rules are already applied but the list still overflows.

image

windingwind:
Thanks! It works for me.

As @AbeJellinek suggested, I moved and renamed the `zotero/ingester/selectitems` to `scaffold/select`. Let me know if it's improper.

PR #4750, opened 2024-10-13
Check for admin privleges in UninstallOld installer routine

tnajdek:
This PR adds a check for whether the user is an admin in the `UninstallOld` routine. If not, and the directory still exists at the end of the routine, it displays an error message and exits.

However, we don’t check if the user selected an empty directory for installation, so just re-running the installer will work. That makes me wonder if removing old Zotero should be a "best effort" approach. We could remove `/REBOOTOK` from that section, and if some files are left over, we could just ignore them and proceed with the installation.

This solution should address the most common issue causing the reboot loop, but we should also consider other cases, especially for users who have already run the previous version of the installer and are now stuck in the loop. I'll look into these other cases next.

https://github.com/user-attachments/assets/2ca460e6-a621-4e8e-92b8-e5186a393636

### DIFF
--- a/app/win/installer/installer.nsi
+++ b/app/win/installer/installer.nsi
@@ -236,7 +236,19 @@ Function UninstallOld
     Delete /REBOOTOK "$3\platform.ini"
     Delete /REBOOTOK "$3\precomplete"
     Delete /REBOOTOK "$3\voucher.bin"
-    RMDir $3
+    RMDir /REBOOTOK $3
+
+    ${If} ${FileExists} "$3"
+      # If the directory still exists, check if current user is Admin.
+      # For Admin users, /REBOOTOK works and directory will be removed after reboot
+      # For non-Admin users, we display a message and quit
+      UserInfo::GetAccountType
+      pop $0
+      StrCmp $0 "Admin" continue_installation
+      MessageBox mb_iconstop "Previous installation of Zotero could not be removed. Please manually remove the following directory, and then run the installer again: $\n$\n$3"
+      Quit
+    ${EndIf}
+
   continue_installation:
     ; End uninstallation
     SetShellVarContext current
--- a/app/win/installer/installer.nsi
+++ b/app/win/installer/installer.nsi
@@ -245,7 +245,7 @@ Function UninstallOld
       UserInfo::GetAccountType
       pop $0
       StrCmp $0 "Admin" continue_installation
-      MessageBox mb_iconstop "Previous installation of Zotero could not be removed. Please manually remove the following directory, and then run the installer again: $\n$\n$3"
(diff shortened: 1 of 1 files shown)


dstillman:
So this means that anyone with a non-admin install will get this message when reinstalling?

Really, even having to reboot is annoying (and pretty embarrassing for us).

And this is because of the Zotero\Zotero folder? It looks like that's deleted later, before installation:

https://github.com/zotero/zotero/blob/main/app/win/installer/installer.nsi#L319-L320

Can we just delete it earlier?

It's possible we can't delete it as part of uninstallation because uninstallation could be running as part of an in-app update, which maybe runs from that folder. (Not sure.) But I would think we could always delete it from the installer.

tnajdek:
> So this means that anyone with a non-admin install will get this message when reinstalling?
> 
> Really, even having to reboot is annoying (and pretty embarrassing for us).

Yes, if any file (e.g., the Zotero profile cache) persists after running the uninstaller and the [cleanup](https://github.com/zotero/zotero/blob/d7b5c93434cdf37028f7cf7a7e85a7eb54d10cf5/app/win/installer/installer.nsi#L220-L239), the user will see this message.

My suggestion is to remove all `/REBOOTOK` from the cleanup section and also remove the message. This means we will clean up the registry (by running the uninstaller) and do our best to clean up the folder, but if some files remain, that's OK. The installer can still handle installing into an existing Zotero directory. Here's a screencast showing just that on a limited account (I removed registry entry manually to force this behavior):

https://github.com/user-attachments/assets/22d73953-fb1b-4142-8c93-f821f580837f

tnajdek:
Another scenario we should consider is when a limited user runs the installer while Zotero is open. In this case, the `UninstallOld` routine calls the uninstaller of the previous installation (in silent mode), which creates the `moz-delete` file and triggers the reboot loop.

One solution is to check if Zotero is running **before** executing the old version's uninstaller, which should cover most cases. However, if the uninstaller fails for any reason, the user could still end up in a reboot loop.

dstillman:
> My suggestion is to remove all `/REBOOTOK` from the cleanup section and also remove the message. This means we will clean up the registry (by running the uninstaller) and do our best to clean up the folder, but if some files remain, that's OK.

That's probably fine. I don't think these `Delete`/`RmDir` lines are even doing much — I believe the uninstaller uses the `postcomplete` file (which is properly populated in current versions (including updates), but wasn't for a number of years in older versions) to delete all known files. I think that's why the folder is emptied (other than the `Zotero` folder) even though, e.g., a deletion for `.sig` files isn't listed here.

dstillman:
> One option could be to duplicate the executable and immediately recover from the duplicate (or clean it up) to perform the check.

Would be nice if there were a cleaner way to check if the file was locked, but I guess this works.

> Finally, what to do when Zotero is running and the installer is started in silent mode? Since we can't prompt the user to close Zotero, the best approach would be to just abort the installation?

Yes, that makes sense.

PR #4752, opened 2024-10-14
reader.ftl file for a11y strings in the reader

abaevbog:
Reader.ftl file with strings for aria-labels and a11y announcements.

Needed for https://github.com/zotero/reader/pull/144  per https://github.com/zotero/reader/pull/142#issuecomment-2410574199

### DIFF
--- a/chrome/locale/en-US/zotero/reader.ftl
+++ b/chrome/locale/en-US/zotero/reader.ftl
@@ -10,6 +10,8 @@ pdfReader-noteAnnotation = Note Annotation
 pdfReader-textAnnotation = Text Annotation
 pdfReader-imageAnnotation = Image Annotation
 
+pdfReader-find-in-document = Find in Document
+
 pdfReader-a11yMoveAnnotation = Use the arrow keys to move the annotation.
 pdfReader-a11yEditTextAnnotation = To move the end of the text annotation, hold { general-key-shift } and use the left/right arrow keys. To move the start of the annotation, hold { general-key-shift }-{ option-or-alt } and use the arrow keys.
 pdfReader-a11yResizeAnnotation = To resize the annotation, hold { general-key-shift } and use the arrow keys.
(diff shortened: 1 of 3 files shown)


chrome/locale/en-US/zotero/reader.ftl
```diff
@@ -0,0 +1,66 @@
...
+    [highlight] { pdfReader-highlightAnnotation } selected. { pdfReader-a11yEditTextAnnotation }
+    [underline] { pdfReader-underlineAnnotation } selected. { pdfReader-a11yEditTextAnnotation }
+    [note] { pdfReader-noteAnnotation } selected. { pdfReader-a11yMoveAnnotation }
+    [text] { pdfReader-textAnnotation } selected. { pdfReader-a11yMoveAnnotation } { pdfReader-a11yResizeAnnotation }
+    [image] { pdfReader-imageAnnotation } selected. { pdfReader-a11yMoveAnnotation } { pdfReader-a11yResizeAnnotation }
+   *[other] Annotation selected.
+} { $popupVisible -> 
+    [yes] { pdfReader-a11yAnnotationPopupAppeared }
+    *[other] { "" }
+}
```
dstillman, line 39: Since this shows up on Transifex, it's not ideal moving this much logic into Fluent. It might be necessary for a visual string where different languages have different conventions around sentence separators, but since this is only for screen readers anyway, that doesn't really matter — a period and a space should be sufficient. So if possible, probably better to handle most of this in JS and just have the "[ annotation ] selected." part here.

And now that we have placeholders where the order can be changed, we could probably try to just have a single string with `{ $annotation } selected.`, pass in the appropriate Fluent string (`pdfReader-textAnnotation`, etc.) as `$annotation`, and see if anyone complains that that doesn't work.

dstillman:
(I merged this so it can be tested, but I'll wait to push these to Transifex until we see if we can simplify some of the Fluent strings.)

dstillman:
(Also I didn't test this in a screen reader, so you should make sure I didn't break anything.)

abaevbog:
> Since this shows up on Transifex, it's not ideal moving this much logic into Fluent.

Gotcha, so the initial [approach](https://github.com/zotero/reader/commit/ffc5593584aa858f5a4de78197a75066a19de504#diff-71e848bde48eee29444b161f597336255b35fe4ba1bcc80593ae896b97541678R1220) of piecing together the sentences on the go was better and I will revert to that. I assumed it would be more readable and somewhat cleaner if it was a single line in the JS with a bit more logic in `.ftl` file.

> (Also I didn't test this in a screen reader, so you should make sure I didn't break anything.)

Will do!

abaevbog:
That makes sense! In these 2 PRs, the `reader.ftl` is stripped of pretty much all logic, and we piece together announcements by fetching the right annotation type, feeding it into `pdfReader-a11yAnnotationCreated` and `pdfReader-a11yAnnotationSelected` to get the main "_ selected." or "_ created." string, with additional instructions appended after that as needed.

There are also a few small tweaks to give reader access to the modifiers from `zotero.ftl` and a minor edit of the modifier strings for highlight/underline annotations on mac that I must have missed earlier.

PR #4756, opened 2024-10-15
simplify reader.ftl

abaevbog:
Remove logic constructing a11y announcements from reader.ftl. Instead, reader will fetch and put together sentences as needed on the fly.

Followup to https://github.com/zotero/zotero/commit/cde21ac9f2407856d000ea2038ea49cee7e8dd7c per https://github.com/zotero/zotero/pull/4752#discussion_r1800481917

Also, add `{ command-or-alt }` to zotero.ftl and use it for textual annotations instructions. On mac, the end of highlight and underline annotations is resized via Shift+Command+arrows, which is different from the Options modifier used in all other instances (per https://github.com/zotero/reader/pull/138#issuecomment-2373775068). I must have missed that the modifier is different earlier.

Finally, tweak aria-description of "Find in document" as it sounded as if Control+Option/Alt+1 would create either highlight or underline annotation (or both?). But correct me if I'm just mistaken about this part.

### DIFF
--- a/chrome/locale/en-US/zotero/reader.ftl
+++ b/chrome/locale/en-US/zotero/reader.ftl
@@ -13,30 +13,12 @@ pdfReader-imageAnnotation = Image Annotation
 pdfReader-find-in-document = Find in Document
 
 pdfReader-a11yMoveAnnotation = Use the arrow keys to move the annotation.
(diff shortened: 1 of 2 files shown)


dstillman:
I removed `command-or-alt` in favor of dedicated general strings for those two keys and a string in reader.ftl that uses those and won't show up in Transifex at all.

Issue #4755, opened 2024-10-15
Collections pane changes sizes for different items and is unresizable in Stacked mode


dstillman:
https://forums.zotero.org/discussion/118454/bug-in-7-0-7-the-my-library-pane-would-not-resize

Regression from 78077fab9

PR #4757, opened 2024-10-15
Fix item pane width jumping in stacked mode

windingwind:
fix: #4755

### DIFF
--- a/chrome/content/zotero/elements/itemPane.js
+++ b/chrome/content/zotero/elements/itemPane.js
@@ -547,10 +547,17 @@
 				let minHeight = 205;
 				let width = this.getAttribute("width");
 				let height = this.getAttribute("height");
-				if (!width || Number(width) < minWidth) this.setAttribute("width", String(minWidth));
+				let isStackedMode = Zotero.Prefs.get("layout") == "stacked";
+				// Don't set width if in stacked mode, it will be auto-resized
+				if (isStackedMode) {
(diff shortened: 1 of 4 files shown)


windingwind:
Should explicitly set a width for the container element of the item pane, otherwise its inner elements, e.g. title and abstract, can stretch the width of item pane.

dstillman:
This is now kind of more messed up for me:

Screenshot 2024-10-15 at 6 37 32 AM

windingwind:
Does it works for you now? The stacked layout does not rely on persisted width attributes and thus should be more robust.

dstillman:
Fixed, but now the splitter no longer tracks the mouse cursor as you resize it. Can we fix that?

https://github.com/user-attachments/assets/bdfc6939-e7dd-4eb6-acaf-25b2903e57a0

dstillman:
I'll merge this for now, since it's better than before.

windingwind:
> Fixed, but now the splitter no longer directly tracks the mouse cursor as you resize the panes. Can we fix that?
> 
>  splitter.mov

We need all layout components to be flex so that they can be shrunk by window resizing. For flex elements, the assigned `width`/`height` works as a relative ratio to their neighbor elements, instead of an absolute value, while the value of width/height is from the mouse. That's why the splitter does not track the mouse cursor. I guess that's the behavior of the builtin splitter system.

PR #4760, opened 2024-10-15
Improve the Windows installer to prevent users without admin privileges from getting stuck in a reboot loop

tnajdek:
This PR contains modifications to the install process as discussed in https://github.com/zotero/zotero/pull/4750#issuecomment-2410392033

1. The installer now checks if Zotero is running before proceeding. This is a quick check for a running process, which is simpler and faster than the more thorough check in `CheckExistingInstall`, where a list of files is tested for write access. If Zotero is running, the installer shows an error message and quits during normal installations. For silent installations, it attempts to kill `zotero.exe` process and quits unless successful.
2. Removed all instances of `/REBOOTOK` in the `UninstallOld` routine and removed the message introduced in #4750.
3. If a previous incomplete installation is encountered and the user is not an admin, the installer will now attempt to continue with a fresh installation. This might result in leftover junk in the target directory but should still produce a working installation, avoiding a reboot loop.

To check and kill the process, I've introduced a new NSIS plugin called `nsProcess`. The [download-nsis-plugins](https://github.com/zotero/zotero/commit/86f222f6fa68d525d65f32475f56575d583e0780#diff-445a8f541c9419a31e1f0cf35a6f3ca8ba9aa8471c048cac44a1792fb40b91a3) script has been extended to automatically fetch it.

Screencasts:

Install as a non-admin user, while Zotero is running:
https://github.com/user-attachments/assets/9dddd04e-5bbd-4395-ba91-2bf70db1d875

Silent install as non-admin user, while Zotero is running:
https://github.com/user-attachments/assets/bd3fc171-3b6b-4a37-a11a-36502318546f

Install as a non-admin user, while previous installation left `.moz-delete` file:
https://github.com/user-attachments/assets/7d7139bd-e26f-4a50-896c-cec774f202d9

### DIFF
--- a/app/win/download-nsis-plugins
+++ b/app/win/download-nsis-plugins
@@ -46,6 +46,11 @@ for plugin in $plugins; do
 	curl https://hg.mozilla.org/mozilla-central/raw-file/052d53200cf8/other-licenses/nsis/Plugins/$plugin.dll > $plugin.dll
 done
 
+curl https://nsis.sourceforge.io/mediawiki/images/1/18/NsProcess.zip > NsProcess.zip
+7z e NsProcess.zip "Plugin/nsProcessW.dll"
+mv nsProcessW.dll nsProcess.dll
+rm NsProcess.zip
+
 echo
 echo
 echo "Files downloaded to ./Plugins -- move to ${NSIS_DIR}Plugins"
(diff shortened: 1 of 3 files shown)


dstillman:
Merged

Issue #4766, opened 2024-10-18
Show "Convert to Standalone Note" for notes in "Change Parent Item…" dialog


dstillman:
Not "Convert to Standalone Attachment"

PR #4768, opened 2024-10-18
Better label to make item standalone in change parent item dialog

abaevbog:
More specific labels for the button to turn selected item(s) into standalone:
- "Convert to Standalone Sttachment(s)" when only attachments are selected
- "Convert to Standalone Note(s)" when only notes are selected
- "Convert to Standalone" otherwise (e.g. if a note and an attachment are selected together)

Fixes: #4766

### DIFF
--- a/chrome/locale/en-US/zotero/zotero.ftl
+++ b/chrome/locale/en-US/zotero/zotero.ftl
@@ -692,10 +692,18 @@ find-pdf-files-added = { $count ->
 
 select-items-dialog =
     .buttonlabelaccept = Select
+convert-to-standalone = Convert to Standalone
+select-items-convertToStandalone =
+    .label = { convert-to-standalone }
 select-items-convertToStandaloneAttachment =
     .label = { $count ->
-        [one] Convert to Standalone Attachment
-        *[other] Convert to Standalone Attachments
+        [one] { convert-to-standalone } Attachment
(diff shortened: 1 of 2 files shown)


chrome/locale/en-US/zotero/zotero.ftl
```diff
@@ -692,10 +692,18 @@ find-pdf-files-added = { $count ->
...
     .label = { $count ->
-        [one] Convert to Standalone Attachment
-        *[other] Convert to Standalone Attachments
+        [one] { convert-to-standalone } Attachment
+        *[other] { convert-to-standalone } Attachments
+}
+select-items-convertToStandaloneNote =
+    .label = { $count ->
+        [one] { convert-to-standalone } Note
+        *[other] { convert-to-standalone } Notes
```
dstillman, line 706: This is another situation where we really can't be obsessed with DRY. We have no way of knowing how "Convert to Standalone" is going to need to be translated in a given language, and there's no reason to expect that the string for "Convert to Standalone" is going to slow into a "Convert to Standalone Attachment" or "Convert to Standalone Note" string.

abaevbog: Gotcha! I really need to turn off DRY thinking while dealing with localization. I removed the `{ convert-to-standalone }` and swapped it for the proper "Convert to Standalone" string.

Issue #4770, opened 2024-10-21
Child attachment shows up in item pane after being moved to trash


dstillman:
https://forums.zotero.org/discussion/comment/477052/#Comment_477052

I can reproduce.

abaevbog:
In addition to that, restoring the trashed attachment from trash will not update the `attachments-box` as well.

To get the item pane properly updated when an attachment is trashed, it seems enough to update the `notify` to expect the `trash` event [here](https://github.com/zotero/zotero/blob/3e37c44c8221b13d5cf8eb9e7903f91dff8d984c/chrome/content/zotero/elements/attachmentsBox.js#L142), in addition to all other ones.

Updating the `attachments-box` after a trashed attachment is restored into the library is trickier. All sections, including `attachments-box` get re-rendered when the top-level item is selected, but the section is counted as [this._isAlreadyRendered](https://github.com/zotero/zotero/blob/3e37c44c8221b13d5cf8eb9e7903f91dff8d984c/chrome/content/zotero/elements/attachmentsBox.js#L193) and the restored attachment does not appear until we switch to another top-level item and back.

windingwind:
I can try to fix this, but in addition to that, the item tree isn't rendering the attachment row after restoring from trash as well, until switching to another collection and back.

Can reproduce with this:

```js
attachment.deleted = false;
attachment.saveTx();
```

This isn't something users would be able to do with the UI, but it can be triggered by plugins. Should fix that as well.

abaevbog:
> the item tree isn't rendering the attachment row after restoring from trash as well, until switching to another collection and back.

I can't seem to reproduce this one. The row is added back into the itemTree for me and I can see it if I expand the parent item. Is this the rough setup?


https://github.com/user-attachments/assets/d828067f-321d-4466-9797-89736df06955

windingwind:
> > the item tree isn't rendering the attachment row after restoring from trash as well, until switching to another collection and back.
> 
> I can't seem to reproduce this one. The row is added back into the itemTree for me and I can see it if I expand the parent item. Is this the rough setup?
> 
>  Screen.Recording.2024-10-24.at.1.57.05.PM.mov


https://github.com/user-attachments/assets/547046f7-02bf-40f8-8efd-8326cd2d3d60

PR #4777, opened 2024-10-22
Fix child attachment shows up in item pane after being moved to trash

windingwind:
fix: #4770

### DIFF
--- a/chrome/content/zotero/elements/attachmentsBox.js
+++ b/chrome/content/zotero/elements/attachmentsBox.js
@@ -139,9 +139,22 @@
 		}
 
 		notify(action, type, ids) {
-			if (!(action === 'add' || action === 'modify' || action === 'refresh' || action === 'delete')) {
+			if (!['add', 'modify', 'refresh', 'delete', 'trash'].includes(action)) {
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/elements/attachmentsBox.js
```diff
@@ -95,6 +95,13 @@
 			return this._preview;
 		}
 
+		get _renderDependencies() {
+			if (this.item?.isRegularItem()) {
+				return [...super._renderDependencies, this.item.getAttachments().join(',')];
```
dstillman, line 100: Is this really what you mean? If so, can you add a comment explaining it? Because the string concatenation looks really weird here.

windingwind: Yeah, actually it doesn’t matter if it’s a string or something else. It just needs to be something comparable to tell the attachment of the regular item is added/removed. I concat it as a string instead appending all the ids because we might want to keep the array length fixed.

windingwind: Comment added

PR #4778, opened 2024-10-22
macOS: Move Window -> Zotero to platformKeys

### LINKED ISSUE #4171: [Mac] Can't reopen main window after closing if reader windows are open
https://forums.zotero.org/discussion/114657/reopening-library-window-on-mac

There should be a Window menu option for this. E.g., in Thunderbird, `Mail & Newsgroups [Cmd-1]` shows up in standalone message windows.

AbeJellinek:
And change shortcut to Cmd-Shift-L to avoid conflicts with various font size and zoom shortcuts.

This makes Window -> Zotero available in every non-main window that uses `platformKeys.js`, which is Scaffold, the reader, and the basic viewer right now. The second commit extends that to include note windows as well. (Those don't actually show _in_ the Window menu, though, because they're opened with `openDialog()`.)

There's no great way to hide the Window -> Zotero menuitem when the main window is open. Messing with `windowMenu` children in `popupshowing` breaks macOS's auto-population. We could probably get around it with some kind of window observer, but I think it's actually better this way: the keyboard shortcut should work consistently, regardless of whether the main window is already open, and a menuitem allows that. Thunderbird's Tools -> Mail & Newsgroups menuitem also shows all the time.

Closes #4171

### DIFF
--- a/chrome/content/zotero/platformKeys.js
+++ b/chrome/content/zotero/platformKeys.js
@@ -3,7 +3,9 @@ document.addEventListener('DOMContentLoaded', (event) => {
 		return;
 	}
 	
-	// Don't need to depend on Zotero object here
+	const { AppConstants } = ChromeUtils.importESModule('resource://gre/modules/AppConstants.sys.mjs');
+	
+	// Can't depend on Zotero object in hidden window initialization
 	let isWin = AppConstants.platform == 'win';
 	let isMac = AppConstants.platform == 'macosx';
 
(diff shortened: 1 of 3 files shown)


dstillman:
Maybe Window → Library? (Not sure what happens once there are multiple library windows if we can't hide this in library windows…)

dstillman:
Ah, sorry. I thought I saw it in the main window, but I must've imagined that. And also I guess we'll still have the problem (in *other* windows) I mention once we support multiple library windows. But I guess we'll deal with that then.

I think Window → Library still makes sense.

PR #4780, opened 2024-10-23
Add support for importing KOReader & Calibre annotations

AbeJellinek:
We've seen a fair amount of interest in this:
https://forums.zotero.org/discussion/comment/477302/#Comment_477302
https://forums.zotero.org/discussion/comment/474501#Comment_474501
https://forums.zotero.org/discussion/comment/474458#Comment_474458
https://forums.zotero.org/discussion/comment/452541#Comment_452541
https://forums.zotero.org/discussion/comment/471329#Comment_471329
https://forums.zotero.org/discussion/113886/third-party-epub-reader-highlights-not-shown-in-zotero-7-epub-reader
https://forums.zotero.org/discussion/comment/451489#Comment_451489

And also interest in syncing Kindle highlights, which this makes possible via Calibre's Kindle sync:
https://forums.zotero.org/discussion/comment/455086/#Comment_455086

It turned out to be pretty easy to implement one-way import functions for both readers!

Needs https://github.com/zotero/reader/pull/148


mrtcode:
Where could I find an example file for testing?

mrtcode:
It works well!

Although in that case, when annotations are embedded, it didn't feel intuitive to have to select the same EPUB file to import annotations from it.

mrtcode:
Better!

But is it intentional to show one popup when there are no embedded annotations, and after selecting a file, to show another one?

mrtcode:
Looks good!

chrome/locale/en-US/zotero/reader.ftl
```diff
@@ -51,3 +51,23 @@ pdfReader-findInDocumentInput =
...
+pdfReader-import-from-epub-prompt-title = Import Ebook Annotations
+pdfReader-import-from-epub-prompt-text =
+    { -app-name } found { $count ->
+        [1] { $count } { $tool } annotation
+        *[other] { $count } { $tool } annotations
+    }, last edited { $lastModifiedRelative }.
+    
+    Any { -app-name } annotations that were previously imported from this ebook will be updated.
+pdfReader-import-from-epub-no-annotations-current-file =
+    This ebook does not appear to contain any KOReader or Calibre annotations.
```
dstillman, line 66: Our own forum threads aside, I think we should list Calibre first?

https://trends.google.com/trends/explore?date=today%205-y&q=koreader,calibre&hl=en

chrome/locale/en-US/zotero/reader.ftl
```diff
@@ -51,3 +51,23 @@ pdfReader-findInDocumentInput =
...
+    
+    If you have another copy in your KOReader or Calibre library, { -app-name } can import annotations from that file.
```
dstillman, line 68: And do we need to say this part? Can we just assume people can figure out where their files are?

Also sort of weird to list the specific other applications in the first sentence.

What about just this?

```
This ebook does not appear to contain any importable annotations.

{ -app-name } can import ebook annotations created in Calibre and KOReader.
```

l1mey112:
Is there a way to get this support for PDFs?

PR #4785, opened 2024-10-24
Moving items to libraries and my publications via context menu

abaevbog:
- added "My publications" `menuitem` to item context menu to add items to my publications without having to drag-drop
- added "Add to Another Library" `menu` to item context menu to be able to add items to collections from other libraries
- refactored addSelectedItemsToCollection to handle adding items to collections outside of current library
- also, prompt to add a new collection in addSelectedItemsToCollection whenever there is no target, for simplicity

The only question here is if we need to tweak `newCollectionDialog.xhtml` to make it possible to create a collection in non-current library when "New collection..." is selected?


dstillman:
1. "My Publications" should be within the top section below collections, with its proper icon.
2. The menu below should be "Other Libraries", not "Add to Another Library". It's just showing collections from other libraries. (Yes, you can also add to the library roots, but the ostensible function here is "Add to Collection".)

dstillman:
> The only question here is if we need to tweak newCollectionDialog.xhtml to make it possible to create a collection in non-current library when "New collection..." is selected?

Why would we have to tweak newCollectionDialog.xhtml? That's just getting the libraryID from `io`, no?

abaevbog:
> Why would we have to tweak newCollectionDialog? That's just getting the libraryID from io, no?

Sorry I didn't make it clear what I meant. That's just me following up on this comment from earlier: https://github.com/zotero/zotero/blob/2aa6444fb8041093d684337a650c08761540b14a/chrome/content/zotero/zoteroPane.js#L4307-L4309

Currently, we have on "New Collection..." option at the very top of the menu, which opens up `newCollectionDialog`, from where you can add a new collection in this current library. So from the context menu, you can only create a collection in the library where you currently are. Do we want to expand it to make it possible to create a new collection in another group before sending items there?

We could tweak the `newCollectionDialog` to be able to navigate more than one library. Or we could add  "New Collection..." into the menu of every group, which would probably be easier. 

The only reason why I didn't add the "New Collection..." into every group menu right away is that we already have one "special" menuitem on top, above the separator (the `menuitem` for the actual group), so I though it might be too busy?

dstillman:
Oh, I see. Yeah, I think it should just be an option in each library within Other Libraries:

```
Group 1
---------------
New Collection…
---------------
Collection 1
Collection 2
Collection 3
```

It's weird enough that we're using "Add to Collection…" for copying to another library. I think it would be even stranger if you could also trigger this other action via the main New Collection option. This way, you're at least going to "Other Libraries" and selecting another library, consistent with copying to an existing collection in that library.

Or put another way: in the top-level menu, the current library name is implied, so this is just transposing the rest of the menu into each entry in Other Libraries.

abaevbog:
That makes sense! How about this? 

Screenshot 2024-10-25 at 1 00 35 PM

chrome/content/zotero/zoteroPane.js
```diff
@@ -4196,29 +4224,205 @@ var ZoteroPane = new function()
...
+				let objectsForMenuItems = [library, { newCollection: true }, ...Zotero.Collections.getByLibrary(library.libraryID, true)];
+				let linkedItems = {};
+				for (let obj of objectsForMenuItems) {
+					let disableMenuItem = true;
+					for (let item of items) {
+						// Needed (in addition to the check in createMenuForTarget above) to disable "New Collection" menuitem
+						if (!library.editable) continue;
+
+						let movingLinkedFile = item.attachmentLinkMode == Zotero.Attachments.LINK_MODE_LINKED_FILE;
+						let linkedItem = linkedItems[item.id] || await item.getLinkedItem(library.libraryID, true);
```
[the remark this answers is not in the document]
abaevbog: Ah, good catch, thank you! I had that part refactored at the last moment to be more concise and missed this bit

chrome/content/zotero/zoteroPane.js
```diff
@@ -4196,29 +4224,205 @@ var ZoteroPane = new function()
...
+				await collection.addItems(itemIDs);
+			});
+		}
+	};
 
-		await Zotero.DB.executeTransaction(
-			() => collection.addItems(items.map(item => item.id)));
+	this.addSelectedItemsToMyPublications = async function () {
+		let items = Zotero.Items.keepTopLevel(this.getSelectedItems());
+		let io = window.ZoteroPane.showPublicationsWizard(items);
```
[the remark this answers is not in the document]
abaevbog: Ah, yes. I grabbed that bit from `collectionTree` where we have` window.ZoteroPane...` and forgot to change it to `this.showPublicationsWizard`.

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -1613,10 +1613,12 @@ Zotero.Utilities.Internal = {
 	 * 		Receives the event and libraryOrCollection for given item.
 	 * @param {Function} disabledPred If provided, called on each library/collection
 	 * 		to determine whether disabled
+	 * @param {Bool} alwaysMenuForLibrary If true, will always create a menu for library, even if
+	 * 		there are no child collections
 	 *
 	 * @return {Node<menuitem>|Node<menu>} appended node
 	 */
-	createMenuForTarget: function (libraryOrCollection, elem, currentTarget, clickAction, disabledPred) {
+	createMenuForTarget: function (libraryOrCollection, elem, currentTarget, clickAction, disabledPred, alwaysMenuForLibrary) {
```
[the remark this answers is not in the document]
abaevbog: Good point! There were indeed quite a few positional parameters. I changed it to be `createMenuForTarget: function (libraryOrCollection, elem, options)` and refactored instances where we currently use `createMenuForTarget` to not throw any warnings.

To be clear, we want to log `Zotero.warn` here just for future reference? Or is it for plugins that may have used it?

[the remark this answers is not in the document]
abaevbog: Gotcha! That makes sense.

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -1608,15 +1608,24 @@ Zotero.Utilities.Internal = {
...
 	 * @param {Library|Collection} libraryOrCollection
 	 * @param {Node<menupopup>} elem Parent element
-	 * @param {Zotero.Library|Zotero.Collection} currentTarget Currently selected item (displays as checked)
-	 * @param {Function} clickAction function to execute on clicking the menuitem.
+	 * @param {Object} options
+	 * @param {Zotero.Library|Zotero.Collection} options.currentTarget Currently selected item (displays as checked)
+	 * @param {Function} options.clickAction function to execute on clicking the menuitem.
 	 * 		Receives the event and libraryOrCollection for given item.
-	 * @param {Function} disabledPred If provided, called on each library/collection
+	 * @param {Function} options.disabledPred If provided, called on each library/collection
```
dstillman, line 1615: Since this is exposed now, can we use a clearer name? `disabledPredicate`? Or a less-technical name? `isDisabledCheck`? `isDisabledFn`?

abaevbog: Makes sense! We use `disabledCheck` phrasing in `dataObjects.js`, so I went with `disabledCheck` here as well. How does that look?

abaevbog:
One question: I just noticed that after an item A is copied from one library to another as item B, we record that relations as `owl:sameAs`. It is returned when I call `item.getRelations()` on the original item A. Then, if I delete item B from another library, that relations does not go away from item A. If I call  `item.getRelations()` on item A again, nothing is returned but I see a logged message `Referenced linked item 'http://zotero.org/groups/.../items/...' not found in Zotero.Item::getLinkedItem()`. Should the `owl:sameAs` relations of item A be deleted as well when item B is erased? Or am I missing something here?

dstillman:
The + button seems to be broken again now (similar to what we fixed in 62de988cb80c65d3e7953e072bdaba534b578e25).

abaevbog:
Ah, right. Thanks for pointing that out. I had somewhat hardcoded the `menupopup` to which all `menuitems` were being added to make it easier to handle tests (having, admittedly forgotten that it's a different `menupopup` in `LibrariesCollectionsBox`).

It should be fixed now, and I tweaked the tests that should be ok with this change.

chrome/content/zotero/zoteroPane.js
```diff
@@ -4151,74 +4153,277 @@ var ZoteroPane = new function()
...
+		if (this.getSelectedLibraryID() == Zotero.Libraries.userLibraryID) {
+			myPublicationMenuItem = document.createXULElement("menuitem");
+			myPublicationMenuItem.setAttribute("label", Zotero.getString("pane.collections.publications"));
+			myPublicationMenuItem.addEventListener("command", (_) => {
+				this.addSelectedItemsToMyPublications();
+			});
+			myPublicationMenuItem.setAttribute("image", "chrome://zotero/skin/16/universal/publications.svg");
+			popup.append(myPublicationMenuItem);
+			// Cannot add top-level notes/attachment or items that are already in my publications
+			if (items.some(item => item.isNote() || item.isAttachment() || item.inPublications)) {
```
dstillman, line 4242: Better to do `!item.isRegularItem()`, so this isn't accidentally enabled for some other item type later.

chrome/content/zotero/zoteroPane.js
```diff
@@ -4151,74 +4153,277 @@ var ZoteroPane = new function()
...
+					},
+					disabledCheck: (_) => !library.editable,
+					alwaysMenuForLibrary: true
+				}
+			);
+			otherLibrariesPopup.append(libraryMenu);
+			addNewCollectionMenuItem(libraryMenu, libraryMenu.querySelector("menuseparator"), library);
+		}
+		// Disable destinations where items cannot be added. Essentially a duplicate of
+		// major parts of CollectionTree.canDropCheck and CollectionTree.canDropCheckAsyns
```
dstillman, line 4277: → `canDropCheckAsync`

dstillman: For collection copying via context menu, I [had mentioned](https://github.com/zotero/zotero/pull/4420#issuecomment-2406579899) pulling out the `dataType == 'zotero/collection'` stuff to avoid duplicating it and having conditions get out of sync. We didn't do that yet because of time, but ideally we'd do the same thing here for items.

And probably same with the stuff from `onDrop` below. We really don't want to have copying code in multiple places if we can help it, among other things because this stuff will definitely be changing as we improve cross-library copying, and we always want drag-and-drop and the menus to trigger the same UI and logic. This, not localization, is the place for DRY!

abaevbog: Right, totally! I believe, while putting this together, the a11y review was still happening and I wasn't sure if these changes would or would not be required to finish it up. So I did go the same route as we did in #4420, thinking of it as a quick followup to other drag-drop alternatives PRs. But now that VPAT is settled, there's no reason to not refactor things to avoid duplication, for sure. I'll go through this logic to move it to `Zotero.Collection` and have it be used here and in collectionTree during drag-drop.

abaevbog: So maybe something like this? I pulled out the `canDropCheck` and `canDropCheckAsync` into `Zotero.Item.canAddToContainer` and `Zotero.Item.canAddToContainerAsync` that both `collectionTree.onDrop` and contextmenus from `ZoteroPane` share. I think in terms of readability and overall structure, it is a big improvement!

While we are at it, would it make sense to do the same for collections drag-drop as well here? Or is it better to do in a separate PR, since it's technically not relevant to these changes.

dstillman:
Sorry, this will need to be rebased.

chrome/content/zotero/xpcom/data/item.js
```diff
@@ -4423,6 +4423,77 @@ Zotero.Item.prototype.addToCollection = function (collectionIDOrKey) {
...
 };
 
+/**
+ * Check sync conditions for if an item can be added into a given container via drag-drop or via context menu options.
+ * @param {Zotero.Collection|Zotero.Library|Object} target - The object to which we check if an item
+ * can be added. Must be a collection, a library or a collectionTreeRow.ref object { libraryID, treeViewID }
+ * (for "My Publications" pseudo collection).
+ * @return {Boolean} result - true if there is no sync condition preventing an item from being added, or false otherwise.
+ */
+Zotero.Item.prototype.canAddToContainer = function (target) {
```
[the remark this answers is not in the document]
abaevbog: That sounds good, I changed it for `canAddToTarget`. I began with obvious `canAddToCollection` but then figured that it needed to be a bit more general than that to account for libraries and "My Publications".

chrome/content/zotero/xpcom/data/item.js
```diff
@@ -4423,6 +4423,77 @@ Zotero.Item.prototype.addToCollection = function (collectionIDOrKey) {
 	this.setCollections(this._collections.concat(collectionID));
 };
 
+/**
+ * Check sync conditions for if an item can be added into a given container via drag-drop or via context menu options.
+ * @param {Zotero.Collection|Zotero.Library|Object} target - The object to which we check if an item
+ * can be added. Must be a collection, a library or a collectionTreeRow.ref object { libraryID, treeViewID }
+ * (for "My Publications" pseudo collection).
```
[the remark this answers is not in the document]
abaevbog: That's fair, probably a bit cleaner to take `CollectionTreeRow` and then, if it is provided, we'll unwrap it to fetch its `.ref`.

abaevbog:
Rebased!

I do keep wondering about the relations behavior I mentioned [above](https://github.com/zotero/zotero/pull/4785#issuecomment-2463926135). During testing, I repeatedly moved a few items from "My Library" into groups, deleted them, and copied them into groups again. And now, I have an item in My Library with 20 `owl:sameAs` relationship entries that point at items in groups that were earlier deleted. It doesn't seem to cause any actual issues but is that expected behavior?

chrome/content/zotero/xpcom/data/item.js
```diff
@@ -4475,6 +4475,83 @@ Zotero.Item.prototype.addToCollection = function (collectionIDOrKey) {
...
+	}
+	if (this.isAttachment() && targetGroup.libraryType == 'group') {
+		// Cannot move linked attachments to another group
+		if (this.attachmentLinkMode == Zotero.Attachments.LINK_MODE_LINKED_FILE) return false;
+
+		// Cannot move an attachment if group's files are not editable
+		if (!targetGroup.filesEditable) return false;
+	}
+	return true;
+};
```
dstillman, line 4525: I assume all the `Zotero.debug()` lines were removed because they would spam the terminal when building the menu, but they're useful for debugging and when people ask why they can't copy something. Maybe add an `options` argument to these with a `debug` flag and set that to true for drags?

abaevbog: Yes, that's exactly why I hadn't included them initially. But it does make sense that we want to see why something can't be copied over to a collection during drag. So I added an options `{ debug }` to `canAddToTarget` and `canAddToTargetAsync` to log why something can't be added. It's only `true` during drag-drop.

dstillman:
https://forums.zotero.org/discussion/120660/add-item-to-another-library

PR #4790, opened 2024-10-25
Pass document to ItemTree API renderCell

windingwind:
https://groups.google.com/g/zotero-dev/c/4jqa8QIk6DM/m/qbq55IUbBAAJ


windingwind:
It's essential to pass the current `document` to the `renderCell` function when creating elements. 

While you can create elements in a different `document` and insert them into the current one if they share the same privilege level, this can result in potential memory leaks or render failures.

Issue #4782, opened 2024-10-24
Printing bibliography resets dark mode to light mode


dstillman:
https://forums.zotero.org/discussion/118990/printing-bibliography-reset-dark-mode-lost-tags

I can reproduce.

abaevbog:
When a browser is initialized, there is a call to [updateFxaToolbarMenu](https://searchfox.org/mozilla-central/source/browser/base/content/browser-init.js#217), which in turn [tries](https://searchfox.org/mozilla-central/source/browser/base/content/browser.js#814) to call a method from `chrome://browser/content/browser-sync.js`, which tried to load [UIState.sys.mjs](https://searchfox.org/mozilla-central/source/browser/base/content/browser-sync.js#17), which we currently don't have. This is what throws the first error from the stack. So I suppose we'd either want to have `UIState.sys.mjs` available (if it is useful for something else in the future), or maybe we could patch `updateFxaToolbarMenu` to just do nothing.

abaevbog:
Actually, `updateFxaToolbarMenu` error seems to not be what's causing this breakage. There are a few other errors logged, so I'll play around to see if I run into the actual root of the issue.

abaevbog:
Aha, so it looks like [LightweightThemeConsumer](https://searchfox.org/mozilla-central/source/toolkit/modules/LightweightThemeConsumer.sys.mjs) is to blame. If we remove it from browser.js [here](https://searchfox.org/mozilla-central/source/browser/base/content/browser-init.js#112), printing does not change the color scheme for me. And our normal color scheme changes are not affected.

PR #4807, opened 2024-10-31
fix color scheme getting reset during printing

abaevbog:
[LightweightThemeConsumer](https://searchfox.org/mozilla-central/source/toolkit/modules/LightweightThemeConsumer.sys.mjs) interacts with the color scheme, so removing it prevents the color scheme from changing whenever `zoteroPrint` runs.

Fixes: #4782

### DIFF
--- a/app/build.sh
+++ b/app/build.sh
@@ -385,6 +385,8 @@ cat "$CALLDIR/assets/chrome.manifest" >> chrome.manifest
 replace_line 'handle: function bch_handle\(cmdLine\) {' 'handle: function bch_handle(cmdLine) {
   \/\/ TEST_OPTIONS_PLACEHOLDER
   ' modules/BrowserContentHandler.sys.mjs
+# prevent color scheme getting reset to 'light' during printing
+replace_line 'new LightweightThemeConsumer\(document\);' '\/\/new LightweightThemeConsumer\(document\);'  chrome/browser/content/browser/browser.js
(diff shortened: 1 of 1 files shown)


dstillman:
Since this seems to be called in `onBeforeInitialXULLayout()`, not any printing-specific code, do we understand what this is doing, and are we confident disabling this doesn't break anything for normal rendering?

abaevbog:
My understanding is that, most importantly, `LightweightThemeConsumer` is called by `browser-init.js` in [browser.xhtml](https://searchfox.org/mozilla-central/source/browser/base/content/browser.xhtml#111), which is for a firefox window. Normally, we should never be encountering it. `BROWSER_CHROME_URL` is set to `browser.xhtml` by default ([here](https://searchfox.org/mozilla-central/source/browser/moz.configure#20), I believe), but we override it to be `zoteroPane.xhtml` while [fetching xulrunner](https://github.com/zotero/zotero/blob/main/app/scripts/fetch_xulrunner#L118). 

The only reason why we are running into it here is that there is a firefox window that [pops up briefly](https://github.com/zotero/zotero/blob/main/chrome/content/zotero/actors/ZoteroPrintParent.jsm#L39) during `zoteroPrint` handling. 

It looks like there are [a few different ways](https://searchfox.org/mozilla-central/source/browser/base/content/contentTheme.js#155) in which the theme can change in firefox and `LightweightThemeController` can [override](https://searchfox.org/mozilla-central/source/toolkit/modules/LightweightThemeConsumer.sys.mjs#301) the color scheme would have been set otherwise. We seem to just rely on `prefers-color-scheme` to determine if it's a dark scheme or not, so removing this does not conflict with changing the preferences for me.

Issue #4646, opened 2024-08-29
Unloaded reader tab titles don't update on `tabs.title.reader` change


AbeJellinek:
1. Open a bunch of PDFs/EPUBs
2. Change the "Show tabs as" setting

Only loaded tabs' titles will change. Other tabs will change when loaded.

This might be difficult to address with our current system, but easier after unloading becomes the reader's responsibility (https://github.com/zotero/zotero/issues/4228#issue-2345474715).

https://forums.zotero.org/discussion/comment/472698/#Comment_472698

abaevbog:
I just ran into a similar issue that just changing the name of the item (e.g. in itemBox) will not rename unloaded tabs as well. 

Currently, the logic of updating the title is [baked](https://github.com/zotero/zotero/blob/3e37c44c8221b13d5cf8eb9e7903f91dff8d984c/chrome/content/zotero/xpcom/reader.js#L584C2-L584C2) into the reader instance, so if there is no reader, there is no way to change the tab name. At the same time, we wouldn't be able to move this into `Zotero_Tabs`, since it does need to work for standalone reader windows. So yeah, it's tricky! 

Would it make sense to move this into `Zotero.Reader`, and we'd pass the relevant item as a parameter (instead of using `this.item` of the instance, like `Zotero.Reader.updateTitle(item))`? It feels a bit clunky to have it next to very general functions like `getByTabID` but at least that way, tabs and reader windows should be able to share this functionality?

dstillman:
> moving the loading/unloading logic into the reader

Just to note that we'll probably allow notes and other library views in tabs as well, so if there's shared loading/unloading logic, we might want to keep that global.

mrtcode:
Yeah, I was considering moving the tab loading/unloading logic into the reader, but that's a good point about having shared logic that could be used by other tab types.

For now, it's probably better to implement a temporary fix for the title issue, as it's still unclear how the tab loading/unloading will work in the future.

PR #4813, opened 2024-11-04
Update tab titles independently of the reader

abaevbog:
Move handling of tabs' renaming when an item is modified from `Reader` into `Zotero_Tabs`, so that the titles of unloaded tabs can get properly renamed.

Have the `Reader` and `Zotero_Tabs` share `Zotero.Utilities.Internal.constructItemTitle` to handle title updates in both tabs and standalone reader windows.

Fixes: #4646

### DIFF
--- a/chrome/content/zotero/xpcom/utilities_internal.js
+++ b/chrome/content/zotero/xpcom/utilities_internal.js
@@ -2488,7 +2488,56 @@ Zotero.Utilities.Internal = {
 		}
 
 		return textContent;
-	}
+	},
+
+	/**
+	 * Costruct item title for a given attachment.
+	 * @param itemID - itemID of the attachment to be opened in the reader.
+	 */
+	async constructItemTitle(itemID) {
+		let type = Zotero.Prefs.get('tabs.title.reader');
+		let item = Zotero.Items.get(itemID);
+		let readerTitle = item.getDisplayTitle();
+		let parentItem = item.parentItem;
+		if (type === 'filename') {
+			readerTitle = item.attachmentFilename;
+		}
+		else if (parentItem) {
+			let attachment = await parentItem.getBestAttachment();
+			let isPrimaryAttachment = attachment && attachment.id == item.id;
+			
+			let parts = [];
+			// Windows displays bidi control characters as placeholders in window titles, so strip them
+			// See https://github.com/mozilla-services/screenshots/issues/4863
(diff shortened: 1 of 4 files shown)


chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -2488,7 +2488,53 @@ Zotero.Utilities.Internal = {
...
 
 		return textContent;
-	}
+	},
+
+	/**
+	 * Costruct item title for a given attachment.
+	 * @param itemID - itemID of the attachment to be opened in the reader.
+	 */
+	async constructItemTitle(itemID) {
```
dstillman, line 2497: This definitely isn't descriptive enough, but also I don't think this function needs to be in `Utilities.Internal`, since it's not a general-purpose function. Can we put it in `Zotero.Reader` as `getTabTitle()`? If that doesn't work, we could move it to `Zotero.Item` next to `getDisplayTitle()`and plan to support notes or other item types later.

abaevbog: Gotcha, that makes sense. I suppose my thinking was that there may be a tab without a reader loaded in it, so I thought it would have to go somewhere that's not tied to the reader. And then I noticed `Zotero.Utilities.Internal.renderItemTitle` (which, granted, is way more general and used in a number of places) and thought to place this title-related function next to it. 

But there is no reason to not keep in `Zotero.Reader`, if that is more appropriate. I moved it to `Zotero.Reader.getDisplayTitle` with a few of the semantics fixes from above. There was also a small conflict, so I rebased because of it.

dstillman: Actually, maybe `Zotero.Item::getTabTitle()` makes more sense? We'll almost certainly support notes in tabs at some point, and calling a single method is simpler than adding a separate `getTabTitle()` to whatever code handles note tabs and calling each handler's `getTabTitle()` conditionally from each place we need to call this.

abaevbog: Ok, yeah, if we want to account for notes in tabs for the future, `Zotero.Reader` would not be appropriate, since there would be no actual reader in the tab. So I moved it to `Zotero.Item` next to other title-related functions. I think it feels quite natural to have `Zotero.Item::getTabTitle()` next to `Zotero.Item::getDisplayTitle()`, actually!

PR #4816, opened 2024-11-05
Fix item tree hooks break ui

windingwind:


windingwind:
Also updated in https://github.com/zotero/zotero/pull/4232#pullrequestreview-2416025461 so that for future APIs all hooks will be executed with proper error handling.

PR #4826, opened 2024-11-07
added ellipses to 'Install Browser Connector' menu item

### LINKED ISSUE #4693: Add ellipsis to "Install Browser Connector"
https://forums.zotero.org/discussion/comment/474533/#Comment_474533

franklindyer:
fixes #4693


dstillman:
1. It needs to be a proper ellipsis character.
2. Only the en-US locale should be changed. The other locales get changed on Transifex (and might, say, use different spacing or a different character).

PR #4828, opened 2024-11-08
Allow quicksearch to stretch when there is space in the middle pane

abaevbog:
Per https://forums.zotero.org/discussion/119312/search-window-in-zotero-7-is-too-small. I suppose, when the middle pane is wide, we do have plenty of space between toolbar buttons and the quickSearch, so allowing it to stretch a bit does not hurt? Maybe the exact width values need to be tweaked - I started with max-width of `300px`.

https://github.com/user-attachments/assets/8033b74d-7613-41d3-9985-da92de3258c2

Also, have spinner always occupy space, so that it does not push quickSearch out of the bounds of the middle pane when it is very narrow. Respectively, make min-width of quicksearch a bit smaller. This resolves the issue that is present currently on `main`, without these quicksearch stretching tweaks. This is the current behavior I am referring to:

https://github.com/user-attachments/assets/797c6db4-970a-4466-8dc5-95415645d404

### DIFF
--- a/chrome/content/zotero/zoteroPane.xhtml
+++ b/chrome/content/zotero/zoteroPane.xhtml
@@ -1334,7 +1334,7 @@
 												</toolbarbutton>
 												
 												<spacer flex="1"/>
-												<image id="zotero-tb-search-spinner" class="zotero-spinner-16"/>
+												<image id="zotero-tb-search-spinner" class="zotero-spinner-16" style="display: inline;visibility: hidden;"/>
 												<quick-search-textbox id="zotero-tb-search" timeout="250"
 														onkeydown="ZoteroPane_Local.handleSearchKeypress(this, event)"
 														oninput="ZoteroPane_Local.handleSearchInput(this, event)"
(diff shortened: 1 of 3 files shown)


chrome/content/zotero/zoteroPane.xhtml
```diff
@@ -1334,7 +1334,7 @@
 												</toolbarbutton>
 												
 												<spacer flex="1"/>
-												<image id="zotero-tb-search-spinner" class="zotero-spinner-16"/>
+												<image id="zotero-tb-search-spinner" class="zotero-spinner-16" style="display: inline;visibility: hidden;"/>
```
dstillman, line 1337: Can just move this to _quickSearchTextbox.scss, right? We're already referencing this there.

abaevbog: Oh yeah, totally! Somehow I missed that `zotero-tb-search-spinner` was already in `_quickSearchTextbox.scss`.

PR #4837, opened 2024-11-13
Context pane notes list focus handling edits

abaevbog:
- fix the issue of focus not landing on search field of the notes pane when it is opened on tab from the reader
- shift-tab from the `sidenav` when notes pane is opened will focus the search input as well, instead of the last focusable node in the pane (e.g. "X more..." button at the bottom)
- tab from a `note-row` in the list of notes will focus the `sidenav`
- shift-tab from a `note-row` will focus the header of the `collapsible-section`. I thought focusing the reader right away would be too much, since one might need to get to the `+` button to add a note, or something like that. And once you are at a section header, you can jump to the header of the previous section (if there is one) with `arrowUp`, at which point, you are at the top of the notes pane.

Per https://github.com/zotero/zotero/pull/4235#issuecomment-2472397642

### DIFF
--- a/chrome/content/zotero/elements/contextNotesList.js
+++ b/chrome/content/zotero/elements/contextNotesList.js
@@ -185,6 +185,7 @@
 		};
 
 		// ArrowUp/Down navigation between notes
+		// Tab from a note-row focuses sidenav, Shift-Tab from a note focuses the section header
 		_handleKeyDown = (event) => {
 			if (event.target.tagName !== "note-row" && !event.target.classList.contains("more")) return;
 			if (event.key == "ArrowDown") {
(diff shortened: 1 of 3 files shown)


dstillman:
This looks good, except if you're on a note and tab to the sidebar, and then Shift-Tab, you end up back in the note search field instead of on the note you were on before.

dstillman:
Oh, which I guess you're saying was on purpose, but I'm not sure why? Just feels wrong.

dstillman:
And to be clear, I'd expect it to refocus the specific element in the list that was focused before, as we do in the annotations tab of the left sidebar.

abaevbog:
> Oh, which I guess you're saying was on purpose, but I'm not sure why? Just feels wrong.

I noticed that before this tweak, Shift-Tab would move focus back from the sidenav, which would land on the last focusable element (e.g. "X more" button at the bottom) and felt quite weird after playing around with it. So I thought that simulating a tab into the notes list (to land on the search input) would be less weird, that way focus is at least at the top. But it's a great idea to add an extra step to remember what note was last focused (just with a special attribute, maybe) and focusing that, if it exists.

dstillman:
Yeah, I think it's pretty much just identical to the Annotations tab. You can move up/down in that list and Tab or Shift-Tab out of it, and then if you go back in, you're on the same entry.

dstillman:
Similarly, I guess, we might want to do the same when tabbing forward in. So if you use down-arrow to move down the list, Shift-Tab back to the section header disclosure button, and Tab back, you'd end up on the same note row.

This is also, of course, how the collection and item trees work.

dstillman:
Let's get this in and tweak that separately. This is already an improvement.

Issue #4814, opened 2024-11-05
Windows: Citation dialog locator dropdown display issues


adomasven:
Reports: https://forums.zotero.org/discussion/119275/word-plugin-issue-with-changing-page-to-chapter-in-citations#latest
https://forums.zotero.org/discussion/119405/pc-7-09-cursor-not-aligned-with-menu-selection-when-adding-citation

I am unable to reproduce but I'm catching up with some Windows updates.

abaevbog:
@adomasven, I ran into what looks very similar to the issue described in these posts. Setting a `max-height` on the popup where locators are listed does this for me:


https://github.com/user-attachments/assets/e54b04f3-5b8d-405a-94f2-64260c433982

abaevbog:
Not sure why it would occur to other people without having `max-height` changed by plugins and such though.

A few more observations:
- this does not occur for me with other `menulist`s outside of QF. Setting `max-height` on itemType dropdown in itemBox does not cause such issues
- I assumed it might be caused by having a `menupopup` inside of a `panel`. But firefox "Bookmark this page" `panel` also has a `menulist` and changing the height of the `menupopup` seems to behave ok there.
- I tried to remove all stylesheets from QF besides `chrome://global/skin/*`, and this still would occur

abaevbog:
Aha, I think I have something. Removing `sizetopopup="none"` from the [menulist](https://github.com/zotero/zotero/blob/a6c178dfe81e1f44f9a8a33085b5d5fe2fb6695b/chrome/content/zotero/integration/quickFormat.xhtml#L100) prevents the breakage if I add `max-height`. 

I'm going through mozilla codebase and can't find exactly what `sizetopopup` is meant to do... But nothing seems to be broken if I remove it, and most other `menulist`s don't have it. So I wonder if it's worth removing it and checking in with people who were running into this without `max-height` css edits?

abaevbog:
Oh, in addition, removing `sizetopopup` seems to help with the width sizing of the popup https://github.com/zotero/zotero/issues/4805. 

<img width="515" alt="Screenshot 2024-11-14 at 2 33 50 PM" src="https://github.com/user-attachments/assets/5b56b865-e733-4b5f-9614-d07627ecbe5e">

vs

<img width="554" alt="Screenshot 2024-11-14 at 2 42 20 PM" src="https://github.com/user-attachments/assets/792c5705-92b5-4cbf-b531-d788b90dc03b">

abaevbog:
[...]

Issue #4805, opened 2024-10-31
Citation Dialog: Locator select is too narrow and tall


adomasven:
It shouldn't take up all the available vertical space on the screen and locators shouldn't be cut off.

![image](https://github.com/user-attachments/assets/cf9b8190-6cda-4be3-ba1d-84186b717e98)

PR #4846, opened 2024-11-14
QuickFormat: potential fix to locator popup not properly rendering

abaevbog:
This may help with the issue of locator popup sometimes not being properly rendered per https://github.com/zotero/zotero/issues/4814#issuecomment-2477538297

Also, it allows the popup to have a proper width on windows when "Always show scrollbar" setting is on.

Addresses: #4814
Addresses: #4805


dstillman:
Pushing a new beta now, so let's include this and ask affected people to test.

abaevbog:
Ok, great! Not a guarantee but worth a shot.

PR #4847, opened 2024-11-15
Context pane  notes list focus handling followup

abaevbog:
- remember the last focused `note-row` in the `contextPane` and, if possible, refocus it on shift-tab from `sidenav`, on tab from the reader into the notes pane, or on tab into the notes list from the collapsible-section. Otherwise, focus the search input. Per: https://github.com/zotero/zotero/pull/4837#issuecomment-2478108186
- when a note is opened in the context pane, let `shift-tab` from `sidenav` place focus on the `links-box`. That way, it is not skipped during tab navigation.
- fix focus not entering `note-editor` on tab from the "Go back" button and instead wrapping around to the current tab

### DIFF
--- a/chrome/content/zotero/elements/contextNotesList.js
+++ b/chrome/content/zotero/elements/contextNotesList.js
@@ -51,6 +51,8 @@
 		_numVisible = 0;
 		
 		_hasParent = false;
+
+		_lastFocusedNote = null;
 		
 		get notes() {
 			return [...this._itemNotes, ...this._allNotes];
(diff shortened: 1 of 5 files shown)


dstillman:
- Tab to the notes list
- Down-arrow to a given row
- Click back on the content page
- Press Tab. Seemingly nothing has focus.
- Tab again. The note row has focus again.
- If I then press Shift-Tab, focus correctly goes back to All Notes, which was skipped when moving forward.

dstillman:
But this is an improvement, so let's get this in a beta for now.

PR #4850, opened 2024-11-18
Add item tree and info box update API

windingwind:
For APIs that have lifecycle control, e.g. item pane section,
the `update` is passed in the init hook;
For APIs without lifecycle, we cannot pass a value.
Instead, we provide a method to update in the API instance.


test/tests/pluginAPITest.js
```diff
@@ -359,6 +359,34 @@ describe("Plugin API", function () {
...
+			let item = new Zotero.Item('book');
+			await item.saveTx();
+			await ZoteroPane.selectItem(item.id);
+			
+			let rowID = await waitForRegister(defaultOption);
+
+			let rowElem = infoSection.querySelector(`[data-custom-row-id="${rowID}"]`);
+			let valueElem = rowElem.querySelector(".value");
+
+			let oldValue = valueElem.value;
```
dstillman, line 373: Isn't this sort of awkward and brittle? This should be totally abstracted — you shouldn't have to know anything about the DOM to change the value, and we should be able to change the DOM without breaking plugins.

Would it make more sense to return an object from `registerInfoRow()` and allow 1) changing parameters (e.g., multiline, readonly), 2) setting a value manually, and 3) unregistering with a method instead of needing to use `unregisterInfoRow()`.

windingwind: Thanks for the reply, I also want to discuss this. This updating the info box custom row is not in the initial design, it's motivated by https://groups.google.com/g/zotero-dev/c/LlIz_maLBCU/m/A4kMrrXFAAAJ

To clarify:

> you shouldn't have to know anything about the DOM to change the value

This isn't really about modifying a specific value on DOM to update something, it's the `rowID` that you would normally get from the return of the `register` call, and we internally mark that on the DOM element for locating the corresponding element.

In this test, it's just for test - we manually _hack_ the rendered value by DOM query and then call the `update` to see if it's restored, so that we can be sure that the update is called. It's never a way that we would allow plugins use, and the major reason why we do this _hack_ in this test is that we try not to expose the DOM directly to developers. Once they have the DOM in hand, they will insert all kinds of things there.


> Would it make more sense to return an object from registerInfoRow() and allow 1) changing parameters (e.g., multiline, readonly), 2) setting a value manually, and 3) unregistering with a method instead of needing to use unregisterInfoRow()?

I think it's better to keep all APIs in the same/similar shape, it's good for plugin developers as they only need to learn how to use the APIs once, and it's also good for use because that allows us to reuse code and keep the codebase tidy. And since we currently decide all APIs' register call returns a key that you can later use for unreigstering, there's not strong reason to specifically return an object for this API.

For 1), I think generally you won't want to change attributes. It can be weird that a field can become to readonly and suddenly turn back to be editable.

For 2), that's why I made this PR - you can call the `update` and then this specific field will get the latest value from the `onGetData`.

----

Back to the discussion about how to update the components registered by APIs:

1. By design, different APIs can have different behaviors regarding updating the content. For example:
  - The item tree is never able to update a single cell or a single column - it just always updates the whole table and the render is internally handled by React.
  - The item pane sections are only allowed to be manually updated after they are already rendered. Their rendering is controlled by the item details CE.

2. Since these APIs have different behaviors or requirements for updating, we cannot simply propose a function for all APIs that can be called in the same format, that's why we didn't expose that ability to plugins before, and that's also what I am trying in this PR.
3. After this PR, plugins can call the update by:

```js
Zotero.ItemTreeManager.updateColumns();

let rowID = Zotero.ItemPaneManager.registerInfoRow({...});
Zotero.ItemPaneManager.updateInfoRow(rowID);
```

Probably we can add a `Zotero.ItemPaneManager.updateSection()`, but since currently plugins can get an `update` function from the `onInit` callback of the API, that's not in a hurry.

We would want to try our best not to modify or deprecate things once they are out, and I am trying to keep super cautious when pushing the APIs forward.

Welcome any comments!

dstillman: > For 1), I think generally you won't want to change attributes. It can be weird that a field can become to readonly and suddenly turn back to be editable.

I can envision a plugin wanting to do something like this, but maybe it's unlikely enough that just creating a new row would be sufficient.

> For 2), that's why I made this PR - you can call the update and then this specific field will get the latest value from the onGetData.

Oh, I see — I was missing the `onGetData` part. I guess that's fine.

dstillman: Maybe "refresh" makes this clearer? To me, "update" sort of implies that you're changing something more directly, whereas "refresh" implies "do whatever you did originally to get the data". (Emiliano also phrased it in terms of a "refresh function".)

windingwind: Yeah I think refresh makes more sense.

PR #4854, opened 2024-11-19
Fix for loss of focus not closing the quick citation dialog

sparxooo:
When focus is lost from the Quick Citation dialog, it does not currently close. I'm unsure if this is expected behaviour, or whether it is unwanted. 

It appears definitely unwanted if you are editing multiple citations, and click out of the window, then select another citation. The Quick Citation dialog then reflects (and will update) the incorrect citation.

Steps to repro:

1) Open Word
2) Insert two citations (and press enter/accept each one to accept).
3) Select one citation, and press Add/Edit to open the Quick Citation dialog
4) Without pressing escape, or accepting the dialog, click back into word, and select the second citation
5) Notice the dialog is still reflecting the first citation

Screenshot 2024-11-19 192637

With a single line of code I believe I've fixed the issue, there appears to be no problem utilising the dialog from my end, but appreciate I cannot test on other platforms/with other word processors.

OS: Windows 11 23H2 22631.4317
Zotero: 7.0.10-beta.2+dc47650eb (64-bit) (also last stable version)

### DIFF
--- a/chrome/content/zotero/integration/quickFormat.xhtml
+++ b/chrome/content/zotero/integration/quickFormat.xhtml
@@ -42,6 +42,7 @@
 	persist="screenX screenY"
 	onkeypress="Zotero_QuickFormat.onWindowKeyPress(event)"
 	onunload="Zotero_QuickFormat.onUnload()"
+	onblur="Zotero_QuickFormat.onUnload()"
 	drawintitlebar-platforms="linux,win"
 	no-titlebar-icon="true"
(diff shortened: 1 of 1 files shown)


adomasven:
Thanks for this, but we cannot accept it, since it degrades the functionality for many of our users. I've responded to your forum post. We'll want to do something more complicated to improve the experience. https://github.com/zotero/zotero/issues/4855

PR #4859, opened 2024-11-21
Close citation dialog if user interacts with the integration plugin

### LINKED ISSUE #4855: Close citation dialog if user interacts with the integration plugin
Based on https://forums.zotero.org/discussion/119805/incorrect-behaviour-quick-citation-dialog-does-not-close-on-focus-loss

Currently we do not close the citation dialog when its focus is lost. This is reasonable as users might get interrupted while citing to looking something up, or to do any other task, and should be able to resume citing.

On the other hand this means that putting the cursor in another place in the document and pressing Add/Edit Citation means that the citation dialog "opens" for the wrong citation (or in worse instances, nothing happens if the OS doesn't allow Zotero to focus itself).

In those cases we should close the current citation dialog (or document preferences, edit bibliography dialogs) and continue with the new operation. In cases where any changes have been made in the dialog, we should instead pop up an alert saying asking the user whether they want to close the dialog and discard the changes.

We should also display an alert when any other word processor operation is running in the background (we often get questions from frustrated users refreshing long documents on macOS).

adomasven:
Asks the user if they want to discard the changes in their existing dialog if they have made changes.

If there is no citation dialog, but a command is running (like updating the document) an alert is displayed.

Closes #4855

### DIFF
--- a/chrome/content/zotero/bibliography.js
+++ b/chrome/content/zotero/bibliography.js
@@ -32,6 +32,8 @@
 // Class to provide options for bibliography
 // Used by integrationDocPrefs.xhtml and bibliography.xhtml
 
+window.isPristine = true;
+
 window.Zotero_File_Interface_Bibliography = new function () {
 	var _io;
 	
@@ -200,6 +202,11 @@ window.Zotero_File_Interface_Bibliography = new function () {
 		if (hasCheckedAdvancedOption) {
 			this.toggleAdvancedOptions(false);
 		}
+		
+		const checkboxes = document.querySelectorAll('.advanced-checkbox');
+		for (const checkbox of checkboxes) {
+			checkbox.addEventListener('command', _ => window.isPristine = false);
+		}
 	};
 	
 	this.openHelpLink = function () {
@@ -232,6 +239,7 @@ window.Zotero_File_Interface_Bibliography = new function () {
 	};
 
 	this.onDocPrefsWindowStyleChange = function (style) {
+		window.isPristine = false;
 		if (windowType !== "docPrefs") return;
 
 		let isNote = style.class == "note";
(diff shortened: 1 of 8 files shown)


adomasven:
I imagine there may be a better way to use fluent here, but given that the form or gender of "document preferences", "citation dialog" and "bibliography dialog" may be different in different languages and require different wording, I split them into different strings.

dstillman:
I'm getting a communication error from Word while testing this:

1. Click Add/Edit Citation and make a selection, but don't close the dialog.
2. Return to Word and move the cursor elsewhere in the document.
3. Click Add/Edit Citation again.

I get the warning about changes, and then in a second or two Word starts bouncing in my dock (because of the communication-error alert). If I click Cancel — don't discard changes — and then press Return to accept my changes, the progress dialog just hangs, because Word is stuck in the background with the communication-error alert.

adomasven:
Fixed. Mac Word HTTP communication has some caveats that other plugins don't which is why this was failing. Should be fine on all platforms now.

adomasven:
@dstillman https://forums.zotero.org/discussion/121455/word-plug-in-wont-work

dstillman:
I'm still seeing the above behavior on the latest commit.

dstillman:
(But at this point we should just build this into the new citation dialog and not bother with quick-format.)

adomasven:
> (But at this point we should just build this into the new citation dialog and not bother with quick-format.)

The logic here is the same for all dialogs, just going to need to add some `isPristine` flags for the new dialog before merging it. We should still merge this PR for its core logic.

adomasven:
And pushed with the fix for mac, with a rebase again. Not sure why the commit didn't come through the first time.

dstillman:
The message isn't always technically accurate for doc prefs. It warns me if doc prefs is just open, even if I didn't change anything. Can we track changes there? Doesn't even need to make sure that the selected settings are actually different, just that they were changed at one point. So if you change to a different style and back without accepting the dialog, it seems fine to show the warning.

adomasven:
Fixed

Issue #4871, opened 2024-11-27
Advanced search "Any field" empty in "match all" mode


abaevbog:
https://forums.zotero.org/discussion/120016/saved-advanced-searches-using-everything-mode

If I change to "match any", the results do appear.

(Seemingly after https://github.com/zotero/zotero/pull/4845)

PR #4873, opened 2024-11-27
fix advanced search anyField condition breakage

abaevbog:
- ensure that if the "joinMode" condition is present, it is processed first, since subsequent conditions may rely on it
- ensure that joinMode is initialized by default to "ALL", in case the condition is not explicitly
provided

Fixes: #4871


dstillman:
I simplified this to just extract `joinMode` in one place and set it as a property.

PR #4876, opened 2024-12-01
File renaming: Add pascal case. Fix unicode handling.

tnajdek:
This PR adds support for the "pascal" text case option in the file renaming template. It's based on this request: https://forums.zotero.org/discussion/comment/480520.

I've also noticed that our "camel" case (and the new "pascal" case) does not preserve Unicode characters, so I've fixed that.

### DIFF
--- a/chrome/content/zotero/xpcom/attachments.js
+++ b/chrome/content/zotero/xpcom/attachments.js
@@ -2348,7 +2348,11 @@ Zotero.Attachments = new function () {
 					value = value.toLowerCase().replace(/\s+/g, '_');
 					break;
 				case 'camel':
-					value = value.toLowerCase().replace(/[^a-zA-Z0-9]+(.)/g, (m, chr) => chr.toUpperCase());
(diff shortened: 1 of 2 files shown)


dstillman:
I actually kind of doubt there's any reason for non-Pascal camel case here. There's not much harm in supporting both, but the alternative would be just changing "camel" to be Pascal. It's really hard to imagine what the use case for lower camel case would be…

tnajdek:
> I actually kind of doubt there's any reason for non-Pascal camel case here.

I can see that now; perhaps we should have implemented upper camel case in the first place. I don’t know how common it is to differentiate between lower and upper camel case - I’ve always referred to these as "camel" and "pascal" cases. At this point, it’s probably best to just support both.

If you prefer, I can tweak the PR so that "camel" case is uppercase. Either way, I’ll update the docs once this is merged.

tnajdek:
https://forums.zotero.org/discussion/comment/493149/#Comment_493149

Issue #4883, opened 2024-12-03
fx128: Need to turn on svg.context-properties.content.enabled


windingwind:
We are using `-moz-context-properties: fill, stroke` everywhere, but it's by default turned off. Need to turn that on.

https://devdoc.net/web/developer.mozilla.org/en-US/docs/Web/CSS/-moz-context-properties.html

current 115:

<img width="693" alt="image" src="https://github.com/user-attachments/assets/be728b2e-b912-4eae-bc3f-500d47f6aea7">

128:

<img width="684" alt="image" src="https://github.com/user-attachments/assets/3cd35b25-dfea-4ccc-a011-f73d71abb435">

After turning on:

<img width="684" alt="image" src="https://github.com/user-attachments/assets/a3202807-94d7-49b7-9d5d-6b05c41f6eb2">

dstillman:
Not a big deal to change this, but what changed in fx128 here? Are we relying on this for non-chrome/resource URLs?

dstillman:
I guess these plugin icons would be `jar:file:///` URLs. Why did it work before?

windingwind:
No idea why it worked before...

dstillman:
Fixed by #4884

PR #4884, opened 2024-12-03
Fx128: Enable -moz-context-properties for SVG context properties

windingwind:
fix: #4883


dstillman:
Not a big deal to change this, but what changed in fx128 here? Are we relying on this for content (rather than chrome)?

PR #4897, opened 2024-12-04
Item pane scroll pos fix after switching from note

windingwind:
Follow up to #4879
fix: #4276

### DIFF
--- a/chrome/content/zotero/elements/itemPane.js
+++ b/chrome/content/zotero/elements/itemPane.js
@@ -151,8 +151,17 @@
 			return true;
 		}
 
-		renderItemPane(item) {
+		async renderItemPane(item) {
+			let previousMode = this.mode;
 			this.mode = "item";
+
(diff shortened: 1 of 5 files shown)


chrome/content/zotero/elements/itemPane.js
```diff
@@ -151,8 +151,17 @@
...
 
-		renderItemPane(item) {
+		async renderItemPane(item) {
+			let previousMode = this.mode;
 			this.mode = "item";
+
+			// Fix https://forums.zotero.org/discussion/115450/zotero-7-beta-wrong-vertical-position-in-the-item-pane-after-switching-from-a-note
+			if (previousMode === "note") {
+				// Wait for DOM to update and then trigger item-details render
+				await new Promise((resolve) => {
```
windingwind, line 161: Moved this await DOM update from `item-details` to `item-pane` before the sections are rendered and scrolled, to fix https://forums.zotero.org/discussion/120216/blinking-item-pane-when-switching-item-with-pinned-section-in-the-sidenav

and 

https://forums.zotero.org/discussion/120217/title-field-is-not-visible-anymore-after-creating-a-new-item-if-a-section-is-pinned-in-the-sidenav

PR #4901, opened 2024-12-04
fx128: Add `remove_between`, fully fix input styles

AbeJellinek:
`global-shared.css` ended up broken after #4890. Turns out multiline Perl regex substitutions do not work the way I thought they did! Every closing brace in the file was getting removed. That was why inputs still looked wrong. So I've added a new `remove_between` function that removes all lines between (and including) a start pattern and an end pattern, non-greedily, and used that to apply the intended fix.

### DIFF
--- a/app/scripts/fetch_xulrunner
+++ b/app/scripts/fetch_xulrunner
@@ -95,6 +95,22 @@ function remove_line {
 	fi
 }
 
+function remove_between {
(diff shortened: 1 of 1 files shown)


dstillman:
(Sorry, I was skeptical of that regex but didn't actually check the output.)

PR #4903, opened 2024-12-05
Server: Expose actual bound port

AbeJellinek:
And:
- Use a different port for tests so server tests don't break when Zotero is already running
- Remove no-op `httpServer.enabled` pref updates in tests
   - It's already enabled, no test disables it, and any test that did would clean up after itself
   - Updating that pref has no effect without a separate call to `Zotero.Server.init()`
- Remove unused arg to `Zotero.Server.init()`

It's basically just annoying that server tests fail when I forget to close the client first, so this fixes that.

### DIFF
--- a/test/runtests.sh
+++ b/test/runtests.sh
@@ -137,6 +137,7 @@ user_pref("extensions.zotero.firstRunGuidance", false);
 user_pref("extensions.zotero.firstRun2", false);
 user_pref("extensions.zotero.reportTranslationFailure", false);
 user_pref("extensions.zotero.httpServer.enabled", true);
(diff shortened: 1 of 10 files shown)


test/runtests.sh
```diff
@@ -137,6 +137,7 @@ user_pref("extensions.zotero.firstRunGuidance", false);
 user_pref("extensions.zotero.firstRun2", false);
 user_pref("extensions.zotero.reportTranslationFailure", false);
 user_pref("extensions.zotero.httpServer.enabled", true);
+user_pref("extensions.zotero.httpServer.port", 23124);	// ascii "ZT"
```
dstillman, line 140: I never understood this. How is this "ZT" (and how is the original 23119 "ZO")?

PR #4923, opened 2024-12-11
Remove missing-component integration error.

### LINKED ISSUE #4916: Remove missing-component integration error
I forgot to update the Mac Word submodule on my fx128 branch and got this:

Screenshot 2024-12-10 at 3 09 58 AM

That shouldn't happen, of course, but we shouldn't have this error anymore regardless, since the advice it gives is obsolete. We should either update this with a relevant message (though I'm not sure what that would be) or just assume this won't happen on real builds and throw the error to the console for developers with messed-up submodules.

adomasven:
This is no longer a thing since word plugins are not Zotero extensions. If that code throws - it's because we've done something wrong and it's a bug.

I removed the associated string, but I'm not sure if we can do that and things will propagate properly to Transifex.

Closes #4916.

### DIFF
--- a/chrome/content/zotero/xpcom/integration.js
+++ b/chrome/content/zotero/xpcom/integration.js
@@ -230,19 +230,12 @@ Zotero.Integration = new function() {
 		if (agent == 'http') {
 			return new Zotero.HTTPIntegrationClient.Application();
 		}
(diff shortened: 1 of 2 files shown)


dstillman:
Yeah, removing the en-US string is sufficient.

Issue #4885, opened 2024-12-03
fx128: Hide `Run on sites with restrictions` in plugins page


windingwind:
<img width="1000" alt="image" src="https://github.com/user-attachments/assets/5060e374-495b-4e90-ba0b-2aaa3ef108ca">

PR #4925, opened 2024-12-13
fx128: hide 'Run on sites with restrictions' from plugins page

abaevbog:
Fixes: #4885

After re-fetching xulrunner:

Screenshot 2024-12-12 at 4 28 54 PM

### DIFF
--- a/app/scripts/fetch_xulrunner
+++ b/app/scripts/fetch_xulrunner
@@ -326,9 +326,9 @@ function modify_omni {
 	echo '.main-heading { margin-top: 2em; }' >> $file
 	# Hide Details/Permissions tabs in addon details so we only show details
 	echo 'addon-details > button-group { display: none !important; }' >> $file
-	# Hide 'Run on sites with restrictions' row and helper message
-	echo '.addon-detail-row-quarantined-domains { display: none !important; }' >> $file
-	echo '.addon-detail-help-row:has([data-l10n-id="addon-detail-quarantined-domains-help"]) { display: none !important; }' >> $file
+	# Hide "Run on sites with restrictions" row and helper message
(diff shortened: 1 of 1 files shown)


app/scripts/fetch_xulrunner
```diff
@@ -326,6 +326,9 @@ function modify_omni {
 	echo '.main-heading { margin-top: 2em; }' >> $file
 	# Hide Details/Permissions tabs in addon details so we only show details
 	echo 'addon-details > button-group { display: none !important; }' >> $file
+	# Hide 'Run on sites with restrictions' row and helper message
+	echo '.addon-detail-row-quarantined-domains { display: none !important; }' >> $file
+	echo '.addon-detail-help-row:has([data-l10n-id="addon-detail-quarantined-domains-help"]) { display: none !important; }' >> $file
```
dstillman, line 331: No reason for `!important` here, right?

app/scripts/fetch_xulrunner
```diff
@@ -326,6 +326,9 @@ function modify_omni {
 	echo '.main-heading { margin-top: 2em; }' >> $file
 	# Hide Details/Permissions tabs in addon details so we only show details
 	echo 'addon-details > button-group { display: none !important; }' >> $file
+	# Hide 'Run on sites with restrictions' row and helper message
```
dstillman, line 329: Double quotes

app/scripts/fetch_xulrunner
abaevbog: Right, it works just fine without it. I initially thought to add it just for good measure, in case something ever changes in the stylesheet in the future. But yeah, it's not required so I removed it.

PR #4936, opened 2024-12-18
Check editability based on tab when adding attachments

### LINKED ISSUE #4934: Add button in Attachments section of reader item pane depends on collection editability
https://forums.zotero.org/discussion/120597/zotero7-0-11-abnormal-alert-when-click-the-add-file-button-in-the-sidenav-pane

AbeJellinek:
Fixes #4934

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -1934,7 +1934,7 @@ var ZoteroPane = new function()
 			
 			// Tab selection observer in standalone.js makes sure that
 			// updateQuickCopyCommands is called
-			if (Zotero_Tabs.selectedID == 'zotero-pane') {
+			if (Zotero_Tabs.selectedType == 'library') {
 				this.updateQuickCopyCommands(selectedItems);
 			}
 			
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -4491,13 +4491,8 @@ var ZoteroPane = new function()
 	
 	
 	this.addAttachmentFromURI = Zotero.Promise.method(function (link, itemID) {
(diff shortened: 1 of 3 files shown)


chrome/content/zotero/zoteroPane.js
```diff
@@ -4491,7 +4491,8 @@ var ZoteroPane = new function()
 	
 	
 	this.addAttachmentFromURI = Zotero.Promise.method(function (link, itemID) {
-		if (!this.canEdit()) {
+		if (Zotero_Tabs.selectedID === 'zotero-pane' && !this.canEdit()
+				|| !Zotero.Items.get(itemID).library.editable) {
```
dstillman, line 4495: It seems kind of confusing to leave `canEdit()` and `canEditFiles()` checking just the tree but have other logic in this same file check for other situations. Shouldn't we just have those functions check editability based on the current context?

chrome/content/zotero/zoteroPane.js
```diff
@@ -4514,35 +4520,39 @@ var ZoteroPane = new function()
...
+				}
+				else if (collectionTreeRow.isPublications()) {
+					Zotero.alert(
+						null,
+						Zotero.getString('general.error'),
+						Zotero.getString('publications.error.linkedFilesCannotBeAdded')
+					);
+					return null;
+				}
+			}
```
dstillman, line 4539: This is here because there's no way through the UI to add a linked file from a reader tab?

It makes sense to check `isPublications()` here, since that by definition requires a collection tree row, but we should probably move the linked-file check down and use `Zotero.Libraries.get(libraryID).isGroup` instead of `isWithinGroup()`, in case some plugin triggers this function from the reader or it gets exposed in some other way.

PR #4946, opened 2024-12-20
Ensure no child rows in regularOnly itemTree

### LINKED ISSUE #4941: ItemTree displays child rows even with regularOnly=true on right arrow
1. Add `this.regularOnly` to the constructor of `itemTree`.
2. Note that no child rows are now displayed in the itemTree now.
3. Select multiple rows that have child notes or attachments
4. Press `ArrowRight`
5. Note that even though there is no grippy, child attachments and notes still appear.
6. `ArrowLeft` from here will collapse the rows

This is not an issue now, since `regularOnly` is not used by itemTree anywhere. But I ran into it while working on https://github.com/zotero/zotero/pull/4872, so it needs to be resolved before that.

Initially brought up in citation explorer PR: https://github.com/zotero/zotero/pull/3468#issuecomment-2542523224

abaevbog:
Needed to avoid glitches in citation explorer and citation dialog.

Fixes: #4941

Discussed in https://github.com/zotero/zotero/pull/3468#issuecomment-2551945060

### DIFF
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -1640,6 +1640,7 @@ var ItemTree = class ItemTree extends LibraryTree {
 		if (!this.isContainer(index)) {
 			return;
 		}
+		if (this.regularOnly) return;
 
 		this._lastToggleOpenStateIndex = index;
 


dstillman:
We already have a `this.regularOnly` check in `isContainerEmpty()`. Shouldn't `toggleOpenState()` just call `isContainerEmpty()` instead of `isContainer()`?

(Or it might be more proper to call `isContainer()` _and_ `isContainerEmpty()`, but in practice `isContainerEmpty()` includes the same check as `isContainer()`, so there's no real point in doing that.)

abaevbog:
That is a good point. I am feeling somewhat concerned about changing the `isContainer` for `isContainerEmpty` ... The comment indicates that `isContainer` is a check for a somewhat special edge case (that I have not been able to run into) during drag-drop.
https://github.com/zotero/zotero/blob/dcf809961a716938a301574f37759641d36998c7/chrome/content/zotero/itemTree.jsx#L1637-L1642

And the return values for `isContainerEmpty` may differ for different item types. For instance, for a standalone note or attachment row `isContainer` is `false` but `isContainerEmpty` is `false`. So, especially since that edge case described in the comment is not easily reproducible, maybe it does make sense to check both `isContainer` and `isContainerEmpty`?

Another thought I have is that when `regularOnly` is true, don't all rows become non-containers? So would it make sense to add `if (this.regularOnly) return false;` check into `isContainer` instead?

Issue #4840, opened 2024-11-14
Improve tabs list popup


windingwind:
- Add keyboard shortcuts for opening the popup
- Always allow input to search
- Press enter to jump to the selected tab directly
- Always have a tab row selected

PR #4961, opened 2024-12-27
Improve tabs menu popup

windingwind:
- [x] add keybord shortcut (by default, ctrl+shift+;)
- [x] always can input
- [x] always have virtual selection
- [x] always can enter to confirm
- [x] improve tab key navigation (input > current row > close button)
- [x] a11y with description for interaction (test on voiceover)

resolve: #4840


windingwind:
@abaevbog It took me some time to figure out how to implement a11y support with virtual focus, not sure if I am doing it right. It would be perfect if you would have a look and maybe test it with other screen reader tools (I only tested it on voiceover). Thanks!

windingwind:
Also @yexingsha maybe we need some design here for the tabs menu with virtual selection. Here's what it looks like now:

image

image

Thank you very much!

abaevbog:
@windingwind this is great! Besides voiceover, I checked this with NVDA and JAWS, and the tab names all get announced as expected after searching. JAWS tends to be somewhat picky, so I am quite surprised that it just worked!

Just a few additional thoughts and suggestions:

1. We want to add `role=button` attribute to the `X` button, so that screen readers know what it is. Currently, it is announced as default "section". We could alternatively make it a proper `<button>` component but if I remember it got a bit messy with the stylesheet.
2. Since the entire row gets selected, the title of the `X` button "Close tab" is read after the tab name. It almost sounds like the entire tab row entry will close a tab (at least in English). It is clarified by a later announcement via `aria-description` but still might be a bit confusing. This happens because the parent row does not have `aria-label`, so screen readers read out everything inside of it. Perhaps we only want to have the tab name announced when a row is selected? If we add `aria-label` to the row only with the tab name, the "Close tab" should not be read out.
3. Now that Enter will open the selected row when input is focused, is there a need to have first `Tab` from the input focus the tab name? Currently, I type a query, hear a tab name, press Tab and hear the same tab name again. And Enter does the same thing when tab name is focused as right after typing when focus is in the input. Unless there is a specific need for it that I am missing, maybe we just skip the tab name and focus the `X` button right away?

yexingsha:
I'm wondering if it makes more sense for virtual selection to not target the entire row. This can also solve the issues of screen readers reading out "Close tab", the first `Tab` not doing anything, and the awkward focus ring overlaid on blue background when the `X` button is focused. The focus states would look like this:

tabs-menu-light
tabs-menu-dark

- The main change here is to separate the focus states of tab names and the `X` button, just as we do with hover states.
- Use `white` item type icons when focused.
- Also, remove hover/press effects from focused items.

If we are to adopt this approach, then I expect `ArrowRight` would move focus from tab name to `X`, and another `ArrowRight` would move focus to the next tab name; and `ArrowDown`/`ArrowUp` would navigate between different `X`s, which is how it currently works.

windingwind:
Thanks @abaevbog , I will improve it once we decide everything about the design.

re @yexingsha :

This seems to be putting more weight on the close button. Most of the time you would not want to close the tab via the popup, because:

1. if you want to close all (other) tabs, usually you use the tab context menu-> close all tabs
2. If you want to close a tab that’s visible in the tab bar, usually you use the close button on the tab itself 
3. Only when you want to close a specific tab that’s not shown in the tab bar, you may want to use this popup. But how do you know there is a tab you want to close? Since tabs will be automatically unloaded, this is either unnecessary or weird. To be honest I don’t know if this case exists.

The only reason I can think of to use this close button is 1) you want to close the current tab before going to the next tab; 2) you find some tabs that you think they can go away when navigating to the next tab (this isn’t a thing for screen reader users)

Thus, currently I am hiding all close button except for the hovered and selected row.
 For the same reason, I think it’s not necessary to provide the close buttons extra arrow key navigation support (left-right arrow between the row and close button might be useful arguably, but we also have a tab key navigation)

yexingsha:
I actually use the tabs menu to close tabs quite a lot, since I often accumulate open tabs until I feel like I need to close some of them, and it's easier to use the tabs menu instead of scrolling through the crowded tab bar. But I take your point that we should not put equal weight on the tab name and the close button.

If we want virtual focus to target the entire row, then we should give hover states the same treatment. Here is how I think the menu should look like, when one of the rows is virtually selected and the close button focused, and the cursor is hovering over the close button in another row:

tabs-menu-light
tabs-menu-dark

- In inactive rows, since close buttons are hidden, tab names should fill the entire row.
- Close buttons should be nested in their rows, like they are in tabs.
- The current hover/active colors are inconsistent with other UI elements, they should be `fill-quinary` and `fill-quarternary` respectively.
- For close buttons in virtually focused rows, use `#ffffff1a` for hover bg color, and `#ffffff33` for active.
- Use windows style focus ring for close buttons.

windingwind:
Thank you @yexingsha ! It looks like this now:

image

yexingsha:
Looks great! Only issues left are:

- In virtually focused rows, use `white` item type icons, and change the color of library/collection icon to `accent-white`.
- For close buttons in virtually focused rows, use `#ffffff1a` for hover bg color, and `#ffffff33` for active.

windingwind:
Thanks @yexingsha , should be fine now:

image

abaevbog:
@windingwind great, it all sounds good now!

As far as the last point, I didn't actually realize that a hovered `X` is also focusable on tab.  I think that it is a very edge case that we shouldn't worry about too much, as I would assume most users relying on tab for navigation are not going to to also be hovering around tab rows with their mouse.

If we were to make the entire row focusable and in that instance do input -> close button -> selected row -> close button, would we also focus the entire row on the first tab from the input, when no hovered `X` button is visible? So input -> selected row -> close button? If so, wouldn't it functionally be identical to the current setup, where you would have to tab twice to get to the `X`, and screen readers would end up announcing the title twice?

What would be an issue with input -> close button -> close button? This seems to be what VScode does in the search bar, for example. (Or at least what it is meant to do, as the focus gets kinda lost for me on the first tab before landing on one of the row's buttons on the second tab).

Finally, I see Firefox search bar has some similar functionality to what we are after here. If I type a query, there is a "Firefox suggests" section with recent matches that I can also navigate with arrows from the input. Those matches have a `...` icon that appears when the row is selected or hovered. In Firefox, though, Tab keypress will only focus the `...` icon of the currently selected `X` icon, while the `...` of the hovered row is ignored. Could we consider that? Is there a reason why we keep hovered row's `X` in taborder (apologies if this was covered earlier and I missed it)?

I don't think I have a very strong opinion on which way is better, so these are just some considerations I could think of to avoid having an extra tabstop. Let me know what your thoughts are on these!

windingwind:
Thank you @abaevbog, I think your points make sense. I'm following your suggestion and 1) ignore the hovered close button; 2) skip the focus of the row title on first tabbing after each selection change.

Do you think that sounds good? Thank you very much.

windingwind:
Also the drag-drop is disabled when the filter is active, as it doesn't make any sense to rearrange the tabs when some tabs are hidden by the filter (you never know where the tabs should be inserted)

abaevbog:
Great! Yeah, I think not including the hovered row's `X` into tab navigation feels more straightforward (and consistent with Firefox too).

My only remaining question is why is title skipped only on the first Tab, and not on the subsequent tabs? It leads to some odd behaviors. For example, if I open the tabs panel when the library tab is open, and press Tab once, focus will remain in the input. But if I press Tab again, then it moves on the title. It's somewhat odd? Why not just make the tab title always un-focusable and only focus the `X` always? 


> (Or at least what it is meant to do, as the focus gets kinda lost for me on the first tab before landing on one of the row's buttons on the second tab).

If this was due to that comment of mine, I only mentioned it because it looks like there is a bug in VSCode where the focus gets lost on the first Tab before landing on the `X`, in case you were to try it in VSCode and see that too. Meaning that it should (probably) not be happening and the focus should just go from the input to an `X` there. Apologies if I didn't phrase it right!

windingwind:
Oh, sorry maybe I misunderstood. It now only focuses the close button of the selected row. It is indeed much clearer than before. Thank you for the comment!

windingwind:
@dstillman We could try to merge this, since it is a blocker to the menus API, or at least a cause of conflicts. This PR decouples tabs menu popup from the Zotero_Tabs, and the menus API will support adding custom menus to the tabs contextmenu in Zotero_Tabs. Thank you very much.

dstillman:
Are the padding changes intentional?

Before:

Screenshot 2025-04-16 at 2 04 27 AM

Now:

Screenshot 2025-04-16 at 2 04 24 AM

We don't need the focus ring — we could experiment with running the input straight to the edge on the top and sides — but if we're going to have

dstillman:
Should we just do Cmd/Ctrl-; for the shortcut? Our Shift shortcuts like Cmd-Shift-K mostly date to when Zotero was a Firefox extension and we couldn't use simpler ones, but we now have, e.g., Cmd-F and could probably have Cmd-L instead of Cmd-Shift-L.

I think the main reason not to do that would be if we were saving Cmd-; for something else. E.g., in my editor, I have Cmd-; bound to a fast-open thing that searches all files in the project, while Cmd-Shift-; opens the tabs menu. But we could also change this later if we wanted Cmd-; for something else.

windingwind:
Thanks for the comments.

> Should we just do Cmd/Ctrl-; for the shortcut? Our Shift shortcuts like Cmd-Shift-K mostly date to when Zotero was a Firefox extension and we couldn't use simpler ones, but we now have, e.g., Cmd-F and could probably have Cmd-L instead of Cmd-Shift-L.

Our customizable shortcuts system seems to be only using ctrl/cmd - shift - key. If we don’t care about customizable shortcuts for the panel, then without shift makes more sense for me.

For other things I will check.

dstillman:
In the Advanced pane of the prefs, you mean? We haven't touched that in many years, and we can certainly update it at some point to handle other formats — no need to stick to Shift for that. (Mac users can also customize any menu item through system settings.)

windingwind:
Yes - that’s also what I wanted to do.

We can for now simply use ctrl/cmd - ;. I can take the shortcut customization part later.

windingwind:
Fixed, thanks.

The current popup with larger padding:

image

chrome/locale/en-US/zotero/preferences.dtd
```diff
@@ -133,6 +133,7 @@
 <!ENTITY zotero.preferences.keys.newItem					"Create a New Item">
 <!ENTITY zotero.preferences.keys.newNote					"Create a New Note">
 <!ENTITY zotero.preferences.keys.toggleAllRead				"Mark All Feed Items As Read/Unread">
+<!ENTITY zotero.preferences.keys.toggleTabsMenu				"Toggle Tabs Menu">
```
dstillman, line 136: No more .dtd or .properties strings!

If we want to add this to the prefs, we'll need to move these strings to preferences.ftl (or special-case this and the new zotero.properties ones for now).

Also, while this does hide the menu too, "Toggle" doesn't seem right for a menu. I think we can just say "Show Tabs Menu".

chrome/locale/en-US/zotero/zotero.ftl
```diff
@@ -721,3 +721,28 @@ mac-word-plugin-install-dont-ask-again-button =
     .label = { general-dont-ask-again }
 
 connector-version-warning = The { -app-name } Connector must be updated to work with this version of { -app-name }.
+
+tabs-menu-row =
+    .aria-description = You are currently on tab { $index } of { $total }.
+    To search for tabs, type the keyword. To jump to the currently selected tab, press Enter.
```
dstillman, line 727: - Has to be `{ return-or-enter }` to account for macOS
- "keyword" seems a little unclear. "To search for a tab, type part of its name."? (Below too.)

chrome/locale/en-US/zotero/zotero.ftl
```diff
@@ -721,3 +721,28 @@ mac-word-plugin-install-dont-ask-again-button =
...
+            *[other] previous
+    }
+    tab, press Arrow Up.
+    To navigate to the 
+    {
+        $isLast ->
+            [true] first
+            *[other] next
+    }
+    tab, press Arrow Down.
```
dstillman, line 741: "Arrow Down" → "Down Arrow"

windingwind:
Thanks. Should all be addressed.

chrome/locale/en-US/zotero/zotero.ftl
```diff
@@ -13,6 +13,11 @@ command-or-control =
         [macos] { general-key-command }
         *[other] { general-key-control }
     }
+command-or-control-shift =
+    { PLATFORM() ->
+        [macos] { general-key-command } + { general-key-shift }
```
dstillman, line 18: Mac shortcuts are given using hyphens (or en dashes if there's a multi-word key) and no spaces:

https://support.apple.com/en-us/102650

Not sure how it will look without spaces with the input fields, but let's see.

(Will need to adjust `preferences-advanced-keys-modifier` too.)

chrome/locale/en-US/zotero/preferences.ftl
```diff
@@ -99,3 +99,30 @@ preferences-advanced-custom-data-dir =
...
+    .value = Mark All Feed Items As Read/Unread
+preferences-advanced-keys-showTabsMenu =
+    .value = Show Tabs Menu
+preferences-advanced-keys-changesTakeEffect =
+    .value = Changes take effect after restart
+preferences-advanced-keys-modifier =
+    .value = { $modifierType ->
+        [ctrl] { command-or-control } + 
+        *[ctrl-shift] { command-or-control-shift } + 
+    }
```
dstillman, line 128: Using `ctrl`/`ctrl-shift` is a little awkward here (and as a Mac user I object on principle). `$modifier` and `with-shift`/`without-shift`?

dstillman (commented):
Screenshot 2025-05-03 at 1 54 30 AM

Can we right-align these?

chrome/locale/en-US/zotero/zotero.ftl
windingwind: Thanks, fixed.

We now have a `shortcut-separator` in zotero.ftl.

MacOS:

image

Win:

image

windingwind:
Thanks. Should all be addressed

dstillman:
Firefox has a Show All Tabs option in the View menu to open the tabs menu. We should probably do something similar. No one's going to discover this otherwise.

And after the trouble of updating the prefs section, I wonder if we should even bother adding those new strings. We're going to want to make keyboard shortcuts fully customizable, with the modifier keys included in the input field — at which point we'll use abbreviations or symbols.

windingwind:
I agree. If we plan to do the shortcut soon, I guess it's better to do it in a separate PR. I can revert this part and maybe create an issue if we didn't have one (for customizing shortcut prefs)

The menu sounds like a good idea.

windingwind:
Done:

image

dstillman:
At least for me, setting the key's `key` attribute dynamically causes the menu not to show the shortcut key on first start. I have to switch to another tab and back before it shows. I've added a back-and-forth toggle of `hidden` to force it to update.

dstillman:
Do we like "Show All Tabs" or "Show Tabs Menu" for the menu item?

windingwind:
> At least for me, setting the key's key attribute dynamically causes the menu not to show the shortcut key on first start.

oh indeed. Same for me on Windows (but not MacOS? Weird). Nice catch!

> Do we like "Show All Tabs"

I vote for Show Tabs Menu personally

dstillman:
Screenshot 2025-05-07 at 1 33 56 AM

dstillman:
A little unfortunate that there are two menu items beginning with "Show" that work totally differently — one a checkbox setting and the other a command — but I'm not sure there's anything we can do about that.

dstillman:
Merged in 61b62d0

Issue #4932, opened 2024-12-17
Remove aria-autocomplete attribute from search-textbox


abaevbog:
https://forums.zotero.org/discussion/120552/quick-search-doesnt-actually-have-popups

`search-textbox` has `aria-autocomplete` set by mozilla [here](https://searchfox.org/mozilla-esr128/source/toolkit/content/widgets/search-textbox.js#136), which ends up being not appropriate for our use cases. It tells screen readers that there should be an autocomplete popup, similar to when you edit tags, that can be navigated with arrow keys. Different screen readers announce it differently. For example, JAWS says that there is a popup, and NVDA says that there is autocomplete. In any case, it needs to be removed from all instances of `search-textbox` to avoid confusion. 

Another point in the linked post is that there is no audible announcement when itemTree results are updated after typing in quickSearch. That is something that would be great to add, though currently `aria-live` regions (which is the most reasonable way to do it) do not reliably work in chrome windows.

PR #4966, opened 2025-01-03
remove wrong aria-autocomplete from search-textbox

abaevbog:
[searchButton setter](https://searchfox.org/mozilla-esr128/source/toolkit/content/widgets/search-textbox.js#129) in `search-textbox` element adds `aria-autocomplete` to the `<input>`. It makes screen readers announce that there is an autocomplete popup and is misleading. This overrides the `searchButton` setter to remain unchanged but with the line where `aria-autocomplete` is set commented out.

Addresses: #4932

For comparison, this is the JAWS announcement before (note the incorrect "Has popup"):

https://github.com/user-attachments/assets/e1ebb692-be34-43ed-b036-297f3ec1d78b

This is the JAWS announcement after this change (note no "Has popup"):

https://github.com/user-attachments/assets/ec8d268c-d83b-4c2b-bfd6-592d62608a5a

### DIFF
--- a/app/scripts/fetch_xulrunner
+++ b/app/scripts/fetch_xulrunner
@@ -416,6 +416,9 @@ function modify_omni {
        }
        popupset.appendChild(this._autoScrollPopup);' chrome/toolkit/content/global/elements/browser-custom-element.js
     
+	# Remove aria-autocomplete unnecessarily added to search-textbox to not confuse screen readers
+    remove_line 'this.inputField.setAttribute\("aria-autocomplete' chrome/toolkit/content/global/elements/search-textbox.js
+
     # Remove non-native text input styles
     remove_between 'html\|input\:where\(' '^}' chrome/toolkit/skin/classic/global/global-shared.css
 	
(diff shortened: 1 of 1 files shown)


dstillman:
Instead of this, which could break silently if Mozilla makes some change to the component, can we just disable it in the source with `remove_line` in `fetch_xulrunner`?

```bash
remove_line 'this.inputField.setAttribute\("aria-autocomplete' chrome/toolkit/content/global/elements/search-textbox.js
```

(I guess there's nowhere in Zotero where a native search textbox is used that we want autocomplete, right?)

abaevbog:
> Instead of this, which could break silently if Mozilla makes some change to the component, can we just disable it in the source with remove_line in fetch_xulrunner?

Oh yeah, that is definitely cleaner! I re-pushed that initial commit to just remove that one line.

> (I guess there's nowhere in Zotero where a native search textbox is used that we want autocomplete, right?)

Right. We only use `<search-textbox>` directly in collections search, notes search in the context pane and preferences. In all other cases, we use `<quick-search-textbox>` where we currently don't expect for autocomplete, and where it does not really make sense I think. Firefox use case is, for example, Bookmarks > Search Bookmarks. There, you can type in the `<search-textbox>` but then also use arrows to navigate the search results. And as you navigate them, the value of the input changes for what has been selected. I don't think we have anything like type of autocomplete that at the moment.

Issue #4975, opened 2025-01-07
fx128: Remove "Report" option for plugins


dstillman:
https://forums.zotero.org/discussion/120959/report-plugin

PR #4976, opened 2025-01-09
hide "Report" button from plugins context menu

abaevbog:
Fixes: #4975

Screenshot 2025-01-09 at 10 20 48 AM


app/scripts/fetch_xulrunner
```diff
@@ -284,6 +284,9 @@ function modify_omni {
 	# Hide "Run on sites with restrictions" row and helper message
 	echo '.addon-detail-row-quarantined-domains { display: none; }' >> $file
 	echo '.addon-detail-help-row:has([data-l10n-id="addon-detail-quarantined-domains-help"]) { display: none; }' >> $file
+	# Hide "Report" button from plugins context menu
+	echo '[data-l10n-id="report-addon-button"] { display: none }' >> $file
+	echo 'addon-card addon-options hr:first-of-type { display: none }' >> $file
```
dstillman, line 289: The `hr` is above the button, right? If so, should move that line above the other.

abaevbog: Good point, thank you. I re-pushed that commit with the lines swapped

PR #4979, opened 2025-01-10
Fix dense call to setItemDownloadPercentage freezes app

windingwind:


dstillman:
This should be `throttle`, not `debounce`. With `debounce`, it just never gets called until the download finishes.

PR #4983, opened 2025-01-15
Unset Mozilla environment variables after launch

### LINKED ISSUE #4981: Unset Mozilla environment variables after launch to prevent launching incorrect Firefox profile
https://forums.zotero.org/discussion/comment/472732/#Comment_472732

dstillman:
Fixes #4981

### DIFF
--- a/chrome/content/zotero/xpcom/fileHandlers.js
+++ b/chrome/content/zotero/xpcom/fileHandlers.js
@@ -149,12 +149,16 @@ Zotero.FileHandlers = {
 		handler = handler || systemHandler;
 		if (handler) {
 			if (Zotero.isMac) {
+				Zotero.Utilities.Internal.Environment.clearMozillaVariables();
 				try {
 					await Zotero.Utilities.Internal.exec('/usr/bin/open', ['-a', handler, path]);
 					return true;
 				}
 				catch (e) {
 					Zotero.logError(e);
+				}
+				finally {
+					Zotero.Utilities.Internal.Environment.restoreMozillaVariables();
 				}
 			}
 			
@@ -498,8 +502,13 @@ Zotero.FileHandlers = {
 		if (!Zotero.File.pathToFile(command).isExecutable()) {
 			throw new Error(`${command} is not an executable`);
 		}
+		
+		Zotero.Utilities.Internal.Environment.clearMozillaVariables();
+		
 		// Do not await
-		Zotero.Utilities.Internal.exec(command, args);
+		var promise = Zotero.Utilities.Internal.exec(command, args);
+		
+		promise.finally(() => Zotero.Utilities.Internal.Environment.restoreMozillaVariables());
 	},
 };
 
--- a/chrome/content/zotero/xpcom/utilities_internal.js
+++ b/chrome/content/zotero/xpcom/utilities_internal.js
@@ -2899,6 +2899,85 @@ Zotero.Utilities.Internal.getProcessID = function () {
 };
 
 
+Zotero.Utilities.Internal.Environment = {
+	/**
+	 * Unset an environment variable
+	 *
+	 * Since nsIEnvironment doesn't have an unset method
+	 */
+	unset: function (varName) {
+		Components.utils.import("resource://gre/modules/ctypes.jsm");
+		
+		let lib;
+		let result;
+		let success;
+		// Windows
+		// https://learn.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-setenvironmentvariablew
+		if (Zotero.isWin) {
+			lib = ctypes.open("kernel32.dll");
+			let SetEnvironmentVariable = lib.declare(
+				"SetEnvironmentVariableW",
+				ctypes.winapi_abi,
+				ctypes.int, // Return type
+				ctypes.char16_t.ptr, // Argument type: LPCTSTR (variable name)
+				ctypes.char16_t.ptr // Argument type: LPCTSTR (value, NULL to unset)
+			);
+			result = SetEnvironmentVariable(varName, null);
+			success = result != 0; // "If the function succeeds, the return value is nonzero."
+		}
+		// macOS or Linux
+		else {
+			lib = ctypes.open(Zotero.isMac ? "/usr/lib/libSystem.dylib" : "libc.so.6");
+			let unsetenv = lib.declare(
+				"unsetenv",
+				ctypes.default_abi,
+				ctypes.int, // Return type
+				ctypes.char.ptr // Argument type: const char *
+			);
+			result = unsetenv(varName);
+			success = result == 0;
+		}
+		
+		// Check the result
+		if (success) {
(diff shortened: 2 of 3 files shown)


adomasven:
Doesn't seem to work. Still opens in a new Firefox profile. MOZ_LEGACY_PROFILE seems to be the important env variable for me.

For testing if your default browser is not Firefox:

```sh
# make note of default browser
xdg-settings get default-web-browser
xdg-settings set default-web-browser firefox.desktop
# ensure the link is opened in firefox
xdg-open https://www.zotero.org
```

dstillman:
Yeah, `MOZ_LEGACY_PROFILES` would be the relevant variable for the reported problems, but we also probably don't want to launch people's normal Firefox installs in downgraded profiles by accident, so might as well unset both.

I only tested this on Windows, and it did fix this problem with PDFs launched in Dev Edition for me: https://forums.zotero.org/discussion/118931/zotero-7-0-8-firefox-is-default-pdf-viewer-but-zotero-opens-pdfs-into-wrong-profile

So I'm surprised this didn't fix things on Linux for you. I'll see if I can reproduce.

Separately, though, this solution (if it did work) might be too risky:

We want to make sure we don't somehow end up launching another instance of ourselves and not having the variables set properly. I don't think that could happen on macOS or Windows, because anything that relaunches Zotero uses the modified launcher that sets these variables to begin with.

On Linux, though, where we set these variables in the `zotero` shell script, it seems like an app restart probably launches `zotero-bin`, not `zotero`. I can trigger the Mozilla downgrade warning with these steps:

1. Launch `zotero` (doesn't need to be this PR).
2. Modify compatibility.ini in the profile to have a higher major version for `LastVersion`.
3. Clear these variables (if not running this PR):
```javascript
var env = Cc["@mozilla.org/process/environment;1"].getService(
	Ci.nsIEnvironment
);
env.set("MOZ_ALLOW_DOWNGRADE", "");
env.set("MOZ_LEGACY_PROFILES", "");
```
4. Restart with `Zotero.Utilities.Internal.quit(true);`.

This probably would never be a problem with `MOZ_ALLOW_DOWNGRADE`, because the app updates compatibility.ini when it starts, and it should never otherwise be changed to some higher version. But `MOZ_LEGACY_PROFILES` is just a scary setting in general, since it silently creates a new profile, and we won't necessarily always build from an ESR build, etc.

So maybe it's safer to clear these in the few places where we launch URLs and files and then set them again after?

FWIW, other places I was thinking of, which I think are all probably OK because they'd use the modified launcher:

1. Word processor integration on Windows
2. Dev tools
3. `zotero://` URLs launched externally?
4. Opening a RIS or CSL file
5. Restarting (other than the Linux case)
6. Upgrading

dstillman:
@adomasven: What's your specific profile setup that's causing it to use a new profile to begin with?

adomasven:
image


Default profile is Test and what runs with xdg-open

Zotero ran with esr-60 profile. 

I see in `about:support` that both ENV variables are present as seen by Firefox, and actually running `MOZ_LEGACY_PROFILES= xdg-open https://www.zotero.org` opens it in the wrong profile too, so setting an empty environment variable doesn't seem to disable this behavior.

> So maybe it's safer to clear these in the few places where we launch URLs and files and then set them again after?

This would make sense to protect us from this issue in the future.

dstillman:
> I see in `about:support` that both ENV variables are present as seen by Firefox, and actually running `MOZ_LEGACY_PROFILES= xdg-open https://www.zotero.org` opens it in the wrong profile too, so setting an empty environment variable doesn't seem to disable this behavior.

Oh, so that's the difference. Environment variables set to empty strings don't show up in about:support on Windows. (On Windows, environment variables are also only passed when we launch PDFs. The URL-launching mechanism seems to start with new environment.)

I guess this is actually kind of [documented](https://searchfox.org/mozilla-central/source/xpcom/threads/nsIEnvironment.idl):

```
     * - For Unix/Linux platforms we follow the Unix definition:
     * An environment variable exists when |getenv()| returns a non-NULL value.
     * An environment variable does not exist when |getenv()| returns NULL.
     * - For non-Unix/Linux platforms we have to fall back to a
     * "portable" definition (which is incorrect for Unix/Linux!!!!)
     * which simply checks whether the string returned by |Get()| is empty
     * or not.
```

But that's a problem, because nsIEnvironment doesn't provide a way to unset a variable. I guess I can try to do this via js-ctypes…

dstillman:
OK, try that. (I didn't yet move this to the places where we actually launch things.)

adomasven:
Yep, that does it.

dstillman:
OK, that should do it.

moetayuko:
@dstillman I use https://aur.archlinux.org/packages/zotero-bin on Arch, and the issue persists in 7.0.15. zotero starts a new firefox instance when opening a URL, where the environment variables are still set:
```
❯ strings /proc/2836008/environ | rg MOZ
MOZ_ENABLE_WAYLAND=1
MOZ_ASSUME_USER_NS=1
MOZ_LAUNCHED_CHILD=
MOZ_CRASHREPORTER_EVENTS_DIRECTORY=/home/xxx/.zotero/zotero/cd262deh.default/crashes/events
MOZ_APP_SILENT_START=
MOZ_ALLOW_DOWNGRADE=1
MOZ_LEGACY_PROFILES=1
```

dstillman:
@moetayuko: You'll need to test with the official tarball.

moetayuko:
> @moetayuko: You'll need to test with the official tarball.

That package does use the official tarball https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=zotero-bin#n18

dstillman:
And makes modifications to it. If you want our help, you need to test with the official tarball.

moetayuko:
> And makes modifications to it. If you want our help, you need to test with the official tarball, extracted into a folder without modifications.

Following your suggestion, I downloaded the tarball from the official website and moved the old ~/.zotero directory to start with a new profile. Unfortunately, the problem persists.

I tried two ways to start zotero:
1. ./zotero
2. ./zotero-bin -app app/application.ini

firefox instances spawned by both of them have the two environment variables

dstillman:
How exactly are you triggering this within Zotero?

moetayuko:
> How exactly are you triggering this within Zotero?

tried two ways:
1. click URL in the builtin pdf viewer
2. Help -> About Zotero, then click any link in it

moetayuko:
@dstillman this appears to be a wayland-specific issue. I managed to reproduce it in debian live cd:
1. boot with debian-live-12.10.0-amd64-kde.iso
2. logout, choose plasma (wayland) at the bottom left of sddm, then login
3. install libdbus-glib-1-2 with apt, which is a shared dependence of zotero
4. launch the preinstalled firefox, set it as the default browser
5. download the official zotero tarball, run with ./zotero, click and link,  and observe the mentioned environment variables in the newly launched firefox instance

PR #5001, opened 2025-01-22
Item box: Fix missing Copy/Paste menuitems on URL fields

### LINKED ISSUE #5000: Missing Copy/Paste context menu options on populated URL field
https://forums.zotero.org/discussion/121306/allow-for-copy-paste-in-an-items-url-field

Image

(And editing mode too, I guess.)

AbeJellinek:
And do a little related cleanup.

Fixes #5000

### DIFF
--- a/chrome/content/zotero/elements/itemBox.js
+++ b/chrome/content/zotero/elements/itemBox.js
@@ -63,6 +63,7 @@
 			this._selectFieldSelection = null;
 			this._addCreatorRow = false;
 			this._switchedModeOfCreator = null;
+			this._popupNode = null;
 
 			this._lastUpdateCustomRows = "";
 			// Keep in sync with itemPaneManager.js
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/zoteroPane.js
```diff
@@ -4408,46 +4408,6 @@ var ZoteroPane = new function()
...
-			
-			var noteElem = document.getElementById('zotero-note-editor')
-			noteElem.focus();
-			return true;
-		}
-		
-		return false;
-	});
-	
-
```
dstillman, line 4450: I assume this isn't supposed to be in this PR?

dstillman: It's not related, though. `document.popupNode` was just a [Mozilla property](https://udn.realityripple.com/docs/Web/API/Document/popupNode). It has no connection to context menus in the item pane — it was getting the selected text in the browser pane and adding it to the selected note in Zotero. That's obviously long since irrelevant, but we might resurrect it at some point (hence the comment to move it to server_connector.js), and in case, it doesn't make sense to remove it as part of this.

Issue #4949, opened 2024-12-20
Light and dark themes for reader content


mrtcode:
@dstillman @yexingsha @AbeJellinek 

As I’m working on a dynamic dark mode for PDF view that supports custom `background` and `foreground` colors, automatically adjusts other colors for optimal contrast, and even works with scanned papers, I think this is a good chance to introduce themes for both dark and light modes (which is often requested as well).

I think @AbeJellinek said that we can use `background` and `foreground` colors for EPUB view. Not sure about snapshots.

This is what I suggest:

1. No longer invert rendered image/ink annotation images in the reader sidebar, attachment annotation preview, and Zotero notes. Instead, simply darken the images slightly in dark mode.
2. Add two new prefs: `reader.lightTheme`, `reader.darkTheme`. Pref example:
    ```js
    {
      "name": "Solarized Dark",
      "background: "#FDF6E3",
      "foreground": "#657B83",
      // …potentially other properties in future…
    }
    ```
    If the pref is empty, it means the reader will not perform any color blending for that light or dark mode.
3. Add a pref `reader.themes` with default themes for example `Snow Storm`, `Sepia`, and `Solarized Dark`, allowing users or plugins to add new themes or modify the existing ones.
4. Eliminate `Use Dark Mode for Content` option from the View menu.
5. Redesign the EPUB appearance popup and use it for all PDF, EPUB, and snapshot view options. Include not only theme options but also options from the View menu, such as vertical/horizontal/wrapped scrolling, continuous/scrolled mode, no/odd/even spreads, split view, and more. These options are currently unavailable in the web library, and it is common for various types of readers to have an appearance popup that consolidates all these settings.

    Existing appearance popup:
    <img width="340" src="https://github.com/user-attachments/assets/98e618a2-ddea-4c0d-be92-c8361e09fc91" />

    ---

    Examples:

    <img width="295" src="https://github.com/user-attachments/assets/399df721-b7fa-4e57-a02c-424b9b7f3fde" />
    <img width="159" src="https://github.com/user-attachments/assets/fbee3bbb-edbd-4b44-b5ce-d7aa57f367b1" />
    <img width="155" src="https://github.com/user-attachments/assets/fc01ffe7-d68e-4a56-b91d-80041c8c88bc" />
    <img width="160" alt="Screenshot 2024-12-20 at 18 12 24" src="https://github.com/user-attachments/assets/e1ddc97a-c835-489d-8752-f0ff6a9ad95d" />

### Example themes
[...]

Issue #5003, opened 2025-01-23
Snapshot: Support reader mode?


AbeJellinek:
https://forums.zotero.org/discussion/118885/saving-web-site-snapshots-in-reading-format

I think this is a good idea. We would probably need to modify one of the existing open-source reader libraries (Mercury Parser, Mozilla Readability, or whatever) to maintain a mapping between positions in the output text and input DOM nodes in order to support annotations.

PR #5005, opened 2025-01-24
Implement appearance popup with themes

mrtcode:
Implements #4949

Rendered Ink and Image annotation images in the item pane and reader sidebars are no longer inverted. Instead, 80% brightness filter is applied.

`Use Dark Mode for Content` is now completely removed. Instead, you can directly pick a theme in dark mode.

Themes also work with scanned papers, as they directly replace pixels close to dark and close to white with the theme background and foreground colors. This only works well if the image is pretty sharp, without compression artifacts. Otherwise, it just darkens the page with a slightly visible theme background color. If we don't like that, we can always just reduce page brightness or invert it.

Only custom themes can be added, modified, or deleted, because it's better for users not to interfere with default themes. There's a high chance that we'll add more themes or properties in the future (i.e., adjust image rendering in PDF pages, or still support color inversion in some cases, or even use a custom CSS filter instead of background/foreground colors).

Themes have a name that serves as their ID and label, and are treated case-insensitively.

New prefs:

1. `zotero.reader.customThemes` with default value `[]`.
    ```js
    [
        {
            "name": "Custom 1",
            "background": "#F4ECD8",
            "foreground": "#5B4636"
        }
    ]
    ```
2. `zotero.reader.lightTheme` default value ''.
3. `zotero.reader.darkTheme` default value 'Dark'.

Original theme is effectively no theme.

Default themes are also localized:
```
pdfReader.theme.original = Original
pdfReader.theme.snow = Snow
pdfReader.theme.sepia = Sepia
pdfReader.theme.dark = Dark
```

Let me know what you think.


mrtcode:
Thanks for the great observations.

> * "Scroll" feels kind of stilted to me. "Scrolling"?

Changed.

> * Not seeing any way to tab into the popup (but I don't think you could tab into the EPUB appearance popup either)

Fixed.

> Edit dialog:
> * Text fields have no focus halo
> * No `:active` style on the buttons
> * Tabbing from a color picker swatch to the input field after it doesn't select the text in the field, but preserves the previous cursor position (or doesn't show a cursor if it hadn't previously been focused)
> * Escape doesn't cancel, Enter doesn't save

Fixed.


> * Opening the View menu logs an error because there's still code trying to enable/disable `view-menuitem-use-dark-mode-for-content`

Fixed.


> * If I'm on a PDF page with a link on it, and I press Tab to focus the link, then Tab again to move to the other split pane / the item pane, the focus halo around the link doesn't go away

Yes, that is a known issue. I'm thinking about how to fix it.

> * Switching spread modes with text selected doesn't reposition the selection popup

I somehow can't reproduce that. When I do it over the appearance popup, the text is unselected, and when I do it from the View menu, the popup position updates.

> * How do theme IDs work with localization? The ID is the label, but the label can be localized? I sort of feel like we should have a cleaner separation between the unique ID and the label for built-in themes (that can't be edited anyway). Treating the label as case-insensitive so it can also be an ID might cause more confusion than it's worth.

Ok, I added IDs. Default themes have `sepia`, `snow`, `dark`, while custom themes get `custom1`, `custom2`, etc.

---

Now, I guess this can be implemented for the EPUB view as well. Snapshots also need to be supported in one way or another, either using the old CSS filter methods or replacing text and background colors. However, the reading mode (#5003) might make more sense for that.

mrtcode:
> OK, pushed EPUB/snapshot support on that branch.

Snapshots work better than I expected.

> Spotted one more small issue in the PDF view: resize handles are barely visible in dark themes.

Fixed.

---

I think this is ready to be merged, but let's wait a few days in case @dstillman wants to take a final look.

PR #5006, opened 2025-01-25
Optimize context menus with collection popups

### LINKED ISSUE #4963: Slow "Add to Collection" menu with many collections
https://forums.zotero.org/discussion/120789/add-to-collection-error-1755011704

abaevbog:
- more efficiently check if a collection can be moved to another collection without calling `getDescendents` for every single candidate
- when fetching descendents of a collection, only fetch those of the relevant type
- It turns out that adding many top-level menuitems with icons into a menupopup takes a while to process and may freeze the UI on macOS. This issue only emerges when icons are present. To avoid it, the icons are hidden for all but the first few menuitems and then are dynamically added when the browser is idle after the popup appears.

Fixes: #4963

For comparison, this is with 2000 top level collections. The icons may take a while to display but at least the UI is not frozen for multiple seconds. Alternatively, we can consider not showing the icon on macOS for these context menus? Or Refactor them to only show the first few collections and hide the rest under "Show more..." submenu?

Before:

https://github.com/user-attachments/assets/65300282-e85e-4ba1-970a-ea5e840d20a3

After:

https://github.com/user-attachments/assets/1e5c64ff-027a-40a5-a04b-e1c1d0c77120

### DIFF
--- a/chrome/content/zotero/xpcom/utilities_internal.js
+++ b/chrome/content/zotero/xpcom/utilities_internal.js
@@ -1715,7 +1715,7 @@ Zotero.Utilities.Internal = {
             return longest;
         }, null);
 		// The images are removed from all menuitems but the first few
-		let menuItemsWithHiddenIcon = menuItemsWithIcons.slice(75);
+		let menuItemsWithHiddenIcon = menuItemsWithIcons.slice(100);
 		for (let node of menuItemsWithHiddenIcon) {
 			// Do not hide the image of the longest row so the menu has the right width from the start
 			if (node == longestNode) continue;
--- a/chrome/content/zotero/xpcom/utilities_internal.js
+++ b/chrome/content/zotero/xpcom/utilities_internal.js
@@ -1700,7 +1700,7 @@ Zotero.Utilities.Internal = {
 	 * Then, run a job through requestIdleCallback to make icons visible on remaining menuitems whenever the browser is free.
 	 * @param {Node<menupopup>} popup containing menuitems with icons
 	 */
-	showIconsDynamically(popup) {
+	showMenuIconsOnIdle(popup) {
 		// Here, the images are removed from all menuitems but the first 30
 		let menuItemsWithHiddenIcon = [...popup.childNodes].filter(n => n.getAttribute("image")).slice(30);
 		for (let node of menuItemsWithHiddenIcon) {
--- a/chrome/content/zotero/xpcom/utilities_internal.js
+++ b/chrome/content/zotero/xpcom/utilities_internal.js
@@ -1703,9 +1703,22 @@ Zotero.Utilities.Internal = {
 	 * @param {Node<menupopup>} popup
 	 */
 	showMenuIconsOnIdle(popup) {
-		// Here, the images are removed from all menuitems but the first 30
-		let menuItemsWithHiddenIcon = [...popup.childNodes].filter(n => n.getAttribute("image")).slice(30);
+		let topLevelMenuItems = [...popup.childNodes].filter(n => n.tagName == "menuitem");
+		let menuItemsWithIcons = topLevelMenuItems.filter(n => n.getAttribute("image"));
+		// If there are only a few menuitems with icons, just let them render normally
+		if (menuItemsWithIcons.length < 250) return;
+		// Find the node in the entire menu with the longest text. The icon is kept
+		// visible on that node from the start to let menu set the right width
+		let longestNode = topLevelMenuItems.reduce((longest, node) => {
+            if (!longest) return node;
+            if (node.textContent.length > longest.textContent.length) return node;
+            return longest;
+        }, null);
+		// The images are removed from all menuitems but the first few
+		let menuItemsWithHiddenIcon = menuItemsWithIcons.slice(75);
 		for (let node of menuItemsWithHiddenIcon) {
+			// Do not hide the image of the longest row so the menu has the right width from the start
+			if (node == longestNode) continue;
 			node.setAttribute("hidden-image", node.getAttribute("image"));
 			node.removeAttribute("image");
 		}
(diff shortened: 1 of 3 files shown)


chrome/content/zotero/zoteroPane.js
```diff
@@ -4054,6 +4054,8 @@ var ZoteroPane = new function()
 		
 		// Build menus for each top-level collection of this library
 		let collections = Zotero.Collections.getByLibrary(this.getSelectedLibraryID());
+		let children = selected.getDescendents(true, "collection");
```
dstillman, line 4057: Shouldn't this be `false`?

abaevbog: You're right! Sorry about that, I very much misunderstood what the `nested` param meant.

chrome/content/zotero/xpcom/utilities_internal.js
```diff
@@ -1693,6 +1693,35 @@ Zotero.Utilities.Internal = {
...
 		return menu;
 	},
+
+	/**
+	 * Adding many top level menuitems specifically with icons freezes UI on macOS.
+	 * To fix it, hide the icons from all menuitems except for the first few and insert them into the DOM.
+	 * Then, run a job through requestIdleCallback to make icons visible on remaining menuitems whenever the browser is free.
+	 * @param {Node<menupopup>} popup containing menuitems with icons
+	 */
+	showIconsDynamically(popup) {
```
dstillman, line 1703: `showMenuIconsDynamically`? `showMenuIconsOnIdle`?

dstillman:
1) We should probably only use `showMenuIconsOnIdle()` if there are actually a huge number of collections. E.g., I have 205 top-level collections in my test library and it loads totally fine. With this, loading just looks worse.

2) We should probably render more icons to start. On this monitor, I see 69 collections, so if it were set to 75 or 100 they'd at least all be rendered at first. (There's still a downside in that the width of the menu changes as it renders below the fold, though I wonder if we could fix that by scanning for the longest row and keeping its icon.)

abaevbog:
> We should probably only use showMenuIconsOnIdle() if there are actually a huge number of collections. E.g., I have 205 top-level collections in my test library and it loads totally fine. With this, loading just looks worse.

That makes sense. On my end, with 250 collections, there is a small delay that is noticeable but not disruptive. With 500 collections, though, the delay is ~1.5-2 seconds, which feels already a bit unpleasant. I assume this delay will also differ depending on the system, so I thought we err on the safe side and set 250 as the cutoff? For reference, this is the delay I see with 500 collections rendered normally without these optimization:

https://github.com/user-attachments/assets/f588b20c-ef00-4f5b-8764-29701bb75220

> We should probably render more icons to start. On this monitor, I see 69 collections, so if it were set to 75 or 100 they'd at least all be rendered at first.

Yeah, it does feel nicer to let the icons for all `menuitem`s currently visible on the screen render first. I increased it to 100, which does add to a small delay but I think it's worth it.

> (There's still a downside in that the width of the menu changes as it renders below the fold, though I wonder if we could fix that by scanning for the longest row and keeping its icon.)

Right, keeping the icon on the longest `menuitem` actually does help! I also changed the `buildAddItemToCollectionMenu` a bit to reuse the new  "New Collection..." `menuitem`. It's label is set asynchronously after initial icons are rendered, so for a moment it would appear empty as in my earlier recordings. If all collections have short names, "New Collection..." would also increase the width when the label is set. Now, after the menu was opened, it will be set from the very beginning.

With these updates, this is what the menus behave like now with 2500 collections:

https://github.com/user-attachments/assets/5d8247cb-306f-4394-9025-0b45843cf1b0

abaevbog:
> I also changed the buildAddItemToCollectionMenu a bit to reuse the new "New Collection..." menuitem. It's label is set asynchronously after initial icons are rendered, so for a moment it would appear empty as in my earlier recordings.

Oh, though we do have to update the `command` listener, otherwise it will keep adding the first selected item... This "New Collection..." bit is turning out a bit less trivial than I initially thought but I feel like it's still worth it.

abaevbog:
I think that should do it. We'll still recreate the "New Collection..." `menuitem` to set the right `command` listener but we'll also set the label directly if the "New Collection..." was previously created. That way, it will still appear before fluent runs after the icons appear.

abaevbog:
To revisit, this issue still exists in current beta build. With 30,000 collections, macOS UI still freezes for me for 15-20 seconds when I hover over the "Add to Collection" menu option. Rebasing these changes on `main` still helps - with icons rendered lazily, the UI does not freeze.

After some investigation, the issue from https://forums.zotero.org/discussion/128804/5s-delay-before-new-collection-new-subcollection-dialog-appears-with-large-collection seems slightly different. The icons were not that big of a problem on windows to begin with. With 30,000 collections I do see a delay after hovering over "Add to Collection" on current Beta. But the delay is <5 seconds, and completely removing collection icons from the menu does not speed it up.

When I click on "New Collection", this is what the delay looks like:

https://github.com/user-attachments/assets/5f3305a8-621d-4374-8e82-2bd8351c12e7

And, as expected, there will be no delay if I comment out the loading of collections into the dropdown. All of the above makes me think that delays on windows are just caused by adding thousands of nodes into the DOM.

I'm looking into the best way to optimize the "New Collection" dialog - something like opening the dialog first and then loading the collections list helps the UI feel more responsive. But I am wondering if handling menupopups with thousands of collections is actually worth it? Finding a specific collection in a context menu of thousands options is incredibly challenging. Would it be worth to consider replacing the collections menulist with something like a collectionTree? It would allow us to lazily render only relevant collections avoiding loading delays, and likely make it easier to actually find a collection?
Same idea could apply to other menus that render collections lists (e.g. "Add to Collection", "Move to", "Copy to") - we could render at most, say, 500 collections, and hide the rest behind a "More collections..." menu option that would open a dialog with a similar collection tree?

PR #5012, opened 2025-01-28
Negative selection for tags in tag selector

### LINKED ISSUE #3858: Negative tag selection in tag selector
https://forums.zotero.org/discussion/101433/negative-tag-selection-in-the-tag-selector and many others

An age-old request, and probably trivial to implement  — we just need to decide how we want to do it.

abaevbog:
- added context menu option "Exclude tag" that will exclude the items with the selected tag from search results
- search condition also includes `includeChildren:true`, otherwise if a parent has a tag but a child attachment does not, those rows would still remain in `itemTree`, even though the intention is probably to have them gone.
- Backspace on a focused tag will also mark it as excluded
- clicking on a tag removes the "excluded" status

There is likely a number of other ways to approach the interface but this felt like the most straightforward starting point.

Fixes: #3858 

https://github.com/user-attachments/assets/41d306f5-35e7-4f13-b8c3-feac4134aaf6

### DIFF
--- a/chrome/content/zotero/components/tagSelector.jsx
+++ b/chrome/content/zotero/components/tagSelector.jsx
@@ -49,6 +49,7 @@ class TagSelector extends React.PureComponent {
 					fontSize={this.props.fontSize}
 					lineHeight={this.props.lineHeight}
 					uiDensity={this.props.uiDensity}
+					excludeTag={this.props.excludeTag}
 				/>
 				<div className="tag-selector-filter-pane">
 					<div className="tag-selector-filter-container">
@@ -90,6 +91,7 @@ TagSelector.propTypes = {
 	}),
 	onSelect: PropTypes.func,
 	onKeyDown: PropTypes.func,
+	excludeTag: PropTypes.func,
 	onTagContext: PropTypes.func,
 	loaded: PropTypes.bool,
 	width: PropTypes.number.isRequired,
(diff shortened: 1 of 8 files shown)


bloomingseed:
From a hope to have a NOT [tag] feature in mind, i came across these posts on the forums [\[1\]](https://forums.zotero.org/discussion/64955/tags-not-negation-operator)[\[2\]](https://forums.zotero.org/discussion/658/filter-by-does-not-have-tag-as-well-as-by-has-tag) (the post [\[2\]](https://forums.zotero.org/discussion/658/filter-by-does-not-have-tag-as-well-as-by-has-tag) was posted even as far as in 2007!)

Finally arrived at seemingly the first PR for this feature.  Really hope it gets merged for a release soon.

abaevbog:
@AbeJellinek thanks for the notes!

> Or should we continue our trend of not adding the target object noun to every context menu? So just "Exclude"?

I only added "tag" because other menuitems ("Rename Tag...", "Delete Tag...") have it too. I thought it's important to remain consistent within the same menu, because otherwise it might seem that "Exclude" does not refer to tags in the same way that "Rename Tag..." does. So should we do "Exclude Tag" to avoid removing "Tag" from all other menuitems?

> For consistency, should we also add the standard left-click action to the menu as well ("Include"?)

That makes a lot of sense. Not sure about the naming though ... To me, "Include" implies that the tag itself would be included somewhere, no? By this logic, "Exclude" may be not much better to be honest. To avoid any confusion, would it be too long to say something like "Keep/Display items with this tag" for normal selection and "Hide items with this tag" or negative selection?

Hirnmoder:
I'd really love this feature to be shipped with Zotero. It's such a qol improvement!

tobiweisss:
That would be a really great feature that I would love to see in Zotero

PR #5015, opened 2025-01-28
Local API: Add write endpoints, track local data object versions

### LINKED ISSUE #5011: Local API version handling
I broke #5004 by changing a request in a test from `since=-1` to `since=0`. (Sorry @AbeJellinek!) Objects from the web API would only have versions >0, so a tool written for the web API would use `since=0`. New local objects don't have versions, though, so those don't match `since=0`.

We could:

A) Make `since=0` match all objects, including unsynced local objects, because that's the intention of `since=0`
B) Require local API consumers to use `since=-1`, because…I don't know? It's a way to differentiate between unsynced and synced objects?

But the larger problem here is that, as we've mentioned in the context of supporting write requests, versioned requests to the local API just don't really make sense. New local objects don't have versions, and the app can't increment versions for write requests. So then I guess another option might be:

C) Have the client to maintain separate local versions, separate from the online versions

I'm not sure if that would be worth it, and I've given exactly zero thought to whether that would actually work, but it'd probably be the only way to handle write requests locally.

AbeJellinek:
I haven't run the entire test suite locally, so we'll see what happens on CI, but this is intended to be a relatively unobtrusive set of changes to track a monotonically increasing library/data object version number.

When the local API is enabled, modifying a data object (item, collection, or search) increments the library's `clientVersion` (once per transaction), then sets the data object's `clientVersion` to the same value.

Also updated `toResponseJSON()` to return local storage properties as well as versions.

Closes #5011 (option C)

### DIFF
--- a/chrome/content/zotero/xpcom/data/dataObject.js
+++ b/chrome/content/zotero/xpcom/data/dataObject.js
@@ -23,6 +23,8 @@
     ***** END LICENSE BLOCK *****
 */
 
+const { XPCOMUtils } = ChromeUtils.importESModule("resource://gre/modules/XPCOMUtils.sys.mjs");
+
 let lazy = {};
 XPCOMUtils.defineLazyPreferenceGetter(
 	lazy,
--- a/chrome/content/zotero/xpcom/data/dataObject.js
+++ b/chrome/content/zotero/xpcom/data/dataObject.js
@@ -23,15 +23,6 @@
     ***** END LICENSE BLOCK *****
 */
 
-const { XPCOMUtils } = ChromeUtils.importESModule("resource://gre/modules/XPCOMUtils.sys.mjs");
-
-let lazy = {};
-XPCOMUtils.defineLazyPreferenceGetter(
-	lazy,
-	'shouldTrackClientVersions',
-	'extensions.zotero.httpServer.localAPI.enabled',
-);
-
 /**
  * @property {String} (readOnly) objectType
  * @property {String} (readOnly) libraryKey
@@ -1307,14 +1298,12 @@ Zotero.DataObject.prototype._finalizeSave = async function (env) {
 		Zotero.logError("skipCache is only for new objects");
 	}
 
-	if (lazy.shouldTrackClientVersions) {
-		let libraryClientVersion = await this.library.incrementClientVersion();
-		await Zotero.DB.queryAsync(
-			`UPDATE ${this.ObjectsClass.table} SET clientVersion = ? WHERE ${this.ObjectsClass.idColumn}=?`,
-			[libraryClientVersion, this.id]
-		);
-		this._clientVersion = libraryClientVersion;
-	}
+	let libraryClientVersion = await this.library.incrementClientVersion();
+	await Zotero.DB.queryAsync(
+		`UPDATE ${this.ObjectsClass.table} SET clientVersion = ? WHERE ${this.ObjectsClass.idColumn}=?`,
+		[libraryClientVersion, this.id]
+	);
+	this._clientVersion = libraryClientVersion;
 };
 
 
--- a/chrome/content/zotero/xpcom/data/dataObject.js
+++ b/chrome/content/zotero/xpcom/data/dataObject.js
@@ -1524,6 +1524,9 @@ Zotero.DataObject.prototype.toResponseJSON = function (options = {}) {
 		meta: {},
 		data: this.toJSON(options)
 	};
+	// Keep the data block's version consistent with the top-level version (toJSON() reports
+	// the synced version, which differs from clientVersion when syncedVersionProperty is false)
+	json.data.version = json.version;
 	if (options.version) {
 		json.version = json.data.version = options.version;
 	}
--- a/chrome/content/zotero/xpcom/data/dataObject.js
+++ b/chrome/content/zotero/xpcom/data/dataObject.js
@@ -1462,6 +1462,8 @@ Zotero.DataObject.prototype._initErase = function (env) {
 };
 
 Zotero.DataObject.prototype._finalizeErase = async function (env) {
+	await this.library.incrementClientVersion();
+	
 	// Delete versions from sync cache
 	if (this._objectType != 'feedItem') {
 		await Zotero.Sync.Data.Local.deleteCacheObjectVersions(
--- a/chrome/content/zotero/xpcom/data/dataObject.js
+++ b/chrome/content/zotero/xpcom/data/dataObject.js
@@ -1499,10 +1499,15 @@ Zotero.DataObject.prototype._finalizeErase = async function (env) {
 
 
 Zotero.DataObject.prototype.toResponseJSON = function (options = {}) {
+	// Default to showing synced properties, since that's what the API does, and this function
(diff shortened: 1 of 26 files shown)


EwoutH:
Thanks for working on this! Tracking object versions locally is an important step towards local API write support right?

What's needed for this PR now? Review from maintainers?

chrome/content/zotero/xpcom/data/dataObject.js
```diff
@@ -23,6 +23,15 @@
...
 */
 
+const { XPCOMUtils } = ChromeUtils.importESModule("resource://gre/modules/XPCOMUtils.sys.mjs");
+
+let lazy = {};
+XPCOMUtils.defineLazyPreferenceGetter(
+	lazy,
+	'shouldTrackClientVersions',
+	'extensions.zotero.httpServer.localAPI.enabled',
+);
```
dstillman: Gating on the pref being enabled would mean that if you disabled the API, made a change, and re-enable it, clients would miss updates. I think we originally had some plan for dealing with a `since=` from before the earliest available version — I guess specifically for deletes, so that we could drop ancient delete-log entries — but I don't think we ever implemented that, and I don't think any clients would handle a 409 or whatever that would throw. We could do that, but it's probably fine to just drop the gate and always update clientVersion. I ran some quick benchmarks and it didn't seem like it had any noticeable effect on performance, even for 2000-item adds as separate transactions.

chrome/content/zotero/xpcom/data/item.js
```diff
@@ -6170,6 +6161,12 @@ Zotero.Item.prototype.toResponseJSONAsync = async function (options = {}) {
 	else if (this.isImportedAttachment()) {
 		json.links.enclosure.length = await getFileSize(this);
 	}
+	
+	if (this.isStoredFileAttachment() && !options.skipStorageProperties) {
+		json.data.mtime = await this.attachmentModificationTime ?? null;
+		json.data.md5 = await this.attachmentHash ?? null;
+	}
+
```
dstillman, line 6217: If we're adding these in the async path here, why are we changing the default toJSON() behavior to emit them? That could affect a lot of other things, and is at least untested.

dstillman:
pyzotero uses `Authorization: Bearer <key>` and `params=` file uploads, so I added support for those.

dstillman:
(Its `attachment_simple()` function still won't work because that expects `/items/new` to work, which we don't support and probably don't want to.)

dstillman:
A couple more fixes. I ran a battery of tests with a ([fixed](https://github.com/urschrei/pyzotero/pull/331)) version of pyzotero, and it looks good! Just the couple issues mentioned above.

chrome/content/zotero/xpcom/data/dataObject.js
AbeJellinek: Sure, I have no problem with that.

chrome/content/zotero/xpcom/data/item.js
AbeJellinek: Oh, yeah, that was not a good change. I'll pull it out of `toJSON()`.

AbeJellinek:
Great!

It occurs to me that we'll need to mention this as a breaking change for the local API - any code that caches versions returned by the local API could break, and mixing local API and web API code (as I think some people do, since the local API doesn't support writes!) becomes more complicated.

Should we add a marker indicating that versions returned in the response are local, so API consumers know whether they're talking an updated version of Zotero?

dstillman:
You mean because before we were returning the server version, and now we're returning a local version? Arguably caching was just broken before, because an item could change locally without the version being bumped (if it hadn't yet synced), right?

But yes, trying to read from the API and then write to the web API with that version would not go well…

AbeJellinek:
> Arguably caching was just broken before, because an item could change locally without the version being bumped (if it hadn't yet synced), right?

That's true, and hopefully that discouraged local API consumers from relying on it. But syncing (when set up) is frequent enough that I don't think we can count on that being the case.

It could be a header - `Zotero-Object-Version-Context`? `Zotero-Object-Version-Partition`? Or something? - that contains a random ID, persisted per client in a pref, that consumers:

- Can read from responses, cache, and compare to know whether they're looking at versions that exist in the same world as ones they received previously
- Need to send back in every write request to the local API, or their request will immediately be rejected
- CANNOT send to the web API, or it will reject their request (requires a small change in the dataserver)

AbeJellinek:
Or the dataserver returns its own magic value, which could just be the server hostname or something like that, and which it doesn't require but does validate if it's provided.

dstillman:
Well, there's a broader idea where the API includes a random identifier that allows clients to know if they're talking to the same endpoint in general, which could also be used to better support custom servers (real or development). That could be advisory or I guess something the server validates. I was assuming it would just be the former — tell syncing clients to perform a full sync and make clear to other consumers that versions aren't equivalent.

AbeJellinek:
`Zotero-API-Endpoint-ID`?

I guess it could just be advisory. Even if we wanted to require it in the local API, I think it would only make sense to require it for writes, and those weren't supported before anyway.

AbeJellinek:
But we could validate it if provided in a request - no harm in that, and it allows clients to avoid making an initial request to check the ID if they think they already know it.

dstillman:
`Zotero-Server-ID`?

AbeJellinek:
Sure.

AbeJellinek:
Alright, see if those do the trick.

chrome/content/zotero/xpcom/server/server_localAPI.js
```diff
@@ -65,6 +91,147 @@ That said, there are benefits:
...
+ * Get this server's persistent ID, generating and storing it on first use.
+ * @returns {string} A 12-character server ID
+ */
+function getLocalAPIServerID() {
+	let id = Zotero.Prefs.get('httpServer.localAPI.serverID');
+	if (!id) {
+		id = Zotero.Utilities.randomString(12);
+		Zotero.Prefs.set('httpServer.localAPI.serverID', id);
+	}
+	return id;
```
dstillman: This has to be in the DB, not in a pref, since it's tied to the data in the database. We can store it in the `settings` table.

One issue would be DB backup restores, which could change the data out from under the consumer. If we want to handle that, we could store a DB version in a pref that gets updated monotonically along with a version stored in the DB after any DB write. At startup, check whether pref-version > DB-version and roll a new server id at that time, to force consumers to perform a full sync. Wouldn't handle all cases — someone could do a Time Machine restore that restored profile dir and data dir together — but would handle the most common ones.

AbeJellinek: OK, in the settings table now.

How concerned are we about the DB restore case? I think that check could lead to some confusion, but I see why we might want it, and it would be easy to put in.

dstillman:
Rebased, and some fixes added

dstillman:
Merged in 9dd17a212ed63ef5eb72fc1c04d1d21cb228b537, 77f2432103e60e2b4545b1add6ac78d9cb8a5997, a37a9e782a515c26e04e67e8e692983d94753b41

PR #5016, opened 2025-01-29
Macos progress bar

### LINKED ISSUE #1923: Display the progress bar for macOS doc refresh
This is requested every now and then, and we should at least reconsider it, since it creates quite an inconvenience for people with hour-long refreshes https://forums.zotero.org/discussion/86458/feature-request-refresh-progress-indicator#latest.

This is the original reasoning for why we chose to not show it: 
https://github.com/zotero/zotero/blob/8fc316f72796d8b7ff4e29e5f8bc3123d8d2a7e7/chrome/content/zotero/xpcom/integration.js#L3030-L3039

adomasven:
Closes #1923 

The issue with displaying the progress bar on macOS is that when you close it (to then open a citation dialog), the main Zotero window is raised (appears in the foreground with the citation dialog focused on top of it). This is because when you close a window of an application on macOS it automatically focuses the last focused window of the app, which in our case after closing the progress bar is the Zotero window.

In the past we haven't been showing the citation dialog on macOS due to this, as it's quite jarring, especially if the Zotero window raises under the quick format dialog

This PR makes it so that Refresh, Unlink Citations and Add/Edit Bibliography show a progress bar. After Refresh and Unlink Citations is finished, Document.activate is called, which means that Word receives focus and the Zotero window is not raised. For Add/Edit Bibliography, if a bibliography already exists, then the Edit Bibliography dialog is displayed, with the Zotero library window raised, which is not perfect, but fine, since Edit Bibliography dialog is large. Since Add/Edit Bibliography refreshes the document and may take a while, this is still better than no progress bar.

For Add/Edit Citation, Add Note and Doc Prefs we do not display an initial progress bar, because these operations are generally faster and only require fetching doc prefs to open their respective dialogs. A progress bar will be displayed after confirming these dialog. This brings the Zotero window into the foreground, but if you are changing citation styles or such, it's much more important to know when the progress has finished than to not get any progress dialog at all.

Note that after all these operations finish, we call Document.active, which focuses the relevant doc editor, so in the end the fact that the Zotero window was raised doesn't cause huge issues.

I believe, that a variety of "intermittent" errors that we see from macOS are actually errors from users who are having their documents updated with no indication (progress bar) from Zotero, and removing or editing citations while Zotero is working on their document. This change should make it much easier for us to help or troubleshoot the errors for those users, as well as generally improve the usability of the integration experience with larger docs where operations may take a long time but display no indication that Zotero is working.

### DIFF
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -1295,10 +1295,10 @@ var Zotero_QuickFormat = new function () {
 			editor.style.width = `${editorDesiredWidth}px`;
 		}
 	}
-	async function _resizeWindow() {
+	async function _resizeWindow(height, width) {
 		let box = document.querySelector(".citation-dialog.entry");
-		let height = box.getBoundingClientRect().height;
-		let width = WINDOW_WIDTH;
+		height = height || box.getBoundingClientRect().height;
+		width = width || WINDOW_WIDTH;
 		// Force resizing if there is no max height set on the window
 		let maySkip = !!document.documentElement.style.maxHeight;
 		if (Math.abs(height - window.innerHeight) < 5 && maySkip) {
@@ -1524,9 +1524,12 @@ var Zotero_QuickFormat = new function () {
 	/**
 	 * Called when progress changes
 	 */
-	function _onProgress(percent) {
+	async function _onProgress(percent) {
 		var meter = document.querySelector(".citation-dialog.progress-meter");
-		if(percent === null) {
+		meter.style.display = "";
+		await Zotero.Promise.delay(10);
+		_resizeWindow(meter.parentElement.parentElement.getBoundingClientRect().height);
+		if (percent === null) {
 			meter.removeAttribute('value');
 		} else {
 			meter.value = Math.round(percent);
@@ -1542,6 +1545,7 @@ var Zotero_QuickFormat = new function () {
 		try {
 			_updateCitationObject();
 			document.querySelector(".citation-dialog.deck").selectedIndex = 1;
+			_onProgress(null);
 			io.accept(_onProgress);
 		} catch(e) {
 			Zotero.debug(e);
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -1544,7 +1544,8 @@ var Zotero_QuickFormat = new function () {
 		accepted = true;
 		try {
 			_updateCitationObject();
-			document.querySelector(".citation-dialog.deck").selectedIndex = 1;
+			document.querySelector(".citation-dialog.main").hidden = true;
+			document.querySelector(".citation-dialog.progress-container").hidden = false;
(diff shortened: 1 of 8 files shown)


abaevbog (commented):
Everything is working as expected for me! These are just a couple of small interface-related suggestions:

chrome/content/zotero/integration/quickFormat.js
```diff
@@ -1524,9 +1524,12 @@ var Zotero_QuickFormat = new function () {
 	/**
 	 * Called when progress changes
 	 */
-	function _onProgress(percent) {
+	async function _onProgress(percent) {
 		var meter = document.querySelector(".citation-dialog.progress-meter");
-		if(percent === null) {
+		meter.style.display = "";
+		await Zotero.Promise.delay(10);
+		_resizeWindow(meter.parentElement.parentElement.getBoundingClientRect().height);
```
abaevbog, line 1531: For me, the deck (`meter.parentElement.parentElement`) retains the height of the `editor`. If there are many lines of bubbles, `meter.parentElement.parentElement.getBoundingClientRect().height` is the same as the current height, so this line does not change the window height for me.

Screenshot 2025-01-29 at 9 43 16 AM

Do we want to resize the window to only have enough space for the progress bar? We could run `editor.replaceChildren()` in `accept` to remove all bubbles after the citation object is updated to let `editor` (and the parent deck) go back to its original single-line height?

chrome/skin/default/zotero/integration.css
```diff
@@ -216,8 +216,12 @@ span.zotero-bubble-input {
 	-moz-window-dragging: no-drag;
 }
 
-.citation-dialog.progress-meter[value] {
+.citation-dialog progress {
 	visibility: visible;
+	border-radius: 10px;
+	margin: 0;
+	flex: 1;
+	height: 10px;
```
abaevbog, line 224: Should we also add `align-self: center;` so that the progress bar is in the center of the dialog, as opposed to being pushed into the upper-half?

Screenshot 2025-01-29 at 10 17 16 AM

abaevbog (commented):
Everything is working as expected for me! These are just a couple of small interface-related suggestions:

chrome/content/zotero/integration/quickFormat.js
adomasven: I'm generally not happy with the need to rely on the XUL deck here, since it seems to replace the items in some async way. I guess it's slower when there's more elements. But it's easy to switch to a more reliable approach.

chrome/skin/default/zotero/integration.css
adomasven: Not relevant since the dialog just has to (and will now) resize to only contain the progress bar.

adomasven:
@abaevbog addressed your comments. The dialog will now correctly resize when finalizing and displaying the progress bar so that only it is visible which was the intention in the first place but the deck element was acting weird.

abaevbog:
Great, yeah it all works well for me now!

Issue #4691, opened 2024-09-17
[Linux] Install and configure .desktop file on Zotero launch


adomasven:
Many Linux applications do this. On running Zotero we should check if we're running from a user-writeable directory (installed in userspace and as such probably not via the .deb package), and if so, install the .desktop file into `~/.local/share/applications` with the correctly set path to the Zotero binary and then place the Zotero icon into `~/.local/share/icons`. Then the Icon entry can simply be `Icon=Zotero`.

dstillman:
So the "installation" process would be putting the folder somewhere and running `./zotero` once from the command line?

adomasven:
Well installation is putting the folder somewhere. This is setting up a .desktop entry so that it's listed in the application manager of user's choice.

dstillman:
Installation as in what we'll say on the [Installation page](https://www.zotero.org/support/installation#linux). I'm just saying that instead of 2 and 3 on there, we'll just say to run `./zotero` once.

dstillman:
What are some examples of apps that do this?

adomasven:
Intellij IDEs do this. Wine creates entries there for Windows software that you run/install. I had a bunch of entries from all kinds of software in `.local/share/applications` but I recently removed them as part of general cleanup of buildup from the last 8 years. It's not uncommon and the intended way for userspace-installed-software to make itself available to launchers.

cadadr:
In 20 years on Linux I have never encountered this method of installation. Nevertheless I think it's quite nice.

It might be a nice idea to (1) make sure it respects whatever `XDG_...` variable governs `~/.local/share/applications`, (2) more importantly, instead of doing it automatically, to ask for user's confirmation via a dialog in the first run, and (3) perhaps a way to pick an alternate directory, because you can have more than one directory where these .desktop files go, by adding extra directories to `XDG_DATA_DIRS`. For example, I used to have a separate one for my personal .desktop files set in it.

I admit (3) is a bit extra and you can probably expect a user knowledgeable enough to know about `XDG_DATA_DIRS` to be able to manage it themselves. But I would really like an option to say no to the installation of the file by Zotero itself because it might end up, in a few scenarios, adding an extra Zotero entry to the application menu, which might then have problematic results if the two different installations diverge somehow, e.g. in their version.

adomasven:
[...]

PR #5019, opened 2025-01-30
Add script to install .desktop file on Linux

### LINKED ISSUE #5014: `install` script for Linux
To replace `set_launcher_icon` and manually symlinking the desktop file to the launcher folder (which can just be `"$XDG_DATA_HOME/applications"`, falling back to `~/.local/share/applications/` if `$XDG_DATA_HOME` isn't set and prompting the user if `~/.local/share/applications/` doesn't exist).

`zotero.desktop` should be moved out of the root folder to avoid confusion, or maybe just be embedded in the `install` script, and `install` should write it to the launcher folder with a hard-coded path to `zotero` and the icon file (which will solve lingering problems with https://github.com/zotero/zotero/issues/4163 on different Ubuntu versions).

The install script should say what it did and clarify that the desktop file is pointing to the current folder (so people don't think they can just delete the folder after running `install`).

We should also try to refresh the launcher, which probably varies by distro.

If there are common distros that don't support desktop files, we should try to detect those and bail.

This can probably supersede https://github.com/zotero/zotero/issues/4691, since if you can run `./zotero` you can run `./install`, and it's not even clear if you should run `./zotero` or `./zotero-bin` without reading instructions.

dstillman:
Closes #5014


dstillman:
Oh, I guess we could just use `desktop-file-install` instead of copying and calling `update-desktop-database` directly…

dstillman:
Eh, I guess not. I thought `desktop-file-install` would automatically use a local dir if you don't run it with sudo, but that doesn't seem to be true — it tries to install to `/usr/share/applications` unless you pass `--dir`. So we could check ourselves whether we're running as root, but probably fine to just stick to a manual local install.

adomasven:
We could also consider adding actions, like launching in troubleshooting mode, etc.

dstillman:
So, this works, but it also shows if Zotero is open, which is kind of confusing:

```
DESKTOP_CONTENT="""[Desktop Entry]
Name=${APP_NAME}
Comment=Your personal research assistant
Exec=${APPDIR}/zotero -url %U
Icon=${APPDIR}/icons/icon128.png
Type=Application
Terminal=false
Categories=Office;
MimeType=text/plain;x-scheme-handler/zotero;application/x-research-info-systems;text/x-research-info-systems;text/ris;application/x-endnote-refer;application/x-inst-for-Scientific-info;application/mods+xml;application/rdf+xml;application/x-bibtex;text/x-bibtex;application/marc;application/vnd.citationstyles.style+xml
X-GNOME-SingleWindow=true
Actions=SafeMode

[Desktop Action SafeMode]
Name=Start in Troubleshooting Mode
Exec=${APPDIR}/zotero -safe-mode
"""
"""
```

Does that work for you?

dstillman:
It's also pretty rare that someone can't start up enough to get to the menu, so actually, this probably isn't worth it.

Starting with the debug logging window open would be another thing for troubleshooting, but, I dunno, we've made it this far. And this is Linux — people really should be able to run from the command line if they really need to.

adomasven:
Sure. Otherwise this is great.

dstillman:
Reverted in favor of #4691

PR #5022, opened 2025-01-31
Track changes in Style Editor and Scaffold

AbeJellinek:
With new general-purpose `DocumentManager` class.

The two editors have a similar (lack of) document model - rather than storing the current style/translator and updating it after user interactions, they just rebuild it from scratch as needed based on the state of the UI. I wrote this to support that design, but it should still work just as well with a document-based model.

### DIFF
--- a/chrome/content/zotero/documentManager.mjs
+++ b/chrome/content/zotero/documentManager.mjs
@@ -26,6 +26,8 @@
 var { Zotero } = ChromeUtils.importESModule("chrome://zotero/content/zotero.mjs");
 
 export class DocumentManager {
+	_editorName;
+	
 	_onSave;
 	
 	_currentState;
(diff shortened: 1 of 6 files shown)


chrome/content/zotero/documentManager.mjs
```diff
@@ -0,0 +1,147 @@
...
+		}
+		if (this._currentState === this._cleanState) {
+			return true;
+		}
+
+		let title = this._metadata.title
+			? await Zotero.ftl.formatValue('document-manager-confirm-dialog-title', {
+				title: this._metadata.title
+			})
+			: await Zotero.ftl.formatValue('document-manager-confirm-dialog-title-new');
```
dstillman, line 120: This doesn't work on Windows, where the dialog title goes in the title bar:

Screenshot 2025-02-13 at 1 51 56 AM

(Not sure about Linux.)

We can use this on macOS and use something else (just the name of the editor?) for the title on Windows(/Linux) with the title string in the description, and then either add the second line after two newlines or just leave it out, since it's just there to copy macOS.

Issue #4952, opened 2024-12-21
Auto-expand libraries and collections during drag and drop


dstillman:
https://forums.zotero.org/discussion/113826/zotero-7-beta-expand-collections-or-libraries-when-doing-a-drag-drop

(And, perhaps, close again when the drag is done, like "spring-loaded folders" in macOS Finder.)

PR #5024, opened 2025-02-03
Expand collectionTree containers on dragover

abaevbog:
- when something is dragged over a collapsed collection or group, it will be expanded to reveal children so one could drop the dragged item/collection/ into them
- when drag is over, the expanded collections will collapse, except for the ancestors of the newly selected collection
- trigger collapse of expanded collections on `dragend` on the entire document (in case `dragend` fires not on the collection tree), or from `onDrop` handler, from where we will wait for selection to change before collapsing rows

Fixes: #4952

https://github.com/user-attachments/assets/6a89984f-8747-430c-a27c-f6cdbcbac203

### DIFF
--- a/chrome/content/zotero/collectionTree.jsx
+++ b/chrome/content/zotero/collectionTree.jsx
@@ -1473,7 +1473,7 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 							this._flashingRow = index;
 							this.tree.invalidateRow(index);
 							// wait for the flashing to finish and then expand the container
-							await Zotero.Promise.delay(300);
+							await Zotero.Promise.delay(300); // 0.2s CSS animation length * 1.5 runs
 							this._flashingRow = null;
 							this._expandedRowsOnDrag.add(treeRow.id);
 							if (!this.isContainerOpen(index)) {
--- a/chrome/content/zotero/collectionTree.jsx
+++ b/chrome/content/zotero/collectionTree.jsx
@@ -107,6 +107,7 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 		this._filterInitialCollapsedRows = [];
 		this._treeWasFocused = false;
 		this._hiddenFocusedRow = null;
+		this._expandedRowsOnDrag = new Set();
 		
 		this.onLoad = this.createEventBinding('load', true, true);
 	}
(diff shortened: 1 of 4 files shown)


abaevbog:
There was a number of reports of target collection opening up if you drop an item into it. Most of the from a few years ago but there is a recent report of it happening in zotero 7 https://forums.zotero.org/discussion/comment/490017/#Comment_490017.

I never ran into this issue but in this PR, opened collections are recorded before drag and restored after drop, so it might be able to help with that as well.

dstillman:
I haven't tried this yet, but going by the video, this shouldn't be immediate — there should be a bit of a delay to allow dragging through the pane without crazy things happening. See folders in macOS Photos for an example. There's a delay before opening a hovered folder, and if you move off of it, there's a delay before it closes again.

There's also a double-flash effect (also done in Finder) when opening, which we could try to replicate, but that's less important.

dstillman:
No… Not sure what you're seeing, but it's a second or less in Photos. Slightly longer to close.

https://github.com/user-attachments/assets/c323c575-8239-4bea-b0fe-0a421aedf9d5

abaevbog:
> No… Not sure what you're seeing, but it's a second or less in Photos. Slightly longer to close.

Yeah, this looks way better than what I see on my end! In Photos, it seemingly takes me about 3 seconds before the folders open. It is possible that I don't keep the mouse positioning steady, while also pressing on mac's trackpad. Or it could be that my OS is a bit out of date. In any case, this behavior is a good blueprint to replicate.

I updated this to only expand a row if you drag over it and keep the mouse over that row for a second. That way, we should avoid opening irrelevant containers in the way of the drag. Also, if you drag something out of the `collectionTree`, all expanded collections will auto-collapse after one second. They will not auto-collapse if you drag back into the `collectionTree` before it happens.

Finally, I also added the flashing effect similar to Photos. The color of the flash is different depending on the background of the row (`--accent-blue50` if `.drop` is set which adds a dark background, and `--accent-azure` otherwise). What do you think? This is the updated behavior:

https://github.com/user-attachments/assets/cffbb3eb-aaa3-48cb-b102-36121693e5e8

dstillman:
I slowed down the animation and reduced it to 2 flashes:

https://github.com/user-attachments/assets/e2e88592-8647-4476-92a2-d73c7740a2a1

I didn't adjust the total timeout, so the animation is actually stopped halfway through the second run, but that's actually what Apple does, so… 🤷‍♂️ Charitably, it makes it all feel like more of a single animation than one animation finishing and then another being triggered. Looks OK to me, in any case.

abaevbog:
Ok, yeah, that makes sense. I tried this now and I fully agree that now the animation feels less seizure-inducing.

PR #4861, opened 2024-11-22
Update JSDoc for pluginAPI

windingwind:

PR #5028, opened 2025-02-05
Implement custom menu API

windingwind:
- [x] Implement custom menus API for:
  - [x] main window menubar
  - [x] main window toolbar menus
  - [x] main window library context menus
  - [x] reader window menubar
  - [x] item pane info box rows context menu
  - [x] notes pane menu
  - [x] sidenav locate menu
  - [x] main window tabbar context menu (awaits #4961)
- [x] add jsdoc (awaits #4861)
- [x] add tests


windingwind:
The major part is ready, only minor points awaiting other PRs. Should be ready for review, mainly the API format and some implementations details need to be discussed.

- Do we want to (automatically) add separators before/after the menus registered by one call, like a concept of *menu grouping*? Arguably we should do that, otherwise plugins usually do not handle the show/hide of their separators very well (multiple separators/no separators); Otherwise, do we want to restrict plugins to only be able to insert menus to several *slots* (menu groups)? Maybe @yexingsha will have some advices, thank you very much.
- To test the API, call:


```js
menuID = Zotero.MenuManager.registerMenu({
  menuID: "test",
  pluginID: "zotero@zotero.org",
  target: "itemPane/info/row",
  menus: [
    {
      menuType: "menuitem",
      l10nID: "menu-print",
      onShowing: (event, context) => {
        Zotero.debug("onShowing");
        Zotero.debug(Object.keys(context));
      },
      onCommand: (event, context) => {
        Zotero.debug("onCommand");
        Zotero.debug(Object.keys(context));
      },
    },
    {
      menuType: "submenu",
      l10nID: "menu-print",
      menus: [
        {
          menuType: "menuitem",
          l10nID: "menu-print",
          onShowing: (event, context) => {
            Zotero.debug("onShowing submenu");
            Zotero.debug(Object.keys(context));
          },
          onCommand: (event, context) => {
            Zotero.debug("onCommand submenu");
            Zotero.debug(Object.keys(context));
          },
        },
      ],
    },
  ],
});
```

All valid registration targets: https://github.com/windingwind/zotero/blob/b2f9a2ddcf885c49adfd66bd065ddd507eff2648/chrome/content/zotero/xpcom/pluginAPI/menuManager.js#L30-L60

Valid menu types: menuitem, separator, submenu.

yexingsha:
I think we definitely need some menu grouping, but I'm not sure if that prevents potential issue of too many plugin menuitems in one single menu. We can take an approach similar to firefox and chrome: 

- In each menu, each plugin can either:
	- Insert one menuitem for a single action;
	- Or Insert one submenu for multiple actions, with separator(s) if necessary.
- Each plugin menuitem leads with its plugin icon.
- The plugin menuitems are all inserted at the bottom of the target menu and grouped together, i.e. with a separator above them.

windingwind:
Thanks for the reply!

>  We can take an approach similar to firefox and chrome:

I am worried that it would be too strict and plugins will not want to using the API. Currently, we do not have a way to prevent plugins from directly inserting menus.

I think in most cases the menu being too long is only a problem with the item list context menu. Would it be an option to collapse the rest of custom menus in a submenu if the menu popup is longer than a certain threshold?

yexingsha:
> I am worried that it would be too strict and plugins will not want to use the API. Currently, we do not have a way to prevent plugins from directly inserting menus.

If that is the case, then grouping inserted menus is the second best option. Though I wonder:

> add separators before/after the menus registered by one call, like a concept of menu grouping

If multiple plugins insert items into one menu, does that mean A) all of them will be grouped together with one separator above them, or B) each plugin's items will be in its own group, with separators between them? In B), when there are lots of plugins with one item each, the separators can get tiresome. Also, both grouping approaches can be disrupted by the plugins' own separators. It's a bigger issue in A), where one plugin's separators can cause its items to appear to be grouped with other plugins' items. So we might need to restrict separators to submenus if we take one of these approaches.

For these reasons I'm still leaning toward the firefox approach, but maybe you are right that's too restrictive for plugin developers to adopt this API.

> I think in most cases the menu being too long is only a problem with the item list context menu. Would it be an option to collapse the rest of custom menus in a submenu if the menu popup is longer than a certain threshold?

Yes, that makes sense.

windingwind:
Thank you for the suggestions.

I think A) would be better, since most plugins only have 1-2 menus and having too many separators is definitely a bad idea...

I will try to implement this, maybe just for the item/collection list context menu. The API will ignore separators from plugin and simply grouping all menus in a group. When the popup becomes too long, some custom menus will be moved to a submenu. Do you think 80% of the screen height would be a good threshold for collapsing custom menus?

windingwind:
Regarding the behavior of collapsing custom menus, shall we 1) collapse all custom menus or 2) only the ones that will make the popup too long?

For 1), users will definitely blame us for making a similar stupid move on menu collapsing as the one in Windows 11 (https://answers.microsoft.com/en-us/windows/forum/all/restore-old-right-click-context-menu-in-windows-11/a62e797c-eaf3-411b-aeec-e460e6e5a82a);

But for 2), it can be sometimes unjustified for plugins whose name/id has disadvantage in alphabet sorting, which means their menus are always hidden in the submenu. Having menu position be dynamically dependent on the number of plugins enabled is also not very good.

windingwind:
Or, will it make more sense to provide a section in the settings window, where users can control (enable/disable/sort) the UI parts, including the menu, item pane section, custom columns, etc., So that both the plugin developers and users will be happy?

yexingsha:
Option A and setting the threshold to 80% screen height sounds good to me.

> Or, will it make more sense to provide a section in the settings window, where users can control (enable/disable/sort) the UI parts, including the menu, item pane section, custom columns, etc., So that both the plugin developers and users will be happy?

I think this plus menu collapsing behavior 2) is the best solution if we can implement it. If I understood it correctly, it will enable users to sort the "UI priority" of plugins, which controls the order of *all* UI elements of different plugins, right? Then I'm not sure if enable/disable is useful here, since I can't think of a use case where you would want to disable the UI elements of a plugin but not disable the plugin itself.

windingwind:
> Option A and setting the threshold to 80% screen height sounds good to me.

Thanks, I am implementing it this way.

>  If I understood it correctly, it will enable users to sort the "UI priority" of plugins, which controls the order of all UI elements of different plugins, right

Yes

> Then I'm not sure if enable/disable is useful here, since I can't think of a use case where you would want to disable the UI elements of a plugin but not disable the plugin itself.

A plugin can register multiple menus, and these menus should be able to be enabled/disabled separately. We should always encourage or even force plugins to provide ways to disable any UI elements they add to the interface. For example, the translate plugin can have `Translate Title` and `Translate Abstract` menus, maybe a user will only use one and want to hide another.

windingwind:
The items list/collections list's context menu is now auto-grouped and collapsed:

image


Run the code below to test. It will create 50 menu rows in the items list's context menu. The menu popup should take up around 80% height of the screen, and the rest menu rows will be collapsed under the submenu.

```js
menuID = Zotero.MenuManager.registerMenu({
  menuID: "test2",
  pluginID: "zotero@zotero.org",
  target: "main/library/item",
  menus: 
    Array.from({length: 50}, () => ({
      menuType: "menuitem",
      l10nID: "menu-print",
      onCommand: (event, context) => {
        Zotero.debug("onCommand");
        Zotero.debug(Object.keys(context));
      },
    })),
});
```

windingwind:
So, after some conflict resolution, I added tests that should mostly cover the API. Now all the planned features are implemented, and any future updates to support new menus/deprecating existing menus would be quite easy.

We can consider merging this.

windingwind:
rebased

PR #5029, opened 2025-02-05
Add JXA bridge and macOS Vision interface

AbeJellinek:
Try out text recognition like this:

```js
var { recognizeTextInImage } = ChromeUtils.importESModule('chrome://zotero/content/vision.mjs');
var imageBytes = await fetch('https://ajellinek.s3.us-east-1.amazonaws.com/2025_02_05_17_31_09.png')
	.then(r => r.arrayBuffer());
await recognizeTextInImage(imageBytes)
```

### DIFF
--- a/chrome/content/zotero/jxa.mjs
+++ b/chrome/content/zotero/jxa.mjs
@@ -1,0 +1,83 @@
+var { Subprocess } = ChromeUtils.import("resource://gre/modules/Subprocess.jsm");
+var { Zotero } = ChromeUtils.importESModule("chrome://zotero/content/zotero.mjs");
+
+/**
+ * Run a JavaScript function with osascript (JXA), giving it access to an Objective-C bridge.
+ *
+ * @param {Function} fn The function. The function will be called with the
+ * 		arguments passed in args, but it won't have access to closures or any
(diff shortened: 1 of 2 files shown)


mrtcode:
Absolutely amazing work!

However, there are a few issues with macOS Vision that slightly complicate its use:
- In your example image, the OCRed text is missing part of a paragraph. This issue can be reproduced as well by opening the image in Preview.app and selecting all the text. Interestingly, if I take a screenshot of the image, it is recognized correctly.
- Multi-column recognition sometimes joins lines from different columns. This means that if we want to use it for quality text layer creation, we need to perform post-processing (on the pdf-worker side) to regroup lines into paragraphs and columns.

I think we can merge this, and I can immediately start using it to improve PDF recognition for non-OCRed or poorly OCRed PDF files, where these issues don’t matter.

The next step is probably to support Windows and figure out what to do with Linux.

mrtcode:
So, I guess we can either try using a small OCR package or just wait for better times for Windows/Linux and use OCR only on macOS.

mrtcode:
So, once this is merged, I can start using it to recognize non-OCRed PDFs on macOS.

PR #5032, opened 2025-02-06
Local API: Fix out-of-repo styles, make `Zotero.Styles.init()` behave

### LINKED ISSUE #5030: Local API: file:// URL not working for `style`?
https://forums.zotero.org/discussion/105488/feature-request-url-scheme-to-get-citation-bibliography-for-an-item

Not tested

AbeJellinek:
This should've been a simple fix, but something absolutely bizarre was going on with `Zotero.Styles.init()` in tests. `init()` was running multiple times in parallel, even when I moved the `if (_initializationDeferred)` block all the way to the top of the method. That led to race condition issues like `_visibleStyles` getting frozen (line 89) before being sorted (line 85). I have absolutely no idea how that's even possible with JS promises. Maybe invisible multithreading due to the server?

But replacing the deferred promise-based serialization with `Zotero.Utilities.Internal.serial()` was enough to make it behave in all cases, and it simplified the code. `_initializationDeferred` wasn't accessible from the outside so this should be a seamless change for callers.

Fixes #5030

### DIFF
--- a/chrome/content/zotero/xpcom/localAPI/server_localAPI.js
+++ b/chrome/content/zotero/xpcom/localAPI/server_localAPI.js
@@ -87,6 +87,18 @@ const exportFormats = new Map([
  */
 class LocalAPIEndpoint {
 	async init(requestData) {
+		try {
+			return await this._initInternal(requestData);
+		}
+		catch (e) {
+			if (!(e instanceof BadRequestError)) {
+				throw e;
+			}
+			return this.makeResponse(400, 'text/plain', e.message);
+		}
+	}
+	
+	async _initInternal(requestData) {
 		if (!Zotero.Prefs.get('httpServer.localAPI.enabled')) {
 			return this.makeResponse(403, 'text/plain', 'Local API is not enabled');
 		}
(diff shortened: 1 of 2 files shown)


dstillman:
(Changed `method` → `coroutine` in commit message)

dstillman:
I had to revert the `Zotero.Styles.init()` commit in https://github.com/zotero/zotero/commit/9c6b3368f21ec77d35c6419382754f63c43341dd. `Styles.init()` awaits `schemaUpdatePromise`, but `schemaUpdatePromise` involves calling `Styles.init()`, so a `serial()` means it can hang forever if you cause it to be called before `schemaUpdatePromise` is resolved.

I'm not clear whether that breaks out-of-repo styles from the local API.

dstillman:
> I'm not clear whether that breaks out-of-repo styles from the local API.

Apparently it breaks all style installation, though I'm not totally sure why or how this was working before:

https://github.com/zotero/zotero/actions/runs/13364956220/job/37320779826

But I think it would've been running `install()` on any style short name (e.g., `chicago-note-bibliography`) rather than using an already-installed style.

Issue #1943, opened 2021-01-19
Allow dragging items to the trash


dstillman:
(And collections/searches once that's allowed)

PR #5033, opened 2025-02-07
Allow dragging items, collections, and searches into trash

abaevbog:
Fixes: #1943

https://github.com/user-attachments/assets/6d2a833c-fb59-4403-94d0-546c35ef4100

### DIFF
--- a/chrome/content/zotero/collectionTree.jsx
+++ b/chrome/content/zotero/collectionTree.jsx
@@ -1568,20 +1568,20 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 		// Directly on a row
 		if (orient === 0) {
 			if (!treeRow.editable) {
-				// Can drop collections, searches and items into trash
+				// Can drop collections, searches, and items into trash of their own library
 				if (treeRow.isTrash()) {
-					if (dataType == 'zotero/item') {
-						var items = Zotero.Items.get(data);
-						let allInSameLibrary = !items.find(item => item.libraryID != treeRow.ref.libraryID);
-						// Cannot drop items from another library
-						return allInSameLibrary;
+					let objects = [];
+					if (dataType === 'zotero/item') {
+						objects = Zotero.Items.get(data);
 					}
-					if (dataType == 'zotero/search' || dataType == 'zotero/collection') {
-						// Can only drop collections and searches into trash of their library
-						let id = data[0];
-						let obj = dataType == 'zotero/collection' ? Zotero.Collections.get(id) : Zotero.Searches.get(id);
-						return treeRow.ref.libraryID == obj.libraryID;
+					else if (dataType === 'zotero/collection') {
+						objects = Zotero.Collections.get(data);
 					}
+					else if (dataType === 'zotero/search') {
+						objects = Zotero.Searches.get(data);
+					}
+					let allInSameLibrary = objects.every(object => object.libraryID === treeRow.ref.libraryID);
+					return allInSameLibrary;
 				}
 				// Zotero.debug("Drop target not editable");
 				return false;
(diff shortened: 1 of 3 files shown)


chrome/content/zotero/collectionTree.jsx
```diff
@@ -1567,6 +1568,21 @@ var CollectionTree = class CollectionTree extends LibraryTree {
...
+						let allInSameLibrary = !items.find(item => item.libraryID != treeRow.ref.libraryID);
+						// Cannot drop items from another library
+						return allInSameLibrary;
+					}
+					if (dataType == 'zotero/search' || dataType == 'zotero/collection') {
+						// Can only drop collections and searches into trash of their library
+						let id = data[0];
+						let obj = dataType == 'zotero/collection' ? Zotero.Collections.get(id) : Zotero.Searches.get(id);
+						return treeRow.ref.libraryID == obj.libraryID;
+					}
```
dstillman, line 1584: Do these need to be different? Isn't it the same check?

chrome/content/zotero/collectionTree.jsx
```diff
@@ -1567,6 +1568,21 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 		// Directly on a row
 		if (orient === 0) {
 			if (!treeRow.editable) {
+				// Can drop collections, searches and items into trash
+				if (treeRow.isTrash()) {
+					if (dataType == 'zotero/item') {
+						var items = Zotero.Items.get(data);
```
dstillman, line 1574: `var` → `let`

chrome/content/zotero/collectionTree.jsx
```diff
@@ -1567,6 +1568,21 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 		// Directly on a row
 		if (orient === 0) {
 			if (!treeRow.editable) {
+				// Can drop collections, searches and items into trash
```
dstillman, line 1571: `searches and` → `searches, and` (house style)

chrome/content/zotero/collectionTree.jsx
```diff
@@ -1567,6 +1568,21 @@ var CollectionTree = class CollectionTree extends LibraryTree {
...
+				if (treeRow.isTrash()) {
+					if (dataType == 'zotero/item') {
+						var items = Zotero.Items.get(data);
+						let allInSameLibrary = !items.find(item => item.libraryID != treeRow.ref.libraryID);
```
dstillman, line 1575: `every(==)` instead of `!find(!=)`?

chrome/content/zotero/collectionTree.jsx
```diff
@@ -2096,6 +2112,27 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 			childFileAttachments: Zotero.Prefs.get('groups.copyChildFileAttachments'),
 			annotations: Zotero.Prefs.get('groups.copyAnnotations'),
 		};
+
+		// Droppping items, collections or searches into trash
```
dstillman, line 2116: - `ppp` → `pp`
- `collections or` → `collections, or`

chrome/content/zotero/collectionTree.jsx
```diff
@@ -2096,6 +2112,27 @@ var CollectionTree = class CollectionTree extends LibraryTree {
...
+			let objects = [];
+			if (dataType == 'zotero/collection') {
+				objects = await Zotero.Collections.getAsync(data);
+			}
+			else if (dataType == 'zotero/item') {
+				objects = await Zotero.Items.getAsync(data);
+			}
+			else if (dataType == 'zotero/search') {
+				objects = await Zotero.Searches.getAsync(data);
+			}
```
dstillman, line 2127: Order should be collection, search, item

chrome/content/zotero/collectionTree.jsx
abaevbog: Apologies, I'm not sure what exactly you mean... 

In this `canDropCheck` bit, `treeRow` is the trash row, and a given `id` from `data` is the id of whatever object is being dragged. So we want to make sure that the search or collection being dragged is allowed to be dropped into the trash of only its own library. And I'm checking against the `dataType` to decide if it's a `Zotero.Collection` or `Zotero.Search` where the actual object should be fetched from.

Alternatively, it could have been something like
```
if (dataType == 'zotero/collection') return Zotero.Collections.get(id).libraryID == treeRow.ref.libraryID;
if (dataType == 'zotero/search') return Zotero.Searches(id).libraryID == treeRow.ref.libraryID;
```

dstillman: I mean why do items and collections/searches use different logic?

abaevbog: Oh, gotcha. Yeah, it really is the same logic. I think I wanted to avoid too many `if-else` statements which did lead to this part looking off. I refactored it to just fetch the objects based on the `dataType` and compute `allInSameLibrary` for that array. That should make more sense now!

chrome/content/zotero/collectionTree.jsx
abaevbog: We are in the oxford comma camp, got it!

PR #5039, opened 2025-02-11
Apply macOS radio/checkbox normalization to import wizard

AbeJellinek:
And replace the left margin with `overflow: visible` so controls stay aligned.

### DIFF
--- a/scss/components/_import-wizard.scss
+++ b/scss/components/_import-wizard.scss
@@ -1,4 +1,6 @@
 .import-wizard {
+	@include macOS-normalize-controls;
+	
 	min-width: 600px;
 	min-height: 400px;
 	font-size: 12px;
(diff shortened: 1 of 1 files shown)


tnajdek (commented):
Thanks! I like the idea of a mixin for form/control normalization!  

There are a couple of issues I've noticed after normalization. First, the controls are cramped vertically, is this intentional or can we add some padding between the lines? Second, the checkboxes now have mismatched sizes, probably due to some other hacky or legacy CSS that can now be removed—both issues are visible in the screenshot below.

It's probably easier for me to take care of these, just please let me know if we want to add any vertical padding.

Screenshot 2025-02-12 at 12 15 05

From what I can see, the mismatched sizes are not due to a focus ring, but I haven't investigated the cause yet.

Issue #1706, opened 2019-06-26
Tag autocomplete in target selector


dstillman:
https://forums.zotero.org/discussion/78009/tag-auto-completion-in-zotero-connector-for-chrome

Needs a new connector-server endpoint that takes tag prefixes and returns possible matches in the selected library

adomasven:
Should probably do https://github.com/zotero/zotero/issues/1400 at the same time

HarshTrivedi:
Tag auto-completion in the connector would be very useful.

manbackintown:
This feature is also requested here zotero/zotero-connectors#310

Dadoopant:
Any update on this issue? Any way to use a third party code or add -on to achieve this outcome?

github-account1111:
@Dadoopant https://github.com/zotero/zotero-connectors/pull/529

PR #5040, opened 2025-02-11
Return tags together with collections in /connector/getSelectedCollection

abaevbog:
It does not sound quite semantically correct, since tags don't have much to do with a selected collection but it is the endpoint called to fetch the info from the client when the connector popup appears, and it already returns not just the selected collections but all collections from all editable groups, so it felt more appropriate than adding a new tags-specific endpoint. Perhaps, `/connector/getSelectedCollection` could be renamed for something more general at some point?

Addresses: #1706
Needed for tags autocomplete in the connector: https://github.com/zotero/zotero-connectors/pull/529


iagogv3:
> since tags don't have much to do with a selected collection

Actually, it would be great collections having associated tags: https://forums.zotero.org/discussion/104606/fr-allow-sub-collections-metadata#latest

PR #5049, opened 2025-02-17
Update eslint ecmaVersion to es2022

windingwind:
Current firefox version supports all features of es2022
Fix errors with features like WeakRef


windingwind:
See also: https://caniuse.com/?search=es2022

PR #5050, opened 2025-02-18
Fix broken browser window on print + remove many Mozilla files

### LINKED ISSUE #5048: Remove unnecessary Mozilla files
For Zotero 7 (?), we switched to bundling much more of Firefox, so there are surely lots of unnecessary files now. E.g., all of pdf.js, as mentioned in https://github.com/zotero/zotero/pull/5038#discussion_r1954643551.

The two omni.ja files are 11 MB and 34 MB, so the potential byte savings here are pretty limited, but multiplied by all downloads it might be significant. (Also not impossible that there could be some memory savings from avoiding loading things, though now that Mozilla is lazy-loading all modules that's probably less the case.) Deleting unneeded files is fun, in any case.

AbeJellinek:
This removes a bunch of Mozilla files that we didn't need - mostly actors, which don't usually break things when removed, but some modules, too. We now remove the entire `browser/chrome/browser/` directory! 362.7 MB -> 354.6 MB! 🎉🎉🎉 Yes, you read that right, a full 2.23% reduction.

But the biggest improvement here is that we no longer show a broken browser window when printing. Turns out all we needed was to set `window.browserDOMWindow` to something implementing `nsIBrowserDOMWindow`. The weird part is that the methods we implement are apparently never actually called(???) - we just need to convince Firefox that a browser window of some kind is already open.

Closes #5048

### DIFF
--- a/app/build.sh
+++ b/app/build.sh
@@ -232,6 +232,47 @@ elif [ $BUILD_LINUX == 1 ]; then
 fi
 set -e
 cd $omni_dir
+
+# Remove Firefox files that we don't need
+rm actors/AboutLogins{Parent,Child}.sys.mjs
+rm actors/AboutMessagePreview{Parent,Child}.sys.mjs
+rm actors/AboutNewTab{Parent,Child}.sys.mjs
+rm actors/AboutPocket{Parent,Child}.sys.mjs
+rm actors/AboutPrivateBrowsing{Parent,Child}.sys.mjs
+rm actors/AboutProtections{Parent,Child}.sys.mjs
+rm actors/AboutReader{Parent,Child}.sys.mjs
+rm actors/AboutTabCrashed{Parent,Child}.sys.mjs
+rm actors/AboutWelcome{Parent,Child}.sys.mjs
+rm actors/ASRouter{Parent,Child}.sys.mjs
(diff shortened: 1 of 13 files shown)


app/build.sh
```diff
@@ -376,12 +414,6 @@ replace_line 'handle: function bch_handle\(cmdLine\) {' 'handle: function bch_ha
   ' modules/BrowserContentHandler.sys.mjs
 export CALLDIR && perl -pi -e 'BEGIN { local $/; open $fh, "$ENV{CALLDIR}/assets/commandLineHandler.js"; $replacement = <$fh>; close $fh; } s/\/\/ TEST_OPTIONS_PLACEHOLDER/$replacement/' modules/BrowserContentHandler.sys.mjs
 
-# Don't try to initialize Places, since we've removed its files
-replace_line 'this._placesInitialized = true;' 'if (true) return; this._placesInitialized = true;' modules/BrowserGlue.sys.mjs
```
dstillman, line 380: Did you remove this on purpose? This avoids an error being logged at shutdown:

> JavaScript error: resource:///modules/BrowserGlue.sys.mjs, line 3581: Error: Failed to load resource://gre/modules/PlacesUtils.sys.mjs

Issue #2376, opened 2022-03-02
Warn if a library hasn't been synced in a while


dstillman:
Follow-up to #1429. In #2354 we added warnings for syncing not being set up and for auto-sync being disabled, but I [laid out some additional situations](https://github.com/zotero/zotero/issues/1429#issuecomment-748949566):

> 4. Sync set up with auto-sync enabled but all libraries disabled in the libraries-to-sync dialog (
https://github.com/zotero/zotero/issues/1913 but for auto-sync)
> 5. Sync set up with auto-sync enabled but at least one enabled library that hasn't been synced in a while due to a sync error
> 6. Same as 5 but for file sync errors (including quota errors)

dstillman:
We really need to figure out (6) in particular. The quota icon just isn't prominent enough, so a lot of people end up losing files when they switch to new computers:

https://forums.zotero.org/discussion/99823/sync-issue-lack-of-every-pdf-file

Seems more important to handle the default case where file syncing using Zotero Storage is enabled but errors have occurred, since if you've gone out of your way to disable file syncing that's sort of on you.

This is a little tricky because a quota error could also be for a group library owned by someone else, which maybe doesn't merit the same level of bright yellow banner?

Quota errors are detected here:

https://github.com/zotero/zotero/blob/0d09618865a938721ee35b5c91f41acd8f95cbb1/chrome/content/zotero/xpcom/storage/zfs.js#L631-L637

With user/group testing available from `item.library.libraryType`, as in `getQuotaError()`:

https://github.com/zotero/zotero/blob/0d09618865a938721ee35b5c91f41acd8f95cbb1/chrome/content/zotero/xpcom/storage/zfs.js#L1057-L1077

I think we don't currently have group owner information in the client, but I guess we could add `Zotero-Storage-Owner: [userID]` to the 413 response so that clients could determine whether the current user's quota is the relevant one.

dstillman:
(I can adjust your quota down to force a quota error if you want that for testing, or you can fake the 413 [here](https://github.com/zotero/zotero/blob/0d09618865a938721ee35b5c91f41acd8f95cbb1/chrome/content/zotero/xpcom/storage/zfs.js#L444-L454).)

AbeJellinek:
Sure, can you lower the quota for `AbeJellinek_dev`?

AbeJellinek:
Not seeing any errors yet...

PR #5053, opened 2025-02-21
Show banner when sync fails due to quota error

AbeJellinek:
This handles (6) in #2376.

Also:
- Move strings to Fluent
- Fix overflow when window is narrow

Simulate a quota error locally for testing:
```diff
--- a/chrome/content/zotero/xpcom/storage/zfs.js
+++ b/chrome/content/zotero/xpcom/storage/zfs.js
@@ -449,6 +449,21 @@ Zotero.Sync.Storage.Mode.ZFS.prototype = {
 						debug: true
 					}
 				);
+				// Don't merge me! Simulate a quota error for testing:
+				Object.defineProperty(req, 'status', { value: 413 });
+				let getResponseHeader = ((original, obj) => ((header) => {
+					if (header === 'Zotero-Storage-Usage') {
+						return 999;
+					}
+					if (header === 'Zotero-Storage-Quota') {
+						return 1000;
+					}
+					if (header === 'Zotero-Storage-UserID') {
+						return Zotero.Users.getCurrentUserID();
+					}
+					return original.call(obj, header);
+				}))(req.getResponseHeader, req);
+				Object.defineProperty(req, 'getResponseHeader', { value: getResponseHeader });
 			}
 			catch (e) {
 				if (e instanceof Zotero.HTTP.UnexpectedStatusException) {
```

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -2808,7 +2808,8 @@ var ZoteroPane = new function()
 				'api-key');
 		}
 
-		// If both reminders are disabled, we don't need an observer
+		// If both reminders that show on item modification are disabled,
+		// we don't need an observer
 		if (!Zotero.Prefs.get('sync.reminder.setUp.enabled')
 				&& !Zotero.Prefs.get('sync.reminder.autoSync.enabled')) {
 			if (_syncRemindersObserverID) {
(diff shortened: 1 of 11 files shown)


chrome/content/zotero/zoteroPane.js
```diff
@@ -2839,16 +2830,17 @@ var ZoteroPane = new function()
...
 			{
 				notify: (event) => {
-					if (!eventTypes.includes(event)) {
+					if (!eventTypes.has(event)) {
 						return;
 					}
 					setTimeout(() => {
 						this.showSetUpSyncReminder();
 						this.showAutoSyncReminder();
+						this.showQuotaErrorReminder();
```
dstillman, line 2843: I don't think this makes sense here. We show the `setUp` and `autoSync` reminders on item modification because it means you're making changes in your library, but that's not really an appropriate time to show the quota warning. I think we can just show it if 1) you hit a quota error and 2) you haven't said not to show the banner.

(Honestly, we should probably do the same for most unexpected, non-5xx sync errors — the idea being that a sync error should be dealt with and a banner is the way to communicate that. The icon then is just there so that there's still an indication that something is wrong if you dismiss the banner.)

So I think we could take out the `lastErrors` tracking and just show this immediately if not explicitly hidden.

chrome/locale/en-US/zotero/zotero.ftl
```diff
@@ -773,3 +774,16 @@ mac-word-plugin-install-remind-later-button =
...
 mac-word-plugin-install-dont-ask-again-button =
     .label = { general-dont-ask-again }
+
+sync-reminder-setUp-message = Back up your library with { -app-name } syncing.
+sync-reminder-setUp-action = Set Up Syncing
+sync-reminder-autoSync-message = { -app-name } hasn’t synced in a while. Do you want to enable automatic syncing?
+sync-reminder-autoSync-action = { general-enable }
+sync-reminder-quotaError-message = { $libraryType ->
+   *[user] You have reached your Zotero File Storage quota. Files you add may not sync.
+   [group] The owner of “{ $libraryName }” has reached their Zotero File Storage quota. Files in the group may not sync.
```
dstillman, line 784: → `Zotero Storage`

chrome/content/zotero/zoteroPane.js
```diff
@@ -2945,6 +2981,14 @@ var ZoteroPane = new function()
 				case 'autoSync':
 					Zotero.Prefs.set(`sync.autoSync`, true);
 					break;
+				case 'quotaError':
+					if (options.library instanceof Zotero.Group) {
+						this.collectionsView.selectLibrary(options.library.libraryID);
+					}
+					else {
+						Zotero.launchURL('https://www.zotero.org/settings/storage');
```
dstillman, line 2989: We should put this in config.js, since we'll probably use it elsewhere.

chrome/content/zotero/zoteroPane.js
```diff
@@ -2945,6 +2981,14 @@ var ZoteroPane = new function()
 				case 'autoSync':
 					Zotero.Prefs.set(`sync.autoSync`, true);
 					break;
+				case 'quotaError':
+					if (options.library instanceof Zotero.Group) {
+						this.collectionsView.selectLibrary(options.library.libraryID);
```
dstillman, line 2986: I don't think there's much point to focusing the group. We should probably provide the owner `userID` in the server response so that the client could know if it's the current user. If it was, it could take you to your account settings, and otherwise it should probably just use the space to show a longer message that says to contact the group owner.

PR #5067, opened 2025-02-26
Fix attachment preview background

windingwind:
regression from 9c41874


windingwind:
Before:

image

After:

image

PR #5071, opened 2025-02-27
Convert `customThemes` from a pref to a sync setting

tnajdek:
Enables syncing custom themes, allowing the same custom themes to be used in the web library and mobile apps.  I've also included code to migrate existing pref to a synced setting.

I don't think we should sync the actual selected theme (and this PR doesn't do that)—users might prefer different themes on different devices—but it would be nice if the same custom themes were available across devices.

Depends on https://github.com/zotero/dataserver/pull/168

### DIFF
--- a/chrome/content/zotero/xpcom/reader.js
+++ b/chrome/content/zotero/xpcom/reader.js
@@ -150,6 +150,8 @@ class ReaderInstance {
 		// Set `ReaderTab` title as fast as possible
 		this.updateTitle();
 
+		await Zotero.SyncedSettings.loadAll(Zotero.Libraries.userLibraryID);
+
 		let data = await this._getData();
 		let annotationItems = this._item.getAnnotations();
 		let annotations = (await Promise.all(annotationItems.map(x => this._getAnnotation(x)))).filter(x => x);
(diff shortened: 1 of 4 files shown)


chrome/content/zotero/xpcom/reader.js
```diff
@@ -1986,6 +1987,17 @@ class Reader {
 				}
 			}
 		}
+		else if (type === 'setting') {
+			let id = ids[0];
+			let newCustomThemes = Zotero.SyncedSettings.get(Zotero.Libraries.userLibraryID, 'readerCustomThemes') ?? [];
```
dstillman, line 1992: Might as well move this within the conditional

chrome/content/zotero/xpcom/reader.js
```diff
@@ -548,7 +550,12 @@ class ReaderInstance {
 				}
 			},
 			onSaveCustomThemes: (customThemes) => {
-				Zotero.Prefs.set('reader.customThemes', JSON.stringify(customThemes));
+				if (customThemes?.length) {
+					Zotero.SyncedSettings.set(Zotero.Libraries.userLibraryID, 'readerCustomThemes', customThemes);
+				}
+				else {
+					Zotero.SyncedSettings.clear(Zotero.Libraries.userLibraryID, 'readerCustomThemes');
+				}
```
dstillman, line 558: `set` and `clear` are async, so we should technically make this function async both here and in the reader. Should at least do it here for now.

chrome/content/zotero/xpcom/reader.js
tnajdek: good point, fixed!

chrome/content/zotero/xpcom/reader.js
tnajdek: Sure thing, converted here for now.

tnajdek:
@dstillman Thanks for the review! I’ve addressed the comments and rebased on top of main.

PR #5079, opened 2025-03-05
Scaffold: Add item pane-like preview of translated items

AbeJellinek:
https://github.com/user-attachments/assets/5649e52a-0e40-4ac9-9aca-d8c4531f31a0

With a diff showing changes made during cleaning.


dstillman:
This is awesome!

I'm not sure I understand the diff, though, and I'm pretty sure no one else will:

Screenshot 2025-03-27 at 3 53 44 AM

`repository` is a base-mapped field of `publisher`, and that shows up, while `archiveID` is `number`, and that doesn't…

PR #5089, opened 2025-03-11
Fix update checker _checker.stopCurrentCheck is not function

northword:
Zotero 7.1-beta12, when closing the ‘check update’ window, I get the following error: 

PixPin_2025-03-11_11-38-36

I have not tested it and am not sure if the modifications make sense.

may ref: https://searchfox.org/mozilla-central/source/toolkit/mozapps/update/UpdateService.sys.mjs#46

### DIFF
--- a/chrome/content/zotero/update/updates.js
+++ b/chrome/content/zotero/update/updates.js
@@ -644,6 +644,7 @@ var gCheckingPage = {
 	 * this so we can cancel the update check if the user closes the window.
 	 */
 	_checker: null,
+	_checkId: null,
 
 	/**
 	 * Initialize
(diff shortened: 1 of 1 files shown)


dstillman:
This seems like an improvement, though testing in a source build with the debugger open to [force the hang](https://forums.zotero.org/discussion/122530/jsdebugger-flag-breaks-app-update), subsequent checks still don't work for me (i.e., fail the way they're supposed to fail in a source build) after closing the update window and the debugger. Not sure if that's due to some other state that we're not cleaning up or because the debugger breaks things more permanently. In any case, better than an error. Thanks!

PR #5094, opened 2025-03-11
Draggable item pane sections

windingwind:
- [x] Fix item pane sections init/uninit, since they will be connected/disconnected to/from DOM for multiple times during dragging/reordering
- [x] Draggable item pane section button
- [x] Item pane section button context menu for reordering
- [x] Add new parameter to item pane custom section API (options->sidenav->orderable). Non-orderable sections always stay at the end of the item pane. This is to avoid unnecessarily caching temporary custom sections' order.
- [x] Add tests for item pane reordering

https://github.com/user-attachments/assets/c7cce9d7-dfd4-49ff-afc1-4566b4130b9d


yexingsha:
This is great! My thoughts so far:

- It seems that after reordering, the item pane always scrolls back up top, then scrolls down to the section that was just reordered. This quick upward then downward scroll can be disorienting. Also, I'm not sure it's even necessary to scroll to the reordered section, since people are likely to do a bunch of reordering at once to achieve the section order they prefer. I think the ideal behavior is:
	- Maintain the item pane's scroll position as much as possible;
	- Render the reordering of section positions if possible. If not, just display the updated section positions.
- Let's use the accent color for the drop indicator.
- The `Locate` button is now colored wrong; it should be `fill-secondary`.
- I'm not sure about this, but do you think we should make the sections themselves draggable too? I think this would be useful to people who have trouble remembering which icon represents which section. Perhaps the icon area of each section can serve as the drag handle, and the cursor changes to `move` when hovering over it to make this feature more discoverable. And we can add "move section up/down" options to the section header context menu as well.

windingwind:
Thanks for the feedback!

> Maintain the item pane's scroll position as much as possible;

Would that mean maintaining the absolute scroll distance to the top of the pane's container or keeping a certain pane (reordered pane?) at the same position in view?

> Render the reordering of section positions if possible.

Like highlighting the section header?

> I'm not sure about this, but do you think we should make the sections themselves draggable too? I think this would be useful to people who have trouble remembering which icon represents which section. Perhaps the icon area of each section can serve as the drag handle, and the cursor changes to move when hovering over it to make this feature more discoverable.

Usually sections are pretty long vertically (e.g. the info, the attachments, the notes), holding the mouse for quite a while to drag across these long sections might not be a very good experience. Maybe we can collapse all sections once dragging starts, but that will change the vertical position of all sections (since they are collapsed) suddenly. If you by accident triggered dragging (for example, the abstract pane) and release the mouse immediately without moving, possibly the abstract pane will go to the end of the item pane (since everything is collapsed, the mouse position originally has the dragged section header will now be after the last section). This could be quite annoying or even be regarded as a bug...

> And we can add "move section up/down" options to the section header context menu as well.

Yes that's a good idea. We can do that.

yexingsha:
> Would that mean maintaining the absolute scroll distance to the top of the pane's container or keeping a certain pane (reordered pane?) at the same position in view?

I'm thinking of maintaining the absolute scroll distance, since I feel like when I reorder a section, my frame of reference is more often the entire item pane rather than the section being reordered. For example, if I move abstract down to the very bottom because I don't care about it, I will not want the scroll position to follow it all the way down.

> Like highlighting the section header?

I was more thinking of animating the swapping of section positioning a little, but perhaps that's not really necessary. Let's hold off on that and see how maintaining absolute scroll distance feel like without transition.

Re: issues with dragging sections themselves, I agree these are very good points. There could be ways to alleviate them (e.g. delay entering drag mode, fine tune the drag distance threshold for changing ordering, etc.) but they might not be worth it. So let's hold off on that too.

windingwind:
Thank you, all adressed.

image

yexingsha:
I feel like no scroll after reordering works fine, what do you think?

Another small issue I found is that when dragging sections to the very top/bottom, it doesn't work when the cursor moves up into the tab bar or down below the divider, which can feel like a bug because the drop indicator is still visible. Can we either make drag-and-drop still work when the cursor moves out of the destination area, or make the the drop indicator disappear when that happens?

Also, still need to add "move section up/down" into the section header context menu.

windingwind:
Done:

image

And the section header context menu is also added.

BTW: we'll need icons for the context menu for Windows and Linux

Thank you!

yexingsha:
I think simple up and down arrows should work: 
down
up

yexingsha:
Looks good to me!

dstillman:
We should probably have a Reset Section Order option when different from the default that would clear the pref.

test/content/support.js
```diff
@@ -353,6 +353,17 @@ function waitForNotifierEvent(event, type) {
 	return deferred.promise;
 }
 
+async function waitForPrefsChange(key, global) {
```
dstillman, line 356: → `waitForPrefChange`

dstillman:
Though I'm not sure how we're handling the order of plugin-created buttons. Since there's no inherent order for those, Reset Section Order shouldn't change those, only the default set.

windingwind:
> Though I'm not sure how we're handling the order of plugin-created buttons. Since there's no inherent order for those, Reset Section Order shouldn't change those, only the default set.

We are currently recording plugins‘ order as well. For the reset, maybe simply 1) remove any plugins‘ id that is no longer registered 2) remove builtin id and add them back to the start of the list in the default order. So that the plugins‘ order are kept but moved to the end.

windingwind:
> Reset Section Order

@yexingsha hi yexing, we need a new icon here, thank you very much.

yexingsha:
Reset icon: 
reset

windingwind:
Thanks!

image

Currently:
1. Clean up persisted order when reordering happens
2. Add a menu for reset order if the current order of default buttons is changed. Plugins' buttons will be kept at their original position if possible.

Should we move plugins' buttons to the end when resetting the order or only show the reset menu on default buttons?

yexingsha:
I think we should move the plugin buttons to the end when resetting the order, since that's what the default order looks like.

dstillman:
I might put Pin Section at the top above a separator? That seems like the option you'd want to use the most, followed by collapse/expand. The move/reset options are really just there for accessibility/discoverability — people should drag if they can.

windingwind:
All adressed:

image

dstillman:
There should be a separator below "Pin Section" in both context menus.

windingwind:
Thanks, done:

image

Issue #5091, opened 2025-03-11
Citation dialog: Locator menu isn't aligned with text field on macOS, label is too low


AbeJellinek:
![Image](https://github.com/user-attachments/assets/26d1418c-b0f0-4304-8b7d-9010ba44820d)

Text field is below menu, and the label ("Page") on the menu is too far down. (The latter happens when you make a native `<select>` or `<menulist>` taller than normal. We fixed it for menulists with `line-height: normal`, but I don't know if there's a workaround for `<select>`.)

abaevbog:
Good catch! Indeed, I increased the height so it roughly matches that of the inputs. And it is messing up the positioning in such a way that adding a small margin to make it align with the input affects the spacing of everything else. So that's something to explore.

At least the documentation for `<select>` is honest. From https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#styling_with_css: "The `<select>` element is notoriously difficult to style productively with CSS. ... it is hard to do things like line different types of form element up with one another in a column."

PR #5096, opened 2025-03-11
Citation dialog: move locator label lower

abaevbog:
So it aligns with inputs a bit better. Positioning is tricky with native select elements but with a bit extra margins, it looks better I think.

Fixes: #5091

Before:
Screenshot 2025-03-11 at 4 14 42 PM

After:
Screenshot 2025-03-11 at 4 13 58 PM

### DIFF
--- a/scss/components/_citationDialog.scss
+++ b/scss/components/_citationDialog.scss
@@ -623,7 +623,10 @@
 							height: 22px;
 							// <select> with increased height is hard to position but this helps
 							// to align it with <input> better
-							margin-bottom: -2px;
+							margin-bottom: -3px;
+							// push label text a bit higher 
+							padding-top: 0;
+							padding-bottom: 3px;
 						}
 						@media (-moz-platform: windows) {
 							border: 1px solid var(--color-border);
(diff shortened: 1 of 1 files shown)


dstillman:
For me, on non-Retina, `-3px` gets it closer to aligned:

Non-Retina `-3px`:

Screenshot 2025-03-12 at 12 59 53 AM

Compared to `-2px`:

Screenshot 2025-03-12 at 1 04 58 AM

On Retina, it's a wash — it's 1px off on either the top or the bottom:

`-3px`:

retina-3px

`-2px`:

retina-2px

(A separate problem is that the text inside our Mac selects is vertically misaligned — "Page" is too low here.)

abaevbog:
Ok, yeah, I see it. How is it looking now? I moved the `<select>` another 1px lower and tweaked padding to push the text inside of it a bit higher. 

Screenshot 2025-03-12 at 4 42 44 PM

dstillman:
Sorry, the vertical alignment of the text was just a side note and shouldn't be fixed here — that's a general problem in all our native selects on macOS (e.g., in the prefs).

abaevbog:
Gotcha! I dropped the adjustment of `<label>`'s positioning from here. This is what it looks like on my end with just `margin-bottom: -3px;`.

Screenshot 2025-03-12 at 10 00 12 PM

PR #5087, opened 2025-03-10
Fix space before ellipsis in editable-text

AbeJellinek:
Before:
Screenshot 2025-03-10 at 4 38 02 PM

After:
Screenshot 2025-03-10 at 4 38 11 PM

Fixes #4716

PR #5103, opened 2025-03-12
Make the info box ellipsis behave

AbeJellinek:
https://github.com/zotero/zotero/pull/5087#issuecomment-2717115891

https://github.com/user-attachments/assets/4e03bb27-b843-46e0-b27f-2ccde4228f75

There's still a little extra space after the *comma* when the field is clipped on a wide character. But after a lot of trial and error, I think this is the best we're going to get.


dstillman:
Now I'm missing the comma altogether for an item:

Screenshot 2025-03-12 at 9 56 09 PM

PR #5105, opened 2025-03-12
citation dialog: deduplicate open items

abaevbog:
So that if there are multiple tabs opened for the same top level item, it appears only once in search results.

### DIFF
--- a/chrome/content/zotero/integration/citationDialog/searchHandler.mjs
+++ b/chrome/content/zotero/integration/citationDialog/searchHandler.mjs
@@ -248,7 +248,9 @@ export class CitationDialogSearchHandler {
 			}
 			return Zotero.Cite.getItem(itemID);
 		});
-		return items;
+		// Return deduplicated items since there may be multiple tabs opened per the same
+		// top-levle item (duplicate tabs or a multiple attachments belonging to the same item)
+		return [...new Set(items)];
 	}
 
 	_getSelectedLibraryItems() {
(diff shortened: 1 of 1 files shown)


abaevbog:
To handle situations like this, where i opened two duplicate tabs for the PDF, as well as the snapshot, all for the same item.

Before:
Screenshot 2025-03-12 at 2 45 41 PM

After:

Screenshot 2025-03-12 at 2 48 56 PM

chrome/content/zotero/integration/citationDialog/searchHandler.mjs
```diff
@@ -248,7 +248,18 @@ export class CitationDialogSearchHandler {
...
+		// item (duplicate tabs or a PDF and a snapshot belonging to the same item)
+		let uniqueItems = [];
+		let seenItemIDs = new Set();
+		for (let item of items) {
+			if (!seenItemIDs.has(item.id)) {
+				uniqueItems.push(item);
+				seenItemIDs.add(item.id);
+			}
+		}
+		return uniqueItems;
```
dstillman, line 262: I would think `return [...new Set(items)];` should be sufficient.

abaevbog: Oh, that's completely right! Because if there are two duplicate items in the array, it's just two pointers to the same object, as opposed to having two different objects. I re-pushed with this simpler approach.

PR #5112, opened 2025-03-13
Windows: Reduce input style specificity to fix directory fields

### LINKED ISSUE #4886: Windows: Data Directory file chooser icon overlaps text
image

Likely too-specific/`!important` Windows text input styles, but I haven't looked closely.

AbeJellinek:
Everything that isn't an element selector is in a `:where()` (zero specificity) block now, so the overall specificity is that of an element selector, allowing it to be overridden by class selectors. Safer than making the whole thing zero specificity.

Fixes #4886

### DIFF
--- a/scss/win/components/_input.scss
+++ b/scss/win/components/_input.scss
@@ -13,6 +13,7 @@
 @mixin windows-input-active {
 	outline: none;
 	background-clip: border-box, padding-box;
+	background-repeat: no-repeat;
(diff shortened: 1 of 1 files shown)


dstillman:
I'm seeing a regression in the settings search bar after this.

Before:

Screenshot 2025-03-27 at 3 31 04 AM

After:

Screenshot 2025-03-27 at 3 32 21 AM

PR #5120, opened 2025-03-14
Update metadata

AbeJellinek:
Supersedes #1582 (because I can't force-push to mrtcode/zotero). This is basically @mrtcode's work plus my tweaks from #1582, squashed to make rebasing less terrible, and updated to work on Z7. Ready for design review and bug testing.


dstillman:
We have an older version of Diff Match and Patch from https://github.com/zotero/zotero/commit/a3654652b21071955ae01ba8fcdc38765e1740a4. Let's just update that one. (We can check whether it still works when we get back to https://github.com/zotero/zotero/pull/3468.)

dstillman:
Just starting to look at this, but just from the first 30 seconds:

Screenshot 2025-03-18 at 2 47 33 PM

1. This Abstract diff is…not usable. Do we have any control over this?
2. Done vs. Apply All is definitely not going to work. I'll let @yexingsha make some suggestions here, but we need to structure the expected interaction flow better, and I think we also want to optimize for the almost certainly default case of wanting to select all items in your library, selecting Update (or its better future name), and having it update all items to whatever you would get with the magic wand, completely overwriting anything you have, without thinking about it. We want to give the people who want more fine-grained control more control and try to avoid data catastrophes (wiping out custom plugin data in Extra? deleting abstracts that were added manually?), but most people are just going to want an automatic process.

dstillman:
((2) probably also has implications for rate limiting, progress display, and maybe even caching. I would expect people to repeatedly select all items and re-run this.)

yexingsha:
> "Update" feels too short in the context menu, but "Update Metadata" is too technical and "Update Item" is weird (and goes against our resolution to stop putting "Item" in every menu item). We could do something a little longer but more descriptive, like "Search for Updates"? Or "Find Updates" to match "Find Full Text"?

I think we can call it "Update Info", since "Info" is the name for the metadata section in ItemPane.

I agree that we should optimize for the use case of "updating multiple items without thinking about it". When there are multiple items being updated, all details should be collapsed by default; when there is only one item being updated, then we can expand it by default. The heuristics we use to deselect certain updates does complicate things a bit, though I can see the utility in e.g. not applying destructive changes. I think we should expose these heuristics as options, such as "update everything", "do not delete existing fields", etc.

I made some designs in [figma](https://www.figma.com/design/y8aZzEDwsxsgVh7Zui63to/Zotero?node-id=12828-138820&t=sK8sHnrBU6xLYekG-1) exploring how best to represent the updates when all details are collapsed. This is the approach I'm most happy with:

in-progress

finished

expanded

- The main action buttons "Cancel" and "Apply Updates" are moved to the bottom right. I removed the buttons for each individual item since they seem redundant. I added "Expand All" and "Collapse All" to the bottom left for easier individual selections.
- In the default collapsed state, you can choose to read the list of updates and expand items to modify them, or simply ignore them and apply whatever updates your chose option provides. Some further considerations:
    - How to display fields where updates are available but not selected. Here I chose to display them with a strikethrough.
    - Whether to include the second line of details when there is no update. I chose to put "No Update" there for now.
    - When there are a lot of updated fields, whether to wrap the list into multiple lines. I guess cutting it off would be too arbitrary?
- The expanded state is similar to the current implementation, with a few changes:
    - I think right now it's not clear enough that clicking on a diff will select/unselect it, so I added checkmarks before each field.
    - Let's restrict all diffs to one line, since you can always expand them.
    - Changed background color to `tag-red` and `accent-green` at 50% opacity.

yexingsha:
> "Update" setting could be hard to follow sometimes. For instance, if the item type changes, we might *need* to clear some fields.

I think we need to describe the update heuristic as much as we can. Something like "Default" would be too vague and confusing. What about "Avoid Clearing Existing Fields"?

> In this design, do you still expand the diff by clicking the ellipsis? Now that we have a checkbox for each field, maybe we can remove the click-to-toggle functionality and make a click anywhere on the diff expand it instead?

Yes, that's a good idea.

> Should items have their own checkboxes so you can quickly choose not to apply any of the updates it's suggested? Overriding and disabling the field-specific checkboxes?

That is a good point. I tried it and found that checkboxes don't go well with the item type icons, so I just got rid of the icons:

in-progress

finished

expanded

This also means items without update will stand out more. What do you think?

dstillman:
For controlling the clearing of existing fields, a checkbox might be more straightforward than radio buttons (and we might have other options at some point)?

yexingsha:
> For controlling the clearing of existing fields, a checkbox might be more straightforward than radio buttons (and we might have other options at some point)?

Then I guess it would be a checkbox labeled "Avoid Clearing Existing Fields" that's checked by default. Though maybe it gives too little warning about how dangerous unchecking it would be?

dstillman:
The URL value is appearing a line down from the label:

Screenshot 2025-04-17 at 11 45 35 PM

And, actually, I'm seeing it for some other fields too:

Screenshot 2025-04-17 at 11 52 16 PM

I'm also not totally clear what's going on with the truncation. This is what I get if I click on the above URL — it definitely had space:

Screenshot 2025-04-17 at 11 45 43 PM

dstillman:
Closing the window with the red stoplight button on macOS doesn't cancel the process. The icon in the top right keeps spinning, and if I then try to update info for another item, it reopens the pane with all the previous items that I cancelled.

dstillman:
> Maybe we should keep Cancel around, so closing with the red stoplight won't clear the queue, but reopening it and then explicitly clicking Cancel will.

OK, yes, that would work. I think we definitely want a get-me-out-of-here button for an operation like this.

Issue #5119, opened 2025-03-14
Citation dialog doesn't persist screen position


dstillman:
It opens in the center of the screen each time.

Probably also causing https://forums.zotero.org/discussion/122594/zotero-beta-about-the-new-citation-dialog-window ?

PR #5122, opened 2025-03-14
citation dialog: persist screen position

abaevbog:
Do not explicitly move the window to the center of the screen.

Fixes: #5119

With this, the dialog also appears on an external monitor if it was last closed there for me (https://forums.zotero.org/discussion/122594/zotero-beta-about-the-new-citation-dialog-window). There was some [special handling in QF](https://github.com/zotero/zotero/blob/4915f1fadf773f855507bd801137e87770583ec6/chrome/content/zotero/integration/quickFormat.js#L273) to move the dialog into the center if it is not properly visible - but such handling is not present in any other dialog, and I have no been able to get the dialog to appear without being properly visible. So I assume it's just no longer needed.

### DIFF
--- a/chrome/content/zotero/integration/citationDialog.js
+++ b/chrome/content/zotero/integration/citationDialog.js
@@ -74,13 +74,6 @@ async function onLoad() {
 	// keypress from the top-most row in the items table
 	doc.addEventListener("keydown", event => KeyboardHandler.captureKeydown(event), true);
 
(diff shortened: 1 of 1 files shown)


dstillman:
Without this code, for the initial open, it seems like it's at least close to centering it with `centerscreen` in List Mode, but Library Mode is too far down. I guess that's something with how the window is resized after the initial render? Maybe there's a way to improve that… Ideally we would center it properly only when not using a persisted position.

Issue #5118, opened 2025-03-14
Citation dialog: Open popup on Down Arrow if bubble has focus


dstillman:
https://forums.zotero.org/discussion/comment/487524/#Comment_487524

I personally find Left-Down faster and easier than Left-Space. If we think this is really going to mess people up as as they try to navigate to the rest of the dialog, we could not do it, but I don't think it's that confusing to have different behavior on the bubbles. And this obviously helps with continuity for existing users. (If we do this, we might as well support Cmd/Ctrl-Down as well, which will help people like adamsmith there still using the pre-Z7 shortcut.)

abaevbog:
The only reason why arrow down as an option to open popups had to be removed was to enable navigation like this between bubble rows:

https://github.com/user-attachments/assets/1beaef71-a9cb-4f33-9bac-5f0a9a49d97c

If this is causing inconvenience, would it be correct to say that this navigation between bubble rows was not worth it? If we don't need it, we can just let arrow down open the itemDetails popup again. I suspect, few people end up with citations big enough for navigating between bubble rows to be necessary.

abaevbog:
I brought back arrowDown on bubble behavior of opening the itemDetails popup in the linked PR and removed the navigation between bubble rows.

I think I might like it better as well. Using arrowDown to open the popup has a benefit of allowing us to use arrowUp to close the popup. And pressing arrowDown + arrowUp is way more convenient for a quick peak at the cited item's details than space + Enter/Escape just because of where the keys are positioned on the keyboard.

PR #5123, opened 2025-03-14
citation dialog: restore arrowDown on bubbles opening details popup

abaevbog:
- restore earlier behavior of arrowDown on a bubble opening itemDetails popup
- arrowUp from inside itemDetails closes the popup (except when locator dropdown is focused)
- remove arrow navigation between rows of bubbles on arrowUp/down
- remove no longer needed edge case of arrowDown handling on the last row of bubbles

Fixes: #5118

### DIFF
--- a/chrome/content/zotero/elements/bubbleInput.js
+++ b/chrome/content/zotero/elements/bubbleInput.js
@@ -288,9 +288,9 @@
 				input.value += event.key;
 				input.dispatchEvent(new Event('input', { bubbles: true }));
 			}
-			// Space on bubble simulates a click
-			if (event.key == " ") {
-				event.target.click();
+			// Space or arrowDown on a bubble open item details popup
+			if (event.key == " " || event.key == "ArrowDown") {
+				Utils.notifyDialog("show-details-popup", { dialogReferenceID: bubble.getAttribute("dialogReferenceID") });
 				event.preventDefault();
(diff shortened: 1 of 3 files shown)


dstillman:
I don't think Up Arrow needs to or should close the popup. We have a prominent, nicely styled Done button. People can press Return/Enter.

*Maybe* we could special-case Up Arrow in an empty locator value text field, just so you can press Up immediately after Down. But otherwise it interferes with native text field behavior, and it's also just sort of strange. Did your changes save? Who knows?

dstillman:
From https://github.com/zotero/zotero/issues/5118#issuecomment-2725467158:

> And pressing arrowDown + arrowUp is way more convenient for a quick peak at the cited item's details

If we wanted to preserve this, we could maybe say that Up works whenever the cursor is at the beginning of the locator value field, even if it had a value. So if no value, you could press it once, and if it did, you'd have to press it twice (one to go back to the beginning of the line (at least on macOS) and again to close the dialog). But I don't think we should interfere with native field behavior. And this would still have the problem of making it unclear whether changes were saved. (I.e., is Up Done or is it Cancel?)

I don't think Left + Down + peek + Esc-with-the-other-hand is too unreasonable. You could make the same argument for Left + Space, I suppose, but that just feels more awkward to me than hammering on Esc to close a popup.

Issue #5121, opened 2025-03-14
Citation dialog: do not await for cited items


abaevbog:
https://forums.zotero.org/discussion/122628/zotero-beta-citation-dialog-appears-but-nothing-shows-up-when-i-write-something

If the document is very large, `SearchHandler._getCitedItems()` can take a very long time, so we should not wait for it to load before setting up all the IO handling. 

Also IO operations like cancelling the dialog should probably be pulled out of the normal `IOManager` workflow. The should be initialized and available from the very beginning regardless of what else happens later.

abaevbog:
@adomasven so as a general rule, I should assume that any call to `await io.getItems()` can take an arbitrary amount of time, right? Accounting for arbitrary large documents with lots of citations.

That means I should also avoid discarding cited items cache and re-fetching cited items via `io.getItems()` when the dialog is re-focused.

adomasven:
See https://github.com/zotero/zotero/blob/2e9b9d40708cf5940c06140ba3b9f90c9b983097/chrome/content/zotero/xpcom/integration.js#L1695-L1780

This is passed in as io into the dialog. `io.getItems()` will return fast once the promise for items resolves, but it won't return something new each time, so there's no reason to keep calling that.

PR #5127, opened 2025-03-17
Citation dialog: interactability before cited items are loaded

abaevbog:
- Initialize asynchronous loading of cited items via `io.getItems` when the dialog appears. The dialog is usable when the cited items are not yet loaded, they just don't appear in search results.
- Once cited items are loaded, the list of items is refreshed to include matching items and all subsequent searches will include them too.
- During the refresh immediately after cited items load, try to preserve focused/selected state of items.
- Minor refactoring of `SearchHandler` to handle refreshing of cited items separately from selected and open items. Some cleanup.

I added a `Zotero.Promise.delay(4000)` to `SearchHandler._getCitedItems()` to test out the behavior. Note the cited items appearing after a delay:

https://github.com/user-attachments/assets/664edf11-2158-4366-ad3a-a152df249e34

As a separate but related tweak, move initialization of accept/cancel functionality of the dialog into a separate function and call it in the very beginning of `onLoad`. `IOManager.init()` relies on layouts being loaded, which is why it is ran in the very end. But one should be able to close the dialog immediately, even if `IOManager.init()` doesn't run for a while or doesn't run at all (due to some unexpected error).

Fixes: #5121

### DIFF
--- a/chrome/content/zotero/integration/citationDialog.js
+++ b/chrome/content/zotero/integration/citationDialog.js
@@ -67,6 +67,9 @@ async function onLoad() {
 	libraryLayout = new LibraryLayout();
 	listLayout = new ListLayout();
 
+	// initialize most essential IO functionality (e.g. accept/cancel)
+	// remaining listeners that rely on layouts being loaded are added later in IOManager.init
+	IOManager.preInit();
 	// top-level keypress handling and focus navigation across the dialog
 	// keypresses for lower-level bubble-specific behavior are handled in bubbleInput.js
 	doc.addEventListener("keydown", event => KeyboardHandler.handleKeydown(event));
@@ -137,6 +140,7 @@ function cancel() {
 }
 
 function cleanupBeforeDialogClosing() {
+	if (!currentLayout || !libraryLayout) return;
 	Zotero.Prefs.set("integration.citationDialogLastClosedMode", currentLayout.type);
 	if (currentLayout.type == "library") {
 		Zotero.Prefs.set("integration.citationDialogCollectionLastSelected", libraryLayout.collectionsView.selectedTreeRow.id);
@@ -862,6 +866,15 @@ class ListLayout extends Layout {
 const IOManager = {
 	sectionExpandedStatus: {},
 
+	// most essential IO functionality that is added immediately on load
+	preInit() {
+		_id("accept-button").addEventListener("click", accept);
+		_id("cancel-button").addEventListener("click", cancel);
+
+		doc.addEventListener("dialog-accepted", accept);
+		doc.addEventListener("dialog-cancelled", cancel);
+	},
+
 	init() {
 		// handle input receiving focus or something being typed
 		doc.addEventListener("handle-input", ({ detail: { query, eventType } }) => this._handleInput({ query, eventType }));
@@ -878,10 +891,6 @@ const IOManager = {
 		doc.addEventListener("select-items", ({ detail: { startNode, endNode } }) => this.selectItemNodesRange(startNode, endNode));
 		// update bubbles after citation item is updated by itemDetails popup
 		doc.addEventListener("item-details-updated", () => this.updateBubbleInput());
-		
-		// accept/cancel events emitted by keyboardHandler
-		doc.addEventListener("dialog-accepted", accept);
-		doc.addEventListener("dialog-cancelled", cancel);
 
 		doc.addEventListener("DOMMenuBarActive", () => this._handleMenuBarAppearance());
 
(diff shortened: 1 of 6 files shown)


dstillman:
I guess this was probably the same in QF with large documents, but it's a little weird that Cited just appears all of a sudden.

Ideally we would show a Cited section sooner with a spinner (@yexingsha), but we probably don't want to have that flash on and disappear in every new document. Could we _not_ show it for new documents where we just set a style but otherwise show it with a spinner, and if they deleted all citations, it disappears and that's fine?

abaevbog:
> Could we not show it for new documents where we just set a style but otherwise show it with a spinner, and if they deleted all citations, it disappears and that's fine?

Right, we can record when the session is being initialized that cited items should not be appearing, and pass it as one of `io` properties to the citation dialog. I'm afraid it might still briefly appear and disappear in many cases if your document is not very big and there are a few citations - because it would take just a moment for the `io.getItems` to resolve. But let's give it a shot!

One other, a bit more general, suggestion could be to add a small debounce-type delay specifically for the cited items section. What if we only show the loading cited items spinner if the `io.getItems` promise is not yet resolved after, say, half a second? That way, if there are no items or just a couple, we don't show the spinner at all. But it would appear only if it's actually taking some time to load. Obviously, in case if it takes 0.75 of a second to load, then it would look like the spinner is flashing. But my assumption is that there are more instances when cited items would load quite quickly, so it would be a reasonable trade off?

dstillman:
> I'm afraid it might still briefly appear and disappear in many cases if your document is not very big and there are a few citations - because it would take just a moment for the io.getItems to resolve.

Why would it disappear?

> What if we only show the loading cited items spinner if the io.getItems promise is not yet resolved after, say, half a second?

Yeah, we can try this. We might want different timings for different implementations (Word for Mac, Google Docs, etc.).

dstillman:
> I'm afraid it might still briefly appear and disappear in many cases if your document is not very big and there are a few citations - because it would take just a moment for the io.getItems to resolve.

Oh, you mean the spinner would appear and then be replaced almost immediately by results? Yeah, we can try to avoid that with a delay.

abaevbog:
> Oh, you mean the spinner would appear and then be replaced almost immediately by results? Yeah, we can try to avoid that with a delay.

Right, that is what I meant. Sorry I wasn't clear.

But having tried this out, I am wondering if we actually do need any delay or extra logic for the placeholder here. The cited items do not appear unless you type something. The same should apply to the placeholder of cited items, right? In that case, the small delay I was thinking about adding before potentially showing the placeholder (e.g. 0.5 seconds) is about the amount of time one would need to start typing after the dialog appears. So then, right now, if there are not too many cited items (or none at all), by the time you type your query, they should all be loaded, so no placeholder would appear.

I was testing it by adding an artificial `Zotero.Promise.delay()` before `io.getItems()`. This is what 0.5 delay looks like. Here, the cited items are loaded and appear by the time I type anything.

https://github.com/user-attachments/assets/7906dcbf-1aae-4318-b99b-3575ac193d4e

There is an edge case if it takes slightly longer than a fraction of a second but less than several seconds. Because we don't know when the cited items can appear, there is always a small chance they will be loaded right as soon as the the placeholder appears. But from my testing, i never ran into that specific instance, and I am not sure we can completely avoid it, unless we have a way of knowing when `io.getItems` would resolve. This what placeholder with 1 second artificial delay looks like:

https://github.com/user-attachments/assets/8d6a075c-5d91-4115-8ccd-364b51d633d4


Finally, for reference, this is with a slightly longer 7 seconds delay before `io.getItems()`:

https://github.com/user-attachments/assets/a1c0ef8b-0c22-4f77-aa7d-449910e982f6

abaevbog:
Oh and let me know how the placeholders look! I am using the same section as the proper cited items would have but with a spinner after the title to indicate that something is loading. I am also adding a mock item node with "Loading..." title. It is mainly required for the library mode to not have a header seemingly floating in the air with nothing underneath. But I thought it can also go into the list mode because the transition when placeholder is gone and the real items appear looks a bit nicer when there is already something there.

abaevbog:
CC: @yexingsha about the placeholders note above

yexingsha:
I think the placeholder looks fine, though it should be "Cited Items" + spinner, without the "(0)".

I'm unsure whether the mock item is necessary; I guess in library mode the blank space will look weird if it's following selected/open items, so let's keep it for now.

abaevbog:
> I'm unsure whether the mock item is necessary; I guess in library mode the blank space will look weird if it's following selected/open items, so let's keep it for now.

Yeah, I thought that this would just looks like something is broken with all that empty space below below the cited label:

Screenshot 2025-03-19 at 9 51 58 PM

Especially when other cards are present:

Screenshot 2025-03-19 at 9 54 46 PM

With the list layout it's less needed and without it the list would look like this (which seems fine). But if we have the placeholder in library, might as well have it everywhere? Besides, I think the transition is nicer when items do appear if something is already there.

Screenshot 2025-03-19 at 9 56 11 PM

dstillman:
Can we just show "Loading…" like we do in List Mode and not the mock item? Fine to fill the space with something, but I don't think the mock item looks great, and it's weird to have "Loading…" shown in two different ways switching between modes.

dstillman:
Merging this for now, though, to get the functional improvement

PR #5129, opened 2025-03-18
Fix plugin scope not loaded if plugin is disabled when starting Zotero

windingwind:
Since the bootstrap plugin framework, from time to time I ran into cases where some plugins don't start.

It turned out that the sandbox scope is not loaded at all if the plugin is

1. installed before Zotero starts
2. disabled before Zotero starts

It is amazing that no one ever reported/found this bug, which is a quite serious bug... Maybe everyone would blame the plugin developers for not having the plugin well-written.


dstillman:
How would something be installed or disabled before Zotero starts?

dstillman:
You mean if a plugin was installed and then disabled, and then you restart Zotero and enable it?

windingwind:
> You mean if a plugin was installed and then disabled, and then you restart Zotero and enable it?

Yes, sorry missed the message

Issue #2712, opened 2022-07-20
Add a dropmarker to Quick Format citation bubbles


dstillman:
https://twitter.com/grace_lao/status/1549544846913933313

I've been thinking for a while that we need to make the popup more obvious. People in the forums pretty regularly have no idea that there's more to the dialog.

We probably want a light-blueish triangle on the right-hand side. Will slightly ruin the asymmetry of the bubbles, but I think it's worth it for the increased usability.

dstillman:
https://twitter.com/WorldImagining/status/1563156252729180163

yexingsha:
Since many people aren't aware they can click on bubbles to customize citations, let's add dropmarkers to them:

<img width="1280" alt="Image" src="https://github.com/user-attachments/assets/5ef05bb3-2d5a-4aaf-8929-9abb5c9bba22" />

The dropmarker uses `chevron-6.svg`. The gap between the text and the dropmarker is 4px. 

Hovering over a bubble will no longer show the `X` button and mask, since people can always delete a bubble using either the keyboard or the `Remove` button in the popup.

yexingsha:
> But one question I have is this: isn't not knowing that a bubble is clickable a one-time problem? But once you find it out, then you just know. Then, after you do discover it, the dropmarkers would be no longer useful, and you'd rather have the X button that can actually perform an action in its place (regardless how often you would use it)?

I agree this is the case, but it seems to be a problem for a so many people that the utility of dropmarkers might outweigh the utility of the occasionally relevant X button.

> what if we just ... open the itemDetails popup for the user ourselves at some point? Maybe, if one opens the citation dialog for the very first time, as soon as they add their first item in the citation, instead of refocusing the input, we open the bubble's popup? That would be the most direct way to indicate that there is a popup associated with a bubble .

I think the inconsistency could be confusing rather than illuminating, making people wonder why this popup automatically appeared before and not now. And it can run into the same problem as first-run guidance panels, which is that people sometimes just don't pay attention. @dstillman what do you think?

dstillman:
> Does it feel slightly off to be explicitly removing an existing feature?

I mean, I wouldn't really call it an "existing feature" — it's been out in a beta for a week. People have gotten by without the X for many years, and I don't recall this being a source of confusion. Conversely, people not realizing that you could and should customize citations via Quick Format has been a major problem for years, and I don't think the new button style really does much to make that clearer.

[...]

PR #5130, opened 2025-03-19
Citation dialog: add a dropmarker to bubbles

abaevbog:
- add dropmarker to bubbles after the text to indicate that bubbles are clickable
- remove X remove button that was there before
- no more mask over the bubble on hover, since dropmarker is just always visible
- 4px distance between the dropmarker and bubble text

Fixes: #2712

Screenshot 2025-03-18 at 11 26 52 PM

### DIFF
--- a/chrome/content/zotero/elements/bubbleInput.js
+++ b/chrome/content/zotero/elements/bubbleInput.js
@@ -233,16 +233,10 @@
 			text.className = "text";
 			bubble.append(text);
 			
-			let deleteBtn = document.createElement("div");
-			deleteBtn.className = "delete-btn";
-			let cross = document.createElement("span");
-			cross.className = "icon icon-css icon-x-8 icon-16";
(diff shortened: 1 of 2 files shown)


abaevbog:
@yexingsha how is this tweak looking to you? Here, the `X` button is gone and instead we have a `chevron-6` dropmarker 4px after the bubble's text. It is always visible, so the the mask effect and etc. on hover is removed too.

yexingsha:
Looks good to me!

dstillman:
The dropmarker goes away if you customize the citation (and even if you remove the customization):

Screenshot 2025-03-20 at 2 44 21 AM

Screenshot 2025-03-20 at 2 44 30 AM

abaevbog:
Sorry about that! Should be fixed now:

Screenshot 2025-03-20 at 8 56 01 AM

PR #5133, opened 2025-03-19
Add ability to delete attachment from item pane

### LINKED ISSUE #5041: Add ability to delete attachment from item pane
https://forums.zotero.org/discussion/121808/from-client-deleting-a-web-link-in-attachments and some others

Some people have been confused by this, and it's reasonable, since you can add attachments from there and remove tags/related/etc. I think we can just show the (-) and trigger the usual confirmation prompt?

tnajdek:
https://github.com/user-attachments/assets/814a1bc6-e97c-4e4c-8265-e999be3aba28

Closes #5041.

### DIFF
--- a/chrome/locale/en-US/zotero/zotero.ftl
+++ b/chrome/locale/en-US/zotero/zotero.ftl
@@ -462,6 +462,7 @@ section-attachments-annotations =
                  [one] { $count } Annotation
                  *[other] { $count } Annotations
              }
+section-attachments-move-to-trash-message = Are you sure you want to move “{ $title }” to the trash?
 section-notes =
     .label = { $count ->
                  [one] { $count } Note
--- a/scss/elements/_attachmentRow.scss
+++ b/scss/elements/_attachmentRow.scss
@@ -43,6 +43,15 @@ attachment-row {
 			}
 		}
 
-		
+		.zotero-clicky-minus {
+			margin-inline-start: auto;
+			& > image {
+				visibility: hidden;
+			}
+		}
+
+		&:is(:hover, :focus-within) .zotero-clicky-minus > image {
+			visibility: visible;
+		}
 	}
 }
--- a/scss/elements/_attachmentRow.scss
+++ b/scss/elements/_attachmentRow.scss
@@ -42,16 +42,7 @@ attachment-row {
 				color: var(--fill-secondary);
 			}
 		}
-
-		.zotero-clicky-minus {
-			margin-inline-start: auto;
-			& > image {
-				visibility: hidden;
-			}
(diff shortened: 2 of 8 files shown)


chrome/locale/en-US/zotero/zotero.ftl
```diff
@@ -462,6 +462,7 @@ section-attachments-annotations =
                  [one] { $count } Annotation
                  *[other] { $count } Annotations
              }
+section-attachments-move-to-trash-message = Are you sure you want to move "{ $title }" to the trash?
```
dstillman, line 465: Typographic quotes

scss/elements/_attachmentRow.scss
```diff
@@ -43,6 +43,15 @@ attachment-row {
...
+		.zotero-clicky-minus {
+			margin-inline-start: auto;
+			& > image {
+				visibility: hidden;
+			}
+		}
+
+		&:is(:hover, :focus-within) .zotero-clicky-minus > image {
+			visibility: visible;
+		}
```
dstillman, line 55: I don't know anything about it, but are we able to reuse the `show-on-hover` class instead of custom CSS for this? We seem to be using that for some other minus buttons.

tnajdek: Thanks for pointing this out! I wasn't aware of this class, but it doesn't seem to fit the bill in this case. It appears that the `.show-on-hover` class only works within a `meta-table` mixin, which I believe is meant for tabular data, such as in the "Info" or "Attachment Info" panes.  

For a toolbar button, it seems we don't use a mixin and instead just repeat the code (see below). We could introduce a mixin for these, but I don't feel like it's particularly warranted in this scenario.

https://github.com/zotero/zotero/blob/929147b33ab9610fdee36050de8db71d3b3e26f5/scss/elements/_librariesCollectionsBox.scss#L53-L55
https://github.com/zotero/zotero/blob/929147b33ab9610fdee36050de8db71d3b3e26f5/scss/elements/_notesBox.scss#L42-L44
https://github.com/zotero/zotero/blob/929147b33ab9610fdee36050de8db71d3b3e26f5/scss/elements/_tagsBox.scss#L75-L77

chrome/locale/en-US/zotero/zotero.ftl
tnajdek: Thanks, fixed!

scss/elements/_attachmentRow.scss
tnajdek: Or actually, maybe it's worth making a new mixin for these, let me try.

PR #5141, opened 2025-03-20
Let autocomplete popup get wider

### LINKED ISSUE #5140: Autocomplete popup is too narrow
I think this may have been a Z7 change. No reason it can't be wider IMO.

https://forums.zotero.org/discussion/122794/zotero7-suggestions-for-autocompletion-are-too-short

AbeJellinek:
The popup tries to size itself based on the width of the input that opened it (input's width or 100px, whatever is greater). This removes that constraint so it uses the standard popup layout algorithm. It still doesn't grow past the edge of the `info-box` for some reason (which is good).

This is the simple approach. I also have alternative code to patch `autocomplete-popup.js` so it still explicitly calculates its width, but allows a parent (e.g. the info-box row) to set itself as the anchor in place of the input. Can go with either approach.

Fixes #5140

### DIFF
--- a/app/scripts/fetch_xulrunner
+++ b/app/scripts/fetch_xulrunner
@@ -404,6 +404,12 @@ function modify_omni {
 		overflow: var(--moz-text-control-overflow);
 	}' >> chrome/toolkit/res/forms.css
 	
+	# By default, an autocomplete popup's width is calculated based on the input that opened it.
+	# Allow an ancestor to designate itself as the width container instead. (For creator inputs.)
+	replace_line 'aElement.getBoundingClientRect\(\).width' \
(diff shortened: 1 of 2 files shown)


dstillman:
Are we able to set a min-width to the width of the input? I feel like, just visually, it's much more standard for the popup to be the full width of the input. Now it can be much narrower if the suggestions are short, which looks a little weird.

It might also make sense to limit this change to creator names. That's the primary issue, right? And without this it just uses the width of the input (as shown in https://github.com/zotero/zotero/issues/5140#issuecomment-2742178675)? Allowing the popup to go past the inputs that extend across the info pane doesn't seem particularly necessary.

Issue #4233, opened 2024-06-13
Disable snapshot saving from Zotero


adomasven:
For single-item translations snapshots should come from the browser, even if the translator specifies a different snapshot url.

For search result and other multiple-item saves snapshots will not be saved. We might attempt to restore this in Connector https://github.com/zotero/zotero-connectors/issues/481

PR #5148, opened 2025-03-21
Receive attachments from Zotero Connector

adomasven:
Connector-backwards-incompatible changes to Zotero which change Zotero Connector server to expect attachments to be saved by Connector instead of Zotero.

This started with replacing our custom server with mozilla's bundled httpd to support saving binary streams directly to file without hopefully having to hold attachments in memory in multiple places (Connector + Zotero).

Also refactors session saving and removes some old connector server endpoints that are no longer needed.

Also includes a couple of misc test changes as I was having running tests locally. Some of them still don't pass, but not more than with the main branch.

Closes https://github.com/zotero/zotero/issues/4233


dstillman:
Files aren't being automatically renamed.

Before:

Screenshot 2025-03-27 at 7 19 58 AM

This PR:

Screenshot 2025-03-27 at 7 19 53 AM

chrome/content/zotero/xpcom/server/server_connector.js
```diff
@@ -0,0 +1,1241 @@
...
+
+/**
+ * Checks if the item has OA attachments (in case PDF saving in connector failed).
+ * Also checks custom resolvers.
+ * 
+ * Accepts:
+ *		sessionID - A session ID previously passed to /saveItems
+ *		itemID - The ID of the item to save alternative attachment for
+ */
+Zotero.Server.Connector.HasOAAttachments = function () {};
```
dstillman, line 588: This seems like not the right name. It's not actually checking whether there are OA attachments. It's checking if there are potential resolvers based on the available fields.

adomasven: Yes. `hasAttachmentResolvers`? Should the `getOAAttachments` endpoint change too?

dstillman: `hasAttachmentResolvers` works.

> Should the `getOAAttachments` endpoint change too?

You mean `saveOAAttachment`? Yeah. Um… `saveAttachmentFromResolver`?

chrome/content/zotero/xpcom/server/server_connector.js
```diff
@@ -0,0 +1,1241 @@
...
+		let item = session.getItemByConnectorKey(data.itemID);
+		let resolvers = Zotero.Attachments.getFileResolvers(item, ['oa', 'custom'], true);
+
+		let attachment = await Zotero.Attachments.addFileFromURLs(item, resolvers);
+
+		if (attachment) {
+			return [201, "text/plain", attachment.getDisplayTitle()];
+		}
+		else {
+			return [503, "text/plain", "Failed to save an attachment"];
```
dstillman, line 640: 500? Not that it really matters, but 503 doesn't seem right here.

chrome/content/zotero/xpcom/attachments.js
```diff
@@ -1010,21 +1010,21 @@ Zotero.Attachments = new function () {
...
 
 		if (parentItemID && collections) {
 			throw new Error("parentItemID and collections cannot both be provided");
 		}
 		
+		// Create a temporary file
+		let parentItem = Zotero.Items.get(parentItemID);
+		let fileBaseName = this.getFileBaseNameFromItem(parentItem, { attachmentTitle: title });
+		let ext = this._getExtensionFromURL(url, contentType);
+		let filename = Zotero.File.truncateFileName(`${fileBaseName}.${ext}`, 100);
```
dstillman, line 1022: I don't think we need to truncate. We only do that currently when the filename comes from the URL, to handle cases where the last path component is very long. Here we're doing standard metadata-based renaming, and the renaming format already handles title truncation.

I can't remember if we handle filesystem errors from too-long filenames, but that would be handled generally, not when assigning the name.

dstillman: And if `_getExtensionFromURL` is a concern, we should just add reasonable truncation (20 characters?) to that.

chrome/content/zotero/xpcom/file.js
```diff
@@ -446,6 +446,68 @@ Zotero.File = new function(){
...
+	 * @param {nsIInputStream} inputStream - The input stream to read from. This
+	 *        stream should implement nsIAsyncInputStream.
+	 * @param {string} path - The file path where the data will be written.
+	 * @param {number} byteCount - The expected number of bytes to write.
+	 *
+	 * @returns {Promise<number>} A promise that resolves with the number of bytes
+	 *          written when the operation is complete, or rejects with an error
+	 *          if any issues occur during reading or writing.
+	 */
+	this.putNetworkStream = async function (path, stream, byteCount) {
```
dstillman, line 464: Why do we need this when `putContentsAsync()` already takes an `nsIInputStream`? Can we just make any modifications to that if necessary?

chrome/content/zotero/xpcom/intl.js
```diff
@@ -126,28 +126,29 @@ Zotero.Intl = new function () {
...
+			// TEMP: The this.strings check prevents "TypeError: this.strings is undefined"
+			// from an early Zotero.getString() call, but I'm not sure why. this.strings is
+			// set using defineLazyGetter, but lazy doesn't mean asynchronous...
+			//
+			// https://forums.zotero.org/discussion/117812/issue-with-installing-zotero-7
+			if (this.strings && this.strings[name]) {
+				return this.strings[name];
+			}
+			else {
+				l10n = bundle.GetStringFromName(name);
```
dstillman, line 151: This is causing a [test failure from unreplaced placeholders](https://github.com/zotero/zotero/actions/runs/14188744079/job/39748715233) and lots of logged errors:

```
zotero(3)(+0000024): NS_ERROR_FAILURE Exception: Component returned failure code: 0x80004005 (NS_ERROR_FAILURE) [nsIStringBundle.formatStringFromName]

    Zotero.Intl</this.getString@chrome://zotero/content/xpcom/intl.js:133:23
    this.getString@chrome://zotero/content/xpcom/zotero.js:1540:24
    get searchModePopup@chrome://zotero/content/elements/quickSearchTextbox.js:140:45
    connectedCallback@chrome://zotero/content/elements/quickSearchTextbox.js:85:25
```

```
zotero(3)(+0000002): NS_ERROR_FAILURE Exception: Component returned failure code: 0x80004005 (NS_ERROR_FAILURE) [nsIStringBundle.formatStringFromName]

    Zotero.Intl</this.getString@chrome://zotero/content/xpcom/intl.js:133:23
    this.getString@chrome://zotero/content/xpcom/zotero.js:1540:24
    getLocalizedName@chrome://zotero/content/xpcom/data/searchConditions.js:738:36
    Zotero.SearchConditions</this.init<@chrome://zotero/content/xpcom/data/searchConditions.js:667:25
```

```
zotero(3)(+0000000): NS_ERROR_XPC_CANT_CONVERT_OBJECT_TO_ARRAY Exception: Cannot convert JavaScript object into an array arg 1 [nsIStringBundle.formatStringFromName]

    Zotero.Intl</this.getString@chrome://zotero/content/xpcom/intl.js:133:23
    this.getString@chrome://zotero/content/xpcom/zotero.js:1540:24
    versionWarning@chrome://zotero/content/xpcom/server/server_connector.js:131:28
    init@chrome://zotero/content/xpcom/server/server_connector.js:1111:31
```

Etc.

dstillman: Yeah, this broke .properties placeholder substitution altogether. You're calling `l10n = bundle.formatStringFromName(name, params, params.length);` and then replacing it with `l10n = bundle.GetStringFromName(name);`.

chrome/content/zotero/xpcom/file.js
adomasven: Not in my attempts. `putContentsAsync()` only works with streams that have EOF, which this does not. We need to copy a specific number of bytes and then the http server code writes the response on the same channel. And I could not find a utility function in XPCOM that copies a specific number of bytes from a stream as a single function call.

chrome/content/zotero/xpcom/attachments.js
adomasven: `truncateFileName` already treats long extensions (20 chars) as part of filename. And in my testing this truncation often does get triggered, and also mimics previous behaviour, and we probably don't want to produce super-long filenames.

dstillman: Not sure what you mean here.

We use `truncateFileName` in two places: `importFromDocument` and `importFromSnapshotContent`. And in both cases we're using it because the "filename" is just coming from the URL. We don't use it for renamed-attachment saving, so we shouldn't be using it here.

I'm saying that 1) if `_getExtensionFromURL` can result in a very long string (because there's no known Content-Type and the URL has a period followed by lot of characters?), we should just handle that there and 2) if there's some problem with overly long filenames being saved for renamed attachments, it should be handled centrally where it applies to all uses of `getFileBaseNameFromItem`, not here. (But the default rename string already truncates titles to 100 characters.)

adomasven: Ah, you're right. I got mixed up when working on this. There are a lot of renaming methods here and every attachment import source does something slightly different!

I didn't identify any specific issues, was just following what we do in other places, so I don't think there's any need to do something special here, since importFromURL which was used previously to save attachment files didn't truncate either.

adomasven:
Addressed in the Connector for saving failure which shouldn't happen, and the wrongly shown OA line. OA failure can still be tested on ScienceDirect or other pages where institutional login is required.

Issue #5151, opened 2025-03-21
Citation dialog: remove mock item when cited items are loading


abaevbog:
Per https://github.com/zotero/zotero/pull/5127#issuecomment-2742166779:

> Can we just show "Loading…" like we do in List Mode and not the mock item? Fine to fill the space with something, but I don't think the mock item looks great, and it's weird to have "Loading…" shown in two different ways switching between modes.

PR #5152, opened 2025-03-21
citation dialog: remove cited item placeholder in library mode

abaevbog:
Keep the placeholder node to retain the positioning of the text consistent but remove the border so the item card is invisible.

Fixes: #5151

Screenshot 2025-03-21 at 2 29 08 PM

### DIFF
--- a/scss/components/_citationDialog.scss
+++ b/scss/components/_citationDialog.scss
@@ -233,6 +233,10 @@
 							&.selected {
 								background-color: var(--selected-item-background) !important;
 							}
+							// no visible border around the placeholder
+							&[disabled] {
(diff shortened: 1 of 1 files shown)


abaevbog:
I thought it's easiest to keep the placeholder item card to have the same centered positioning for "Loading..." as for an item card without the description. But then we can just hide the border so it does not look like an item card. @dstillman does this look like what you had in mind?

dstillman:
Do we really need the spinner and "Loading…"? Particularly since there's another spinner in the dialog, it seems like we could just take that out. We use "Loading items…"/etc. elsewhere.

PR #5157, opened 2025-03-24
Potential fix to sciencedirect infinite captcha

abaevbog:
Before opening the PDF, open sciencedirect homepage where a number of cloudflare cookies (including cf_clearance indicating that this is a real user) are set. Wait for the cookies to be added to the sandbox, and then proceed to open the PDF. In that case, captcha should not appear at all.

This has been working consistently for me but I'm not sure if this is the best way to do it (e.g. is it worth to try to do it when we download via HiddenBrowser first?) and if this is something that should be more generalized, as opposed to being just a sciencedirect special case.

### DIFF
--- a/chrome/content/zotero/xpcom/browserDownload.js
+++ b/chrome/content/zotero/xpcom/browserDownload.js
@@ -212,6 +212,24 @@ Zotero.BrowserDownload = {
 		// make sure that you are using the browser you are claiming to be.
 		delete options.cookieSandbox?.userAgent;
 		
+		// Going to the URL of a sciencedirect PDF immediately opens a page with captcha stuck in
+		// an infinite loop. Opening a home page first and waiting a bit before redirecting to the PDF file
+		// seems to not trigger the captcha and should successfully download the PDF.
+		// When homepage opens, CF sets a number of cookies (like cf_clearance) which are
+		// indicators that this is a real user and are required for the PDF to be downloaded.
+		if (uri.host.includes("sciencedirect") && options.cookieSandbox) {
+			let homepage = new URL("https://sciencedirect.com");
+			win = Zotero.openInViewer(homepage, { cookieSandbox: options.cookieSandbox });
+			let loopCount = 0;
+			let receivedCookies;
+			// wait for cf_clearance cookie to be set but not more than 10 seconds
+			do {
+				receivedCookies = options.cookieSandbox.getCookiesForURI(Services.io.newURI(homepage)) || {};
+				await Zotero.Promise.delay(100);
+			} while (!receivedCookies.cf_clearance && loopCount++ < 100);
+			win.close();
+			// At this point, all the right cookies should be set but if not - just try to open the PDF anyway
+		}
 		try {
 			wmListener = {
 				onOpenWindow(xulWindow) {
--- a/chrome/content/zotero/xpcom/browserDownload.js
+++ b/chrome/content/zotero/xpcom/browserDownload.js
@@ -212,24 +212,10 @@ Zotero.BrowserDownload = {
 		// make sure that you are using the browser you are claiming to be.
 		delete options.cookieSandbox?.userAgent;
 		
-		// Going to the URL of a sciencedirect PDF immediately opens a page with captcha stuck in
-		// an infinite loop. Opening a home page first and waiting a bit before redirecting to the PDF file
-		// seems to not trigger the captcha and should successfully download the PDF.
-		// When homepage opens, CF sets a number of cookies (like cf_clearance) which are
-		// indicators that this is a real user and are required for the PDF to be downloaded.
-		if (uri.host.includes("sciencedirect") && options.cookieSandbox) {
-			let homepage = new URL("https://sciencedirect.com");
-			win = Zotero.openInViewer(homepage, { cookieSandbox: options.cookieSandbox });
-			let loopCount = 0;
-			let receivedCookies;
-			// wait for cf_clearance cookie to be set but not more than 10 seconds
-			do {
-				receivedCookies = options.cookieSandbox.getCookiesForURI(Services.io.newURI(homepage)) || {};
-				await Zotero.Promise.delay(100);
-			} while (!receivedCookies.cf_clearance && loopCount++ < 100);
-			win.close();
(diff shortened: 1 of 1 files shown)


adomasven:
How are you testing this? It doesn't work for me when saving from Chrome on Linux and I get the captcha. You are checking for the cf_clearance cookie, which Firefox and Chromium browser will pass to the Zotero client, so I'm not sure how this would help.

Furthermore, we should probably make this explicitly a ScienceDirect workaround, like in a BrowserDownload.ScienceDirect namespace or something, and also open the page in a hidden browser, instead of a visible dialog, if some version of this actually helps.

Not sure if this is related to your proposed approach or something else, but I got blocked by ScienceDirect after a few requests.

abaevbog:
> How are you testing this? It doesn't work for me when saving from Chrome on Linux and I get the captcha. 

I'm just going to one of the sciencedirect pages and clicking on the connector icon. This is a recording of what I see saving it after clearing caches everywhere: [video](https://www.dropbox.com/scl/fi/6yq0r72uwj2l6ou09sk60/sciencedirect_demo.mov?rlkey=erkupkdmaegn6vf142ebms5g4&st=ioyrhrs8&dl=0). I get this behavior on every platform that I have (macOS, windows 11 VM and older ubuntu) on firefox and chrome.

> You are checking for the cf_clearance cookie, which Firefox and Chromium browser will pass to the Zotero client, so I'm not sure how this would help.

Right, let me add some more context on this. cf_clearance cookie seems important. My earlier observation on this topic was that rerunning a cancelled POST request from the captcha page would help it pass. cf_clearance is one of the cookies that would get set after that curious request rerun. It is also a part of [CF anti-bot detection](https://developers.cloudflare.com/fundamentals/reference/policies-compliances/cloudflare-cookies/#cf_clearance-cookie-for-cloudflare-bot-products) (which may be the mechanism causing issues here).

To be honest, I am not 100% confident that cf_clearance is really at the root of the issue here - all of the above are educated guesses. But I got this idea after I tried to download a sciencedirect file from `basicViewer` by opening the homepage first, navigating to the item and clicking "View PDF". It worked just fine, which means there is nothing about `basicViewer` preventing us from getting the file, it's just about how we get to the file. After that, I tried to open homepage first, wait for 7 seconds and then go to the PDF file url - that worked too. Next, I tried to open the homepage and only wait till the window loads (`win.addEventListener('load', resolve);` same as what we do with the PDF window) but that did not work and I got the captcha loop. So this means we need to get to the main sciencedirect page, wait for ... something, and then get the PDF. There are numerous cookies being set after loading, and cf_clearance is the one that seemed to have mattered the most during my experiments with the POST request, so I figured we can try to wait for it.

> Furthermore, we should probably make this explicitly a ScienceDirect workaround, like in a BrowserDownload.ScienceDirect namespace or something, and also open the page in a hidden browser, instead of a visible dialog, if some version of this actually helps.

Gotcha! To be clear, you mean make it a special case when we decide to [download the PDF](https://github.com/zotero/zotero/blob/6787dcd596f987fdae5ed08d09f85da4242e7806/chrome/content/zotero/xpcom/attachments.js#L1119) from Zotero.Attachments? That makes sense. But you said it's not working on your end on linux with chrome... What about firefox? I have tried to run into issues on linux on both firefox and chrome but have not been able to - this just seems to work for me. I'll try to see if I can reproduce the captcha thing with these changes - because it probably doesn't make sense to add a fix that only works for me.

> Not sure if this is related to your proposed approach or something else, but I got blocked by ScienceDirect after a few requests.

Yeah, also not sure what could be going on here ... I have been cautiously downloading files quite a bit over the span of several days and have not ran into any issues with that

abaevbog:
One more thing to note here is this does not help me with sciencedirect on Safari. I get a red cross next to the PDF name in the connector right away instead of the captcha window. But it also looks like `Zotero.BrowserDownload.downloadPDF` is just never hit when I try to add the item with Safari, so this behavior likely is not affected by changes here.

abaevbog:
> No, I mean the cookieSandbox that is used there to save the page, by the time we arrive to the code which opens the ScienceDirect homepage, already has a cf_clearance cookie set

I see what you mean, thank you for the explanation! I think I have an idea of why sciencedirect is breaking to begin with. cf_clearance cookie is never actually passed by the connector to the cookie sandbox. If I `console.log(options.cookieSandbox)` in `downloadPDFViaViewer`, cf_clearance is not there. This must be the reason why visiting the homepage before going to the PDF url helps - that's how we get the missing cf_clearance.

cf_clearance is a partitioned cookie, so when we [fetch all cookies](https://github.com/zotero/zotero-connectors/blob/fda9af45427b7389e44fa9e91dd96c04f67c505b/src/common/connector.js#L228) from the connector to pass it over to the desktop, cf_clearance is not returned by `browser.cookies.getAll({ url: tab.url })`.  Per [cookies.getAll()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/cookies/getAll#firstpartydomain) documentation, adding `firstPartyDomain: null` to [cookieParams](https://github.com/zotero/zotero-connectors/blob/fda9af45427b7389e44fa9e91dd96c04f67c505b/src/common/connector.js#L206C1-L208C6) allowed me to get cf_clearance cookie in addition to all the other ones.

With that small change to the connector, downloading PDFs works for me on Firefox without having to go to the homepage and without showing the captcha page. It does not yet work with chrome though.

>  Note that after a successful save Zotero will store the cache the good cookies, so you need to restart for tests with Chrome after it works with Firefox.

Interestingly, chrome still worked fine for me after briefly opening the homepage before going to the PDF right after a fresh restart (my ubuntu version is quite old so maybe that is relevant). But the reason why we open sciencedirect homepage should just be in order to get the missing cf_clearance cookie. If the connector sends it, it should not be required. But Chrome still sends me to the captcha loop for some reason. I am wondering if this  has something to do with the fact that when we send cf_clearance from the connector, we essentially discard the fact that it has a partition key. I am now looking into tweaking the connector to record which cookies are partitioned and to handle that in the Zotero.CookieSandbox.

abaevbog:
The real fix is in https://github.com/zotero/zotero-connectors/pull/535, after which this hacky workaround would not be needed.

PR #5162, opened 2025-03-26
Tweak column behaviour on visibility and order changes

### LINKED ISSUE #4993: itemTree column headers become misaligned with content when more columns are added
The more columns are added, the more misaligned they become. Resizing the columns fixes the alignment, but reordering the columns does not. As far as I can tell, this issue appears in all instances of itemTree, including main window, add related dialog, and the citation dialog that's being worked on.

https://github.com/user-attachments/assets/cbaccd5a-0b35-4eca-ad2e-fd03e00c70af

tnajdek:
When column visibility is changed, we now redistribute the available space between the columns. I believe this fixes #4993.

### DIFF
--- a/chrome/content/zotero/components/virtualized-table.jsx
+++ b/chrome/content/zotero/components/virtualized-table.jsx
@@ -1598,6 +1598,10 @@ var Columns = class {
 		this._virtualizedTable.props.storeColumnPrefs(prefs);
 	}
 
+	// Injects extra width into the first column if that column has
+	// `staticWidth` or `minWidth` configured, in order to accommodate space for
+	// the item type icon and the chevron.The original width is restored if the
+	// column is reordered to a different position.
 	_adjustColumnWidths = () => {
 		if (!this._virtualizedTable.props.firstColumnExtraWidth) {
 			return;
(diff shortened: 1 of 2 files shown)


adomasven:
Not sure if I'm up-to-date to review this. There's been column sizing changes to this code by Tom since the redesign, like the addition of `_adjustColumnWidths()`. Generally it's just a bit complicated to get all pieces of this working nicely, given we want flexibility in customization, but then have to handle layout manually.

@tnajdek Would you add comments to what and why you're doing here, and maybe in other places that are responsible for column sizing, and extract more constants if that would help?

tnajdek:
`_adjustColumnWidths()` injects extra width into the first column to accommodate the item type icon and the chevron. It currently only does this for columns with `staticWidth` or `minWidth`, but arguably, a similar logic should apply to every column. Without this, `hasAttachment` would be rendered at 32px, but if it's the first column, it needs to render at 60px to allow for the item type icon and the chevron. However, this isn't directly relevant to this change.

This change is a workaround for how the flex model distributes space for columns. Whenever a user resizes columns, we set `flex-basis` to the exact number of pixels the column takes on screen. In this scenario, the sum of all `flex-basis` values equals the header's width. However, as soon as a new column is introduced or the available space is reduced, the sum of `flex-basis` values exceeds the width of the table (and vice versa when removing columns).

That's generally fine because we allow for the growing/shrinking of cells, and proportions are maintained, so everything works. However, when the content becomes very dense and the sum of all `flex-basis` values substantially exceeds the table width, the flex model starts distributing space differently for the table body and the header. I'm not 100% sure about this (the flex model is a complex algorithm), but I believe this difference stems from the fact that the header is wider than the table body and/or from the intrinsic sizes of the cells.  

One way to fix this is to recalculate `flex-basis` for each column so that the sum of `flex-basis` values matches the width of the table. In this PR, I've introduced a piece of code to do just that whenever a column is toggled. We should probably also do something similar when the available space is reduced due to window or pane resize.

adomasven:
This is good. Would you add some of this where appropriate in the code just so it's clearer what's going on? One of the reasons why it's hard to review this change is that it's hard to know what is done where in this code, how to test for edge cases that could arise from this and what is the justification regarding how this is drawn with flex, etc.

tnajdek:
I've added some comments and extracted the new piece of logic into a separate function.

tnajdek:
Maybe I misunderstood the problem, here is what I'm seeing and what I'm trying to fix:

Without this PR:

https://github.com/user-attachments/assets/df02fe4e-7365-461d-90d8-fc1d1cbff856

I took a snapshot from the final state in the video and added red lines for visibility:

4993-columns-not-redistributed-snapshot-with-lines


With this PR:


https://github.com/user-attachments/assets/d11791b0-859a-47d8-8787-a6e0f6c2e009


Again, a snapshot from the final state, with red lines added for visibility:


4993-columns-redistributed-snapshot-with-lines

adomasven:
It's still an issue at least with Ubuntu 24.04. I'm testing by adding a few columns and dragging the resizer to see if columns are reflown, which they are in my case.

tnajdek:
@adomasven I've tweaked the logic that redistributes widths to account for fixed and static-width columns. Please give it a try and let me know if it resolves the issue.

tnajdek:
@adomasven Thank you for your help testing this! The issues with `NaN`s should be fixed, let me know if this works for you now.  

I've also tweaked the `itemType` column config, increasing its width from 40 to 56 to account for column padding and making it a `staticWidth`.

tnajdek:
How would you measure a width of a column that did not render yet?

adomasven:
Do it in the next event loop (i.e. `setTimeout()` without a timeout arg), or in `requestAnimationFrame()`, although that might make this behaviour visible, which may be jarring, although given that we are changing columns and reflowing stuff it may not be too bad, since this is a minor adjustment. Otherwise we could pre-calculate this in a `visibility: hidden` element, althought at that point it probably gets as complicated as this.

tnajdek:
Pre-calculating the column widths and updating the DOM once will always perform better than updating the DOM, measuring it, and then updating it again. While the code for this approach may be more complex, I’ve already implemented it. If you can point out a scenario where it doesn’t work correctly, I’m happy to fix it. However, unless there’s a compelling reason to switch approaches, I’d prefer to stick with this one.

If you strongly believe your DOM-measuring-based solution is better, please feel free to take over this issue.

dstillman:
@yexingsha Does this fix the issue you were seeing?

yexingsha:
The issue seems fixed for adding/removing columns, but it still shows up on reordering columns and resizing the window/itemTree.

tnajdek:
Correct, I only implemented it for toggling columns but we should be able to wire the same function to resolve the remaining issues.

dstillman:
We can handle those other situations in a separate commit.

dstillman:
Any chance https://forums.zotero.org/discussion/123877/chaotic-itemtree-in-beta-when-register-numerous-columns is related to this, or does this resize the full column, not individual cells?

tnajdek:
This might be related. If indeed this used to work in 7.0 but broke in 7.1 then maybe it's worth revisiting the fix.

Issue #5161, opened 2025-03-26
Citation dialog: item not yet loaded error


abaevbog:
If there is a item from a group that is not "My Library" opened in a tab, "Item ... not yet loaded" error can be thrown when the dialog just opens.

PR #5163, opened 2025-03-26
citation dialog: fix unloaded items errors

abaevbog:
- get opened items asynchronously and load their data in case if there are items opened across multiple
libraries, and some of them are unloaded.
This fixes: #5161, which makes the dialog unresponsive upon initial load.
- enforce `io.filterLibraryIDs` in library mode, the same way it already is being applies in list mode. This is currently used to only show items from the current library when adding citations in a note.
- account for the change above in collectionTree by focusing the first library from `io.filterLibraryIDs`, if those are present, and only default to `userLibraryID` if `io.filterLibraryIDs` is empty.

### DIFF
--- a/chrome/content/zotero/collectionTree.jsx
+++ b/chrome/content/zotero/collectionTree.jsx
@@ -118,7 +118,10 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 			var selected = await this.selectByID(lastViewedID);
 		}
 		if (!selected) {
-			await this.selectByID('L' + Zotero.Libraries.userLibraryID);
+			// If the last viewed folder was not selected, default to the first library from
+			// filterLibraryIDs (if any), or the user library
+			let libraryToSelect = ((this.props.filterLibraryIDs || [])[0] || Zotero.Libraries.userLibraryID);
+			await this.selectByID('L' + libraryToSelect);
 		}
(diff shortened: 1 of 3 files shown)


abaevbog:
I am hoping this may resolve the issue described in https://forums.zotero.org/discussion/comment/488420/#Comment_488420. In the debug log provided in the last post, there are no errors, except for a few non-descriptive `[JavaScript Error: "uncaught exception: Object"]` at the top. I could reproduce a similar behavior (non-responsive dialog but it works fine if you are inserting a note) with two opened unloaded tabs belonging to different libraries. Then, I got `"Item ... not yet loaded"` error coming from the item in the library that's not currently opened.

In the last beta, if I generate a debug output in the scenario above, I get an output similar to the one provided in the forums post - with the `[JavaScript Error: "uncaught exception: Object"]` at the top and no stacktrace. However, if I do the same on a build from `main`, that line will be followed by an actual stacktrace like `[JavaScript Error: "uncaught exception: Object" {file: "chrome://zotero/content/xpcom/data/dataObjects.js" line: 123}]...`. 

The stacktrace is helpful, so I am wondering why that could be the case? Is it intentional? Does it dependent on if the debugger is enabled?

abaevbog:
https://forums.zotero.org/discussion/comment/488885/#Comment_488885 awesome, that was it!

PR #5173, opened 2025-03-31
File renaming: add support for counting creators

### LINKED ISSUE #4575: File renaming: enable counting authors
Based on the feedback we've been receiving, it would be useful to allow running certain parts of the template based on the number of authors.

https://forums.zotero.org/discussion/116752/specific-zotfile-workflow-with-zotero7-file-renaming-rules
https://forums.zotero.org/discussion/116854/option-for-et-al-instead-of-et-al-in-renaming-rules

To support this, we could add a new variable, authorsCount, and new "common" attributes such as `isEqual`, `isGreaterThan`, etc., making the following syntax possible (without using the regex hack I've posted in the above threads):

````
{{ if {{ authorsCount isGreaterThanOrEqual="2" }} }}
{{ authors max="1" suffix=" et al" }}
{{ else }}
{{ authors join=" & " }}
{{ endif }}
````

Thoughts?

### PR DESCRIPTION
**Pull request by tnajdek**
Adds support for `authorsCount`, `editorsCount`, and `creatorsCount` variables, along with a set of attributes to allow relational operations. This enables the ability to format file names differently depending on the number of authors, a feature that has been requested multiple times on the forums.

Close #4575

### CODE REVIEW

File: chrome/content/zotero/xpcom/attachments.js [hunk 1]
```diff
@@ -2339,6 +2339,24 @@ Zotero.Attachments = new function () {
 			return value;
 		};
 
+		const relationalOperators = (value, { isGreaterThan = null, isGreaterThanOrEqual = null, isLessThan = null, isLessThanOrEqual = null } = {}) => {
```
dstillman, line 2342: "Equal" → "EqualTo"

tnajdek: Fixed!

But thinking about this again, maybe it would be better to modify the template engine to allow relational operators so the syntax would look like this:  

`{{ if authorsCount >= 2 }}`  

The reason for this is that it's confusing to use `==` for equality while relying on descriptive arguments for inequality.  

Instead, I could extend template engine here:

https://github.com/zotero/zotero/blob/7a467b8bfa24ebd78bedbce389aa9678a8150914/chrome/content/zotero/xpcom/utilities_internal.js#L2197-L2203

and we would document this feature only for numbers. (For strings, it would follow JavaScript's default behaviour of comparing strings alphabetically, but we would not document that.)

I've now convinced myself that this would be a better approach. Thoughts?

tnajdek:
@dstillman please see my comment underneath your review for why I think #5175 is a better approach.

Issue #5180, opened 2025-04-03
Run JavaScript: Show something other than "undefined"?


AbeJellinek:
This confuses people:

https://forums.zotero.org/discussion/comment/488999/#Comment_488999
https://forums.zotero.org/discussion/comment/487195/#Comment_487195
https://forums.zotero.org/discussion/comment/483577#Comment_483577

"Completed successfully"? Or just make the text green? I don't know!

PR #5181, opened 2025-04-03
runJS: better message if nothing is returned

abaevbog:
Display "Completed successfully" instead of current undefined, which is a bit misleading.

Fixes: #5180

@AbeJellinek I agree that showing `undefined` has always been a bit misleading. Displaying "Completed successfully" at the very least conveys that everything ran and nothing broke.

While we are at it, I wonder if it would be worth to maybe also add a timestamp before the return value? While rerunning the code more than once, I find that it's not quite clear that rerunning actually happened, since the return value does not change. If there is a prefixed timestamp, something would look different on every run. Or would it just be confusing to have something go before the return value?

### DIFF
--- a/chrome/content/zotero/runJS.html
+++ b/chrome/content/zotero/runJS.html
@@ -28,8 +28,9 @@
 			</div>
 			
 			<div class="textbox-container">
-				<div class="textbox-header">
-					<label id="result-label" class="textbox-label" for="result" data-l10n-id="runJS-result"/>
+				<div class="textbox-header result-header">
+					<label id="result-label" class="textbox-label" for="result" data-l10n-id="runJS-result"></label>
+					<div id="loading-spinner" class="zotero-spinner-16"></div>
 				</div>
 				<textarea id="result" readonly aria-disabled="true"></textarea>
 			</div>
--- a/chrome/content/zotero/runJS.js
+++ b/chrome/content/zotero/runJS.js
@@ -14,6 +14,8 @@ async function run() {
 	var isAsync = document.getElementById('run-as-async').checked;
 	var result;
 	var resultTextbox = document.getElementById('result');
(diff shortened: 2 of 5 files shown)


abaevbog:
Right, good point! I think spinner might work better than just a timestamp because a timestamp doesn't help to indicate that a long-running operation is in progress. This is the behavior now, what do you think? Note that I am hiding the spinner after a small delay when everything completes so it briefly appears even if the code executes instantly. That way, there is always a visible indication that the code did run:

https://github.com/user-attachments/assets/afc84d39-3f4f-4c7f-9825-e8f34cea7e16

dstillman:
I don't think we should hide "undefined" — if you run `Zotero.foo`, you should see that it's an undefined value, not "Completed successfully". And given that we display string results without any adornment, it also just seems a little weird to display a message there.

But we could maybe replace `===>undefined<=== (undefined)` with `===>undefined<=== (completed successfully)`? Inconsistent with `===>1<=== (number)`, and may not help, but better than nothing, particularly in combination with the spinner?

abaevbog:
Yeah, that's fair. We should keep at least some kind of indication that the return value is `undefined`, as opposed to something else. `===>undefined<=== (completed successfully)` sounds like a great compromise to me - you get the correct return value but at the same time there is a hint that everything ran successfully. It is inconsistent but I think the added clarity is worth it.

This is what it looks like now:


https://github.com/user-attachments/assets/34221d91-fe13-4b3d-a46c-384eb4e8316b


The one alternative approach can be to add a verbal "Completed" where the spinner goes after the run is finished? That way, we would say that everything ran ok without any inconsistencies in terms of what is returned.

Something like this, maybe? Not sure if this looks strange though.

https://github.com/user-attachments/assets/21ec5038-0b4c-417c-ac58-b1f06f0aad2a

PR #5182, opened 2025-04-03
Citation dialog: search by doi, issn or isbn

abaevbog:
- if the search string looks like an identifier, use it as a search condition. https://forums.zotero.org/discussion/comment/489309/#Comment_489309
- have library use "All fields and tags" search mode. Followup to https://github.com/zotero/zotero/commit/b78bf8e24b1285855c9ff9ac680e6d07dfed268a

https://github.com/user-attachments/assets/3ef57f15-914d-496a-8bb4-31aa64eb4a67

### DIFF
--- a/chrome/content/zotero/integration/citationDialog.js
+++ b/chrome/content/zotero/integration/citationDialog.js
@@ -645,7 +645,7 @@ class LibraryLayout extends Layout {
 			},
 			isSearch: () => true,
 			isSearchMode: () => true,
-			setSearch: str => collectionTreeRow.setSearch(str),
+			setSearch: (searchText, mode) => collectionTreeRow.setSearch(searchText, mode),
 			ref: collectionTreeRow.ref
 		});
 		await this.itemsView.setFilter('search', SearchHandler.searchValue);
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/integration/citationDialog/searchHandler.mjs
```diff
@@ -195,7 +195,13 @@ export class CitationDialogSearchHandler {
 
 	cleanSearchQuery(str) {
 		str = str.replace(/ (?:&|and) /g, " ", "g").replace(/^,/, '');
-		str = this._cleanYear(str);
+		let issn = Zotero.Utilities.cleanISSN(str);
```
dstillman, line 198: An ISSN identifies a publication, not an item, so that shouldn't be here.

chrome/content/zotero/integration/citationDialog.js
```diff
@@ -645,7 +645,7 @@ class LibraryLayout extends Layout {
 			},
 			isSearch: () => true,
 			isSearchMode: () => true,
-			setSearch: str => collectionTreeRow.setSearch(str),
+			setSearch: (type, str) => collectionTreeRow.setSearch(type, str),
```
dstillman, line 648: Wouldn't matter, but wrong order. Should be `(searchText, mode)`.

chrome/content/zotero/integration/citationDialog/searchHandler.mjs
abaevbog: Ah, sorry about that. I removed it and repushed the commit to fix the message.

chrome/content/zotero/integration/citationDialog.js
abaevbog: Right, thank you. I swapped them in the repushed commit.

Issue #5199, opened 2025-04-10
Improve opening of a link in note editor via keyboard


abaevbog:
https://forums.zotero.org/discussion/122292/feature-request-allow-alt-enter-to-open-links-in-notes-for-accessibility

It is possible to open a link via keyboard but quite cumbersome and also not clear how if you rely on screen readers. One can navigate the popup that appears when the cursor is inside of a link by pressing Shift-Tab 3 times and then press Enter to open the link. The problem is there is no announcement when the popup appears, so if you rely on screen readers, you can only hope that it's there. Also, the popup does not appear when the cursor is right after the link - though screen readers will still announce that position as being on a link component.

So perhaps we should add an announcement that the popup is there, and have it appear when the cursor is right after the link too?

Maybe opening the link on Option/Ctrl+Enter keypress when the cursor is on the link is a nice addition as well? So that one does not have to navigate the popup to open it.

abaevbog:
These 2 PRs add an announcement when the popup appears telling the user that the poup is open. I think this already would make a difference. 

We can then consider additional edits, like opening the link on Option/Ctrl+Enter or opening the popup when the cursor is right after (but still adjacent to) the link, if needed.

PR #5205, opened 2025-04-11
add string for note-editor a11y link popup message

abaevbog:
String read out by screen readers when the link popup appears instructing users to navigate it.

Needed for https://github.com/zotero/note-editor/pull/65
Addresses: #5199

Not sure if the message needs to be more detailed (e.g. "Use Shift-Tab to navigate it to remove/edit/open the link") or if it's better to just be concise.

### DIFF
--- a/chrome/locale/en-US/zotero/zotero.properties
+++ b/chrome/locale/en-US/zotero/zotero.properties
@@ -1355,6 +1355,7 @@ noteEditor.insertColumnAfter = Insert Column Right
 noteEditor.deleteRow = Delete Row
 noteEditor.deleteColumn = Delete Column
 noteEditor.deleteTable = Delete Table
+noteEditor.a11yLinkPopupAppearedAlert = Link popup appeared. Use Shift-Tab to navigate it.
 
 pdfReader.annotations = Annotations
 pdfReader.showAnnotations = Show Annotations


dstillman:
Do we not yet have a way to use FTL strings in note-editor?

abaevbog:
I don't believe so... Fluent is not used anywhere in `note-editor` currently, and if I do `document.getElementById("editor-view").contentDocument.l10n` when a note is opened, `null` is returned. That is in contract to top-level `document.l10n` or `document.querySelector(".reader").contentDocument.l10n`, which both return a fulfilled promise of `DocumentL10n`.

dstillman:
I think reader is not remote (currently, though we need to change that), so direct access is possible there. For note-editor, we might need to set up async message-passing to get the strings. If note-editor expects to have strings synchronously, that could be a little ugly.

Or we could have a dedicated .ftl file, make that available to the editor, and use https://github.com/projectfluent/fluent.js/, but that's almost certainly more complicated.

We can merge this for now, but @mrtcode, would be good to think about switching to .ftl, since I imagine we'll need to at some point.

abaevbog:
Oh, wait a second. If I add `<link rel="localization" href="zotero.ftl" />` to the `editor.zotero.html`, it actually does seem to work! I may have been looking at the wrong `.html` file from `note-editor` before. So I guess we would want to do that instead, and then fetch the correct string via `l10n.formatValue`?

abaevbog:
Well, actually though, I'm not sure how this fluent approach would work on other platforms? We could do a check similar to what we do in the reader, where if `document.l10n` exists, we set the message, and otherwise do nothing - at least for now. It would likely not be relevant for ios and android, so it's just about how web library would be able to handle the `document.l10n`.

dstillman:
And without that, it just uses the bundled English strings, right? So we'll need to do something different to localize the reader and editor on other platforms, but that's it.

abaevbog:
Yes, without that (as it does in the current state of the linked PR), it uses `intl.formatMessage`, which fetches bundled `zotero.properties` strings, at least on desktop.

So we could do something like that? Use `document.l10n` if it's there, or default to `intl.formatMessage` otherwise.
```
{ document.l10n ? 
    <div id="a11y-alert" aria-live="polite" data-l10n-id={editorState.link.popup.active ? "note-editor-a11y-link-popup-appearedalert" : ""}></div>
    : <div id="a11y-alert" aria-live="polite">{editorState.link.popup.active ? intl.formatMessage({ id: 'noteEditor.a11yLinkPopupAppearedAlert' }) : ""}</div>
}
```

dstillman:
Sorry, I mean the duplicate bundled English strings in the reader (and editor?) submodule. We're not using localization at all in the other environments.

mrtcode:
> I think reader is not remote (currently, though we need to change that)

We need to be sure it's worth doing this, because the PDF view uses a worker for part of the processing, and I believe the EPUB view does something similar—possibly using a separate iframe (@AbeJellinek can explain this better). Losing synchronous access to the reader would be quite inconvenient for both us and plugin developers.

Additionally, memory usage is a bigger concern.

abaevbog:
> Sorry, I mean the duplicate bundled English strings in the reader (and editor?) submodule. We're not using localization at all in the other environments, so it needs those strings in order to show anything.

Ah, right! I was a bit confused about this mechanism. Now I see that in the desktop app, localized strings are [passed](https://github.com/zotero/zotero/blob/a7778b93e841ce62773efc65cd86576c7bfe8af1/chrome/content/zotero/xpcom/editorInstance.js#L197) from `zotero.properties`, and `en-us.strings.js` is used for all other platforms (e.g. [web](https://github.com/zotero/note-editor/blob/ae92348364daec4b6877e6805b54a5cbcf145741/src/index.web.js#L7)). 

> So I'm saying if we can use Fluent in Zotero and continue using the in-submodule strings otherwise for now, that's fine.

Great. I updated the `note-editor` PR to use fluent in the desktop app and fetch strings via `document.l10n.formatValue`. If `document.l10n` is not available, we'll fetch strings from `en-us.strings.js` via `intl.formatMessage`, the way it's done currently. I moved all of this into a new `A11yAlert` component, to not clutter `src/ui/editor.js`.

Issue #5202, opened 2025-04-10
Cannot navigate collections dropdown via keyboard in "New Collection" dialog on windows


abaevbog:
https://forums.zotero.org/discussion/123329/accessibility-creating-new-collections-keyboard-problem

This does not occur on a mac. On windows, once the collections dropdown is focused, space does not open it. Strangely, arrow up/down can change the selected collection once or twice (without opening the popup), after which they also stop having any effect.

abaevbog:
Looks like on windows, space does not open a `<menulist>`  in general but Alt+ArrowDown does. It works to open the collections menu but arrows do not navigate the menu.

abaevbog:
You see the menu jumping behind the dialog too? Great! I wasn't sure if it would end up being related to keyboard nav or not, so I added a separate #5204 for it

abaevbog:
I'll also note that navigating `menulist`s on windows in general feels quite off to me. If the menu does open (e.g. in preferences, or item type dropdown in `itemBox`), attemting to navigate menuitems with arrowUp/Down immediately changes the selection of the `menulist`, which then closes the menu. So I seemingly can't move farther than 1 menuitem from where I started before the menu hides.

For instance, to get to "Bill" item type, I would need to press ArrowUp twice. But as soon as I press ArrowUp once, this is what happens:

https://github.com/user-attachments/assets/47d3d3f7-4d73-4f27-a2db-ec937452c662

PR #5208, opened 2025-04-14
Fixed non-working selection of collections in "New Collection" dialog via keyboard on Windows.

abaevbog:
Fixed non-working selection of collections in the dropdown via keyboard on Windows.

Fixes: #5202

### DIFF
--- a/chrome/content/zotero/newCollectionDialog.js
+++ b/chrome/content/zotero/newCollectionDialog.js
@@ -48,17 +48,18 @@ var Zotero_New_Collection_Dialog = {
 	_updateMenu() {
 		let createInField = document.querySelector('#create-in');
 		let menupopup = createInField.firstElementChild;
-		let id = menupopup.id;
-		// Fascinatingly, clearing the children of the menupopup isn't enough here.
-		// We have to completely recreate it or it will no longer be willing to open.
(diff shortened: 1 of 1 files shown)


abaevbog:
@AbeJellinek I slightly reworked the "New Collection" dialog here to not rebuild the entire menu on every `command` event, which was breaking keyboard accessibility on windows. The issue is that on windows, the items actually get selected and `command` is fired every single time you press arrow up or down to navigate items of `menulist`. How is this looking to you?

Also, on my mac and on windows VM, the popup now appears for me properly if I just clear its children without having to re-create the `menupopup`. So I removed the `menupopup.replaceWith(menupopup = document.createXULElement('menupopup'));` bit here as well.

abaevbog:
Good point! This is the config on `7.0` branch, so it looks like we are on 115 in stable releases, yes.

https://github.com/zotero/zotero/blob/bbf4692b99b8f92618d58dafb897f7b797a987ab/app/config.sh#L3-L7

I checked out `7.0` branch, and cherry-picked this one commit. There I still had no issues with reopening the popup.

https://github.com/user-attachments/assets/66fa6821-ac79-4777-a710-fb6cb7c44be0

Issue #1819, opened 2020-05-04
Multi-section bibliographies


adomasven:
As noted in https://github.com/citation-style-language/documentation/issues/71 this is a frequently requested feature, which require a both a technical and GUI solution in CSL/citeproc clients.

Since the GUI is likely to inform the technical implementation, I wanted to discuss it separately in more detail.

(1) If we were to do this I think we should split the Add/Edit Bibliography button into two separate buttons: Add and Edit; where Add would add a full bibliography of all items in the document at the cursor location, and you could add multiple bibliographies, while Edit would open the Bibliography editor for the bibliography at the cursor and allow you to uncite items. This would be enough to make fully customizable bibliography sections.

The Edit Bibliography dialog can be further enhanced per user requests based on the use cases of multi-section bibliography, to filter for certain item types, tags, etc. (2) We could also allow users to give names to bibliography sections and allow them to mark which sections an item should belong to when inserting/editing items via the citation dialog.

For (1) we probably just would need to update uncited items for each bibliography generation separately. For (2) we might want to use the citeproc bibliography generation [filter functionality](https://citeproc-js.readthedocs.io/en/latest/running.html#selective-output-with-makebibliography) and would need to store some additional information in bibliography field codes.

dstillman:
Why not just keep Add/Edit Bibliography and have it work like Add/Edit Citation? Going back to two buttons doesn't make much sense to me.

If we're concerned that people are going to click Add/Edit Bibliography by mistake and insert a second bibliography at the cursor position, we could just prompt whether they want to insert a second bibliography when the cursor isn't within an existing bibliography, and if they say yes either insert another regular bibliography or automatically open the Edit Bibliography window so that they can set a filter. If it's a Zotero (as opposed to Word) prompt, and there's a single bibliography in the document, we could even give them a choice between inserting a new bibliography or editing the existing one.

adomasven:
> we could just prompt whether they want to insert a second bibliography when the cursor isn't within an existing bibliography

We could although it would be a jarring change in the behavior of the button without no apparent UI changes. It would also become more apparent that you can now insert multiple bibliographies if the buttons were separate.

> If it's a Zotero (as opposed to Word) prompt, and there's a single bibliography in the document, we could even give them a choice between inserting a new bibliography or editing the existing one.

We can have 3 buttons in Word prompts, so this could be done with them too.

denismaier:
[...]

PR #5210, opened 2025-04-15
Edit bibliography dialog redesign in works

### LINKED ISSUE #5179: Update edit bibliography dialog?
Now that we have a new citation dialog, perhaps it's worth updating the `editBIbliography` dialog too? 

First, it looks slightly off with all the space above the `collectionTree` and no differentiation in background color between collections and items table. There is a mockup in Figma that looks much nicer.

Second, the `->` and `<-` icons used to add/remove items are somewhat glitchy and also are hard to use with just a keyboard. The same mockup in Figma replaces the two arrow buttons with `+` in the itemTree to add the items and `-` icon in the right-most list to remove it, which is more intuitive.

I am realizing that `editBibliography` dialog is very similar to now-removed `classicCitationDialog`. It has very similar functionality to the current `citationDialog` of searching for items and adding/removing them from the list.  Besides, the Figma mockup has the same layout as the library mode of the `citationDialog`. So I am wondering if it would make sense to adapt the `citationDialog` to accommodate the bibliography edits, instead of working with `editBIbliographyDialog` as a separate entity and rebuilding most of the library mode there? Could we display the items in bibliography as bubbles and fully reuse the existing library layout? `simpleEditor` can be still added at the bottom, if the `citationDialog`'s approach of adding prefix and suffix is not sufficient.

Not sure if there are any additional details I am not aware of so correct me if I am wrong (cc @adomasven @dstillman )

abaevbog:
- updated the dialog's layout to remove empty space around the edges
- replaced the references list with a proper itemTree
- removed selectItemsDialog.js as a dependency. Some behaviors of selectItemsDialog.js conflict with what needs to happen in editBibliographyDialog. For example, custom logic is needed to save column preferences of two itemTrees in the same window. Also, onActivate handler of itemTree does not need to accept the entire bibliography dialog, even though it makes sense when a single item is being selected.
- move handling of clickable buttons from citationDialog into itemTree. It allows us to also add +/- buttons to itemTrees in this dialog.
- library itemTree has a column with + button to add the item into the list of references. It disappears if the item is already included
- references itemTree has a column with - button to remove the item from the list of references. If the item is not cited, it is removed from this table. If the item is cited, it will remain but will become
faded. It's column button will change to +, which will re-add that item into bibliography.
- references itemTree also has a "Edited" column which indicates of the citation has been manually changed by the user or not, and "Cited" column which indicates if the item is cited in the document or if it was added without a citation.
- there is a splitter between library trees and the references itemTree. It can be dragged to make
more space for references and can be collapsed to hide library trees completely.

Additional minor fixes:
- do not display tooltip when a button in virtualized-table is hovered over
- cleanup of integration.css to remove unused styles from editBibliographyDialog and citationDialog

Fixes: #5179

https://github.com/user-attachments/assets/da158511-28ac-4f0c-b62d-5ef0acdcc930


abaevbog:
@adomasven this is the first iteration of the updated edit bibliography dialog. It still needs some polishing but I wanted to get your feedback on the overall functionality. 

Among a few questions I had were:
- what's the best way to indicate if an item was updated with custom text? For now, I just put "yes" or "no" into the "Edited" column, but maybe we would want some kind of icon in there? At the same time, nothing is as expressive as yes/no, so not exactly sure.
- the cited items always appear in the references list, they just become faded if they are excluded. Do you think it is clear what exactly is going on without additional explanation?
- we would like to show which items are added without a citation, so I added another yes/no column "Cited" for that. With "Edited" and +/-, we end up having 3 supplemental columns, which may seem like too much? At the same time, I can't think of a more effective way to do this. We could highlight those rows with some additional color, I guess? Or, since this dialog isn't used too often, perhaps a few extra columns is not too big of an issue?
- I checked if there were any older issues related to edit bibliography dialog, in case we could cover them here as well. There was a discussion on multi-section bibliographies a while back https://github.com/zotero/zotero/issues/1819. Is it still somewhat relevant? There was a discussion of additional filters type, and etc., so let me know if there's anything helpful we could add here for it.

adomasven:
> * what's the best way to indicate if an item was updated with custom text? For now, I just put "yes" or "no" into the "Edited" column, but maybe we would want some kind of icon in there? At the same time, nothing is as expressive as yes/no, so not exactly sure.

I had checkmark and cross for https://github.com/zotero/zotero/pull/3468, but arguably it makes less sense here. In citation explorer you want to make sure items are linked to your library, so that's the correct "check" state, whereas here it's more informational.

> * the cited items always appear in the references list, they just become faded if they are excluded. Do you think it is clear what exactly is going on without additional explanation?

It's pretty clear, and overall still better than what we have now where they disappear completely from the list

> * we would like to show which items are added without a citation, so I added another yes/no column "Cited" for that. With "Edited" and +/-, we end up having 3 supplemental columns, which may seem like too much? At the same time, I can't think of a more effective way to do this. We could highlight those rows with some additional color, I guess? Or, since this dialog isn't used too often, perhaps a few extra columns is not too big of an issue?

I think it's fine. Colors would be too ambiguous and the columns don't take up that much space.

> * I checked if there were any older issues related to edit bibliography dialog, in case we could cover them here as well. There was a discussion on multi-section bibliographies a while back [Multi-section bibliographies #1819](https://github.com/zotero/zotero/issues/1819). Is it still somewhat relevant? There was a discussion of additional filters type, and etc., so let me know if there's anything helpful we could add here for it.

It would be a good feature, but quite niche. On the other hand implementing it from the technical viewpoint is non-trivial both in terms of UI/UX solutions and handling of citation fields in docs. Let's not think about it for now.

chrome/content/zotero/itemTree.jsx
```diff
@@ -3160,8 +3162,9 @@ var ItemTree = class ItemTree extends LibraryTree {
 					this.onDrop(e, index);
 				}, { passive: true });
 			}
-			div.addEventListener('mousedown', this._handleRowMouseUpDown, { passive: true });
-			div.addEventListener('mouseup', this._handleRowMouseUpDown, { passive: true });
+			div.addEventListener('mousedown', this._handleRowMouseUpDown, true);
+			div.addEventListener('mouseup', this._handleRowMouseUpDown, true);
```
adomasven, line 3166: Changed event listener from passive to capture. See docs https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

I don't remember why I added passive here specifically, but it had something to do with improving performance, maybe drag performance, and should be adjusted carefully and only with good reason.

chrome/content/zotero/itemTree.jsx
```diff
@@ -3193,6 +3196,55 @@ var ItemTree = class ItemTree extends LibraryTree {
 		if (this.collectionTreeRow.isDuplicates() && !modifierIsPressed) {
 			this.duplicateMouseSelection = true;
 		}
+
+		// Handle mouseup and mousedown events on a row in itemTree to enable clicking on +/- button
+		// On mousedown, add .active effect to the +/- button
+		// On mouseup, add/remove the clicked item from the citation
+		// This specific handling is required, since :active effect fires on the row and not the child button
```
adomasven, line 3203: The :active event doesn't fire on the children because CSS `pointer-events: none` is set. See https://github.com/zotero/zotero/commit/8e2790e2d2a1d8b15efbf84935f0a80d58db4e44 . Since we don't need to support drag and drop here, we might be able to simplify the handling by enabling pointer-events. Furthermore, maybe with the current version of firefox setting `pointer-events: none` is no longer even needed.

dstillman:
Just going by the screen recording above, this doesn't really seem usable to me — there's just not enough horizontal space to have two item trees with full columns. You shouldn't have to slide a splitter around for basic usage of the dialog.

I agree that people mostly shouldn't be using this dialog and therefore we shouldn't be spending much time on it, but since we're here:

Instead of giving the whole bottom of the dialog to an editor that we really don't want people to use, we could use the bottom for the references. What if we showed the actual bibliography, with buttons/checkmarks on the side? So then you'd see each entry right there without needing to click on it, and if you really wanted to edit it, you could do so inline, and the row would then update to reflect that it was edited, with an option to revert.

@adomasven, do we have reliable entry markers (`.bib-entry`?) for all citation styles, such that we could add buttons to each row?

@yexingsha, do you want to look at this?

adomasven:
> Instead of giving the whole bottom of the dialog to an editor that we really don't want people to use, we could use the bottom for the references.

I've considered something like this, but some things are not as obvious to do, like showing entries that are excluded from the bibliography, or highlighting which ones are not cited. Overall seems like too much work for this dialog.

> @adomasven, do we have reliable entry markers (`.bib-entry`?) for all citation styles, such that we could add markers/buttons to each row?

citeproc returns a sorted list of bibliography entries, so we can do whatever here. I think implementation-wise it would be easier to keep them as separate input fields, instead of making the whole bibliography a single editable textarea. But also an editable bibliography list might give people wrong impressions, like trying to remove entries by removing them in said textarea.

dstillman:
> some things are not as obvious to do, like showing entries that are excluded from the bibliography, or highlighting which ones are not cited

I think it could be the same as what we have here — we'd add columns for cited/excluded (and an excluded item could be gray, for example).

> I think implementation-wise it would be easier to keep them as separate input fields, instead of making the whole bibliography a single editable textarea

Yes, they would be separate rows, in the same sorted order as the bibliography. I was just asking whether we're always able to pull out the separate entries regardless of the citation style.

dstillman:
I think we also have to be realistic about people editing entries, and this would be a pretty elegant way of doing that — and much more transparent than what we have now. You could just start editing any bibliography entry inline in the bottom pane, and once you did, the Edited column would show a checkmark and a revert icon next to it that would reset it back to its unedited state. (We could also show the warning at the bottom once any entry was edited.)

dstillman:
We're settled on showing a diff in https://github.com/zotero/zotero/pull/5120 too, but a diff probably gets in the way of making the entries directly editable.

abaevbog:
All of this makes a lot of sense! My understanding is that the course of actions now is:
- reverting most itemTree-related changes here, since we will won't have two itemTrees
- removing the editor at the bottom
- instead of the editor, I'll add something like a `richlistbox`, where each row will have a title, a mini-editor with rendered bibliography entry for that item, and columns for edited/cited/not cited
- add a horizontal splitter to be able to make that list taller. Not sure if we need to keep the vertical splitter between the itemTree and collectionTree, now that they will occupy the entire width of the dialog. 

Correct me if I am missing anything.

> I've considered something like this, but some things are not as obvious to do, like showing entries that are excluded from the bibliography, or highlighting which ones are not cited. Overall seems like too much work for this dialog.

I suppose it does end up being a bit more work than initially anticipated. At the same time, I do see [forums posts like this one ](https://forums.zotero.org/discussion/comment/489802/#Comment_489802 ) every now and then which just boil down to the dialogs UI overall not being very intuitive. So over time, it might be worth the effort?

dstillman:
> I'll add something like a `richlistbox`, where each row will have a title, a mini-editor with rendered bibliography entry for that item, and columns for edited/cited/not cited

There wouldn't be a title and rendered bibliography entry — just the latter. I'm imagining something that basically looks like a regular bibliography, except there are a couple icons/buttons next to each entry.

We should get a mockup from @yexingsha for the details, though.

yexingsha:
I really like this idea! I tried two different approaches. The first one is largely based on the current implementation:

table-hovered

table-editing

- I think the biggest issue with this approach is it's too table-like and doesn't really look like a bibliography. The rows look like they are selectable but not editable.
- To compensate for this, I think we would need to make the bib entries into something like metadata fields, with an hover effect to indicate they can be edited.
- If an entry is edited, a `revert` icon button appears at the right of the entry column. Clicking on this button will revert the entry and remove the button.
- The Cited column, the add/remove buttons, and the faded text to indicate excluded citations are the same as in the current implementation.
- I think a floating editor toolbar makes the most sense in this layout. A potential issue is people won't know its existence until they select some text. We can auto select all the text when people click on an entry so the toolbar always appears, but that might be too bothersome.
- I'm trying out background colors in the item tree to better highlight items included in the bib. Blue means cited items, which is the same as in citation dialog; green means uncited items that are manually added to the bib. But we might need tooltips to explain this.

The second one (which I prefer) is more bibliography-like:

list-colored

list-uncolored

- No need for column headers, so editor buttons can stay where they are. I also put the warning text in there to save some space.
- The bib entries are visually not in distinct text fields, so no hover or focus effects.
- Revert and Add/Remove buttons work the same as in the first approach.
- Cited/Uncited become labels at the beginning of each entry. I'm trying out colored labels that match the highlighted rows in itemTree, and uncolored labels/rows. Let me know which one you prefer.

adomasven:
Great work @yexingsha! The second version with colors is great! Some notes:

- I think we will want tooltips to appear for the "Cited" and "Uncited" entries, with explanatory text like "This entry appears in the bibliography because it is cited in the body of the document" and "This entry appears in the bibliography because you manually added it in this dialog". We might also want a label for excluded entries, "Removed" or "Excluded", and possibly a third color - red?
- +/- Buttons should be available for cited and uncited entries in the library view
- It would probably be useful to synchronize the focus of top and bottom sections, such that if you select a line in one, the other also changes selection

yexingsha:
> I think we will want tooltips to appear for the "Cited" and "Uncited" entries, with explanatory text like "This entry appears in the bibliography because it is cited in the body of the document" and "This entry appears in the bibliography because you manually added it in this dialog".

Sounds good, but maybe these tooltips should only appear when hovering over the labels. It might be too annoying when they appear over the entry text. 

> We might also want a label for excluded entries, "Removed" or "Excluded", and possibly a third color - red?

> +/- Buttons should be available for cited and uncited entries in the library view

These are great suggestions, since a third color also means clicking the +/- button on cited items in itemTree will produce an obvious effect. I think "Excluded" works better, and orange is warning-like color without being too much like an error:

bib

> It would probably be useful to synchronize the focus of top and bottom sections, such that if you select a line in one, the other also changes selection

I guess the synchronization will look like this (please correct me if I understood incorrectly):

- Editing an entry below (which simply means clicking while the cursor is on the entry text, with no visible focus state other than a blinking text cursor) will select the corresponding item above. This can involve switching to a different collection and scrolling when necessary;
- Selecting a colored item above will select the corresponding entry text below.

dstillman:
1. I think the "If you edit an entry" line should stay at the bottom, even if it's a less efficient use of space. It's a bit distracting and messy to have it in between the two panes.
2. I worry that the revert icon alone may not be clear enough to convey that a row has been edited. An alternative might be a diff icon that popped up a diff in a little panel on hover, with a revert button in the panel. But maybe that's overkill and a tooltip on the revert button would be sufficient.
3. The colors are fun, though it's worth keeping in mind that the default cause will involve 100% cited items. So we just want to make sure we're OK with this styling even when there are 300 rows that all say "Cited" next to them.
4. Synchronizing selection from top to bottom makes sense, but I don't think the reverse is necessary. Seems weird and annoying that it would discard your current selection and potentially switch to My Library just because you clicked within the bibliography. And I'm not sure what good that does you, whereas if you have a long bibliography it definitely might be quicker to find a bibliography entry via the library browser.

abaevbog:
This looks great and makes a lot of sense to me! I am focused on just putting together the overall functionality at the moment but a few questions/notes I have so far:
- sorry if this is obvious but what exactly is the expected behavior with the synchronization of focus between top and bottom parts of the dialog? If I select a cited row from itemTree, is the idea to immediately move focus to the editor for that entry? I think I am missing something here.
- I agree with the colors - they look nice but even the mockup looks slightly busy to me with 3 different colors. It would actually be 4 colors, including the default dark blue selected row. It does get the message across love effectively though, so I suppose we'll just have to see what it feels like.
- for the editors, am I correctly going with the same `simpleEditor.xhtml` iframes? I added a few minor modifications for this use case and it generally works ok. But it's no longer used anywhere else, so I want to make sure there is nothing I should be using instead.

dstillman:
> If I select a cited row from itemTree, is the idea to immediately move focus to the editor for that entry?

I think we'd want to scroll the bibliography entry into the middle of the view and highlight it somehow so you could make changes to it quickly but not actually move focus there.

> am I correctly going with the same simpleEditor.xhtml iframes?

It looks like simpleEditor.xhtml is just a wrapper around contenteditable, which seems appropriate to me. We'll just have to make sure it doesn't fall down with 300+ rows initialized at once. If it does, we could maybe set only a window of rows as contenteditable.

yexingsha:
> I worry that the revert icon alone may not be clear enough to convey that a row has been edited. An alternative might be a diff icon that popped up a diff in a little panel on hover, with a revert button in the panel. But maybe that's overkill and a tooltip on the revert button would be sufficient.

Yeah, I think that would be overkill. By default you don't see the revert buttons, and once you edit an entry you will see it appear immediately. I think that plus a tooltip should be sufficient to convey how it works.

> The colors are fun, though it's worth keeping in mind that the default cause will involve 100% cited items. So we just want to make sure we're OK with this styling even when there are 300 rows that all say "Cited" next to them.

That is true, and usually you shouldn't exclude cited items from your bibliography either. So what if by default we don't show the labels at all, and they show up only when you have excluded or uncited items? We can always keep the light blue background for cited items in itemTree, since it's helpful and consistent with the citation dialog. So a default case where cited items match 100% with the bibliography would look like this (with the helper text moved to the bottom):

bib-default

abaevbog:
> I think we'd want to scroll the bibliography entry into the middle of the view and highlight it somehow so you could make changes to it quickly, but not actually move focus there. And nothing in the other direction.

Ok, yeah, that makes sense!

> It looks like simpleEditor.xhtml is just a wrapper around contenteditable, which seems appropriate to me. We'll just have to make sure it doesn't fall down with 300+ rows initialized at once. If it does, we could maybe set only a window of rows as contenteditable.

Right, it's just `contenteditable` with some additional tweaks for bold/italics/underline styling. It does rely on some deprecated methods like `document.execCommand` - but I see it being used in firefox, so it's probably ok. I'll make sure to check that it doesn't break when we have a large number of items.

dstillman:
> That is true, and usually you shouldn't exclude cited items from your bibliography either. So what if by default we don't show the labels at all, and they show up only when you have excluded or uncited items?

Yeah, I think that makes sense. But then we'd probably want to move the excluded/uncited labels to the right side and just wrap the row before the label?

And, actually, that would be a solution to my concerns about the revert button too — since we could then easily have multiple labels, we _could_ add an "Edited" label, immediately to the left of the revert button.

dstillman:
Though I guess showing a label would affect wrapping, so it might be awkward to do that while you were editing…

yexingsha:
> But then we'd probably want to move the Excluded/Uncited labels to the right side and just wrap the row before the label?

> And, actually, that would be a solution to my concerns about the revert button too — since we could then easily have multiple labels, we could add an "Edited" label, immediately to the left of the revert button.

I guess this is what it would look like:

bib-right-label

I think placing the labels to the right and omitting "Cited" can work, but "Edited" together with the revert button just feels like stating the obvious. And changing wrapping while editing is indeed very awkward.

dstillman:
OK, let's try labels on the right but no Cited or Edited.

abaevbog:
OK, so this is the first iteration of the new interface with the list of editors at the bottom. What does everybody think?

Screenshot 2025-04-24 at 3 10 35 PM


@yexingsha, I think the colors don't look that busy after all. I think I quite like it. The only thing I am wondering about is if green and red sends a somewhat wrong message? When both uncited and excluded items are present, it almost looks like success and failure indications. The fact that colors on the rows match the label clarifies this a bit but still I wonder if maybe instead of green we do something like an even lighter shade of blue? (I don't feel strongly about this, just a thought).

@dstillman, I tested the performance of this with a bibliography of 350 items. Overall, loading many iframes with simpleEditor is not too bad but noticeable. It was particularly clear on initial load, where it would take ~4 seconds for the dialog to load vs ~1 second with 10 items. There are many ways we can optimize this (e.g. a windowed list or using `simpleEditor.js` directly in dialog's DOM, without any `iframe`s). For now, I went with creating each bibliography row with an empty placeholder instead of the `iframe` editors, then loading the first 10 `iframe`s immediately in `onLoad` listener, and all remaining iframes are loaded in `requestIdleCallback`. With this, the initial loading time goes back to 1-2 seconds for me. What do you think about this? I know a bibliography can be very big - but at which point would you say we would be approaching an extreme number of items? 1000? Besides that, I am noticing a bit of extra delay when separate items are being added and removed - but that feels similar to the delay in the current dialog.

@adomasven, now the cited items that are excluded still appear in the bibliography list - they are just faded out. So we need to be able to get the bibliography entries for those items too, even though they are not in the actual bibliography. To get their bibliography entries in the right order, I ended up creating a temp duplicate `cslEngine`, which creates a bibliography from both actual reference items and excluded cited items. Does this make sense? Is there a more correct way to get the text of cited items and their position?
https://github.com/zotero/zotero/blob/17ca79395a35d42f6257e3d4f7c9fedd4abeb98e/chrome/content/zotero/integration/editBibliographyDialog.js#L497-L505

chrome/content/zotero/itemTree.jsx
abaevbog: Same as below, apologies for late reply!

Yes, these are just listeners from citation dialog that I combined with the existing `_handleRowMouseUpDown` that sets `this.duplicateMouseSelection = true;`:
https://github.com/zotero/zotero/blob/3144314d20e51fb8cfb73d65548122c9fe1eb9ed/chrome/content/zotero/integration/citationDialog.js#L582-L584

In addition to them not being passive, I also changed them to be capturing. The listeners are capturing, so we can call `event.stopPropagation()` to not select the row whose button is clicked. And they are not passive because we have to be able to call `event.preventDefault()`, because otherwise focus will move into the itemTree when the button is clicked (even if it's propagation is stopped). I suppose we could handle it with additional `focusin` listeners but that seems messy.

What if, to make it less likely to cause issues, I keep the current `_handleRowMouseUpDown` from `itemTree` unchanged, and move these mouse up/down handlers  into a separate function. Then, we would only add it as capturing additional `mouseup` and `mousedown` listeners if there are columns with clickable buttons (so currently, only here and in citation dialog)?

dstillman:
I don't think we want to highlight the whole bib entry with the same color when it's selected above — that's a bit overwhelming and will impact readability. I would just have some sort of outline, similar to a focus ring (which it then would have if you tabbed from the items list).

> at which point would you say we would be approaching an extreme number of items? 1000?

Yes, 1000 would be pretty extreme but not unprecedented.

If we can drop the iframes easily, we should do that, but we can probably save additional optimizations for later.

yexingsha:
Right, the orange still looks too reddish, and the green does look too much like a success indication. Let's try yellow and cyan instead:

bib-colors

> I don't think we want to highlight the whole bib entry with the same color when it's selected above — that's a bit overwhelming and will impact readability. I would just have some sort of outline, similar to a focus ring (which it then would have if you tabbed from the items list).

I agree we shouldn't use that light blue color as background, but I'm unsure about the outline/focus ring part. Right now when you tab from the items list, you go first to the editor toolbar, then straight to the text field of the first displayed entry without giving it a focus ring, and then to its button. This feels mostly correct to me, aside from the fact that when there is a visible highlighted entry, the focus should just go to it instead of the first visible entry. But since the focus never goes to an entire entry, the outline around the highlighted entry will just stay there as long as the selected item doesn't change, and never turn into a focus ring. Which I guess can work? It just needs to be obviously different from a focus ring, and not clash with the focus ring around the +/- button that can appear inside it. So maybe a dotted line:

bib-highlight-outline

Or, maybe just change the background color to`fill-quarternary`, which is the same as when the selected item loses focus:

bib-hightlight-bg

I know this is still in progress, but here are my first impressions:

- We should increase the min height of the window to 600px, and increase the min height for the library area to at least 200px, maybe even 250px.
- We can also consider enabling resizing between the two areas, since it's not obvious to me how we should allocate the space when the window gets resized.
- The width of the collection tree should be increased to 200px.
- I'm thinking the labels don't take up a column, so that cited items can utilize that space. Otherwise, when you only have cited items (which is the norm), there is a wide column of empty space that seems wasteful.
- The entries should use the global line height (18px at normal font size). The serif font already seems smaller than the system font, so I think it wouldn't look too tight.
- Looks like the entries have a slightly larger bottom padding than the top.
- There should be 8px margin above the first entry and below the last entry, so that they aren't directly against the border when they are highlighted.
- The editor toolbar doesn't seem to work very well at the moment.

dstillman:
If the "highlight" remains when focus is in the entry's input field, I'd go for the blue dashed line, or something like it. I think it would be unpleasant to type on the gray background, and it would also be sort of weird to have something that looks like an unfocused selection state when it has focus.

chrome/content/zotero/itemTree.jsx
adomasven: > In citation dialog, one can drag the row into the `bubble-input`

Huh! I didn't even realize that. Shame adding clickable cell buttons is impossible this without restoring to these kinds of workarounds.

chrome/content/zotero/itemTree.jsx
adomasven: Testing this I don't see any performance issues, so we can try going in this direction and rethink if something comes up.

abaevbog:
> If we can drop the iframes easily, we should do that, but we can probably save additional optimizations for later.

Interestingly enough, `contenteditable` seems to not behave right in chrome windows. Without a `<browser>` or `<iframe>` to contain it, many keyboard operations just don't work right. That is the case in firefox as well - so I think we'll still need a single `iframe` but it will contain the entire list of entries.

abaevbog:
I circled back here to rebase post-fx140 and finish up the remaining edits and style fixes. I also refactored the `simpleEditor` more to be more aligned with how it is used in #5312. I believe this should be close to the desired behavior we discussed above.

As a summary, this is the current behavior:
- items in the bibliography appear in the bottom resizable section rendered as actual bibliography entries. One can edit them inline right there, after which a `revert` will appear to its right to undo those changes
- cited entries that are explicitly excluded from the bibliography will still appear in that list - their entries are just greyed out. One can re-add them back by pressing `+` icon to the right of the editor. They are also marked with an `excluded` label.
- uncited entries can be added into the bibliography by pressing `+` in the `itemTree`. They will be marked with an `uncited` label.
- items that are excluded or highlighted will also have their itemTree rows highlighted with the same color as the labels
- clicking on an itemTree row of an item that is in the bibliography will scroll to it and place an outline around the editor
- from my experiments here and while working on editable annotation cards, keeping `simpleEditor` inside of the iframe seems to be the most reliable approach. It does mean we have have many iframes in the same dialog if the bibliography is large. I added some optimization to avoid redundant re-rendering whenever possible, so when I test with a bibliography of 900 entries, it performs quite OK on my end. We could add further optimization (via windowed list maybe) if needed.
- I removed creating and using a dumplicate `cslEngine` from the earlier revision. Instead, when the dialog opens, we record `bibEditInterface.bibliography.omittedItemIDs` in a separate set, clear `omittedItemIDs` and generate the bibliography as if nothing is omitted. That way, we know where to display the cited but excluded entries. When the dialog is accepted, we remove omitted items (potentially updated in the dialog) from the bibliography. That way, there is no duplicate `cslEngine` used, and it seems to perform better too.

abaevbog:
@yexingsha, your most recent notes from https://github.com/zotero/zotero/pull/5210#issuecomment-2829333749 should also be addressed here (including the funky editor toolbar that should be more reliable now). Let me know if this looks better!

Screenshot 2025-08-29 at 11 46 27 AM

yexingsha:
Looks good! Just a few issues:

- The default height of the bib area is too small, and it shouldn't revert to its default height every time the dialog closes and reopens.
- All the confirmation dialogs have "Cancel" as the primary button, which is weird.
- Clicking the `-` button in the library seemingly deletes the bib entry, which doesn't actually happens because the citation is untouched. It should exclude the bib entry instead.
- The label and buttons are still misaligned with the first line of the bib text:
Screenshot 2025-08-29 at 15 19 19

abaevbog:
Thank you for these notes!

> The default height of the bib area is too small, and it shouldn't revert to its default height every time the dialog closes and reopens.

I increased the initial height of the bottom area to 200px. Before closing the dialog, we now also store the height in a pref, and we'll use that height on next dialog open.

> All the confirmation dialogs have "Cancel" as the primary button, which is weird.

That is a good observation. I don't see why we would want "Cancel" as the primary button, so I changes it for "OK" in remove and revert confirmation dialogs.
Screenshot 2025-08-29 at 2 20 23 PM
Screenshot 2025-08-29 at 2 20 16 PM



> Clicking the - button in the library seemingly deletes the bib entry, which doesn't actually happens because the citation is untouched. It should exclude the bib entry instead.

Ah, I was fixing an edge case elsewhere, and it caused a glitch where it would remove the entry from the bib instead of excluding it. Should be fixed now.

https://github.com/user-attachments/assets/469f43c2-f2bf-4eeb-9e2d-ecbdbb295664

> The label and buttons are still misaligned with the first line of the bib text:

I shifted some things around a bit more. I think it should look better now!

Screenshot 2025-08-29 at 2 01 06 PM

yexingsha:
Looks great now!

abaevbog:
Awesome! Thanks for looking into this with me!

dstillman:
@abaevbog: Sorry, can you rebase?

dstillman:
Er, never mind, it's just a couple comments. I'll do it.

dstillman:
1. If you select a row in the items list, it draws a dashed line around the entry in the list below. But if you click on an entry in the list below, I think it should probably show the dashed line and also select the item in the list above and scroll it into view. There's no particular functional reason for that, but it's a little weird to be able to edit an item other than the one with the focus ring.
2. Not sure if we discussed this above, but I think we need + buttons in the list to add uncited items, not just rely on double-click/Enter. It's not a common action, but right now there's no indication that you can even do it.
3. After adding an uncited item, it should be selected and get a focus ring and scroll below. (If multiple, I guess just the first one.) Right now it can silently appear below the fold.
4. I still think the gray revert button alone is just too subtle. We have the colored labels for Excluded and Cited, but editing will probably be the most common action, and it seems inconsistent not to have a similar colored label for that. I don't really see the downside, and the text description will make the state much clearer.
5. "Excluded" is a little close to the text here:
Screenshot 2025-10-10 at 12 21 05 PM

6. I think we can get rid of this prompt (which has a typo anyway) — since we don't particularly want people to edit bibliography entries to begin with, I don't think we need to scare them away from undoing it (and if we do add an Edited text label, it will also be clearer what you're undoing):

Screenshot 2025-10-10 at 12 15 37 PM

abaevbog:
> But if you click on an entry in the list below, I think it should probably show the dashed line and also select the item in the list above and scroll it into view. 

That's fair. Now, clicking into the editor will highlight that editor's bibliography row and also select that item's row in the itemTree. I agree that it looks better even though functionally it doesn't add much.

> Not sure if we discussed this above, but I think we need + buttons in the list to add uncited items, not just rely on double-click/Enter.

I may be misunderstanding something but the rows of items in the itemTree that are uncited should already have the `+` button to include that item into the bibliography. Is that not what you are seeing on your end? Or are you referring to another component?

Screenshot 2025-10-14 at 10 55 39 AM

> After adding an uncited item, it should be selected and get a focus ring and scroll below. 

Right, that makes sense. When any item is added into the citation, we'll select the itemTree row and scroll the editor into view.

> I still think the gray revert button alone is just too subtle. We have the colored labels for Excluded and Cited, but editing will probably be the most common action, and it seems inconsistent not to have a similar colored label for that.

Cool, so I added the "Edited" label that will appear along side of the revert button if the text is modified. We used to reserve the space for the revert button and the label to appear without shifting the text. I think now we should not longer do that, since we can have up to two labels (Uncited and Edited), and reserving that space for all rows would result in too much empty space. This is what it looks like now:

Screenshot 2025-10-14 at 11 19 45 AM

> "Excluded" is a little close to the text here:

Right, I added 5px extra padding after the editor.

> I think we can get rid of this prompt (which has a typo anyway) — since we don't particularly want people to edit bibliography entries to begin with, I don't think we need to scare them away from undoing it 

I agree, I removed that confirmation dialog. I assume we still want to keep one for "Revert All"? That one does not just reset the values of all bibliography entries, but it also removes all uncited items. It feels like a kind of destructive action that we should still show a warning for.

dstillman:
> I may be misunderstanding something but the rows of items in the itemTree that are uncited should already have the + button to include that item into the bibliography.

I'm not seeing that column. I haven't tried resizing the window.

Screenshot 2025-10-14 at 2 34 04 PM

abaevbog:
I see. Does the column appear in the View > Columns menu for you? It is technically just a column like any other, so it could be hidden. And if a column is hidden, that state of the itemTree is saved when the dialog closes and on the next open, it will not appear by default.

In any case, we should probably force that column to appear no matter what when the dialog reopens. Or prevent it from being hidden to begin with.

Screenshot 2025-10-14 at 11 37 55 AM

dstillman:
Yeah, it's unchecked there, but I definitely didn't uncheck it.

abaevbog:
Gotcha! I'm not sure why the `+/-` column would not be showing but I added a special handling to ensure that it always appears, no matter what.

And I removed it from the View > Columns menu, since it's never supposed to be removed.

abaevbog:
I rebased again just to make sure the fix from https://github.com/zotero/zotero/commit/5b82a38383ee2689b959c656713beb3e8accdea6 works fine with the changes in this dialog.

PR #5212, opened 2025-04-16
Move note-editor popup a11y strings to zotero.ftl

abaevbog:
Per https://github.com/zotero/zotero/pull/5205#issuecomment-2809055908

Needed for https://github.com/zotero/note-editor/pull/65
Addresses: https://github.com/zotero/zotero/issues/5199

This also adds a new string to announce when citation popup opens because it should be announced as well.

### DIFF
--- a/chrome/content/zotero/xpcom/intl.js
+++ b/chrome/content/zotero/xpcom/intl.js
@@ -85,6 +85,7 @@ Zotero.Intl = new function () {
 			'zotero.ftl',
 			'reader.ftl',
 			'integration.ftl',
+			'note-editor.ftl',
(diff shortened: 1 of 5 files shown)


dstillman:
Let's make a new note-editor.ftl file (like reader.ftl in cde21ac9f2407856d000ea2038ea49cee7e8dd7c).

abaevbog:
That makes sense. I moved the strings to `note-editor.ftl` and added it as a localization link to `editor.zotero.html`. I kept the `zotero.ftl` link in there as well, even though we don't currently use it. That's just in case we want to use any of the general strings from there in the future.

PR #5219, opened 2025-04-23
Scaffold: Integrate Firefox devtools

AbeJellinek:
I think I was iffy on this at some point in the past, but it turned out to be really straightforward to implement, and I think it's a nice workflow - basically no need to use an external browser for translator development anymore. You can pick elements directly from the Scaffold browser pane, view and edit/resend XHR requests, and navigate and debug page JS.

No ability to set breakpoints within the translator code yet, but that would be very cool, so I'll see if it's possible.

Screenshots:

image

image


dstillman:
Incredible stuff.

1. I guess we're including the devtools in all builds these days, and the `-t` pref is only determining whether we enable remote debugging?
2. Is there ever anything other than Info when not on the Browser tab? I see that we're force-switching to Console, but should we also force-enable Info and hide the other options? A little weird that you can turn off logging, which is the only reason that pane exists. We could also hide the (i) icon before each row.
3. If I go to the Browser tab, load nytimes.com, enable the inspector, click on the page, and then Cmd-Q, the app hangs:

```
JavaScript error: , line 0: NotFoundError: No such JSProcessActor 'BrowserToolboxDevToolsProcess'
JavaScript error: , line 0: NotFoundError: No such JSProcessActor 'BrowserToolboxDevToolsProcess'
JavaScript error: , line 0: TypeError: NetworkError when attempting to fetch resource.
[Parent 5618, IPC I/O Parent] WARNING: process 5655 exited with status 1: file /builds/worker/checkouts/gecko/ipc/chromium/src/base/process_util_posix.cc:341
[Parent 5618, IPC I/O Parent] WARNING: process 5669 exited with status 1: file /builds/worker/checkouts/gecko/ipc/chromium/src/base/process_util_posix.cc:341
[Parent 5618, IPC I/O Parent] WARNING: process 5668 exited with status 1: file /builds/worker/checkouts/gecko/ipc/chromium/src/base/process_util_posix.cc:341
[Parent 5618: Main Thread]: D/WorkerShutdownDump Found ActiveWorker (dedicated): resource://devtools/client/shared/source-map-loader/worker.js
[Parent 5618: Main Thread]: D/WorkerShutdownDump Principal: [System Principal]
[Parent 5618: Main Thread]: D/WorkerShutdownDump LoadingPrincipal: [System Principal]
[Parent 5618: Main Thread]: D/WorkerShutdownDump CrashInfo: Canceled


[delay]


Possibly unhandled rejection:

Database permanently closed; not re-opening

Error: Database permanently closed; not re-opening
    Zotero.DBConnection.prototype._getConnection@chrome://zotero/content/xpcom/db.js:1211:11
    Zotero.DBConnection.prototype.queryAsync@chrome://zotero/content/xpcom/db.js:600:21
    Zotero.Feeds</this.updateFeeds<@chrome://zotero/content/xpcom/data/feeds.js:319:41
From previous event:
    this.init@chrome://zotero/content/xpcom/zotero.js:410:25
    getContentsFromURL@chrome://zotero/content/xpcom/file.js:339:13
    Zotero.Intl</<@chrome://zotero/content/xpcom/intl.js:102:35
    Zotero.Intl</this.getString@chrome://zotero/content/xpcom/intl.js:146:18
    this.getString@chrome://zotero/content/xpcom/zotero.js:1540:24
    ZoteroPane<@chrome://zotero/content/zoteroPane.js:3178:19
    @chrome://zotero/content/zoteroPane.js:31:18
    @chrome://zotero/content/zoteroPane.xhtml:81:25
RunWatchdog: Mainthread nested event loops during hang:
 --- (no nested event loop active)
[…]/app/scripts/build_and_run: line 86:  5618 Segmentation fault: 11  "$APP_ROOT_DIR/staging/$command" "${profile_args[@]}" -ZoteroDebugText -purgecaches $PARAMS "$@"
```

dstillman:
I get the hang without -jsdebugger, FWIW.

dstillman:
FWIW, I was only talking about the console output when not in the Browser tab — I wasn't suggesting we hide site output in the Browser tab. Though it certainly may not be useful for translator development in any way. But then that seems like all the more reason to hide the Console subtabs that would never be used.

dstillman:
Hmm, it still seems a little convoluted.

if you're in the Tests tab and run a test, it force-switches you to the Console tab, but the XHR and Requests toggles aren't actually used at all. And they're not even used in the Browser tab if you run translation — only if you click in the URL field and press Enter.

Filtering the Console output also doesn't work for translation output.

Meanwhile, you can't actually use the Network tab to see the requests being made when running translation. Maybe that's fine, because we mostly just want to emulate the real requests made during normal browser interactions, but 1) maybe it would be useful (?) and 2) if not, should we even be showing those other tabs?

So one option would be to hide everything except for the Console field when running translation — no tabs, no filter bar, no subtabs, no input field — and then show the rest of the devtools UI when doing a webpage load in the Browser tab. Maybe that would be distracting, but it might be more clear. We're also already showing the fake item pane for translation, so the UI changes anyway.

dstillman:
Yes, if it actually logged requests/XHR (and, even better, if the Network tab still worked) during translation, that would probably be fine.

PR #5245, opened 2025-05-01
Fix unintentional title transfer between attachments on item switch

### LINKED ISSUE #5244: Title of attachment A1 saved to attachment A2 when switching from A1 to A2
https://forums.zotero.org/discussion/123955/title-of-attachment-a1-saved-to-attachment-a2-when-switching-from-a1-to-a2

abaevbog:
To avoid re-focusing that fields for new attachment after refresh, which leads to the value of old attachment's field being transferred to the new attachment.

Fixes #5244

https://github.com/user-attachments/assets/d718043b-8ac6-4320-b73d-330e8af24069

### DIFF
--- a/chrome/content/zotero/elements/attachmentBox.js
+++ b/chrome/content/zotero/elements/attachmentBox.js
@@ -221,9 +221,9 @@
 			if (!(val instanceof Zotero.Item)) {
 				throw new Error("'item' must be a Zotero.Item");
 			}
-			// Blur editable-text of old attachment. Otherwise, _handleTitleBlur would fire after
-			// the new item is set but before updateInfo sets the title field to have the new item's value.
-			// Such race condition would cause the old item's title to be set on the new item.
+			// Blur editable-text of old attachment. Otherwise, _handleTitleBlur() would fire after
+			// the new item is set but before updateInfo() sets the title field to have the new
+			// item's value, causing the old item's title to be set on the new item.
 			if (this._item && val.id != this._item.id && document.activeElement.closest("editable-text")) {
 				document.activeElement.blur();
 			}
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/elements/attachmentBox.js
```diff
@@ -221,6 +221,10 @@
 			if (!(val instanceof Zotero.Item)) {
 				throw new Error("'item' must be a Zotero.Item");
 			}
+			// Blur editable-text of old attachment so it is not refocused for new attachment after refresh
```
dstillman, line 224: Why did this cause the old value to be transferred to this one, though? Should clarify that it's avoiding that. (A test for this — and the same thing in item details, if we don't have that — would also be good here.)

abaevbog: > Why did this cause the old value to be transferred to this one, though?

Apologies I wasn't clear, I updated the comment so it should hopefully be more detailed. Essentially, it's a race condition. Select attachment A -> focus title -> select item B -> title blurs -> current value of `#title` field (still from item A) is saved into current item B -> `updateInfo` places title from item B (already same as title A) into the `#title` field. As an experiment, when I added a delay into the blur handler, the `updateInfo` placed item B's title into the `#title` field before item saving runs, which would not cause the title to transfer. I added a test case for that too.

> and the same thing in item details, if we don't have that 

Sorry, do you mean to add the same handling of blurring focused `editable-text` into `itemDetails` when item changes? It might be helpful to generalize this, so we don't need to do that in each component separately. I was cautious to do that because it would affect all the components in the `itemPane`. Or to just add a few additional test cases for components that do the same thing? There is a test for `tagsBox` already but it doesn't look like there is one for `itemBox`, `abstractBox` or `itemPaneHeader`

dstillman: Yes, I just mean a test for this item details.

abaevbog: Gotcha! I added a similar test for the item details table.

Issue #5246, opened 2025-05-01
Ghost annotation rows after changing attachment's parent item


abaevbog:
https://forums.zotero.org/discussion/123994/ghost-annotations-in-the-items-list-after-changing-parent-item#latest

This occurs specifically if the original parent of the attachment has no other children but the attachment being moved.

https://github.com/user-attachments/assets/247e6aa6-a72b-4da8-83f1-33310f8ab00c

abaevbog:
This is a regression after a7964f7116fc0a386710f51ad55805d4a5c54246, which itself fixed another regression from 48fd23ccec73156855f4c59170b332af2e4357dd

abaevbog:
Also https://forums.zotero.org/discussion/comment/491496/#Comment_491496

PR #5248, opened 2025-05-02
Fix leftover annotation rows and disappearing attachment rows after drag-drop

abaevbog:
- collapse all rows before removing them. It makes sure that annotation rows will be cleaned up properly when an annotation row is moved into another parent.
- move handling of inserting a new row when attachment is moved to be a top-level row from under a parent into the `item.isTopLevelItem()` condition. Otherwise, it never gets reached and attachment row just disappears when dragged to the top level.

Fixes: #5246

https://github.com/user-attachments/assets/cb277e7f-1330-497c-89f8-0dc57c218081

### DIFF
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -686,9 +686,14 @@ var ItemTree = class ItemTree extends LibraryTree {
 					if (row !== undefined) {
 						let parentItemID = this.getRow(row).ref.parentItemID;
 						let parentIndex = this.getParentIndex(row);
-
+						
+						// If item moved from top level to under another item, remove the old row
+						if (parentIndex == -1 && parentItemID) {
+							this._closeContainer(row);
+							this._removeRow(row);
+						}
 						// If moved from under another item to top level, remove old row and add new one
-						if (parentIndex != -1 && !parentItemID) {
+						else if (parentIndex != -1 && !parentItemID) {
 							this._closeContainer(row);
 							this._removeRow(row);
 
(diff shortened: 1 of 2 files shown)


dstillman:
Can we add tests?

chrome/content/zotero/itemTree.jsx
```diff
@@ -693,26 +693,29 @@ var ItemTree = class ItemTree extends LibraryTree {
...
+							// If moved from under another item to top level, remove old row and add new one
+							else if (parentIndex != -1 && !parentItemID) {
+								this._closeContainer(row);
+								this._removeRow(row);
+	
+								let beforeRow = this.rowCount;
+								this._addRow(new ItemTreeRow(item, 0, false), beforeRow);
+	
+								sort = true;
+							}
```
dstillman, line 705: Oh, but the problem here is that I changed the meaning of the "Top-level item" block in a7964f7116fc0a386710f51ad55805d4a5c54246. It's supposed to be for top-level items being changed, and it was that way because the `row` in `this.isContainer(row)` was coming from the row map that hadn't been updated yet.

So the proper fix here is a reverting to `this.isContainer(row)` and adding an additional condition to prevent that from applying to attachments, which have become containers.

dstillman: Haven't tested, but just `this.getLevel(row) === 0`, maybe?

dstillman: `parentIndex == -1` would also work, because that's just testing for level being 0.

dstillman: Oh, and we're using `parentIndex == -1` below. So we'd need to combine with the following block and add a sub-block for the `parentItemID` test.

dstillman: (And then we can remove the `parentIndex != -1` from the following conditions, since it will be redundant.)

abaevbog: I reverted to `this.isContainer(row)` but then I realized that now handling of moving a row between/to/from top level items should apply equally to containers (attachments) and to non-containers (notes). So I refactored this to handle change in parent regardless of if it is an container or not, with the Unfiled items being a special case at the bottom. Does this make sense?

I added test cases for moving between/to/from top-level items of attachments and notes. I checked and with earlier iterations (before a7964f7116fc0a386710f51ad55805d4a5c54246, after, and with the initial commit from this PR), at least one of them would be failing.

PR #5254, opened 2025-05-04
Fix table columns misaligned after resizing

### LINKED ISSUE #5198: Table columns misaligned after resizing
It's possible to get misaligned columns by resizing the window, pane, or even a single column:

https://github.com/user-attachments/assets/2d1cb6bb-b511-4c0c-b2f8-ce6d975ba0d0

For the scenario in the video above, this is most likely due to the minimal width of the first column not being adjusted for extra content (item type icon and chevron). For other scenarios (window/pane resize), a fix similar to #5162 might be needed.

tnajdek:
This PR is a follow-up to #5162 and fixes misaligned columns that sometimes appear after:

- [x] Resizing the first column to the minimal width  
- [x] Shrinking the items tree table pane  
- [x] Shrinking the entire window

Resolves #5198

### DIFF
--- a/chrome/content/zotero/components/virtualized-table.jsx
+++ b/chrome/content/zotero/components/virtualized-table.jsx
@@ -842,7 +842,10 @@ class VirtualizedTable extends React.Component {
 		event.stopPropagation();
 		const result = this._getResizeColumns();
 		if (!result) return;
+		const columns = this._getVisibleColumns();
 		const [aColumn, bColumn, resizingColumn] = result;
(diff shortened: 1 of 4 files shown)


tnajdek:
I've reverted my previous fix, which was working around the issue rather than addressing the core problem. What I hadn't noticed before is that we had introduced additional padding rules that applied only to the header or the body of the table. These differences caused the flex model to assign different actual `px` widths in some scenarios.

https://github.com/user-attachments/assets/175198fd-03e4-4baf-8595-47467b0b48cf

Issue #5257, opened 2025-05-05
Drag-drop of a child item between annotation rows makes it top-level


abaevbog:
https://forums.zotero.org/discussion/124017/drag-and-drop-a-child-attachment-to-the-annotations-interline-makes-it-a-standalone-attachment

Though it's not specific to annotation rows. One can drop a child attachment or note between any child rows of another item, which would also make them standalone. The intended behavior is likely the same as with standalone notes or attachments, which cannot be dropped between child rows at all.

https://github.com/user-attachments/assets/9924fae3-fe1b-4075-a4c2-901d03161f9c

PR #5258, opened 2025-05-05
Disable itemTree row drop between child rows

abaevbog:
Fixes: #5257

https://github.com/user-attachments/assets/fc9b1d15-304b-4145-8eef-ed00cc8805c3

### DIFF
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -2345,7 +2345,7 @@ var ItemTree = class ItemTree extends LibraryTree {
 
 			// In library, allow children to be dragged out of parent
 			else if (collectionTreeRow.isLibrary(true) || collectionTreeRow.isCollection()) {
-				let targetRow = row !== -1 ? this.getRow(row) : null;
+				let targetRow = row != -1 ? this.getRow(row) : null;
 				for (let item of items) {
 					// Don't allow drag if any top-level items
 					if (item.isTopLevelItem()) {
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -2374,9 +2374,9 @@ var ItemTree = class ItemTree extends LibraryTree {
 					if (orient == 1 && targetRow.isOpen) return false;
 					// Cannot drop after the last child of a parent container
 					if (orient == -1) {
-						var parentIndex = this._rowMap[item.parentItemID];
-						var nextParentIndex = null;
-						for (var i = parentIndex + 1; i < this.rowCount; i++) {
+						let parentIndex = this._rowMap[item.parentItemID];
+						let nextParentIndex = null;
+						for (let i = parentIndex + 1; i < this.rowCount; i++) {
 							if (this.getLevel(i) == 0) {
 								nextParentIndex = i;
 								break;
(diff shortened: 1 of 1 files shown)


chrome/content/zotero/itemTree.jsx
```diff
@@ -2365,21 +2365,18 @@ var ItemTree = class ItemTree extends LibraryTree {
...
-					// Don't allow children to be dragged within their own parents
 					var parentItemID = item.parentItemID;
 					var parentIndex = this._rowMap[parentItemID];
-					if (row != -1 && this.getLevel(row) > 0) {
-						if (this.getRow(this.getParentIndex(row)).ref.id == parentItemID) {
-							return false;
-						}
-					}
-					// Including immediately after the parent
+					let targetItem = this.getRow(row)?.ref;
```
dstillman, line 2370: We already have a `rowItem` from above.

abaevbog: We do. However in this conditional branch `rowItem` is `undefined`. `rowItem` is only set if we drag directly over the row:

https://github.com/zotero/zotero/blob/8b006fd6b705831b543449e8fc71fa725f40d12e/chrome/content/zotero/itemTree.jsx#L2306-L2312

And if it is set, it is handled in an earlier conditional:
https://github.com/zotero/zotero/blob/8b006fd6b705831b543449e8fc71fa725f40d12e/chrome/content/zotero/itemTree.jsx#L2317-L2326

It is a bit convoluted but I thought it's best to not refactor how the `canDropCheck` works for a smaller fix like this. Would it be worth to tweak how `rowItem` is set to be more cohesive here? Alternatively, if we wanted to streamline this a bit more, #4785 moves `canDropCheck` for collectionTree to `Zotero.Item.prototype.canAddToTarget`. We could also expand its `canAddToTarget` to handle objects from itemTree as well.

dstillman: Oh, sorry, I misread that.

Why do we have the `?` here, though? Is that just for `row` being `-1`? Seems like we should check for that explicitly and otherwise expect `getRow(row)` to work.

(We should probably also move the `getRow(row)` or `getLevel(row)` calls out of `items` loop for good measure, though I doubt there's a meaningful effect on performance.)

abaevbog: > Why do we have the ? here, though? Is that just for row being -1?

Yes, exactly, that's in case you drag over the part of itemTree that does not have any more rows. 

>  Seems like we should check for that explicitly and otherwise expect getRow(row) to work.

That makes sense! I moved setting of `targetRow` to the very top of this conditional, and if it doesn't exist, we'll just `continue` that loop because dropping over empty space should always make child items top level.

> (We should probably also move the getRow(row) or getLevel(row) calls out of items loop for good measure, though I doubt there's a meaningful effect on performance.)

Right, the `targetRow` will be fetched before the loop. `this.getLevel(i)` is called to check if we are dragging right after the last child of a parent of any of the items being dragged, so it has to be in this way or another a part of `for (let item of items)` loop. We could pull it out of the big loop for better readability? But I believe we would still end up having to iterate over all the items and, for each item, going through multiple rows to find the last child of that item's parent.

dstillman:
This is great.

PR #5263, opened 2025-05-06
Always show item/context pane toggle

### LINKED ISSUE #5260: Update/add button to toggle item pane
Though there is already such a button in reader, many people want it to stay in the same place when it's toggled (https://forums.zotero.org/discussion/comment/491769/#Comment_491769). This requires putting this button at the top of the sidenav when it's visible, and at the right of the toolbar (4px right margin to achieve the exact same position) when the sidenav is hidden.

For consistency's sake, we need to also add this button to the top of the sidenav in library view, so that buttons in the sidenav don't visibly shift up/down when people switch tabs. This also makes it easier to close the item pane in library view. The desired effect looks like this:

https://github.com/user-attachments/assets/54eec07e-271a-485c-bafd-5cf6be8e3a06

In stacked view, this button will always stay at the right of the toolbar, in the exact same position. Its behavior is the same as currently in reader.

Image

AbeJellinek:
And:

- Make everything in `ZoteroContextPane` an instance method to reduce confusion about what `this` means
- Remove redundant `if (Zotero_Tabs.selectedIndex > 0)` check and use tab type instead of index

Needs zotero/reader#164, closes #5260

### DIFF
--- a/chrome/content/zotero/contextPane.js
+++ b/chrome/content/zotero/contextPane.js
@@ -145,6 +145,9 @@ var ZoteroContextPane = new function () {
 			// needed for standard layout
 			_contextPane.style.width = 'auto';
 			_contextPaneInner.style.removeProperty("min-height");
+
+			// Propagate state to standard splitter
+			_contextPaneSplitter.setAttribute('state', this.collapsed ? 'collapsed' : 'open');
 		}
 		else {
 			_contextPaneSplitter.setAttribute('hidden', false);
@@ -160,6 +163,9 @@ var ZoteroContextPane = new function () {
 			// force it to occupy all height available
 			_contextPaneInner.style.minHeight = `100%`;
 			_contextPane.style.width = `${_contextPane.getAttribute("width")}px`;
+
+			// Propagate state to stacked splitter
+			_contextPaneSplitterStacked.setAttribute('state', this.collapsed ? 'collapsed' : 'open');
 		}
 		
 		Zotero.Reader.setContextPaneOpen(!this.collapsed);
--- a/chrome/content/zotero/contextPane.js
+++ b/chrome/content/zotero/contextPane.js
@@ -57,11 +57,9 @@ var ZoteroContextPane = new function () {
 			_contextPaneInner.setAttribute('collapsed', !!collapsed);
 			_contextPaneSplitter.setAttribute('state', collapsed ? 'collapsed' : 'open');
 			_contextPaneSplitterStacked.setAttribute('state', collapsed ? 'collapsed' : 'open');
-			_update();
+			this.update();
 		}
 	});
-
-	this.update = _update;
 
 	this.focus = () => {
 		return _contextPaneInner.handleFocus();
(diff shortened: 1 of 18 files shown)


yexingsha:
1. In library view, the top border of the itemPane/sidenav in Stacked view isn't rendered consistently. Steps to reproduce:

	- When item pane is open in Standard view, switch to Stacked view. Close the item pane. Now the top border of the sidenav is darker than it should be. If I switch back to Standard view in this state, an extra border will show up:

	Screenshot 2025-05-06 at 17 31 47

	Screenshot 2025-05-06 at 17 36 09

	- When item pane is closed in Standard View, switch to Stacked view. Open the item pane. Now the top border of the item pane isn't rendered. If I switch back to Standard view in this state, the corresponding border isn't rendered either:

	Screenshot 2025-05-06 at 17 37 35

	Screenshot 2025-05-06 at 17 46 16

2. In reader, in Standard view, close the item pane. Switch to Stacked view. The first click on the toggle button doesn't open the item pane.

3. In reader toolbar, let's get rid of the divider between the search button and the toggle button, and set the gap between these buttons to 9px, so that the search button is aligned with the tab overview button up top.

yexingsha:
I see this in my `prefs.js`:

```user_pref("extensions.zotero.pane.persist", "{\"zotero-reader-sidebar-pane\":{\"collapsed\":\"false\",\"width\":\"240\"},\"zotero-layout-switcher\":{\"orient\":\"horizontal\"},\"zotero-items-splitter\":{\"state\":\"\",\"orient\":\"horizontal\"},\"zotero-context-splitter\":{\"state\":\"collapsed\"},\"zotero-context-splitter-stacked\":{\"state\":\"open\"},\"zotero-collections-pane\":{\"width\":\"246\"},\"zotero-tags-splitter\":{\"state\":\"\"},\"zotero-tag-selector-container\":{\"height\":\"290\"}}");```

After I quit Zotero, remove this line, and open Zotero again, I still see the same issues.

yexingsha:
1) and 3) looks fine. 2) still persists, and now when I close the context pane in Standard view, the toggle button disappears.

windingwind:
Nice work!

Since we're reconsidering this part, maybe we also want to show the toggle button when the note is selected. Could either 1) show the sidenav for note as well; or 2) show it in the note editor's top-right corner

windingwind:
For example:

When no item is selected, I can see the toggle button, and I, as a user, feel safe even if the item pane is hidden. I know I can open it any time:

image

But after selecting the note,  nothing happens, and the button is gone! I will definitely feel panic for losing part of the UI:

image

windingwind:
We are currently hiding the sidenav intentionally for notes: https://github.com/zotero/zotero/blob/143829065f79553f4051cc81ba640d2fce37ec02/chrome/content/zotero/elements/itemPane.js#L126-L127

Simply removing that line would make the sidenav available.

yexingsha:
Looks good to me now!

And yes, we shouldn't hide the sidenav for notes in the current setup. I was trying to find the discussion we had about what do with the sidenav for notes but couldn't. I think maintaining the sidenav state when no item is selected should be fine.

windingwind:
Looks good!

dstillman:
In the reader, the button is shifting for me:


https://github.com/user-attachments/assets/12bc135b-39fe-4068-b657-936d96bd65f2

yexingsha:
@dstillman: pull https://github.com/zotero/reader/pull/164

dstillman:
Better, sorry!

PR #5270, opened 2025-05-07
Make XULElementBase into a mixin to reduce code duplication

AbeJellinek:
As Mozilla does with theirs. This way, we can augment existing CE superclasses with our `XULElementBase` methods without duplicating code.

And check whether `super` has `connectedCallback`/`disconnectedCallback`. Fixes `TypeError: super.disconnectedCallback is not a function` on quit.


windingwind:
Looks good to me!

PR #5271, opened 2025-05-07
Fix annotation item pane title word cutoff

windingwind:


windingwind:
Before:

image


After:

image

abaevbog:
Oh, yeah, that looks much nicer now!

PR #5274, opened 2025-05-08
Fix dropping or copy-pasting of annotation rows into note editor

### LINKED ISSUE #5235: Drag-drop of annotation rows into note editor
https://forums.zotero.org/discussion/123899/drag-and-drop-annotations-from-the-items-list

Currently, dropping a row into a separate `note-editor` window does nothing, even though dropping the annotation into text editor does paste the annotation as expected.

Also, copying annotation via Edit > Copy Annotation and pasting it into note will add an empty `{citation}` placeholder after annotation text.

abaevbog:
- set `zotero/annotation` data when annotation rows are dragged into the note editor, so that it can add a citation, the same way as when dragging from the reader
- when copying annotations to the clipboard as a note, set `zotero/annotation` data so that note-editor can handle it. `zotero/annotation` does not appear by itself in `event.clipboardData` on paste, so it will be passed into note-editor iframe in `window.clipboardData` object, similar to how drop events are handled
- tweak `Zotero.Annotations.toJSONSync` to optionally add fields used to serialize annotations in `Zotero.EditorInstanceUtilities.serializeAnnotations`. Set annotation.id to be item.key, which will add
the &annotation= param into the link to the zotero://open-pdf/ link after the citation.

Fixes: #5235
Needs: https://github.com/zotero/note-editor/pull/66 to handle pasting of annotation rows into note-editor

https://github.com/user-attachments/assets/c5487602-b423-4a42-8cdd-09eb2ebd119a

### DIFF
--- a/chrome/content/zotero/elements/noteEditor.js
+++ b/chrome/content/zotero/elements/noteEditor.js
@@ -83,6 +83,18 @@
 						'zotero/item': event.dataTransfer.getData('zotero/item')
 					}, this._iframe.contentWindow);
 				}, true);
+
+				// If annotations are copies from itemTree, zotero/annotation will not
+				// be included in event.clipboardData, so we fetch it manually
+				// from the clipboard and provide it to the note-editor
+				this._iframe.contentWindow.addEventListener('paste', (event) => {
+					let stringifiedAnnotations = Zotero_File_Interface.getDataFromClipboard("zotero/annotation");
+					console.log('paste', stringifiedAnnotations);
+					if (!stringifiedAnnotations) return;
+					this._iframe.contentWindow.wrappedJSObject.clipboardData = Components.utils.cloneInto({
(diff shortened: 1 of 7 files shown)


abaevbog:
This should address all points from https://forums.zotero.org/discussion/123899/drag-and-drop-annotations-from-the-items-list except for the inconsistencies in keyboard shortcuts between annotations in the reader and itemTree. The reader simply has more options (with copying of the bibliography and the citation). But most importantly, reader's CMD-C is equivalent to Shift-Cmd-A on itemTree rows. There seems to be no instances of Cmd-C used elsewhere in itemTree, so using that (instead of Shift-Cmd-A) may be a bit inconsistent. But on the other hand, having different shortcuts that do the same thing is not consistent either. So maybe we should copy annotation rows on Cmd-C instead of Shift-Cmd-A?

PR #5275, opened 2025-05-08
Improve attachment row annotation button click behavior

### LINKED ISSUE #5249: Expand and scroll to attachment annotations section when clicking annotation count in item pane
https://forums.zotero.org/discussion/123992/expand-annotations-when-accessing-from-the-item-pane

windingwind:
fix: #5249

For now, I'm doing the selecting annotation items in the library, since I believe that's the interface where interactions between user and annotations would happen, for example, now we have the add note from annotations button there. If we select the attachment item with the annotations section in the item pane, there are no operations actually available.


abaevbog:
Looks good to me! Having thought about it, selecting all rows is probably indeed the most expected behavior.

Issue #5277, opened 2025-05-09
Citation dialog: Enter/Return should always trigger Done in citation popup


dstillman:
https://forums.zotero.org/discussion/124129/comment-suggestion-about-the-ignore-author-field-in-the-citation-dialog-options-and-the-keyboard

Even if Omit Author has focus, instead of toggling the checkbox

(The report is about Z7, but it's still the case in 7.1.)

PR #5281, opened 2025-05-09
Citation dialog: updated Enter handling in panels

abaevbog:
Unhandled Enter keypress from inside of a panel will now close it. Enter on "Omit Author" checkbox or on the inactive locator label dropdown will close the item details popup.
Enter on "Keep sources sorted" checkbox of settings popup will close it for consistency.

Also, allow Cmd/Ctrl+Enter to accept the dialog even from inside of a panel.

Fixes: #5277

Enter on item details buttons at the bottom will still just perform the click, as it would be strange if Enter on a focused "Remove" button would not actually remove the item.

### DIFF
--- a/chrome/content/zotero/integration/citationDialog/keyboardHandler.mjs
+++ b/chrome/content/zotero/integration/citationDialog/keyboardHandler.mjs
@@ -86,11 +86,11 @@ export class CitationDialogKeyboardHandler {
 	_handleTopLevelKeydown(event) {
 		let handled = false;
 		let tgt = event.target;
+		let isKeyboardClickable = tgt.classList.contains("keyboard-clickable") || (Zotero.isWin && tgt.tagName.localName == "button");
 		// Space/Enter will click on keyboard-clickable components.
 		// On macOS, focused buttons are only clickable with Space (not Enter),
 		// and on Windows they are clickable with both.
 		// This distinction determines if Enter on a button in a panel will perform a click or close it.
-		let isKeyboardClickable = tgt.classList.contains("keyboard-clickable") || (Zotero.isWin && tgt.tagName.localName == "button");
 		if (["Enter", " "].includes(event.key) && isKeyboardClickable) {
 			tgt.click();
 			handled = true;
--- a/chrome/content/zotero/integration/citationDialog/keyboardHandler.mjs
+++ b/chrome/content/zotero/integration/citationDialog/keyboardHandler.mjs
@@ -86,8 +86,11 @@ export class CitationDialogKeyboardHandler {
 	_handleTopLevelKeydown(event) {
 		let handled = false;
 		let tgt = event.target;
-		let isKeyboardClickable = tgt.classList.contains("keyboard-clickable") || (Zotero.isWin && tgt.tagName.includes("button"));
-		// Space/Enter will click on keyboard-clickable components
+		// Space/Enter will click on keyboard-clickable components.
+		// On macOS, focused buttons are only clickable with Space (not Enter),
+		// and on Windows they are clickable with both.
+		// This distinction determines if Enter on a button in a panel will perform a click or close it.
+		let isKeyboardClickable = tgt.classList.contains("keyboard-clickable") || (Zotero.isWin && tgt.tagName.localName == "button");
 		if (["Enter", " "].includes(event.key) && isKeyboardClickable) {
 			tgt.click();
 			handled = true;
--- a/chrome/content/zotero/integration/citationDialog/keyboardHandler.mjs
+++ b/chrome/content/zotero/integration/citationDialog/keyboardHandler.mjs
@@ -86,8 +86,8 @@ export class CitationDialogKeyboardHandler {
 	_handleTopLevelKeydown(event) {
 		let handled = false;
 		let tgt = event.target;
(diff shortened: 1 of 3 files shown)


dstillman:
> as it would be strange if Enter on a focused "Remove" button would not actually remove the item

No, closing the panel is the correct behavior on a Mac. You don't trigger a focused element with Return on macOS — you trigger it with space. E.g., go to File → "Save Page As…" right on this page, tab to New Folder, and press Return. It saves the page — it doesn't prompt you to create a new folder.

On Windows, Enter does trigger the active button. Not sure about Linux.

I also don't think we need the Cmd/Ctrl-Enter thing. You can just press Enter twice.

abaevbog:
Gotcha! I remember we did override some default behaviors for, what we thought was, convenience.

How about this? On Mac, Enter will no longer press the focused button, so it will just close the panel. On windows, Enter will still trigger button press, since this is the convention.

I also reverted the Cmd/Ctrl-Enter shortcut to accept the entire dialog from the panel.

chrome/content/zotero/integration/citationDialog/keyboardHandler.mjs
```diff
@@ -86,12 +86,17 @@ export class CitationDialogKeyboardHandler {
 	_handleTopLevelKeydown(event) {
 		let handled = false;
 		let tgt = event.target;
-		let isKeyboardClickable = tgt.classList.contains("keyboard-clickable") || tgt.tagName.includes("button");
-		// Space/Enter will click on a button or keyboard-clickable components
+		let isKeyboardClickable = tgt.classList.contains("keyboard-clickable") || (Zotero.isWin && tgt.tagName.includes("button"));
```
dstillman, line 89: `includes()` because there's a namespace prefix? `.localName == 'button'`?

chrome/content/zotero/integration/citationDialog/keyboardHandler.mjs
```diff
@@ -49,7 +49,7 @@ export class CitationDialogKeyboardHandler {
 	// lower-level components
 	captureKeydown(event) {
 		let cmdOrCtrl = Zotero.isMac ? event.metaKey : event.ctrlKey;
-		// Cmd/Ctrl-Enter will always accept the dialog regardless of the target (unless within a panel)
+		// Cmd/Ctrl-Enter will always accept the dialog regardless of the target (unless within a panel
```
dstillman, line 52: Lost the trailing `)`

chrome/content/zotero/integration/citationDialog/keyboardHandler.mjs
```diff
@@ -86,12 +86,17 @@ export class CitationDialogKeyboardHandler {
 	_handleTopLevelKeydown(event) {
 		let handled = false;
 		let tgt = event.target;
-		let isKeyboardClickable = tgt.classList.contains("keyboard-clickable") || tgt.tagName.includes("button");
-		// Space/Enter will click on a button or keyboard-clickable components
+		let isKeyboardClickable = tgt.classList.contains("keyboard-clickable") || (Zotero.isWin && tgt.tagName.includes("button"));
+		// Space/Enter will click on keyboard-clickable components
```
dstillman, line 90: Move back above (and probably explain the Windows vs. Mac thing)

chrome/content/zotero/integration/citationDialog/keyboardHandler.mjs
abaevbog: Yes, exactly, there is one `xul:button`. And yes, `.localName` is way better!

chrome/content/zotero/integration/citationDialog/keyboardHandler.mjs
```diff
@@ -86,12 +86,20 @@ export class CitationDialogKeyboardHandler {
 	_handleTopLevelKeydown(event) {
 		let handled = false;
 		let tgt = event.target;
-		let isKeyboardClickable = tgt.classList.contains("keyboard-clickable") || tgt.tagName.includes("button");
-		// Space/Enter will click on a button or keyboard-clickable components
+		let isKeyboardClickable = tgt.classList.contains("keyboard-clickable") || (Zotero.isWin && tgt.tagName.localName == "button");
+		// Space/Enter will click on keyboard-clickable components.
+		// On macOS, focused buttons are only clickable with Space (not Enter),
+		// and on Windows they are clickable with both.
+		// This distinction determines if Enter on a button in a panel will perform a click or close it.
```
dstillman, line 93: Oh, sorry, maybe I misread the diff. `isKeyboardClickable` should be below the comment.

chrome/content/zotero/integration/citationDialog/keyboardHandler.mjs
abaevbog: It took me a second to phrase it - how about this now:
```
  // Space/Enter will click on keyboard-clickable components.
  // On macOS, focused buttons are only clickable with Space (not Enter),
  // and on Windows they are clickable with both.
  // This distinction determines if Enter on a button in a panel will perform a click or close it.
```

chrome/content/zotero/integration/citationDialog/keyboardHandler.mjs
abaevbog: Moved it over!

Issue #5282, opened 2025-05-09
Citation dialog: prefix and suffix values persist on item in the citation


abaevbog:
Prefix and suffix on a citation item will stick to it. So if the item is deleted but then re-added into the citation, suffix and prefix remain.

https://github.com/user-attachments/assets/3a4f6b1c-3425-4bdb-bd3b-0feed8a9e829

More curious, this behavior will persist if I close and reopen the dialog.

PR #5289, opened 2025-05-12
Citation Dialog: fix added suffix/prefix properties persisting on items

abaevbog:
`CitationDataManager.items` is an array with `zoteroItem` and `citationItem` objects used to represent cited items. `zoteroItem` is an instance of `Zotero.Item`, and `citationItem` is an object with citation data (locator/suffix/prefix, cslItemData, etc.).

If an ordinary (non-cited) item from the library is added to the citation, `citationItem` [would end up](https://github.com/zotero/zotero/blob/ed24cce4a1aaa0d78a0e2ec0be6ab003d6ccb876/chrome/content/zotero/integration/citationDialog.js#L1470) having an instance of `Zotero.Item`. In that case, if a suffix/prefix/etc. is added to the `citationItem`, it would land on the instance of `Zotero.Item` and persist even if the bubble is removed and then re-added. If two bubbles for the same item are added, both of them would share those properties too.

This PR ensures that instances of `Zotero.Item` are only placed into `zoteroItem`, and `citationItem` is maintained in the format that is equivalent to `io.citation.citationItems`. This better separation allows for some streamlining of how items are being processed. `CitationDataManager.addItems` now expect an array with both `zoteroItem` and `citationItem`, instead of checking if a provided citation item (which is often already an instance of `Zotero.Item`) [can be converted to zoteroItem](https://github.com/zotero/zotero/blob/ed24cce4a1aaa0d78a0e2ec0be6ab003d6ccb876/chrome/content/zotero/integration/citationDialog.js#L1466). That allows handlers that call `CitationDataManager.addItems` to only perform necessary transformations using `_zoteroToCitationItem` (when adding a library item) and `_citationToZoteroItem` (when building a citation with cited items on initial load). As an added benefit, when `io` citation object is [being updated](https://github.com/zotero/zotero/blob/ed24cce4a1aaa0d78a0e2ec0be6ab003d6ccb876/chrome/content/zotero/integration/citationDialog.js#L1506-L1518), no transformations are now necessary - we can essentially just use the array of `citationItems`. I think this makes the flow of how items are handled significantly more straightforward.

Fixes: #5282

### DIFF
--- a/chrome/content/zotero/elements/bubbleInput.js
+++ b/chrome/content/zotero/elements/bubbleInput.js
@@ -50,21 +50,20 @@
 		 * are not present, remove bubbles whose citation items were removed, rearrange bubbles
 		 * if the items were moved, update bubble text if locator/prefix/suffix was changed.
 		 * Make sure that there is an input for user to type in before and after every bubble.
-		 * @param {Object[]} combinedItems - array of objects { zoteroItem, citationItem, dialogReferenceID, selected }.
-		 * zoteroItem - Zotero.Item, citationItem - object from io.citation.citationItems
-		 * dialogReferenceID - String ID of this citation entry, selected - Boolean indicator if bubble should be highlighted
+		 * @param {Object[]} bubblesConfig - array of objects { dialogReferenceID, bubbleString, selected}
+		 * representing each bubble.
 		 */
-		refresh(combinedItems) {
+		refresh(bubblesConfig) {
 			// Remove bubbles of items that are no longer in the citations
 			for (let bubble of this.getAllBubbles()) {
 				let bubbleDialogReferenceID = bubble.getAttribute("dialogReferenceID");
-				let itemExistsForBubble = combinedItems.find(({ dialogReferenceID }) => dialogReferenceID == bubbleDialogReferenceID);
+				let itemExistsForBubble = bubblesConfig.find(({ dialogReferenceID }) => dialogReferenceID == bubbleDialogReferenceID);
 				if (!itemExistsForBubble) {
 					bubble.remove();
 				}
 			}
 			// Ensure each item in the citation has a bubble in the right position
-			for (let [index, { dialogReferenceID, bubbleString }] of Object.entries(combinedItems)) {
+			for (let [index, { dialogReferenceID, bubbleString }] of Object.entries(bubblesConfig)) {
 				let allBubbles = this.getAllBubbles();
 				let bubbleNode = allBubbles.find(candidate => candidate.getAttribute("dialogReferenceID") == dialogReferenceID);
 				// Create bubble if it does not exist and append to the input
@@ -95,7 +94,7 @@
 			// Highlight bubbles selected in the library view
 			for (let bubble of this.getAllBubbles()) {
 				let bubbleDialogReferenceID = bubble.getAttribute("dialogReferenceID");
-				let itemObj = combinedItems.find(({ dialogReferenceID }) => dialogReferenceID == bubbleDialogReferenceID);
+				let itemObj = bubblesConfig.find(({ dialogReferenceID }) => dialogReferenceID == bubbleDialogReferenceID);
 				if (itemObj) {
 					bubble.classList.toggle("has-item-selected", !!itemObj.selected);
 				}
--- a/chrome/content/zotero/integration/citationDialog.js
+++ b/chrome/content/zotero/integration/citationDialog.js
@@ -181,8 +181,8 @@ class Layout {
 		let sections = [];
 
 		// Tell SearchHandler which currently cited items are so they are not included in results
-		let citedItems = CitationDataManager.getCitationItems();
-		let searchResultGroups = SearchHandler.getOrderedSearchResultGroups(citedItems);
+		let citedIDs = CitationDataManager.getCitedLibraryItemIDs();
+		let searchResultGroups = SearchHandler.getOrderedSearchResultGroups(citedIDs);
 		for (let { key, group, isLibrary } of searchResultGroups) {
 			// selected items become a collapsible deck/list if there are multiple items
 			let isGroupCollapsible = key == "selected" && group.length > 1;
@@ -440,13 +440,8 @@ class LibraryLayout extends Layout {
 	updateSelectedItems() {
 		if (!libraryLayout.itemsView) return;
 		let selectedItemIDs = new Set(libraryLayout.itemsView.getSelectedItems().map(item => item.id));
-		for (let itemObj of CitationDataManager.items) {
-			if (selectedItemIDs.has(itemObj.zoteroItem.id)) {
-				itemObj.selected = true;
-			}
-			else {
-				itemObj.selected = false;
-			}
+		for (let bubbleItem of CitationDataManager.items) {
+			bubbleItem.selected = selectedItemIDs.has(bubbleItem.id);
 		}
 		IOManager.updateBubbleInput();
 	}
@@ -711,10 +706,9 @@ class LibraryLayout extends Layout {
 		if (event.key == "Backspace") {
 			let itemsToRemove = this.itemsView.getSelectedItems();
 			for (let item of itemsToRemove) {
-				let citationItems = CitationDataManager.getItems({ zoteroItemID: item.id });
-				for (let citationItem of citationItems) {
-					let { dialogReferenceID } = citationItem;
-					IOManager._deleteItem(dialogReferenceID);
+				let items = CitationDataManager.getItems({ itemID: item.id });
+				for (let item of items) {
+					IOManager._deleteItem(item.dialogReferenceID);
 				}
 			}
 		}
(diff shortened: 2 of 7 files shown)


adomasven:
This works, but it's pretty hard to figure out what's this full data structure, and why the need for it, even when I know how integration works just looking at the code. With the comment above the justification makes more sense, but could we make this more natural?

It's natural for me to think of there being a mapping between bubbles in the citation dialog and some data-structure/object/model that represents those in code. I see that you've made that an object with two members `citationItem` and `zoteroItem`. The naming is quite confusing. The object which corresponds to the bubble should probably be called `citationItem`, and contain or hold a reference to all data required to build it. Whereas `zoteroItem` here, is actually either a reference to a real Zotero Item in a library, or a virtual item (when the original item is removed and we are reconstructing from document-embedded data).

I assume you wanted to keep the stuff in `citationItem` from `zoteroItem` separate because that's the format that citeproc consumes? I think it makes the data structure here a little too confusing and cumbersome.

Could we have a CitationItem object, that has citation-specific prefs + a reference to Zotero Item (`_dataItem`, or something), either virtual or real, and store CitationItems as an array that we can reference, access, etc? CitationItem can have a method to export it to its final form (`serialize()` or `toJSON()`), so that we can get it into its final form for other consumers.

abaevbog:
Thank you for the feedback!

> I assume you wanted to keep the stuff in citationItem from zoteroItem separate because that's the format that citeproc consumes?

Right, exactly. To elaborate on my mental model, currently I see two types of objects here.
- When the dialog is loaded on an exiting citation, we get `io.citation.citationItems` objects (currently referred to as `citationItem`). For each of those objects, we have to get an instance of `Zotero.Item` (currently referred to as `zoteroItem`) to build bubbles.
- When an item from the library is being added, we already have a `Zotero.Item`. But then, locator/prefix/suffix added to it need to be stored somewhere, and it has to be something that is not `Zotero.Item` because otherwise we end up exactly with https://github.com/zotero/zotero/issues/5282. `citationItem` seems like the best place for it because this is what we already may get from `io.citation.citationItems` and what citeproc will consume when dialog is submitted.
- When we are accepting the dialog, we don't need `Zotero.Item`s anymore. We need objects with locator/suffix/prefix/etc, and maybe `cslItemData`, and `cslURIs`. Even though `cslItemData` and `cslURIs` are initially part of a surrogate `Zotero.Item` (for deleted but cited items), it makes sense to me to keep them with the locator/suffix/prefix in `citationItem` as well because they would eventually have to be together in `io.citation.citationItems` when the dialog is accepted.

> Could we have a CitationItem object, that has citation-specific prefs + a reference to Zotero Item

I agree this setup is somewhat convoluted. I thought this refactoring was helpful because it introduces a more rigid structure as to how the data is being converted, and it happens in fewer places. Moving all of this conversion and transformation into a new object is a great idea, and a very appropriate extension of what I was trying to do here. With that wrapper, we may not need that awkward `citationItem` too.

abaevbog:
@adomasven, what do you think about this approach?

I moved all data describing a citation entry into instances of a new `BubbleItem` class. Now `CitationDataManager.items` is an array of instances of that class instead of having objects like `{zoteroItem: ..., citaitonItem: ...}`, so each `bubbleItem` directly maps to a bubble node. There are two static methods to create a `BubbleItem`:
- `BubbleItem.fromCitationItem` used when a citation is being edited and we need to build bubbles from `io.citation.citationItems` on initial load
- `BubbleItem.fromItem` used in all other cases - when a new `Zotero.Item` (real or virtual) is being added into the citation. 

And then `bubbleItem.getCitationItem` will return a `citationItem` object for citeproc to consume. None of this changes any logic but the way data is stored should be easier to follow. `BubbleItem.fromCitationItem` is definitely better than such array construction from the first commit:
```
let citationAndZoteroItems = citationItems.map((citationItem) => {
	return {
		citationItem,
		zoteroItem: this._citationToZoteroItem(citationItem),
	};
});
```

A few side notes:
- I picked `BubbleItem` instead of `CitationItem` to avoid confusion with entries from `io.citation.citationItems`. I thought `BubbleItem` indicates that each instance of this corresponds to a bubble but I am also more than open to changing it to something else if you have a better idea.
- I added `cslItemID`, `cslURIs`, and `cslItemData` to be added to the virtual item created in `io.customGetItem` to be consistent with `Zotero.Integration.Citation.loadItemData`. That way it seems more consistent because we don't need to also check for those properties in `citationItem` in the [constructor of BubbleInput](https://github.com/zotero/zotero/blob/945690a1b9176a6ea7e4a59440e67a80177d1ef3/chrome/content/zotero/integration/citationDialog.js#L1451-L1453).
- These changes ended up being a bit more extensive than I first thought, so I also included the initial test suite for the citation dialog. It's about time to add some tests for the citation dialog anyways.

adomasven:
Seems like cited items do not appear in the "Cited" section unless the item is deleted from the library.

adomasven:
This is good. BubbleItem is perhaps not ideal, but let's not let that get in the way of good. Overall this makes the code much clearer.

> These changes ended up being a bit more extensive than I first thought, so I also included the initial test suite for the citation dialog. It's about time to add some tests for the citation dialog anyways.

Great! Let's just iron out the kinks and get this merged.

abaevbog:
> When running runtest.sh -g "Citation Dialog" I am getting local errors at least on Ubuntu 24.04.

Can you elaborate on what errors you are seeing? I get no issues on my mac. I ran the tests on my old Ubuntu laptop as well, and got no errors there either. Granted, it is on very old 16.04. which is way past EOL. I'll try to update to 24.04 to also run it there meanwhile.

> Seems like cited items do not appear in the "Cited" section unless the item is deleted from the library.

Can you share more details on how to reproduce this? I have a cited item in the library "A case study of ..." , and it appears in the cited tab below:

Screenshot 2025-05-15 at 10 54 29 AM

Existence of a cited item in the library is also part of the tests:
https://github.com/zotero/zotero/blob/5ba7da4477c352cddba46ea61dd1d39a69bac140/test/tests/citationDialogTest.js#L450-L460

While looking into this, I did notice the deduplication wasn't being applied to cited items. So if a cited item exists in the library and has an opened tab, the item card would be appearing twice. I don't think this was intentional (likely just a consequence of the confusing setup of real vs virtual cited items that we are fixing here), so I pushed a change to this as well.

adomasven:
<details>
<summary>Failing test run</summary>

```test
  Citation Dialog
Missing chrome or resource URL: chrome://zotero/skin/16/white/loading.svg
    Manage entries in the citation
    ❌️ [FAIL] should add an item to citation
      expected 0 to equal 1 at
@resource://zotero-unit-tests/citationDialogTest.js:166:14

    ❌️ [FAIL] should remove an item from citation
      expected 0 to equal 2 at
@resource://zotero-unit-tests/citationDialogTest.js:174:14

    ✅ should build citation with a cited item in library (1 ms)
    ✅ should build citation with a cited item not in library (1 ms)
    ❌️ [FAIL] should add a locator/suffix/prefix to a bubble
      expected 0 to equal 2 at
@resource://zotero-unit-tests/citationDialogTest.js:213:14

    ❌️ [FAIL] should change the order of bubbles
      bubbles[0] is undefined at
@resource://zotero-unit-tests/citationDialogTest.js:252:20

    ❌️ [FAIL] should sort the citation
      firstBubbleItem is undefined at
@resource://zotero-unit-tests/citationDialogTest.js:290:20

    ❌️ [FAIL] should update io.citation.items from bubbles
      bubbleItems[2] is undefined at
@resource://zotero-unit-tests/citationDialogTest.js:313:7

    UI
    ✅ should switch dialog mode (1 ms)
    ✅ should highlight bubbles whose items are selected (116 ms)
    ✅ should highlight rows of items in the citation (72 ms)
    Search
    ✅ should perform search in list mode (16 ms)
    ✅ should perform search in library mode (33 ms)
    Helpers.extractLocator
      Invalid locators
      ✅ has no locator label with numeric locator value (1 ms)
      ✅ has no locator label with textual locator value (1 ms)
      ✅ has no quotes around textual locator value (2 ms)
      ✅ has no locator value (0 ms)
      ✅ has textual locator value that does not follow locator label (0 ms)
      ✅ has numeric locator value that does not follow locator label (0 ms)
      ✅ has textual locator value not in the end of the string (1 ms)
      ✅ has numeric locator value not in the end of the string (1 ms)
      Valid locator labels
      ✅ is a valid numeric locator by itself (0 ms)
      ✅ is a valid textual locator by itself (0 ms)
      ✅ is a valid numeric locator with other text (8 ms)
      ✅ is a valid textual locator with other text (2 ms)

19/25 tests passed
JavaScript error: resource://devtools/client/shared/theme.js, line 99: NS_ERROR_ILLEGAL_VALUE: Component returned failure code: 0x80070057 (NS_ERROR_ILLEGAL_VALUE) [nsIObserverService.removeObserver]
JavaScript error: , line 0: NotFoundError: No such JSProcessActor 'BrowserToolboxDevToolsProcess'
```
</details>

These tests don't fail on macOS for me.

The cited item not appearing in the cited section is an issue on main too, so I've created a separate issue for it.

abaevbog:
The test question is odd. I guess I will have to bring my laptop up to 24.04 in hopes to reproduce this. But at the same time, github actions runs tests on Ubuntu 24.02 (which can't be that different) and they all pass there.

One guess I do have is this: we have a check that prevents items from being added when search is still running. If you just remove the  line with`SearchHandler.searching` check, does this have any effect on tests?

https://github.com/zotero/zotero/blob/5ba7da4477c352cddba46ea61dd1d39a69bac140/chrome/content/zotero/integration/citationDialog.js#L998-L999

The dialog re-runs search on initial load, when the dialog mode changes, and on window focus. And when the search runs for the first time, we also wait to make sure that the itemTree is fully loaded (otherwise, it throws an error initially somewhere deeper in libraryTree).

https://github.com/zotero/zotero/blob/5ba7da4477c352cddba46ea61dd1d39a69bac140/chrome/content/zotero/integration/citationDialog.js#L310-L314

So maybe the citation items are not getting added because search it taking longer to finish?

adomasven:
The tests pass now. Let's squash this and then merge.

abaevbog:
Great! I squashed everything into one commit with some minor cleanup and comment tweaks. Does this look good?

PR #5290, opened 2025-05-13
Fix itemTree broken keyboard navigation after deleting child row during quickSearch

### LINKED ISSUE #5280: Broken arrow up/down navigation in itemTree after trashing item during search
https://forums.zotero.org/discussion/comment/491934/#Comment_491934

I have been able to reproduce it somewhat reliably, though I am not sure about the exact steps to get it to happen. As described in the post, this occurs after deleting a child row when quickSearch is active. It does not always occur but after a few attempts, I ran into the `TypeError: this._rowOffsets[idx] is undefined` error that seems to be causing this.

This is the full video with my attempts to reproduce it - the error is thrown at the very end: [recording](https://www.dropbox.com/scl/fi/wybrpj30fags2ajpimubc/broken_keyboard_nav_in_search.mov?rlkey=9fijgshabl29gel89dxy3a26a&st=q2r1jk7y&dl=0)

abaevbog:
Fixed `itemTree.selection.focused` being set to `NaN` while trying to restore selection to a no-longer existing child row after it is trashed when `quickSearch` is active. Then, trying to find the next/previous row following/before row with index `NaN` would result in broken keyboard navigation of the tree.

Fixes: #5280

Before. Note, no row is selected, and `itemTree.selection.focused` is `NaN`. If at this point, up/down arrow is pressed, an error will be thrown in `windowed-list`, and keyboard navigation will be broken.

https://github.com/user-attachments/assets/079f3f0e-39a4-467a-aef2-3a113b9f6421

After. Since the child row is gone, the parent is re-selected:

https://github.com/user-attachments/assets/e31e8512-7b39-4f2e-abeb-4ce0781e4efc

### DIFF
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -3908,7 +3908,14 @@ var ItemTree = class ItemTree extends LibraryTree {
 						if (expandCollapsedParents) {
 							await this._closeContainer(this._rowMap[parent]);
 							await this.toggleOpenState(this._rowMap[parent]);
-							toggleSelect(selection[i].treeViewID);
+							// Re-select original row if it exists
+							if (this._rowMap[selection[i].treeViewID] != null) {
+								toggleSelect(selection[i].treeViewID);
+							}
+							// If it does not (e.g. child item moved to trash), select the parent
+							else {
+								toggleSelect(parent);
+							}
 						}
 						else {
 							!this.selection.isSelected(this._rowMap[parent]) &&
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -894,13 +894,29 @@ var ItemTree = class ItemTree extends LibraryTree {
 			}
 			// On removal of a selected row, select item at previous position
 			else if (savedSelection.length) {
+				// TEMP: trashing an item emits two events: modify and trash.
+				// When not in search mode, on 'modify', we attempt to process rows manually, which does nothing.
+				// The next 'trash' event reaches here and the selection is properly handled.
+				// In search mode, however, trashing an item reruns the search and sets madeChanges=true,
+				// which brings us here on the first 'modify' event. If not handled below, it will try
+				// to restore the selection, which may select the parent of the trashed item instead of the previous/next row.
+				// To avoid losing row selection, for now, assume that a 'modify' event in search
+				// could also represent item deletion/removal.
+				let potentialDeletionInSearch = collectionTreeRow.isSearchMode() && action == 'modify';
 				if ((action == 'remove'
 						|| action == 'trash'
 						|| action == 'delete'
-						|| action == 'removeDuplicatesMaster')
+						|| action == 'removeDuplicatesMaster'
+						|| potentialDeletionInSearch)
 					&& savedSelection.some(o => this.getRowIndexByID(o.id) === false)) {
+					// TEMP: see note above. If this is a 'modify' event that could be a deletion but is not,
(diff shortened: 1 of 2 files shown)


dstillman:
Why select the parent, though? Don't we want the same behavior you get when there's not a search, where the next row (i.e., the row that takes the place of the deleted one) becomes selected?

abaevbog:
That is a good point! I pushed a tweak to try to shift the selection on deletion to the next/previous row.

https://github.com/user-attachments/assets/0c42c3dc-524a-49ef-b339-72bde92d1a2f

One thing worth mentioning that in `_restoreSelection` (where the error originated from), we just try to restore selection to provided itemIDs. The logic of shifting the selection to the next/previous row is handled in `notify`:

https://github.com/zotero/zotero/blob/1f8e05bfd82187ccc7fc3edc73d51ee4e465f858/chrome/content/zotero/itemTree.jsx#L895-L935

With the last change to `notify`, we should no longer run into that edge case of trying to call `_restoreSelection` with a non-existent item, so the first commit is technically no longer necessary. But I think we should still keep it for the future because this is where the keyboard navigation breakage originated from, and we never need to re-select a non-existent row.

I tried to explain the nature of the last change in the (fairly) long comment in the last commit. But, to make sure it's clear what I have found, this is a summary:
- when an item is trashed, there are two events that are emitted: `modify` and `trash`. `modify` is higher in event order, so it is handled first
- when not in search, on `modify` rows are [processed manually](https://github.com/zotero/zotero/blob/1f8e05bfd82187ccc7fc3edc73d51ee4e465f858/chrome/content/zotero/itemTree.jsx#L674-L731), which does nothing. Then, the `trash` event is handled, which [removes the row ](https://github.com/zotero/zotero/blob/1f8e05bfd82187ccc7fc3edc73d51ee4e465f858/chrome/content/zotero/itemTree.jsx#L622-L648), sets `madeChanges=true`, and the previous/next row [is selected in the end of notify.](https://github.com/zotero/zotero/blob/1f8e05bfd82187ccc7fc3edc73d51ee4e465f858/chrome/content/zotero/itemTree.jsx#L895-L935)
- when in search, the first `modify` event is not a noop. It [causes a refresh](https://github.com/zotero/zotero/blob/1f8e05bfd82187ccc7fc3edc73d51ee4e465f858/chrome/content/zotero/itemTree.jsx#L657-L669), which also sets `madeChanges=true`. This means that we get to the part of `notify` that moves selection to the next/previous row on the `modify` event (not `trash`), which does not pass [this conditional](https://github.com/zotero/zotero/blob/1f8e05bfd82187ccc7fc3edc73d51ee4e465f858/chrome/content/zotero/itemTree.jsx#L895-L901) and sends us to [_restoreSelection](https://github.com/zotero/zotero/blob/1f8e05bfd82187ccc7fc3edc73d51ee4e465f858/chrome/content/zotero/itemTree.jsx#L936-L939) below. At that point, the selection will be moved (because the row no longer exists), so when `trash` event is handled next, we wouldn't find the proper next/previous row
- the best solution I found is to treat the first `modify` event in search as potentially equivalent to `trash` (if the item is deleted), so that we can move selection to the next/previous row then.
- alternatively, if we were to change the order of events to fire `trash` before `modify`, that would also fix this. But i assume this is not something we want to touch `Zotero.Notifier` for, and it should be fixed on the `itemTree` level.
- we also have #3435 where, among other things, we manually process updated rows in search mode. The approach from there might be a proper, not-hacky solution for this. The logic then could be somewhat the same regardless of if we are in search mode or not. This is why I marked the current fix as "temp".

Does this make sense?

PR #5292, opened 2025-05-13
Allow plugin to uninstall itself with update_url

windingwind:
Use discontinue: true in update_url json
"updates" that is valid for the installed plugin
to uninstall the plugin the next time it checks update

### DIFF
--- a/app/scripts/fetch_xulrunner
+++ b/app/scripts/fetch_xulrunner
@@ -287,7 +287,7 @@ function modify_omni {
 					return;
 				}
 				else {
-					logger.error(`AddonUpdateChecker: addon ]${aId} with discontinue flag not found or cannot be uninstalled`);
+					logger.error(`AddonUpdateChecker: addon \${aId} with discontinue flag not found or cannot be uninstalled`);
 					sendUpdateAvailableMessages(this, null);
 					return;
 				}
--- a/app/scripts/fetch_xulrunner
+++ b/app/scripts/fetch_xulrunner
@@ -234,10 +234,10 @@ function modify_omni {
 	replace_line "id: TOOLKIT_ID," "id: '$APP_ID'," modules/addons/AddonUpdateChecker.sys.mjs
 	replace_line 'lazy.AddonManagerPrivate.webExtensionsMinPlatformVersion' '"7.0"' modules/addons/AddonUpdateChecker.sys.mjs
 	replace_line 'result.targetApplications.push' 'false && result.targetApplications.push' modules/addons/AddonUpdateChecker.sys.mjs
-	# Set discontinue flag to true if the update config has it set
-	replace_line 'let result = \{' \
-		'let result = {
-			discontinue: getProperty(update, "discontinue", "boolean"),' modules/addons/AddonUpdateChecker.sys.mjs
+	# Set uninstall flag to true if the update config has it set
+	replace_line 'targetApplications: \[appEntry\],' \
+		'targetApplications: \[appEntry\],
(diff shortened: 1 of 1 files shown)


windingwind:
A test example: 
[make-it-red-1.9.zip](https://github.com/user-attachments/files/20182630/make-it-red-1.9.zip)

The update URL points to https://gist.githubusercontent.com/windingwind/060b2ebb027c481857ded581027e84ef/raw/2d576b75e075cc2b972882f3f09291751ef38666/sample-discontinue-update.json, where the `discontinue` flag is set.

Currently, the plugin is uninstalled silently. We can decide if we want to show an alert or banner (we're having a lot of banners recently)

app/scripts/fetch_xulrunner
```diff
@@ -266,6 +270,29 @@ function modify_omni {
...
+					try {
+						addon.uninstall();
+					} catch (err) {
+						logger.error(`AddonUpdateChecker: error uninstalling addon \${aId} with discontinue flag`, err);
+					}
+					sendUpdateAvailableMessages(this, null);
+					return;
+				}
+				else {
+					logger.error(`AddonUpdateChecker: addon ]${aId} with discontinue flag not found or cannot be uninstalled`);
```
AbeJellinek, line 290:
```suggestion
					logger.error(`AddonUpdateChecker: addon \${aId} with discontinue flag not found or cannot be uninstalled`);
```

I'm assuming.

But is this escaping for Perl or the shell? If it's the shell, we don't need it in a single-quote string:

```bash
$ echo '$PATH'
$PATH
$ echo '\$PATH'
\$PATH
```

windingwind: Thanks, sorry for missed that! I don’t know where it came from, but we need it here.

app/scripts/fetch_xulrunner
```diff
@@ -234,6 +234,10 @@ function modify_omni {
 	replace_line "id: TOOLKIT_ID," "id: '$APP_ID'," modules/addons/AddonUpdateChecker.sys.mjs
 	replace_line 'lazy.AddonManagerPrivate.webExtensionsMinPlatformVersion' '"7.0"' modules/addons/AddonUpdateChecker.sys.mjs
 	replace_line 'result.targetApplications.push' 'false && result.targetApplications.push' modules/addons/AddonUpdateChecker.sys.mjs
+	# Set discontinue flag to true if the update config has it set
+	replace_line 'let result = \{' \
```
dstillman, line 238: This seems a little fragile. Let's use one of the more distinctive properties.

app/scripts/fetch_xulrunner
```diff
@@ -234,6 +234,10 @@ function modify_omni {
...
+		'let result = {
+			discontinue: getProperty(update, "discontinue", "boolean"),' modules/addons/AddonUpdateChecker.sys.mjs
```
dstillman, line 240: How about `uninstall`?

windingwind:
Updated. Here's the updated example:

[make-it-red-1.9.zip](https://github.com/user-attachments/files/20231101/make-it-red-1.9.zip)

dstillman:
This isn't working for me, but maybe just because https://gist.githubusercontent.com/windingwind/060b2ebb027c481857ded581027e84ef/raw/fe8d8776f3d2edd121f801c11ff6a979881a5ce9/sample-discontinue-update.json has `"strict_max_version": "6.999"`?

windingwind:
Oops, sorry. Indeed that's the cause. Updated:

[make-it-red-1.9.zip](https://github.com/user-attachments/files/20239909/make-it-red-1.9.zip)

dstillman:
Works. Awesome.

PR #5297, opened 2025-05-13
Fix issue where subsequent updates provide stale changed object data

tnajdek:
It seems that after 2bc0c48, the `changed` object does indeed contain previous values (thanks!), but this works only once—subsequent updates provide stale `changed` object data.

I encountered this while working on #3860, so I’ve extended the test to cover this scenario and attempted to fix the problem by clearing `_previousData` in `postSave`—no idea if that's the correct approach.


tnajdek:
It seems tests are failing so I suspect this is not the right approach!

dstillman:
(I'm looking into the test failure.)

dstillman:
OK, I think that will do it. `_clearChanged()` wasn't clearing primary or item data in `_previousData` when called explicitly for those categories. The relations failure was an unrelated bug that just surfaced because clearing `_changed` manually in `_postSave()` was incorrect and prevented relations from being reloaded. (Reloading data after every save is a major performance drain and likely unnecessary, but that's another matter.)

PR #5298, opened 2025-05-14
Fix locate button for annotation item

windingwind:
For any selected annotation items in library, the locate button will show operations for its parent item, instead of showing nothing.

### DIFF
--- a/chrome/content/zotero/locateMenu.js
+++ b/chrome/content/zotero/locateMenu.js
@@ -40,7 +40,7 @@ var Zotero_LocateMenu = new function() {
 			locateMenu.removeChild(locateMenu.firstChild);
 		}
 		
-		var selectedItems = _getSelectedItems();
+		var selectedItems = await _getSelectedItems();
 		
 		if(selectedItems.length) {
 			await _addViewOptions(locateMenu, selectedItems, true, true, true);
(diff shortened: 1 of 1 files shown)


chrome/content/zotero/locateMenu.js
```diff
@@ -349,7 +349,12 @@ var Zotero_LocateMenu = new function() {
 		var selectedItems = [];
 		while (selectedItems.length < 50 && allSelectedItems.length) {
 			var item = allSelectedItems.shift();
-			if (!item.isNote() && !item.isAnnotation()) selectedItems.push(item);
+			if (item.isAnnotation()) {
+				selectedItems.push(item.parentItem);
```
dstillman, line 353: If the attachment has a parent regular item and the attachment is the primary attachment (`getBestAttachment()`, for now), we should probably use the regular item instead. The Locate menu doesn't make a ton of sense for attachments — the open options are fine, but Google Scholar, which currently shows for everything, is totally pointless, since it will do a search for literally "Full Text PDF".

We can still use the attachment for non-primary attachments (and then, separately from this, we should probably just hide all locate engines for non-regular items).

windingwind:
Thanks, updated.

chrome/content/zotero/locateMenu.js
```diff
@@ -344,12 +344,23 @@ var Zotero_LocateMenu = new function() {
...
-	function _getSelectedItems() {
+	async function _getSelectedItems() {
 		var allSelectedItems = ZoteroPane.getSelectedItems();
 		var selectedItems = [];
 		while (selectedItems.length < 50 && allSelectedItems.length) {
 			var item = allSelectedItems.shift();
-			if (!item.isNote() && !item.isAnnotation()) selectedItems.push(item);
+			if (item.isAnnotation()) {
+				let attachment = item.parentItem;
+				if (attachment.parentItem && attachment.id === await attachment.parentItem.getBestAttachment()?.id) {
```
dstillman, line 354: Should be able to just compare the object — no need to compare the ids.

windingwind: Thanks, updated

PR #5301, opened 2025-05-16
Overhaul testing in Scaffold

AbeJellinek:
I'm blocked on #5219 until I can figure out how to redirect network events, so in the mean time, here are some improvements that take advantage of (and update Scaffold to be compatible with) the `TranslatorTester` rewrite:

image

- Use `TranslatorTester` on a fresh page in `HiddenBrowser` to create tests
	- This unifies the code for creating tests and running them, and makes it harder to create a test that will fail later
	- `"defer": true` is set automatically when necessary
- New UI for tests:
	- Double-click to run
	- After running, you get a short summary of changes like
	image
    and you can click (or right-click -> Apply Updates) to apply them. "Update" is no longer a separate action, but something you can choose to do after running, which I think makes more sense.
- Test results show in a mock item pane with a colored diff, just like outputs from browser runs do
- No more "Not run" test status, to reduce visual noise
- Add striped background to listbox rows
- No more Run/Update/Delete buttons, just as we don't have big buttons to open or delete items in the item tree. Apply Updates is in the context menu and can be triggered by clicking the diff summary; all the rest are accessible via context menu and keyboard.
- Fix misaligned columns in test listbox

### DIFF
--- a/chrome/content/scaffold/scaffold.js
+++ b/chrome/content/scaffold/scaffold.js
@@ -26,10 +26,12 @@
 var { E10SUtils } = ChromeUtils.import("resource://gre/modules/E10SUtils.jsm");
 var { Subprocess } = ChromeUtils.import("resource://gre/modules/Subprocess.jsm");
 var { RemoteTranslate } = ChromeUtils.import("chrome://zotero/content/RemoteTranslate.jsm");
-var { ContentDOMReference } = ChromeUtils.import("resource://gre/modules/ContentDOMReference.jsm");
 
 var { Zotero } = ChromeUtils.importESModule("chrome://zotero/content/zotero.mjs");
 var { FilePicker } = ChromeUtils.importESModule('chrome://zotero/content/modules/filePicker.mjs');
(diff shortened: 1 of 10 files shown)


AbeJellinek:
Rebased and changed target to `fx140`.

dstillman:
I haven't yet tested this, but it seems a little confusing not to have a visible button for updating the test? It just seems like a common action shouldn't be buried in the context menu.

dstillman:
Or maybe there is a visible element after you click on "Data mismatch"?

AbeJellinek:
There is - the `Data mismatch` link. You update after running the test, so you know what the new item data will be (versus the old system, where it would rerun the test in order to update it, which could fail/return different data).

AbeJellinek:
When you click Data mismatch, it pops up a prompt asking if you want to update.

dstillman:
I think that's still a bit hidden and unexpected, but this is awesome in general, so merging for now. I'll open separate issues for a couple minor things.

AbeJellinek:
Thank you!

PR #5302, opened 2025-05-16
Persist open state of attachment rows in itemTree

### LINKED ISSUE #5252: Convert between highlight and underline annotation collapses attachment row
https://forums.zotero.org/discussion/123934/convert-between-highlight-and-underline-hides-the-annotation-from-the-items-list

Specifically if the attachment has no other annotations but the one being converted. It seems like the reason is that the highlight annotation is deleted first, which will remove the only highlight annotation row child from`itemTree` and trivially collapse the attachment. When the underline annotation is created, the attachment row just remains collapsed.

abaevbog:
- record in a persistent cache which itemTree rows were toggled open. When a row is being expanded, check if its children are meant to be open, and if so - expand them as well
- that way, if one expands an attachment, collapses the parent item, and then expands the parent item, the annotations will be visible
- if an attachment is expanded and its annotations are all removed in the reader, it will technically count as a closed container. But if some new annotations are added later, the attachment will reopen. This also applies to
attachments with just one annotation that is converted between highlight and underline, since it essentially deletes the old annotation and creates a new one
- when a child item (e.g. note) is modified, the parent is collapsed and re-opened during sorting. If another child item is an expanded attachment, this will ensure that it remains open
- for consistency, if an attachment is expanded and is moved to/from top-level items, it will remain open

https://forums.zotero.org/discussion/123933/persistent-expand-collapse-state-for-annotations-in-the-items-list
https://forums.zotero.org/discussion/123934/convert-between-highlight-and-underline-hides-the-annotation-from-the-items-list

Fixes: #5252

Demo: [recording](https://www.dropbox.com/scl/fi/r2lkhil2er562onznqyib/persistent_attachment_rows_open_state.mov?rlkey=3svsob8316d6ttasetdrg5b6i&st=zexvq4t1&dl=0)

### DIFF
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -165,6 +165,7 @@ var ItemTree = class ItemTree extends LibraryTree {
 		}
 		
 		this._itemTreeLoadingDeferred = Zotero.Promise.defer();
+		this._openedRowsCache = new Set();
 	}
 
 	unregister() {
@@ -259,6 +260,7 @@ var ItemTree = class ItemTree extends LibraryTree {
 		
 		try {
 			Zotero.CollectionTreeCache.clear();
+			this._openedRowsCache.clear();
 			// Get the full set of items we want to show
 			let newSearchItems = await this.collectionTreeRow.getItems();
 			if (this.collectionTreeRow.isTrash()) {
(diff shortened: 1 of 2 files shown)


abaevbog:
This also addresses: https://forums.zotero.org/discussion/124581/attach-file-collapses-the-annotations-in-the-items-list

PR #5304, opened 2025-05-19
Fix sidenav drop indicator position

windingwind:
regression after #5263


windingwind:
Before:

image

After:

image

PR #5307, opened 2025-05-19
Fix sidenav notes button not switching to notes list in context pane

### LINKED ISSUE #5306: Sidenav notes button in reader tab does not display notes list on Space
https://forums.zotero.org/discussion/124326/cannot-open-notes-pane-from-keyboard

When the notes button in sidenav of the reader tab is activated with Space, focus just moved to the context pane without displaying the list of notes.

This was initially implemented in https://github.com/zotero/zotero/pull/4235.

abaevbog:
Also, try to not loose focus when toggling context pane open/closed. If one toggle-pane button disappears, and the other one appears - move focus to the new button.

Fixes: #5306

https://github.com/user-attachments/assets/147f3e0f-d9fb-47f2-9a1e-f3c9abd2dee1

### DIFF
--- a/chrome/content/zotero/elements/itemPaneSidenav.js
+++ b/chrome/content/zotero/elements/itemPaneSidenav.js
@@ -665,21 +665,25 @@
 					});
 					firstBtn.dispatchEvent(clickEvent);
 				}
-				setTimeout(() => {
-					// If notes are visible, tab into them
-					if (this._contextNotesPaneVisible) {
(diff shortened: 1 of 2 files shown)


abaevbog:
The issue of focus being lost on keyboard-click of toggle-pane button was also reported here: https://forums.zotero.org/discussion/124325/lost-focus-on-toggle-context-pane

Issue #5247, opened 2025-05-01
Render markup in annotation rows?


AbeJellinek:
https://forums.zotero.org/discussion/123995/annotations-in-the-item-pane-cannot-show-comments-formatting

Via `Zotero.Utilities.Internal.renderItemTitle()`, probably. #4607 extends that method to handle arbitrary HTML, so we could rename the new mode to `'html'`(?), merge that PR, and use it for annotations as well.

abaevbog:
I was just looking at the discussion we had about this a while back https://github.com/zotero/zotero/pull/3416#discussion_r1727252473. Crazy timing!

We would likely also need this in order to make the comments in annotation cards editable #5229

abaevbog:
My thinking was to use `simpleEditor.html` as the rich text editor, with a few modifications. We wouldn't want to insert the `iframe` with it for every annotation card, so I thought we would only do it if one clicks on the `div` with the comment. Then we would hide the `div` and create and insert the `iframe`. But for it to visually make sense, `div` with the comment still needs to have the content properly rendered, which is where the title rendering routing would come in. Does this make sense?

Issue #5241, opened 2025-04-30
Sidenav visibility inconsistent with annotation in item pane


windingwind:
Select note -> annotation: invisible
Select item -> annotation: visible.

Actually, we should probably make the item pane context available to plugins, and only disable the sidenav if there are no other (custom) sections in this view.

I could take this if we want to expose this interface to plugins. What do you think? (maybe @abaevbog)

abaevbog:
That makes sense to me! I suspect there may be a number of use cases for customization of `annotationsItemPane`.

PR #5312, opened 2025-05-22
Editable annotation cards in itemPane

### LINKED ISSUE #5229: Make annotations editable in item pane
https://forums.zotero.org/discussion/123889/adding-tags-and-comments-to-annotations-from-the-items-list

Do we have any way of sharing the component (in a separate library that we pull into each?), or do we just want reimplement some of the editable parts?

### LINKED ISSUE #5262: External annotations can be deleted from item tree
And don’t show a lock icon to indicate that they’re external.

https://forums.zotero.org/discussion/124063/external-annotations-can-be-deleted-from-the-items-list

abaevbog:
- tags in the annotation cards will open a popup with tags section, same as in the reader. The functionality of creating and opening the popup is shared in new `Zotero.Annotations.insertAnnotationsTagsPopup`.
- options button in the top-right corner of the annotation card will open a menu with colors and "Delete" option
- keyboard navigation somewhat similar to one in the sidenav of the reader: arrowUp/Down will navigate
between annotations, tab will navigate through annotation comment and tags. Next tab will move focus out of the annotations list.
- annotation comment can be edited. On click, `simpleEditor.html` iframe will be inserted instead of the comment div. There, one has rich text editor capabilities, same as in the reader. `simpleEditor.html` iframe is used because `contenteditable` in the root chrome window does not work properly (this issue must have been the reason why the original `quickFormat` dialog had the `iframe` too) . Edits are saved on blur, or on Cmd/Ctrl+Enter. Escape cancels the edits. 
- refactoring of `simpleEditor` to allow for easier customization via attributes placed on the iframe. Possible attributes are: `actions` (to specify which actions should appear), `tight` (to indicate that the buttons should be smaller than default), and `toolbar-below-editor` (to indicate where toolbar should be placed). This way, simpleEditor can be customized here without affecting editBibliographyDialog. For convenience, moved simpleEditor stylesheet into new _simpleEditor.scss. I expect this customization will also help in the redesign of the `editBibliographyDialog`.
- do not set `aria-expanded` on non-expandable header of `collapsible-section`, since that implies the section can be expanded.

Notes: 
- marking this as a draft for now until https://github.com/zotero/zotero/issues/5247 is resolved. Currently, we strip all html tags from annotations. But https://github.com/zotero/zotero/pull/4607 should allow us to properly render them safely and perform any conversions necessary, as in the reader. Functions like `_commentFromHTML` and `_commentToHTML` of  `annotation-row` would likely be affected by it too, and for now are just minimal implementations to get some editing functionality working. 
- there is also a question about editable and delete-able annotations discussed in #5262. In reader, external annotations or annotations made by other users cannot be edited or deleted. However, external annotations are editable and delete-able on the `item.js` level. Annotations made by other users can not be edited but still can be deleted. I am not exactly sure about the intended behavior here. For now, external and other user's annotation cards cannot be interacted with, same as in the reader. Backspace on external annotation rows also will not delete them but backspace on group annotation rows will (as this is explicitly done in https://github.com/zotero/zotero/commit/df64a16b55c17264949f30cfc9edeb4c098f169a). Correct me if this is off.

Fixes: #5262
Fixes: #5229

### DIFF
--- a/chrome/content/zotero/elements/annotationItemsPane.js
+++ b/chrome/content/zotero/elements/annotationItemsPane.js
@@ -106,6 +106,7 @@
 					let row = document.createXULElement('annotation-row');
 					row.annotation = annotation;
 					row.container = this._body;
+					row.draggable = true;
 					section.querySelector('.body').append(row);
 				}
 			}
--- a/chrome/content/zotero/elements/annotationItemsPane.js
+++ b/chrome/content/zotero/elements/annotationItemsPane.js
@@ -105,6 +105,7 @@
 					if (this.querySelector(`annotation-row[annotation-id="${annotation.id}"]`)) continue;
 					let row = document.createXULElement('annotation-row');
 					row.annotation = annotation;
+					row.container = this._body;
 					section.querySelector('.body').append(row);
 				}
 			}
--- a/chrome/content/zotero/elements/annotationRow.js
+++ b/chrome/content/zotero/elements/annotationRow.js
@@ -37,7 +37,6 @@
 			<html:div class="body">
 				<html:img/>
 				<html:div class="quote"/>
-				<html:div class="comment keyboard-clickable" tabindex="0" role="button"/>
 			</html:div>
 			<html:div class="tags keyboard-clickable" tabindex="0"/>
 		`);
@@ -85,10 +84,16 @@
 			this._body = this.querySelector('.body');
 			this._tags = this.querySelector('.tags');
 			this._options = this.querySelector('.head toolbarbutton');
-			this._comment = this.querySelector('.comment');
 			this._img = this.querySelector('img');
 			this._quote = this.querySelector('.quote');
 			this._commentInEditor = null;
+
+			// <html:div> placed directly in content string will strip all
+			// html tags when .innerHTML is set. But a node created dynamically won't.
+			this._comment = document.createElement('div');
+			this._comment.className = "comment keyboard-clickable";
+			this._comment.setAttribute('role', 'button');
+			this._quote.after(this._comment);
 
 			this.addEventListener('keydown', this._handleKeyDown.bind(this));
 			this._tags.addEventListener('click', this._handleTagsClick.bind(this));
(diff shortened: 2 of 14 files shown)


windingwind:
That sounds natural. Thanks for making this.

I have something about annotations in the item pane in mind, as we discussed in #5241, and I think the easiest way is to use the current `ItemDetails` for single annotation selection, where other custom sections can automatically work; for multiple annotation selection, it uses the `annotationsItemPane`.

The thing is, do we want to use the already-there tags box in the `ItemDetails` if only one annotation is selected? I suppose no, and we stick to the implementation here to edit the annotation tags, right?

abaevbog:
Gotcha, yeah that makes sense! Thank you for the explanation. 

I see that `itemDetails` has all the logic to handle custom sections. Do I understand correctly that the idea is to avoid duplicating that logic in `annotationItemPane`?

>  if any custom section for single annotation selection is added, the single annotation selection item pane becomes the way it works with other item types, like with several item pane sections and collapsible/pinnable.

What about custom sections for when multiple annotations are selected? Is that something we are not considering at the moment?

abaevbog:
> Yes, and besides that, the custom sections and the API by design don't accept multiple items

Oh, I see! That makes sense. It actually clarifies the main question for me.

I kept thinking about multiple items being selected because it is really the main value of the having annotation rows in itemTree (e.g. so you can find and do something to all annotations fitting some criteria across many items). But if custom sections for multiple annotations is not something we are after at the moment, using `itemDetails` for a single selection makes perfect sense to me, thank you for the explanation!

abaevbog:
Need to add ability to select annotation cards, and perform operations via open context menu, similar to itemTree rows. Might as well add drag-drop for consistency as well.

abaevbog:
The last set of updates contains improved navigation/selection of annotation rows. Now, one can select multiple rows and use the context menu to create/add a note, change colors, or delete the selected annotations. Also, one can drag-drop selected rows into an editor for quickCopy. This is the brief demo of the current behavior:

https://github.com/user-attachments/assets/de133dfe-7e42-49ee-8c1c-4445136a2729


This does have only a subset of what is available in the reader but I also think there's no need to replicate everything (e.g. changing font-size or page number is probably not relevant here). Let me know if there's anything else that is possible in the reader that is also worth including here.

PR #5313, opened 2025-05-23
Undo Scholaread changes to `user.js`

AbeJellinek:
Scholaread sets `extensions.zotero.fileHandler.pdf` in `<profile>/user.js`, not `prefs.js`, so changing the pref via the UI does nothing (`user.js` takes precedence). This PR checks `user.js` for `user_pref()` directives, and if any are found, removes them, shows an alert, and restarts.

As far as I can tell (I haven't run it), Scholaread asks the user before adding its pref, so this _shouldn't_ happen every time you start Zotero. But if it does, maybe that's more impetus for them to stop setting it in the wrong location?

### DIFF
--- a/chrome/content/zotero/xpcom/prefs.js
+++ b/chrome/content/zotero/xpcom/prefs.js
@@ -560,22 +560,34 @@ Zotero.Prefs = new function() {
 			return;
 		}
 		
-		const PREF_RE = /^\s*user_pref\s*\(\s*['"]extensions\.zotero\..+$/gm;
-		if (PREF_RE.test(userJS)) {
-			Zotero.debug('user.js contains prefs:');
-			Zotero.debug(userJS);
-			
-			await Zotero.File.putContentsAsync(userJSPath, userJS.replace(PREF_RE, ''));
-			
-			Zotero.alert(null,
-				Zotero.getString('general-error'),
-				Zotero.getString('userjs-pref-warning'
-					+ (userJS.toLowerCase().includes('scholaread') ? '-program' : ''))
-			);
-			Services.startup.quit(
-				Components.interfaces.nsIAppStartup.eAttemptQuit
-				| Components.interfaces.nsIAppStartup.eRestart
-			);
+		const PREF_RE = /^\s*user_pref\s*\(\s*['"]extensions\.zotero\..+$/g;
+		
+		let updatedUserJS = userJS
+			.split('\n')
+			.filter((prefLine) => {
+				if (PREF_RE.test(prefLine)
+						&& prefLine.includes('extensions.zotero.fileHandler.')
+						&& prefLine.toLowerCase().includes('scholaread')) {
+					Zotero.debug('user.js contains disallowed pref: ' + prefLine);
+					return false;
+				}
+				return true;
+			})
+			.join('\n');
+		
+		if (updatedUserJS === userJS) {
+			return;
 		}
+		
+		await Zotero.File.putContentsAsync(userJSPath, updatedUserJS);
+		
+		Zotero.alert(null,
+			Zotero.getString('general-error'),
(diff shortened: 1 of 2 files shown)


dstillman:
A program definitely shouldn't be setting itself as a file handler in user.js, but I'm not sure we want to block use of user.js altogether. Maybe enough to just look for the handler settings specifically and warn on those?

AbeJellinek:
This doesn't complain about/remove all `user.js` customizations, just ones that set prefs. And why not block those? Is there any pref you can set in `user.js` that you can't set in `prefs.js`? If not, since `user.js` prefs override `prefs.js` prefs and block changes made from the UI (as we've seen with the Scholaread situation), it just seems like a bad idea to allow them.

As a less pushy option, maybe we could silently move prefs set in `user.js` to `prefs.js`, unless one of them is a known bad override (like the Scholaread file handler pref), in which case we delete it and pop up the alert.

AbeJellinek:
Well, I guess still kind of pushy, but less annoying.

dstillman:
> This doesn't complain about/remove all user.js customizations, just ones that set prefs.

But that's what user.js is, right? It's just a mechanism for overriding prefs in a way that resets them at every startup.

I'm just saying, since this is a valid Firefox configuration mechanism, I'm not sure why we'd need to block it — for all we know people have valid use cases for it. We just don't want shady programs hijacking the handler settings.

AbeJellinek:
Yeah, good point. I didn't realize it could *only* override prefs - I think I had it confused with `userChrome.js`.

I'll restrict it to the one pref for the moment.

chrome/content/zotero/xpcom/prefs.js
```diff
@@ -542,4 +544,50 @@ Zotero.Prefs = new function() {
...
+		}
+		
+		const PREF_RE = /^\s*user_pref\s*\(\s*['"]extensions\.zotero\..+$/g;
+		
+		let updatedUserJS = userJS
+			.split('\n')
+			.filter((prefLine) => {
+				if (PREF_RE.test(prefLine)
+						&& prefLine.includes('extensions.zotero.fileHandler.')
+						&& prefLine.toLowerCase().includes('scholaread')) {
```
dstillman, line 570: I was suggesting we just check for the handler prefs here, not Scholaread specifically. This is one of the prefs that's most likely to be abused, and there's no legitimate reason for it to be overridden here.

chrome/content/zotero/xpcom/prefs.js
```diff
@@ -542,4 +544,50 @@ Zotero.Prefs = new function() {
...
+				}
+				return true;
+			})
+			.join('\n');
+		
+		if (updatedUserJS === userJS) {
+			return;
+		}
+		
+		await Zotero.File.putContentsAsync(userJSPath, updatedUserJS);
```
dstillman, line 582: This should probably be in a try/catch as well. Since this more or less by definition will involve the file being written by another program, it's not impossible it could become unwritable, and we don't want to break startup while trying to fix it.

AbeJellinek:
OK, now it looks for fileHandler prefs specifically, and I brought back the Scholaread-specific warning since it seems better to be less vague when we know what's at fault for overriding the pref. Also added a try-catch.

Let me know if that's still not what you mean!

dstillman:
How about this?

Once this is in place, no program is going to try to do this again, so it's sort of pointless to have a separate message about it, and the other message was — as you say — kind of vague. So how about we just say "Some Zotero settings have been overridden using an unsupported method. Zotero will revert them and restart."?

AbeJellinek:
Yeah, I mean it’s still a little vague and ominous, but maybe that’s as good as we can do. It’ll definitely stop anyone from trying this again!

Issue #5314, opened 2025-05-23
Citation dialog: allow inserting annotations


abaevbog:
https://forums.zotero.org/discussion/100781/can-we-make-a-bibliography-from-a-note

Currently, one has to create a note from annotations and then use "Insert note" dialog to insert those annotations. Now, we can allow one to insert annotation directly from the itemTree but selecting the annotation row in library mode and skip the extra step of making a note.

It is a bit more complicated with the list mode. We would want to be able to search annotation by their top-level item, and display them in an hierarchical fashion as child rows of the parent items as well. This is quite similar to #2812 and #2389.

A few questions/thoughts about the use cases:
- Should adding annotations behave similar to adding a citation or adding a note? If you select an item, does the dialog immediately auto-accept, or do you select multiple annotations, which appear as bubbles, and then press Enter?
- What is the best way to differentiate between "Insert annotation" and "Add note"? They are somewhat similar, though a bit different. We could add "Insert annotation" as a new button into the Google docs/Word/Libre Office toolbar, and open the dialog in the right mode based on that. 
- Per the forums post, the distinction of a note vs annotation is at times blurry because you can already insert annotations as long as you create a note with those annotations first. So, instead of forcing differentiation, we could consider keeping these two modes as one? So you can either select a note or an annotation. Describing this functionality might be hard but at least you wouldn't have to think which of the two buttons to click. I am noticing that even now, having "Add citation" and "Insert note" buttons sometimes causes confusion (e.g. https://forums.zotero.org/discussion/124082/zotero-is-not-working) - adding "Insert annotation" may lead to more confusion down the road.

dstillman:
> Should adding annotations behave similar to adding a citation or adding a note?

In between? Because maybe you want multiple, but there's no need to order them — they'd be in document order — so there's no need for them to bubbleize, and that would be an annoying extra step for single annotations. So it kind of seems like you should be able to select multiple and then insert them directly. But then we'd probably have to remove the + buttons in the items list.

You should also be able to select a top-level item or attachment and get all the annotations under it, in document order. And I guess even multiple top-level items or attachments.

Later, we may even want to add the tag selector to the citation dialog so that you could insert all annotations with a given tag across a collection or library.

Basically, most of the things you had to create a note for should be possible to do directly. People can still create notes if they want to work further within Zotero, but it shouldn't be necessary.

> What is the best way to differentiate between "Insert annotation" and "Add note"?

[...]

PR #5315, opened 2025-05-23
Preference to hide context annotation rows

### LINKED ISSUE #5264: Consider a pref to hide context annotation rows?
https://forums.zotero.org/discussion/comment/491797/#Comment_491797 and a few earlier posts in the same threat.

While working on the performance optimization for attachments with many annotations, I noticed this issue myself. Suppose you want to review all annotations for a specific subject. You can get all the annotations by using a search, which will also expand all attachments. At that point, I find the `itemTree` somewhat hard to use because it is overwhelmed by non-matching context annotation rows.

https://github.com/user-attachments/assets/de98e9bc-ce63-4068-8db6-0c4211582da9

Maybe it's worth considering a separate pref to "Omit non-matching annotation rows", so that if one performs many searches across annotations, `itemTree` is not cluttered by non-matches? It is also possible that it won't be a problem once you can edit annotations in `itemPane` and when we add more detailed info from parent item into the annotation sections.

abaevbog:
This may be helpful while searching across many annotations to not be overwhelmed by non-matching rows. In the example below, I have a PDF with a few hundred annotations, and while searching for "evolution", almost the entire visible scrollbox is filled with irrelevant context rows. With this setting, only rows related to "evolution" remain.

https://github.com/user-attachments/assets/fe0bede8-2162-4233-ab61-e416c2f87c47

It does go against the currently existing practice of just displaying all child rows. But given how many annotation child rows there can be, maybe it's worth it?

Fixes #5264

### DIFF
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -154,6 +154,10 @@ var ItemTree = class ItemTree extends LibraryTree {
 				'itemTreeView',
 				50
 			);
+			this._prefsObserverID = Zotero.Prefs.registerObserver('hideContextAnnotationRows', async () => {
+				await this.refresh();
+				this.tree.invalidate();
+			});
 		}
 		
 		this._itemsPaneMessage = null;
@@ -170,6 +174,7 @@ var ItemTree = class ItemTree extends LibraryTree {
 	unregister() {
 		this._uninitialized = true;
 		Zotero.Notifier.unregisterObserver(this._unregisterID);
+		Zotero.Notifier.unregisterObserver(this._prefsObserverID);
 		this._writeColumnPrefsToFile(true);
 	}
 
@@ -1724,6 +1729,9 @@ var ItemTree = class ItemTree extends LibraryTree {
 
 		if (item.isFileAttachment()) {
 			annotations = item.getAnnotations();
+			if (Zotero.Prefs.get("hideContextAnnotationRows") && this._searchMode) {
+				annotations = annotations.filter(annotation => this._searchItemIDs.has(annotation.id));
+			}
 		}
 		var newRows = [];
 		if (attachments.length && notes.length) {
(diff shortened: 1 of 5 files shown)


dstillman:
I think we'd want this to be a setting in View, next to Show Items from Subcollections, and I don't think it needs to be specific to annotations — it could just hide all context rows that don't have matching descendant rows. "Show Only Matching Items"?

abaevbog:
That sounds good! Making the pref available through the view menu definitely makes it easier to access.

How does this new behavior look? You'll notice the context PDF attachments and notes that don't have "evolution" are now also removed during filtering:

https://github.com/user-attachments/assets/098b0422-a5db-4816-869f-afac225c816f

abaevbog:
Yeah, I guess in case of "Show Only Matching Items", the main keyword would actually be "Only". "Hide Non-Matching Items" is a bit more expressive, right?

dstillman:
Yeah, exactly.

dstillman:
This seems to be causing lots of `Error: getAnnotations() can only be called on file attachments  item.js:4171:11` lines to be thrown when I turn it on.

abaevbog:
Thank you for pointing it out! Non-file attachments were not being skipped when checking for matching annotations, which is where the error came from. It took me a second to run into it because I didn't have any non-file attachments. This should be fixed now

abaevbog:
We will need this update for the annotations in citation dialog PR https://github.com/zotero/zotero/issues/5314 to be able to hide irrelevant non-matching child rows of parent items (e.g. attachments without annotations or notes).

That made me realize that this "Show Only Matching Items" option may be useful not only in ZoteroPane, and not only in citation dialog - but everywhere the itemTree is used. Would it make sense to move that menu from `zoteroPane.xhtml` to `itemTreeMenuBar`, so it is always be available in all windows that have an itemTree, similar to View > Columns and View > Sort By?

abaevbog:
Rebased!

https://github.com/user-attachments/assets/af8a37a5-aa12-43f5-b60e-db532417020d

PR #5317, opened 2025-05-27
Tags autocomplete: fix stale results appearing in large libraries, add pref for fuzzy search

### LINKED ISSUE #1647: Search all words in tags for tag autocomplete
This has been requested fairly regularly. For performance reasons, we'll probably still want to limit this to individual words, maybe by using SQLite FTS.

https://forums.zotero.org/discussion/76084/can-tag-suggestions-prompts-search-whole-field and others

abaevbog:
- do not add autocomplete results for queries that are no longer relevant. For example, if the user
types "test", and row handler is still called with results for "te" - ignore them. Otherwise, results with "te" but not "test" may appear in libraries with many tags
- add optional fuzzy search for tags with a visible pref checkbox

Fixes: #1647 

---

To look into how fuzzy tags autocomplete would perform now, I created a collection with 10,000 items, each with 100 random tags (example: `8-6QqFw-ULQ44-TG9KD`). I assume a library with 1,000,000 tags total should be close to the upper bound of the number of tags one would have because then I also started running into other performance issues, like a longer delay when switching between collections. For reference, this is the [exported RDF file of that collection](https://www.dropbox.com/scl/fi/6ddrfav0lunglsvwxukxc/million_tags_collection.rdf?rlkey=db7tryqbc330xanftv4lvvpff&st=r6358o5c&dl=0).

With so many tags, on current `main`, the autocomplete runs fast but it also includes stale search results that are irrelevant for the current query:

https://github.com/user-attachments/assets/3e9c14af-b6d5-4343-8f19-042d5c745075

When we only include results fetched for the latest input (as in this PR), I get search results after ~2 seconds if fuzzy matching is disabled:

https://github.com/user-attachments/assets/f38f3c29-5e74-4350-a5fe-5f6fc2882cd8

With fuzzy search results, it takes a bit longer but not by that much - 3 seconds or so:

https://github.com/user-attachments/assets/e0e3d528-b78e-4dd3-99da-ffb7c52d360d

We could consider more optimization with SQLite FTS as described in the linked issue - let me know if that would be worth it!

### DIFF
--- a/chrome/content/zotero/zotero-autocomplete.mjs
+++ b/chrome/content/zotero/zotero-autocomplete.mjs
@@ -32,25 +32,86 @@ const Cr = Components.results;
 
 import { Zotero } from "chrome://zotero/content/zotero.mjs";
 
+const MAX_RESULTS = 50;
+
+let searchResultsCache = {};
+let currentSearchConfig = null;
+let currentQuery = null;
+let nextQuery = null;
+let lastQueryTs = 0;
+
 /*
  * Implements nsIAutoCompleteSearch
  */
 export function ZoteroAutoComplete() {}
 
+// Entrypoint to autocomplete search called on each input change
 ZoteroAutoComplete.prototype.startSearch = async function (searchString, searchParams, previousResult, listener) {
-	// FIXME
-	//this.stopSearch();
-	
+	// stopSearch is not always called before startSearch. We could
+	// ensure that any ongoing queries are cancelled here. But instead,
+	// we let the old query finish, in hope that some query results will
+	// match the new search string, since they will be added faster.
+	// this.stopSearch();
+
 	var result = Cc["@mozilla.org/autocomplete/simple-result;1"]
 					.createInstance(Ci.nsIAutoCompleteSimpleResult);
 	result.setSearchString(searchString);
-	
+
 	this._result = result;
-	this._results = [];
+	this._includedSearchResults = new Set();
 	this._listener = listener;
-	this._cancelled = false;
-	
-	Zotero.debug("Starting autocomplete search with data '"
+
+	let queryID = Zotero.Utilities.randomString();
+
+	Zotero.debug("Start autocomplete query " + queryID + " with data '"
+		+ searchParams + "'" + " and string '" + searchString + "'");
+
+	// startSearch is called on each typed character. We don't want to start running
+	// a new query unless the old one finished to avoid spamming many queries running
+	// at the same time, which poorly affects performance.
+	// We maintain a record of the current query and the next query that should run
+	// after the first one finished.
+	// Suppose we get "t", "te", "tes", and "test" searches, and queries take a somewhat long time.
+	// "t" query will be started, "te" and "tes" will be  skipped, and we'll only run "test"
+	// after "t" finishes.
+	nextQuery = { queryID, searchString, searchParams, cancelled: false };
+	if (!currentQuery) {
+		this._processQueries();
+	}
+};
+
+// Fetch search results one query at a time
+ZoteroAutoComplete.prototype._processQueries = async function () {
+	while (nextQuery) {
+		// Move next query to current
+		currentQuery = nextQuery;
+		nextQuery = null;
+		let currentTime = (new Date()).getTime();
+		
+		let { queryID, searchString, searchParams } = currentQuery;
+
+		// If the search config has not changed since last search, see if any
+		// earlier search results match this query
+		let searchConfig = JSON.stringify(searchParams);
+		let isExpired = currentTime - lastQueryTs > 10 * 1000;
+		if (currentSearchConfig === searchConfig && !isExpired) {
+			this._applyCachedResults(searchString);
+		}
+		// If the search config has changed, clear the cache
+		else {
+			searchResultsCache = {};
+			currentSearchConfig = searchConfig;
+		}
+		lastQueryTs = currentTime;
+
+		await this.executeQuery(queryID, searchString, searchParams);
+
+		currentQuery = null;
+	}
+};
+
+ZoteroAutoComplete.prototype.executeQuery = async function (queryID, searchString, searchParams) {
+	Zotero.debug("Execute autocomplete query with data '"
 		+ searchParams + "'" + " and string '" + searchString + "'");
 	
 	searchParams = JSON.parse(searchParams);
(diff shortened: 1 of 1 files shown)


chrome/content/zotero/zotero-autocomplete.js
```diff
@@ -246,12 +247,19 @@ ZoteroAutoComplete.prototype.startSearch = Zotero.Promise.coroutine(function* (s
...
 				Zotero.debug("Cancelling query");
 				cancel();
 				return;
 			}
+			// Ignore results from queries that are no longer relevant
+			// (e.g. if user types "test", ignore results that came from when they just typed "te")
+			if (currentSearchString != this._result.searchString) {
+				cancel();
+				return;
+			}
```
dstillman, line 262: This is what `stopSearch()` — currently unused with a couple `FIXME`s — is supposed to be doing, so we should clean that up at the same time if this logic is just better. And to clarify, this isn't just "ignoring" results from other queries — it's actually stopping the SQLite query (but see my other note about this).

defaults/preferences/zotero.js
```diff
@@ -28,6 +28,7 @@ pref("extensions.zotero.downloadAssociatedFiles",true);
 pref("extensions.zotero.findPDFs.resolvers", '[]');
 pref("extensions.zotero.reportTranslationFailure",true);
 pref("extensions.zotero.automaticTags",true);
+pref("extensions.zotero.tagsFuzzyAutocomplete", false);
```
dstillman, line 31: Let's just optimize this as best as we can. No need to pref it (unless we start getting reports that it's causing problems).

chrome/locale/en-US/zotero/preferences.ftl
```diff
@@ -59,6 +59,9 @@ preferences-locate-library-lookup-intro = Library Lookup can find a resource onl
 preferences-locate-resolver = Resolver:
 preferences-locate-base-url = Base URL:
 
+preferences-tags-fuzzy-autocomplete =
+    .label = Fuzzy tags autocomplete
```
dstillman, line 63: (This wouldn't be "fuzzy" matching. We could actually do that, but that would imply finding strings that were similar but not exact. This is just a non-left-bound search.)

dstillman (commented):
One optimization we could do here would be to store matching tags in memory after the first query and then simply filter that list in JS, as long as the search string contains the original string (and has the same libraryID/itemID params). We'd then store the next list, and so on. We'd set a short timer to clean up old lists.

The one catch is that (despite my other comment) we wouldn't actually want to cancel the initial query anymore — we'd want to let it run and avoid running additional queries for supersets of the running query string, so that we could filter the results as soon as they were available.

To the user, this would hopefully just mean that the autocomplete list might not appear after the first few characters but then would start appearing later as it was filtering successively smaller lists.

dstillman:
(And I don't think FTS would help us here. That was still about searching for individual words, not arbitrary substrings.)

abaevbog:
Briefly circling back here when other PRs are in works.

> The one catch is that (despite my other comment) we wouldn't actually want to cancel the initial query anymore — we'd want to let it run and avoid running additional queries for supersets of the running query string, so that we could filter the results as soon as they were available.

I tried the approach of letting the first query fully run, and then loading all those results into an in-memory cache, and then filtering that on all subsequent queries. If we try to get the superset of all subsequent matches in the first query, we have to remove `LIMIT` from the SQL. This, however, can take a very long time to run with small queries (e.g. a 1-2 character queries take forever and end up throwing errors). So, I think, we would still need a reasonable limit to set. Granted, I have here way more tags than one probably should (1,000,000) but still whatever solution we go with, it should probably handle that.

But then the first query (even if it runs fully) is not guaranteed to return matches for all subsequent sub-queries. For example, if we get 1000 results for "t", there may or may not be some matches for "test" among them. So we would still have to keep checking the database.

An option I explored was to not try to cancel the queries. Instead, `onRow` saved results into cache, and also added them into the autocompelte result, as long as the value matches the current search string. That gave me the best performance in terms of how fast the results appeared. But the issue there is that a new query may start without waiting for the existing query to finish, which leads to potentially several queries still running even after the autocomplete popup is gone. That is the case now, and it can lead to unfortunate behaviors as in #5403.

I am now looking into combining the two approaches. If `stopSearch` is called, the id of the last started query is added to the set of cancelled queries, which is checked in `onRow` handler. That way, we are at least somewhat cleaning up old queries. At the same time, as `onRow` handler runs, it will save results to the cache, so that if there is another query started, we can first include those in-memory cached results from the old query that match. I think it's worth exploring a small debounce as well.

abaevbog:
I think something like this would work here! These are the optimizations I added:

- there is an in-memory cache that gets cleared out after a while. Before running a new query, we check that cache to see if there is anything matching the new query. It is stored as an object `{"searchString": ["matchOne", "matchTwo"]}`. So if we are searching for `"test"`, we'll check cached results for `"tes"`, `"te"`, and `"t"` if they exist. Those will be found fast and immediately added into the popup
- we still have to run the SQL query after. As described above, I attempted to load the entire superset of a shorter query into cache (by removing the `LIMIT` from SQL) - but that can take too long for small queries. Without `LIMIT`, queries for one or two characters take forever. Because of this, cache may not always have all the matches but it serves mainly to provide a faster visual feedback.
- after the SQL runs, any additional results are added into the popup and stored in cache as well for subsequent queries

One issue I noticed is that we ended up spawning multiple SQL queries simultaneously, since `stopSearch` didn't really work (e.g. "t", "te", "tes", "test" would all start a new query, regardless if the old one finished). It ended up affecting performance after typing a few characters, and it also meant that the autocomplete searches kept running for a while even after the autocomplete popup was closed.

To tackle that, I added proper `stopSearch` handling and separated `startSearch` (called when user types) from `executeQuery` that runs the actual search to ensure that we only run one query at a time. If there are multiple queries that are send by the user while a query is already loading, the last query will be executed after the first one is finished (via cancellation or otherwise), and all other ones will be discarded.

This is an example of events now:
- user types "t" (`startSearch` called)
- SQL is being executed for searchString "t"
- user types "e"
- `stopSearch` may be called (but not always?), which will mark "t" query as cancelled. `startSearch` with "te" is called.
- user types "s" (`stopSearch` may be called, `startSearch` called)
- user types "t" (`stopSearch` may be called, `startSearch` called)
- SQL for "e" is finished via `cancel` in `onRow`
-  if there were any search results for "t" added into the cache that match "test", they will appear in the popup
- SQL is being executed for searchString "test" ("te", and "tes" are ignored)
- SQL finishes for "test", popup updated

This is what the search looks like on my end with 1,000,000 tags library:

https://github.com/user-attachments/assets/2aa88f79-f624-4d38-87f1-ba2f5e1bbb29

PR #5319, opened 2025-05-28
Improve tabs menu keyboard handle

### LINKED ISSUE #5283: Tabs menu: Space bar should activate close button
After #4961, space bar no longer activates the close button when it has focus (presumably because it's reserved for the input field, but if the close button has focus, that should take precedence).

### LINKED ISSUE #5284: Tabs menu: Focus next close button after closing tab?
It would be nice if it were easier to quickly close multiple tabs via the keyboard. If the close button has focus and you close the tab, maybe it should focus the close button of the next row instead of returning focus to the input field?

windingwind:
Use space to close tab if the close button is active
Focus next close button after closing tab with keyboard
fix #5283
fix #5284

### DIFF
--- a/chrome/content/zotero/elements/tabsMenuPanel.js
+++ b/chrome/content/zotero/elements/tabsMenuPanel.js
@@ -57,7 +57,7 @@
 
 		_filterInput;
 
-		_prevFocusCls = null;
+		_prevFocusClass = null;
 
 		get visible() {
 			return ["showing", "open"].includes(this.state);
@@ -94,7 +94,7 @@
 		/**
 		 * Create the list of opened tabs in tabs menu.
 		 * @param {Object} [options] - Options for refreshing the list
-		 * @param {boolean} [options.ignorePrevSelection=false] - If true, ignore the previous selection and select the currently selected tab
+		 * @param {boolean} [options.ignorePreviousSelection=false] - If true, ignore the previous selection and select the currently selected tab
 		 * @returns {void}
 		 */
 		refreshList(options = {}) {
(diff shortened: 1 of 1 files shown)


abaevbog:
If I follow these steps:
- select one of reader tabs
- open the tabs menu
- press tab to focus the current tab's `X` icon
- press Space to close the current tab

Then the tab does not get closed and there is an error thrown `Uncaught TypeError: panel.focusEntry is not a function`. This does not happen on Enter keypress, in which case the focus just goes back to the input.

windingwind:
Thanks for the detailed steps! It should be fixed now.

abaevbog:
Great! It works right on my end now.

One last thing I just noticed is that if you close the last tab via keyboard, the focus will end up leaving the tabs menu and going back to the library without closing the tabs panel. I assume if we mean to refocus the library tab because there is nothing else left in the menu, we should close the panel as well. But maybe it's best to just return focus into the input field? It may be better from accessibility perspective because you don't necessarily expect focus to leave the tabs menu when you press `X`.

For reference, this is what I mean:

https://github.com/user-attachments/assets/919b083d-178f-41c6-82b5-7fbd3ac4ccba

windingwind:
Thanks! The focus handling is a bit complex with the tabs. Should be fixed now.

dstillman:
This doesn't seem right to me. After I press space bar on a close button, focus is just jumping back to the row for the selected tab in the tab bar. This isn't a problem if I'm closing a tab that I'm actually viewing, because the next logical tab gets selected in the tab bar anyway, but if I've moved focus in the menu to a different tab from the one I'm viewing, focus jumps back rather than staying on the same menu row and letting me quickly go down the list from where I started.

windingwind:
Thanks, makes sense. Should be fixed now.

PR #5324, opened 2025-06-04
Annotation filters in tag selector

### LINKED ISSUE #5316: Add filtering of annotation rows by color
https://forums.zotero.org/discussion/comment/492839/#Comment_492839

And maybe author as well?

The main question is where this functionality can go in terms of layout. To filter by the author, maybe we can just include it as one of the conditions for quickSearch lookup. Not sure about the colors though. They act somewhat similar to tags, so maybe we move the color filter into the tag selector?

Also mentioned in https://github.com/zotero/zotero/issues/5229#issuecomment-2835721236

abaevbog:
- added new component annotationFiltersSelector.jsx to display annotation colors followed by a list of annotation authors
- added a splitter between the annotations filters and the actual tag selector, so that one can adjust the heights of these two subsections
- search box of the tag selector will also filter the annotation authors
- added new search conditions to accommodate searching by annotation color and author
- the logic of which tag colors/authors are visible is somewhat different from tags, and a bit more convoluted. Filtering by many annotation colors or authors is done with OR operator, not AND as with tags. It means that if one annotation author is selected, other annotation authors should not be removed from the list of options, so that one can select more than one author.
The current search logic can be described as below: (tag1 AND tag2) AND (color1 OR color2) AND (author1 OR author2)

Current behavior:

https://github.com/user-attachments/assets/3cadad02-436d-417b-93be-2fca036454e3

Still in works - need to add keyboard nav with a11y properties, tests, and polish out the UI.

Fixes: #5316

### DIFF
--- a/chrome/content/zotero/components/tagSelector.jsx
+++ b/chrome/content/zotero/components/tagSelector.jsx
@@ -28,24 +28,171 @@
 const React = require('react');
 const PropTypes = require('prop-types');
 const TagList = require('./tagSelector/tagSelectorList');
+const AnnotationFiltersSelector = require('./tagSelector/annotationFiltersSelector');
 const { Button } = require('./button');
 const { CSSIcon } = require('./icons');
 const Search = require('./search');
 
+const annotationColorsHeight = 20;
+const minTagsListHeight = 150;
+const annotationsVerticalPadding = 14;
+const annotationsAuthorsFirstRow = 22 + 6; // 22px for first row, 6px for margin between colors and authors
+const minAnnotationSectionHeight = annotationColorsHeight + annotationsAuthorsFirstRow;
+const splitterHeight = 9;
+
 class TagSelector extends React.PureComponent {
+	constructor(props) {
+		super(props);
+		this.state = {
+			annotationSectionHeight: minAnnotationSectionHeight,
+			isDragging: false,
+			dragStartY: 0,
+			dragStartHeight: 0
+		};
+	}
+
+	// Resize the annotation section height when the annotation authors are added or filtered out
+	componentDidUpdate() {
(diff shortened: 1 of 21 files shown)


abaevbog:
Some odd test failures happening at the moment. Last test run failed at `should update tag when pressing Enter in textbox` - but I assume it's an intermittent failure that I ran into in the past a few times.

Before that, `should return detection result passed to Zotero.done() synchronously` would not pass. If i run `test/runtests.sh -f translateTest.js` locally I get the same error. But I also get it on `main`, so that's odd

abaevbog:
Ok, seems like it was because one of the submodules was old

abaevbog:
https://forums.zotero.org/discussion/124680/add-or-export-annotations-conditional-on-colour

abaevbog:
I believe all the smaller glitches should be ironed out now here.

There is still the question if it's worth rendering annotation authors inside of a `<Collection>`, same as with tags. I wasn't sure if it's worth it, so to not over-optimize they are just rendered as regular nodes as the moment.

abaevbog:
I stumbled upon a feature request to be able to see a list of unique authors and filter items list by them: https://forums.zotero.org/discussion/86874/feature-request-author-view-list. This sounds very similar to what we are already doing here with annotation authors. I imagine another conditionally visible section of the tag selector container that would display unique authors in the selected collection would be appropriate for that. Is this something we are considering for the future? 

If so, I may want to pull some of the functionality here into separate components to make it easier to share between annotations and that future authors section (e.g. splitter resizing, hide/display logic, etc.). We also could consider renaming tag selector for something more general in that case.

But if it's not something we want to worry about now, none of the above is necessary.

PR #5333, opened 2025-06-10
Do not exclude image annotations when creating a standalone note from multiple items

abaevbog:
This was a regression from 48fd23ccec73156855f4c59170b332af2e4357dd. Image annotations are not excluded in other cases, like if a child note is created or if a standalone note is created from selected image annotation rows (not attachments).


mrtcode:
Thanks!

PR #5339, opened 2025-06-11
Do not open hidden Zotero window on connector save

abaevbog:
And fix regression where selecting a collection in connector would always switch to the library tab, even if a reader tab is selected.

It was originally added in https://github.com/zotero/zotero/pull/5020 but was lost after https://github.com/zotero/zotero/pull/5148


adomasven:
I don't get it. Did some change in our code started triggering this, because I cannot find it? Or update to underlying Firefox version?

abaevbog:
I believe it's all about this commit from a while ago: https://github.com/zotero/zotero/commit/45e681431bfd6b77bfaf8f55ce87f04fedb4a475. It is not present in `7.0.15`, so reopening a hidden window just doesn't work on the stable release. I went back to a commit before that one on `main` to verify, and the window would not get reopened there as well.

PR #5341, opened 2025-06-12
Use annotation box for single annot selection

### LINKED ISSUE #5241: Sidenav visibility inconsistent with annotation in item pane

windingwind:
Allow plugins to register section in annotation item pane context
resolve: #5241

### DIFF
--- a/chrome/content/zotero/customElements.js
+++ b/chrome/content/zotero/customElements.js
@@ -68,6 +68,7 @@ Services.scriptloader.loadSubScript('chrome://zotero/content/elements/itemTreeMe
 		['attachments-box', 'chrome://zotero/content/elements/attachmentsBox.js'],
 		['attachment-row', 'chrome://zotero/content/elements/attachmentRow.js'],
 		['attachment-annotations-box', 'chrome://zotero/content/elements/attachmentAnnotationsBox.js'],
+		['annotation-box', 'chrome://zotero/content/elements/annotationBox.js'],
 		['annotation-row', 'chrome://zotero/content/elements/annotationRow.js'],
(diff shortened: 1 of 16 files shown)


windingwind:
Single annotation (currently the tags box is enabled to illustrate the look with custom sections):

image


Multi annotations:

image

If this proof-of-concept works, I will let the item details element disable section toggling and the sidenav when only one section is enabled, so that if there are no custom sections, it looks exactly the same as the previous implementation of `AnnotationItemsPane`.

abaevbog:
Yeah, I think this works! I was concerned that the switch between `itemDetails` and `annotationItemsPane` would be noticeable but it looks totally fine.

I only have a few minor comments for now. I'll keep playing around with this to see if there's anything else I notice.

- we probably want the header of the collapsible section on `annotationBox` to have the name of the top level item, same as in `annotationItemsPane`. Just so that they look the same
- for the same reason, maybe the annotations collapsible section should also not be collapsible? Even if there are custom sections, I think the annotation should be always visible. Otherwise, you can collapse it, and then if the second annotation is selected, it looks as if the section is forced to expand.
- when one annotation is selected, we only show the button for annotations `collapsible-section` (plus any additional sections from plugins and such). When multiple annotations are selected, the sidenav switches to have all default sidenav buttons that are all disabled. Maybe we should only show the same annotations sidenav button when many annotation rows are selected too, just for consistency?

Issue #5329, opened 2025-06-05
Tabs menu: Focus previous/next close button when using up/down arrow?


dstillman:
After #5319, I feel like if you have a close button focused and you use up/down arrow, it should focus the close button for the previous/next row rather than focusing the row itself.

PR #5346, opened 2025-06-19
Keep tabs menu focus when using up/down

windingwind:
fix: #5329


windingwind:
Currently, when going from other rows -> library row -> other rows, the focus will change from close button -> input -> close button, because the library row does not have the close button. A bit strange, but it feels better than losing the focus target (close button -> input -> input).

PR #5347, opened 2025-06-19
qf: fix breakage with unloaded tab

abaevbog:
Properly load all item data for items opened in tabs to account for the fact that they can belong to
libraries that are not yet loaded.

Similar fix as in https://github.com/zotero/zotero/commit/4421de63de709c238a5418cb9747aeb2085fcd93

### DIFF
--- a/chrome/content/zotero/integration/quickFormat.js
+++ b/chrome/content/zotero/integration/quickFormat.js
@@ -656,13 +656,22 @@ var Zotero_QuickFormat = new function () {
 		}).map(t => t.data.itemID);
 		if (!itemIDs.length) return [];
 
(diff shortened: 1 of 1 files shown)


abaevbog:
The simplest way to reproduce the issue this is aimed to solve is to:
- begin with no tabs opened
- open an item in a group (not My Library) as a reader tab
- select My Library tab
- close and reopen Zotero
- that reader tab is now unloaded and the group the item of that tab belongs to is not loaded either
- open QuickFormat dialog (e.g. in word). Without this change, there is an error thrown that the item `_getMatchingReaderOpenItems` tried to access is not yet loaded.

PR #5348, opened 2025-06-20
Fix reader sidenav buttons wrongly disabled

windingwind:
fix: #5332


windingwind:
To reproduce: pin a section in the reader - switch to library - switch back.

Issue #5350, opened 2025-06-23
Citation dialog: Citation popup overflows for non-English locale


dstillman:
https://forums.zotero.org/discussion/comment/494148/#Comment_494148

![Image](https://github.com/user-attachments/assets/d430f36f-5fd8-4bc2-9edc-c9165e42a9bd)

PR #5354, opened 2025-06-23
Citation dialog: fix broken item details popup with some non-english locales

abaevbog:
Make sure that `<select>` element does not stretch and cutoff the popup for locales that have longer locator labels than English.

Fixes: #5350

Screenshot 2025-06-23 at 9 34 11 AM

### DIFF
--- a/scss/components/_citationDialog.scss
+++ b/scss/components/_citationDialog.scss
@@ -604,7 +604,7 @@
 					margin: 16px 0;
 					display: grid;
 					grid-template-columns: auto 1fr;
-					gap: 8px 2px;
+					gap: 8px 1px;
 					align-items: center;
 					.details-label:not(select) {
 						text-align: end;
@@ -625,7 +625,7 @@
 					}
 					#label {
 						// align <select>'s end with the end of prefix/suffix labels better
-						margin-inline-end: -3px;
+						margin-inline-end: -1px;
 						@media (-moz-platform: macos) {
 							height: 22px;
 							// <select> with increased height is hard to position but this helps
--- a/scss/components/_citationDialog.scss
+++ b/scss/components/_citationDialog.scss
@@ -606,8 +606,8 @@
 					grid-template-columns: auto 1fr;
 					gap: 8px 1px;
 					align-items: center;
-					.details-label {
-						text-align: right;
+					.details-label:not(select) {
+						text-align: end;
 					}
 					.details-data {
 						@media not (-moz-platform: windows) {
--- a/scss/components/_citationDialog.scss
+++ b/scss/components/_citationDialog.scss
@@ -591,8 +591,8 @@
 			}
 		}
 		#itemDetails {
-			width: 400px;
 			.popup {
+				width: 364px; // with padding and border, the panel is 400px
 				.details-header {
 					.icon {
 						margin-inline-end: 5px;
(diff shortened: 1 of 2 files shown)


scss/components/_citationDialog.scss
```diff
@@ -632,7 +632,9 @@
 						padding-inline-start: 100px;
 					}
 					#label {
-						flex: 1;
+						// prevent left-edge of <select> from expanding past the "Remove" button
+						// for locales that have long labels
+						max-width: 138px;
```
dstillman, line 637: What is 138px?

Did you check whether this is enough space for all locales not to be truncated?

abaevbog:
I ended up completely revising this to use the grid in the item details popup, and I think this gives us the most flexible layout with all the elements still remaining aligned. Now, the inputs will occupy all available space that is left by the `<select>` component, while still fitting into the 400px of the popup.

In English locale, it looks almost identical (I added a tweak to right-align the text in `<select>`, so it looks similar to Prefix/Suffix labels).

Before:
before_eng

After:
after_eng

And then if the locator dropdown needs more space, it will just grow and shrink all inputs by the same amount. So `<select>` can fully fit and the inputs remain aligned. For example, this is in Portugese that has the longest locator:

after_port


How does this look?

dstillman:
The drop-down text should remain left-aligned. I think right-aligning that would be pretty non-standard and weird. More important to follow the convention for the component itself than to align with the surrounding text.

abaevbog:
Gotcha. I pushed a tweak to not change the text alignment in the `<select>`. That idea came from the long dropdown in locales like Portugese. `<select>` there is stretched a lot due to one long label, though most remaining labels are fairly short. In that case, it seemed to me, that pushing the text to the right makes it look a bit nicer.

But, it's more of an edge case, so it's probably best not to change text alignment just for that.

Screenshot 2025-06-24 at 10 07 23 PM
Screenshot 2025-06-24 at 10 04 37 PM

dstillman:
Yeah, kind of awkward there, but I think that's just the way it has to be.

Looks like the space between the label and input is now slightly narrower? Can we add 1px or so back?

abaevbog:
Yeah, sorry I was trying to measure out what the distance was like before vs what it was now. It looks like on current beta, the gap between the input and label is very-very small. Looks like just 1px. With 2px now, the distance is slightly larger than it was. But this feels better now to me, so I think it's an improvement.

For comparison, this is a very zoomed in Prefix label and input in beta: 
Screenshot 2025-06-24 at 11 06 08 PM
And this is the Prefix label and input that i see now:
Screenshot 2025-06-24 at 11 06 47 PM

The gap is slightly bigger but I think it looks better that way.

dstillman:
Oh, no, it's just the `margin-inline-end: -3px` on the drop-down. At least on macOS, that's causing the `select` to push against the focus ring in the first input.

dstillman:
This is reverting the gap to 1px and changing the select's margin-inline-end to -1px:

Screenshot 2025-06-25 at 2 12 00 AM

abaevbog:
Gotcha, I see what you mean.
On beta:
Screenshot 2025-06-24 at 11 12 10 PM
And here, before the last commit:
Screenshot 2025-06-24 at 11 12 44 PM

I was trying to align the end of select with the labels a bit better. Which doesn't really pay off when the focus ring appears...

dstillman:
Yeah, what you were trying to do makes total sense, but since the first input gets focused by default, I think making that not look bad is more important.

dstillman:
We can tweak further, but merging this now to get the locale fix in the beta. Much better!

abaevbog:
Sounds good!

PR #5357, opened 2025-06-25
Fix all cookies being lost during remote translation

AbeJellinek:
We need to use system XHR in order to allow translators to make cross-origin requests, but by default, system XHR objects have no information identifying the source browser. If CookieSandbox is unable to determine the source of the request, it goes into "paranoid" mode and clears the `Cookie` header. This led us to lose cookies on all requests made by web translators in the client, causing issues like https://forums.zotero.org/discussion/124692/issue-with-saving-items-from-wires-climate-change.

We now override the channel's `nsILoadInfo` with a document-specific one, so CookieSandbox can see where the request is coming from.

Separately, we now add the CookieSandbox observer at Zotero startup, not just after the first CookieSandbox is created. Because CookieSandbox touches *all* requests, not just ones from tracked sources, the previous behavior didn't really make sense - initially, all requests would keep their cookies, then some code somewhere would initialize a CookieSandbox, and all future requests everywhere would have their cookies managed until Zotero was restarted.

Fixes zotero/translate#41

### DIFF
--- a/chrome/content/zotero/xpcom/cookieSandbox.js
+++ b/chrome/content/zotero/xpcom/cookieSandbox.js
@@ -60,6 +60,10 @@ Zotero.CookieSandbox = function (browser, uri, cookieData, userAgent) {
 	}
 };
 
+Zotero.CookieSandbox.init = function () {
+	Zotero.CookieSandbox.Observer.register();
+};
+
 /**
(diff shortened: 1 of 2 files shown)


AbeJellinek:
Merged

PR #5358, opened 2025-06-25
Fix all cookies being lost during remote translation (for 7.0)

AbeJellinek:
https://github.com/zotero/zotero/pull/5357


AbeJellinek:
Merged

PR #5359, opened 2025-06-26
Display the number of items that would be deleted when replacing the online library

### LINKED ISSUE #5078: Show clearer warning for Replace Online Library?
https://forums.zotero.org/discussion/121383/library-accidentally-overwritten-after-migrating-to-a-new-pc
https://forums.zotero.org/discussion/122361/i-replaced-my-online-library-with-nonexistant-local-files-on-my-new-computer

tnajdek:
This compares item keys in the remote library with those in the local DB and prints an additional warning if at least one remote key is not found locally. In a fully synced library, there is no change in behaviour, but if I remove `zotero.sqlite`, I’ll now see:

Screenshot 2025-06-26 at 14 24 46

Few notes:

* I've added `asKeys` to `getAll` in `items.js` as an optimisation. If we don't want to affect the core API for this, I can retrieve all items and map them to keys instead.
* We probably need a spinner in the Preferences window to provide feedback while remote keys are being fetched.

Fix #5078

### DIFF
--- a/chrome/content/zotero/preferences/preferences_sync.jsx
+++ b/chrome/content/zotero/preferences/preferences_sync.jsx
@@ -833,6 +833,7 @@ Zotero_Preferences.Sync = {
 			
 			case 'restore-to-server': {
 				const CHECKBOX_THRESHOLD = 10;
+				const CONFIRMATION_TEXT_MAX_ITEMS = 5;
 				
 				let apiKey = await Zotero.Sync.Data.Local.getAPIKey();
 				let client = Zotero.Sync.Runner.getAPIClient({ apiKey });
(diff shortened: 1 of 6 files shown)


dstillman:
> I've added `asKeys` to `getAll` in `items.js` as an optimisation. If we don't want to affect the core API for this, I can retrieve all items and map them to keys instead.

Yeah, I would just use `asIDs` and map with `Zotero.Items.getLibraryAndKeyFromID(id)`.

chrome/locale/en-US/zotero/preferences.ftl
```diff
@@ -99,3 +99,12 @@ preferences-advanced-custom-data-dir =
...
     .value = (Default: { $directory })
     .aria-label = Default location
+
+preferences-sync-reset-restore-to-server-body = { -app-name } will replace data in “{ $libraryName }” on { $domain } with data from this computer. { $remoteItemsDeletedCount ->
+    [0] { "" }
+    *[other] { $remoteItemsDeletedCount } { $remoteItemsDeletedCount -> 
+        [1] item
+        *[other] items
+    } in the online library will be deleted!
+}
```
dstillman, line 109: Probably better to do this as a separate string that gets added after two newlines if >0.

And I think we can omit the exclamation mark. Arguably appropriate for thousands of items, but it might say "1 item", and in general it's probably better to be matter-of-fact about it.

chrome/content/zotero/preferences/preferences_sync.jsx
```diff
@@ -831,21 +831,24 @@ Zotero_Preferences.Sync = {
...
-					Zotero.getString('zotero.preferences.sync.reset.restoreToServer.button'),
-					null, null, null, {}
-				);
+			case 'restore-to-server': {
+				let apiKey = await Zotero.Sync.Data.Local.getAPIKey();
+				let client = Zotero.Sync.Runner.getAPIClient({ apiKey });
+				var keyInfo = await Zotero.Sync.Runner.checkAccess(client, { timeout: 5000 });
+				let { keys: remoteKeysArray } = await client.getKeys('user', keyInfo.userID, { target: 'items/top' });
+				let remoteKeys = new Set(remoteKeysArray);
+				let localKeys = new Set(await Zotero.Items.getAll(Zotero.Libraries.userLibraryID, true, false, false, true));
```
dstillman, line 840: This is using `onlyTopLevel`, so in theory you could be wiping out hundreds of attachments, notes, and/or annotations. Probably too confusing to count annotations as "items", but we should probably at least count attachments and notes. Can just omit `asIDs` (contra my other comment), filter to `isRegularItem() || isNote() || isFileAttachment()`, and map to`.key`.

We could also show annotations separately, but that's probably getting too complicated, since the main use case here is just preventing wiping out a whole library.

tnajdek:
> > I've added `asKeys` to `getAll` in `items.js` as an optimisation. If we don't want to affect the core API for this, I can retrieve all items and map them to keys instead.
> 
> Yeah, I would just use `asIDs` and map with `Zotero.Items.getLibraryAndKeyFromID(id)`.

I've reverted this change

dstillman:
OK, currently going from this:

Screenshot 2025-09-12 at 1 11 13 AM

To this:

Screenshot 2025-09-12 at 1 27 31 AM

Screenshot 2025-09-12 at 1 27 35 AM

(Unfortunately I don't think this can be a sheet anymore, only a dialog, but at least we're getting an exclamation icon.)

If fewer than 10 remotely deleted items, it doesn't show the checkbox.

But I'm working on changing it to show this if the local library has ~0 items, to account for the most common failure pattern here, and to use a checkbox when there are more local items.

Screenshot 2025-09-12 at 3 36 29 AM

Screenshot 2025-09-12 at 3 36 36 AM

If only a few remote items will be deleted, it will just show this:

Screenshot 2025-09-12 at 3 35 37 AM

dstillman:
OK, something like that? Need to think more through the actual conditional logic we want.

tnajdek:
Looks good to me. I’ve tested a few scenarios, and just to reiterate:

1. If the local library is a superset of the remote library, the dialog behaves as it did before this PR.
2. If the local and remote libraries differ slightly, the number of items deleted in the remote library is shown, and the dialog can be confirmed immediately.
3. If the local and remote libraries differ significantly, the dialog requires checking a checkbox.
4. If the local library is empty or nearly empty while the remote contains significant number of items, the dialog requires entering text to confirm replacement.

This logic seems sound to me.

dstillman:
Great, thanks!

PR #5366, opened 2025-07-03
Expose tag split option in item tags box

### LINKED ISSUE #5323: Expose tag split option in item tags box
Same as https://github.com/zotero/zotero/issues/1545 but in the item tags box, so you can paste in comma-separated tags and quickly split them

tnajdek:
In the item box tag input, if the pasted string matches the intercept regex—currently set to two words separated by one of three delimiters (`,`, `;`, or `--`)—Zotero will now open a tag splitter UI, with the delimiter automatically set to the one that was detected. Pressing **"Save Tags"** will save the selected tags from the list; pressing **"Cancel"** will save the tag as-is, without splitting.

Close #5323 

https://github.com/user-attachments/assets/38863c4b-e708-4439-a15c-44f7eb6fb7f3

### DIFF
--- a/chrome/content/zotero/elements/tagsBox.js
+++ b/chrome/content/zotero/elements/tagsBox.js
@@ -368,6 +368,15 @@
 				});
 				event.preventDefault();
 			}
+			let delimiters = [',', ';', '--'];
+			let interceptRegex = new RegExp(`\\w+\\s*(${delimiters.join('|')})\\s*\\w+`, 'i');
+			let match = str.match(interceptRegex);
+			if (match) {
+				event.preventDefault();
+				setTimeout(() => {
+					this.openTagSplitterWindow(str, match[1], textbox);
+				});
+			}
 		};
 
 		makeMultiline(editable, value) {
@@ -381,6 +390,30 @@
 		makeSingleLine(editable) {
 			editable.noWrap = true;
 			editable.multiline = false;
+		}
+
+		openTagSplitterWindow(tagString, delimiter, textbox) {
+			const dataIn = {
+				oldTag: tagString,
+				isLongTag: false,
+				delimiter
+			};
+			const dataOut = { result: null };
+
+			window.openDialog(
+				'chrome://zotero/content/longTagFixer.xhtml',
+				'',
+				'chrome,modal,centerscreen',
+				dataIn, dataOut
+			);
+
+			if (!dataOut.result) {
+				textbox.value = tagString;
+			}
+
+			if (dataOut?.result?.op === 'split') {
+				textbox.value = dataOut.result.tags.join('\n');
+			}
 		}
 
 		saveTag = async (event) => {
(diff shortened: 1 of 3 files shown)


dstillman:
Cool! A few things:

1. If you erase the delimiter, it should show the full string on one line. Right now there's no way to tell it to use the full string other than to press Esc or enter some delimiter that doesn't appear in the string.
2. If you press Esc, currently it saves the string as is, but probably it should cancel the paste?
3. When the dialog is open, the new tag line is empty with just an icon. Not a big deal, but it looks a little wrong, and maybe also better to show the full string there for context? That would be another way to see the delimiter, if it guesses wrong.

dstillman:
(Also need to revert the reader submodule change)

tnajdek:
@dstillman Thanks for the feedback. Here are the changes:

* When the delimiter is empty or the split produces a single tag, the tag appears in the listbox and the accept button stays enabled
* If you press `Esc` or `Cancel`, the pasted tag is removed
* When the dialog is open, the new tag input contains the whole tag as pasted
* I've introduced a new button to the dialog with the label `Don’t Split` to allow quickly dismissing the dialog
* Removed submodule change - definitely shouldn't have been part of this PR
* Rebased

https://github.com/user-attachments/assets/03aaf769-7596-4688-b22d-477a60383777

chrome/content/zotero/elements/tagsBox.js
```diff
@@ -368,6 +368,14 @@
 				});
 				event.preventDefault();
 			}
+			let delimiters = [',', ';', '--'];
```
dstillman, line 371: What was your thinking behind `--`? If we want to actually include em dashes, we should include an actual em dash too, but I wonder if we actually want that. I do feel like em dashes are sometimes used, but I'm not sure I have any examples, and the one instance I see in zotero/translators is in National Academies Press, where the em dashes actually appear *within* valid tags rather than between them (and for some reason we [convert them to two hyphens](https://github.com/zotero/translators/blob/4d2bfbd39d39377c6b03e43b16b99a61e5c5229b/National%20Academies%20Press.js#L109), which seems wrong — maybe @AbeJellinek remembers why).

tnajdek: I included `--` because of this thread: https://forums.zotero.org/discussion/73146/how-to-split-a-tag, but I didn't do any research on how commonly it is used as a separator beyond this one case.

dstillman: > Most library catalogs use -- to separate levels in standard taxonomies

Right, but would those be separate tags? It seems like we should be consistent here. To use one of the [NAP examples](
https://nap.nationalacademies.org/catalog/26186/the-use-of-limited-access-privilege-programs-in-mixed-use-fisheries
):

Screenshot 2025-07-30 at 10 02 04 AM

Should that be three tags or six? We're currently importing three in the translator. If we're going to split on `--`/`—` here, we should probably split in a translator like this as well.

In any case, we should include `—` if we're including `--`.

(And yeah, I don't think we should replace good typography with bad typography.)

PR #5371, opened 2025-07-07
Fix trashed collection sometimes treated as an item row

abaevbog:
If the id of the collection happens to be the same as an id of an existing item, those objects may end up being used interchangeably in `itemTree`. The trashed collection could appear in "My Library" instead of the item, or the not-deleted item could appear in the trash instead of the collection. This is addressed by using `.treeViewID` instead of the `.id` of the objects in `ItemTree.refresh` and `ZoteroPane.itemSelected`.

@AbeJellinek, does this fix the issue of weird trashed collections behavior for you?


AbeJellinek:
It does!

PR #5378, opened 2025-07-09
Unify styles for headers of similar custom elements

### LINKED ISSUE #5372: Trashed item / collection buttons have slightly different sizes
When switching between an item and a collection in the trash:

https://github.com/user-attachments/assets/41fe017c-71d9-44ab-a6f1-819cc82cc5bf

tnajdek:
Unify very similar styles used by four different custom elements and move the shared code into a mixin.

The most noticeable improvements are:

* Buttons no longer "jump around" (see video in #5372)
* The background colour remains consistent — previously, some elements used the toolbar background, while others used the sidepane background.

@windingwind I've removed some of your styles to fix buttons "jumping around" (`button` and `.button-text` in `scss/elements/_itemPaneHeader.scss`). If you believe this breaks something, please let me know.

Before:

https://github.com/user-attachments/assets/c3f1c7f5-c6b6-4855-8d6d-00fa2a421af1

After:

https://github.com/user-attachments/assets/498a4a59-774f-47fc-80a5-d3d833223050

I've also tested on Windows, where this issue is also present (though perhaps slightly less noticeable).

Close #5372

### DIFF
--- a/chrome/content/zotero/elements/itemPaneHeader.js
+++ b/chrome/content/zotero/elements/itemPaneHeader.js
@@ -199,7 +199,7 @@
 			if (this._item.isAttachment()) {
 				headerMode = 'title';
 			}
-			
+
 			this.title.hidden = true;
 			this.creatorYear.hidden = true;
 			this.bibEntry.hidden = true;
--- a/scss/abstracts/_mixins.scss
+++ b/scss/abstracts/_mixins.scss
@@ -406,3 +406,30 @@
 		visibility: visible;
 	}
 }
+
+@mixin elements-custom-head {
+	.custom-head {
+		display: flex;
+		flex-direction: row;
+		align-self: stretch;
+		gap: 8px;
+		padding: 6px 8px;
+		background: var(--material-sidepane);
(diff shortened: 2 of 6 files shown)


windingwind:
Thanks! I've been always wanted to fix that and this seems to be a more generalizable solution.

Unfortunately, I'm still seeing some jumping around, see the misaligned below:

image

It's these two images:


image

image

tnajdek:
@windingwind I can't seem to reproduce this - could it be that some styles didn’t rebuild when you were testing? Or maybe I’m looking at the wrong thing?

I’ve adjusted Zotero to look more like yours and used a note and a collection inside the trash to compare (I slowly adjust opacity between the two screenshots in this video):

https://github.com/user-attachments/assets/fac2df52-ab36-4230-b30f-0655840bed95

At 50% overlay:

Screenshot 2025-07-10 at 14 57 26

windingwind:
Thanks! I checked again and it indeed looks fine. It turned out that the terminal was using Node 16 and the building didn't run correctly. Sorry for the misinformation.

Looks good to me.

Issue #5391, opened 2025-07-17
Tab with empty title is missing in tab menu


AbeJellinek:
https://forums.zotero.org/discussion/125504/list-all-tabs-ignores-tabs-with-empty-tab-header

https://github.com/zotero/zotero/blob/51a62022e91fa43e7ab5fdee6ec9368a8964dfb4/chrome/content/zotero/elements/tabsMenuPanel.js#L116

windingwind:
Thanks!

Not sure why this is there. Might be inherited from the old implementation before the improvement...

PR #5392, opened 2025-07-17
Fix tabs menu filtering out empty title

windingwind:
Fix: #5391


windingwind:
By the way, should we give the empty tabs a default name? For example, `Empty item - $libraryID/itemKey`, otherwise it reads confusing if you have multiple tabs with empty title.

AbeJellinek:
I think including the library ID/item key in the tab title would be a bit too technical for users who don't know anything about the Zotero data model, but maybe we could fall back to the attachment title, and then to "Untitled" or something like that.

PR #5394, opened 2025-07-19
fix the  missing "yield" in line 158 of storageEngine.js

seasideccm:
Fix the issue where Zotero.Sync.Runner.sync({background: true}) cannot obtain the itemIDs

截图 2025-07-19 17-12-08

截图 2025-07-19 17-10-21


dstillman:
Oops! Thank you!

(This wouldn't affect most users these days because of the built-in reader, but for people using an external reader or otherwise modifying local files on disk, it would prevent regular checking for changes on recently modified or recently opened files, and they would've only been checked every 3 hours with a full scan.)

seasideccm:
Thank you!

PR #5405, opened 2025-07-24
Cache CSL Engine for item pane header

### LINKED ISSUE #5404: Cache CSL.Engine for item pane header when using Bibliography Entry mode
Otherwise there's a delay on the first click on an item

We currently cache the engine within itemPaneHeader.js, so to do this ahead of time we have to store it somewhere else.

AbeJellinek:
Once styles are initialized. `ItemPaneHeader#init()` runs at startup because the item pane is always in the DOM, even if an item isn't being viewed, so this is sufficient to cache the Engine at startup.

Also extract prefs as constants.

Closes #5404

### DIFF
--- a/chrome/content/zotero/elements/itemPaneHeader.js
+++ b/chrome/content/zotero/elements/itemPaneHeader.js
@@ -26,6 +26,10 @@
 "use strict";
 
 {
+	const PREF_HEADER_MODE = 'itemPaneHeader';
+	const PREF_BIB_ENTRY_STYLE = 'itemPaneHeader.bibEntry.style';
+	const PREF_BIB_ENTRY_LOCALE = 'itemPaneHeader.bibEntry.locale';
+
 	let htmlDoc = document.implementation.createHTMLDocument();
 	
 	class ItemPaneHeader extends ItemPaneSectionElementBase {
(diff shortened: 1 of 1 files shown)


dstillman:
But this would mean caching per window if/once we allowed multiple windows, right?

AbeJellinek:
Not really, because we're also caching on the XPCOM Style object now, and that persists across windows. Caching in `itemPaneHeader.js` is now mostly redundant, but it does prevent the Engine state from being reset unnecessarily every time (which we don't need since disambiguation isn't an issue here).

dstillman:
Not sure I follow. Are we storing an engine both in the style and in the header? What causes the state to be reset?

AbeJellinek:
Sorry, ignore that, I don't think I came to the right conclusion or explained my reasoning very well! I'll write something clearer.

AbeJellinek:
As of #5399, every time you call `Style#getCiteProc()`, `style.js` checks whether the locale, format, and journal abbreviation settings match a previous call, and it returns a cached CSL Engine instance if they do:

https://github.com/zotero/zotero/blob/d4a5fbbefe6def44d116bcf016e1d8841c9abca0/chrome/content/zotero/xpcom/style.js#L714-L727

`itemPaneHeader.js` has been caching its own (per-window) CSL Engine instance since before we added the global cache.

`itemPaneHeader.js` also caches the citations generated by its CSL Engine instance.

This PR adds a call to `_getCSLEngine()` in `itemPaneHeader.js`'s initialization code. `_getCSLEngine()` calls `Style#getCiteProc()`, which causes `style.js` to populate its global cache with a CSL Engine instance matching the item pane header settings, and then it copies that instance to `itemPaneHeader.js`'s own cache.

We don't strictly need the `itemPaneHeader.js` cache now that we have a global cache. That said, the global cache is a little more risk-averse than we need - it fully resets the Engine's state every time it's returned. That shouldn't be necessary for `itemPaneHeader.js`, and we can skip it by caching locally. We also already need to store the cache keys in `itemPaneHeader.js` in order to cache generated citations. So there's no real downside to this double-caching situation, and it saves a bit of computation on each item pane render since it skips the full state reset that `style.js` would do.

As you said, the `itemPaneHeader.js` cache is per-window. But every time it's populated using the same style and locale settings, it's going to pull from the `style.js` cache, which is global.

AbeJellinek:
Does that clarify it at all? Sorry, still a bit of a complicated explanation.

dstillman:
OK, so it's not actually storing a separate copy. It's just avoiding the state reset that style.js does by getting the local copy.

AbeJellinek:
Yeah, exactly.

PR #5407, opened 2025-07-24
If custom file handler is set to Zotero itself, use reader

### LINKED ISSUE #5406: Prevent setting custom PDF handler to Zotero
https://forums.zotero.org/discussion/125660/unable-to-open-pdf-in-zotero-mackbook-pro

AbeJellinek:
Fixes #5406


dstillman:
Very unlikely, but I guess not impossible that someone could have a custom script that had 'zotero' in the name, or had 'zotero' in the path to their handler of choice? Maybe better to limit to `Zotero.*.app` on macOS, `zotero.exe` on Windows, and a suffix of 'zotero' on Linux?

AbeJellinek:
Sure, like that?

Issue #5408, opened 2025-07-25
New Collection dialog: Sub-collection menus don't expand when dialog was opened via right-click (Windows)


AbeJellinek:
https://forums.zotero.org/discussion/125635/cannot-expand-sub-collections-in-new-collection-from-a-right-click-in-the-collections-pane

I can't reproduce this on macOS.

abaevbog:
On macOS, even though the child collections expand when the parent is hovered, clicking away does seem not hide the collections menu. So once the collections menu is open, the only way to close it is via Escape. It may or may not be related.

AbeJellinek:
And/or related to #5204.

abaevbog:
Oh, I think it has the same root cause as #4891. I suspect there is some kind of race condition going on while trying to handle a dialog and a context menu that confuses mozilla. 

If I add a `Zotero.Promise.delay()` before opening the dialog to create a subcollection, it allows the context menu to fully close before opening the dialog and then it seems to work as expected. The menu will close when the dialog is clicked, and the children expand when parent collection is hovered.

https://github.com/zotero/zotero/blob/735922a2bfcbc8eff4df5a7b0586e3ce05857ad4/chrome/content/zotero/zoteroPane.js#L3226-L3229

@AbeJellinek, does this work on your end?

```
{
	id: "newSubcollection",
	oncommand: async () => {
		await Zotero.Promise.delay();
		this.newCollection(this.getSelectedCollection().key);
	}
}
```

If so, we could just add `Zotero.Promise.delay()` before every relevant instance of `window.openDialog` - but that seems a bit hacky and it's something easy to forget in the future. Maybe it would be cleaner to tweak the `window.openDialog` function so that it calls the original `window.openDialog` after a `setTimeout` if there is an opened `menupopup`?

AbeJellinek:
This seems to work for me, and avoids monkey-patching:

```js
Services.obs.addObserver({
	observe() {
		Services.ww.activeWindow
			?.document.querySelector('menupopup[open="true"]')
			?.hidePopup(/* cancel */ true);
	}
}, 'domwindowopened');
```

abaevbog:
Hmm, I'm afraid this observer doesn't seem to work for me... Am I correctly understanding that `domwindowopened` fires when the window is already opened? Because my idea for the delay came from the observation that the new collection dialog works just fine if opened via icon click but has issues when the context menu is still present. So the intention is to ensure that the menu closes _before_ trying to open the dialog.

[...]

Issue #4891, opened 2024-12-03
fx128: Popups no longer auto-hide in dialogs on macOS


AbeJellinek:
May be related to non-native menupopup (@windingwind)? I'm not sure.

1. Right-click an item -> Export Item...
2. Click the Format menulist
3. Click somewhere else, or Cmd-Tab

Popup should disappear, but it stays open until you press Escape. Other non-native popups work fine - this seems to only affect popups in dialog windows.

windingwind:
No idea why's that... Sorry I did some tries but forgot to reply. Should not be related to our custom element mixins.

mrtcode:
After testing the fx128 branch and then switching back to the main branch using the same profile, I found that it had set `ui.popup.disable_autohide` and `devtools.popup.disable_autohide` to `true`.

It also changed `extensions.zotero.openReaderInNewWindow` to `true`.

So this may be related. Also who knows what other properties might have been altered.

dstillman:
I don't think this is related to the `disable_autohide` prefs — those haven't been changed for me, and that setting would affect all popups.

abaevbog:
Are we sure this is specific to `fx128`? I am observing the same behavior on both `fx128` branch and in the latest beta where the first time a dialog is opened, its popup will auto-hide if you click elsewhere but on subsequent times, it won't.

This is what I see on both fx128 and the current beta:
- start Zotero
- Right-click on an item -> Export Item -> Format (popup appears)
- click on the body of the dialog
- it does go away
- click cancel, dialog closes
- Right-click on an item -> Export Item -> Format (popup appears)
- click on the body of the dialog
- popup does not go away

Another observation I made: 
- start Zotero
- focus some other window
- return focus to Zotero window
- Right-click on an item -> Export Item -> Format (popup appears)
- click on the body of the dialog
- even this first time, it does not go away

Maybe it has something to do with focus handling on the main window? Perhaps it's related to this other dialog focus-related issue: https://github.com/zotero/zotero/issues/4819?

abaevbog:
Oh, one more thing: `newCollectionDialog` can be opened by pressing the `toolbarbutton` above collectionTree or by pressing `+` in itemPane "Libraries and Collections" section and selecting "New Collection" in the popup. I noticed that in the first case, the popup with collections will auto-hide as expected (e.g. if I click on the dialog). But if I open `newCollectionDialog` through "New Collection" context menu, then menulist popup will not auto-hide after the first time as described above.

[...]

PR #5409, opened 2025-07-25
Fix glitchy menupopups in dialogs opened from a context menu

abaevbog:
Opening a dialog window via a command from a `menupopup` (e.g. context menu) leads to unexpected glitched in the dialog's popups. For example, collections menu in `newCollection` dialog on macOS won't go away when you click away, and on windows, child collections don't appear when a parent is hovered. This is due to some kind of race condition that arises when the dialog appears when the popup is still closing.
This workaround captures command events from `menupopup`s, closes the popup, stops propagation of those events, and then dispatches a copy of that event after a delay. This gives the popup enough time to close.

Fixes: #5408
Fixes: #4891

### DIFF
--- a/chrome/content/zotero/customElements.js
+++ b/chrome/content/zotero/customElements.js
@@ -109,6 +109,50 @@ Services.scriptloader.loadSubScript('chrome://zotero/content/elements/itemTreeMe
 		}
 	});
 
+	// Opening a dialog window via a command from a menupopup leads to
+	// unexpected glitched in the dialog's popups (e.g. collections menu
+	// in newCollection dialog on macOS won't go away when you click away,
+	// and on windows, child collections don't appear when a parent is hovered).
+	// This is due to some kind of race condition that arises when the dialog
+	// appears when the popup is still closing.
+	// This workaround captures command events from menupopups, closes the popup,
+	// stops propagation of those events, and then dispatches a copy of
+	// that event after a delay. This gives the popup enough time to close.
+	document.addEventListener("command", (event) => {
+		// Only handle command events from menupopups
+		if (!event.target.closest("menupopup")) return;
+
+		// If the source event of this 'command' event is another 'command' event,
+		// it must be a delayed event re-dispatched by the code below, so let it propagate
+		if (event.sourceEvent?.type === "command") return;
+
+		// Hide the popup and don't let anything else to happen
+		event.target.closest("menupopup").hidePopup(true);
+		event.stopPropagation();
+		event.preventDefault();
+
+		// Create a copy of the 'command' event and re-dispatch it after a delay
+		let originalTarget = event.target;
+		let delayedEvent = document.createEvent("XULCommandEvent");
+		delayedEvent.initCommandEvent(
+			event.type,
+			event.bubbles,
+			event.cancelable,
+			event.view,
+			event.detail,
+			event.ctrlKey,
+			event.altKey,
+			event.shiftKey,
+			event.metaKey,
+			0,
+			event,
+			0
+		);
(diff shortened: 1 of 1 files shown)


abaevbog:
Hm, CI is failing on `ruby/setup-ruby` step before tests get to run. That's something new ...

AbeJellinek:
Do we even use Ruby for anything anymore?

But maybe it'll work on a rerun.

AbeJellinek:
> `/usr/bin/cat: null: No such file or directory`

Hmmmm. What?

abaevbog:
Yeah, I'm not sure. github status page shows no errors. I'm rerunning another random PR now to see if it happens there.

abaevbog:
Ok, yes, it happened here too: https://github.com/zotero/zotero/actions/runs/16509556110. So my guess is that CI build is just generally broken now.

abaevbog:
I think the issue is in the actual `setup-ruby` action step - there are some commits being reverted right now: https://github.com/ruby/setup-ruby/commit/0ecad18fe538ef70f6b82773daecc6af1a7fe58a.

abaevbog:
Ok, locally I have all tests passing. CI just keeps tripping on that problematic tagsbox test `should update tag when pressing Enter in textbox`.

abaevbog:
All is finally passing now!

AbeJellinek:
Could we check whether the menupopup is inside a menulist, and if it is, let the event through synchronously? That would handle the failing tests, I think. I might be forgetting something, but I can't think of a case where we open a dialog from a menupopup that isn't the native file picker (which this issue shouldn't affect).

abaevbog:
Oh yeah, great point. I can't think of any instances of `menulists` opening dialogs, so just stopping if we get a `command` from one of those sounds good to me.

chrome/content/zotero/customElements.js
```diff
@@ -109,6 +109,53 @@ Services.scriptloader.loadSubScript('chrome://zotero/content/elements/itemTreeMe
...
+		if (!event.target.closest("menupopup")) return;
+
+		// Ignore commands from within menulists, since no dialogs are opened from them
+		if (event.target.closest("menulist")) return;
+
+		// If the source event of this 'command' event is another 'command' event,
+		// it must be a delayed event re-dispatched by the code below, so let it propagate
+		if (event.sourceEvent?.type === "command") return;
+
+		// Hide the popup and don't let anything else to happen
```
AbeJellinek, line 132:
```suggestion
		// Hide the popup and don't let anything else happen
```

AbeJellinek:
I think this looks good!

chrome/content/zotero/customElements.js
abaevbog: Thanks! English can be hard...

abaevbog:
@AbeJellinek, I ran into a new UI glitch on fx140 on windows. If a context menu option opens a dialog, it can sometimes (pretty often on my VM) lead to this weird looking effect with an empty menu:

Screenshot 2025-07-29 at 2 06 12 PM

I assume what happens is the modal blocks the main thread before the menu is fully gone. Once the dialog is dismissed, the menu finally goes away. I could not reproduce it on current beta on windows, so it seems new to fx140. Are you seeing it as well?

After rebasing to fx140, these changes here seem to fix that issues as well. What do you think I should do? Should I rebase and make a PR to fx140? Or just leave this as it is now.

abaevbog:
Hmm, though now I don't see it anymore ... I guess it's not fx140-specific after all! Let's leave this as is then, and if it does emerge again, I can rebase then.

Issue #5414, opened 2025-07-28
fx140: Replace `this._searchDebouncePromise.isPending()` in citationDialog.js


dstillman:
Synchronous Promise inspection no longer available after Bluebird removal

dstillman:
(The `.cancel()` call below it also need to be removed.)

abaevbog:
This PR should keep the existing behavior without using `.isPending()` or `.cancel()`. It relies on timestamps to decide if search should run or not after the delay, which I thought is cleaner than trying to do something else with promises.

PR #5416, opened 2025-07-28
Fx140: do not use bluebird methods in citation dialog

abaevbog:
Instead, decide if the search should be skipped after the delay based on the timestamps.

Fixes: #5414

### DIFF
--- a/chrome/content/zotero/integration/citationDialog.js
+++ b/chrome/content/zotero/integration/citationDialog.js
@@ -197,7 +197,7 @@ function _id(id) {
 class Layout {
 	constructor(type) {
 		this.type = type;
-		this._searchDebouncePromise = null;
+		this._lastSearchTime = null;
 	}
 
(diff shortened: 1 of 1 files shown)


chrome/content/zotero/integration/citationDialog.js
```diff
@@ -319,15 +319,17 @@ class Layout {
...
-		}
 		if (!skipDebounce) {
-			this._searchDebouncePromise = Zotero.Promise.delay(SEARCH_TIMEOUT);
-			await this._searchDebouncePromise;
+			// record when the search started
+			let searchStartTime = Date.now();
+			this._lastSearchTime = searchStartTime;
+			
+			// wait a moment
+			await new Promise(resolve => setTimeout(resolve, SEARCH_TIMEOUT));
```
dstillman, line 328: We can still use `Zotero.Promise.delay()` for this sort of thing.

Issue #5430, opened 2025-07-30
fx140: linux arm has transparent gap between tag selector and its separator


abaevbog:
<img width="251" height="81" alt="Image" src="https://github.com/user-attachments/assets/958f1014-e9fa-4246-aa65-f1acf0040451" />

abaevbog:
Not specific to tag selector:

<img width="858" height="224" alt="Image" src="https://github.com/user-attachments/assets/0700fb42-bcfc-4c50-b112-0c3d242a1dc5" />

PR #5432, opened 2025-07-30
fx140: fix separators appearance on linux

abaevbog:
- fix transparent gap between tag selector and separator
- remove double border below separator of context pane in stacked mode

Fixes: #5430

Before (double border above stacked context pane):
before_context

Before (transparent gap):
before_tag

After (context pane):
after_context
After (tag selector):
Screenshot 2025-07-30 at 3 49 04 PM


abaevbog:
Oh, I guess this is not new in fx140, I just never noticed it earlier on my x86 ubuntu laptop. But when I built and ran fx140 on arm linux VM, it looked much more apparent there.

PR #5436, opened 2025-07-31
Improve item pane test stability

windingwind:


windingwind:
Failures like https://github.com/zotero/zotero/actions/runs/16637597361/job/47081529179 probably come from the `waitForPreviewBoxRender` getting immediately resolved by the flags from previous render before the current render actually starts. Try to fix this by awaiting the flag and matching by item id.

dstillman:
Looks like it still failed

dstillman:
Oh, wait, sorry, it failed later on an unrelated test, so maybe it's fine.

PR #5443, opened 2025-07-31
Fix Find Full Text failing on ProQuest

### LINKED ISSUE #5442: Find Full Text ignores `Set-Cookie` header before redirect
Which breaks ProQuest.

```
(5)(+0000000): CookieSandbox: Not touching channel for www.proquest.com

(5)(+0000320): CookieSandbox: Not touching channel for www.proquest.com

(3)(+0000001): HTTP GET https://www.proquest.com/intermediateredirectforezproxy succeeded with 302

(3)(+0000001): HTTP GET https://www.proquest.com/intermediateredirectforezproxy

(5)(+0000000): CookieSandbox: Not touching channel for www.proquest.com

(5)(+0000501): CookieSandbox: Not touching channel for www.proquest.com

(3)(+0000001): HTTP GET https://www.proquest.com/intermediateredirectforezproxy succeeded with 302

(3)(+0000001): HTTP GET https://www.proquest.com/intermediateredirectforezproxy

(5)(+0000001): CookieSandbox: Not touching channel for www.proquest.com

(5)(+0000323): CookieSandbox: Not touching channel for www.proquest.com

(3)(+0000001): HTTP GET https://www.proquest.com/intermediateredirectforezproxy succeeded with 302

(3)(+0000001): Too many redirects to https://www.proquest.com/intermediateredirectforezproxy -- stopping
```

We handle redirects manually in `Zotero.Attachments.downloadFirstAvailableFile()`, so we should use a `CookieSandbox` to track cookies between requests.

Example URL (put this in the URL field of a new item): https://www.proquest.com/docview/2778644623?sourcetype=Dissertations%20&%20Theses

Reported on the forums: https://forums.zotero.org/discussion/comment/496558/#Comment_496558

AbeJellinek:
And probably some other sites with auth redirects that set cookies.

- Wrap Find Full Text requests and translation in a CookieSandbox
	- Pre-fx115, our XHRs would've implicitly used global cookies. Now system XHRs [are anonymous](https://bugzilla.mozilla.org/show_bug.cgi?id=1881800) (`mozAnon`) by default, so we need to use a CookieSandbox in order to track cookies between requests. (Redirects are separate requests here because we handle them manually.)
- ProQuest-specific: When you first connect to ProQuest using a proxy/VPN, it sends you through https://www.proquest.com/intermediateredirectforezproxy, which uses a JS redirect to bring you back to the page you originally requested. We can't detect that, so just treat a 200 response from `/immediateredirectforezproxy` like a redirect back to the original URL.
- Fix JSDoc comment, remove unused variable

Fixes #5442

### DIFF
--- a/chrome/content/zotero/xpcom/attachments.js
+++ b/chrome/content/zotero/xpcom/attachments.js
@@ -39,6 +39,7 @@ Zotero.Attachments = new function () {
 	
 	var _findFileQueue = [];
 	var _findFileQueuePromise = null;
+	var _findFileCookieSandboxes = new Map();
 	
 	var self = this;
 	
@@ -2052,6 +2053,8 @@ Zotero.Attachments = new function () {
 			if (!url && !pageURL) {
 				continue;
 			}
+
+			let cookieSandbox = _getFindFileCookieSandboxForSite(url || pageURL);
 			
 			if (urlResolver.referrer) {
 				options.referrer = urlResolver.referrer;
@@ -2069,7 +2072,7 @@ Zotero.Attachments = new function () {
 				while (tries-- > 0) {
 					try {
 						await beforeRequest(url);
-						await this.downloadFile(url, path, options);
+						await this.downloadFile(url, path, { ...options, cookieSandbox });
 						afterRequest(url);
 						return { url, props: urlResolver };
 					}
@@ -2092,7 +2095,6 @@ Zotero.Attachments = new function () {
 				try {
 					Zotero.debug(`Looking for file on ${pageURL}`);
 					
-					let redirects = 0;
 					let nextURL = pageURL;
 					let req;
 					let blob;
@@ -2125,7 +2127,8 @@ Zotero.Attachments = new function () {
 										responseType: 'blob',
 										followRedirects: false,
 										// Use our own error handling
-										errorDelayMax: 0
+										errorDelayMax: 0,
+										cookieSandbox,
 									}
 								);
 							}
@@ -2223,7 +2226,7 @@ Zotero.Attachments = new function () {
 					}
 					// Otherwise translate the Document we parsed above
 					else if (doc) {
-						({ title, mimeType, url } = await Zotero.Utilities.Internal.getFileFromDocument(doc));
+						({ title, mimeType, url } = await Zotero.Utilities.Internal.getFileFromDocument(doc, { cookieSandbox }));
 					}
 				}
 				catch (e) {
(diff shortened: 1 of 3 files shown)


dstillman:
I worry that using a new sandbox each time will cause sites/Cloudflare to block more, since there will be repeated requests from the same IP with no previous cookies, which looks real bot-y.

Using global cookies with no way to reset them was always a little unsettling, but one could imagine that it improved download results, particularly from something with cross-internet visibility like Cloudflare. Of course, one could also imagine the inverse: that seeing the same device download huge numbers of files could get it blocked more quickly. (Sites obviously have the IP address regardless, but the cookie makes clear it's not just some exit node.) And old cookies might have caused weird, hard-to-debug failures.

A compromise here might be to maintain a cookie sandbox per domain ([eTLD+1](https://developer.mozilla.org/en-US/docs/Glossary/eTLD)), and persist that in memory, maybe with an inactivity timeout (24 hours?). That would fix this, roughly mirror Firefox's own behavior, and stay relatively privacy preserving compared to global cookies. (I don't know if there's some way we can just piggyback on Firefox's Total Cookie Protection, but our needs are probably simpler.)

AbeJellinek:
OK, let me know how that looks.

dstillman:
Do we want to clear expired sandboxes proactively? This will only clear them on the next use for that site. Are the sandboxes holding on to any memory?

dstillman:
(Also, `Services.eTLD.getSite(url)` — nice!)

dstillman:
Oh, though apparently it [doesn't like some input](https://github.com/zotero/zotero/actions/runs/16677917704/job/47209057928?pr=5443):

```
       Component returned failure code: 0x80070057 (NS_ERROR_ILLEGAL_VALUE) [nsIEffectiveTLDService.getSite] at
_getFindFileCookieSandboxForSite@chrome://zotero/content/xpcom/attachments.js:2289:30
        Zotero.Attachments</this.downloadFirstAvailableFile@chrome://zotero/content/xpcom/attachments.js:2057:59
```

AbeJellinek:
OK, that should fix it.

> Do we want to clear expired sandboxes proactively?

We could. The sandboxes don't use a lot of memory, just an object for the cookies. No harm in it, though.

AbeJellinek:
OK, added. It only clears out entries when `_getFindFileCookieSandboxForSite()` is called, but I think that's good enough.

AbeJellinek:
OK, the relevant tests are passing. I'll squash into fewer commits.

PR #5444, opened 2025-07-31
Cleanup opened windows before loading new zotero pane window in tests

abaevbog:
Some tests may cause windows or dialogs to open without closing them. This can interfere with other tests
that rely on `zoteroPane` becoming the active window when loaded. This will ensure that before `zoteroPane` is loaded, all windows (besides `runtests.html`) are be closed.


abaevbog:
I investigated a bit more what's been happening with timeouts around `should update tag when pressing Enter in textbox` test, and these are the two findings I made:

I don't believe that `Zotero.Utilities.Internal.activate(win)` called from `activateZoteroPane()` has any effect during test runs on linux x86. This is where `_x11BringToForeground` is called:
https://github.com/zotero/zotero/blob/69fdf0b14580a5ea64f0317eec68c2b0846a4403/chrome/content/zotero/xpcom/utilities_internal.js#L2971-L2976
`_x11BringToForeground` will be only called after the `load` event on the window fires. So, if the window is already loaded and `Zotero.Utilities.Internal.activate(win)` is called, nothing will happen.

I tried to workaround the above by calling `_x11BringToForeground` immediately if `window.document.readyState == "complete"`. This, however, still gave me timeouts. Then, I added a few `dump()` statements into the tests, and it looks like the issue is that `Services.focus.activeWindow` in `tagsboxTest` is `chrome://zotero/content/progressQueueDialog.xhtml`... For reference, this is the active window logged during test run: https://github.com/abaevbog/zotero/actions/runs/16658572103/job/47149767581#step:13:2669.

I'm not sure why but perhaps some of the progress queue dialogs don't end up being closed, so as a solution I added `closeAllWindows` to close all windows opened during the test run. That way, when `loadZoteroPane` runs, nothing should get in the way of it becoming the active window.

@AbeJellinek, does the above make sense to you?

dstillman:
> Then, I added a few dump() statements into the tests, and it looks like the issue is that Services.focus.activeWindow in tagsboxTest is chrome://zotero/content/progressQueueDialog.xhtml

Nice!

For the fix, though, I think we'll just want to identify the tests that are leaving windows open and close them there, since there could break other things unexpectedly. In this case, if you run `./test/runtests.sh -s server_connector`, you can see that those tests leave the progress dialog open. I've fixed that in #5446. That's still running, and there might be other tests that leave it open, but I think that's what we'll want.

abaevbog:
Understood! To check if there are any more tests that do not close their windows, I logged opened windows after every test, and I ran into two more instances:
- `hyphenationTest` opens the dialog but never closes it
-  `should clear sync errors from the toolbar after logging in` in `preferences_syncTest.js` opens a zoteroPane without properly closing it in time. So, if you run `test/runtests.sh -s preferences_syncTest.js`, when subsequent tests open their own `zoteroPane`, we end up with 2 instances of `zoteroPane` windows at the same time

Both should be fixed in the linked PR. Hope fully this will take care of all the random timeouts!

PR #5460, opened 2025-08-01
Close more leftover windows in tests

abaevbog:
- close `hyphenationTest window.
- properly close `zoteroPane` window in `preferences_syncTest`. `after()` hook inside of the test would not close the window until the very end of the entire test run. Without this, during subsequent tests that open their own `zoteroPane` window, there are two instances of `zoteroPane` open at the same time.
- close open tabs in `readerTest` cleanup, since subsequent tests expect library tab to be selected.

Followup to https://github.com/zotero/zotero/pull/5444#issuecomment-3141879892

### DIFF
--- a/test/tests/hyphenationTest.js
+++ b/test/tests/hyphenationTest.js
@@ -1,8 +1,14 @@
 describe("Hyphenation", function () {
+	let win;
+
+	after(function() {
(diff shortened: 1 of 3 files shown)


dstillman:
Great!

PR #5461, opened 2025-08-01
Warn users running non-ARM version on ARM CPU on Linux

### LINKED ISSUE #5448: Update architecture warning to include non-ARM Linux build on ARM hardware
Probably very few people are actually doing this, but the ones who are should definitely switch!

(Unfortunately I just destroyed my VM that was set up to test this. @tnajdek, if you don't have a VM set up this way, you can just do it blind and we can see if someone else can test.)

tnajdek:
I wasn't able to test it as a whole - I set up a Parallels VM running Ubuntu with the Rosetta layer for this and jumped through some hoops trying to get it to work, but the x64 binary segfaults on that system (7.x does as well).

Resolve #5448

### DIFF
--- a/chrome/content/zotero/xpcom/zotero.js
+++ b/chrome/content/zotero/xpcom/zotero.js
@@ -1419,6 +1419,35 @@ const { CommandLineOptions } = ChromeUtils.importESModule("chrome://zotero/conte
 			return false;
 		}
 	}
+
+	this.isLinux64EmulatedOnArm = async function () {
+		if (!this.isLinux) {
+			return false;
+		}
+
+		if (this.arch !== "x86_64") {
+			// We only check if x86_64 build is running on ARM
+			return false;
+		}
+		
+		try {
+			let uname = Cc["@mozilla.org/file/local;1"].createInstance(Ci.nsIFile);
+			uname.initWithPath("/bin/bash");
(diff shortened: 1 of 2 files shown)


dstillman:
I asked ChatGPT for a js-ctypes version that avoids the shell call. Not tested with an emulated build, but `machine` is indeed `aarch64` for me.

tnajdek:
Yeah, that will work even better!

AbeJellinek:
Has anyone actually gotten Zotero running through emulation on ARM Linux? I've never seen any forum posts about it. And in my own tests, it doesn't seem like Rosetta for Linux can handle GUI apps, and [Zotero doesn't work on box64](https://github.com/ptitSeb/box64/issues/1826).

dstillman:
Unless I'm totally misremembering, I had it running, after installing amd64 sources and some amd64 libraries. But that's the VM that self-destructed the other day. If I recall, it was very, very slow.

dstillman:
But yes, it is quite possible this code will be run by no one ever. Would have to be someone who was already set up to run x86_64 binaries and then downloaded the x86_64 version by mistake. Which we're hopefully going to get better at preventing with the use of [User-Agent Client Hints](https://developer.mozilla.org/en-US/docs/Web/API/User-Agent_Client_Hints_API) (at least for Chrome/Edge).

But hopefully this is a very lightweight call and there's no real downside.

Issue #5463, opened 2025-08-01
Copy collection to library root sends it to wrong destination


abaevbog:
Trying to copy a subcollection to library root via "Copy collection" context menu will instead create a copy under the same parent collection.

https://forums.zotero.org/discussion/125634/copy-sub-collection-to-my-library-going-to-the-wrong-location

PR #5464, opened 2025-08-01
Fix subcollection not being copied to library root via context menu

abaevbog:
Fixes: #5463


abaevbog:
Reported in https://forums.zotero.org/discussion/125634/copy-sub-collection-to-my-library-going-to-the-wrong-location

PR #5465, opened 2025-08-04
Remember last import settings

### LINKED ISSUE #5303: Remember last import settings
For "Place imported collections and items into new collection" and copy/link

tnajdek:
* Stores import options as `import.shouldCreateCollection` and `import.fileHandling` preferences
* Preferences are set whenever the value changes, so if the user configures import options then cancels, the preference is still stored
* If we ever decide to enable folder import, we’ll probably want to preserve folder importing options
* While testing, I managed to open two instances of the file picker dialog, so I’ve fixed that as well

Close #5303


dstillman:
The pref can just be `import.createCollection`.

dstillman:
> Preferences are set whenever the value changes, so if the user configures import options then cancels, the preference is still stored

Doesn't really matter, but I don't think it should do this. It would be different for a persistent UI state like showing advanced options via a disclosure triangle, but I don't think something that would actually affect the operation should be persisted unless the operation is performed.

tnajdek:
Thanks, updated!

PR #5467, opened 2025-08-06
Fix linux form element styling. Closes #5435

### LINKED ISSUE #5435: fx140: Linux form control styling issues
fx128: 

Image

fx140:

Image

Seems like all form elements (text, button, select) have lost padding and have a darker border.

adomasven:
Fixes form element (button, menulist, input field) styling, which better matches the styling of the form controls in the OS widgets and that of fx128. Tested in light and dark mode. Uses default colors to ensure compatibility with as little additional maintenance as possible.

Before:
Image

After:
image


scss/xulElementPatches/dialog-unix.scss
```diff
@@ -0,0 +1,22 @@
+// Override/enhance mozilla <dialog> styles on Windows. See customElements.js
```
dstillman, line 1: On Linux?

PR #5469, opened 2025-08-06
Hide text editing caret in read-only fields

AbeJellinek:
To align with normal web content / other apps.

### DIFF
--- a/scss/elements/_editableText.scss
+++ b/scss/elements/_editableText.scss
@@ -96,6 +96,10 @@ editable-text {
 			text-align: inherit;
 		}
 		
+		&:read-only {
+			caret-color: transparent;
+		}
+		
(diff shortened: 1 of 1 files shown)


dstillman:
Chrome and Safari do a version of this, but I'm not sure why. Firefox shows the caret.

In Chrome/Safari, it means you can't select a substring with the keyboard. Compare https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/readonly in Chrome and Firefox.

In our case, since we're just hiding the caret rather than disabling it, you can still select with the keyboard, but you don't know where your cursor is until you do.

I don't really see the downside to showing it?

AbeJellinek:
Yeah, fair. I thought it made it less clear that the Scaffold item details preview was read-only, but it’s not too bad.

PR #5471, opened 2025-08-07
create new notes with related items from selected items

lukeburns:
- adds a new context menu item "Create Note From Items" which will create a new note with selected items as related items
- if multiple items are selected, the new note hotkey will create note from selected items


dstillman:
Thanks, but this is a bit too limited and easy to replicate with existing functionality for a core feature. You can look into the Actions and Tags plugin for a way to assign this to a shortcut key.

PR #5472, opened 2025-08-07
Send deleted annotations to trash

### LINKED ISSUE #5466: Send deleted annotations to trash
https://forums.zotero.org/discussion/125908/recover-deleted-annotations-from-trash

abaevbog:
Annotations will be sent to trash instead of being immediately erased, similar to other item types. That is the case when deleting from both `itemTree` and the reader. From trash, annotations can be permanently deleted or restored. `Reader.AnnotationManager.undo` will also restore annotations from trash.

Needs: https://github.com/zotero/reader/pull/173
Fixes: #5466

### DIFF
--- a/chrome/content/zotero/xpcom/data/item.js
+++ b/chrome/content/zotero/xpcom/data/item.js
@@ -2167,6 +2167,29 @@ Zotero.Item.prototype.numChildren = function (includeTrashed) {
 	return this.numNotes(includeTrashed) + this.numAttachments(includeTrashed);
 }
 
+/**
+ * Get all descendents of an item.
+ * @param	{Boolean}	includeTrashed		Include trashed items
+ * @return	{Zotero.Item[]} Array of Zotero.Item objects, whose parent, or parent's parent, is this item
+ */
+Zotero.Item.prototype.getAllDescendents = function (includeTrashed) {
+	let attachments = [];
+	let notes = [];
+	let annotations = [];
+	if (this.isRegularItem()) {
+		attachments = Zotero.Items.get(this.getAttachments(includeTrashed));
+		notes = Zotero.Items.get(this.getNotes(includeTrashed));
+		for (let attachment of attachments) {
+			if (!attachment.isFileAttachment()) continue;
+			annotations = annotations.concat(attachment.getAnnotations(includeTrashed));
+		}
+	}
+	else if (this.isFileAttachment()) {
+		annotations = this.getAnnotations(includeTrashed);
+	}
+	return [...attachments, ...notes, ...annotations];
+};
+
 
 /**
  * @return	{String|FALSE}	 Key of the parent item for an attachment or note, or FALSE if none
--- a/chrome/content/zotero/xpcom/data/item.js
+++ b/chrome/content/zotero/xpcom/data/item.js
@@ -2460,28 +2460,30 @@ Zotero.Item.prototype.numChildren = function (includeTrashed) {
 
 
 /**
- * Get all descendents of an item -- child notes and attachments of a regular item,
- * annotations of its file attachments, and annotations of a standalone file attachment
+ * Get descendant items -- child notes and attachments of a regular item, plus
+ * annotations of its file attachments; annotations of a standalone file attachment
  *
- * @param	{Boolean}	includeTrashed		Include trashed items
- * @return	{Zotero.Item[]}
+ * @param {Object} [options]
+ * @param {Boolean} [options.includeTrashed=false] Include trashed items
+ * @param {Boolean} [options.includeAnnotations=true] Include annotations of file attachments
+ * @return {Zotero.Item[]}
  */
-Zotero.Item.prototype.getAllDescendents = function (includeTrashed) {
-	let attachments = [];
-	let notes = [];
-	let annotations = [];
+Zotero.Item.prototype.getDescendantItems = function ({ includeTrashed = false, includeAnnotations = true } = {}) {
+	let descendents = [];
 	if (this.isRegularItem()) {
-		attachments = Zotero.Items.get(this.getAttachments(includeTrashed));
-		notes = Zotero.Items.get(this.getNotes(includeTrashed));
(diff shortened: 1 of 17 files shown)


abaevbog:
@mrtcode, this PR somewhat touches on changes from https://github.com/zotero/reader/commit/de1f3e69e214f8ea4aea054ae788a349e307fada, so I wanted to check in with you if this makes sense.

Currently, if you delete an annotation, it is immediately erased, so when you `undo` it, a fresh annotation item is created. That fresh annotation item gets a fresh `key` to avoid sync errors.

With this, a deleted annotation appears in trash, and `undo` removes it from there, essentially just toggling `annotation.deleted`. For this to work, I had to remove the part where annotation would get a newly generated key on each `undo` in the linked reader PR (because otherwise, we would create a duplicate annotation instead of restoring the trashed one). This is the reason for the CI test failure here - it should pass with the linked reader tweak. When one finally erases the annotation from trash, it is removed from the edit history. Otherwise, one could trash and erase the annotation, sync, and then re-create that annotation with the same key via `undo` (which, I believe, is what we are trying to avoid by regenerating a new key on each `undo` now).

Does this sound reasonable to you?

mrtcode:
Yes, it looks good.

Just restoring annotations from trash doesn't seem to work. Some annotations have their context menu with all options disabled, and for others, clicking "Restore to Library" does nothing.

abaevbog:
> Just restoring annotations from trash doesn't seem to work. Some annotations have their context menu with all options disabled, and for others, clicking "Restore to Library" does nothing.

Can you share more on what you are seeing? The intended behavior at the moment is that if a single annotation row is trashed, all of its non-trashed siblings also appear in the trash (similar to how non-trashed attachments or notes are visible if one of their siblings is trashed). Such non-trashed annotation rows can't be deleted or restored, so all context menu options are disabled for them. And then, once you restore a trashed annotation row, it still remains in the trash, as long as there are other trashed siblings - it just stops being a `.context-row`.

https://github.com/user-attachments/assets/5f34f3a6-9359-4f40-bbea-bcf0a14a94d4

Once all trashed rows are restored or deleted, they all go away from trash with the top-level item. I agree it doesn't feel too interactive that if you restore an annotation, nothing visibly happens - but that's how other child rows behave as well. It also may change after #5315 when context rows can be hidden.

mrtcode:
I'm not sure why, but restoring works for some PDFs and not for others.

https://github.com/user-attachments/assets/3b89ed98-b940-4632-979f-66e5f20bcd98

abaevbog:
I believe this could have happened with standalone attachments. Does this fix it for you now?

mrtcode:
Looks good now!

abaevbog:
Awesome! Thank you for checking this.

abaevbog:
@AbeJellinek, I applied these older changes on top of your unified undo history commit. How does it look to you? It needs a small change to the reader in linked PR https://github.com/zotero/reader/pull/173.

chrome/content/zotero/xpcom/data/item.js
```diff
@@ -2459,6 +2459,32 @@ Zotero.Item.prototype.numChildren = function (includeTrashed) {
...
 
 
+/**
+ * Get all descendents of an item -- child notes and attachments of a regular item,
+ * annotations of its file attachments, and annotations of a standalone file attachment
+ *
+ * @param	{Boolean}	includeTrashed		Include trashed items
+ * @return	{Zotero.Item[]}
+ */
+Zotero.Item.prototype.getAllDescendents = function (includeTrashed) {
```
AbeJellinek: We'll definitely have some uses for this elsewhere! Though we won't always want second-level descendants (aka descendents). I know there's a lot of code scattered around that concatenates all the attachments and notes of an item and repeats the exact same `isRegularItem()` check. If we're going to introduce this method, I think we should make it usable for those cases as well.

Maybe add a second optional argument, an `options` object with `maxLevel` (1 for only direct children, >= 2 or null for second-level children as well)? That would make the "all" in the method name misleading (and I think it's redundant as is), but `getDescendents()` risks confusion with the complicated `Zotero.Collection.getDescendents()` signature. But we could call it, like, `getDescendantItems()`?

That or just leave it, and callers can filter the returned array to remove annotations if they want.

(Also, for what it's worth, notes with images have attachment children, which this currently ignores.)

chrome/content/zotero/xpcom/reader.js
```diff
@@ -338,14 +356,18 @@ class ReaderInstance {
...
-						let annotation = Zotero.Items.getByLibraryAndKey(libraryID, key);
-						// Make sure the annotation actually belongs to the current PDF
-						if (annotation && annotation.isAnnotation() && annotation.parentID === this._item.id) {
-							this.annotationItemIDs = this.annotationItemIDs.filter(id => id !== annotation.id);
-							await annotation.eraseTx({ notifierQueue });
+					// Deleted annotations are sent to the trash
+					await Zotero.DB.executeTransaction(async () => {
+						for (let key of keys) {
+							let annotation = Zotero.Items.getByLibraryAndKey(libraryID, key);
+							// Make sure the annotation actually belongs to the current PDF
```
AbeJellinek:
```suggestion
							// Make sure the annotation actually belongs to the attachment
```

which is not necessarily a PDF!

chrome/content/zotero/xpcom/data/item.js
abaevbog: That's a fair point - we do have this pattern of iterating over children in a few places.

I am concerned though that introducing level parameter may end up being confusing because of standalone attachments. If we specify `maxLevel=1`, then for regular items we would return their attachment and note children but for standalone attachments, it would imply that annotations should be returned, since it's the next level? 

It looks like in most cases, it's just about if annotations need to be returned or not, so what do you think about `getDescendantItems = function ({ includeTrashed = false, includeAnnotations = true } = {})`? I think it fits into existing spots where we iterate over children (e.g. restoring items from trash) just fine.

Issue #3463, opened 2023-10-29
Remove "Create Timeline"


dstillman:
This code hasn't been updated in many years, and there's no reason for it to be built in. If someone wants to maintain a timeline feature, they should do it as a plugin.

PR #5484, opened 2025-08-14
Remove timeline related code

windingwind:
Since timeline feature is remove in Zotero 7 (https://www.zotero.org/support/timelines), there's no reason to keep these code.


windingwind:
resolve: #3463

Issue #5481, opened 2025-08-13
Item relation not removed when target item is deleted


dstillman:
https://forums.zotero.org/discussion/125399/ghost-related-items

1. Add Book item.
2. Create Book Section item from Book item.
3. Move Book Section to trash.
5. Empty trash

Book item shows "1 Related" with nothing under it:

<img width="305" height="41" alt="Image" src="https://github.com/user-attachments/assets/78218753-6430-44b6-a40d-676e1a355371" />

Because it still has the relation:

```json
{
    "key": "Z8892IRF",
    "version": 43151,
    "itemType": "book",
    "title": "A",
    "creators": [],
    "tags": [],
    "collections": [],
    "relations": {
        "dc:relation": [
            "http://zotero.org/users/6/items/92BPI74Q"
        ]
    },
    "dateAdded": "2025-08-13T03:56:36Z",
    "dateModified": "2025-08-13T03:56:37Z"
}
```

Unlike https://github.com/zotero/zotero/issues/1780, this is for actual deletions. I assume this is a regression? Hard to imagine it's always been like this.

abaevbog:
I'm struggling to reproduce this so far ... If I follow the steps, once the trash is emptied, the erased item is removed from `item.relatedItems` and is not returned in `item.getRelations()`. The entry is also, naturally, removed from `itemRelations` table. Have you been able to reproduce this?

I'll keep trying to run into it but let me know if I am missing in any of the steps below:

https://github.com/user-attachments/assets/a1ae100a-8216-49bc-8d12-f177885b25f5

abaevbog:
I can see it now. An important condition for me to reproduce it was sync. I have automatic sync disabled, so then issue does not occur. I have to sync once after trashing the book section and one more time before clicking back on the remaining book item.

abaevbog:
I could reproduce a similar issue with merging related items: https://forums.zotero.org/discussion/111552/zotero-7-beta-merging-creates-a-self-related-item

abaevbog:
https://forums.zotero.org/discussion/109545/zotero-7-beta-ghost-related-item

abaevbog:
Per discussion in https://github.com/zotero/zotero/pull/5486#issuecomment-3190793335 this should be fixed by https://github.com/zotero/dataserver/pull/172

PR #5486, opened 2025-08-14
Fix relations to erased items re-appearing after sync

abaevbog:
Fix ghost relations to erased items re-appearing after sync. The dataserver ensures that `dc:relation` relations are bidirectional. On the client side, a simple way to verify this is to remove relation from one item, save it, sync - the relation will be restored. When a related item is erased, syncing would first upload the update to the remaining related item (where the relation is removed), followed by the deletions. While handling update upload, the dataserver would return the old, should-be-removed, `dc:relation` entry for that item because it does not know that the other item is erased.

To handle this, swap the order of operations during sync: upload deletion first and then upload updates after. Then, when the dataserver handles the update of the non-erased item, there is no related item to enforce bidirectional relations to.

Fixes: #5481

### DIFF
--- a/chrome/content/zotero/xpcom/sync/syncEngine.js
+++ b/chrome/content/zotero/xpcom/sync/syncEngine.js
@@ -1094,7 +1094,7 @@ Zotero.Sync.Data.Engine.prototype._startUpload = async function () {
 				objectType, objectDeletions[objectType], libraryVersion
 			);
 		}
-
+		
 		Zotero.debug(JSON.stringify(objectIDs));
 		for (let objectType in objectIDs) {
 			this._statusCheck();
--- a/chrome/content/zotero/xpcom/sync/syncEngine.js
+++ b/chrome/content/zotero/xpcom/sync/syncEngine.js
@@ -1086,21 +1086,21 @@ Zotero.Sync.Data.Engine.prototype._startUpload = async function () {
 	}
 	
 	try {
-		Zotero.debug(JSON.stringify(objectIDs));
-		for (let objectType in objectIDs) {
-			this._statusCheck();
-			
-			libraryVersion = await this._uploadObjects(
-				objectType, objectIDs[objectType], libraryVersion
-			);
-		}
-		
 		Zotero.debug(JSON.stringify(objectDeletions));
 		for (let objectType in objectDeletions) {
 			this._statusCheck();
 			
 			libraryVersion = await this._uploadDeletions(
 				objectType, objectDeletions[objectType], libraryVersion
+			);
+		}
+
+		Zotero.debug(JSON.stringify(objectIDs));
+		for (let objectType in objectIDs) {
+			this._statusCheck();
+			
+			libraryVersion = await this._uploadObjects(
+				objectType, objectIDs[objectType], libraryVersion
 			);
 		}
 	}
--- a/test/tests/syncEngineTest.js
+++ b/test/tests/syncEngineTest.js
@@ -3537,29 +3537,31 @@ describe("Zotero.Sync.Data.Engine", function () {
 			assert.ok(Zotero.Items.get(item1.id));
 			assert.isFalse(Zotero.Items.get(item2.id));
 		});
-
-		it("should not return relation to a deleted item in response", async function () {
+		
+		
+		// https://github.com/zotero/zotero/pull/5486
+		it("should not cause server to return relation to a deleted item", async function () {
 			({ engine, client, caller } = await setup());
-
+			
 			let library = Zotero.Libraries.userLibrary;
 			let libraryVersion = 5;
 			library.libraryVersion = libraryVersion;
 			await library.saveTx();
-
-			var itemOne = await createDataObject('item');
-			var itemTwo = await createDataObject('item');
-			itemOne.addRelatedItem(itemTwo);
-			itemTwo.addRelatedItem(itemOne);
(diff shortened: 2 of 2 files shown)


abaevbog:
I suppose another approach to this without changing the order of uploads could be to tweak the dataserver itself, since the issue really comes from there. Maybe we should just not enforce bidirectional relations if an item had a relation and then it was removed? Then, if the relation is not returned from dataserver, it would not matter if we upload update or deletion first.

dstillman:
Sorry, I shouldn't have merged this. This is a pretty major change to sync logic, and I suspect it would break things in unexpected ways. At the very least, we need to test it much more thoroughly. But I suspect we'll need a different solution.

When interrelated objects are deleted, Zotero makes various changes to both sides, and prior to this change, those would get uploaded first. Obviously that's a problem here, but in general it's what we want, because in lieu of those changes, the dataserver has to perform its own actions to ensure data integrity. So, for example, if you move collection B out of its parent collection A and then delete the collection A, the collection B modification would get uploaded first, and then the collection A deletion could just go through without affecting anything else. But if collection B is still a subcollection of collection A when A is deleted, the dataserver probably deletes collection B and modifies all items within it, which would trigger 1) a conflict between the local collection B modification and the remote deletion and 2) a remote modification to tons of items that haven't actually changed locally. And probably a similar situation for parent/child items, so we probably can't just limit this to items.

dstillman:
Let me look at what exactly the server is doing here.

abaevbog:
> So, for example, if you move collection B out of its parent collection A and then delete the collection A, the collection B modification would get uploaded first, and then the collection A deletion could just go through without affecting anything else. 

Right, that is a good example. To be honest, I suspected that with this change something would break elsewhere in some of the tests and was quite surprised that it did not happen (which gave me some reassurance).

Ultimately, on the dataserver side everything ends up being correct - it's just the returned relation that is the issue. If I end up with that ghost relation, and then I modify that item again and re-sync, then the ghost relation will be properly gone. We could instead try to handle this on the client side by ignoring the returned relation if we know that the relation was removed. But that can get messy if there are were multiple updates to the relations from different clients. Also, I'm not sure if it would mean that iOS and Android would have to handle it separately as well.

abaevbog:
Without running the full dataserver setup yet, I assume this all comes from here: https://github.com/zotero/dataserver/blob/3ed926b85e62dd7226d34decc4b14b2227dc5601/model/Item.inc.php#L5023-L5033

So maybe it's a matter of un-doing the inclusion of these relations while handling item update when that relation is excluded?

dstillman:
I'm not actually sure why we're including reverse relations anyway. The clients set the reverse relations explicitly (or are supposed to), so I'm not sure why we need to fake them when they're not yet set.

abaevbog:
It looks like was explicitly added in https://github.com/zotero/dataserver/commit/fbaa602bdd32b0f43b8997f51aa13686986eab01. More than a decade ago! Not exactly sure about the context there.

abaevbog:
And yes, as far as I can see, whenever related items are being handled, we explicitly add both directions. For example, `add` in relatedBox:
https://github.com/zotero/zotero/blob/bf59a1e2f80dd49d1e7622bf5ddbff3d6b236e56/chrome/content/zotero/elements/relatedBox.js#L175-L185

dstillman:
If I had to guess — which I do — the goal was to have the dataserver enforce the bidirectionality itself, with the idea that it could be used by unofficial clients that don't necessarily set both sides. But having the dataserver modify another item itself risks sync conflicts (see above), so we had to fake it and not actually modify the other item. But that makes for a very weird situation:

https://github.com/zotero/dataserver/commit/fbaa602bdd32b0f43b8997f51aa13686986eab01#diff-ebf103ea0ed19fcf4f2aa89452c23ef62ab13181339338f7b3c39587242faefbR103-R106

So the best option here might be to just let clients enforce bidirectionality on their own.

abaevbog:
That makes sense. I guess the clients have to deal with this bidirectionality question in this way or another - by either enforcing it themselves or by tackling the situations like here with ghost relations that do not actually exist. 

Should I make a dataserver PR with removed bidirectional relations and updated tests?

dstillman:
Yes, that'd be great.

abaevbog:
Sorry about the delay - it looks like the dataserver is now on php 8, and all of my local container setup/scripts/etc. was for php 7, so it took me a second to get everything up and running again.

I added the linked PR to remove the bidirectional nature of related items and updated the phpunit tests. Let me know if it makes sense!

I didn't touch the in-works mocha test suite from a while back - I think there have been a few other changes since the last time I looked at it, so that whole PR would have to be gone over completely in any case.

One minor thing I noticed is that we do not allow to relate an item to itself in [Zotero_Item.setRelatedItems](https://github.com/zotero/dataserver/blob/3ed926b85e62dd7226d34decc4b14b2227dc5601/model/Item.inc.php#L4972-L4975). But this is not enforced in lower level [Zotero_DataObjects.setRelations](https://github.com/zotero/dataserver/blob/3ed926b85e62dd7226d34decc4b14b2227dc5601/model/DataObject.inc.php#L393), which is what gets called when items are being [updated from JSON](https://github.com/zotero/dataserver/blob/3ed926b85e62dd7226d34decc4b14b2227dc5601/model/Items.inc.php#L1797). So at the moment, I can relate an item to itself, at least with my local setup. Is that something worth preventing as well?

PR #5495, opened 2025-08-19
Don't reopen collection/item pane on restart

### LINKED ISSUE #5490: Stop reopening panes on restart
https://forums.zotero.org/discussion/comment/497522/#Comment_497522

And remove `extensions.zotero.reopenPanesOnRestart`

AbeJellinek:
(We'll see if this breaks any tests...)

Closes #5490

### DIFF
--- a/chrome/content/zotero/xpcom/prefs.js
+++ b/chrome/content/zotero/xpcom/prefs.js
@@ -46,7 +46,7 @@ Zotero.Prefs = new function () {
 
 		// Process pref version updates
 		var fromVersion = this.get('prefVersion');
-		var toVersion = 15;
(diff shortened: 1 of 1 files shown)


dstillman:
I guess we should add a pref migration step with a `clear()` so that the dozen people in the world who ever changed this don't have a line sticking around in prefs.js for all eternity…

AbeJellinek:
OK, done, and tests should hopefully be fixed. (We now clear `pane.persist` by default every time test code wants a new main window, since there's lots of test code that can mess with pane states, and it doesn't seem like we'd normally want that to carry over to new windows.)

dstillman:
Merged

PR #5499, opened 2025-08-19
Add Relate Items to context menu

### LINKED ISSUE #5477: Relate selected items
As seen in [Zutilo](https://github.com/wshanks/Zutilo)

I think this is a pretty popular feature, and it seems fairly useful. We should probably just add it?

Select multiple items → right-click → Relate Items

AbeJellinek:
We show the Relate Items option for all non-feed items, but disable it when all selected items are already related. That keeps things consistent with the UI we use for Add to Collection, and I think it's a little less confusing than hiding the menu item entirely based on a non-obvious condition (relations between all selected items).

Will throw on mixed multi-selections in the trash until #5497 is merged.

Closes #5477

### DIFF
--- a/chrome/content/zotero/zoteroPane.js
+++ b/chrome/content/zotero/zoteroPane.js
@@ -2324,6 +2324,29 @@ var ZoteroPane = new function () {
 	};
 	
 	
+	this.relateSelectedItems = async function () {
+		if (!this.canEdit()) {
+			this.displayCannotEditLibraryMessage();
+			return;
+		}
+		
+		let selectedItems = this.getSelectedItems();
+		await Zotero.DB.executeTransaction(async () => {
+			for (let index1 = 0; index1 < selectedItems.length; index1++) {
+				for (let index2 = index1 + 1; index2 < selectedItems.length; index2++) {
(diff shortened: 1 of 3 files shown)


dstillman:
https://github.com/zotero/zotero/blob/8eb803292e6ca2cab864371bbe21693cb6529f65/chrome/content/zotero/elements/relatedBox.js#L177-L179

Should use `skipDateModifiedUpdate` as in relatedbox.js

dstillman:
In lieu of Date Modified changing, we probably could use some sort of UI feedback here. Just "[x] items were related" in a `Zotero.ProgressWindow`?

AbeJellinek:
"[x] items were related" is kind of ambiguous because it could mean that [x] items were related _before_ the operation (and so didn't change). Maybe "[x] items are now related"? "Added relations between [x] items" (maybe too technical)?

AbeJellinek:
And maybe we actually want a way to undo this? "Undo Relate Items", visible only when all the items in the selection are related to one another? Since there's no confirmation for this, it could be relatively easy to add relations between a very large number of items and then have no easy way to reverse that.

PR #5500, opened 2025-08-19
Keep Zotero in the User-Agent, add Firefox

AbeJellinek:
And:
- Remove `rv:` replacement after fx140 upgrade
- Remove special handling for tests, since we now run tests in Zotero

### DIFF
--- a/chrome/content/zotero/xpcom/zotero.js
+++ b/chrome/content/zotero/xpcom/zotero.js
@@ -2046,26 +2046,17 @@ Zotero.VersionHeader = {
 	},
 	
 	/**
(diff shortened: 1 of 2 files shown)


dstillman:
> Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:140.0) Gecko/20100101 Zotero/8.0.SOURCE.ba65b1dcc Firefox/140.0

 I think we'd want to follow the ~standard convention here and append `Zotero/` to the end, after `Firefox/`.

AbeJellinek:
OK, updated.

Issue #5501, opened 2025-08-19
Citation dialog: "Omit author" always looks unchecked on linux


abaevbog:
https://forums.zotero.org/discussion/126205/possible-bug-in-zotero-8-linux-arm

abaevbog:
This seems to be the case for `<input type="checkbox">` not specific to the citation dialog. Fortunately, there is only one more instance of `<input type="checkbox">`, which is in `runJS.html`

PR #5503, opened 2025-08-19
Fix checkbox inputs not looking right when checked on linux

abaevbog:
And set slightly smaller default border width of buttons.

Post-fx140. Followup to  https://github.com/zotero/zotero/commit/06ef9ab657af0f509d211fa431c962c98e288345.

Fixes: #5501

Before `appearance: none` on inputs would make `<input type="checkbox">` always look the same, no matter if it is checked or not. Also, the default border on buttons looks very wide.

Screenshot 2025-08-19 at 4 40 45 PM

After:
Screenshot 2025-08-19 at 4 41 30 PM

### DIFF
--- a/scss/_zotero.scss
+++ b/scss/_zotero.scss
@@ -55,6 +55,7 @@
 @import "components/rtfScan";
 @import "components/runJS";
 @import "components/search";
+@import "components/select";
 @import "components/selectItemsDialog";
 @import "components/spinner";
 @import "components/syncButtonTooltip";
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -42,8 +42,7 @@ button {
 		// fx140 form element styling fix
 		appearance: none;
 		border-radius: 5px;
-		border-color: ThreeDShadow;
-		border-width: 1px;
+		border: 1px solid ThreeDShadow;
 		padding: 0.3em 1.5em;
 	}
 }
--- a/scss/components/_citationDialog.scss
+++ b/scss/components/_citationDialog.scss
@@ -637,6 +637,9 @@
 							border-radius: 3px;
 							height: 25px;
 						}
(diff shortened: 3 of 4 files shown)


adomasven:
Hmm. The select and button styling seems to be bad in this dialog overall, not just the checkbox. It seems that some of these form elements are XUL, and some - HTML. Can we normalize and fix them all?

abaevbog:
> It seems that some of these form elements are XUL, and some - HTML. Can we normalize and fix them all?

We have a XUL "Done" button so that it can be a native-looking primary button on each platforms.

https://github.com/zotero/zotero/blob/4f1589132c7d6f99d895b7aa9b2d8a5ba5251168/chrome/content/zotero/integration/citationDialog.xhtml#L131-L132

Would it be better to try to replicate the looks of the buttons with platform-specific css instead? Besides that, the only other XUL elements we have here are `panel`s and the `bubble-input`.

adomasven:
> Would it be better to try to replicate the looks of the buttons with platform-specific css instead? Besides that, the only other XUL elements we have here are `panel`s and the `bubble-input`.

That could work, but really, we need CSS rules that work for both HTML and XUL versions of form elements on Linux giving a unified look. I know the XUL button there is blue on macOS, but a default Done/Close button isn't a thing on Windows or Linux, so that's actually a mac-specific customization. Still, HTML and XUL form elements should have a consistent look. As I am Linux with Ubuntu Gnome native, I can take this over if you want.

abaevbog:
Ok, I see what you meant about the consistently looking form components. The border color was different on buttons, and the `<select>` had a weird border and sizing as well. I added a style for the `<select>` component to look closer to `<menulist>` and to fit the popup in the dialog better, as well as tweaked the buttons borders. What do you think about this now?

Screenshot 2025-08-20 at 12 06 12 PM

PR #5505, opened 2025-08-20
Title-case "List All Tabs"

AbeJellinek:


dstillman:
Sentence case is intentional for tooltips. See https://developer.apple.com/design/human-interface-guidelines/offering-help#macOS-visionOS

AbeJellinek:
Then why are all the other toolbar tooltips (Filter Collections, New Item, Underline Text, Find in Document, Toggle Sidebar, ...) title-cased?

AbeJellinek:
I think this is the only sentence-case one we have.

AbeJellinek:
Not that Apple is consistent with this either, I guess:

image

But if sentence case is recommended, should we sentence-case all the other tooltips?

dstillman:
I'm sure it's sometimes just an oversight, but there's a distinction between a description of an action (often as an imperative sentence) and the name of a command. "Edit Tags" is a command, as is "New Tab" in Safari, as are our other ones. But "List all tabs" is a sentence, so it gets sentence case.

AbeJellinek:
In TextEdit and Font Book, Apple uses sentence case for commands, too: "Underline text", "Increase font size". (In Calendar they seem to have forgotten to add tooltips at all.) Not sure I like these guidelines, but they are the guidelines.

PR #5516, opened 2025-08-25
Stop using `restoreProcessorState()` to fix Chinese punctuation

### LINKED ISSUE #5512: Chinese full stop rendered as period after caching citeproc engines
https://github.com/zotero-chinese/styles/issues/551
https://forums.zotero.org/discussion/126148/zotero-7-0-23-chinese-full-stop-rendered-as-english-in-references
https://forums.zotero.org/discussion/126296/bug-chinese-legal-citation-manual-style-footnotes-incorrectly-end-with-a-period-instead-of-a-chines

I can reproduce with the data from https://github.com/zotero-chinese/styles/issues/551#issue-3280190032 and the style from https://forums.zotero.org/discussion/126296/. It was caused by d4a5fbbefe6def44d116bcf016e1d8841c9abca0. It goes away if I revert the changes in `restoreProcessorState()` in integration.js and comment out the `this._cachedEngines.set(cacheKey, citeproc);` line in style.js, but I didn't check beyond that. I assume something about the locale is being cached incorrectly.

AbeJellinek:
`restoreProcessorState()` breaks locale-specific punctuation (and is deprecated).

`rebuildProcessorState()` is enough if we're passing an empty citation list; if we're reinitializing with a new non-empty list, call `updateItems()` first.

Also:
- Add tests for this issue and for potential regressions in disambiguation handling
- Fix `Zotero.Style.install()` still prompting after validation error in silent mode (`handbook-of-legal-citations-zh.csl` has some validation errors)

Fixes #5512

### DIFF
--- a/test/tests/data/handbook-of-legal-citations-zh.csl
+++ b/test/tests/data/handbook-of-legal-citations-zh.csl
@@ -1,0 +1,1724 @@
+<?xml version="1.0" encoding="utf-8"?>
+<style xmlns="http://purl.org/net/xbiblio/csl" class="note" version="1.0" initialize-with=". " demote-non-dropping-particle="sort-only" default-locale="en-US">
+  <info>
+    <title>法学引注手册（多语言，重复引用不省略）</title>
+    <id>https://www.zotero-chinese.com/styles/法学引注手册（多语言，重复引用不省略）</id>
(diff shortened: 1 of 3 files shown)


dstillman:
Merged in 08e4d6b06 and 5c95c35743

PR #5517, opened 2025-08-25
Itembox: restore unsaved input value after refresh

abaevbog:
If `itemBox` refresh happens when an input is focused and has some unsaved changes, restore the value of the input after the refresh.

Potential solution for:
https://forums.zotero.org/discussion/126273/bug-data-that-is-input-is-not-saved-and-needs-to-be-entered-again

---

I have not been able to naturally reproduce the issue from the forums post above. However, I do see the following in the debug log: `Showing editor for publisher ... Showing editor for publisher ... Hiding editor for publisher`.  One `Hiding editor for publisher` is missing. That means that if you have some changes typed in after the first `Showing editor for publisher`, on the second `Showing editor for publisher` (which is refresh and refocusing of the previously focused field), those changes are just discarded. We already restore focus and selection, so perhaps it's also worth trying to restore the value for cases like this.

A simple way I mocked this is by running in the console `setTimeout(() => {document.querySelector("info-box")._renderInternal();console.log("Refresh")}, 5000)` to simulate an unexpected refresh. Then, I would focus the field, type something, wait for the refresh to happen, Tab - and then the edits were gone.

This situation happens, I think, due to the fact that right before the first `Showing editor for publisher`, a new DB transaction is started. Once it finishes, the itemBox is refreshed. So ... maybe that's just what can happen if transactions take a while to run?

### DIFF
--- a/chrome/content/zotero/elements/itemBox.js
+++ b/chrome/content/zotero/elements/itemBox.js
@@ -60,6 +60,7 @@
 			this._initialVisibleCreators = 5;
 			this._draggedCreator = false;
 			this._selectField = null;
+			this._selectFieldValue = null;
 			this._selectFieldSelection = null;
 			this._addCreatorRow = false;
 			this._switchedModeOfCreator = null;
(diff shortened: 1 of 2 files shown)


abaevbog:
@AbeJellinek, what do you think about this? Does this seems like a plausible explanation of what's happening in the linked forums post?

AbeJellinek:
Yeah, this seems good!

abaevbog:
Awesome! Thank you for checking.

PR #5528, opened 2025-09-01
Add note as tab

### LINKED ISSUE #5178: New tab type: note
https://forums.zotero.org/discussion/123130/

To be discussed:

- Should consider making note remote from the very beginning, otherwise the main window will be overloaded
- Left pane: Outline for the note (same as the BN plugin)
- Right pane: item pane with tags and related section. I want to propose again refactoring the UI logic to share only one item pane instance among the library and different tabs and moving the logic of layout (stacked/standard) into Zotero_Tabs for maintainability. For our current logic, every time we create a new tab type, we will need to replicate this part.
- For plugins: APIs to access and inject script into the note editor; APIs to get the outline; APIs to add custom elements to the left pane (the right pane is already managed by ItemPaneManager).
- Something else: since this will be a big change, and it's related to the tabs, maybe we should merge #4961 ?

FYR, the BN's current note tab:

Image

windingwind:
Resolve #5178


windingwind:
So the new tab type of note is ready. It looks like this:


https://github.com/user-attachments/assets/d5a14466-e46e-44fd-bdf3-0e6d9f52ff22

What's changed:

- Unify the tab logic and abstract relevant tab event hooks to be independent from the reader
- Create a new note tab type with support to auto load & unload
- Update the note-editor to match the current design pattern of the context pane
- Add a note info section in the context pane. Added a 3rd-party script (from https://github.com/thecodrr/alfaaz) for word count, which was built by me with esbuild (they don't provide ESModule script). It's open source and the code is relatively simple, mainly only regexes to match word count for different languages.
- Add prefs to control if the notes are opened in tabs
- Add tests for note tabs

Here we'll need some suggestions/assistance:

- @yexingsha Please take a look and see if it's good design-wise. We might need a new icon for the note info section (currently it's reusing the one from the attachment info section)
- For the note info section, I only added the word count and modification/creation time. What else do we want to show there? For example, for group notes, we can show authors; number of images/tables/citations
- The note context is enabled, as people would definitely like the idea of modifying two notes side by side (one in tab and one from search in the note context).
- What else to show in the context pane: cited items; options to export
- What else to show in the left side pane: currently we don't have it. In the plugin BN, it's showing the outline of the note and its variants like mind map.



For note editor, please use https://github.com/windingwind/note-editor/tree/add-note-tab-type and build locally

yexingsha:
This is great! My thoughts so far:

- I made the icons in the format of item/attachment info, but I worry that they look too similar to the actual note icon down below. I think we need to either switch to a different color (maybe accent-blue?) or a different icon (maybe just an info circle, but could be something else). Both options break the logic a little, but maybe a different icon is acceptable. Let me know what you think: [icons.zip](https://github.com/user-attachments/files/22107895/icons.zip)

- I think parent item should be the first row in note info. Clicking the row should take you back to library tab and focus on the item. If there is no parent item, then the row should say "None" and clicking it should take you back to library and focus on the note itself. Not sure if we should enable add/switch parent item in there as well.

- It seems that the note context is also showing up in library tab.

- Cited Items as another section in the context pane sounds good to me.

- If we offer note export, then that's going to be available regardless of whether the note is in a context pane, a tab, or a separate window, right? Then the export button would make more sense in the note editor toolbar, rather than in the context pane.

- In the note editor toolbar, maybe we should switch the position of the search button and the overflow button, so that the position of the search button stays consistent? Either way, it's going to mess with people's muscle memory a little.

- I'm wondering what we should do with the Locate button in the sidenav. Maybe just disable it?

- We should update some context menus. For notes in library tab, we should add "Open in New Tab" and "Open in New Window"; in the overflow menu of the item/context pane note editor, we should add "Edit in a New Tab"; and in the note tab toolbar, maybe "Edit in a Separate Window" should pop out the tab, though I'm less sure about that.

windingwind:
Thanks @yexingsha !

> icon

I think the (i) icon might be better compared to the attachment-info-like icon with a different color. It looks like:

image


> parent item should be the first row in note info

done

> note context is also showing up in library tab

fixed

> Cited Items as another section

Can go to another PR, as this one is already including tons of changes...

> note export

agree with you. let's not do this for now, at least.

> switch the position of the search button and the overflow button

Considering the UI in reader tabs, this might be good to unify the interface. We can discuss if we want to do that after this PR.

> Locate button in the sidenav

We can use `open in window`. Currently, even for reader tabs, they still show `open in tab`, which doesn't feel right. I'll start a new issue for that.

For the note tabs, I would suggest to keep the locate button as we have an API for adding menu to locate popups. Plugins might want to do something here.

> update some context menus

Done. For the note-editor side changes, I try to keep them minimized in this PR so that it would be easier to review... If we want to add the note editor menu items in this PR, I can do that as well.

yexingsha:
Let's display Parent Item in multiple lines (same as Title and Filename in attachment info), but otherwise this looks good to me.

windingwind:
> Parent Item in multiple lines

It's now multiple-line:

image

windingwind:
@dstillman this PR is ready for review and merge.

abaevbog:
This is great! When trying this out, I noticed just a few small things, mainly with keyboard navigation:

1. Tab from the sync button seems to do nothing. We would want to move focus into the note's toolbar.
2. Similarly, Shift-tab from the note's toolbar should focus the sync button. Currently, focus goes to the "Locate" button in the sidenav.
3. If I tab through the context-pane, the value of the parent item (`#parentItem`) is skipped. It should get focused on Tab from the twisty of the collapsible section and on shift-tab from the "Word Count" value.
4. If the note has no parent item and I click on the parent item's value ("None"), errors on the DB level get thrown.
5. Locate menu in the sidenav has a new "Open Note in New Tab" menu item that seemingly does nothing?
6. note-editor-toggle-context-pane button is missing a tooltip when the context pane is collapsed. Currently, it just shows "note-editor-toggle-context-pane" when I hover over.

windingwind:
Thank you @abaevbog ! All addressed.

> Tab from the sync button seems to do nothing.
> Shift-tab from the note's toolbar should focus the sync button.

https://github.com/user-attachments/assets/db5bd08e-c6d8-436a-8613-a7692e28f511

> parent item should get focused on Tab from the twisty of the collapsible section and on shift-tab from the "Word Count" value.

image

> If the note has no parent item and I click on the parent item's value ("None"), errors on the DB level get thrown.

Fixed

> Locate menu in the sidenav has a new "Open Note in New Tab" menu item that seemingly does nothing?

That's happening with the reader tabs before this PR. Should now be fixed.

> note-editor-toggle-context-pane button is missing a tooltip when the context pane is collapsed.

Thanks for finding this out. I updated the locale file.

windingwind:
Based on the #5554 it now has libraries & collections section in the note tab, which looks great:

image

dstillman:
I think sticking to an icon like the attachment-info icon makes sense. Just the (i) looks like some sort of warning to me. Maybe the Notes icon down below needs to have multiple documents?

abaevbog:
The keyboard focus seems to be stuck again in the note editor. As I keep pressing Shift-Tab, it jumps between the toolbar and the actual editor. I think we want Shift-Tab from the toolbar to move focus to the sync button, same as in the reader tab: 

https://github.com/user-attachments/assets/27a00ff2-fe74-4cfe-843f-0a2b981f8e91

windingwind:
re icon: 

> Maybe the Notes icon down below needs to have multiple documents?

That sounds good.

re keyboard stuck:

I cannot reproduce, it's working for me. Maybe check if you're on the latest note-editor submodule (https://github.com/windingwind/note-editor/tree/add-note-tab-type) and if it's actually built?

abaevbog:
> Maybe check if you're on the latest note-editor submodule

Ah, sorry about that - I hadn't updated the `note-editor` submodule. Working as expected now!

yexingsha:
@dstillman re: icons, do you mean something like this? If yes, I think it would make sense for the notes tab in reader sidenav to always use this icon.

Sidenav

note-info-16
note-info-20
notes

Another thing I noticed: when the context pane is toggled off, the "More" button in the editor toolbar is too close to the toggle button. It should be directly under and aligned with the tab menu button up top.

dstillman:
Yes, something like that, but it's a little intense with three. Two? Or three but spaced out more?

And yes, we'd change this icon in all states.

yexingsha:
Spacing the layers out loses the "stack" effect. We can try different configurations with two layers:

notes
notes-1
notes-2

windingwind:
tried the first icon:

image

second:

image

third:

image

dstillman:
Can we see the other two in context as well?

windingwind:
Sorry forgot to update the other two. Updated in the comment above.

windingwind:
Do we have any updates regarding the icon etc.?

windingwind:
Requires submodule change: https://github.com/zotero/note-editor/pull/67

mrtcode:
I recall earlier discussions about centering the note editor with larger margins once the note tab is implemented, and increasing the font size to create a more focused writing experience, similar to those fancy macOS writing apps like Ulysses. Because neither the whitespace on the right nor the extremely long lines are very useful there.

windingwind:
> I recall earlier discussions about centering the note editor with larger margins once the note tab is implemented, and increasing the font size to create a more focused writing experience, similar to those fancy macOS writing apps like Ulysses. Because neither the whitespace on the right nor the extremely long lines are very useful there.

Thanks for the comment. Yes I do remember that discussion happened some time ago.

For the larger font size in note tab, it might be tricky, as we currently explicitly let users decide the font size by number. We need to change it to the same behavior as client font size (bigger/smaller), or allow users to set font size differently for the tab and other places.

For the whitespace, I'm not very sure. Maybe @yexingsha will have some comments here on how the interface should behave in the tab.

yexingsha:
Yes, for a better writing experience, we should limit the line width to 70ch (so that it scales with font size), center the editor area, and offer a separate font size setting for the note tab. 16px should be a good default size.

windingwind:
The font size and editor width are updated: 

image
image

yexingsha:
Screenshot 2025-12-04 at 23 06 20

The scrollbar now appears at the edge of the 70ch text area. It should still be at the edge of the entire editor.

Also, the background color is not covering the entire editor area, which you can see in dark mode.

windingwind:
Oops, sorry. Forgot to check the scroll bar.

Should be fixed now: 
image

windingwind:
The drag handle position is fixed (for both ltr & rtl):


https://github.com/user-attachments/assets/5e53baa5-1676-4846-969a-a05a65ced0fc

yexingsha:
Looks good now!

dstillman:
This seems to be centering text in a new note in the item pane:



If I keep typing, it ends up left-justified, and new paragraphs start on the left.

dstillman:
Oh, it's doing the same thing in a new tab too. The note just no longer has its full width until it has content.

dstillman:
When you open a note in a tab, the note should be focused so that the cursor is at the beginning of the editor, the same as in a new note window.

windingwind:
re layout: fixed

image

windingwind:
re note tab focus: fixed

dstillman:
Do we want to share the second font size pref between tabs and windows? Not sure how we would label the prefs in that case ("Note Tab/Window Font Size" is pretty clumsy), but probably you want the bigger text in the windows as well?

dstillman:
Bug:

1. Open note in tab.
2. Open context pane.
3. Click Note Info button.
4. Click Toggle Context Pane button to close the context pane.
5. Click in editor.

The context pane opens back up (after a slight delay).

windingwind:
> Do we want to share the second font size pref between tabs and windows? Not sure how we would label the prefs in that case ("Note Tab/Window Font Size" is pretty clumsy), but probably you want the bigger text in the windows as well?


We can probably say "standalone editor note size"?

windingwind:
> Bug:

I'm not sure if I can reproduce... you mean the context pane opens itself when you focus the editor?

dstillman:
> you mean the context pane opens itself when you focus the editor?

Actually doesn't require focusing the editor.

https://github.com/user-attachments/assets/8b13570b-d195-4a4a-8207-4788a1491973

windingwind:
Thanks! Now I get it. Should be fixed now.

Rebased on the current main.

dstillman:
I realized a note in a standalone window may become a note tab anyway — with the idea that everything should just be a tab — so "Note Tab Font Size" is probably OK.

dstillman:
(And we should share that pref, but we can do that separately.)

PR #5529, opened 2025-09-03
Citation dialog: updates to locators shortcuts

abaevbog:
- after a new bubble is added to the citation, it is recorded as a just-added bubble. The next locator typed without any additional search query and accepted via Enter will go to that item, instead of going to the bubble right before where the locator was typed.
- just-added bubble is visually marked, so it is more apparent why a locator would go into that bubble.
- the record of just-added bubbles is cleared on `focusout` or` keypress` of an arrow key. That way, it's discarded if the user is almost certainly not intending to immediately type a locator.
- added a special case to recognize a numeric value as a page locator if it is typed right after a bubble is added
- added a special case to recognize ":<number>" as a locator in the same circumstances that "page <number>" is currently recognized

Also, minor refactoring of `cleanSearchQuery` to make it less likely that relevant punctuation (e.g. colon for locators) is removed.

---

In the demo below, I add a page locator to the bubble before the input by typing ":10-15". Then I add a bubble and add a page locator to that newly added bubble by typing a number. Then I type a number 2019 meaning to search by a year - the search does not happen because it is initially treated as a locator. I press ArrowRight, which clears the record of a just-added bubble, and the 2019 is then treated as a search query.

https://github.com/user-attachments/assets/8d55dace-ca8a-4180-853c-f723069da445

Comments:
- I added the visual border around a last-added bubble because I thought it could be helpful to indicate what is happening with the locators. I also found it useful to show which bubbles were just added because it is not always apparent after sorting which bubbles came from the items one clicked on. This is not essential, so if it just makes it busy, we can remove it.
- In QuickFormat dialog, the locators would be added immediately after you type them. Now that we support more locators, including ones that can be quite long (e.g. `chapter "some title of the chapter"`), to apply the locator one needs to press Enter. The same applies to newly added numeric locators for consistency.
- just-added bubbles are cleared if you press arrow keys. This covers navigation through the list of items while the focus remains in the input. I also thought it's an effective to "cancel" the effect of just-added bubble on numeric input without necessarily moving the cursor from the input (as in the demo). We could use something else instead, such as whitespace, for example.
- we briefly discussed adding Cmd/Ctrl-Z to undo a locator going into the bubble and return it into the input. I am not sure how useful it would be with the current setup, given that it's pretty easy to "cancel" the just-added status of a bubble, as in the demo. But let me know if it is something we would still want.

### DIFF
--- a/chrome/content/zotero/elements/bubbleInput.js
+++ b/chrome/content/zotero/elements/bubbleInput.js
@@ -432,7 +432,6 @@
 			// Handle drag-drop of items from the citationDialog into bubble-input to add them
 			if (itemIDs) {
 				itemIDs = itemIDs.split(",");
-				console.log(itemIDs);
 				let newIndex = 0;
 				if (this.dragOver) {
 					newIndex = [...this.bubbleInput.querySelectorAll(".bubble")].findIndex(node => this.dragOver == node);
@@ -587,6 +586,11 @@
 			this.bubbleInput._body.appendChild(span);
 			let spanWidth = span.getBoundingClientRect().width;
 			span.remove();
+			// set min-width of 1px if the input is focused to ensure
+			// that the cursor is always visible
+			if (document.activeElement == input && !spanWidth) {
+				spanWidth = 1;
+			}
 			return spanWidth;
 		},
 
--- a/chrome/content/zotero/integration/citationDialog.js
+++ b/chrome/content/zotero/integration/citationDialog.js
@@ -35,6 +35,7 @@ var Helpers, SearchHandler, PopupsHandler, KeyboardHandler;
 
 const ITEM_LIST_MAX_ITEMS = 50;
 const SEARCH_TIMEOUT = 250;
+var NUMERIC_LOCATOR_TIMEOUT = 500; // exposed to tests
 
 var { CitationDialogHelpers } = ChromeUtils.importESModule('chrome://zotero/content/integration/citationDialog/helpers.mjs');
 var { CitationDialogSearchHandler } = ChromeUtils.importESModule('chrome://zotero/content/integration/citationDialog/searchHandler.mjs');
@@ -356,8 +357,7 @@ class Layout {
 		let timer = new Zotero.Integration.Timer();
 		timer.start();
 		Zotero.debug("Citation Dialog: searching");
-		_id("loading-spinner").setAttribute("status", "animate");
-		_id("accept-button").hidden = true;
+		IOManager._showLoadingSpinner();
 		SearchHandler.searching = true;
 		// search for selected/opened items
 		// only enforce min query length in list mode
@@ -398,8 +398,7 @@ class Layout {
 		}
 
 		SearchHandler.searching = false;
-		_id("loading-spinner").removeAttribute("status");
-		_id("accept-button").hidden = false;
+		IOManager._hideLoadingSpinner();
 		let searchTime = timer.stop();
 		Zotero.debug(`Citation Dialog: searching done in ${searchTime}`);
 		if (this.forceUpdateTablesAfterRefresh && this.type == "library") {
@@ -941,6 +940,7 @@ class ListLayout extends Layout {
 //
 const IOManager = {
 	sectionExpandedStatus: {},
+	_skipInputAcceptOnEnterUntil: 0,
 
 	// most essential IO functionality that is added immediately on load
 	preInit() {
@@ -995,6 +995,11 @@ const IOManager = {
 			if (event.pointerType === "mouse") return;
 			setDialogType(event.target.closest(".option").getAttribute("value"));
 		});
+		// clear record of just-added bubbles to which next locator would go
+		doc.addEventListener("keydown", event => this._clearJustAddedBubbles(event));
+		_id("bubble-input").addEventListener("focusout", event => this._clearJustAddedBubbles(event));
+		// handle cmd/ctrl-z pressed from the input to undo added locator to a just-added bubble
+		_id("bubble-input").addEventListener("keydown", event => this._handleInputUndo(event));
 	},
 
 	// switch between list and library modes
@@ -1099,6 +1104,13 @@ const IOManager = {
 				bubbleItem.locator = locator.locator;
 				bubbleItem.label = locator.label;
 			}
+			// Do not record just-added bubbles if locator is already provided
+			this._clearJustAddedBubbles();
+		}
+		else {
+			// If no locator is provided, record which bubbles were just added.
+			// If a locator is typed next, these bubbles will receive it.
+			this._justAddedBubbles = bubbleItems;
 		}
 		await CitationDataManager.addItems({ bubbleItems, index });
 		// Refresh the itemTree if in library mode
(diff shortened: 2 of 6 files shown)


dstillman:
I think the border is too distracting/confusing. There might be some other effect — some subtle temporary animation? — we could use to show where the new bubble is for sorted multi-item citations (maybe @yexingsha can suggest something), but most of the time 1) there will be a single bubble or 2) people will just be searching for another item and a border on another item would be confusing. Sorted multi-item + typed locator is going to be inherently weird, but I don't think we need to worry about it too much. This is an advanced trick.

abaevbog:
That's fair. Keeping a border when one is searching for the next item may indeed be confusing. We could do a brief highlight for bubbles that were just added, similar to this?

https://github.com/user-attachments/assets/5dae1365-54d9-41b1-81bd-341c40ffb213

But ultimately, my intention was to avoid confusing moments that occurred a few times with QuickFormat when it wasn't clear where a newly typed locator would go. In the dialog now (last example in the recording), when I type "2019" after adding a bubble, it's not apparent why the search is not running. The border around the bubble is at least some explanation as to why. And when an arrow is pressed, the border disappears and the search runs. If we only do a brief highlight, that would not help with this scenario - you would type "2019" and nothing would tell you why nothing happens.

This is a concern only for purely numeric queries - if you type a proper locator (e.g. "p. 50-100", or even shorter ":50-10"), you are almost certainly not searching. Maybe the way to address this is to just make the numeric regex stricter? It currently allows whitespaces and punctuation. What if we stop counting it as a locator if there is more than 1 consecutive whitespace, or if a whitespace does not follow a comma, period, or semicolon? Then, in the scenario here, I suspect, one would type "2019", nothing would happen, and then they would press space (that's what I instinctively do if something that should happen on input did not happen), after which "2019 " would not count as a locator and search would run.

Does the reasoning above make sense? Because if it sounds like a way to address the "2019" locator/query situation, we may not need any highlighting at all.

yexingsha:
> Then, in the scenario here, I suspect, one would type "2019", nothing would happen, and then they would press space (that's what I instinctively do if something that should happen on input did not happen), after which "2019 " would not count as a locator and search would run.

I think making the numeric regex stricter is good, but I'm not sure everyone would press space here; presumably a lot of people will just think something is broken, and try a different search or browse the library. But maybe we can just... not allow pure numeric queries? I don't know how many people that's going to affect, but probably fewer than the number of people confused by why searching for "2019" doesn't work.

If that's too disruptive, then here are some ideas for indicating the last added bubble. We can put a "second caret" at the end of the text, either a triangle or an underline:

triangle
underline

To minimize distraction, I think they should only show up when you are typing a locator.

abaevbog:
> but I'm not sure everyone would press space here; 

That's a good question for people's usage patterns. Maybe my approach of trying to mash different buttons isn't universal... Sometimes it does help though!

> But maybe we can just... not allow pure numeric queries? I don't know how many people that's going to affect, but probably fewer than the number of people confused by why searching for "2019" doesn't work.

I did not include them during the big citation dialog redo. But the issue there is that it's something that had been a part of QF for a long time, and it seems that some people are quite used to it. Then, when they try to do it with the new dialog, from their perspective, it doesn't work right. At the same time, I agree, we should be careful to not interfere with searching, which is ultimately the main feature here.

> If that's too disruptive, then here are some ideas for indicating the last added bubble. We can put a "second caret" at the end of the text, either a triangle or an underline

I like these! This gets the message across. I wonder if this is too subtle though? Once you know what it is, it makes sense. But if you don't, I feel like many would not even notice?

---

I am beginning to wonder if how we treat "2019" is something that could go into a pref? Then, it would not affect the majority of people but those few who rely on it as part of their workflow can still enable it? It would also allow us to not have to worry about visual indication at all: if you don't have the pref enabled, it does not apply but if you do, you already know what's going on.

yexingsha:
> I wonder if this is too subtle though? Once you know what it is, it makes sense. But if you don't, I feel like many would not even notice?

The triangle can be a little larger, but the underline can't be longer without stretching the gap between the text and the dropdown icon. We can try giving them a soft blinking animation to see if that helps.

> I am beginning to wonder if how we treat "2019" is something that could go into a pref? Then, it would not affect the majority of people but those few who rely on it as part of their workflow can still enable it?

I suppose with this pref, pure numeric queries would be off by default? Then it might have a little less impact than disabling "2019" altogether, depending on how many people would think of checking the prefs.

abaevbog:
> I suppose with this pref, pure numeric queries would be off by default? Then it might have a little less impact than disabling "2019" altogether, depending on how many people would think of checking the prefs.

Yeah, that's what I am thinking. Those who don't know are not affected but those who do can enable it if it matters for them.

One more alternative, of course, is to go back to how it was in QF and add locator to the last-added bubble(s) without Enter. Once you type a number, it will be counted as a locator after a second or so automatically. I don't personally think it would fit here, since with other longer locators Enter is somewhat necessary, so it would not be consistent. It also would still get in the way of trying to search for "2019" after adding a bubble - but at least it's clearer why.

This is from the current stable build:

https://github.com/user-attachments/assets/24e80c85-b3ca-4399-86ce-d3d9f5367066

dstillman:
No pref! A pref is an abdication of responsibility!

I think the little indicators are too distracting and confusing, for much the same reason that the yellow border was. (The indicators might be even more confusing. The border at least made sense just as an indicator of where the last-created bubble went. People won't understand why there's a random mark within that bubble.)

I remain comfortable with privileging page numbers, since that's what we supported for so many years, so I think it would be OK to continue automatically adding page numbers to the last bubble as you typed them. Yes, it would be inconsistent with other locators, where you'd have to 1) understand that they would apply to the last-created bubble and 2) press Enter after, but the fact that page numbers were automatically transferred to the bubble would also somewhat convey the concept of a last-created bubble that locators could be added to.

I also think this is where we could use Cmd/Ctrl-Z to remove the automatically transferred numbers from the last bubble, move them back to the search query, and disable automatic page number transfer for that query. So if you really wanted to search for a subsequent item and begin with a number, there'd be a way.

Ultimately, you'd probably have to read the documentation to know how to use this hack for other locators or search for things beginning with a number, but that's fine. We shouldn't break automatic page number transfer to accommodate much rarer cases.

abaevbog:
> No pref! A pref is an abdication of responsibility!

Understood! That is a good way to put it, I'll keep that in mind for the future.

I re-pushed with the updated behavior. Now, typing a number locator after adding a bubble will append those values as a page locator to that bubble, same as it was done in QF. One can press Cmd/Ctrl + Z to clear that locator and move that number string back into the input. Then, just-added bubble will be forgotten, same way as it is forgotten on `focusout` or an arrow keypress. (Clearing just-added bubble on an arrow keypress is no longer as important as during the previous setup but perhaps we should still keep it for edge cases like if one presses arrowDown to navigate the list of items in list mode. Then, I think we can be quite certain that whatever will be typed next is meant to be a search query.)

https://github.com/user-attachments/assets/fae83192-8b78-4a6c-a8f3-5b54ea4bb82d


---

One additional small change worth noting here is that [in QF](https://github.com/zotero/zotero/blob/54b6216ecc69f246ca37bee2486e3323fe2ef543/chrome/content/zotero/integration/quickFormat.js#L508-L514) and in [the current version of the dialog](https://github.com/zotero/zotero/blob/e896fd0137c267fb6dd738d94cf5f3f4f9971997/chrome/content/zotero/integration/citationDialog/searchHandler.mjs#L384-L393), there is some processing of the numeric value, which splits the potential year range (e.g. "2010 - 2020") and keeps only the first part ("2010"). I always found it a bit confusing but carried it over from QF for consistency. I ran into a few moments where it somewhat conflicted with the locator logic above, so I removed it. Is there any reason not to? My thinking is that if one types "2010 - 2020", that is what should be searching for?

dstillman:
Can we add a short debounce on the transfer, like we had before? That way you see what you're typing, instead of their appearing in another place.

Obviously, if you keep typing and fail the regexp before the delay triggers, it should cancel the transfer.

abaevbog:
Totally, how does this look? I thought that adding a locator immediately would be more responsive but I guess it is better to give one a moment to finish typing. I set the debounce to half of a second (vs 250 ms for search) so that it's more likely that the typing is done. If after this debounce the input is no longer a numeric locator, nothing happens with the locator, search runs, and just-added bubbles are cleared.

I added a spinner during debounce as well, just to indicate that something is happening. 


https://github.com/user-attachments/assets/b90f6a7e-ca20-4843-9702-792c2bd75bd6

dstillman:
This is good, but if I press Enter while the spinner is spinning, it cancels the transfer and the number can no longer be added. If it's a valid numeric page number that would be automatically transferred, pressing Enter should add it automatically, so you don't have to wait 1/2 a second if you don't want to.

abaevbog:
>  If it's a valid numeric page number that would be automatically transferred, pressing Enter should add it automatically, so you don't have to wait 1/2 a second if you don't want to.

I see what you mean. I added an extra step to where we handle Enter keypresses to run the processing of the potential numeric locator without debounce for such cases. Does this feel better now?

--- 

While testing this, I did run into the following edge case: if I press Enter intending to confirm the numeric locator after a small delay (but while spinner is still on), it might land right after the locator is transferred to the bubble via debounced input handling. In that case, there would be no numeric locator when `_handleInputEnter` runs, so it would just be treated as an Enter keypress on an empty input and accept the dialog, despite my intention. It happened to me a few times, so I feel like it could happen often to those who know of this trick. The best solution I could come up with here is to ignore Enter keypresses that would otherwise accept the dialog after the numeric locator is transferred to the bubble for 500ms. Does that seem reasonable?

dstillman:
> The best solution I could come up with here is to ignore Enter keypresses that would otherwise accept the dialog after the numeric locator is transferred to the bubble for 500ms. Does that seem reasonable?

Yeah, I think that's good.

abaevbog:
Rebased! As a brief summary, these are the main updates here:
- after a bubble is added without a locator, it is recorded as a just-added bubble. It receives the next submitted locator. 
- just-added bubble record is cleared once the user moves focus anywhere or adds a locator to it by typing and pressing Enter
- special case: treat ":{number}" the same way as "p{number}" in all instances. ":10-15" will add a page locator after the bubble or as a part of the search query.
- special case: if just-added bubbles record is not empty, typing a number will append a page locator with that value to just-added bubble after a small debounce. This does not clear just-added record, so one can type "1", followed by "2", which will add "12" as a page locator. cmd-Z will clear numeric page locators and return that numeric value to the input, which will become a search term.

PR #5544, opened 2025-09-15
Fix menuManager removedkeys.join is not function on plugin shundown

northword:
Fix follow error:

7fc209df31f5fd6ef9d9cb23e3189820


northword:
Fixed in https://github.com/zotero/zotero/commit/913d15632b8dab14e53344279bb4584702feb08b .

@windingwind does jsdoc of `PluginAPIBase._unregisterByPluginID` need update? => https://github.com/zotero/zotero/pull/5544/commits/904788c498c501f65d17219dda3678bd2dfa6712

PR #5552, opened 2025-09-19
Insert note dialog: match notes by parent item

### LINKED ISSUE #2812: Allow searching for parent item title/creator/year in Add Note dialog
https://forums.zotero.org/discussion/99611/inserting-notes-in-word-how-the-search-bar-works

And show all of the item's notes. (Relatedly, we should show the parent item title as well, similar to how we show it in the PDF reader notes pane.)

abaevbog:
- allow to filter notes by their parent item
- display notes in list mode nested under their parent item
- in library mode, add parent title to the item card

Fixes #2812

Screenshot 2025-09-19 at 12 23 13 PM
Screenshot 2025-09-19 at 12 23 32 PM

### DIFF
--- a/chrome/content/zotero/integration/citationDialog.js
+++ b/chrome/content/zotero/integration/citationDialog.js
@@ -25,6 +25,9 @@
 
 
 const CollectionViewItemTree = require('zotero/collectionViewItemTree');
+const React = require('react');
+const ReactDOM = require('react-dom');
+const VirtualizedTable = require('components/virtualized-table');
 const { getCSSIcon } = require('components/icons');
 const { COLUMNS } = require('zotero/itemTreeColumns');
 var doc, io, ioReadyPromise, ioIsReady, accepted;
@@ -47,10 +50,12 @@ var { CitationDialogKeyboardHandler } = ChromeUtils.importESModule('chrome://zot
 var DIALOG_STATE = {
 	type: null, // 'citation' or 'add-note'
 	loaded: false,
+	isTestRun: false,
 
 	isCitingItems: () => DIALOG_STATE.type == 'citation',
 	isAddingNote: () => DIALOG_STATE.type == 'add-note',
 	isAddingAnnotations: () => DIALOG_STATE.type == 'annotations',
+	getCurrentLayoutType: () => currentLayout?.type,
 };
 
 
@@ -75,7 +80,7 @@ async function onLoad() {
 	let timer = new Zotero.Integration.Timer();
 	timer.start();
 
-	Helpers = new CitationDialogHelpers({ doc, io });
+	Helpers = new CitationDialogHelpers({ doc, dialogState: DIALOG_STATE });
 	SearchHandler = new CitationDialogSearchHandler({ dialogState: DIALOG_STATE, io });
 	PopupsHandler = new CitationDialogPopupsHandler({ doc, dialogState: DIALOG_STATE });
 	KeyboardHandler = new CitationDialogKeyboardHandler({ doc });
@@ -123,6 +128,7 @@ async function onLoad() {
 	// init library layout after bubble input is built since bubble-input's height is a factor
 	// determining initial library layout height
 	await libraryLayout.init();
+	await listLayout.init();
 	// fetch selected items so they are known
 	// before refreshing items list after dialog mode setting
 	await SearchHandler.refreshSelectedAndOpenItems();
@@ -263,7 +269,7 @@ async function setDialogType(type) {
 	}
 
 	// Set proper settings availability depending on the type
-	_id("bubble-input").sortable = DIALOG_STATE.isCitingItems();
+	_id("bubble-input").sortable = DIALOG_STATE.isCitingItems() || DIALOG_STATE.isAddingNote();
 	_id("keepSorted").disabled = !io.sortable || !DIALOG_STATE.isCitingItems();
 	_id("keepSorted").checked = !_id("keepSorted").disabled && !io.citation.properties.unsorted;
 	if (DIALOG_STATE.isCitingItems()) {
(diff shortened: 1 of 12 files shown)


abaevbog:
This adds the suggestions outlined in https://github.com/zotero/zotero/issues/2812#issuecomment-2895999678 except for displaying note titles across all columns (same as with annotations). I was concerned that it would be inconsistent with the way note rows are rendered in itemTrees of all other windows. Unless we want to just display note title rows across all columns everywhere? Notes don't have any additional information to display besides title (and, trivially, item type), do they?

This would certainly have some conflicts with the insert-annotations PR, so it would probably need to happen after. But let me know if in general this is the behavior we are after, @adomasven @yexingsha.

yexingsha:
Yes, this looks like a good start!

As for displaying note in item trees, I'm thinking of something similar to annotations: title in regular font size, followed by content in smaller font size, spanning across all columns. This will apply to all item trees everywhere. The only other relevant note fields I can think of are Date Added and Date Modified, but they are probably less useful than simply displaying more note content.

adomasven:
Date Added and Modified are actually probably important enough to keep, especially for standalone notes where that's their only other identifier besides note content.

abaevbog:
> Date Added and Modified are actually probably important enough to keep, especially for standalone notes where that's their only other identifier besides note content.

That's fair. In that case, I don't think the note title could span all columns, right? I do agree that it is a bit challenging to navigate notes based on just a single narrow(ish) title column. If one has more than a single note, it's hard to tell which one is which.

What if we take inspiration from the insert-annotations PR and add a read only note editor as a side pane? Then, if a note is selected, we could render a preview there? Or is it too much?

yexingsha:
A read only note editor side pane sounds good to me! I don't think it would be too much, especially since we are using the side pane for annotations anyway.

abaevbog:
Cool! Something like that?

Screenshot 2025-09-22 at 4 48 43 PM
Screenshot 2025-09-22 at 4 48 48 PM

yexingsha:
Yes, with smaller margins. 12px should be about right.

abaevbog:
I see what you mean, the default padding around text of note-editor is indeed a bit too large. Since note-editor is an iframe, and we want to add this padding change just in this specific case, I think patching the stylesheet by just inserting a `<style>` node is reasonable (vs changing the css that goes in the actual iframe).

This is with padding of 12px:

Screenshot 2025-09-22 at 11 17 53 PM

yexingsha:
Looks good! Some thoughts as I tested this a bit more:

- Maybe we can use the default font size in the note editor, instead of the note font size (which tends to be bigger), so that a bit more text can be displayed per row.
- We can also get rid of the hover effect on citations in the note editor as well, since it doesn't do anything here.
- We should replicate the Selected Items & Open Documents behavior of Add Annotations: selected or open items are displayed if they have child notes, selected notes and selected/open items are put in different sections in library mode with different cards, etc. The only difference I can think of right now is that selected/open parent items should be fully expanded in list mode.
- In list mode, the second line of parent item is now misaligned with the first line.

abaevbog:
Thank you for the notes! The style sheet and font issues should be fixed now.

I also pushed the refactoring of opened and selected items.  In list mode, selected and opened items sections now include items that have child notes. In library mode, selected items are broken into "selected top-level items" and " selected notes" sections. Clicking on the "selected note" will immediately insert it. Clicking on a "selected top-level item" or an opened item card will select it in the itemTree, same as in the "Insert Annotations" PR.

How does this behavior look?

Screenshot 2025-09-25 at 2 36 35 PM

Screenshot 2025-09-25 at 2 36 30 PM

yexingsha:
I think we can slightly increase the width of the note preview area, but otherwise this looks good!

abaevbog:
Sure, no reason not to make it a bit wider. I bumped the width to 250px:

Screenshot 2025-09-26 at 12 10 33 PM

abaevbog:
@yexingsha, I rebased this PR that we worked on a while back, now that "Add annotations" is merged. It also has an additional tweak: you can now select multiple notes as bubble before inserting them all together. Bubbles can be reordered, same as with citations. Do you mind having another look to make sure it looks right?

(In this PR I also refactored list mode to use virtualized-table instead of sharing DOM nodes generation with library mode. Now that we want collapsible item containers with child nodes, that sharing no longer really works. Also virtualized-table will be necessary anyways for potential future list mode of "Add annotation" dialogs , e.g. for mobile, so I thought might as well do it now. All of that is to say that the way list mode works is completely different here but I tried to preserve the look).

yexingsha:
The bubble note input looks pretty good to me! Just a few side issues:

- In the note dropdown popover, there needs to be a 4px gap between the item name and the note content, and a 16px gap between the note content and the buttons.
- I don't remember if/when we decided to not show anything in the list mode when there is no selected and open items; currently the missing result section looks pretty weird. This situation can happen in the citation dialog as well, but it will be quite common in the insert note dialog when none of the selected and open items have notes. I vaguely remember an iteration where we had some placeholder text there for this situation? But it might be better if we show the most recently modified/read notes/items instead. We can even consider adding a "recently modified/read" as a persisting section in the list mode, so that it will always show a number of suggestions when there is no active search.
- There is a resizing issue where if the bubbles occupy one line in library mode, but wrap to two lines when switching to list mode, the height calculation will be off and the bottom area will be cut off:

Screenshot 2026-04-17 at 16 12 32

abaevbog:
Thank you for the notes! The popup spacing and the window resizing should be fixed now. When we shrink the window from library to list mode, we resize the window diagonally, so it's hard to predict on the fly if the `bubbleInput` will become taller if bubbles overflow to the second row or not. But we already run window resizing twice when the dialog mode is toggled, so the second run will stretch the window if it appears too small.

> I don't remember if/when we decided to not show anything in the list mode when there is no selected and open items; currently the missing result section looks pretty weird.

Yeah, it's just an empty looking dialog now. We could add a placeholder saying something similar to what we show now in library mode for suggested items, e.g. "No notes match the current query" during search and "No selected/open notes" when search is empty?

> But it might be better if we show the most recently modified/read notes/items instead. We can even consider adding a "recently modified/read" as a persisting section in the list mode, so that it will always show a number of suggestions when there is no active search.

"Recently read" for notes is somewhat hard to define since you could "read" notes in the itemPane preview without opening them in a tab or window. But then counting any note as "read" if you ever selected its row in itemTree doesn't seem right either. 

Also, after a period of time without reading or modifying any notes, the "Recently read" or "Recently modified" would be empty, so we could still end up with the empty dialog window. 

Let me try to plug in the placeholder and we can see if we need anything else after.

abaevbog:
With the added "No results" message in list mode, this is what the dialog looks like now when there's nothing to show. What do you think?

Screenshot 2026-04-24 at 12 33 44 PM
Screenshot 2026-04-24 at 12 32 28 PM

yexingsha:
I think it works, though the wording should be consistent with library mode ("No selected or open notes" versus "No notes are selected"). And when the dialog reloads results (e.g. when switching focus from a different window back to the dialog), the message should stay displayed, just like in library mode.

abaevbog:
> ("No selected or open notes" versus "No notes are selected")

Thanks for pointing that out. Fixed, we now have "No selected or open notes" and "No selected or open items" everywhere.

> And when the dialog reloads results (e.g. when switching focus from a different window back to the dialog), the message should stay displayed, just like in library mode.

Oh yeah, that was because on refocus we would rerun search which refreshes items list twice: one after quick refresh of selected/open items, and the second after a potentially longer refresh of library search results. The message generally is rendered after second rerun to not flash the message mid-search. But in the case of refocus, we'll render the message on the first run too, which eliminates the blinking.

PR #5556, opened 2025-09-26
Move quick copy to cite tab in preferences

### LINKED ISSUE #5555: Move Quick Copy preferences into Cite tab
Per https://github.com/zotero/zotero/issues/5511#issuecomment-3325651871

abaevbog:
- Move Quick Copy from Export tab to Cite tab in Preferences
- Delete Export tab
- Leave JS logic related to quick copy unchanged but rename `Zotero_Preferences.Export` to `Zotero_Preferences.QuickCopy`

Fixes: #5555

Screenshot 2025-09-25 at 5 19 36 PM


dstillman:
Probably should make Quick Copy a separate section, like Word Processors?

abaevbog:
Yeah, we can make it a separate `main-section`. I think then we have to move Quick Copy below other citation-related prefs (Citation dialog mode, Citation options and Tools) because otherwise they would be floating between the "Word Processors" and "Quick Copy" main sections without a top-level header.

How does this look now?

https://github.com/user-attachments/assets/de7e2bb6-6a6d-4144-817f-a21e5a9dba8c

We now don't have the benefit of the having the quick copy style dropdown right next to the style table but it's still in the same tab. If we want to keep Quick Copy and Style Manager next to each other, we could wrap Citation dialog mode, Citation options and Tools into their own `main-section` - just what would the title be? I can't think of anything better than "Other".

For reference, this is if we just make the Quick Copy a section right below the Style Table, and "Citation dialog mode", "Citation options" and "Tools" don't have the top-level section label:


https://github.com/user-attachments/assets/92d485da-633d-4c14-b96b-2933122a07c2

dstillman:
Oh, sorry, I just remembered why we had Quick Copy under Export — because it supports export formats as well as bibliography formats.

So now that there are no other Export settings, maybe Quick Copy just needs to be its own top-level tab? We don't get the benefit of the two drop-downs, but at least it will be easier to find.

abaevbog:
Oh, I see. That explains it.

Is there any chance that at some point in the future we would want to add some additional export-related preferences? QuickCopy is super specific. If we do keep it as a separate tab, maybe there's some value in keeping the tab's name somewhat general, so that we can use it for additional prefs in the future?

PR #5560, opened 2025-10-02
Fix: Resolve 500 error on single item export via local API

umpa-loomp:
This pull request fixes a `500 Internal Server Error` that occurs when exporting a single item via the local API using any format that relies on translators (e.g., `?format=bibtex`, `?format=csljson`).

The root cause of the issue is a `TypeError: items.slice is not a function` within the `exportItems` function in `server_localAPI.js`.

This happens because when a single item is requested (e.g., `/api/users/0/items/ITEM_KEY`), the function receives a single Zotero.Item object. However, the code attempts to call `.slice()` on it, which is an array method, causing the server to crash and return a 500 error.

The solution is to ensure that the `items` variable within the `exportItems` function is always an array. This is achieved by adding a simple check: if the input is not an array, it is wrapped in a new array containing that single item.

This makes the function robust and capable of handling both single-item objects and arrays of items, aligning it with the behavior of the similar `citeprocToHTML` function.

I have thoroughly tested this fix by:
1.  Checking out the `main` branch and successfully reproducing the `500 Internal Server Error` for a single item export.
2.  Checking out this branch (`fix-local-api-export`).
3.  Confirming that the same request now returns a `200 OK` status with the correctly formatted data for both `?format=bibtex` and `?format=csljson`.

This fix resolves the bug and makes the local API more reliable for single-item exports.

### DIFF
--- a/chrome/content/zotero/xpcom/server/server_localAPI.js
+++ b/chrome/content/zotero/xpcom/server/server_localAPI.js
@@ -978,14 +978,14 @@ async function citeprocToHTML(itemOrItems, searchParams, asCitationList) {
 /**
  * Export items to a string with the given translator.
  *
- * @param {Zotero.Item[]} items
+ * @param {Zotero.Item|Zotero.Item[]} itemOrItems
  * @param {String} translatorID
  * @returns {Promise<String>}
  */
-function exportItems(items, translatorID) {
-	if (!Array.isArray(items)) {
-   	 items = [items];
- 	 }
+function exportItems(itemOrItems, translatorID) {
+	let items = Array.isArray(itemOrItems)
+		? itemOrItems
+		: [itemOrItems];
 	return new Promise((resolve, reject) => {
 		let translation = new Zotero.Translate.Export();
 		translation.setItems(items.slice());


chrome/content/zotero/xpcom/server/server_localAPI.js
```diff
@@ -983,6 +983,9 @@ async function citeprocToHTML(itemOrItems, searchParams, asCitationList) {
  * @returns {Promise<String>}
  */
 function exportItems(items, translatorID) {
+	if (!Array.isArray(items)) {
+   	 items = [items];
+ 	 }
```
dstillman, line 988: Looks like a mix of tabs and spaces here, but this should probably get fixed where `exportItems` is actually being called (presumably [here](https://github.com/zotero/zotero/blob/fefad8a9e34a8f5e7d76fc125fe3175a8ab8d5ab/chrome/content/zotero/xpcom/server/server_localAPI.js#L365)) or where the data is initially set to a single object instead of an array. I'll let @AbeJellinek comment on that.

AbeJellinek:
Thanks! This looks good.

Issue #5573, opened 2025-10-09
Collapsed tag selector not loading after restart


abaevbog:
1. Collapse tag selector
2. Close Zotero
3. Open Zotero
4. Expand tag selector

It is stuck in "Loading Tags..." state and "can't access property "setMode", ZoteroPane_Local.tagSelector is undefined" error is logged.

<img width="258" height="374" alt="Image" src="https://github.com/user-attachments/assets/ae1d11de-d47c-4a3f-bdd6-7182c3b2d0c4" />

abaevbog:
It looks like this may be causing CI test failures every now and then: https://github.com/zotero/zotero/actions/runs/18379943909/job/52363684943.

I never saw that error during CI test runs before so it must be something relatively new. https://github.com/zotero/zotero/commit/2e3deb0a215e7d27febde90084773836be325e32 changed how tag removals are saved from the tags box - this should have nothing to do with the tag selector but it's the most relevant change from the recent commits.

PR #5574, opened 2025-10-09
Ensure tag selector is initialized before using

abaevbog:
If the tag selector is not initialized in `setScope`, initialize it before proceeding. This fixes the tag selector being stuck in loading state after expanding.

Fixes: #5573


abaevbog:
Even though this fixes the tag selector error, the CI failures must be caused by something else. After a few reruns, I got a test failure here seemingly because an error popup appeared: https://github.com/zotero/zotero/actions/runs/18386393140/job/52395986684?pr=5574#step:13:542

PR #5576, opened 2025-10-10
Improve attachments section test

windingwind:
Cancel pending discard to avoid race condition
Cancel pending render/discard after tests

This should fix the unstable `should discard attachments pane preview after becoming invisible` test.

After extensive debugging, it is clear that the failure is due to render tasks sometimes being queued by the previous test. These tasks can execute after the previous test finishes, updating the preview unexpectedly. When the current test attempts to discard the preview, the condition check finds that the preview is newer than the one it was asked to discard. As a result, the discard operation is skipped and the test fails.

### DIFF
--- a/chrome/content/zotero/elements/attachmentBox.js
+++ b/chrome/content/zotero/elements/attachmentBox.js
@@ -93,6 +93,10 @@
 
 		_previewDiscarded = false;
 
+		_discardTimeoutID = null;
+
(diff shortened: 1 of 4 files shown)


windingwind:
I ran the tests with this fix 15 times (mainly in https://github.com/windingwind/zotero/actions/runs/18378970833) and none of the failures were caused by this test, which I assume indicates that the test can now be passed reliably.

PR #5577, opened 2025-10-10
Enhance copying between libraries and linked collections

### LINKED ISSUE #3045: Make it easier to copy an entire library to another library
https://forums.zotero.org/discussion/103877/how-to-copy-my-entire-zotero-library-to-group-zotero-library

Since you can't select multiple collections, it's pretty annoying to copy an entire library to another library. We've talked about supporting multi-collection selection, but even if we have that there should still just be a command to copy an entire library via the library context menu.

abaevbog:
- allow one to copy a linked collection of one library into another linked collection after the first copy. This will add new subcollections and items (including child attachments, notes, and annotations) into the linked collection without editing already linked items. This is a fairly safe options that one can use regularly to sync most of items between collections without risking that any data is going to be changed or lost.
- provide a new option to "replicate" a collection in another library that will not only copy new items
and subcollections but also update metadata of items, rearrange subcollections to match those in the source library and delete items (including children) and collections that are not linked. It allows one to essentially sync the state of one collection with another. This is a destructive operation that has to be confirmed via a hard confirmation dialog.
- when a linked collection exists, context menu of collection tree "Copy to" will only show the menu of that linked collection. It has two menuitem options - to copy collection or to fully replicate it
- allow one to copy a library into another library, as well as to replicate it, same way as with collections above
- when a collection is selected, highlight the linked collections in other libraries with faint blue, so that there is some indication which collections are linked
- refactoring to move the logic of copying collections/items between libraries from collectionTree into dedicated methods in Zotero.Libraries, Zotero.Collections and Zotero.Items.

Fixes: #3045

---
Marking it as a draft to allow for additional testing and to write additional tests.

Copy collection:

https://github.com/user-attachments/assets/dd5093e6-85f2-493b-923d-3df2393d1676

Copy library:

https://github.com/user-attachments/assets/d117b5da-7be2-414a-9a82-b5699b575c60

Replicate collection:

https://github.com/user-attachments/assets/d4f240bc-39c9-4e19-aa17-cb33ec3f6170

Replicate library:

https://github.com/user-attachments/assets/5660a513-4aab-429c-91e1-5291b348b86b

### DIFF
--- a/chrome/content/zotero/collectionTree.jsx
+++ b/chrome/content/zotero/collectionTree.jsx
@@ -112,6 +112,7 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 		this._expandedRowsOnDrag = new Set();
 		this._expandRowOnHoverTimer = null;
 		this._collapseExpandedRowsTimer = null;
+		this._linkedToSelected = [];
 		
 		this.onLoad = this.createEventBinding('load', true, true);
 	}
@@ -231,6 +232,9 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 				}
 			}
 		}
+		// when selection changes, highlight collections linked to the one currently selected
+		this.highlightLinkedCollections();
+
 		if (shouldDebounce) {
 			this._onSelectionChangeDebounced();
 		}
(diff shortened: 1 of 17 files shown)


abaevbog:
@dstillman, this adds two possible ways of copying collections and libraries that you mentioned in https://github.com/zotero/zotero/issues/3045#issuecomment-3329484911. 

There is a "safe" way to copy that will just add new items and collections without moving existing collections, deleting or modifying data. It's something one can do on a regular basis without being worried about accidentally overriding or deleting anything.

And then the is a "destructive" way to copy (that I named "replicate" for now) that aims to replicate the state of selected collection or library in the linked collection (or target group). It includes moving collections, updating items' metadata, removing top-level items not in the collection and trashing child items, if their linked item is not there. This action has to be confirmed with a hard confirmation dialog.

I'll spend some time on testing this out but I wanted to check in if that's the functionality/interface you had in mind? The screen recordings of 4 main use cases are above.

abaevbog:
I think now this is behaving as expected and the tests should cover all the use cases I could think of

Issue #5582, opened 2025-10-15
Citation dialog: allow to switch between modes from within the dialog


abaevbog:
We should allow one to switch between the "Add Note" and "Add/Edit Citation" (and "Insert Annotations" later) from within the citation dialog. There is a frequent confusion when one clicks on "Add Note", meaning to click "Add/Edit Citation" and then can't find the item to cite. For example: https://forums.zotero.org/discussion/127350/zotero-not-allowing-citation-to-be-placed-in-paper.

In the new redesigned dialog, it is still not explicitly clear which mode the dialog is opened in, so we should make it clearer and allow one to "recover" from selecting the wrong dialog by changing the mode from within the dialog. 

Probably the best place for it would be to add a prominent dropdown in the bottom left corner. Something like this:
<img width="893" height="501" alt="Image" src="https://github.com/user-attachments/assets/b8d365ba-bf94-4c1a-8535-70fcae9ae4b0" />

We can also add coloring to some of the components to make it more visually apparent. In the stable build, "Add Note" dialog is yellow, and "Add/Edit Citation" is red. We could add similar coloring to the bottom area maybe? Alternatively, we can add the newly designed icons that will go into the word processor toolbar https://github.com/zotero/zotero/issues/5360 into the bottom left corner. Having the same icon as in the word processor can get the message across and I think would look pretty good.

dstillman:
I don't think it should be a drop-down. I think there should just be three icons for the different modes — an item icon, a note icon, and an annotation icon — in some sort of tab component. I don't think these should be the same as the icons for the word processor toolbar — those are designed to fit into those apps, not Zotero. I think these could really just be our usual item, note, and annotation icons, because that's what you're inserting.

And then I think we also might want to consider a text label on the right — e.g., "Switch to Library Mode" or a [Library | List] toggle button. I worry that people aren't going to find the current button, particularly people who would've been inclined to use Classic Mode before, and it would also reduce the number of random icons along the bottom. But we can consider that separately.

yexingsha:
@dstillman is this what you have in mind?

<img width="896" height="592" alt="Image" src="https://github.com/user-attachments/assets/8bbd8ccd-38ef-4893-a73d-bad51123fea3" />

[...]

PR #5586, opened 2025-10-17
Citation dialog: allow to switch between "Add Note" and "Add/Edit Citation"

abaevbog:
Allow one to switch between "Add/Edit Citation" and "Add Note" modes in the citation dialog via segmented control in the bottom left corner.

One cannot switch modes when editing an existing citation or when inserting a citation in the note-editor.

Use the same segmented control UI to switch between list and library modes for clarity.

Collect some of the state variables into DIALOG_STATE object for better organization.

Fixes: #5582

https://github.com/user-attachments/assets/a807044d-2bcd-4725-a765-b0f3dfc8bb23

### DIFF
--- a/chrome/locale/en-US/zotero/integration.ftl
+++ b/chrome/locale/en-US/zotero/integration.ftl
@@ -39,7 +39,7 @@ integration-citationDialog-lib-message-citation = { $search ->
 }
 integration-citationDialog-lib-message-add-note = { $search ->
     [true] No selected notes match the current search
-   *[other] No notes are selected.
+   *[other] No notes are selected
 }
 integration-citationDialog-settings-keepSorted = Keep sources sorted
 integration-citationDialog-btn-settings =


abaevbog:
@yexingsha, how is this looking to you? I may have gotten lost in figma because I couldn't find your mockup from the linked issues. So I went with the colors/sizing just based on the pictures - correct me if anything looks off.

We obviously will have the "Insert Annotations" type as well later - but until that PR is merged, it's just "Add Note" and "Add/Edit Citation".

yexingsha:
This looks pretty good! Just need a few small tweaks:

- Segmented controls should be 28px in height, so that they are the same height as the settings button.
- The segmented controls should have 5px corner radius.
- The item/note switch should have 8px margin to the left instead of 12px.
- For the List/Library switch, the inactive part does not need to use `fill-secondary` instead of `fill-primary`, since it might give the impression that the option is disabled.
- If List is the default mode, then we should put List on the left and Library on the right.
- We should prevent the dialog from resizing when these switches are clicked.

abaevbog:
Thank you for the notes! They should be addressed with these two caveats:

> The segmented controls should have 5px corner radius.

That was supposed to be 5px to begin with from https://github.com/zotero/zotero/blob/fc30f4dbda27bc2daa2d7ccd58bab272d37361d1/scss/components/_citationDialog.scss#L495-L501 Does that not look like 5px on your end?

> We should prevent the dialog from resizing when these switches are clicked.

I'm not sure how we could do that ... We record the sizing of the window in library mode and, if one switches to list mode and then back, original library mode's height is restored. At the same time, list mode height is variable, as it depends on how many results there are to avoid having a long almost empty window when there's just one search result. After changing the type or mode of the dialog, search results are updated, so list mode's height must change. This logic is not specific to the dialog type toggles from this PR - it's how it works currently when switching between list and library modes too.

I think in my recording above, height of library mode happened to be close to max height of list most, so the height change was small and maybe looked unintentional. But this is what dialog height changes are more likely to be, I think:

https://github.com/user-attachments/assets/c7405a7c-65af-40a5-8341-74b4db87596e

yexingsha:
> That was supposed to be 5px to begin with from

Ah I see, I think the larger height of the original segmented controls made me thought that radius was smaller.

Speaking of the segmented controls, it seems that their content shift slightly when toggled. I think it's because the different borders of active and inactive labels cause active labels to be 1px narrower.

Re: dialog height changes, the reason I mentioned that was because at 0:02 of the first video you posted, there was a small shift in height when the dialog was toggled between Note + Library and Note + List, which looked like a bug to me. I think it was actually cause by library mode's adjusted height being slightly different from the list mode's default max height. Also, after further testing, the fact that we don't record height resizing in list mode feels a bit strange, since now you can switch between dialog types and see the list mode height revert to default. So I think we should record all manual height resizing, both in list mode and library mode, and keep that as the maximum height for the dialog regardless of its configuration. This would not impact the auto resizing in list mode. Does that make sense?

abaevbog:
> Speaking of the segmented controls, it seems that their content shift slightly when toggled. I think it's because the different borders of active and inactive labels cause active labels to be 1px narrower.

Good catch! This should be fixed now.

> I think it was actually cause by library mode's adjusted height being slightly different from the list mode's default max height.

Right, I must have just stretched my window to be close to the max height of list mode, so it would shift back and forth by a tiny bit, which looked not fully intentional (though it was).

> Also, after further testing, the fact that we don't record height resizing in list mode feels a bit strange, since now you can switch between dialog types and see the list mode height revert to default. 

The main reason why we did that is to avoid having a lot of empty space in the window. The list mode always occupies the amount of space that it's search results need (capped at 400px for the results).

Are you suggesting to record the resized height of the list mode independently of library mode or just maintain the resized height of the window unchanged regardless of which mode is being used? In any case, if I switch to list mode  and it's height is not made to match the number of search matches, I could end up with this strange big empty window? To me, this much empty space would look like something got broken. 

https://github.com/user-attachments/assets/4fab101f-1e48-40b9-b866-b0889cc9b33a

yexingsha:
My idea is that whenever you manually resize the dialog, that sets the maximum height for the dialog in both list mode and library mode. So it doesn't affect the auto height adjustment in list mode, but it does avoid the height shifting when the dialog is in its initial state (no search input, maximum height).

abaevbog:
Ok, I think I see what you mean! We really do often resize the window unnecessarily. Like if the window in list mode is 525px, and one switches to library mode, there's no point to restore the previous height of, say, 500px, right?

 I pushed a few tweaks to resizing. This is the behavior I'm after now:
- we'll record window height resized by the user. If window height is moved below 300px (e.g. in list mode with few/no entries), we don't record that, since it's most likely to small to be useful.
- in library mode, resizing happens only if the window is smaller than a reasonable min height of. And we'll resize to either that min height OR to the last height of the window stretched by the user (whichever is larger).
- in list mode, we allow the window to shrink as needed. And if the window needs to become larger, it can stretch to the last height set by the user

That means that if you switch between modes, and list mode has very few entries it will still shrink to avoid having empty space. But if there are many entries, it's height will remain as is. Something like this:

https://github.com/user-attachments/assets/359d3057-16c9-48c2-aa53-2c26313348f5

Is that close to what you had it mind?

yexingsha:
Yes, this feels much better now!

abaevbog:
Great! Thanks for looking into this with me!

abaevbog:
@dstillman how is this looking to you? I think that the segmented control was the way to go here. I also updated the placeholders in `bubble-input` and messages in empty selected/open/cited section of library mode to refer to notes (as opposed to generic "items") in "Add Note" dialog for a bit more context.

One small thing worth noting is that now there is an inconsistency that as a result of calling `Zotero.Integration.addNote`, one can end up adding a citation, or the other way around. But as far as I can tell, opening the dialog in one type, and then switching to another works fine on the `Zotero.Integration` side. `addNote` and `addEditCitation` are essentially identical, so I suppose it's OK.

dstillman:
Screenshot 2025-12-04 at 11 31 38 AM

The tabs here feel a little weird/unresponsive. Can we either 1) add an an active effect when clicking on one or 2) add a hover effect and switch on mousedown instead of mouseup?

macOS Settings tabs do the former:

https://github.com/user-attachments/assets/7c6f26b2-95f2-45ca-8c13-c1fa03e5e243

More traditional tabs (e.g., Safari) do the latter:

https://github.com/user-attachments/assets/04f7f1e6-4c90-4e0c-a136-db92b6e3fb0c

This component looks more like the former, but I guess I'd be inclined to do the latter just to keep things feeling fast?

abaevbog:
I see what you mean about the tabs. I added a hover effect and changed the switch to happen on `mousedown` vs `click`. How is this looking now?

https://github.com/user-attachments/assets/2025644b-ab7d-428c-b53e-27230665a0bc

dstillman:
I'll let @yexingsha take a look. It's possible a hover effect is sort of incompatible with this kind of tab. For the macOS Settings example, it can show a slightly grayed-out version of the final tab, but that might be weird on hover rather than click. But just having the gray background within the unbounded inactive tab looks kind of gross to me — the corners look like a mistake.

abaevbog:
Something I am noticing with mac OS settings tabs is that the tabs there are not touching - there is 1 or 2 px gap between them. It's different from the tabs in Scaffold (which I was somewhat replicating). So one suggestion I have is to maybe add a bit of spacing between the tabs?

Now:

https://github.com/user-attachments/assets/5962b666-05a9-4381-8795-facb541057de

With Gap:

https://github.com/user-attachments/assets/c8ab7ac0-4c2b-4a0b-a10d-b3e5363d9a16

dstillman:
Hmm, I don't think it really solves the problem. It's mostly the corners that stick out to me when I try it, and I think that may remain the case without an actual outline. But the outline isn't totally appropriate for a hover effect…

abaevbog:
For reference, this is what it would look like if we added a faint border on hover as well:

https://github.com/user-attachments/assets/b18d3e04-abf4-41b1-9ffe-cfe040743cc6

dstillman:
As a temporary measure, so we're not waiting on this, let's keep that outline and narrow the gap slightly:

Screenshot 2025-12-04 at 1 54 36 PM

dstillman:
Or, I guess, remove the gap as long as we have an outline there, which seems to be what Settings has. We can fix this more later after @yexingsha weighs in.

abaevbog:
Sounds good, how about this? The effect is still applies on `:hover` not `:active`, right?

I also realized that the `.option` nodes were seemingly overlapping a bit on hover due to flexbox layout, which I think contributed to the button borders looking off. I think it looks a bit crispier now than before. What do you think?

Now:

https://github.com/user-attachments/assets/b21e4d4f-bf09-4724-9687-a2cfe720a7de

Before:

https://github.com/user-attachments/assets/b1923495-201b-4e17-a179-52a6487e747b

yexingsha:
If we want to add a hover effect for this style of segmented control, the hover background needs to be 1px offset (so 1px gap on all sides), with its corner radius decreased accordingly:

https://github.com/user-attachments/assets/94ec7106-b688-469f-a4a2-c72d96d364d6

Or, we can try a different style of segmented control:

https://github.com/user-attachments/assets/62e8f81d-4b03-4f44-9a67-ff188c7d57ea

(These are largely based on macOS Sequoia components. We can omit the vertical dividers if they are too tricky)

dstillman:
I don't love the hover effect on the first one, but I think we have to use that style of control — it's makes it much clearer which mode is selected (particularly for List/Library).

abaevbog:
@yexingsha how is this hover effect (first option) looking to you?

https://github.com/user-attachments/assets/a9cd83bc-9e8e-4707-b4b1-0f8d2e3fb426

yexingsha:
Looks good to me!

dstillman:
I find it a little confusing that:

1) I can't select a parent item to add all of its child notes — I press Return and the selection just goes away

2) I can't select multiple notes

3) clicking the (+) button just adds the note straight to the document without adding multiple to the input field like in the other mode.

Did we discuss adding multiple notes? I know it's not particularly different from just clicking Add Note again, and if we add multiple (particularly by selecting the parent item) we have the problem of sorting, but it does seem like this is something people are going to want to be able to do. (And for annotations, you'll be able to select multiple, and we have a similar problem of sorting there, right?)

abaevbog:
> Did we discuss adding multiple notes? 

We never really discussed adding multiple notes - that's the behavior remaining from original QuickFormat dialog. I suppose there's no reason not to do it now, especially if we bubbleize annotations. It's probably best to handle it not here but in the PR for annotations, since most of the logic for it is already there.

> if we add multiple (particularly by selecting the parent item) we have the problem of sorting

Since relations between notes and their ordering is not explicit, we can just insert notes in the sequence in which they are added (or in the sequence in which they appear below the parent) but still allow user to reorder them. 

We ran into issues while experimenting with the behavior of adding all children from parent in the annotations PR... But I guess there are never as many notes as there are annotations, and wanting to add all notes is way more common than wanting to add all annotations, so it's worth considering. 

Somewhat relevant improvements to the Insert Note dialog were discussed in https://github.com/zotero/zotero/pull/5552 . Those changes display the parent item in list mode (plus search by parent item in list mode is possible). So with those changes, we could just add an "Add all" button to parent items in list mode, as well as add all child notes on `+` click in library mode. I can add those changes to that PR.

> (And for annotations, you'll be able to select multiple, and we have a similar problem of sorting there, right?)

For annotations, sorting is currently disabled. They are grouped by their parent item and sorted in order in which they appear in the document

dstillman:
Oh, sorry, for some reason I was thinking that this was adding note support to the new citation dialog, but we already had that. This is just adding mode switching. We can just get this in, then.

dstillman:
There's one conflict that has to be resolved.

abaevbog:
Rebased!

> Oh, sorry, for some reason I was thinking that this was adding note support to the new citation dialog, but we already had that. This is just adding mode switching. We can just get this in, then.

Those notes-related comments are still valid - we'll add those tweaks in the remaining notes-related PRs

Issue #5580, opened 2025-10-14
Sync Reset Options: Radio buttons are misaligned


dstillman:
<img width="1132" height="714" alt="Image" src="https://github.com/user-attachments/assets/a482d558-ac08-4a02-98fc-736f07f2e598" />

PR #5587, opened 2025-10-20
Fix misaligned sync reset options radio button labels

abaevbog:
Keeping the spacing between the radio buttons and the description text below.

Fixes: #5580

Before:

Screenshot 2025-10-20 at 1 37 32 PM

After:

Screenshot 2025-10-20 at 1 36 45 PM

### DIFF
--- a/scss/preferences/_sync_reset.scss
+++ b/scss/preferences/_sync_reset.scss
@@ -61,10 +61,6 @@
 	margin-top: 1em;
 }
 
-#sync-reset-radiogroup > div radio .radio-check {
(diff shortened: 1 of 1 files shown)


dstillman:
The spacing between the radio button and the text isn't intentional, so we can fix that too, in line with other radio buttons in the prefs.

abaevbog:
Gotcha, I thought it was intentional for better readability, or something like that. I removed that extra spacing, this is what it looks like now:

Screenshot 2025-10-21 at 9 20 37 AM

PR #5588, opened 2025-10-21
Don't cache `CSL.Engine` for integration

### LINKED ISSUE #5581: Add Bibliography inserts bibliography from previously updated document
When working with multiple documents, inserting a bibliography in one document can insert the bibliography from another document.

1. Open 2 empty word documents - A and B
2. Insert a citation into document A
3. Insert a different citation into document B
4. Generate bibliography for document B
5. Go to document A, and generate the bibliography there

Note that the bibliography inserted is the same as from document B.

Demo with the latest beta build on macOS: https://www.dropbox.com/scl/fi/7092ov9799igaap31u54v/bibliography_stuck.mov?rlkey=w2qcybxu3wxczzqtmamukx51y&st=7nv5ucrq&dl=0

AbeJellinek:
Replacing `automaticJournalAbbreviations` with a backwards-compatible `options` object, as is the way of things. Integration code relies too much on internal `CSL.Engine` state to be able to share instances.

This does not revert the `restoreProcessorState()` -> `updateItems([])` change made in 08e4d6b06b06084968f3403464d118a04ba3c107; the latter works just as well and avoids a deprecated method call.

Fixes #5581

### DIFF
--- a/chrome/content/zotero/xpcom/integration.js
+++ b/chrome/content/zotero/xpcom/integration.js
@@ -1927,7 +1927,10 @@ Zotero.Integration.Session.prototype.setData = async function (data, resetStyle)
 			if (this.style && this.style.free) {
 				this.style.free();
 			}
-			this.style = getStyle.getCiteProc(data.style.locale, this.outputFormat, data.prefs.automaticJournalAbbreviations);
+			this.style = getStyle.getCiteProc(data.style.locale, this.outputFormat, {
+				automaticJournalAbbreviations: data.prefs.automaticJournalAbbreviations,
+				noCache: true,
+			});
 			this.styleClass = getStyle.class;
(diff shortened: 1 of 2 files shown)


adomasven:
My only concern is whether caching should be a default behaviour or not. If you're calling `getCiteProc()`, presumably you expect an instance that yours (at least that was the previous behaviour), including the internal state of the processor. Maybe this is fine, but maybe something like BBT may also be impacted by this.

AbeJellinek:
I imagine that almost all `getCiteProc()` calls, even in plugins, are one-and-done citation/bibliography generations.

BBT doesn't hold onto any CSL.Engine instances, as far as I can tell. Only [three `getCiteProc()` calls](https://github.com/search?q=repo%3Aretorquere%2Fzotero-better-bibtex%20getciteproc&type=code) anywhere in its code.

PR #5592, opened 2025-10-23
Fix broken new collection dialog opened from context menu of a group

abaevbog:
Fix new collection dialog appearing broken when opened via "New Collection" option of context menu on "My Library" or a group in collection tree.

On macOS, the popup of collections would never leave and on windows, subcollection would never appear.

https://github.com/user-attachments/assets/fae27dcd-2b49-4409-b147-758de40788ec

This is a followup to https://github.com/zotero/zotero/pull/5409 that fixed this issue for all collections. The reason why it didn't work for groups is that the command event would fire not on the `menuitem` but on the [command](https://github.com/zotero/zotero/blob/5b82a38383ee2689b959c656713beb3e8accdea6/chrome/content/zotero/zoteroPane.xhtml#L117) node, which is not what the workaround expects (e.g. `command` as a target node does not have the `menupopup`, and etc.). Instead of tweaking the workaround (that was already hacky enough), I thought it's best to just skip the `command` node for this specific option, which solves this on my end.

### DIFF
--- a/chrome/content/zotero/customElements.js
+++ b/chrome/content/zotero/customElements.js
@@ -126,7 +126,7 @@ Services.scriptloader.loadSubScript('chrome://zotero/content/elements/itemTreeMe
 		// event with <menuitem> as target in event.sourceEvent.
 		// This is the event we want to work with, since we do not expect events with .sourceEvent
 		// and <command> has no menupopup ancestor.
-		if (originalTarget.tagName === "command" && event.sourceEvent?.type === "command") {
+		if (originalTarget.localName === "command" && event.sourceEvent?.type === "command") {
 			event = event.sourceEvent;
 			originalTarget = event.target;
 		}


abaevbog:
@AbeJellinek, I ran into another instance of a broken "New Collection" dialog opened from a context menu that we worked around in https://github.com/zotero/zotero/pull/5409 a while back.

To me, this seems like the safest way to fix it. How does this look to you?

abaevbog:
A good question is why does "New Saved Search" dialog not have any issues, even though it is opened the same way via a `command`:
https://github.com/zotero/zotero/blob/5b82a38383ee2689b959c656713beb3e8accdea6/chrome/content/zotero/zoteroPane.js#L3279-L3286

My guess is because `ZoteroPane.newSearch` takes a bit longer to open `searchDialog.xhtml` because it loads all searches in the library first.

In any case, I could not find any more instances of this behavior besides `newCollection` dialog for the group row, so hopefully this is the last issue of this kinda.

AbeJellinek:
> because `ZoteroPane.newSearch` takes a bit longer to open `searchDialog.xhtml`

Yeah, I guess, but what if it doesn't? Just not great to be in a situation where adding more caching to `Zotero.Searches.getAll()` (or just using a machine with very fast storage/a lucky disk cache hit) could break a popup menu in some dialog some of the time. Or where we need to remember not to use the `command` attribute on menuitems that open some dialogs in the future.

My assumption is that we could handle `<command>`s in our existing `command` capture hack without adding that much complexity. But if you tried doing it and it seemed too messy, I get it!

abaevbog:
> Yeah, I guess, but what if it doesn't?

That is a good point... "New Search" works fine for me but that's no guarantee that it would work for you. And it does mean that essentially all instances of `command:` should be `oncomand: () => ... `, which is not good.

So I looked over it again. We seem to be able to get the original event dispatched on the `<menuitem>` from `event.sourceEvent` and use that to proceed with the rest of the execution just fine. I repushed that commit - does it make sense to you?

Also, let me know if the comment is clear enough - it has a lot of different "command"s, so I ended up rewriting it several times before settling on this version.

AbeJellinek:
I think we should check `localName` instead of `tagName` so namespace prefixes don't interfere, but other than that, looks good to me!

AbeJellinek:
Looks good.

Issue #5584, opened 2025-10-16
Test fails: should refresh on item update


windingwind:
While debugging the tests, another weird thing happens. The Item pane -> Info pane -> `should refresh on item update` test from time to time fails, and it turned out that the 
```js
item.setField('series', 'Test'); 
await item.saveTx();`
```
(L262 in itemPaneTest.js)

has a chance to trigger an item-modify update with a null value of field `series`: https://gist.github.com/windingwind/b2c1c3410772117b363fc1e1d618c38d#file-ci_log-txt-L216

```
Notifier.trigger('modify', 'item', [465], {"465":{"changed":{"series":null}}}) called [observers: 30]
```

which leads to test failures. It seems like there's something from a lower level that's causing this.

Full test logs see https://github.com/windingwind/zotero/actions/runs/18560770287/job/52912718306

abaevbog:
Yeah, this test failure is incredibly annoying! I keep running into it as well. Thank you for the logs, it's a great starting point for troubleshooting.

My understanding is that the notifier is behaving as expected. When data fields (e.g. "series") are changed, the data in the notifier event has what the field was before the update, not the new updated value of the field. So if a "series" field was empty, `{"series":null}` is what we should get when it is changed to be not empty.  (At least, that's what I see after manually updating an item too).

I am inclined to blame the failing test on this line from the logs: https://gist.github.com/windingwind/b2c1c3410772117b363fc1e1d618c38d#file-ci_log-txt-L88. It implies that the item being tested is not selected in the itemTree.

Later, there is a line indicating that "My Library" is getting selected in collectionTree: https://gist.github.com/windingwind/b2c1c3410772117b363fc1e1d618c38d#file-ci_log-txt-L134. That's odd because it is supposed to be selected to begin with.

One of earlier tests selects the trash and then supposedly goes back to "My Library":
https://github.com/zotero/zotero/blob/5b82a38383ee2689b959c656713beb3e8accdea6/test/tests/itemPaneTest.js#L214-L236

But in the logs, I am seeing this line around the time when that test completes:
```
2025-10-16T12:56:51.6574749Z zotero(3)(+0000002): CollectionTree.selectWait(): selectEventsSuppressed. Not waiting to select row 0
```

The selection in `collectionTree` gets suppressed during refreshing or handling of notifier events. So I am wondering if the test could be failing because of a race condition where `await selectLibrary(win); ` from the test above doesn't actually end up selecting the library because selection in collection tree is suppressed.

[...]

PR #5595, opened 2025-10-23
Potential fix to test 'should refresh on item update' failing sometimes

abaevbog:
"should refresh on item update" test would sometimes fail, most likely due to the library sometimes not getting re-selected in the previous test 'should update custom header for items in the trash'.

A likely explanation is that the selection event in collectionTree would still be suppressed when selectLibrary is called, so make sure to wait for item deletion to go through before trying to re-select the library. More context: https://github.com/zotero/zotero/issues/5584#issuecomment-3438330589

Fixes: #5584

### DIFF
--- a/test/tests/itemPaneTest.js
+++ b/test/tests/itemPaneTest.js
@@ -230,8 +230,13 @@ describe("Item pane", function () {
 			
 			assert.notExists(win.document.querySelector('#zotero-item-message .custom-head .item-restore-button'));
 			
-			await item1.eraseTx();
-			await item2.eraseTx();
+			let promise = waitForItemEvent('delete');
+			await Zotero.DB.executeTransaction(async function () {
+				await item1.erase();
+				await item2.erase();
+			});
+			await promise;
+
 			await selectLibrary(win);
 		});
 	});


abaevbog:
I have rerun the tests about 10 times in total, and `should refresh on item update` has not failed on me in any of those runs. So I hope this is it!

windingwind:
Still seeing it after merging changes from this PR...

https://github.com/windingwind/zotero/actions/runs/18823362498/job/53746130564 (and other failures attempt 9 - 13)

abaevbog:
Oh man, that's unfortunate. It looks like the item is still not being selected. I am seeing the same `Couldn't find row for item ... -- not selecting` in those logs as well. I suppose it's something else then.

Oddly enough, I still haven't run into this test breakage after that fix though. Maybe having the logs enabled while running tests `xvfb-run test/runtests.sh -f -d 5` somehow helps to trigger it.

windingwind:
fully understand that's really unfortunate. The only differences between the commit where the logs are made and the current main branch is there i added more logging and some fixes to the tests after this one, which shouldn't really affect.

i ran for ~15 times since the merge and only in the recent 4-5 times it continuesly fails. i assume it might get affected by if the host where the ci is running on is slow or fast. (i was almost never able to reproduce the test failure when i try to fix locally even with 100+ times run but on ci it fails 10-20%)

Issue #5601, opened 2025-10-28
Can't navigate to related item without tag when filtering by tag


AbeJellinek:
More generally, I guess we should make tag filtering act like quick search, so `selectItems()` clears the filter if the target items don't match.

https://forums.zotero.org/discussion/127714/blocked-access-to-related-items-if-not-matching-tag

PR #5602, opened 2025-10-28
Clear tag selector when itemTree selection changes, if needed

abaevbog:
Similar to how quickSearch is being cleared already.

To avoid rerunning itemTree refresh twice (once for clearing quickSearch and once for clearing tag filter),
clear tags and search text directly on collectionTreeRow. And then call itemTree.refreshAndMaintainSelection to refresh everything. Without it, tag selector sometimes would find itself out of sync with current search results and display tags as if quickSearch was not cleared.

Fixes: #5601


abaevbog:
https://github.com/user-attachments/assets/083b733c-ff7d-401c-88a0-a3943f700d59

abaevbog:
@AbeJellinek, does this approach seem reasonable to you?

PR #5604, opened 2025-10-29
Align buttons with the top line of the field

### LINKED ISSUE #5268: In item pane, align buttons with the top line of the field
See https://github.com/zotero/zotero/pull/3860#issuecomment-2853372843.

- In meta data and attachment rows, the buttons are currently aligned with the center of the whole field.

- In notes, libraries and collections, and related rows, the buttons are currently aligned with the top of the field. It seems correct at the default font size, but when font size changes, the button becomes misaligned with the top line of the field.

tnajdek:
(Screenshots below have significantly increased font size)

Metadata table:
Screenshot 2025-10-29 at 15 18 25

Attachments list:
Screenshot 2025-10-29 at 15 18 38

Tags list:
Screenshot 2025-10-29 at 15 24 09

"Re-index" icon tweaked size and position to match:
Screenshot 2025-10-29 at 15 18 17

Close #5268

### DIFF
--- a/chrome/content/zotero/elements/attachmentBox.js
+++ b/chrome/content/zotero/elements/attachmentBox.js
@@ -64,7 +64,7 @@
 						<html:div id="indexStatusRow" class="meta-row">
 							<html:div class="meta-label"><html:label id="index-status-label" class="key" data-l10n-id="attachment-info-index"/></html:div>
 							<html:div class="meta-data">
-								<html:label id="index-status"/>
+								<html:span id="index-status"/>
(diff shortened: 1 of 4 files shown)


dstillman:
Sorry, I didn't mean to suggest this as the solution. In general, if the font size is increased, I think we want to either increase the button sizes proportionally or, if we're going to keep them the same size, keep them aligned with the center of the top row of text.

tnajdek:
Oh, I read “top line” as the top edge, not the first line of the actual text. This way it actually makes more sense:

Screenshot 2025-10-30 at 00 56 07

Screenshot 2025-10-30 at 00 55 50

Screenshot 2025-10-30 at 01 13 06

Screenshot 2025-10-30 at 00 20 59

I also undid all the changes to the alignment in the tag list.

Issue #5607, opened 2025-10-30
Connector save hangs when an item is selected in Advanced Search


AbeJellinek:
https://forums.zotero.org/discussion/127760/connector-conflict-with-advanced-search

abaevbog:
Very strange!

From my observation, selecting an item in the itemTree is not required - saving will hang as long as the search results are not empty. ~I can also reproduce it in other dialogs where itemTree search is applicable. The same happens if I have `selectItemsDialog` open to add a related item, as long as quick search at the top is active.~

abaevbog:
Still figuring out the best way to work around this but the culprit is here:

https://github.com/zotero/zotero/blob/2d6deda04a8a874fd63ec4f07a87beb635ff27c2/chrome/content/zotero/itemTree.jsx#L952-L957

It seems like the itemTree notifier tries to wait for selection to change, which never happens in the advanced search window, which hangs the notifier and DB transactions as a result.

PR #5609, opened 2025-10-31
Fix connector save timing out with Advanced Search

adomasven:
Report: https://forums.zotero.org/discussion/127760/connector-conflict-with-advanced-search

Also restore Advanced Search tests that depend on the item tree. They were disabled due to React timing issues. Hopefully with these changes the initialization of the tree is more robust.


AbeJellinek:
(#5607)

AbeJellinek:
Tested, looks good to me!

PR #5618, opened 2025-11-07
Fix missing labels for options under “Display citations as”

### LINKED ISSUE #5617: Missing footnotes/endnotes label in Bibliography dialog
Image

tnajdek:
Resolves #5617


dstillman:
This is fine, but I can't actually reproduce this. Are you getting this just by running "Create Bibliography from Items" after startup?

tnajdek:
Correct - for me, the strings from `integration.ftl` are always missing in that dialog. It works in Zotero 7, but that’s because those two strings were previously located in `zotero.ftl` as far as I can tell.

Issue #5624, opened 2025-11-11
Test failure: `should discard attachment pane preview after becoming invisible`


dstillman:
https://github.com/zotero/zotero/actions/runs/19271995328/job/55109489316

Not to be confused with the "attachment**s** pane preview" test fixed in #5589

dstillman:
Alas, still happening:

https://github.com/zotero/zotero/actions/runs/20276686631

dstillman:
https://github.com/zotero/zotero/actions/runs/20316551730/job/58361593131

dstillman:
https://github.com/zotero/zotero/actions/runs/20359476444/job/58501857458

windingwind:
That's really unfortunate, likely after some recent commits it is happening again. I'm trying to figure out...

PR #5630, opened 2025-11-12
Fix `should discard attachment pane preview after becoming invisible`

windingwind:
fix: #5624

(need to run more rounds to test)


windingwind:
0 failures caused by this test in 20 runs: https://github.com/zotero/zotero/actions/runs/19306025389

PR #5635, opened 2025-11-13
Fix missing icons in "Libraries and Collections"

### LINKED ISSUE #5634: Missing icons in rows of Libraries and Collections section
Image

The icons are there but they have no height. This looks like a regression from https://github.com/zotero/zotero/commit/e09e99308136994faa480458021d78ce40fa314e

tnajdek:
Resolve #5634

### DIFF
--- a/scss/abstracts/_mixins.scss
+++ b/scss/abstracts/_mixins.scss
@@ -115,7 +115,7 @@
 
 @mixin clicky-item {
 	display: flex;
-	align-items: baseline;
+	align-items: flex-start;
 	gap: 4px;
 	padding-inline-start: 4px;
 	overflow: hidden;
@@ -130,7 +130,8 @@
 	}
 
 	.icon {
-		line-height: calc(1.3333333333 * var(--zotero-font-size));
+		height: calc(1.3333333333*var(--zotero-font-size));
+		padding-top: 0.25em; 
 	}
 
 	.label {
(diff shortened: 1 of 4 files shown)


abaevbog:
Great!

What do you think about the alignment of the icons in relations to the text? They seem to hover somewhat above the text, with the bottom of the icon being a few pixels above the bottom of the label.

Screenshot 2025-11-13 at 2 27 34 PM

I don't know if it's related to the recent changes or if it's actually always been like that - but would it look better if we center-aligned them in the row?

Screenshot 2025-11-13 at 2 31 10 PM

tnajdek:
The reason I've changed alignment from center to baseline in few places was to tackle the problem of multi-line rows, where icons should be aligned to the middle of the first line of text, like in here:

Screenshot 2025-11-14 at 01 00 30

and here:

https://github.com/zotero/zotero/pull/5604#issuecomment-3465447697

I can see that my solution still has problems like the one you've outlined. Also I believe the creator row now has icons that are just slightly off the center. I'll try to come up with some new strategy that covers all the cases.

tnajdek:
So I believe this now aligns correctly but I might need to move some style definitions around.

abaevbog:
Yeah, the icons look much nicer now.

Like you pointed out, do the buttons of creator rows look pushed maybe a bit too close to the top (vs being centered)? I believe they shifted after https://github.com/zotero/zotero/commit/e09e99308136994faa480458021d78ce40fa314e as well

Now:
Screenshot 2025-11-13 at 5 03 23 PM

Before:
Screenshot 2025-11-13 at 5 15 08 PM

tnajdek:
Ok, I believe this is good to go!

In multi-line rows, icons align with the top line of the text:
Screenshot 2025-11-14 at 13 50 58
Screenshot 2025-11-14 at 13 51 15
Screenshot 2025-11-14 at 13 51 25

In single-line rows, icons appear centered:

Screenshot 2025-11-14 at 13 52 17
Screenshot 2025-11-14 at 13 51 32

"Libraries and Collections" no longer has missing icons:
Screenshot 2025-11-14 at 13 54 05

And alignment is correct with larger font size:
Screenshot 2025-11-14 at 13 52 47
Screenshot 2025-11-14 at 13 52 42

tnajdek:
@abaevbog thank you for helping test this so far! Could you please have one more look and if it's OK, I'll merge this.

abaevbog:
For sure! Yeah, now everything is looking nice and aligned to me!

PR #5636, opened 2025-11-17
Citation dialog: warn about cross-library citations

### LINKED ISSUE #5631: Citation Dialog: Warn if citing items from multiple libraries
Based on [Dan's comment](https://github.com/zotero/zotero/pull/3468#issuecomment-3523994165) and [user requests](https://forums.zotero.org/discussion/127786/add-new-advanced-option-to-avoid-common-mistake-in-collaborative-documents), we might also want to warn users when citing items in a document, if they all come from one library, and then the user chooses items from a different one.

Not sure about the UX. Probably should pop up when you add the item into the citation dialog from the list, and say that all the cited items in this document are from a different library. It should allow to cancel add, or proceed, with a checkbox to not warn about that in this document again.

abaevbog:
When a new item is added into the citation, and all other cited items are all from another library, display a warning saying that cross-library citations are not recommended. If the document has cited items from multiple libraries, no warning is shown.

If there are many cited items, and `io` is still loading, the warning will appear if needed as soon as `io` is ready.

If the user clicks "Cancel" in the warning dialog, all added items from other libraries are removed.

Fixes: #5631

---

For reference:

https://github.com/user-attachments/assets/0667ee52-0c7c-4ebc-ad70-7c79fb4408b5

If io takes a long time to load, and the user adds some items and clicks `Accept`, the warning will be shown before the dialog is accepted:

https://github.com/user-attachments/assets/f04800e5-9826-4c5c-9c49-e761ce714b68

I wasn't sure about the use case for the pref to not show this warning in this document again, as mentioned in the linked issue. Once the user adds at least one citation from another library, next time the dialog is used, all cited items will not belong to a single library, so the warning won't be shown already.

### DIFF
--- a/chrome/content/zotero/integration/citationDialog.js
+++ b/chrome/content/zotero/integration/citationDialog.js
@@ -59,6 +59,16 @@ async function onLoad() {
 	}
 	ioReadyPromise.then(() => ioIsReady = true);
 	isCitingNotes = !!io.isCitingNotes;
+	// set default cited libraries info, if not provided
+	if (!io.getCitedLibraryInfo) {
+		io.getCitedLibraryInfo = () => {
+			return {
+				citedLibrariesURIs: [],
+				citedLibrariesNames: [],
+				citedLibrariesIDs: []
+			};
+		};
+	}
 	window.isPristine = true;
 
 	Zotero.debug("Citation Dialog: initializing");
--- a/chrome/content/zotero/integration/citationDialog.js
+++ b/chrome/content/zotero/integration/citationDialog.js
@@ -124,7 +124,7 @@ async function onLoad() {
 		await Zotero.Integration.currentSession.refreshCitedLibraries(true);
 		let { citedLibrariesIDs } = io.getCitedLibraryInfo();
 		if (citedLibrariesIDs.length) {
-			await libraryLayout.collectionsView.setTopGroup(citedLibrariesIDs, "Cited in this document");
+			await libraryLayout.collectionsView.setCitedGroup(citedLibrariesIDs);
 		}
 		Zotero.debug("Citation Dialog: io loaded cited data");
 		await SearchHandler.refreshCitedItems();
@@ -670,16 +670,16 @@ class LibraryLayout extends Layout {
 			onActivate: () => {},
 			filterLibraryIDs: io.filterLibraryIDs
 		});
-		// Wait for collectionTree to get fully loaded before moving cited libraries to the top
+		// Wait for collectionTree to get fully loaded before setting cited groups
 		let waitCount = 10;
 		while (!this.collectionsView.getSelectedLibraryID() && waitCount < 50) {
 			await Zotero.Promise.delay(10);
 			waitCount++; // sanity check to avoid infinite loop if something goes wrong
 		}
-		// Move cited libraries to the top of the collection tree
+		// Move cited groups to the top
 		let { citedLibrariesIDs } = io.getCitedLibraryInfo();
 		if (citedLibrariesIDs.length) {
-			await libraryLayout.collectionsView.setTopGroup(citedLibrariesIDs, "Cited in this document");
+			await libraryLayout.collectionsView.setCitedGroup(citedLibrariesIDs);
 		}
 		// Add aria-description with instructions on what this collection tree is for
 		// Voiceover announces the description placed on the actual tree when focus enters it
@@ -849,6 +849,19 @@ class ListLayout extends Layout {
 		let collapsibleSection = doc.querySelector(".section.expandable");
 		if (collapsibleSection) {
 			collapsibleSection.querySelector(".header-label").addEventListener("click", () => IOManager.toggleSectionCollapse(collapsibleSection, null, true));
+		}
+
+		// Add "Cited" badge to headers of cited libraries
+		for (let section of [..._id("list-layout").querySelectorAll(".section")]) {
+			let id = section.id;
+			let libraryID = id.match(/^list-(\d+)-items$/)?.[1];
+			if (!libraryID) continue;
+			let { citedLibrariesIDs } = io.getCitedLibraryInfo();
+			if (citedLibrariesIDs.includes(parseInt(libraryID))) {
+				let citedBadge = Helpers.createNode("span", {}, "cited-badge");
+				document.l10n.setAttributes(citedBadge, "integration-citationDialog-cited-label");
+				section.querySelector(".header").appendChild(citedBadge);
+			}
 		}
 		if (!options.skipWindowResize) {
 			this.resizeWindow();
--- a/chrome/content/zotero/integration/citationDialog.js
+++ b/chrome/content/zotero/integration/citationDialog.js
@@ -679,7 +679,7 @@ class LibraryLayout extends Layout {
 		// Move cited groups to the top
 		let { citedLibrariesIDs } = io.getCitedLibraryInfo();
 		if (citedLibrariesIDs.length) {
-			await libraryLayout.collectionsView.setCitedGroup(citedLibrariesIDs);
+			await libraryLayout.collectionsView.setCitedGroup(citedLibrariesIDs, !ioIsReady);
 		}
 		// Add aria-description with instructions on what this collection tree is for
 		// Voiceover announces the description placed on the actual tree when focus enters it
@@ -859,7 +859,13 @@ class ListLayout extends Layout {
 			let { citedLibrariesIDs } = io.getCitedLibraryInfo();
 			if (citedLibrariesIDs.includes(parseInt(libraryID))) {
 				let citedBadge = Helpers.createNode("span", {}, "cited-badge");
-				document.l10n.setAttributes(citedBadge, "integration-citationDialog-cited-label");
+				citedBadge.textContent = Zotero.getString('integration-citationDialog-cited-label');
+				// Untill the cited items are fetched, add a spinner to the cited badge to indicate
+				// that the library is not confirmed as cited
+				if (!ioIsReady) {
+					citedBadge.classList.add("loading");
+					citedBadge.setAttribute('title', Zotero.getString('integration-citationDialog-cited-title-loading'));
+				}
 				section.querySelector(".header").appendChild(citedBadge);
 			}
 		}
--- a/chrome/content/zotero/integration/citationDialog.js
+++ b/chrome/content/zotero/integration/citationDialog.js
@@ -134,17 +134,12 @@ async function accept() {
 	if (accepted || SearchHandler.searching || !CitationDataManager.items.length) return;
 	accepted = true;
 	Zotero.debug("Citation Dialog: accepted");
-	_id("library-layout").hidden = true;
-	_id("list-layout").hidden = true;
-	_id("bubble-input").hidden = true;
-	_id("bottom-area").hidden = true;
-	_id("progress").hidden = false;
+	changeProgressDisplay(true);
 	let progressHeight = Helpers.getSearchRowHeight();
 	// The minHeight is not just removed so that windows doesn't make the window too small
 	document.documentElement.style.minHeight = progressHeight + "px";
-	setTimeout(() => {
-		window.resizeTo(window.innerWidth, progressHeight);
-	});
+	await Zotero.Promise.delay();
+	window.resizeTo(window.innerWidth, progressHeight);
 	// If items were added before sorting was ready, we must wait to sort them here.
 	// Otherwise, if the dialog is opened again, bubbles will not be in the correct
 	// order, even though the citation itself will look right.
@@ -152,6 +147,24 @@ async function accept() {
 		await ioReadyPromise;
 		await CitationDataManager.sort();
 	}
+	// Refresh cited libraries and check if cross-library citation warning
+	// needs to be shown again. Needed to handle rare cases when
+	// recorded cited libraries in document prefs get outdated.
+	await Zotero.Integration.currentSession.refreshCitedLibraries(true);
+	let canProceed = IOManager.checkCrossLibraryCitations(CitationDataManager.items.map(obj => obj.item));
+	if (!canProceed) {
+		// Remove added items and un-accept the dialog
+		let { citedLibrariesIDs } = io.getCitedLibraryInfo();
+		for (let item of CitationDataManager.items) {
+			if (!citedLibrariesIDs.includes(item.item.libraryID)) {
+				await IOManager._deleteItem(item.dialogReferenceID);
+			}
+		}
+		changeProgressDisplay(false);
+		accepted = false;
+		return;
+	}
+	
 	CitationDataManager.updateCitationObject(true);
 	cleanupBeforeDialogClosing();
 	io.accept((percent) => {
@@ -190,6 +203,20 @@ function dialogNotPristine() {
 // shortcut used for brevity
 function _id(id) {
 	return doc.getElementById(id);
+}
+
+// Switch from displaying the actual citation dialog to loading indicator, or the other way around.
+function changeProgressDisplay(progressVisible) {
+	_id(`${currentLayout.type}-layout`).hidden = progressVisible;
+	_id("bubble-input").hidden = progressVisible;
+	_id("bottom-area").hidden = progressVisible;
+	_id("progress").hidden = !progressVisible;
+	if (!progressVisible) {
+		_id("zotero-collections-tree").replaceChildren();
+		libraryLayout.collectionsView = null;
+		libraryLayout._initCollectionTree();
+		_id("bubble-input").refocusInput();
+	}
 }
 
 
@@ -620,12 +647,20 @@ class LibraryLayout extends Layout {
 	
 	async _initCollectionTree() {
 		const CollectionTree = require('zotero/collectionTree');
+		// Display cited libraries at the top of collection tree, separate
+		// from all other ones
+		let topGroupConfig = null;
+		let { citedLibrariesIDs } = io.getCitedLibraryInfo();
+		if (citedLibrariesIDs.length) {
+			topGroupConfig = { libraryIDs: citedLibrariesIDs, headerLabel: "Cited in this document" };
+		}
 		this.collectionsView = await CollectionTree.init(_id('zotero-collections-tree'), {
 			onSelectionChange: this._onCollectionSelection.bind(this),
 			hideSources: ['duplicates', 'trash', 'feeds'],
 			initialFolder: Zotero.Prefs.get("integration.citationDialogCollectionLastSelected"),
 			onActivate: () => {},
-			filterLibraryIDs: io.filterLibraryIDs
+			filterLibraryIDs: io.filterLibraryIDs,
+			topGroupConfig: topGroupConfig
 		});
 		// Add aria-description with instructions on what this collection tree is for
 		// Voiceover announces the description placed on the actual tree when focus enters it
@@ -887,6 +922,7 @@ class ListLayout extends Layout {
 //
 const IOManager = {
 	sectionExpandedStatus: {},
+	crossLibraryCitationsAllowed: false,
 
 	// most essential IO functionality that is added immediately on load
 	preInit() {
@@ -1017,6 +1053,13 @@ const IOManager = {
 		// If multiple items are being added, only add ones that are not included in the citation
 		if (items.length > 1) {
 			items = items.filter(item => !(item.id && CitationDataManager.getItems({ itemID: item.id }).length));
+		}
+
+		// Potentially show an alert if the user is attempting a cross-library citation
+		let canProceed = this.checkCrossLibraryCitations(items);
+		if (!canProceed) {
+			_id("bubble-input").refocusInput();
+			return;
 		}
 
 		// If the last input has a locator, add it into the item
@@ -1217,6 +1260,28 @@ const IOManager = {
 			desiredMode = "list";
 		}
 		this.toggleDialogMode(desiredMode);
+	},
+
+	// Check if provided items belong to a library different from
+	// the library of all other cited items. If so, show a warning message.
+	checkCrossLibraryCitations(items) {
+		let { citedLibrariesURIs, citedLibrariesNames } = io.getCitedLibraryInfo();
+		// If there is no single group that all cited items belong to, do nothing
+		if (!citedLibrariesURIs.length) return true;
+		// If the dialog was already shown and confirmed, do nothing
+		if (IOManager.crossLibraryCitationsAllowed) return true;
+		let itemGroups = items.map(item => Zotero.Libraries.get(item.libraryID)).filter(Boolean);
+		let notCitedGroups = itemGroups.filter(group => !citedLibrariesURIs.includes(Zotero.URI.getGroupURI(group)));
+		if (notCitedGroups.length === 0) return true;
+		Zotero.debug("Citation Dialog: show dialog to confirm cross-library citations");
+		// Show a confirmation dialog
+		let canProceed = PopupsHandler.showCrossLibraryCitationWarning(citedLibrariesNames);
+		if (canProceed) {
+			// If confirmed, set a flag to not show this warning again on Zotero.Integration level
+			IOManager.crossLibraryCitationsAllowed = true;
+		}
+		Zotero.debug(`Citation Dialog: Allow cross-library citations - ${canProceed}`);
+		return canProceed;
 	},
 	
 	// handle drag start of item nodes into bubble-input
--- a/chrome/content/zotero/integration/citationDialog.js
+++ b/chrome/content/zotero/integration/citationDialog.js
@@ -120,6 +120,12 @@ async function onLoad() {
 	// wait to call functions that rely on io.getItems() or io.sort() till all cited data is loaded
 	ioReadyPromise.then(async () => {
 		if (accepted) return;
+		// Move cited collections to the top, when ready
+		await Zotero.Integration.currentSession.refreshCitedLibraries(true);
+		let { citedLibrariesIDs } = io.getCitedLibraryInfo();
+		if (citedLibrariesIDs.length) {
+			await libraryLayout.collectionsView.setTopGroup(citedLibrariesIDs, "Cited in this document");
+		}
 		Zotero.debug("Citation Dialog: io loaded cited data");
 		await SearchHandler.refreshCitedItems();
 		currentLayout.refreshItemsList({ retainItemsState: true });
@@ -657,21 +663,24 @@ class LibraryLayout extends Layout {
 	
 	async _initCollectionTree() {
 		const CollectionTree = require('zotero/collectionTree');
-		// Display cited libraries at the top of collection tree, separate
-		// from all other ones
-		let topGroupConfig = null;
-		let { citedLibrariesIDs } = io.getCitedLibraryInfo();
-		if (citedLibrariesIDs.length) {
-			topGroupConfig = { libraryIDs: citedLibrariesIDs, headerLabel: "Cited in this document" };
-		}
 		this.collectionsView = await CollectionTree.init(_id('zotero-collections-tree'), {
 			onSelectionChange: this._onCollectionSelection.bind(this),
 			hideSources: ['duplicates', 'trash', 'feeds'],
 			initialFolder: Zotero.Prefs.get("integration.citationDialogCollectionLastSelected"),
 			onActivate: () => {},
-			filterLibraryIDs: io.filterLibraryIDs,
-			topGroupConfig: topGroupConfig
+			filterLibraryIDs: io.filterLibraryIDs
 		});
+		// Wait for collectionTree to get fully loaded before moving cited libraries to the top
+		let waitCount = 10;
+		while (!this.collectionsView.getSelectedLibraryID() && waitCount < 50) {
+			await Zotero.Promise.delay(10);
+			waitCount++; // sanity check to avoid infinite loop if something goes wrong
+		}
+		// Move cited libraries to the top of the collection tree
+		let { citedLibrariesIDs } = io.getCitedLibraryInfo();
+		if (citedLibrariesIDs.length) {
+			await libraryLayout.collectionsView.setTopGroup(citedLibrariesIDs, "Cited in this document");
+		}
 		// Add aria-description with instructions on what this collection tree is for
 		// Voiceover announces the description placed on the actual tree when focus enters it
 		if (Zotero.isMac) {
--- a/chrome/content/zotero/collectionTree.jsx
+++ b/chrome/content/zotero/collectionTree.jsx
@@ -113,7 +113,7 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 		this._expandRowOnHoverTimer = null;
 		this._collapseExpandedRowsTimer = null;
 		
-		this._topGroupConfig = null;
+		this._citedGroupLibraryIDs = new Set();
 		
 		this.onLoad = this.createEventBinding('load', true, true);
 	}
@@ -140,20 +140,18 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 	}
 
 	/**
-	 * Set libraries to display at the top of the tree
+	 * Set cited groups to display at the top of the groups section
 	 *
-	 * @param {Array<number>} libraryIDs - Array of library IDs to show at the top
-	 * @param {string} headerLabel - Label for the header row
+	 * @param {Array<number>} groupLibraryIDs - Array of group library IDs to show at the top of groups
 	 */
-	async setTopGroup(libraryIDs, headerLabel) {
-		this._topGroupConfig = { libraryIDs, headerLabel };
-		let selected = this.getRow(this.selection.focused);
-		console.log(this.selection.focused, selected);
-		await this.refresh();
-		if (selected) {
-			await this.selectByID(selected.id);
-			this.tree.invalidate();
+	async setCitedGroup(groupLibraryIDs) {
+		this._citedGroupLibraryIDs = new Set(groupLibraryIDs);
+		for (let [index, row] of this._rows.entries()) {
+			if (row.isLibrary(true) && !this._citedGroupLibraryIDs.has(row.ref.libraryID) && row.isOpen) {
+				await this.toggleOpenState(index);
+			}
 		}
+		await this.refresh();
 		this.selection.selectEventsSuppressed = false;
 	}
 
@@ -351,7 +349,7 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 		// For non-userLibrary/feed items that are drawn under headers
 		// we do not draw the arrow and need to move all items 1 level up
 		if (Zotero.isMac && !treeRow.isHeader() && !treeRow.isFeed() && treeRow.ref
-			&& (treeRow.ref.libraryID != Zotero.Libraries.userLibraryID || this._topGroupConfig?.libraryIDs.includes(treeRow.ref.libraryID))) {
+			&& treeRow.ref.libraryID != Zotero.Libraries.userLibraryID) {
 			depth--;
 		}
 		// Ensures the feeds and separator rows have no padding
@@ -433,11 +431,16 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 		cell.appendChild(twisty);
 		cell.appendChild(icon);
 		cell.appendChild(label);
-		div.appendChild(cell);
-
-		if (treeRow.isSeparator() && treeRow.ref?.bottomBorder) {
-			cell.classList.add("bottom-border");
+		
+		// Add "Cited" badge for cited groups
+		if (treeRow.isLibrary(true) && treeRow.ref && this._citedGroupLibraryIDs.has(treeRow.ref.libraryID)) {
+			let citedBadge = document.createElement('span');
+			citedBadge.className = 'cited-badge';
+			document.l10n.setAttributes(citedBadge, 'integration-citationDialog-cited-label');
+			cell.appendChild(citedBadge);
 		}
+		
+		div.appendChild(cell);
 		
 		// Accessibility
 		div.setAttribute('aria-level', depth+1);
@@ -547,29 +550,10 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 			let libraryIncluded, groupsIncluded, feedsIncluded;
 
 			//
-			// Selected groups moved to the top (e.g. in citation dialog)
-			//
-			if (this._topGroupConfig) {
-				let groupHeader = new Zotero.CollectionTreeRow(this, 'header', {
-					id: "top-group-header",
-					label: this._topGroupConfig.headerLabel,
-					libraryID: -1
-				});
-				newRows.splice(added++, 0, groupHeader);
-				for (let groupID of this._topGroupConfig.libraryIDs) {
-					let type = groupID == Zotero.Libraries.userLibraryID ? 'library' : 'group';
-					newRows.splice(added++, 0, new Zotero.CollectionTreeRow(this, type, Zotero.Libraries.get(groupID), 1, true));
-					added += await this._expandRow(newRows, added - 1);
-				}
-				newRows.splice(added++, 0, new Zotero.CollectionTreeRow(this, 'separator', { bottomBorder: true }, 0));
-			}
-
-			//
 			// Add "My Library"
 			//
 			libraryIncluded = this._includedInTree({ libraryID: Zotero.Libraries.userLibraryID });
-			let libraryInTopGroup = this._topGroupConfig?.libraryIDs.includes(Zotero.Libraries.userLibraryID);
-			if (libraryIncluded && !libraryInTopGroup) {
+			if (libraryIncluded) {
 				newRows.splice(added++, 0,
 					new Zotero.CollectionTreeRow(this, 'library', Zotero.Libraries.userLibrary));
 				newRows[added - 1].isOpen = true;
@@ -578,9 +562,16 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 			
 			// Add groups
 			var groups = Zotero.Groups.getAll();
-			// Don't include special groups already placed at the top
-			if (this._topGroupConfig) {
-				groups = groups.filter(group => !this._topGroupConfig.libraryIDs.includes(group.libraryID));
+			// Sort cited groups to the top if configured
+			if (this._citedGroupLibraryIDs.size > 0) {
+				groups.sort((a, b) => {
+					let aIsCited = this._citedGroupLibraryIDs.has(a.libraryID);
+					let bIsCited = this._citedGroupLibraryIDs.has(b.libraryID);
+					if (aIsCited && !bIsCited) return -1;
+					if (!aIsCited && bIsCited) return 1;
+					// Both cited or both not cited - sort alphabetically
+					return a.name.localeCompare(b.name);
+				});
 			}
 			groupsIncluded = groups.some(group => this._includedInTree(group));
 			if (groups.length && groupsIncluded) {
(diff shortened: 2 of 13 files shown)


abaevbog:
@adomasven, is this close to what what you had in mind for the cross-library citation warning?

chrome/content/zotero/integration/citationDialog.js
```diff
@@ -1218,6 +1249,53 @@ const IOManager = {
...
+		let citedItemsByID = {};
+		if (io._citationsByItemIDPromise) {
+			citedItemsByID = await io._citationsByItemIDPromise;
+		}
+		let citedItems = Object.keys(citedItemsByID).map(id => Zotero.Cite.getItem(id));
+		// If the document already has cited items from multiple libraries, or not cited items at all, do nothing
+		let citedLibraries = new Set(citedItems.map(item => item.libraryID).filter(l => l));
+		if (citedLibraries.size !== 1) return true;
+		// If the items being added belong to the same library as all cited items, do nothing
+		let citedLibraryID = citedLibraries.values().next().value;
```
adomasven: Won't this return the libraryID of the newly added citation if it's being cited as the first item of the first citation in the document, i.e. being citation insert before all other citations in the doc?

adomasven:
Yep, I had something like this in mind, although as I said in the issue, I'm not sure if this is the best/only UX solution we should do for this, but it's probably a good start.

abaevbog:
I guess one concern here is that if you frequently cite from multiple groups, and don't care about collaborators and such, will it be annoying that this warning will appear on every new document? It will appear only on the first cross-library citation attempt but still. We could add a checkbox to the warning dialog to never show it again but the flip-side is that one can silence the warning once, and then in the distant future end up not be warned when it would have been applicable.

abaevbog:
Now that I have tested this more, though, I feel like the warning itself isn't particularly helpful...

Suppose you cited only from Group A in a document, and you want to cite an item, say, opened in the reader tab that belongs to Group B. You select it, get a warning that it's not recommended, and ... then what? Do you go back to the main window, drag the item from group B to group A, and then go back to the citation dialog to cite that copied item from group A? Feels like a lot of extra steps.

I wonder if we should also add an option in the warning dialog to copy the item to group A and cite that instead? Essentially, it would save the user the trip to the main Zotero window and make the process a bit smoother.

In that case, though, the warning could potentially appear multiple times for the same document, which is certainly more annoying. Could we also consider adding a document-level pref of a "Target Group"? If that is set, and one is about to cite an item from a different library, we could copy the item to the "Target Group" and cite that right away without showing the warning dialog. I think it might also help with some of the issues the citation explorer is meant to fix, since it should make it more likely that all items come from the same group (when it is applicable)?

@adomasven, @dstillman, what do you think about this idea?

abaevbog:
Along the same lines, when the user is adding an unlinked item from the "Cited" section, we could suggest to save the item to their library, or just save it right away if the pref is set. That way, it should reduce likelihood that one will end up with a bunch of unlinked items in citations.

dstillman:
> Do you go back to the main window, drag the item from group B to group A, and then go back to the citation dialog to cite that copied item from group A?

I think so? Copying is within scope for the Citation Explorer because the whole point of that is to help get you out of a jam, but it's still almost always a mistake to be trying to cite items from different libraries in the citation dialog, and adding copying here (even with a warning) would encourage a haphazard approach to library item management. If you cited from the wrong library originally, you should use Citation Explorer to fix your existing citations, and then you won't get this warning going forward. This dialog can have a button to open Citation Explorer.

But we also don't need to wait for a selection to show a warning. Once you cite from a given library in a document, we should probably prioritize that library in the citation dialog and de-emphasize all the others — at least by ordering the used library before others, but maybe also with some sort of flag on other libraries/items… @yexingsha might have some ideas.

And if you've already cited clicked through this warning and cited from multiple libraries, I guess we'd prioritize all of the used libraries, but we could also show some sort of warning indicator in the citation dialog itself that, when clicked, would show a similar warning and link out to Citation Explorer.

I guess the only question before we get too deep here is whether there genuinely are use cases where people intentionally cite from multiple libraries. Maybe, but I'm not sure what they would be…

abaevbog:
> I think so? Copying is within scope for the Citation Explorer because the whole point of that is to help get you out of a jam,

Gotcha! That makes sense. My thinking above was about how to try to proactively prevent people from getting into a jam to begin with. But it indeed would lead to questions as to which collections to place copied items, and etc, so perhaps we can leave it up to the user. All of this was heavily inspired by the [forums post](https://forums.zotero.org/discussion/127786/add-new-advanced-option-to-avoid-common-mistake-in-collaborative-documents) linked in the issue.

> Once you cite from a given library in a document, we should probably prioritize that library in the citation dialog and de-emphasize all the others

That is a very good idea. In list mode, the cited library can appear at the top, above all others. Not sure about the library mode... We could sort the group to the very top, so it appears even above "My Library", maybe? We can also arrange open and cited items to have the correct group first.

But all of the above does run into the issue that cited items in the document can take a while to load, especially with google docs and if the document is large. So when the citation dialog is first open, we may not always immediately know what the cited items are, and therefore can't really check what library all cited items are from ... Correct me if I'm off on this. Having a desired library saved in the documents preference would be one way to work around it. Or we would have to wait for `io` to load before allowing the interactions with the dialog - but that was a source of bad user experience in the past.

> I guess the only question before we get too deep here is whether there genuinely are use cases where people intentionally cite from multiple libraries. Maybe, but I'm not sure what they would be…


That's a big question. From what I have seen, it is rarely intended and most commonly people are trying to get everything into one group, not the other way around. That said, everyone's approach to organization is unique, so who knows? For example, in a scenario like https://forums.zotero.org/discussion/comment/499186/#Comment_499186, one could have sources in two different groups to make searching more manageable? The later, though, could be addressed with other ways (like adding a tag filter in citation dialog), so I feel like pushing users to citing from a single library is not a bad idea.

dstillman:
> Not sure about the library mode... We could sort the group to the very top, so it appears even above "My Library", maybe?

I could imagine sorting the active group to the top and showing a blue star next to it, and then grouping others under an "Other Libraries" header (and removing the "Group Libraries" header), or something like that.

> Having a desired library saved in the documents preference would be one way to work around it.

Yeah, just maintaining the list of used libraries in the doc prefs makes sense. And if it's not present (e.g., for an existing document), we don't do any prioritization unless we get the cited items immediately, I guess.

abaevbog:
Sounds good! I think allowing one to specify the desired library will very much serve as a preventative measure that I initially had in mind.

dstillman:
To clarify, I was saying we would automatically maintain the list of used libraries, not that we would expose it as a distinct pref or let users choose it explicitly. The user's choice is implicit based on their first selection (and if they messed up, that's what the Citation Explorer is for).

abaevbog:
> To clarify, I was saying we would automatically maintain the list of used libraries, not that we would expose it as a distinct pref or let users choose it explicitly. 

Gotcha, understood!

adomasven:
> But all of the above does run into the issue that cited items in the document can take a while to load, especially with google docs and if the document is large. So when the citation dialog is first open, we may not always immediately know what the cited items are, and therefore can't really check what library all cited items are from ... Correct me if I'm off on this. Having a desired library saved in the documents preference would be one way to work around it. Or we would have to wait for `io` to load before allowing the interactions with the dialog - but that was a source of bad user experience in the past.

`Zotero.Integration` maintains a list of session objects that correspond to each document that the user has used with Zotero plugins for citing (mapped via unique IDs in doc prefs). We store/cache this kind of information there.

We could technically store it in document preferences, but we've only ever stored the necessary information to create/format citations. I don't think it's a good idea to store something that may potentially change based between Zotero interactions with the document there.

--------

I agree that the citation dialog shouldn't do automatic user item copying, moving, etc. We don't ever do anything like this in other parts of the UI, and the users should have full ownership and responsibility of organizing their library in a way that makes sense to them, and also allows for good citing behaviour.

If the user is citing an item from the wrong library, they may find that they have saved a bunch of items to the wrong one that they were planning to use, or perhaps they are now writing a shared document, and intended to use some items from their earlier work in personal library. They should copy those items into the group library, and share them with the rest of collaborators.

We risk introducing more user confusion by doing these things automatically.

--------

Something to think about here is collaborators that are just citing from their own personal libraries. For one-off projects, especially for students, it's likely not that uncommon. The items cited by collaborators are unlinked for you.

But also, items that you may have accidentally deleted from your library are also unlinked for you. Will those trigger this warning? Should they? Should we warn users in the citation dialog when that occurs and point them to citation explorer?

abaevbog:
> Zotero.Integration maintains a list of session objects that correspond to each document that the user has used with Zotero plugins for citing (mapped via unique IDs in doc prefs). We store/cache this kind of information there.

Cool, thank you for the pointer. `Zotero.Integration.sessions` does seem like a better place for this kind of metadata/cache, and we can save the cited library (if one exists) as another variable there, say, after the citation dialog closes. Then, it will be available for all subsequent citation dialog invocations.

The only question there is how do we handle the situation when the session does not exist yet, and a new one is created before opening the citation dialog (e.g. after restart, or when citing in a document that already has some citations for the first time)? Then, the cited library wouldn't yet be available to send to the citation dialog, right? My thought is that maybe in that case, it is OK to wait for `io` to be fully loaded before showing the dialog, perhaps with an additional loading message of some kind, highlighting that it only needs to happen on the first run?

dstillman:
> We could technically store it in document preferences, but we've only ever stored the necessary information to create/format citations. I don't think it's a good idea to store something that may potentially change based between Zotero interactions with the document there.

Why not? It's not any different from adding a visible "Target Library" setting to the doc prefs — it's just implicit. We'd only need to update 1) on the first citation write and 2) if a citation from another library was ever added.

Otherwise, as @abaevbog says, for a new Zotero session we have a bad experience where we can't show the preferred library until items are loaded.

abaevbog:
> Something to think about here is collaborators that are just citing from their own personal libraries. For one-off projects, especially for students, it's likely not that uncommon. The items cited by collaborators are unlinked for you.

Right. I imagine they get an alert and are prompted to move the item from "My Library" to the group? But then a valid question is what happens if one doesn't have permissions to edit the group? Do we just not show the alert, and expect someone who does have permissions to use citation explorer to move unlinked items to the group later?

> But also, items that you may have accidentally deleted from your library are also unlinked for you. Will those trigger this warning? Should they? Should we warn users in the citation dialog when that occurs and point them to citation explorer?

At the moment, unlinked items are just not considered in the alert handling. So if you add one citation with one item, and then delete that item, next time you add a new citation, you get no warning, since the unlinked item doesn't have a `.libraryID`. We could totally warn people about unlinked items - I'm not sure about the exact context though. We could show it once per session, saying "It looks like some cited items are missing in your libraries. Use citation explorer to fix it"? I suppose maybe - it would somewhat address the scenario described above. If someone doesn't have access to a group that is being cited in the document and cites something from "My Library", then next time you open the doc, you will be prompted to add whatever items they cited into the group via citation explorer?

adomasven:
> > Something to think about here is collaborators that are just citing from their own personal libraries. For one-off projects, especially for students, it's likely not that uncommon. The items cited by collaborators are unlinked for you.
> 
> Right. I imagine they get an alert and are prompted to move the item from "My Library" to the group? But then a valid question is what happens if one doesn't have permissions to edit the group? Do we just not show the alert, and expect someone who does have permissions to use citation explorer to move unlinked items to the group later?

Well, I was talking about different collaborators citing from their own personal library, so anything you cite from yours, is effectively "unlinked" for me. But it also applies to group libraries that one of the collaborators is not a part of.

We probably should just treat all unliked items the same, and perhaps deal with them in a separate PR, but e.g. displaying an icon in their bubble in the citation dialog could be something to do, and another thing is to display a warning symbol somewhere in the dialog, if there are any citations in the document that are unlinked, that takes you to the citation dialog. Both of these actually make sense after citation explorer is merged.

abaevbog:
> Well, I was talking about different collaborators citing from their own personal library, so anything you cite from yours, is effectively "unlinked" for me. But it also applies to group libraries that one of the collaborators is not a part of.

I see, I was thinking if we have a document where I was citing from a group, and then you were to try to cite from "My Library" on your end. In that case, the warning as discussed earlier seems applicable. But if we are both citing from our respective "My Libraries" in one document, it's trickier, since what would the alert even say? Maybe we do nothing for now, and later add an alert prompting to open the citation explorer, when it is ready.

I ran into one edge case that doesn't quite come together with the setup we have been discussing. Suppose I add several citations from the same group into the document - it will record in the document data the groupID as `citedLibraryID`. Then I (or another user) add(s) one more citation from another group or "My Library", selecting to continue when the warning appears. Then, `citedLibraryID` is cleared, since there are now multiple cited groups in the document. Next, suppose I delete that one citation in word processor.

Ideally, in this case, we would want to update `citedLibraryID` in document data, since now all citations belong to the same group again. But there is no way to do that, unless the user presses "Refresh" button in the toolbar, is there? That means, in that scenario, next time the citation dialog opens, it will look as if there is no cited library preference based on outdated `citedLibraryID`. I assume it's not that big of an issue but still worth highlighting.

adomasven:
> Ideally, in this case, we would want to update `citedLibraryID` in document data, since now all citations belong to the same group again. But there is no way to do that, unless the user presses "Refresh" button in the toolbar, is there? That means, in that scenario, next time the citation dialog opens, it will look as if there is no cited library preference based on outdated `citedLibraryID`. I assume it's not that big of an issue but still worth highlighting.

Right. This was what I meant when saying that this is data that may get outdated when the user edits the document, hence why I don't think we should store it in document preferences, on the other hand, it's better to have outdated data than none at all, since in 99% of the cases it will be correct. I think we should update the UI every time we get all fields from the document.

Regarding `citedLibraryID`, I would think it should be `citedLibraryIDs`. Moreover, this warning should probably apply every time you try to add a citation from a new library to the document. I don't think we should encourage users to cite from multiple libraries, but it's ultimately their choice, and if they do so, the 2 libraries that they are citing from should still be sorted at the top.

abaevbog:
@adomasven thank you for the notes! I just pushed the revised version. This is the behavior we currently have:
- Cited libraries are saved in the document like preferences. I am saving the URI of the libraries, and their name. Using URIs allows us to distinguish between "My Library"-s of different users, and group names allow us to display group names in the warning dialog even if the other user does not have access to the other group.
- `Zotero.Integration.refreshCitedLibraries` is responsible for recording all cited libraries in the document. It is called after citing and in `updateFromDocument`.
- After one saves an item from the group, that group will appear at the top of the collectionTree.
Screenshot 2025-11-21 at 2 20 21 PM

- In list mode, cited libraries will be sorted to the top above all other libraries (though still below selected/open/cited sections)
- If one tries to save an item that belongs to another group, the warning will have names of cited libraries (or usernames, if cited from "My Library" of another user as in the example below):
Screenshot 2025-11-21 at 2 21 23 PM
- If there are multiple cited groups, they all appear at the top of collectionTree, and the warning will not appear for those groups:
Screenshot 2025-11-21 at 2 20 54 PM
- In a scenario where the recorded cited libraries are outdated, `refreshCitedLibraries` will still eventually record cited libraries when fields are fetched, and the warning will appear on `accept`.
- Finally, it's worth noting that if two users are collaborating on the same dialog, and one of them has an older version of Zotero without these changes, naturally, the older version will override whatever cited libraries are saved to document prefs.


What do you think about this? I'll do some more testing between different accounts to see if any more edge cases appear.

---

@yexingsha what do you think about the collection tree setup? I think the header that says "Cited in this document" is important for context but we may want a different icon?

yexingsha:
I think we can add a "Cited" label to cited libraries. This does not change the structure or order of the collection tree, and works for the list mode as well:

library

list

In library mode, when the row is focused, use `#FFFFFF33` as the label's background color, and `accent-white` as its text color.

abaevbog:
> This does not change the structure or order of the collection tree, and works for the list mode as well:

I feel like changing the order of collection tree is a significant part of the idea here... In list mode, after selected/open/cited sections, the cited libraries go first before remaining not-cited libraries. Similarly, it feels appropriate to make cited libraries "special" by placing them at the top, which implies that these are the libraries you are expected to be citing from. Some people end up having a fairly large number of groups (e.g. several tens), in which case it's also more useful to have the currently-cited libraries at the top vs having to scroll through the entire list to locate them.

adomasven:
> * If one tries to save an item that belongs to another group, the warning will have names of cited libraries (or usernames, if cited from "My Library" of another user as in the example below):

Where are we getting usernames? Sync usernames? Not everyone is synced, although I guess in those cases there won't be any group libraries, so maybe an irrelevant consideration.

> - In a scenario where the recorded cited libraries are outdated, `refreshCitedLibraries` will still eventually record cited libraries when fields are fetched, and the warning will appear on `accept`. 

I don't think we should do this. The recorded cited libraries are technically updated between every interaction, since the user or a collaborator may have edited the document in between interactions even if Zotero wasn't shut down. We should only show the warning on accept when we have no cited libraries value at all.

> - Finally, it's worth noting that if two users are collaborating on the same dialog, and one of them has an older version of Zotero without these changes, naturally, the older version will override whatever cited libraries are saved to document prefs.

We should cache this in the integration session and fall back on that in case it's missing from the doc between citation dialog calls.

abaevbog:
> Where are we getting usernames? Sync usernames? Not everyone is synced, although I guess in those cases there won't be any group libraries, so maybe an irrelevant consideration.

Yes, that's right. I assume that in most cases, `Zotero.Users.getName` will be able to give us the names of users who cited from their "My Library". But I also tried to account for the scenario if there are users that don't share any groups or don't sync at all (when `Zotero.Users.getName` would be undefined).

If user A cites in the document, cited libraries in document prefs will be recorded as `{ "http://zotero.org/users/..." : "user_A" }`. Then, if user B comes along and tries to cite, they'll get the "user_A" stored value from document prefs in the warning. I don't know how important it is to warn about the other user's citations from their libraries, since in this case "user B" doesn't really have any way to fix it. But it may prompt "user B" to check in with "user A" to create a group to collaborate in, which is the best practice we are trying to push here? 

>  The recorded cited libraries are technically updated between every interaction, since the user or a collaborator may have edited the document in between interactions even if Zotero wasn't shut down. We should only show the warning on accept when we have no cited libraries value at all.

But what about the scenario we discussed earlier, when there may be 2 cited groups recorded in document prefs, and then all citations from one group are removed? When citation dialog is open, document prefs would still have 2 groups. It may take a bit for all fields to be loaded, so it's possible that you would add an item from the no-longer-cited group without getting any warning. Then you would press Enter, the dialog waits for all fields to load, and cited libraries are updated. Currently, there will follow a warning saying that the library is not cited - any reason why we wouldn't want it?

For reference, this is the recording of the scenario I have in mind:

https://github.com/user-attachments/assets/75694099-dfec-4320-b249-290c0d83febf

> We should cache this in the integration session and fall back on that in case it's missing from the doc between citation dialog calls.

If I cite in a document as user A, and then cite in it as user B, next time I cite as user A, a new session will be created (since the sessionID in the document was last set by user B). So I am under the impression that it's not about caching it in the session (which is already happening in `Zotero.Session.data.citedLibraryURIs`) but reusing the existing session if the document data is changed by another user, maybe? I'll see what we can do here.

abaevbog:
> We should cache this in the integration session and fall back on that in case it's missing from the doc between citation dialog calls.

Correct me if I'm wrong on this but having looked into the session management a bit more, my understanding is that all integration commands are completely independent of each other. The way we can maintain one active session across several cite actions is by saving `sessionID` into document data, and then fetching the session based on that ID on the next command [here](https://github.com/zotero/zotero/blob/49325e27d9b76094ea83e9106746ef8fbe198593/chrome/content/zotero/xpcom/integration.js#L568).

If so, when another user with an older version of Zotero updates document data, cited libraries won't be there, and also `sessionID` will be new. It will prompt us to create a new session and calculate cited libraries from scratch, potentially a few seconds after the citation dialog opens (if fields take a while to load).

My intention behind saving cited libraries into document data was to avoid exactly the scenario above. We had previously ran into one edge case, when cited libraries document data would be stale after removing all citations from a specific group. This is another edge case that, I assume, will be even more common, since not everyone updates Zotero frequently, and even now there is a number of users on Zotero 6.

To account for these issues, we have to reload citation dialog UI when document fields are loaded to account for any inaccuracies. I pushed a tweak that will reload the "Cited groups" section of collectionTree, so it is accurate whenever `io` is ready.

With this, I myself am starting the wonder if saving cited libraries in document data is really worth it (vs just saving data in the session), given that we have to double-check info from document data and reload the citation dialog UI anyways? The only instance when document data approach really helps is if you are opening a document after, say, restarting Zotero AND there are no collaborators who worked on this document. But if you have no collaborators, citing from different libraries is not as much of a concern to begin with.

adomasven:
> But what about the scenario we discussed earlier, when there may be 2 cited groups recorded in document prefs, and then all citations from one group are removed? When citation dialog is open, document prefs would still have 2 groups. It may take a bit for all fields to be loaded, so it's possible that you would add an item from the no-longer-cited group without getting any warning. Then you would press Enter, the dialog waits for all fields to load, and cited libraries are updated. Currently, there will follow a warning saying that the library is not cited - any reason why we wouldn't want it?

Right, but we don't really have a state where we know that the cached state is correct. Say user 1 adds a group library, and also keeps items cited from user libraries of 2 collaborators. Then, user 2 attempts to edit the doc, and adds an item from the already-added group library. Do they get the warning immediately?

So perhaps we treat the cache as something that we use for visual updates in the dialog, but ultimately we only show the warnings once field data comes in from the doc?

> > We should cache this in the integration session and fall back on that in case it's missing from the doc between citation dialog calls.
> 
> If I cite in a document as user A, and then cite in it as user B, next time I cite as user A, a new session will be created (since the sessionID in the document was last set by user B). So I am under the impression that it's not about caching it in the session (which is already happening in `Zotero.Session.data.citedLibraryURIs`) but reusing the existing session if the document data is changed by another user, maybe? I'll see what we can do here.

Uh, I hadn't considered this. We could probably reuse the existing sessionID when loading a document with existing document data. There's some implications for that specifically when transferring docs between doc processors, but it should be handled fine since sessionID is specifically reset in those cases.

> With this, I myself am starting the wonder if saving cited libraries in document data is really worth it (vs just saving data in the session), given that we have to double-check info from document data and reload the citation dialog UI anyways? The only instance when document data approach really helps is if you are opening a document after, say, restarting Zotero AND there are no collaborators who worked on this document. But if you have no collaborators, citing from different libraries is not as much of a concern to begin with.

I think if we keep sessionID stable, it's still worth it. When you reload Zotero, assuming you're working with the document on your own, you immediatelly get UI feedback for cited libraries. And in cases where you're working with a collaborator with an older version, and you cite one after another, you can reload from your session cache (assuming the sessionID remains stable).

abaevbog:
> Right, but we don't really have a state where we know that the cached state is correct. Say user 1 adds a group library, and also keeps items cited from user libraries of 2 collaborators. Then, user 2 attempts to edit the doc, and adds an item from the already-added group library. Do they get the warning immediately?

The initial idea was that there would be no warning in this case.  After citing from the group, `{ user_1_uri: user_1_name, user_2_uri: user_2_name, groupURI: groupName }` would be the new recorded `Zotero.Integration.session.data.citedLibraryURIs`. So if another user tried to cite from the just-cited group again, `groupURI` value would be used as the indication that the group is cited, it would appear at the top of collection tree, and there would be no warning.

Currently, that doesn't quite work because the session is reset between citations of different users - but that was the intention. The behavior now is still very close to this though. After an item is added by another user, cited libraries cache for the document session won't exist. The cross-library check would run when the item is about to be added into `bubble-input` or before the dialog is submitted in `accept` - depending on when the document fields are loaded. In both cases, since an item is already cited from the group, cross-library check would yield no warning.

> So perhaps we treat the cache as something that we use for visual updates in the dialog, but ultimately we only show the warnings once field data comes in from the doc?

One could say that this is sorta how it already works, in a way. Cached cited libraries values fetched via ` io.getCitedLibraryInfo()` from the dialog are used for the UI and warnings in the very beginning when the dialog opens - until the fields are loaded. Then, when the fields are fetched, cited libraries record is updated, which will refresh the UI and determine future warnings. If the user tries to accept the dialog before all fields are loaded, there's a final check on `accept`. I feel like the warnings and UI need to be as synchronized as possible, otherwise it may be confusing for the user why the warning appears in some cases for non-cited groups but not others.

> Uh, I hadn't considered this. We could probably reuse the existing sessionID when loading a document with existing document data.

Hm, how can we find out which of our locally recorded `Zotero.Integration.sessions` should be used in this case? If I am working on multiple documents, I will have several sessions `{ ZftUc0MD: { session_one }, c3qJeYyh: {session_two} }`. So, when I try to  cite in a document after another user, how can we know which sessionID to pick?

abaevbog:
For reference, this is the flow of the logic we have at the moment when the citation dialog opens:
1. fetch cited libraries from document data via `io.getCitedLibraryInfo()`. If provided, move those libraries to the top of collectionTree, prioritize those libraries in list mode results, show warning if one tries to add an item belonging to a library not in that list. It includes user libraries to discourage citing from multiple user libraries in the same doc.
2. when the fields load, update cited libraries cache, and refresh the UI based on that (collectionTree + results in list mode). Future warnings will check against the updated libraries cache too.
3. if the user presses Enter before the the fields are loaded, progress indicator will appear. Before `io.accept` is called, we'll wait for the fields to be loaded and run the cross-library check to potentially display a warning, in case if cached cited libraries are, in fact, no longer cited. If not accepted, the dialog goes back to it's original state
4. if one accepts a cross-library citations warning, it will no longer appear. So if you add a citation from one new group and accept the warning, if you were to go cited from another new group, there would be no warning. (I couldn't think of a the exact use case here but figured if you accept the warning ones, no need to bother the user with it again).

So whatever the cited libraries cache in document data is, it's always double-checked by actual loaded do fields, whenever they are ready. In the scenario when cited libraries document data is erased by the other user, cited libraries appear in the dialog when fields are fetched.

yexingsha:
> Some people end up having a fairly large number of groups (e.g. several tens), in which case it's also more useful to have the currently-cited libraries at the top vs having to scroll through the entire list to locate them.

Can't we just sort the cited libraries to the top of the group libraries? I think I would prefer to keep the My Library / Group Libraries structure consistent, and also use the same visual indicator for both library mode and list mode. But that's mostly an aesthetic preference.

If we do put the cited libraries into a group at the top, we can use `16/universal/cite.svg` for the icon and color it `accent-blue`, and put 8px spacing between the last library and the divider.

Also, this just occurred to me: can we put a "Don't warn me again in this document" checkbox in the dialog, and an equivalent one in Document Preferences? So that when people don't write with a collaborator, they can just cite from wherever they want.

adomasven:
> The initial idea was that there would be no warning in this case. After citing from the group, `{ user_1_uri: user_1_name, user_2_uri: user_2_name, groupURI: groupName }` would be the new recorded `Zotero.Integration.session.data.citedLibraryURIs`. So if another user tried to cite from the just-cited group again, `groupURI` value would be used as the indication that the group is cited, it would appear at the top of collection tree, and there would be no warning.

Right, I guess this is only an issue when User 1 is using a previous version.

> Currently, that doesn't quite work because the session is reset between citations of different users - but that was the intention. The behavior now is still very close to this though. After an item is added by another user, cited libraries cache for the document session won't exist. The cross-library check would run when the item is about to be added into `bubble-input` or before the dialog is submitted in `accept` - depending on when the document fields are loaded. In both cases, since an item is already cited from the group, cross-library check would yield no warning.

There are other ways for things to get out-of-sync though. If another user copies citations into the document from another doc, and the items there are from yet another group, or from their personal library, document data AND session cached data are both out of sync and you wouldn't know until you fetched the fields. In that case you may cite from a library that is "used" in the document, but then after the fields are fetched it turns out it no longer is. When do we warn about that, and more importantly, how do we do it, because our UI said that the library is actually cited.

I'm not too concerned if the warning in this case is confusing to the user, since this is quite an extreme edge case, but we should make sure we know what the edge cases are here and what we do with them.

> Hm, how can we find out which of our locally recorded `Zotero.Integration.sessions` should be used in this case? If I am working on multiple documents, I will have several sessions `{ ZftUc0MD: { session_one }, c3qJeYyh: {session_two} }`. So, when I try to cite in a document after another user, how can we know which sessionID to pick?

No, I mean the current `Integration` code assigns a new session id and stores it to the document every time you restart Zotero or another collaborator edits the document. We should not assign a new session id to the document once it's been assigned an initial one.

> If user A cites in the document, cited libraries in document prefs will be recorded as `{ "http://zotero.org/users/..." : "user_A" }`. Then, if user B comes along and tries to cite, they'll get the "user_A" stored value from document prefs in the warning. I don't know how important it is to warn about the other user's citations from their libraries, since in this case "user B" doesn't really have any way to fix it. But it may prompt "user B" to check in with "user A" to create a group to collaborate in, which is the best practice we are trying to push here?

> 4\. if one accepts a cross-library citations warning, it will no longer appear. So if you add a citation from one new group and accept the warning, if you were to go cited from another new group, there would be no warning. (I couldn't think of a the exact use case here but figured if you accept the warning ones, no need to bother the user with it again).

> Also, this just occurred to me: can we put a "Don't warn me again in this document" checkbox in the dialog, and an equivalent one in Document Preferences? So that when people don't write with a collaborator, they can just cite from wherever they want.

And I guess I want to circle back to this. We should warn if there are citations in the document from unknown group libraries, as well as different user libraries. We should not encourage users to cite from their personal libraries when collaborating, or from multiple different groups either. This creates all kinds of confusion that is not that easy to fix once you've done this related to metadata edits, as well as duplicate in the bibliography.

But we need to make sure we don't do it every time the user restarts Zotero, if they decide to work with a document like that. You cannot store whether the user saw the warning in the document preferences, since this is a per-user thing. Unless we store user prefs with their unique ID or something. But with 50 collaborators that may mean huge document prefs. And that would be fine, except that on Google Docs it may make things very slow. I need to check, but ideally we'd do it differently. Another alternative is to store that in the database or on disk as a pref.

> In the scenario when cited libraries document data is erased by the other user, cited libraries appear in the dialog when fields are fetched.

Do we not use the session cached data in this case?

abaevbog:
Thank you for the feedback here! Also, happy belated thanksgiving (to those in the US).

@yexingsha, I refactored the collection tree setup to sort cited groups to the top, while keeping them together in the "Group Libraries" section. There is also a blue "Cited" badge added to their rows. "My Library" in collection tree remains where it was, it just gets its own "Cited" badge when relevant. Does this look better?

Screenshot 2025-12-01 at 6 20 14 PM
Screenshot 2025-12-01 at 6 20 57 PM

I don't have a particularly strong opinion about this vs moving cited groups to the very top of collection tree. One benefit of this setup is that we don't have to deal with weird positioning of "My Library" if it is not cited. On the other hand, placing all cited groups together at the top (and moving "My Library" lower) may be more effective in terms of deterring users from adding a new citation from uncited "My Library". Another issue now is that if "My Library" is expanded, its collection push all groups lower, so you may not even see the "Cited" badge when the dialog opens. To combat this, I thought we can collapse all uncited groups, including "My Library" - it works but it also effectively discards user's old selection, but it's probably appropriate here anyways.

There is also a question of consistency: before, uncited "My Library" would be moved, which looks very inconsistent with the collection tree of the main window. Now, uncited "My Library" in collectionTree remains at the top, though it will not appear at the top in list mode (it will appear after cited groups). So it's a question of which inconsistency sounds less bad, I suppose.

yexingsha:
This looks great!

> To combat this, I thought we can collapse all uncited groups, including "My Library" - it works but it also effectively discards user's old selection, but it's probably appropriate here anyways.

I agree that this should mitigate most of the issues. As to consistency, since the collection tree is a major part of the main window, whereas the libraries that appear in list mode can vary from search to search anyway, I think the collection tree consistency is more important here.

abaevbog:
> No, I mean the current Integration code assigns a new session id and stores it to the document every time you restart Zotero or another collaborator edits the document. We should not assign a new session id to the document once it's been assigned an initial one.

@adomasven, thank you for the explanation here. I added a tweak here to use the session ID that currently exists in the document if we are creating a new session. That way, even if there is another collaborator with an older Zotero version, we'll just stick with the sessionID they set, and will not set a new random sessionID on the next `Zotero.Integration.execCommand` call. When updating session from document data, if document data has no cited libraries record, we'll keep the cited libraries record in the session. This way, on subsequent integration commands, we'll be able to use the old session cache, even if the user with old Zotero version updates document data in between.

>> In the scenario when cited libraries document data is erased by the other user, cited libraries appear in the dialog when fields are fetched.

> Do we not use the session cached data in this case?

When I was describing this scenario, the session ID would be new, so we wouldn't reliably know which session cached data to use. Now that we try to keep sessionID, it is much more likely we'll find the session with cached data to fallback on. There are still edge cases when it would not work perfectly (e.g. with multiple collaborators who have old Zotero) - but it's certainly better.

> In that case you may cite from a library that is "used" in the document, but then after the fields are fetched it turns out it no longer is. When do we warn about that, and more importantly, how do we do it, because our UI said that the library is actually cited.

Hmm, copy-pasted citations is indeed something I did not consider. Currently, this is what would happen: first, the dialog opens and shows cached cited libraries in collectionTree/list mode search results. If you just do nothing, once `io` loads, we know which libraries are cited, so the collectionTree/list mode will refresh, place "cited" labels on the proper groups, and warnings will appear only for items from actually uncited groups. If you try to add a citation from a cited library of pasted items that was not recorded in integration `session.data` before `io` is loaded, it will show an unwarranted warning.

If you were to delete all records from one group recorded as cited in session data, open the dialog, and try to add an item from that would before `io` is loaded - there will be no warning when a bubble is added into `bubble-input`. A warning will appear when you try to accept the dialog though (better late than never).

A late warning is probably OK - but showing a warning for a library that was not recorded as cited, even though it is after pasting a citation, is pretty bad... And the UI would be confusing. I'm not sure what a good solution here would be. We can add something like a small spinner into the "Cited" badge that follows cited groups in list mode and collectionTree to indicate that something is still loading, with a tooltip explaining why the spinner is there. And we can tweak the alert wording a bit to indicate that the record of cited libraries has not yet been confirmed, until `io` is loaded?


> But we need to make sure we don't do it every time the user restarts Zotero, if they decide to work with a document like that. You cannot store whether the user saw the warning in the document preferences, since this is a per-user thing.

I wonder if this could be considered a per-document (vs per-user) preference? Suppose I have collaborators who diligently tried to cite from a single group. Then suppose I select to silence the warnings, and proceed to cite from a bunch of different groups. The citations in the document would be somewhat messed up because of me, and would likely have to be resolved using citation explorer later on. So if a single person selects to silence the cross-library citation warnings, it is likely to affect everyone working on the document, so why not keep track of it on the document level? There would be a question of how to communicate that distinction to other collaborators though.

Alternatively, I would imagine we could save `sessionID`s of selected documents in a `Zotero.Prefs` field, and suppress warnings based on that. We would just need a mechanism to ensure that old sessionIDs are cleaned up after a while.

adomasven:
> Hmm, copy-pasted citations is indeed something I did not consider. Currently, this is what would happen: first, the dialog opens and shows cached cited libraries in collectionTree/list mode search results. If you just do nothing, once `io` loads, we know which libraries are cited, so the collectionTree/list mode will refresh, place "cited" labels on the proper groups, and warnings will appear only for items from actually uncited groups. If you try to add a citation from a cited library of pasted items that was not recorded in integration `session.data` before `io` is loaded, it will show an unwarranted warning.
> 
> If you were to delete all records from one group recorded as cited in session data, open the dialog, and try to add an item from that would before `io` is loaded - there will be no warning when a bubble is added into `bubble-input`. A warning will appear when you try to accept the dialog though (better late than never).
> 
> A late warning is probably OK - but showing a warning for a library that was not recorded as cited, even though it is after pasting a citation, is pretty bad... And the UI would be confusing. I'm not sure what a good solution here would be. We can add something like a small spinner into the "Cited" badge that follows cited groups in list mode and collectionTree to indicate that something is still loading, with a tooltip explaining why the spinner is there. And we can tweak the alert wording a bit to indicate that the record of cited libraries has not yet been confirmed, until `io` is loaded?

Yes, adding a spinner to cited library status, or just generally somewhere in the dialog as we are fetching citations with an explanation text is not a bad idea not just for this, but in general. Like the "Cited items" section in list mode not appearing until citations are fetched from the document.

I don't think we can avoid these false-negative prompts, although I'm personally more concerned about the confusion arising from citing from what in UI is described as Cited library, only to be greeted by a dialog that it's not cited.

I also don't think that any amount of trying to explain this to an average user in any prompt we could write is going to make them less confused about it. But if someone comes into the forums later asking for explanation we should be able to give it to them.

> > But we need to make sure we don't do it every time the user restarts Zotero, if they decide to work with a document like that. You cannot store whether the user saw the warning in the document preferences, since this is a per-user thing.
> 
> I wonder if this could be considered a per-document (vs per-user) preference? Suppose I have collaborators who diligently tried to cite from a single group. Then suppose I select to silence the warnings, and proceed to cite from a bunch of different groups. The citations in the document would be somewhat messed up because of me, and would likely have to be resolved using citation explorer later on. So if a single person selects to silence the cross-library citation warnings, it is likely to affect everyone working on the document, so why not keep track of it on the document level? There would be a question of how to communicate that distinction to other collaborators though.
> 
> Alternatively, I would imagine we could save `sessionID`s of selected documents in a `Zotero.Prefs` field, and suppress warnings based on that. We would just need a mechanism to ensure that old sessionIDs are cleaned up after a while.

I guess what I said in my previous post is actually asking for a slightly different thing, which is, if User A starts a document and cites items from their library, and then User B joins them, they should get a warning, that this isn't the way to use Zotero and both of them should create a group library. User B can choose to ignore it. When User A returns to the document, we should also show the same warning to them, since now there are citations from User B library. When User C joins and also starts citing, everyone should once again get a warning.

They all should be able to see the warning and ignore it. But then when they restart Zotero, the fact that for that specific document the warning is ignored for that particular user should be restored. I'm not sure if Zotero.Prefs is where we want to store this. Like we store item tree column prefs in a separate file, and we should probably do this for document state of this kind too.

Also, in the scenario above, if a group library is added to the document, we shouldn't warn about it for the first one.

Anyway, this is slightly different and more complicated, and maybe we should check with Dan before proceeding, so I'm happy to create an issue for this for further discussions.

abaevbog:
> Yes, adding a spinner to cited library status, or just generally somewhere in the dialog as we are fetching citations with an explanation text is not a bad idea not just for this, but in general. Like the "Cited items" section in list mode not appearing until citations are fetched from the document

I agree that it might be helpful to clarify this whole situation with warnings. Where to place such loading indicator is a question though... BTW, the "Cited items" section itself is meant to always appear in both modes as long as there is an active search query - if the `io` is not loaded, it just says that the section is loading. If we could always show it while `io` is loading, it might somewhat address this issue actually. But "Cited" section always follows selected and opened items, so it would often be invisible to the user unless they scroll, so probably not.
Screenshot 2025-12-03 at 9 21 32 AM
Screenshot 2025-12-03 at 9 21 28 AM

These are the spots where I though some indication of cited items still being loaded could go:

- The Z icon is not a bad spot, since it's not currently used for anything. Maybe we add a small spinner at the bottom of the Z?
- "Cited" badge in collectionTree (or after group names in list mode) is nice, since it already says "cited", so adding a spinner-type icon would already imply that the loading has something to do with cited items. 
- Finally, we have empty space at the bottom where something like "Cited items are being loaded" message could go. Some of that space is used in #5586 though, so we would need to account for that.

Screenshot 2025-12-03 at 9 08 36 AM

Does any of these options sound better than others? (cc @yexingsha)

> I don't think we can avoid these false-negative prompts, although I'm personally more concerned about the confusion arising from citing from what in UI is described as Cited library, only to be greeted by a dialog that it's not cited.

Fair enough. This might be an argument in favor if putting the spinner after "Cited" badge in collectionTree. We can also display the "Cited" badge as faint-grey-ish until `io` is loaded, and change it to blue once we confirm that the group is indeed 100% cited.

> I guess what I said in my previous post is actually asking for a slightly different thing, which is, if User A starts a document and cites items from their library, and then User B joins them, they should get a warning, that this isn't the way to use Zotero and both of them should create a group library. User B can choose to ignore it. When User A returns to the document, we should also show the same warning to them, since now there are citations from User B library. When User C joins and also starts citing, everyone should once again get a warning.

Gotcha, thank you for the explanation. Yes, that's not exactly the use case I had in mind - currently, user B would actually get a warning saying that all cited items belong to user A if they try to cite from their library. But user A currently gets no warning, unless they try to cite from a different group.

I feel like it's still in the scope of these changes, since they all deal with warnings/accounting for which libraries are cited, and whatever is happening here so far should account for the behavior you described. The outline above will certainly push folks harder to not cite from multiple user libraries. Let me add the following tweaks and see if this feels right:
- a new type of warning will be added that will say something like "This document has cited items from another user's library. To collaborate, make a group and cite from there".
- when the citation explorer is ready, that warning can also have a button to open it, I think
- that warning will appear when you open the citation dialog and the document has items cited from another user's library
- the warning will have two options: close citation dialog or continue/ignore.
- if the warning was ignored, we'll record the sessionID from document data in a separate JSON file, which we'll check before trying to show the warning in the future, in case if Zotero is restarted
- if the document has only cited user libraries, we won't show the original warning if one tries to cite from a group library for the first time

`sessionID` stored in the document is the best way to uniquely identify the document, right? From what I see, the id of the document itself in `Zotero.Integration` code is different on every invocation.

adomasven:
> * that warning will appear when you open the citation dialog and the document has items cited from another user's library

We'd get this from the document preferences? It's probably fine to not make this backwards compatible.

How are we going to account for users who are not synced?

> `sessionID` stored in the document is the best way to uniquely identify the document, right?

Yes, but we should maybe consider making it longer, a UUID or something. Duplicate IDs wasn't a concern before, but here we should make sure they're all unique.

> From what I see, the id of the document itself in `Zotero.Integration` code is different on every invocation.

I'm not sure what you're referring to.

abaevbog:
> We'd get this from the document preferences? It's probably fine to not make this backwards compatible.

Right. The same questions as with the first warning that we discussed are relevant here. We can use the cited libraries cached in document data OR we can wait for `io` to fully load and show the warning only after verifying that items from user libraries actually exist. Or a bit of both, like with the first warning.

> How are we going to account for users who are not synced?

From what I see, when users who are not logged in cite items, the `cslItemURI` item property is still set, and it looks like this `http://zotero.org/users/local/OGF52t5L/items/BSHPT4DB`. We know that it's a user library from the `/users` part of the URI. If we need to warn user A that user B cited from their user library, and user B is not logged in, I think we'll just use a `[Unknown user]` placeholder in the warning?

That said, I'll certainly do some more testing to check the scenario if one cites as a not synced/logged out user, and then syncs their account.

> Yes, but we should maybe consider making it longer, a UUID or something. Duplicate IDs wasn't a concern before, but here we should make sure they're all unique.

Good idea.

>>  From what I see, the id of the document itself in Zotero.Integration code is different on every invocation.

> I'm not sure what you're referring to.

Oh, I was just referring to the `document` that we get in `execCommand` and that is later available in `Zotero.Integration.Sessions[key]._doc`. The `_doc` object there does seem to have an ID but it's not a constant property and changes each time a new session is created. So we'll need to rely on the stored `sessionID` in document data.

https://github.com/zotero/zotero/blob/11223451d27298fdf9f5dc26b20ab2221fe1fce9/chrome/content/zotero/xpcom/integration.js#L271-L279

yexingsha:
> This might be an argument in favor if putting the spinner after "Cited" badge in collectionTree. We can also display the "Cited" badge as faint-grey-ish until io is loaded, and change it to blue once we confirm that the group is indeed 100% cited.

Sounds good to me. It can look like this:

loading

- The spinner is put before the text so that it doesn't shift when the spinner disappears.
- In this state, the badge background is `fill-quarternary`, text is `fill-secondary`, and spinner is `fill-tertiary`.
- The blue that it changes to should be `accent-blue30`. Right now the blue is a bit too dark.

adomasven:
> > I'm not sure what you're referring to.
> 
> Oh, I was just referring to the `document` that we get in `execCommand` and that is later available in `Zotero.Integration.Sessions[key]._doc`. The `_doc` object there does seem to have an ID but it's not a constant property and changes each time a new session is created. So we'll need to rely on the stored `sessionID` in document data.

Oh, that's internal to the integration plugins and probably should be `_id`. Not all of them will have an id at all.

PR #5638, opened 2025-11-20
Fix bibliography modal race condition

### LINKED ISSUE #5619: Broken Bibliography modal if opened too early
The dialog is broken if opened before styles are cached. It eventually corrects itself, though on some occasions I had to manually resize the window because it remained stuck in a truncated state.

https://github.com/user-attachments/assets/f6141d1e-bd21-4e28-ae1f-9ef3ae376fb0

tnajdek:
Fix #5619

This is how it looks like now if opened very early:

https://github.com/user-attachments/assets/d2a5842a-6867-4cbf-83be-5350bee9c39f


dstillman:
Would be better to show a "Loading…" if styles aren't yet loaded and swap it out when the list is ready, but this is certainly an improvement.

PR #5640, opened 2025-11-25
Fix note editor hard refresh after saving note

windingwind:
Fix note editor instances' editing state lost after saving

https://github.com/user-attachments/assets/8caa1e7e-2a5c-4ee0-8f57-b99d6c30297c

This PR is based on #5528.

### DIFF
--- a/chrome/content/zotero/elements/noteEditor.js
+++ b/chrome/content/zotero/elements/noteEditor.js
@@ -182,13 +182,15 @@
 					let state = extraData && extraData[id] && extraData[id].state;
 					if (state) {
 						if (extraData[id].noteEditorID !== this._editorInstance.instanceID) {
-							this.initEditor(state, true);
+							let noteData = { state, html: this._item.note };
+							this._editorInstance.applyIncrementalUpdate(noteData, true);
 						}
 					}
 					else {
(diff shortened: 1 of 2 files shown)


dstillman:
Did you mean to update the note-editor submodule?

windingwind:
> Did you mean to update the note-editor submodule?

Yes, on both the note-editor and the client side. Basically this PR creates a method to update the ProseMirror doc with the editing state (selection etc.) preserved in the note-editor and calls that in the client.

The note-editor branch is https://github.com/windingwind/note-editor/commits/add-note-tab-type/ (last two commits are related to this PR).

I noticed this bug while working on the note-as-tab PR, and recently I figured out the stable working fix. That's why it's based on #5528 and its note-editor branch.

dstillman:
But the PR can't just update the submodule to a random commit in another repo. You'll need to open a PR in the note-editor repo so @mrtcode can review those changes, and then you should note here that this requires that PR. Once that's merged, this can be rebased to include the merged submodule commit.

windingwind:
OK, in that case, I'll also update the note-as-tab's submodule change as a PR to the submodule. Thanks.

The PR in note-editor: https://github.com/zotero/note-editor/pull/67

PR #5644, opened 2025-12-02
Set libraryID on annotations in itemTest

abaevbog:
Fix tests breakage post 84764570f4af1df81fb7de3df4b03b5c9f5a9139

### DIFF
--- a/test/tests/itemTest.js
+++ b/test/tests/itemTest.js
@@ -1600,6 +1600,7 @@ describe("Zotero.Item", function () {
 		describe("#saveTx()", function () {
 			it("should save a highlight annotation", async function () {
 				var annotation = new Zotero.Item('annotation');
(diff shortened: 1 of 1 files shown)


abaevbog:
@AbeJellinek, this fixes the breakage in `itemTest.js` like https://github.com/zotero/zotero/actions/runs/19679243949/job/56368748219 which now happens if you try to call [item.parentItem](https://github.com/zotero/zotero/blob/84764570f4af1df81fb7de3df4b03b5c9f5a9139/chrome/content/zotero/xpcom/data/item.js#L4079) without setting the libraryID on the annotation. It looks like the libraryID is set in all other instances besides tests. What do you think?

AbeJellinek:
Oh, thanks, I should've caught this. Looks good.

PR #5648, opened 2025-12-05
Use column dataKeys for toggling column visibility; make toggleHidden accept both numeric index and dataKey

RProppert:
Column toggling was broken because the resorting of selected columns made the old numeric indices invalid, leading to “column is undefined” errors. I updated the UI to send the column’s dataKey instead, and made toggleHidden() handle both keys and indices so everything stays stable hopefully. Submitted debug log with ID D406833145


dstillman:
Can you provide steps to reproduce the bug you're fixing?

Issue #5655, opened 2025-12-11
Double-clicking in collection tree no longer triggers `onActivate()` (e.g. collection renaming)


AbeJellinek:
Double-clicking does nothing, but Enter and right-click -> Rename still work.

abaevbog:
Regression after https://github.com/zotero/zotero/commit/015769a9f85990a896b288954524f097af307678

PR #5656, opened 2025-12-11
Fix no onActivate handling on double click of collectionTree rows

abaevbog:
After 015769a removed `pointer-events: none` from table cells, clicking on a `collectionTree` row would always trigger a `focusout` event. Then,`ZoteroPane.handleBlur` would call `collectionTree.setHighlightedRows`, which would always redraw the `collectionTree`. This complete tree redraw on every click made it impossible to register double-clicks because nodes were being removed and re-created.

With this change, `setHighlightedRows` won't have any effect if called with the same rows to highlight as before (including no rows). Without constant tree redraws, double-clicks should fire as expected.

Fixes: #5655

### DIFF
--- a/chrome/content/zotero/collectionTree.jsx
+++ b/chrome/content/zotero/collectionTree.jsx
@@ -1081,7 +1081,7 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 
 		// If the highlighted rows are the same as before, do nothing
 		let toHighlight = new Set(ids);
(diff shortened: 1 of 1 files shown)


abaevbog:
@AbeJellinek, does this tweak fix the issue on your end?

AbeJellinek:
It does!

abaevbog:
Ok, great!

PR #5658, opened 2025-12-12
Move Advanced Search and saved search editing to the main window

### LINKED ISSUE #5653: Merge advanced search pane into the main window
One benefit of that would be to support context-menu actions for found items. It would also play nicely with #5652

AbeJellinek:
"Hackathon project" indeed! But I think this is a really nice UX improvement.

Doing a one-off search:

image

Editing a saved search:

image

Needs one commit from https://github.com/zotero/zotero/pull/3435, which I've cherry-picked here.

Closes #5653


abaevbog:
This is awesome! Certainly a UI improvement. 

One question I have is: now that the advanced search window is removed, where would we place the advanced search across libraries, when that is ready? We briefly discussed the possibility of adding a new pseudo collectionTree row (e.g. "Global search"). Would that be the plan now?

A few minor things I noticed on the first try:
- keyboard nav is a bit off. Shift-tab from itemTree won't go into the advanced search tab. Shift-Tab from the collapse/expand button in "Advanced search" tab goes to tag selector, skipping toolbar buttons ("New item"). If the screen is small and itemPane appears at the bottom, Shift-tab from "Toggle item pane" button that appears right after "Advanced search" tab does nothing.
- some conditions do not update the value input type correctly. For example:
Screenshot 2025-12-12 at 9 45 38 AM

- When I open advanced search for the first time, `item type = annotation` seems to do nothing... After clicking around (e.g. changing mode from "AND" to "OR"), it works.
Screenshot 2025-12-12 at 9 47 48 AM

- Now that advanced search is more handy, I expect we will see way more `modify` actions when the search is active. Currently, `itemTree` just refreshes on `notify`, rerunning the search. In some cases (e.g. https://forums.zotero.org/discussion/100103/tagging-from-a-search-result-is-very-slow), it can take a while. One of the tweaks in https://github.com/zotero/zotero/pull/3435 was meant to address this by processing modified items only. I can revise it to make sure it still works, if that would be helpful here?

- If we are making modifications to the quick search field, can we consider removing the shadow DOM from it? I'm not sure it's needed anymore, and there have been weird behaviors reported with the dropdown (e.g. https://forums.zotero.org/discussion/127930/search-problem) that I could never reproduce. I have no reason specifically to blame shadow DOM, except that we never got any reports of it in other components.

AbeJellinek:
> We briefly discussed the possibility of adding a new pseudo collectionTree row (e.g. "Global search"). Would that be the plan now?

Personally I think that seems good. Maybe "All Libraries" (or "All Items"?) so it doesn't seem like it's *only* for search.

> some conditions do not update the value input type correctly

Fixed, thanks.

> When I open advanced search for the first time, item type = annotation seems to do nothing

Fixed!

> One of the tweaks in https://github.com/zotero/zotero/pull/3435 was meant to address this by processing modified items only. I can revise it to make sure it still works, if that would be helpful here?

Sure. Though I imagine that optimization would prevent a *new* match from being added to the view? Not a big issue right now, since the only case in which that could happen would really be syncing. Could be a bigger issue later, once we support multiple library tabs.

> If we are making modifications to the quick search field, can we consider removing the shadow DOM from it?

Sure, I think so. I'll see how hard that looks.

AbeJellinek:
A global pseudo-library should probably be a separate PR, though. There's a lot of Zotero code that thinks it knows whether an item is editable based on the collection tree row that it's being viewed from. I guess we could cheaply sidestep that issue by preventing edits from within the global thing, but that doesn't seem like the right approach.

abaevbog:
> A global pseudo-library should probably be a separate PR, though. There's a lot of Zotero code that thinks it knows whether an item is editable based on the collection tree row that it's being viewed from. I guess we could cheaply sidestep that issue by preventing edits from within the global thing, but that doesn't seem like the right approach.

Yeah, absolutely. It would also require quite a few tweaks in the itemTree to display items within groups nicely. I will make a separate issue for it.

abaevbog:
Just a few additional things I noticed:

- when a condition is removed, the window is unexpectedly shrunk by a little bit
- when a reader tab is open, Edit > Advanced Search menu currently has no effect. We probably want to just focus the main tab then?
- because Advanced search was a standalone window for such a long time, I am afraid some people may forget that they have a collection selected in the collectionTree, try to use advanced search to find items in another collection and report on the forums that nothing is there. Did we discuss adding a collection condition at the top of the advanced search when a collection is selected? It would clarify things a bit I think.
- Should we disable advanced search in trash? It looks like trashed items are always excluded, so in trash advanced search is always empty for me

abaevbog:
> Sure. Though I imagine that optimization would prevent a new match from being added to the view? Not a big issue right now, since the only case in which that could happen would really be syncing. Could be a bigger issue later, once we support multiple library tabs.

It is supposed to add the row of an item that did not match the search but after `modify` does! That PR is one of the first ones I put together so it's a bit unruly and has several different features for no particular reason, including proper annotations handling. So I'll rebase, and try to consolidate some commits. Those tweaks are not essential for this UI update, so it could be merged later too.

dstillman:
> There's a lot of Zotero code that thinks it knows whether an item is editable based on the collection tree row that it's being viewed from.

There sure is! (I'm dealing with this for multi-collection selection, which will also be a way to do an advanced search across multiple collections.)

windingwind:
Cool!

A suggestion after a quick test with this: maybe we should collapse the item pane (`ZoteroPane.itemPane.collapsed = true`) if in stacked mode and the vertical height is not enough to show at least several rows of the item list?

image

abaevbog:
Just to record another good use case for advanced search as a part of the main window: https://forums.zotero.org/discussion/128746/how-to-drag-and-add-literature-from-the-advanced-search-results-to-a-collection

dstillman:
Merged in d2f1c56250 and 00527332c4

dstillman:
@yexingsha: If you have tweaks for the UI, let us know. Along with the main changes here, I moved the advanced search action into a filter button within the quick search bar. Right now that's a little conceptually messy, since it's separate from quick search, but 1) we've had the Advanced Search menu option in the quick search mode menu for a while and 2) we could make it cleaner by automatically creating advanced-search conditions from the current quick search, which would be neat (and could even work if we extended the quick search syntax to allow prefixes/tokens, etc.).

yexingsha:
Here is what I have so far:

- Save Search should open a dialog for name input.
	- I was wondering if we could enable Save Search in collections as well, and there the dialog will also contain a checkbox that says "Include the current collection as a condition" or something like that.
- The height of the Advanced Search area seems to be calculated weirdly. At default height, the space below the buttons is larger than what it should be (8px). When I add any new conditions or Edit Saved Search, the space below the buttons become smaller. After I exit Edit Saved Search, the Advanced Search area no longer returns to its minimal height.
- The Advanced Search area should have a 1px bottom border using `color-panedivider`.
- Various alignment and spacing issues: because some input/combo boxes (but not all) have their own margins from Firefox css, the resulting padding and spacing becomes larger than intended. It would be nice if we could patch out the margins on the individual components, but if not, we will need to fine tune the margins by row.
- Move the filter button 4px to the right (`margin-inline-start: -26px`) so it has the same margin on all sides.

AbeJellinek:
On macOS, the condition value field seems to extend too far down:

image

PR #5663, opened 2025-12-16
Remove `1px solid transparent` border on links

dstillman:
We seem to have had this for years (since at least 2011), but I'm not sure why, since it means that a link at the end of a sentence pushes the period away from the link.

Screenshot 2025-12-16 at 3 32 00 PM

Screenshot 2025-12-16 at 3 32 10 PM


AbeJellinek:
Oh, I had never noticed this. Looks good!

tnajdek:
I subconsciously noticed that in the about dialog but never thought to investigate. Looks good to me!

Issue #5661, opened 2025-12-15
[MacOS] citation dialog doesn't load if no main Zotero window is available


adomasven:
Haven't tested this yet, but it's suggested in https://forums.zotero.org/discussion/comment/504140/#Comment_504140

We occasionally get reports for the progress bar showing, but not leading to citation dialog fully loading from macOS users and I've been wondering what's up with that.

@abaevbog want to take a look at this?

abaevbog:
Right, I can reproduce it if the main window is closed but Zotero is still running and appearing in the dock. It happens both with QuickFormat and the new citation dialog in beta. Thanks for flagging this to me!

PR #5664, opened 2025-12-16
Citation dialog: fix hang if the main window is closed

abaevbog:
Fixes: #5661

### DIFF
--- a/chrome/content/zotero/integration/citationDialog/searchHandler.mjs
+++ b/chrome/content/zotero/integration/citationDialog/searchHandler.mjs
@@ -273,8 +273,18 @@ export class CitationDialogSearchHandler {
 
 	async _getReaderOpenItems() {
 		if (this.isCitingNotes) return [];
+		let tabs = [];
 		let win = Zotero.getMainWindow();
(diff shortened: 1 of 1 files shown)


abaevbog:
@adomasven, I believe it's just this to address the citation dialog hanging when the window is closed.

Do you think we'll need a fix for the older quick format as well, with Zotero 8 being close?

adomasven:
Does this mean we aren't fetching open tab items with the window closed? Can we not get them somewhere else?

> Do you think we'll need a fix for the older quick format as well, with Zotero 8 being close?

No I think it's fine to do it for the new citation dialog only.

abaevbog:
At first I thought there was not much we could do, since `Zotero_Tabs` is just a part of the main window, and all the tabs are being closed when Zotero Pane is destroyed.

But it turns out that we already [save the session state](https://github.com/zotero/zotero/blob/86823e45efe673847807f348e968db2d8951bd24/chrome/content/zotero/zoteroPane.js#L766) when the main window closes, so we can just fallback to tabs recorded in `Zotero.Session.state`, if the main window is missing!

PR #5672, opened 2025-12-22
RTF Scan rework

### LINKED ISSUE #5662: RTF scan fails for references with Unicode characters
To address number 2 listed in the post below we should revisit the regular expressions in the code and, at the least, replace `A-Z`/`a-z` capture groups with modern Unicode-compatible equivalents.

https://forums.zotero.org/discussion/128551/rtf-scan-problems-diacritics-and-missing-authors

tnajdek:
Current RTF Scan functionality is severely limited and will only correctly extract pure ASCII citations with no formatting from an RTF document. Citations with non-ASCII characters will either be ignored or presented incorrectly; in the latter case, searching for a matching document will also use incorrect search criteria and result in no matches.

This PR re-implements the RTF Scan feature with support for the various ways RTF can represent non-ASCII characters. It also improves citation discovery and adds a test suite. Close #5662.

Here is a screenshot from running RTF Scan for a [sample RTF document](https://github.com/user-attachments/files/24532696/demo.rtf) in current Zotero:

Screenshot 2026-01-09 at 18 37 15

And with this PR:

Screenshot 2026-01-09 at 18 38 09

Summary of changes:

- Separate RTF processing logic from the UI-related code
- Add test coverage
- Add support for citations with Unicode characters in authors and in the title (#5662)
- Add support for citations with non-ASCII quotation marks (Unicode left/right and high/low)
- Add support for citations with a title but no author (https://forums.zotero.org/discussion/128551/rtf-scan-problems-diacritics-and-missing-authors)
- Add support for documents encoded with multiple codepages
- Add support for legacy multi-byte encodings
- Add support for font lookup (which may affect encoding!)
- Ignore control sequences and control symbols when looking for citations

### DIFF
--- a/chrome/content/zotero/modules/rtf.mjs
+++ b/chrome/content/zotero/modules/rtf.mjs
@@ -1,0 +1,647 @@
+export const UNMAPPED = 0;
+export const AMBIGUOUS = 1;
+export const MAPPED = 2;
+
+const BIBLIOGRAPHY_PLACEHOLDER = "\\{Bibliography\\}";
+
+// https://en.wikipedia.org/wiki/Rich_Text_Format#Character_encoding
+// https://encoding.spec.whatwg.org/
+const CHARSET_CODEPAGE_LOOKUP = new Map([
+	[0, 'windows-1252'],			// 	Latin alphabet, Western Europe / Americas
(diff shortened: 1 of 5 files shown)


dstillman:
Awesome!

PR #5677, opened 2025-12-24
Fix tab sidebar layout

### LINKED ISSUE #5551: Reader sidebar does not appear after macOS Tahoe update
https://github.com/user-attachments/assets/0328fb7a-04b2-4124-95be-be5fa1bd4c0a

windingwind:
Fix #5551

### DIFF
--- a/chrome/content/zotero/tabs.js
+++ b/chrome/content/zotero/tabs.js
@@ -479,7 +479,7 @@ var Zotero_Tabs = new function () {
 	};
 
 	this._loadSidebarState = () => {
-		let sidebarState = Zotero.Prefs.get('sidebarState') || "{}";
+		let sidebarState = Zotero.Prefs.get('sidebarState') || '{}';
 		try {
 			sidebarState = JSON.parse(sidebarState);
 			for (let tabType in sidebarState) {


windingwind:
Changes: Moved all sidebar layout logic outside of `contextPane.js` into `tabs.js`; Only use `sidebar` everywhere for the left side space of tabs to avoid confusion; Added a min-width limit to sidebar based on the settings in the reader so that they never get 0 width.

mrtcode:
Much better now! Good that you removed `sidePaneState` pref, since we already have `zotero-persist`, which isn't perfect, but introducing another pref for this should be deliberately thought out.

One small thing: `Note Tab Font Size > Reset` doesn't work.

Also, I think we should be more consistent with single vs double quote usage. Right now, even within the same commit, it seems to randomly switch between the two, even when the file previously used consistent quotes.

windingwind:
Well we are using another sidebarState pref (a simple rename). Because for different tab types the same placeholder elem is reused and ideally they should not share the same sidebar state (open/collapsed), thus the persist mechanism won’t work.

i will check the font reset.

re quote: I personally always prefer double, while we usually have single everywhere and also sometimes double. I think we had similar discussions years ago and Dan decided we didn’t have a strict policy on that, if I remember correctly

windingwind:
> One small thing: `Note Tab Font Size > Reset` doesn't work.

fixed in #5678, thanks

PR #5681, opened 2025-12-27
fix zotero://select on cold start

humblehacker:
I recently discovered Zotero and it's almost perfect for my needs. But one thing I really need is the ability to drop links in markdown files that when clicked open the item in whatever tool I'm using to track my sources (currently DevonThink, where this works perfectly). However, Zotero doesn't handle the link properly on a cold start. The app opens, but nothing is selected. I did some debugging and discovered a few places where certain references weren't yet available at the time the link is handled - namely the main window and the item tree. So I've added some readiness promises that are resolved in places that I think are sufficient. It works pretty flawlessly for me now on macOS - I haven't tried any other platforms.

Happy to discuss and work on alternate solutions if this one isn't suitable for any reason.

### DIFF
--- a/chrome/content/zotero/ZoteroProtocolHandler.mjs
+++ b/chrome/content/zotero/ZoteroProtocolHandler.mjs
@@ -637,6 +637,14 @@ export function ZoteroProtocolHandler() {
 			}
 			
 			win.Zotero_Tabs.select('zotero-pane');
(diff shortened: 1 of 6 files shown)


dstillman:
(We can look at this, but we'd first need the hundreds of whitespace changes to be reverted.)

humblehacker:
Thanks for taking a look. Whitespace reverted.

Issue #5688, opened 2026-01-02
Enter on focused tab no longer focuses tab's content


abaevbog:
1. Tab onto a tab (reader or note)
2. Press Enter

It focuses the first button in the toolbar but it is supposed to (as in the stable build) focus the reader content so you can scroll with arrows. For note tabs, it should probably focus the editor, so you can type after that.

Issue #5684, opened 2026-01-02
Cannot scroll reader content with arrows after re-selecting tab


abaevbog:
https://forums.zotero.org/discussion/comment/504747/#Comment_504747

1. Open a reader tab
2. Go back to the library tab
3. Click on the loaded tab
4. Notice that pressing arrow up/down does not scroll content.
5. Arrows will scroll content after pressing Escape

abaevbog:
Similarly, if you select an already-loaded note tab, editor is not automatically re-focused, so you can't type right away.

Issue #5683, opened 2026-01-02
Focus not restored after returning to library tab


abaevbog:
https://forums.zotero.org/discussion/128761/zotero-8-beta-focus-does-not-return-to-center-pane-after-closing-tab

1. Navigate to an item in itemTree
2. Press Enter to open it in a new tab
3. Press Cmd+W to close the newly opened tab
4. Note that focus in library tab is no longer on the itemTree but on the "New Collection" button

abaevbog:
Yeah, this is a post notes-in-tabs regression. From what I see, when a tab is selected, context pane tries to focus its content:

https://github.com/zotero/zotero/blob/58baa44ad446fe6dea2ca30a6c104aa28495c987/chrome/content/zotero/elements/contextPane.js#L263-L282

It ends up emitting `tab-focus` even that reader or note tabs know how to handle. For the library tab, though, we should be using the logic of refocusing `ZoteroPane.lastFocusedElement`, with itemTree as the default fallback. That currently lives in `Zotero_Tabs.tabhooks.refocus.library` - but it seems to be only used as `onRefocus` prop in the tabbar.

https://github.com/zotero/zotero/blob/58baa44ad446fe6dea2ca30a6c104aa28495c987/chrome/content/zotero/tabs.js#L148-L165

windingwind:
Thanks for the investigation.

I think the described behavior of focusing the new collection button is from the `Zotero_Tabs.tabHooks.focus`, which is also the logic inherited from the previous code (just unified into one place). The focus code in the context pane is another old deeply coupled code that I tried not to touch during refactoring. Though it actually calls the `setFocus`, which is also what gets called by `tab-focus` event. This is likely harmless.

Since we are seeing various issues with the tab focus and we used to have multiple places for tab focus handling and they seem to work by sometimes setting timeout to overwrite each other, I think it's time to further integrate them into one place.

Not sure if I summarized it correctly:

When you select a tab:

- It's a newly created tab, or the tab is selected with `keepTabFocused !== true`, trigger the behavior in `tabHooks.focus`, e.g. the library's first button
- If it's not library tab and context pane note editor is active, keep focus
- Otherwise, use tabContent's `tab-focus` handle to let the tab decide where to focus
- A special case: when you navigate tabs with the arrow keys, keep focus in tabs.

Let me know if I missed any part.

PR #5689, opened 2026-01-04
Fix tab focus control

windingwind:
fix: #5683
fix: #5684
fix: #5688

Rename tab hook focus -> focusFirst
Replace tab-focus custom event with tab hook refocus
Move context pane _handleTabReady to Zotero_Tabs
Focus note editor after opening tab
Use setTimeout to ensure the focus does not stay in tab element

### DIFF
--- a/chrome/content/zotero/elements/contextPane.js
+++ b/chrome/content/zotero/elements/contextPane.js
@@ -218,7 +218,6 @@
 					// updated by another select tab event.
 					&& (action === 'select'
 						|| (action === 'load' && Zotero_Tabs.selectedID == tabID))) {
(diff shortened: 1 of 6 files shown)


abaevbog:
I think this is looking and working great! And not having an extra custom `tab-focus` event makes the logic a bit easier to follow as well. 

I am wondering if maintaining `Zotero_Tabs._focusOptions` is still needed, now that selecting an unloaded tab doesn't create and select a completely new loaded tab (which is what `_focusOptions` was meant to address)... Maybe it could be removed at some point to simplify `Zotero_Tabs.select` - but it's not something necessary to try now I think.

PR #5697, opened 2026-01-12
Fix field and abstract changes not refreshing in the library pane if performed from the context pane

### LINKED ISSUE #5694: Abstract not refreshed in library tab when updated in context pane
https://forums.zotero.org/discussion/128812/abstracts-not-saved-when-attachment-open

tnajdek:
`InfoBox` registers an observer for when an item is updated, but if the tab is not visible, this won’t trigger a render. And since it’s not listening for a tab change, it won’t ever refresh the rendered fields. `AbstractBox` has the same problem. There is a separate listener for a tab change already in the `ItemDetails`, but that, as far as I can tell, is unaware of the item change.

In this PR, I’ve introduced an observer for tab changes in the `InfoBox` and `AbstractBox` so that these components can “catch up” after any renders missed due to item changes on other tabs.

Fix #5694

### DIFF
--- a/chrome/content/zotero/elements/abstractBox.js
+++ b/chrome/content/zotero/elements/abstractBox.js
@@ -67,7 +67,7 @@
 		}
 
 		init() {
-			this._notifierID = Zotero.Notifier.registerObserver(this, ['item'], 'abstractBox');
+			this._notifierID = Zotero.Notifier.registerObserver(this, ['item', 'tab'], 'abstractBox');
 
 			this.initCollapsibleSection();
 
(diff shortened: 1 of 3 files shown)


chrome/content/zotero/elements/abstractBox.js
```diff
@@ -86,6 +86,9 @@
 			if (action == 'modify' && this.item && ids.includes(this.item.id)) {
 				this._forceRenderAll();
 			}
+			if (action === 'select' && type === 'tab') {
+				this._handleTabSelect(ids);
```
dstillman: Maybe clearer to pass `ids[0]` and take a single id in `handleTabSelect`? I assume it's an array because other tab events (e.g., closing?) can be multiple, but a select would inherently be a single tab.

tnajdek: Done! I've adapted that code from [ItemDetails](https://github.com/zotero/zotero/blob/e899012e5e2696756ee3f3f7c3a38cf7b9ea0b0d/chrome/content/zotero/elements/itemDetails.js#L787). Perhaps we should tweak it there as well?

windingwind:
looks good to me in general. thanks!

Issue #5704, opened 2026-01-17
Note tab reopen logic


windingwind:
https://forums.zotero.org/discussion/128954/

https://forums.zotero.org/discussion/128917/

PR #5706, opened 2026-01-20
Fix note tab reopen logic

windingwind:
fix: #5704 
Distinguish note editor CE and instance in var name 
Add note editor CE _initPromise 
Await CE init in various places 
Add tests for note tab


dstillman:
(Merging to get this in a beta, but @mrtcode can still take a look.)

mrtcode:
Looks good.

PR #5710, opened 2026-01-20
Add note editor content view mode

windingwind:
Restore wide mode (full-width) for non-tab note editors

Use https://github.com/zotero/note-editor/pull/71


mrtcode:
Looks good.

PR #5716, opened 2026-01-22
Fix `contextPaneOpen` race condition causing duplicate button

### LINKED ISSUE #5715: Reader shows extra context pane toggle when opened with context pane visible
Randomly, like half the time for me.

https://forums.zotero.org/discussion/129107/double-toggle-context-pane-icon-in-pdf-reader

AbeJellinek:
I haven't done a bisect but I believe this was a note-in-tab regression; all the code that set `Zotero.Reader._contextPaneOpen` was removed, but we were still passing it to new reader instances. @windingwind, let me know if this looks OK or if there's a better approach we could use.

Fixes #5715


windingwind:
Thanks for the fix. Looks good!

PR #5719, opened 2026-01-23
Fix color of OK button in MS Word DocPrefs

northword:
https://forums.zotero.org/discussion/129173/zotero-8-accent-color-of-the-ok-button-in-the-document-preferences-pane-of-ms-word-addins#latest


dstillman:
Probably a regression from https://github.com/zotero/zotero/pull/5638

dstillman:
Sorry, merged this by mistake. Fixing to `disabled="true"`

PR #5726, opened 2026-01-24
Update integration.ftl

pan4ratte:
Revised and updated the Russian translation


dstillman:
https://github.com/zotero/zotero/pull/5728#issuecomment-3794725382

PR #5735, opened 2026-01-26
Replace `×` close links with buttons in banners

### LINKED ISSUE #5723: Wrong cursor for close buttons in banners
Image

Should be treated like a button, not a link

tnajdek:
Looks like this:
Screenshot 2026-01-26 at 11 47 42

Behaves like a button on hover (not present on the screenshot but there is no cursor change):
Screenshot 2026-01-26 at 11 47 32

I’ve also added the `aria-label` attribute, though I’m not sure if banners are accessible in screen readers at all, so it might not be relevant.

Fix #5723

### DIFF
--- a/chrome/content/zotero/zoteroPane.xhtml
+++ b/chrome/content/zotero/zoteroPane.xhtml
@@ -1135,7 +1135,9 @@
 									<html:div class="spacer"/>
 									<html:a id="sync-reminder-disable" class="link"/>
 									<html:a id="sync-reminder-remind" class="link"/>
-									<label is="text-link" id="sync-reminder-close" class="close-link">×</label>
+									<html:div role="button" id="sync-reminder-close" class="close-btn" data-l10n-id="banner-close-button">
+										<html:span class="icon" />
+									</html:div>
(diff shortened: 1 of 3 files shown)


windingwind:
Kind of offtopic but tbh do we consider having a banner system that shows one banner at a time and has something like

[Banner message... `previous` `next` `close`]

So that we don't have these stacked rainbow banners? It's really scary to me

dstillman:
No one sees these all at once — these are just turned on for demonstration purposes.

I don't think showing one at a time is desirable. We use these for fairly critical messages that we want users to see. We know from screenshots that people already sometimes leave these untouched, which is why we added a 7-day timeout for the most recent attachment-renaming one. But if there's a message, we want people to see it and deal with it. We could make sure they all have a Remind Me Later option, though.

A system to unify these would be nice, but it's also not that big a deal. We don't create these that often, and copying/pasting a block in zoteroPane.xhtml isn't the end of the world.

dstillman:
`--color-background50` seems a little light here. Could probably be a bit closer?

PR #5736, opened 2026-01-26
Fix missing ellipsis for styles in the style selector

tnajdek:
It might be a regression, but the style selector doesn’t use ellipsis when long style names are truncated. This PR fixes that and also introduces a tooltip attribute so that it’s possible to check the entire style name by hovering over a style label.

Before:

Screenshot 2026-01-26 at 14 24 47

After:

Screenshot 2026-01-26 at 14 20 19

### DIFF
--- a/chrome/content/zotero/elements/styleConfigurator.js
+++ b/chrome/content/zotero/elements/styleConfigurator.js
@@ -86,8 +86,11 @@
 					.replace(/^American Sociological Association/, "American Sociological Association (ASA)");
 				
 				let richlistitem = document.createXULElement('richlistitem');
+				let labelEL = document.createXULElement("label");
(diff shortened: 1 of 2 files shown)


dstillman:
Are you seeing this in the settings too? I get both an ellipsis and a tooltip there before this PR:

Screenshot 2026-01-26 at 10 36 53 AM

tnajdek:
The one in the settings works correctly, but it’s a different component. In the settings, we use our Virtualized Table so that we can include additional columns. In the bibliography dialog (and RTFScan), we use XUL’s `richlistbox`, rendered by our `StyleSelector` custom element.

dstillman:
Ah, right, sorry.

Issue #5738, opened 2026-01-26
Restore first-run guidance panel for citation bubbles


dstillman:
And reset `firstRunGuidanceShown.quickFormat` in prefs.js. Since it's a new dialog, I think we can just show it again for everyone, not just people in Library mode.

PR #5740, opened 2026-01-26
Citation dialog: restore first run guidance panel

abaevbog:
- restore guidance panel with instructions appearing on the first run
- increase preference version to 20, clear `firstRunGuidanceShown.quickFormat` pref, initialize `firstRunGuidanceShown.citation-dialog` pref to false
- set slightly higher arrow padding on win/linux for the guidance panel. If the arrow is too close to the edge, it appears partly disconnected from the panel.

Fixes: #5738

Mac:
Screenshot 2026-01-26 at 10 21 37 AM

Linux:
Screenshot 2026-01-26 at 10 16 55 AM

Windows:
Screenshot 2026-01-26 at 10 11 30 AM

For reference, this is what the windows panel's arrow would look like without increasing the padding a little bit:

Screenshot 2026-01-26 at 10 02 45 AM

### DIFF
--- a/chrome/content/zotero/xpcom/prefs.js
+++ b/chrome/content/zotero/xpcom/prefs.js
@@ -213,10 +213,8 @@ Zotero.Prefs = new function () {
 						break;
 					
 					case 20:
-						// QuickFormat dialog no longer exists
+						// QuickFormat dialog no longer exists -- replaced with `.citationDialog`
 						this.clear("firstRunGuidanceShown.quickFormat");
-						// Replaced by citation dialog
-						this.set("firstRunGuidanceShown.citationDialog", false);
 						break;
 				}
 			}


dstillman:
Oh, and I guess we need to do https://github.com/zotero/zotero/pull/5529 too so this is accurate.

dstillman:
Is that last line still correct with https://github.com/zotero/zotero/pull/5529? I forget where we landed there.

abaevbog:
In that PR we setup the logic of having the locator go to the last added bubble. Suppose you add bubble A, and then you type and add bubble B that is sorted to go before A (so BA). Then, if you type "Line 10-15" and press Enter, it will go to the last-added bubble B, even though you would type after bubble A.

We also brought back purely numeric locators that are treated as a page when typed right after adding a bubble. So if a bubble is added, and then you type "10-15", it will be added as "page 10-15" locator. The last bubble record is cleared once you move focus anywhere. If you type "US history 2024", the last number by itself is still treated as a search query and not a locator to not conflict with years.

With that, the last line "You can also add a page number directly by including it with your search terms or typing it after the bubble and pressing { return-or-enter }" is somewhat true but maybe should be clarified to say not "page number" but a "locator", since now we handle all of them?

abaevbog:
`You can also add a locator directly by including it with your search terms (e.g. "history p. 10-15") or typing it after the bubble and pressing { return-or-enter }`?

abaevbog:
Updated panel string:

Screenshot 2026-01-26 at 12 52 22 PM

abaevbog:
While looking through this, I noticed a glitch that would try to parse the last numeric entry in a string as a year (leftover logic from quickFormat), and strip some of the punctuation, which can loose the last part of a typed locator in search query. It is addressed in https://github.com/zotero/zotero/pull/5529 but it's a small part of that PR.

I can rebase https://github.com/zotero/zotero/pull/5529 if we want to get it in together with guidance panel update, or I can pull out just that small fix into a separate pr. What would be best?

dstillman:
Yeah, I was going to merge #5529 with this for 8.0.1, since people are noticing.

abaevbog:
Ok, great. I'll rebase and squash into one commit that describes the final behavior we settled on there

dstillman:
Simplified it a bit more:

`You can also add a page number or other locator by including it with your search terms (e.g., “history p10”) or by typing it after the bubble and pressing { return-or-enter }.`

abaevbog:
That sounds good! `guidance-panel` ties both pref and the strings fetched from the `.ftl` file to the `about=` attribute. So if we want it to be `citationDialog`, we should also rename the string name to `first-run-guidance-citationDialog` as well.

dstillman:
Ah, thanks. (And yes, looks like we have, e.g., `first-run-guidance-authorMenu` as well.)

dstillman:
Are the prefixes hard-coded or based on the localized locator names? E.g., should the "p10" be localized?

dstillman:
I guess they're based on the localized prefixes from the CSL locales. So then we should pass those in, not have them be localized via Transifex.

dstillman:
I've updated the string in Transifex to use `{ $locator }` instead of `p10`. Let's update the .ftl and the calling code accordingly.

abaevbog:
Yeah, I see what you mean about the CSL locator. I added #5746 to essentially insert the short CSL locator returned from `Zotero.Cite.getLocatorString("page", "short")` as a `{$locator}` variable instead of always using the hard-coded "p".

Issue #5745, opened 2026-01-27
Citation dialog: Pass localized locator to first-run guidance string


dstillman:
https://github.com/zotero/zotero/pull/5740#issuecomment-3803275890

PR #5746, opened 2026-01-27
Citation dialog: localized CSL page locator in guidance panel text

abaevbog:
Use localized short CSL page locator (from `Zotero.Cite.getLocatorString("page", "short")`) in guidance panel text example ("p.10").

Fixes: #5745

---
English - note "p.10" is used. We could remove the "." - but I am not sure if that would still look right for all locales the same as for English.
english

In another locale (Russian), note the locator is "c.10" ("c" for "страница")
russian

Also, fixed dialog mode switch buttons not properly sizing in locales where list/library string is longer.

Before:
Screenshot 2026-01-26 at 10 58 33 PM

After:

Screenshot 2026-01-26 at 10 58 05 PM

### DIFF
--- a/chrome/content/zotero/integration/citationDialog.js
+++ b/chrome/content/zotero/integration/citationDialog.js
@@ -127,8 +127,14 @@ async function onLoad() {
 		}
 	});
 
+	//
 	// Show guidance panel on the first run. Noop on subsequent runs.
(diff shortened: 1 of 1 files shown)


dstillman:
I dunno — stripping the period seems fine to me? I went with that for now.

Issue #5751, opened 2026-01-27
Quick search doesn't match annotation tags on standalone PDFs


dstillman:
https://forums.zotero.org/discussion/comment/506494/#Comment_506494

PR #5756, opened 2026-01-27
Fix quicksearch not finding annotations of standalone attachments

abaevbog:
Fixes: #5751


abaevbog:
https://github.com/user-attachments/assets/dd9d87bf-811c-4b09-9f61-d2f28d2e0f02

PR #5774, opened 2026-02-03
Add UI for per-group file renaming settings

### LINKED ISSUE #5759: Add UI for per-group renaming settings
Group admins should be able to edit. Read-only for others.

Assigning this to myself as well since I'm not sure we know locally whether someone is a group admin, which will be a prerequisite here, but the rest could be worked on with the assumption that we'll have that eventually.

Follow-up to https://github.com/zotero/zotero/issues/5739

tnajdek:
https://github.com/user-attachments/assets/750cf5ac-0e57-451f-85b1-8e545cdacd39

Introduced a new context menu option to allow configuring file renaming per library. To enable this, a new custom element for editing file renaming settings was added, which avoids code duplication. Additionally, in a few places, methods were adjusted to better align with the idea of renaming happening per library. This currently displays the new option to everyone; we still need to tweak this so it is presented only to admins.

Close #5759

### DIFF
--- a/chrome/content/zotero/fileRenamingDialog.js
+++ b/chrome/content/zotero/fileRenamingDialog.js
@@ -25,6 +25,8 @@
 
 var FileRenamingDialog = { // eslint-disable-line no-unused-vars
 	prompted: false,
+	_settingsChanged: false,
+	_currentLibraryID: null,
 
 	init: function () {
 		const { DEFAULT_ATTACHMENT_RENAME_TEMPLATE, DEFAULT_AUTO_RENAME_FILE_TYPES } = ChromeUtils.importESModule("chrome://zotero/content/renameFiles.mjs");
@@ -56,6 +58,14 @@ var FileRenamingDialog = { // eslint-disable-line no-unused-vars
 		this._renameFilesPrefObserver = Zotero.Prefs.registerObserver('autoRenameFiles.done', this._handleDonePrefChange);
 
 		this.loadSettingsForLibrary(initialLibraryID);
+		this._currentLibraryID = initialLibraryID;
+
+		window.addEventListener('unload', () => {
+			if (this._shouldPromptRename(this._currentLibraryID)) {
+				this._promptRename(this._currentLibraryID);
+			}
+			Zotero.Prefs.unregisterObserver(this._renameFilesPrefObserver);
+		});
 	},
 
 	get libraryID() {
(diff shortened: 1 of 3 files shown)


dstillman:
Doing this as a context-menu option seems a little weird to me. Most people don't need to change this, so the context menu feels a little unnecessarily prominent, and it's confusing to keep the main settings when those only apply to your personal library. Can we just add a library drop-down in the settings?

dstillman:
Though I wonder if we need to move all the renaming prefs into that new dialog, even for My Library, if we do that, and put the drop-down at the top. Otherwise we'd need a drop-down both in the main settings section and in the configure window, which is awkward.

I just think having a top-level section that only applies to My Library is confusing.

tnajdek:
I agree this would be nicer with a library dropdown on top. So:
1. In preferences, under "File Renaming," there would only be a "Configure File Renaming" button?
2. This opens a new dialog, similar to the one opened from the group context menu in this PR, but with a library selector on top.
3. The same dialog is used when configuring file renaming from the context menu.
4. The context menu option is removed.

This should work, except the button in 1. might look a bit awkward but I can't think of a better place to put it.

I'll go ahead and tweak the code.

dstillman:
Yeah, I think that's all correct.

tnajdek:
I'm still considering a few UX-related issues:

- The prompt that appears immediately after toggling "Automatically rename files" in the new dialog now feels invasive -- we just opened the dialog and probably want to do some more configuration before triggering the rename.
- We used to prompt the user to rename existing files when closing the preference pane with a modified but unapplied template. We could still do this in the dialog, but it's not clear when to trigger it. Probably two checks: when the user switches libraries, or when they close the dialog with their library selected and a template unapplied?
- I think we should re-implement the prompt. We could extend the logic to track whether any change was made for a given library and prompt for group libraries as well.
- "Rename files" is disabled for the user's library after the new template is applied (tracked via `autoRenameFiles.done`). For group libraries we don't have an equivalent, so it stays enabled — but I think that's correct?

tnajdek:
I've adjusted the workflow:

- The prompt appears if changes were made to the file renaming settings when either closing the dialog or switching to another library
- When switching to another library, "Rename Files..." cancels the switch and "Keep Existing Filenames" continues with it
- For My Library, we use the existing `autoRenameFiles.done` pref
- For group libraries, we keep track of baseline settings and use those to decide whether to show the prompt
- Also, for group libraries, as long as file renaming is enabled, we never hide the "Rename Files..." button, as we can't be sure whether filenames are in sync
- This is UI changes only. For a version of this that sets read-only flag for non-admins see #5862

https://github.com/user-attachments/assets/5e4500a2-35dd-4113-9a95-d7e12711d009

dstillman:
Superseded by #5862

Issue #5775, opened 2026-02-03
Ignore `strict_max_version` for plugins in beta/dev/source builds


dstillman:
(Are we actually enforcing it now? `extensions.strictCompatibility` is set to `false`, but maybe that only applies if `strict_max_version` isn't set.)

This will presumably need to be in `fetch_xulrunner` and won't able to use our flags, but this is how we check the build type:

https://github.com/zotero/zotero/blob/2f064a5604632425b4f4778f9388e723a98f2a42/chrome/content/zotero/xpcom/zotero.js#L233-L236

dstillman:
(I won't change the version to `9.0` until we address this, but it would be good to do so soon for clarity of the beta version numbers.)

windingwind:
After in depth investigation:

> Are we actually enforcing it now?

no. we only force plugins setting a max_strict_version otherwise it directly fails the check.

https://github.com/windingwind/zotero/blob/Skip-plugin-strictCompatibility-check-on-non-stable-releases/app/scripts/fetch_xulrunner#L232-L234

> extensions.strictCompatibility is set to false, but maybe that only applies if strict_max_version isn't set.

not sure where the pref is set, but seems like we are not setting it. do we want to?

it does not apply because even if this pref is set to false, the web extension addon instance has a hard coded strictCompatibility = true, which always triggers the full version check:

https://searchfox.org/firefox-esr140/source/toolkit/mozapps/extensions/internal/XPIDatabase.sys.mjs#592-597

so the most straight forward solution is replacing the hard coded flag on addon instance to false.

PR #5781, opened 2026-02-04
Skip plugin strict_max_version check on non-stable releases

windingwind:
resolve: #5775

When the strictCompatibility of addon instance and the `extensions.strictCompatibility` are both set to false, the plugin version check only compares the min version.

See https://searchfox.org/firefox-esr140/source/toolkit/mozapps/extensions/internal/XPIDatabase.sys.mjs#592-597

### DIFF
--- a/app/scripts/fetch_xulrunner
+++ b/app/scripts/fetch_xulrunner
@@ -244,17 +244,11 @@ function modify_omni {
 	
 	# When installing addon, use app version instead of toolkit version for targetApplication
 	replace_line "id: TOOLKIT_ID," "id: '$APP_ID'," modules/addons/XPIInstall.sys.mjs
-
+	
 	# Set strictCompatibility to false on beta/dev/source builds to skip strict_max_version check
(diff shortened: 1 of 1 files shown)


dstillman:
The version check here won't work — there's no channel when fetching xulrunner. (`version` will only ever be [version].SOURCE.) The actual modified Fx code has to use `Services.appinfo.version` to check the runtime version.

windingwind:
we are not setting the extensions.strictCompatibility, i cannot find it anywhere in our code, unless it's set by firefox but doesn't seem to be so.

do we want to set that flag for non-stable clients, or i would suggest doing it same way in the fetch_xulrunner, so that people don't get polluted profiles when switching between different client channels.

dstillman:
It's set by default in Firefox:

https://searchfox.org/firefox-main/rev/336fbf710daa1af13ff2bb550dd0ea1d4e82b305/browser/app/profile/firefox.js#32-33

Issue #5780, opened 2026-02-04
Citation dialog: Scroll to first cited item in Library mode


dstillman:
https://forums.zotero.org/discussion/comment/507058/#Comment_507058

PR #5782, opened 2026-02-05
Citation dialog: scroll to 1st cited row in library mode on open

abaevbog:
If the dialog opens in library mode, after loading, scroll to the first cited item in this citation in current collection.

Fixes: #5780

Also, fix sorting by `+` column not working due to typo.

https://github.com/user-attachments/assets/83038349-8943-479c-aa8c-c9d4cb4ab119

---

I am not sure I remember what was the scrolling behavior of the classic citation dialog. I thought it makes sense to do it on the initial open but we shouldn't scroll anywhere after, since it may be disorienting as you just try to browse collections. If you want to always see all cited items, you can sort by the column with the `+` sign.

### DIFF
--- a/chrome/content/zotero/integration/citationDialog.js
+++ b/chrome/content/zotero/integration/citationDialog.js
@@ -788,7 +788,7 @@ class LibraryLayout extends Layout {
 		// Scroll to the first cited item
 		let firstCitedRow = this.itemsView._rows.findIndex(row => CitationDataManager.itemAddedCache.has(row.ref.id));
 		if (firstCitedRow == -1) return;
-		this.itemsView.ensureRowIsVisible(firstCitedRow);
+		this.itemsView.ensureRowsAreVisible([firstCitedRow]);
 	}
 
 	// after an item is added, bubble-input's height may increase and push the itemTree down


dstillman:
Seems a little weird that if you're in List and you quickly switch to Library, it doesn't scroll? Since Library is like classic, and classic scrolled on open, the equivalent here would be to make sure we scroll even if someone was in List first and then switched to Library.

dstillman:
(Also, not really for this PR, but does `ensureRowIsVisible()` not try to scroll rows a bit more into view instead of leaving them at the bottom of the list? We've definitely had code before that tried to put them at least a few rows up. Centering would be even better…)

abaevbog:
> Seems a little weird that if you're in List and you quickly switch to Library, it doesn't scroll?

Yeah, we can always try to scroll the first time the library layout is open, regardless if it was the initial layout or if the user switched from list layout to library.


https://github.com/user-attachments/assets/1704a817-4e36-4443-a558-d78db36ec637


> Also, not really for this PR, but does ensureRowIsVisible() not try to scroll rows a bit more into view instead of leaving them at the bottom of the list?

Aha, yes! It turns out we have `itemsView.ensureRowIsVisible` and also `itemsView.ensureRowsAreVisible`, and we want to use the latter here. At first, I also thought that the scrolling was supposed to happen but assumed I was just mistaken. Turns out, it was just `ensureRowsAreVisible` vs `ensureRowIsVisible`.

Even though `ensureRowsAreVisible` is meant to handle multiple rows, if multiple rows are cited, it ends up scrolling past the first cited row for me in the citation dialog. I assume that's because the size of the window is way smaller than in the main window. For this reason, I am still sending just the first cited row to `ensureRowsAreVisible` for now.

Issue #5779, opened 2026-02-04
Select menu option with Space


dstillman:
https://forums.zotero.org/discussion/comment/507016/#Comment_507016

On macOS too

Issue #5633, opened 2025-11-13
Empty context menu sometimes remains after opening a dialog from it


abaevbog:
1. Select some items in the itemTree
2. Open context-menu via right-click
3. Select "Create Bibliography from Items"
4. The empty context menu will appear on top of the dialog

Originally reported: https://forums.zotero.org/discussion/128049/beta-grey-rectangle-on-create-bibliography

I could reproduce it somewhat reliably on windows 11 VM:
<img width="757" height="633" alt="Image" src="https://github.com/user-attachments/assets/bde7eeac-449a-49d0-a7f2-33c58cfb0402" />

This is not specific to "Create Bibliography from Items" option - it happens with several context menu options that open dialogs, for example "Add to collection".

<img width="501" height="479" alt="Image" src="https://github.com/user-attachments/assets/924c6bdd-74a3-47bf-9a8a-613feb4d61c4" />

It seems to happen more frequently during syncing but not strictly then.

I wonder if it's related to the issue of dialogs and context menus that we worked around in https://github.com/zotero/zotero/pull/5409.

abaevbog:
https://forums.zotero.org/discussion/128935/ghost-right-click-menu-window

PR #5783, opened 2026-02-05
Space on an option of menulist dropdown will select it

abaevbog:
Fixes: #5779

For reference, this is what I mean by "empty context menu behind an alert" that the timeout is needed for (https://github.com/zotero/zotero/issues/5633):

Screenshot 2026-02-04 at 4 55 01 PM

### DIFF
--- a/chrome/content/zotero/customElements.js
+++ b/chrome/content/zotero/customElements.js
@@ -360,11 +360,11 @@ Services.scriptloader.loadSubScript('chrome://zotero/content/elements/itemTreeMe
 				}, { once: true });
 				target.open = false;
 			}
-			// No blinking happens on linux or windows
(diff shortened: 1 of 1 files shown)


dstillman:
https://github.com/user-attachments/assets/1fd60e4d-6a98-4684-87e8-6e62c748062d

This isn't properly triggering a selection effect on macOS. The first one is with Return, the second with Space.

abaevbog:
Good catch, thank you. It looks like on macOS, mozilla just adds and removes `_moz-menuactive` attribute 3 times with ~70ms interval. If we do the same, this is what it will look like. The first two keypresses are space, the last two keypresses are Enter:


https://github.com/user-attachments/assets/47b9b9d5-5679-40b6-8fe6-f3bea0f7c470

PR #5785, opened 2026-02-06
Reduce the judgment threshold for checking duplicate items by year

northword:
(I'm not sure how reasonable this is, and if you think this modification is not reasonable, please feel free to close it.)

Now, for the following two example items, Zotero marks them as duplicates, and this can be avoided only when `2021` is changed to `2022`. I think this threshold (one year) is a bit high and may lead to some false positives. Changing it to `>= 1` seems more reasonable.

```json
[
	{
		"id": "http://zotero.org/users/11729930/items/KIEL2LZB",
		"type": "article-journal",
		"language": "en",
		"title": "This is a example title",
		"issued": {
			"date-parts": [
				[
					"2020"
				]
			]
		}
	},
	{
		"id": "http://zotero.org/users/11729930/items/FSRZPSX5",
		"type": "article-journal",
		"language": "en",
		"title": "This is a example title",
		"issued": {
			"date-parts": [
				[
					"2021"
				]
			]
		}
	}
]
```

### DIFF
--- a/chrome/content/zotero/xpcom/duplicates.js
+++ b/chrome/content/zotero/xpcom/duplicates.js
@@ -359,7 +359,7 @@ Zotero.Duplicates.prototype._findDuplicates = async function () {
 			// If both items have a year and they're off by more than one, it's not a dupe
 			if (typeof yearCache[a.itemID] != 'undefined'
 					&& typeof yearCache[b.itemID] != 'undefined'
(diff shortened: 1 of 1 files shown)


AbeJellinek:
Not sure about this one. When working on translators, I've seen many, many cases where a paper gets published online late one year and then in print early the next year. Depending on your metadata source, you might end up with either one of those dates. I think it's correct to treat those as duplicates.

PR #5786, opened 2026-02-06
Add open note in tab context menu

ramonmi:
Since we have added Open Note as tab, this PR
- Introduce `Open Note in New Tab` context menu in both noteEditor toolbar (...) and rightpane context menu.
- Rename *Edit in a Separate Window* to *Open Note in New Window*, unified with the item menu strings.

### DIFF
--- a/chrome/content/zotero/elements/notesContext.js
+++ b/chrome/content/zotero/elements/notesContext.js
@@ -77,7 +77,8 @@
 				
 				<menupopup class="context-pane-list-popup">
 					<menuitem class="context-pane-list-show-in-library" label="&zotero.items.menu.showInLibrary;"/>
-					<menuitem class="context-pane-list-edit-in-window" label="&zotero.context.editInWindow;"/>
+					<menuitem class="context-pane-list-edit-in-tab" data-l10n-id="context-notes-edit-in-tab"/>
+					<menuitem class="context-pane-list-edit-in-window" data-l10n-id="context-notes-edit-in-window"/>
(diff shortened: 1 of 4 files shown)


windingwind (changes requested):
Thanks for the contribution. Please see below for comments.

ramonmi:
Thanks. I have pushed the new commit using the ftl in notesContext.

> I'm not sure if it's wise to change the existing strings.

Modifying a string that has been in use for five years may indeed affect the user's experience. But I don't think unified with the item menu strings (i.e., *Open Note in New Tab* and *Open Note in New Window*) is a bad thing.
Anyway, now the new context menu has been renamed to `Edit in a Built-in Tab`.

chrome/content/zotero/elements/notesContext.js
```diff
@@ -66,18 +66,19 @@
 			
 			<popupset>
 				<menupopup class="context-pane-add-child-note-button-popup">
-					<menuitem class="context-pane-add-child-note" label="&zotero.context.addChildNote;"/>
-					<menuitem class="context-pane-add-child-note-from-annotations" label="&zotero.context.addChildNoteFromAnnotations;"/>
+					<menuitem class="context-pane-add-child-note" data-l10n-id="context-notes-add-child-note"/>
```
windingwind: Maybe better not to touch existing strings, at least not in this PR, as these changes are irrelevant, I assume.

windingwind:
looks good to me.

PR #5787, opened 2026-02-06
Web-based login flow

dstillman:
Replaces the login form with a "Log In" button that creates a dataserver login session and opens the website for logging in. The client polls for session completion every few seconds and also watches for a notification from the streaming server.

TODO:

- Better UI
- Clearer text (i.e., about opening in browser)


dstillman:
Merged

PR #5789, opened 2026-02-09
Use grep -E instead of egrep

Yag000:
While updating my Zotero installation on Arch, I saw in the build output a bunch of warnings from `egrep` saying that it’s outdated:

> warning: egrep is obsolescent; using grep -E

It seems like it might be a good idea to change those calls to directly use `grep -E`, but there might be some scenarios where this could cause errors (I’m not familiar enough with it to ensure that making the change is perfectly safe).

I’ve updated the bash scripts to call `grep -E` instead of `egrep`.


dstillman:
Thanks!

PR #5795, opened 2026-02-11
Batch Editing

### LINKED ISSUE #5652: Batch editing
We have a very early prototype that @windingwind and I built. It still needs a lot of work and we also need to address editing creators which is going to be tricky. We might also want to consider merging advanced search into the main window at the same time but I feel like it could be handled separately.

Related issues:

#2119
#1920

tnajdek:
This PR adds batch editing support for fields, excluding the item type and creators.

https://github.com/user-attachments/assets/10c67274-5038-450c-8757-de971704891f

Close #5652

### DIFF
--- a/chrome/content/zotero/elements/collapsibleSection.js
+++ b/chrome/content/zotero/elements/collapsibleSection.js
@@ -43,7 +43,8 @@
 		set open(newOpen) {
 			newOpen = !!newOpen;
 			let oldOpen = this.open;
-			if (oldOpen === newOpen || this.empty || !this.collapsible) return;
+			if (oldOpen === newOpen || this.empty) return;
+			if (!newOpen && !this.collapsible) return;
 			this.render();
 			
 			// Force open before getting scrollHeight, so we get the right value
@@ -118,9 +119,22 @@
 				this.setAttribute('no-collapse', val);
 			}
 		}
-		
+
+		get showContextMenu() {
+			return !this.getAttribute("no-context-menu");
+		}
+
+		set showContextMenu(val) {
+			if (val) {
+				this.removeAttribute('no-context-menu');
+			}
+			else {
+				this.setAttribute('no-context-menu', val);
+			}
+		}
+
 		static get observedAttributes() {
-			return ['open', 'empty', 'label', 'summary', 'extra-buttons'];
+			return ['open', 'empty', 'label', 'summary', 'extra-buttons', 'no-collapse'];
 		}
 		
 		attributeChangedCallback(name) {
@@ -348,7 +362,7 @@
 		}
 		
 		_saveOpenState() {
-			if (this._disableSavingOpenState) return;
+			if (this._disableSavingOpenState || this._skipSaveOpenState) return;
 			Zotero.Prefs.set(`panes.${this.dataset.pane}.open`, this.open);
 		}
 		
@@ -383,11 +397,11 @@
 		}
 
 		get _disableContextMenu() {
-			return !this._getSidenav() || !!this.closest('annotation-items-pane');
+			return !this._getSidenav() || !this.showContextMenu || !!this.closest('annotation-items-pane');
 		}
 
 		_handleClick = (event) => {
-			if (this._disableCollapsing) return;
+			if (this._disableCollapsing || !this.collapsible) return;
 			if (event.target.closest('.section-custom-button, menupopup')) return;
 			this.open = !this.open;
 		};
@@ -417,7 +431,7 @@
 			}
 			// Space/Enter toggle section open/closed.
 			// ArrowLeft/ArrowRight on actual header will close/open (depending on locale direction)
-			if (["ArrowLeft", "ArrowRight", " ", "Enter"].includes(event.key) && !this._disableCollapsing) {
+			if (["ArrowLeft", "ArrowRight", " ", "Enter"].includes(event.key) && !this._disableCollapsing && this.collapsible) {
 				stopEvent();
 				this.open = ([" ", "Enter"].includes(event.key)) ? !this.open : (event.key == Zotero.arrowNextKey);
 				event.target.focus();
@@ -480,7 +494,7 @@
 			this._title.textContent = this.label;
 			this._summary.textContent = this.summary;
 			let twisty = this._head.querySelector('.twisty');
-			twisty.hidden = this._disableCollapsing;
+			twisty.hidden = this._disableCollapsing || !this.collapsible;
 			document.l10n.setAttributes(twisty, `section-button-${this.open ? "collapse" : "expand"}`, { section: this._paneName || "" });
 		}
 	}
--- a/chrome/content/zotero/elements/editableText.js
+++ b/chrome/content/zotero/elements/editableText.js
@@ -137,6 +137,14 @@
 			this.setAttribute('value', value || '');
 		}
 		
+		get values() {
+			return this._values ? this._values : [this.value];
+		}
+		
+		set values(values) {
+			this._values = values;
+		}
+		
 		get initialValue() {
 			return this._input?.dataset.initialValue ?? '';
 		}
@@ -178,6 +186,23 @@
 		
 		get ref() {
 			return this._input;
+		}
+
+		get multipleValues() {
+			return this.hasAttribute('multiple-values');
+		}
+		
+		set multipleValues(multipleValues) {
+			this.toggleAttribute('multiple-values', !!multipleValues);
+		}
+		
+		// true if the value is unchanged since the last blur event. Relevant for batch editing.
+		get cancelled() {
+			return this.hasAttribute('cancelled');
+		}
+
+		set cancelled(cancelled) {
+			this.toggleAttribute('cancelled', !!cancelled);
 		}
 
 		_resetTextDirection() {
@@ -275,7 +300,15 @@
 				}
 			}
 			this._input.readOnly = this.readOnly;
-			this._input.placeholder = this.placeholder;
+			if (this.readOnly && this.multipleValues) {
+				this._input.tabIndex = -1;
+			}
+			else {
+				this._input.removeAttribute('tabindex');
+			}
+			if (!(this.multipleValues && this.focused)) {
+				this._input.placeholder = this.placeholder;
+			}
 
 			if (this._input.tagName == "textarea") {
 				// Reset to initial state
@@ -381,6 +414,9 @@
 				this._ignoredWindowInactiveBlur = false;
 				return;
 			}
+			
+			this.cancelled = false;
+			this._clearValue = false;
 
 			let valueBeforeFocus = this.value;
 			this.dispatchEvent(new CustomEvent('focus'));
@@ -400,7 +436,15 @@
 			}
 
 			if (!('initialValue' in this._input.dataset)) {
-				this._input.dataset.initialValue = this._input.value;
+				this._input.dataset.initialValue = this.value;
+			}
+			
+			if (this.multipleValues) {
+				this._input.placeholder = '';
+				this._input.value = '';
+				if (this._input.mController) {
+					this._input.mController.startSearch("");
+				}
 			}
 		};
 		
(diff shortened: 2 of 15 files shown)


windingwind:
Thank you! Looks nice as an initial version. Code wise I feel like it's mostly safe to go. I noticed you added the auto complete, which is exactly what we wanted to have last time but wasn't able to due to time. Very nice job!

Some design wise suggestions:

1. The "Enter/exit batch editing button". Two major issues I can see:
  a) the enter and exit buttons are placed in faraway locations, which is not preferred from a user's perspective. Logically, I don't want to walk to the other side of the room to just turn off the light.
  b) The enter batch editing is a bit distracting. The most important info on the interface when multiple items are selected is the number of items, as we had in the past; while the enter batch editing should be a less frequently used feature than counting the numbers. The button is the only interactive element on the interface, which almost distract user's focus from the item counts.

2. The authors. We probably can leave the author batch editing for now, though we might better show them anyway on the surface (even as a readonly placeholder without actual author names), otherwise users can get panic that their author info might be gone when they click the exit batch editing.

3. The font of `Multiple`. I actually liked our blue italic for the special value, currently the gray one simply looks like a readonly effect.

4. Interaction wise. I would suggest making it crystal clear when the modifications will take effect, currently with the `Exit batch editing` it gives me a feeling that there will be a `confirm changes?` prompt after clicking the exit button and I can safely discard my changes, but actually it's saved immediately. Maybe using terms like `Close batch editing panel` with a smaller explanation that changes are immediately saved to selected items underneath if necessary.

tnajdek:
@windingwind thank you for the review and comments!

1a - I personally don’t see this as a problem; it’s more akin to a dialog: you press a button to open it, and once you’re done you press a close button, which is in a completely different place, to close it. Since you did work in the meantime, it doesn’t matter that the exit button is somewhere else. On top of that, selecting any other items exits the batch edit mode, so that button won’t be used too often.

1b - I’m not sure. I don’t see this as distracting, we can consider adding some padding or moving things

2 - We also remove sections other than “Info” and the item type field. I feel like if we communicate clearly enough that this is a different mode, then it’s OK to remove fields that cannot be batch edited.

3 - I agree, and I’ve brought it back. It makes the batch edit mode more distinctive. Updated screenshots below. I’d love to hear from @yexingsha whether we could make this, or something like it, work.

4 - We can tweak the wording if this turns out to be a problem.


Screenshot 2026-02-14 at 12 38 58
Screenshot 2026-02-14 at 12 38 54

yexingsha:
The blue is nice, though we should use `accent-blue`. I'm not sure about the italics, since the one other place where we use this is for annotations in the item tree, and here it has a totally different meaning. But it does accentuate the difference between Multiple and actual values, so maybe it's okay.

As to 4), after testing for a bit, I think the ideal solution is offering a way to discard changes, probably by changing the "Exit Batch Edit Mode" button to a Cancel and a Save button. If that's impractical due to syncing, then we probably need better warning before entering batch edit mode, and different wording for the button.

One other thing I found is that item titles (or other forms depending on the settings) are still showing up in header while in batch edit mode.

SR--:
If I may, a preview of changes before they are action is a normal expectation for batch-editing tools, and complete reversibility would be good too. Or too scary.

tnajdek:
@yexingsha thank you for the review & feedback!

> The blue is nice, though we should use `accent-blue`...

Changed to `accent-blue`.

> As to 4), after testing for a bit, I think the ideal solution is offering a way to discard changes, probably by changing the "Exit Batch Edit Mode" button to a Cancel and a Save button. If that's impractical due to syncing, then we probably need better warning before entering batch edit mode, and different wording for the button.

We currently have no mechanism for this, so implementing it would require more significant changes. Assuming we did implement it, what would happen if a user makes changes in batch editing and then selects another item -- would that cancel the changes? Wouldn't that be unexpected, considering that in "normal" mode changes are persisted?

I actually think a better solution would be to implement undo/redo that would work here and for any other metadata edits, though that's a pretty big task in itself.

> One other thing I found is that item titles (or other forms depending on the settings) are still showing up in header while in batch edit mode.

Thanks, fixed!

yexingsha:
> Assuming we would implement this, what would happen if user makes changes in batch editing then selects another item, would that cancel the changes? Isn't that unexpected, considering in "normal" mode changes are persisted?

For me, the ideal solution is enabling undo for any and all changes. Barring that, I think all changes made in batch editing mode need to be "saved" before going into effect. So in your example, there should be a warning dialog with a summary of unsaved changes, with the option to cancel, discard and continue, and save and continue.

northword:
Nice!

Is there a plan to integrate "batch replace/regular expression replace" in this feature?

This feature would be quite useful in the `title` field, especially in fields like biology, chemistry, materials, and environmental science, where people may have a large number of paper's titles containing subscripts, italics, and other rich text. With batch replace, we could convert `Co3O4` to `Co<sub>3</sub>O<sub>4</sub>`, as there is a lot of term repetition in the same field, which could significantly reduce the workload of maintaining reference information. And, Zotero's built-in sentence case conversion often incorrectly lowercases some proper nouns; batch replace could help mitigate this issue.

Of course, if it's not possible to implement this in Zotero core, we could also achieve it through an plugin.

tnajdek:
> For me, the ideal solution is enabling undo for any and all changes

Would it work if undo worked for all editing operations but not for creating and permanently deleting objects? Or would that be confusing? I've made a POC here: https://github.com/zotero/zotero/pull/5823

yexingsha:
This is fantastic! I think it's okay if undo doesn't work for permanently deleting objects, especially with clear warnings that it would be an irreversible action. As for creating objects, I think undo actions are still broadly useful (e.g. undoing an import that turns out wrong, or a note created by mistake, or adding items to a group library). And if people get used to undoing editing operations, it's likely they will try undoing creating objects at some point and get confused when it doesn't work.

That said, undo/redo for all editing operations is still great. I think it might be okay if we start with that, and then figure out if we need to support undoing object creation and how. @dstillman thoughts?

dstillman:
This works great! Haven't tried the undo yet, but some thoughts:



This seems like way too much new text. Can we just have an "Edit Multiple…" button and ditch the other text?



I actually…do not love this styling. Having differing fields here is the normal case, and something that can appear many times within the pane, so the blue (+ italic) feels totally excessive and jarring. It's not an error case that we need to draw attention to. We also show the current values in the drop-down. Gray non-italicized text seems fine to me?



For date fields with Multiple, the parsed-date tooltip and ymd indicator still appear. Neither should.



Doesn't make sense to collapse or reset order in this context, and if it is collapsed, that should probably be ignored. Just hide the disclosure triangle and context menu and force the pane open if collapsed, without affecting state?



Non-editable fields with Multiple (like all Date Added/Modified) shouldn't be selectable at all.



Wording aside, I don't really see the point of the "Exit Batch Edit Mode" button? As discussed above, you already exit this mode if you click on any item or change the selection at all. You can still see the number of selected items when in this mode. The only reason we have a button to enter this mode in the first place is that selecting multiple items is a very common action in Zotero, so just offering batch editing automatically would be kind of inappropriate and dangerous, but once you've made the choice to enter it, there's no particular reason to leave it short of clicking elsewhere.





When you first click on a Multiple field, the placeholder disappears and the menu appears. If you then type, both disappear. If you backspace to empty, the menu reappears, but so does the placeholder. It should remain hidden.

Also, on macOS, I'm always seeing the scrollbar for multiple rows, even when it's show them all.

tnajdek:
@dstillman Thanks for the review! Good catches - I totally missed those!

I've implemented all the requested changes. The autocomplete scrollbar bug was a pre-exisiting issue so I've fixed it in main.

My concern with the current version is that it too easy to forget that batch edit mode is enabled. With the button in the header and blue `Multiple` labels it was much harder to forget that multiple items are affected when editing a field. But maybe it's OK, I don't know.

dstillman:
Maybe we need some different styling on the bar with the item count at the top? @yexingsha?

dstillman:
Well, and it should even say "Editing 12 items", right?

yexingsha:
The item count at the top is supposed to use the same styling as item title. And I agree that changing the wording to "Editing X items" is clearer.

But I'm still concerned it's too easy to click away prematurely and make accidental batch edits. I think with undo/redo functionality, it might be okay to leave out the cance/save buttons, but otherwise we do need them.

dstillman:
> The item count at the top is supposed to use the same styling as item title

As the item pane header, you mean?

yexingsha:
Yes, I phrased it that way because the header can have different configurations.

dstillman:
Right. OK, let's see how that looks, along with the new wording.

tnajdek:
Updated 
Screenshot 2026-03-10 at 13 38 57
Screenshot 2026-03-10 at 13 38 51

windingwind:
Very innovative discussions and designs above. Some comments motivated by the previous discussions:

1. 
The gray text for `multiple`, speaking for myself, doesn't make it very clear what that means/differs compared with its surrounding actual values, without extra explanations. Even if we may have some first-run popups, I am likely to forget them after a while. I believe a good design explains itself. Another extreme case is when some fields actually have a value of textual "Multiple", which is not of chance 0, it would look confusing.

Since we have candidates from selected items for those `multiple`s, which made it in fact a hybrid of text input and menu list; would it be clearer if we have some menu list style decorations on the UI, like a small drop down (either to by default visible or on hover visible) that indicates this field is actually linked a multiple items' same field, instead of a readonly field with value "Multiple"? Even if we might show readonly fields in batch editing differently, I don't feel like this looks distinguishable enough from a readonly field, because users are already very used to the assumption of "gray === readonly" in mind from many other places.

2. 
The `Editing N items` also looks very close to (exactly the same as) the item title. Similarly, in case item(s) actually have a title like "Editing" or even "Editing N items", which I believe some people would use the former at least, like I myself, sometimes just write random words for items that are manually created to contain PDFs of in-progress materials.

I would suggest putting an icon at the start of the title line, like a pencil or something similar, so that it doesn't look like a regular item title.

dstillman:
Yeah, I think an icon and/or some other styling for the header would be good. This really shouldn't have exactly the same styling as an item title.

For "Multiple", I don't think the gray itself is a problem. Gray doesn't just mean read-only/disabled — it's also just general placeholder text (e.g., in a search field). But showing some sort of icon or indicator that you can click to see the individual values could be helpful. We could also consider presenting the other values in a way that looks less like autocomplete.

It also occurs to me that there's no way (I think?) to clear all values. Mostly people will use that when they shouldn't (e.g., deleting all DOIs), but it should still be possible. Maybe need to add an empty value (presented…somehow) as an option in the menu?

dstillman:
(And the field still has a hover effect. I don't think it's unclear that you can click on it. It's maybe just a little unclear what will happen if you do.)

SR--:
Great to see this development. A couple of suggestions based on my very limited understanding of where the development is at:

1. Consider providing a complete textual preview of all changes and/or a complete difference log after the fact, especially if the Undo is complex and not 100%, maybe also some way to report on what happens during Undo. E.g., see tools such as https://www.joejoesoft.com/vcms/108/ for an example.
2. Consider restricting batch editing ability in groups to some select users. I understand that this might be complex and beyond scope, but in a group with a lecturer and 200 students, you do not want a confused novice destroying the work for many others. Perhaps you could at least let the group admin enable or disable batch editing for the group as a compromise?

Thanks for considering...

tnajdek:
> It also occurs to me that there's no way (I think?) to clear all values.
 
 Pressing enter with an empty field sets the field to an empty value on all selected items. I thought that’s intuitive but maybe not as discoverable as I thought.

dstillman:
Oh, I worry that it's dangerous. Since it just looks like an empty field, I think Enter has to be a no-op, as it is on a normal field. I think you should have to affirmatively choose to clear all values — with a blank row in the drop-down or something else.

yexingsha:
I think a lot of these decisions hinge on whether we will launch this with undo (even if limited to metadata edits). If there is undo, then hitting enter to clear values is no more dangerous than accidentally exiting batch editing with half-typed values. If there is no undo, then I really think it's necessary to put batch edits in a pending state that has to be confirmed with save/cancel buttons, or a dialog if you click outside.

So in the case of launching without undo, a safe batch editing mode could look like this:

item-pane

- Add cancel/save buttons to the header area, which I think sufficiently differentiates it from item titles;
- Use `Multiple…` to better indicate that they can be interacted with;
- Add `(No value)` in gray as an option in the menu;
- Show value counts to provide more context and better distinguish the menu from autocomplete. This also has the benefit of indicating when some of the selected items have this field empty.

dstillman:
I just haven't looked at it yet, but I assume we can have Undo/Redo. I'd really like to avoid Cancel/Save — it's just kind of off-brand for us!

tnajdek:
Updated the UI (see below). We're using a customized auto-complete component to display the options and that can take only one value (the label). I've put the items counter in square brackets. Chasing exact design match here would require some sort of a custom component with floating placement. I don't think it's worth it, is it?

I agree we want this with undo rather than cancel. I'll update the branch with undo and give it another look.

Screenshot 2026-03-27 at 22 13 15

tnajdek:
Superseded by #5823

PR #5797, opened 2026-02-11
ItemTree: + expands all rows one level at a time

abaevbog:
On `+` keypress, expand the rows down one level, instead of expanding all rows at once. If some rows are expanded and some are collapsed, expand to the deepest level currently visible.
E.g. if some regular items are already expanded, expand only remaining collapsed regular items. If there is a row with expanded attachments, expand all rows. That way, after `+` keypress, all opened rows are of the same level.

On `-` keypress, collapse the rows up one level to be consistent.

https://forums.zotero.org/discussion/129596/optional-annotations-in-item-list

https://github.com/user-attachments/assets/a772bb5d-ee96-47c2-87fa-e78ba715ea01

### DIFF
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -1831,8 +1831,16 @@ var ItemTree = class ItemTree extends LibraryTree {
 	 *  @param {Boolean} acrossAllLevels - Expand rows across all levels
 	 */
 	expandAllRows(acrossAllLevels) {
+		// Do nothing if there are no rows
+		if (this.rowCount == 0) return;
 		this.selection.selectEventsSuppressed = true;
 		var selectedItems = this.getSelectedObjects();
+		// Remember scroll position of an anchor row relative to the viewport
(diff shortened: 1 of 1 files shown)


dstillman:
Can we try to preserve the position of the 1) selected row or, if none, 2) the top row in the items list?

abaevbog:
How about this? We set the anchor to be the focused row if it is visible OR the first visible row otherwise. Then, we restore the scroll position of the anchor after expanding or collapsing. After collapsing, it's possible that the anchor was the now-collapsed child row - in that case, we scroll it's parent to the top. I think with this, it feels like the most amount of scroll context is being preserved: 

https://github.com/user-attachments/assets/7a6e7bc0-0fc7-40e1-b0d9-492738e8a108


https://github.com/user-attachments/assets/b90188da-8290-47a5-b70a-b0d7b8bb2e28

dstillman:
https://forums.zotero.org/discussion/130365/expand-all-shift-in-zoteros-behavior-changed

PR #5800, opened 2026-02-16
Memory optimizations: citeproc-js, CSL.Engine caching, and DataObject properties

dstillman:
In my test library, these commits together reduce memory usage at startup by ~260 MB — a 20% reduction.

It should also dramatically speed up CSL.Engine caching at startup (e.g., 8x for one Chicago style).


dstillman:
@AbeJellinek: For the second commit, did we have any particular reason for caching both `text` and `html` engines? All tests pass with just one.

PR #5802, opened 2026-02-19
Item tree refactor

adomasven:
## Summary

This is a huge Item Tree refactor, focused on separating the rendering and data related logic of the Item Tree, as well as making the Item Tree more flexible for displaying Item rows from sources other than Zotero Collections.

The need for this flows out of requirements for https://github.com/zotero/zotero/pull/3468 and other needs springing from more sophisticated uses of ItemTree in various Zotero UIs. With the row data and row rendering separated we can more easily create ItemTree views that mix items and non-items, as well as allow Zotero Items to be children of other items (such as section groupings, or library groupings, as required in Citation Explorer and others).

I may have not taken this refactor up if coding models were not as good, or I may have reduced its scope significantly. I knew that the ItemTree could use with a refactor since original VirtualizedTree implementation. Since then we've added more rendering behaviour to the tree, like annotation rows, which has made it even more complex, and as such the refactor even more warranted.

No actual behaviours are changed here, just the code. I also tried to keep most previous APIs available and signatures unchanged to not break extensions and such.

### 1) Row/model/render refactor
Refactored ItemTree internals into clearer units (`ItemTreeRow`, `ItemTreeRowProvider`, `ItemTreeRowRenderer`)

### 2) Ownership split into `ItemTree` and `CollectionViewItemTree`
ItemTree now doesn't depend on an "attached" collectionTreeRow, and instead expects a custom ItemTreeRowProvider which can implement any custom row data provision logic. Custom non-Zotero.Item rows can be provided and their display and other behaviour configured by subclassing ItemTreeRowRenderer.

### 3) Cache ownership and invalidation improvements
- A full decoupling of notify() behaviour between base ItemTree case, where 'modify' is respected and 'refresh' are respected, and CollecionViewItemTree, where 'add' events are considered.
- Row invalidation (redraw) improvements and normalization across the ItemTree methods.
- Moved CollectionTreeCache into CollecionTreeRow, where it should belong

### 4) Miscellaneous fixes
- Linux context-menu behavior
- ConcurrentCaller unhandled async error fix
- Others

## Tests

- Expanded/refactored tests around:
  - collection-view item tree behavior
  - virtualized table behavior
  - collection tree row cache/error behavior
  - tag selector handling on search failures
- Added regression coverage for toggle-open scroll behavior and sort-context readiness.
- Unskipped and modernized previously skipped ItemTree tests.

## Key architectural decisions:

- Any tree rerendering that needs to occur due to data changes (item data modified, an parent item is expanded/collapsed and new rows appear in the visible tree, etc) is signaled in `ItemTreeRowProvider.onUpdate` and handled by `ItemTree.handleRowProviderUpdate()`.
- Public `ItemTreeRowProvider` methods always issue a `onUpdate` event, private methods do not.
- Any behaviour related to an attached `collectionTreeRow` to the `ItemTree` lives in `CollectionViewItemTree`.

------------------------

It's hard to review these kind of changes, but still I'd like @tnajdek, @AbeJellinek, @abaevbog and/or @windingwind to take a look if you have time and provide comments, if any and merge this without too much delay.

Also @windingwind you are probably best positioned to say whether this impacts plugin developers or ItemTree plugin API.

### DIFF
--- a/chrome/content/zotero/advancedSearch.js
+++ b/chrome/content/zotero/advancedSearch.js
@@ -24,7 +24,7 @@
 */
 
 
-import ItemTree from 'zotero/itemTree';
+import CollectionViewItemTree from 'zotero/collectionViewItemTree';
 import { COLUMNS } from 'zotero/itemTreeColumns';
 
 
@@ -39,6 +39,7 @@ var ZoteroAdvancedSearch = new function () {
 
 	var _searchBox;
 	var _libraryID;
+	var _searchCounter = 0;
 	
 	async function onLoad() {
 		_searchBox = document.getElementById('zotero-search-box');
@@ -61,16 +62,16 @@ var ZoteroAdvancedSearch = new function () {
 			column.hidden = !['title', 'firstCreator', 'year', 'hasAttachment'].includes(column.dataKey);
 			return column;
 		});
-		this.itemsView = await ItemTree.init(elem, {
+		this.itemsView = await CollectionViewItemTree.init(elem, {
 			id: "advanced-search",
 			dragAndDrop: true,
-			persistColumns: true,
 			columnPicker: true,
 			onActivate: this.onItemActivate.bind(this),
 			columns,
 		});
 
 		await this.itemsView.changeCollectionTreeRow({
+			id: 'advanced-search-' + _searchCounter++,
 			ref: _searchBox.search,
 			visibilityGroup: 'default',
 			isSearchMode: () => true,
@@ -103,7 +104,8 @@ var ZoteroAdvancedSearch = new function () {
 		_searchBox.updateSearch();
 		_searchBox.active = true;
 		
-		var collectionTreeRow = {
+		return this.itemsView.changeCollectionTreeRow({
+			id: 'advanced-search-' + _searchCounter++,
 			ref: _searchBox.search,
 			visibilityGroup: 'default',
 			isSearchMode: () => true,
@@ -115,21 +117,8 @@ var ZoteroAdvancedSearch = new function () {
 				search.libraryID = _libraryID;
 				var ids = await search.search();
 				return Zotero.Items.get(ids);
-			},
-			isLibrary: () => false,
-			isCollection: () => false,
-			isPublications: () => false,
-			isDuplicates: () => false,
-			isFeed: () => false,
-			isFeeds: () => false,
-			isFeedsOrFeed: () => false,
-			isRecentlyRead: () => false,
-			isSortable: () => true,
-			isShare: () => false,
-			isTrash: () => false
-		};
-
-		return this.itemsView.changeCollectionTreeRow(collectionTreeRow);
+			}
+		});
 	}
 	
 	
--- a/chrome/content/zotero/collectionTree.jsx
+++ b/chrome/content/zotero/collectionTree.jsx
@@ -26,7 +26,7 @@
 const React = require('react');
 const ReactDOM = require('react-dom');
 const LibraryTree = require('./libraryTree');
-const VirtualizedTable = require('components/virtualized-table');
+const VirtualizedTree = require('components/virtualized-table').VirtualizedTree;
 const { getCSSIcon } = require('components/icons');
 const { getDragTargetOrient } = require('components/utils');
 const { noop } = require("./components/utils");
@@ -70,6 +70,8 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 		this.type = 'collection';
 		this.name = "CollectionTree";
 		this.id = "collection-tree";
+		this._rows = [];
+		this._rowMap = {};
 		this._highlightedRows = new Set();
 		this._unregisterID = Zotero.Notifier.registerObserver(
 			this,
@@ -300,12 +302,6 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 		// Div creation and content
 		let div = oldDiv || document.createElement('div');
 		div.innerHTML = "";
-		// When a hidden focused row is added last during filtering, it
-		// is removed on focus change, which can happen at the same time as rendering.
-		// In this case, just return empty div.
-		if (index >= this._rows.length) {
-			return div;
-		}
 		
 		// Classes
 		div.className = "row";
@@ -463,7 +459,7 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 	}
 	
 	render() {
-		return React.createElement(VirtualizedTable,
+		return React.createElement(VirtualizedTree,
 			{
 				getRowCount: () => this._rows.length,
 				id: this.id,
@@ -478,7 +474,7 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 				isContainer: this.isContainer,
 				isContainerEmpty: this.isContainerEmpty,
 				isContainerOpen: this.isContainerOpen,
-				toggleOpenState: this.toggleOpenState,
+				onToggleOpenState: this.toggleOpenState,
 				getRowString: this.getRowString.bind(this),
 				
 				onItemContextMenu: (...args) => this.props.onContextMenu && this.props.onContextMenu(...args),
@@ -486,7 +482,6 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 				onKeyDown: this.handleKeyDown,
 				onActivate: (...args) => (this.props.onActivate ? this.props.onActivate(...args) : this.handleActivate(...args)),
 
-				role: 'tree',
 				label: Zotero.getString('pane.collections.title')
 			}
 		);
--- a/chrome/content/zotero/collectionViewItemTree.jsx
+++ b/chrome/content/zotero/collectionViewItemTree.jsx
@@ -1,0 +1,1918 @@
+/*
+	***** BEGIN LICENSE BLOCK *****
+	
+	Copyright © 2024 Corporation for Digital Scholarship
+                     Vienna, Virginia, USA
+					http://zotero.org
+	
+	This file is part of Zotero.
+	
+	Zotero is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published by
+	the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+	
+	Zotero is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with Zotero.  If not, see <http://www.gnu.org/licenses/>.
+	
+	***** END LICENSE BLOCK *****
+*/
+
+/**
+ * CollectionViewItemTree - Item tree for collection-based views.
+ *
+ * This class extends ItemTree with behaviors for views backed by a CollectionTreeRow:
+ * - Drag and drop with library/collection awareness
+ * - Duplicates set selection
+ * - Feed-specific sorting
+ * - Colored tag keyboard shortcuts
+ * - Intro/welcome text
+ * - Collection-aware delete behavior
+ * - Quick search integration
+ *
+ * Used in ZoteroPane, Advanced Search, and other collection-based contexts.
+ */
+
+const { getDragTargetOrient } = require("components/utils");
+const React = require('react');
+const ReactDOM = require('react-dom');
+const ItemTree = require('zotero/itemTree');
+const { ItemTreeRowProvider } = ItemTree;
+
+const { OS } = ChromeUtils.importESModule("chrome://zotero/content/osfile.mjs");
+const { ZOTERO_CONFIG } = ChromeUtils.importESModule('resource://zotero/config.mjs');
+
+const COLORED_TAGS_RE = new RegExp("^(?:Numpad|Digit)([0-" + Zotero.Tags.MAX_COLORED_TAGS + "]{1})$");
+
+// Minimal CollectionTreeRow-like object for callers that pass plain objects to
+// changeCollectionTreeRow()/setCollectionTreeRow() (e.g. advanced search).
+const STUB_COLLECTION_TREE_ROW = {
+	// Properties set by CollectionTreeRow constructor
+	view: null,
+	type: null,
+	ref: {},
+	level: 0,
+	isOpen: false,
+	onUnload: null,
+	searchText: "",
+	searchMode: "search",
+	tags: [],
+	
+	// Extra props/methods expected by CollectionViewItemTree
+	visibilityGroup: "",
+	getItems: async () => [],
+	isSearchMode: () => false,
+	isLibrary: () => false,
+	isCollection: () => false,
+	isSearch: () => false,
+	isPublications: () => false,
+	isDuplicates: () => false,
+	isFeed: () => false,
+	isFeeds: () => false,
+	isFeedsOrFeed: () => false,
+	isShare: () => false,
+	isTrash: () => false,
+	isBucket: () => false,
+	isUnfiled: () => false,
+	isRetracted: () => false,
+	isRecentlyRead: () => false,
+	isSortable: () => true,
+	setSearch: () => false,
+	setTags: () => false,
+	clearCache: () => {}
+};
+
+class CollectionViewItemTreeRowProvider extends ItemTreeRowProvider {
+	constructor(itemTree) {
+		super(itemTree);
+		this.collectionTreeRow = null;
+	}
+
+	/**
+	 * Check if the quick search box has a search term.
+	 * @returns {boolean}
+	 */
+	hasQuickSearch() {
+		return this.collectionTreeRow?.searchText.length > 0;
+	}
+
+	/**
+	 * Set a new collectionTreeRow and refresh items.
+	 * This handles the data/model logic; UI orchestration stays in ItemTree.
+	 * @param {Object} collectionTreeRow - The collection tree row to set
+	 * @returns {Promise<void>}
+	 */
+	async setCollectionTreeRow(collectionTreeRow) {
+		// Normalize stub objects to include default CollectionTreeRow methods
+		if (collectionTreeRow.constructor.name == "Object") {
+			collectionTreeRow = Object.assign({}, STUB_COLLECTION_TREE_ROW, collectionTreeRow);
+		}
+		// No-op if same collection
+		if (this.collectionTreeRow && this.collectionTreeRow.id === collectionTreeRow.id) {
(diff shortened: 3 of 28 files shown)


abaevbog:
@adomasven, this is great! I will make sure to go over everything here in more details but so far, I just wanted to note that with this change, the extra right-most column with `+` icons in the citation dialog does not appear for me. It is marked as `hidden`. Does it appear for you (it's not impossible the issue is somewhere on my end)?


After:
Screenshot 2026-02-25 at 1 14 29 PM

adomasven:
@abaevbog Fixed, but you may need to reset your column prefs, as a mistaken hidden=true value may be written to them with no way to undo it in the UI (column not available in picker).

abaevbog:
Great! Yes, that fixes it. I like how this refactor allows to create custom pieces of itemTree-based components. We have a lot of use cases for it. Let me go over the ones I am aware of below to make sure I understand how they could be implemented now with these changes. This is a very big refactor, so I think we might as well try to account for as many future use cases as possible.

1. **Group items by libraries.** For example, in citation explorer we discussed rendering items grouped by their library. We would also require the same for advanced search to allow one to search across all libraries.
2. **Show all hidden annotation rows container**.  Currently, by default, we hide all non-matching annotation rows during search because otherwise context rows overwhelm the UI. [It was reported](https://forums.zotero.org/discussion/comment/508012/#Comment_508012) that some way to access  non-matching annotations on a cases by case basis may still be useful. The most feasible solution would be to hide context rows in a collapsed "Show all" container row that would be a pseudo-child of the attachment.
3. **'Everything' quicksearch previews**. An old request to display [previews](https://forums.zotero.org/discussion/comment/491123) of search results in `Everything` mode of quick search, so one can see where exactly the search match comes from. Search results would likely be pseudo children of the attachment's container row.

To group items by the library, we would create a new class extending `ItemTreeRowProvider` (e.g. `groupedByLibrariesItemTree`). There, we would implement custom `_refresh` to fetch the items and group them by libraries, `_toggleOpenState` to handle opening of the new library container (where regular items would end up starting on level 1), etc. Then, we would also extend `ItemTree` to handle actual rendering, with section rows being treated somewhat specially (like annotation rows). So that seems pretty straighforward! That would work very well for the citation explorer. But if we wanted to do the same for the advanced search, we would want to have the same `notify()` logic in that `groupedByLibrariesItemTree` as in the current `collectionViewItemTree` (so that if you update an item, its update is reflected in groupped advanced search view). Same with drag-drop logic. In that case, to be able to group items by libraries for advanced search, we would have to plug it into existing `collectionViewItemTree`, which seems tricky.

To hide context annotation rows behind a pseudo container "Show all annotation rows", we could adjust the current `_toggleOpenState` to check if a given attachment row has context annotation rows and append a pseudo child "Show all non-matching annotations". In turn, `_toggleOpenState` on that row will display/hide context rows of that attachment. Finally, we would adjust the remaining logic for it in `renderItem`, `_getIcon`, `isContainer`, etc. To be able to show search previews, we would do something very similar to annotation rows, where they are just added as a special type of a child row of an attachment.

--- 

Does this sound correct so far? Is so, these are the two ideas I have:

1. I think it might be nice to be able to more easily extend `CollectionViewItemTreeRowProvider` and `CollectionViewItemTreeRow`, so that one could take advantage of the `notify` handling, drag-drop, and other logic relevant to ZoteroPane, while constructing new types of UI. Grouping by libraries for search across all groups is one example, but I can imaging grouping by other properties (e.g. for plugins). It seems challenging to do it now because the logic of figuring out what rows should be in `collectionViewItemTree` is coupled with how the rows should be arranged (e.g. `_refresh` places all regular item rows as level 0, which would not be right if they're grouped by the library).
Would it be worth considering to separate that logic into 2 steps: fetching which entries need to be present/added/removed in the itemTree (mainly first part of `_refresh`), and how those entries should be arranged? If we were to pull the first step into its own method (e.g. `fetchEntriesToRender`, or something like that), it would be called in `_refresh` to get the right items and the rest of `_refresh` would arrange them as needed. Then, one could create a new class extending `CollectionViewItemTreeRowProvider` (e.g. `GroupedByLibrariesItemTreeRowProvider`) that would override the `_refresh` method to arrange items differently (e.g. by placing libraries as rows with level 0), while still relying on the rest of the `notify` and drag-drop logic necessary in ZoteroPane.

2. Adding custom child (pseudo) rows now requires adding special handling like with annotations across many functions. It may be even trickier because annotations are proper `Zotero.Item`s but, say, search result previews would not be. Would it be worth considering moving some of the functionality related to how rows behave (e.g. how to render, if they are a container, what are the row's children, etc.) into the `ItemTreeRow` itself? It could implement base functions like `isContainer`, `getChildren`, `renderRow`, `getIcon`, etc., and a new type of rows would extend the base class and override functions if necessary. Then, adding a new functionality would be mainly a matter of creating a new class without touching the rest of the itemTree-related logic. Then, for example, `_toggleOpenState` on a row would just call `row.getChildren` to get child rows instead of having to contain all the logic about which rows can have what kind of children. I imagine it could be useful for the plugins to create their own custom rows as well.

---

These are my thoughts! I did my best to go over the main changes in this refactor but it is quite big, so apologies if any of the above is not appropriate.

abaevbog:
On a different, much less abstract, note, these are two additional glitches I recently ran into:

1. Not all rows matching a search have their parent expanded.

https://github.com/user-attachments/assets/763f0718-6a5c-4949-bf04-5a3ec44bc973

2. Currently, `ItemTreeRow` has a `type` that's always "item". This is not always correct - an item tree row can also be a saved search or a collection in the trash. Their behavior though is handled in `collectionViewItemTree` by checking the type or `row.ref`, so it causes no issues. But we probably want it consistent.

windingwind:
I checked the Plugin API part, and it's working well.

adomasven:
> 3\. **'Everything' quicksearch previews**. An old request to display [previews](https://forums.zotero.org/discussion/comment/491123) of search results in `Everything` mode of quick search, so one can see where exactly the search match comes from. Search results would likely be pseudo children of the attachment's container row.

This would be possible, although I don't think that's the correct UI decision.

> To group items by the library, we would create a new class extending `ItemTreeRowProvider` (e.g. `groupedByLibrariesItemTree`). There, we would implement custom `_refresh` to fetch the items and group them by libraries, `_toggleOpenState` to handle opening of the new library container (where regular items would end up starting on level 1), etc. Then, we would also extend `ItemTree` to handle actual rendering, with section rows being treated somewhat specially (like annotation rows). So that seems pretty straighforward! That would work very well for the citation explorer. But if we wanted to do the same for the advanced search, we would want to have the same `notify()` logic in that `groupedByLibrariesItemTree` as in the current `collectionViewItemTree` (so that if you update an item, its update is reflected in groupped advanced search view). Same with drag-drop logic. In that case, to be able to group items by libraries for advanced search, we would have to plug it into existing `collectionViewItemTree`, which seems tricky.

This isn't entirely right at least in my conception.
1. Generally, most things should be doable by using `ItemTree`, and providing a custom `ItemTreeRowProvider` and custom `ItemTreeRowRenderer._refresh()` is a pattern that is only really applicable to `CollectionViewItemTree` due to the fact that the ItemTree view is "attached" to the selected collection tree. If the idea of a "selected" collection tree isn't applicable, then it's up to the ItemTreeRowProvider and whatever controls it to notify the ItemTree that it needs redrawing.
2. For things that need `notify()` logic, depending on the type of interface, custom accommodations will be required, and as such reusing what `CollectionViewItemTree` has is unlikely to be useful. The default `ItemTree.notify()` handles only `refresh` and `modify` events to show updated metadata, but leaves the rest up to the child class to do, if anything is needed.
3. Regarding drag-and-drop, it also doesn't necessarily make sense to handle it if you have an item view that is not attached to a selected collection. I don't think it's a good idea to allow many places to support "reorganizing" the Zotero library, aside from the library window, and also for a custom view that may display results from a variety of sources a drop operation is not necessarily well defined or desirable.

On the other hand, if some or most of the capabilities of `CollectionViewItemTree` are required, it can be easily subclassed, or we can further refactor it if need be. The best part about this refactor is that it separated enough concerns of the ItemTree that it becomes feasible to make variations of it with custom behaviour instead of just stacking everything into the same file with `if ()` statements for an increasingly large number of various custom ways in which ItemTrees need to appear in Zotero views.

As a follow up to this I'll do the first proper non `CollectionViewItemTree` subclass of `ItemTree` in `CitationDialog`, where we'll have some more hashed out details on how this works in practice.



> 1. I think it might be nice to be able to more easily extend `CollectionViewItemTreeRowProvider` and `CollectionViewItemTreeRow`, so that one could take advantage of the `notify` handling, drag-drop, and other logic relevant to ZoteroPane, while constructing new types of UI. Grouping by libraries for search across all groups is one example, but I can imaging grouping by other properties (e.g. for plugins). It seems challenging to do it now because the logic of figuring out what rows should be in `collectionViewItemTree` is coupled with how the rows should be arranged (e.g. `_refresh` places all regular item rows as level 0, which would not be right if they're grouped by the library).
>    Would it be worth considering to separate that logic into 2 steps: fetching which entries need to be present/added/removed in the itemTree (mainly first part of `_refresh`), and how those entries should be arranged? If we were to pull the first step into its own method (e.g. `fetchEntriesToRender`, or something like that), it would be called in `_refresh` to get the right items and the rest of `_refresh` would arrange them as needed. Then, one could create a new class extending `CollectionViewItemTreeRowProvider` (e.g. `GroupedByLibrariesItemTreeRowProvider`) that would override the `_refresh` method to arrange items differently (e.g. by placing libraries as rows with level 0), while still relying on the rest of the `notify` and drag-drop logic necessary in ZoteroPane.

As for specifics of something like this, I think you'd just subclass and create a custom row provider. Generally we have two routes via which the rows displayed in the main display of the item tree change:
1. Change the selected collection tree row (`refresh()` route)
2. A notify event comes in that modifies, adds or removes an item that belongs in the display of the current collection tree row (`notify()` route).

Search views are implemented as a "pseudo-change" of selected collection tree row. But generally what search does, how its items list is built, and what needs to happen when `notify()` events come is fairly different with little overlap than what happens in a collection-row attached view. The current `refresh()` workflow jams both collection views and search views into the same function and loading pipeline, and it doesn't have to be this way.

As such, for any other novel views of items (and they're not just search views, but e.g. for citation explorer a view of items where some of them are part of the library, but some are unlinked, or perhaps one day we could show a preview items list of items being imported from RDF, etc), the flow of information is something like:
1. Update the data layer. In practice, this means something, say a changed filter text, triggers a function in a custom `ItemTreeRowProvider`, which updates the internal `_rows` and issues and `update` event, and the tree updates the display of rows based on that.
  For the item tree in the citation explorer we have an unchanging list of items that are not part of the Zotero databse, that you're filtering over. Right now that is implemented as pseudo collection tree changes. In the future this would be implemented as a function call on the RowProvider and a handled `update` event in the ItemTree that redraws those rows.

> 2\. Adding custom child (pseudo) rows now requires adding special handling like with annotations across many functions. It may be even trickier because annotations are proper `Zotero.Item`s but, say, search result previews would not be. Would it be worth considering moving some of the functionality related to how rows behave (e.g. how to render, if they are a container, what are the row's children, etc.) into the `ItemTreeRow` itself? It could implement base functions like `isContainer`, `getChildren`, `renderRow`, `getIcon`, etc., and a new type of rows would extend the base class and override functions if necessary. Then, adding a new functionality would be mainly a matter of creating a new class without touching the rest of the itemTree-related logic. Then, for example, `_toggleOpenState` on a row would just call `row.getChildren` to get child rows instead of having to contain all the logic about which rows can have what kind of children. I imagine it could be useful for the plugins to create their own custom rows as well.

That's an interesting suggestion. Right now some of these functions are in `RowProvider`, like `isContainer()`, while others like `getIcon` are part of the `RowRenderer`. `ItemTreeRow` however is a rather shallow class, more like a Struct. Rows knowing how to draw themselves, and what behaviours they support (dragging, dropping, toggling open state) might simplify this further in a good way. `RowRenderer` probably would cease to exist, and it would indeed become very easy to implement custom row types, which is good for plugins. There's some logic that depends on the tree state in rendering, e.g. for search results, highlighted items, and this probably doesn't split as cleanly, but overall I think this would be better. I'm going to go ahead and pursue this.

dstillman:
Good points on the future cases, but I'd add:

> 'Everything' quicksearch previews

This isn't necessary just Everything — search snippets in general have been a longstanding, popular request and could be useful even for All Fields & Tags mode. I don't think we'd show them for all searches, at least by default, but definitely something we want to make sure we're able to support.

We can get mockups from @yexingsha for anything where it's unclear exactly how we might want to support it UI-wise (e.g., for a Show All for non-matching rows, or some of the requirements for https://github.com/zotero/zotero/pull/3468#issuecomment-3523994165), in case that affects implementation decisions.

adomasven:
Updated and search issue addressed.

> 2\. Currently, `ItemTreeRow` has a `type` that's always "item". This is not always correct - an item tree row can also be a saved search or a collection in the trash. Their behavior though is handled in `collectionViewItemTree` by checking the type or `row.ref`, so it causes no issues. But we probably want it consistent.

I don't know how I feel about this property. Since rows provide their own renderers, etc, and you can check the `treeRow.ref instanceof`, it feels a little redundant. I might have a more informed opinion once I do Citation Explorer updates. We might want to remove that.

abaevbog:
I think this looks great! I like how new types of rows are just confined to `itemTreeRow`, and the rest of the itemTree doesn't actually need to know much about what kind of item is being rendered.  

> I don't know how I feel about this property. Since rows provide their own renderers, etc, and you can check the treeRow.ref instanceof, it feels a little redundant

I imagine we may have future use cases for the `type` property. I think there's value in having the flexibility to be able to not just render instanced of `Zotero.DataObject`s but to be able to render arbitrary objects. For instance, a container row "Show more" that would contain child annotations that do not match the current search query. Or maybe some kind of search result preview. For those, we would not have the same `.ref` as now, so `.type` could be used to indicate what kind of row it is then.

---

When I worked on adding trashed collections and searches, I had to resort to a workaround to allow collections and searched to pretend to be items when they are displayed in the trash. Many parts of `itemTree` would call item's methods like `.isRegularItem`,  `isAttachment`, `isAnnotation` , etc. that collections and searches do not have. I figured it would be easier to mock those properties in [Zotero.Collection](https://github.com/zotero/zotero/blob/f130e3b3e5869b8d211890290dd3af68d25efb35/chrome/content/zotero/xpcom/data/collection.js#L109-L110) and [Zotero.Search](https://github.com/zotero/zotero/blob/f130e3b3e5869b8d211890290dd3af68d25efb35/chrome/content/zotero/xpcom/data/search.js#L115-L116) via `Zotero.DataObjectUtilities.itemTreeMockProperties`, instead of sifting through every instance of their usage in itemTree. Now, `ItemTreeRow` is already somewhat [responsible for providing row's data](https://github.com/adomasven/zotero/blob/a44c5947cee28aad556f84d222099e73e5c04d53/chrome/content/zotero/itemTreeRow.js#L539-L547). So would it make sense to move `itemTreeMockProperties` to those classes, remove them from `Zotero.Collection` and `Zotero.Search`, and then have `CollectionViewItemTree` provide those methods/properties (same as `CollectionItemTreeRow.getField`) instead of itemTree using `.ref.{method}`?
 
This is mainly for the purpose of a cleanup. But at the same time, it may help us make itemTree more future proof because if we were to add a new type of a row, we would not have to worry about some branch of CollectionItemTreeRow potentially calling a non-existent method of that row's `ref`.  What do you think?

adomasven:
> When I worked on adding trashed collections and searches, I had to resort to a workaround to allow collections and searched to pretend to be items when they are displayed in the trash. Many parts of `itemTree` would call item's methods like `.isRegularItem`, `isAttachment`, `isAnnotation` , etc. that collections and searches do not have. I figured it would be easier to mock those properties in [Zotero.Collection](https://github.com/zotero/zotero/blob/f130e3b3e5869b8d211890290dd3af68d25efb35/chrome/content/zotero/xpcom/data/collection.js#L109-L110) and [Zotero.Search](https://github.com/zotero/zotero/blob/f130e3b3e5869b8d211890290dd3af68d25efb35/chrome/content/zotero/xpcom/data/search.js#L115-L116) via `Zotero.DataObjectUtilities.itemTreeMockProperties`, instead of sifting through every instance of their usage in itemTree. Now, `ItemTreeRow` is already somewhat [responsible for providing row's data](https://github.com/adomasven/zotero/blob/a44c5947cee28aad556f84d222099e73e5c04d53/chrome/content/zotero/itemTreeRow.js#L539-L547). So would it make sense to move `itemTreeMockProperties` to those classes, remove them from `Zotero.Collection` and `Zotero.Search`, and then have `CollectionViewItemTree` provide those methods/properties (same as `CollectionItemTreeRow.getField`) instead of itemTree using `.ref.{method}`?
> 
> This is mainly for the purpose of a cleanup. But at the same time, it may help us make itemTree more future proof because if we were to add a new type of a row, we would not have to worry about some branch of CollectionItemTreeRow potentially calling a non-existent method of that row's `ref`. What do you think?

It doesn't entirely work because some of those are also accessed in `zoteroPane` and `itemPane`. I think we can do that as a follow up PR, if you have better ideas on how to improve how those accesses occur: https://github.com/adomasven/zotero/commit/19327acac9d7a8522cad5c291ee77491ed11bc63. 

I've rebased and think we should merge this now unless @dstillman has some update release/scheduling concerns.

abaevbog:
> It doesn't entirely work because some of those are also accessed in zoteroPane and itemPane. I think we can do that as a follow up PR,

Oh, right. `zoteroPane` and `itemPane` are indeed quite tightly coupled with what is selected in itemTree. This could be a great opportunity to remove these dependencies but, I agree, better to handle that as a separate project.

Besides that, everything looks great to me!

adomasven:
> This isn't necessary just Everything — search snippets in general have been a longstanding, popular request and could be useful even for All Fields & Tags mode. I don't think we'd show them for all searches, at least by default, but definitely something we want to make sure we're able to support.

The virtualized table supports variable-height rows, so we don't have to resort to child rows for search snippets and can do variable height rows for search results.

dstillman:
I'm getting an error at startup:

```
zotero(1)(+0000001): ReferenceError: cacheKey is not defined

    _getColumns@chrome://zotero/content/itemTree.js:2501:5
    _getColumns@chrome://zotero/content/collectionViewItemTree.js:878:18
    setId/this._sortContextReadyPromise<@chrome://zotero/content/itemTree.js:993:12
```

adomasven:
Hmm, sorry! Tests were passing, and I didn't test the live version. Idk how this got past. Anyway, rebased again.

dstillman:
Test is failing

adomasven:
@dstillman ok, it took some time to iron out some test failure issues, and corner cases, but should be good to merge.

adomasven:
@dstillman let's merge this!

PR #5814, opened 2026-03-03
Citation dialog: search by citationKey in list mode

abaevbog:
https://forums.zotero.org/discussion/130090/zotero-picker-i-cant-search-item-by-citation-key

In citation dialog list mode, search by `citaitonKey` in addition to `quicksearch-titleCreatorYear` condition. In library mode, search already uses all data fields.

https://github.com/user-attachments/assets/f90781d4-2b15-4d00-bebc-43e2f5291608


dstillman:
Why not just add to titleCreatorYear? Should probably work from quick search as well.

abaevbog:
That makes sense. We already lookup by `key` in `titleCreatorYear`, so might as well add `citationKey`. I re-pushed.

We still have the extra handling in citation dialog for the in-memory filtering of selected and open items, since that is done separately from the library search.

PR #5818, opened 2026-03-04
Add Read Aloud first-run guidance panel

AbeJellinek:
And fix first-run guidance description not updating when multiple exist in the document.


dstillman:
This will show for new users, though. We need to set a default pref to true, I think?

chrome/content/zotero/xpcom/reader.js
```diff
@@ -1659,6 +1663,20 @@ class ReaderInstance {
...
+
+		// Anchor to the toolbar button in the iframe
+		let readAloudButton = this._iframeWindow.document.getElementById('read-aloud');
+		if (!readAloudButton) return;
+
+		let guidancePanel = this._window.document.createXULElement('guidance-panel');
+		guidancePanel.setAttribute('about', 'readAloud');
+		guidancePanel.setAttribute('position', 'after_end');
+		this._popupset.appendChild(guidancePanel);
+		guidancePanel.show({ forEl: readAloudButton });
```
dstillman: Doesn't really matter, but this is async.

dstillman:
Not really a problem with this PR specifically, but it's pretty easy to close this accidentally by clicking outside it. Guidance panels should probably have an X to explicitly close them…

dstillman:
Oh, I guess we already have a `noautohide` attribute. Not sure what it does instead, though.

AbeJellinek:
OK, there we go. Now we set a default pref value, disable autohide, and hide when Read Aloud is clicked. (Also still possible to click the popup itself to hide.) The `noautohide` attribute was kind of broken, fixed now.

dstillman:
Need some sort of X or "OK" or "Got it" here, though…

AbeJellinek:
Something like... this?

image

AbeJellinek:
Not my favorite design ever, but having the X above the text looked even more awkward.

dstillman:
How about this? (I have this working locally.)

Screenshot 2026-03-05 at 6 55 27 PM

AbeJellinek:
Much better.

AbeJellinek:
The "OK" is kind of funny in the context of a new feature we hope people will be excited about, but still, I think it's fine!

dstillman:
Yeah, that's why I was thinking about something like "Got it", but maybe that's too casual/hard-to-localize?

AbeJellinek:
Might sound totally bizarre in the context of a UI in languages that have stronger formal/informal register separation than English does... But then again, people can just localize it as the equivalent of "OK" if there's no good equivalent.

PR #5823, opened 2026-03-05
Batch editing with undo

### LINKED ISSUE #5652: Batch editing

tnajdek:
This PR adds:

* Batch editing support for fields, excluding the item type and creators
* The undo/redo for common actions, see below for the exact list

### Batch Editing

https://github.com/user-attachments/assets/10c67274-5038-450c-8757-de971704891f

### Undo/Redo

I've made some early design decisions, including drawing a line at what's "undoable", though I'm happy to revisit!

**Design decisions**
  - Sync that contains remote changes to items or collections clears history
  - Redo stack cleared on new change (no branching)
  - Defers to native text-editing undo for focused elements
  - All saves within a single executeTransaction() become one undo step. Batch edits are a single undo.

**Undoable:**
  - Edit item field (including item type change, creators, title-case transform, batch edit)
  - Add or remove tags (including via the since-new-item-tree-dead-code `toggleItemsListTags`)
  - Trash/Restore items, collections, and saved searches
  - Add/Remove items to/from a collection (same-library)
  - Move collection
  - Rename collection inline
  - Change parent item and convert to standalone
  - Related items (add/remove, context menu "Relate Items")
  - Merge items
  - Long-tag-fixer split via tag selector (but not at sync)
  - Normalize Attachment Titles

  **NOT undoable:**
  - Creating new objects
  - Permanent deletion
  - Adding/Removing to/from My Publications
  - File operations -- attachment renames, file syncing, PDF operations
  - Same-library drag-drop copy items (multiple chunked transactions)
  - Cross-library copy
  - Long-tag-fixer at sync
  - Sync-originated changes
  - Document recognition
  - Annotation changes
  - Search condition changes
  - Note text changes (but when a note is focused, its own Undo stack is active)

Thoughts and suggestions welcome!

Close #5652

### DIFF
--- a/chrome/content/zotero/collectionTree.jsx
+++ b/chrome/content/zotero/collectionTree.jsx
@@ -283,7 +283,7 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 		if (!treeRow.editingName) return;
 		treeRow.ref.name = treeRow.editingName;
 		delete treeRow.editingName;
-		await treeRow.ref.saveTx();
+		await treeRow.ref.saveTx({ undoAction: 'undo-action-rename-collection' });
 		window.Zotero_Tabs.rename("zotero-pane", treeRow.ref.name);
 	}
 	
@@ -1330,14 +1330,27 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 			await row.ref.eraseTx();
 		}
 		if (others.length) {
+			let collectionCount = others.filter(r => r.isCollection()).length;
+			let searchCount = others.length - collectionCount;
+			// Use the search label only when nothing but searches is selected;
+			// otherwise describe the action in terms of collections
+			let undoAction, undoActionArgs;
+			if (searchCount && !collectionCount) {
+				undoAction = 'undo-action-trash-search';
+				undoActionArgs = { count: searchCount };
+			}
+			else {
+				undoAction = 'undo-action-trash-collection';
+				undoActionArgs = { count: collectionCount + searchCount };
+			}
 			await Zotero.DB.executeTransaction(async () => {
 				for (let row of others) {
 					row.ref.deleted = true;
 					if (row.isCollection()) {
-						await row.ref.save({ deleteItems });
+						await row.ref.save({ deleteItems, undoAction, undoActionArgs });
 					}
 					else {
-						await row.ref.save();
+						await row.ref.save({ undoAction, undoActionArgs });
 					}
 				}
 			});
@@ -2260,16 +2273,21 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 		// Dropping items, collections, or searches into trash
 		if (targetTreeRow.isTrash()) {
 			let objects = [];
+			let undoAction;
 			if (dataType == 'zotero/collection') {
 				objects = await Zotero.Collections.getAsync(data);
+				undoAction = 'undo-action-trash-collection';
 			}
 			else if (dataType == 'zotero/search') {
 				objects = await Zotero.Searches.getAsync(data);
+				undoAction = 'undo-action-trash-search';
 			}
 			else if (dataType == 'zotero/item') {
 				objects = await Zotero.Items.getAsync(data);
+				undoAction = 'undo-action-trash';
 			}
 			await Zotero.DB.executeTransaction(async function () {
+				Zotero.UndoHistory.stageAction(undoAction, { count: objects.length });
 				for (let obj of objects) {
 					obj.deleted = true;
 					await obj.save();
@@ -2304,7 +2322,7 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 				await Zotero.DB.executeTransaction(async () => {
 					for (let droppedCollection of droppedCollections) {
 						droppedCollection.parentID = targetCollectionID;
-						await droppedCollection.save();
+						await droppedCollection.save({ undoAction: 'undo-action-move-collection' });
 					}
 				});
 			}
@@ -2380,6 +2398,21 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 					await Zotero.DB.executeTransaction(async function () {
 						let collection = await Zotero.Collections.getAsync(targetCollectionID);
 						await collection.addItems(ids);
+						// If moving, remove from source in the same
+						// transaction so it's a single undo step
+						if (dropEffect == 'move' && toMove.length
+								&& sourceTreeRow && sourceTreeRow.isCollection()) {
+							await sourceTreeRow.ref.removeItems(toMove);
+							toMove = [];
+							Zotero.UndoHistory.stageAction(
+								'undo-action-move-to-collection', { count: ids.length }
+							);
+						}
+						else {
+							Zotero.UndoHistory.stageAction(
+								'undo-action-add-to-collection', { count: ids.length }
+							);
+						}
 					}.bind(this));
 				}
 				else if (targetTreeRow.isPublications()) {
--- a/chrome/content/zotero/collectionViewItemTree.jsx
+++ b/chrome/content/zotero/collectionViewItemTree.jsx
@@ -1546,6 +1546,11 @@ class CollectionViewItemTree extends ItemTree {
 				// canDrop() limits this to child items
 				var rowItem = this.getRow(row).ref; // the item we are dragging over
 				await Zotero.DB.executeTransaction(async function () {
+					Zotero.UndoHistory.stageAction(
+						'undo-action-change-parent-item',
+						{ count: items.length }
+					);
+
 					for (let i = 0; i < items.length; i++) {
 						let item = items[i];
 						item.parentID = rowItem.id;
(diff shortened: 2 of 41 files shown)


dstillman:
Haven't tested, but this looks awesome. A couple quick thoughts:

1. Not sure it makes sense for saves to have to opt out via `skipUndo`. Most/all actions that are supported already pass a label via `undoAction`. Shouldn't we just get rid of `skipUndo` and have `undoAction` be the signal that a save should be undoable?
2.  I think having any sync clear the stack is a dealbreaker — auto-sync is on by default, and I don't think it makes sense to have an important feature that only works in a non-default configuration. As an easy fix, what if we clear the stack only if there are remote changes? (We could probably be even more precise and ignore remote changes in other libraries or even for objects that don't affect undo, like synced settings, but only clearing on local changes should get us most of what we need.)

tnajdek:
> Not sure it makes sense for saves to have to opt out via skipUndo. Most/all actions that are supported already pass a label via undoAction. Shouldn't we just get rid of skipUndo and have undoAction be the signal that a save should be undoable?

It's actually the other way round: the majority of undoable actions (~70%+) don't provide `undoAction` at all — they're handled by [inferring logic](https://github.com/zotero/zotero/pull/5823/changes#diff-5c1e60b5f94069ea0c1adc1c9ddb389c86d3391c991ba74c4b07809192b3f2efR272) instead. I'd argue that's preferable: when I implement a new feature that does `item.saveTx`, I shouldn't need to think about the undo action unless it's something unusual. Going the other way would mean sprinkling an explicit, repeated `undoAction` across all the call sites that currently rely on inference.

> I think having any sync clear the stack is a dealbreaker [...] As an easy fix, what if we clear the stack only if there are remote changes?

That's a good idea, I'll implement something like that!

dstillman:
> I'd argue that's preferable: when I implement a new feature that does item.saveTx, I shouldn't need to think about the undo action unless it's something unusual.

I don't think that's true. saveTx() could be called for any reason — it could be some automatic repair operation that should never be undoable, it could be some new property that we added and haven't added undo support for, it could be some complex operation across multiple data types (only one of which would be exposed to the user), or it could be a plugin doing some automated save to Extra to store some data.

The inference logic is neat, but there's just no guarantee that a given item save should actually be undoable, and there's an asymmetry to getting it wrong. If you forget to request undo, that's a missing undo step until someone notices. If you forget `skipUndo` on some post-upgrade data-cleanup operation, the undo stack shows internal operations (maybe hundreds or thousands of them) the user never did, and Cmd-Z reverts something that was never meant to be undone.

The inference logic is also pulling logic from the call sites into the undo code, which creates a somewhat dangerous, hard-to-maintain coupling.

Callers know what they're doing (moving an item to the trash, moving an item to a collection), and making them request undo guarantees both that the label is correct/comprehensive and that undoing the operation is appropriate and has been tested.

tnajdek:
Fair argument, I'll invert the logic: opt-in by providing label explicitly.

tnajdek:
I've implemented the opt-in logic which enabled more accurate undo tracking and better descriptions of the undo action. I also now have more precise list of "undoable" actions (updated in the initial message).

Next, I'm going to look into making the undo stack survive some forms of sync as discussed.

tnajdek:
Done, undo stack now survives sync if there were no remote changes to the items/collections

dstillman:
I changed it to clear the stack if there's an error on undo/redo, which seems safer. Let me know if you think that's not necessary. But I assume that should never fail, and if it does all bets are probably off?

tnajdek:
Generally, I think clearing the stack on undo/redo error is a good idea. Ideally, however, we would never hit it (without plugins/debug JS calls). Here are two cases I thought of so far:

1. Rename a collection then delete it (trash + empty trash). There is a check if the modified object exists, so undoing the rename doesn't throw: it will silently fail but beyond it stack is persisted.
2. Re-parent collection then delete the old parent (trash + empty trash) and try to undo. This time it will throw and the stack will be cleared. Even worse (I'll fix that separately) re-parented collection now has the `parentKey` set to a non-existent collection so a future rename will fail.

I feel like we might want to make the latter behave like the former?

Alternatively, few ideas:

1. Try to selectively modify the stack to remove all entries that reference a removed object
2. Consider extending undo/redo to deletions/creations

dstillman:
> Even worse (I'll fix that separately) re-parented collection now has the parentKey set to a non-existent collection so a future rename will fail.

How is that happening? Seems like that should trigger a save error if it's going through the main data methods.

tnajdek:
The broken `parentKey` is not persisted, only the in-memory version of it is broken. This happens because `_initSave` is outside of the `try`/`catch` block so it never goes through `_recoverFromSaveError`.

I'll add a fix (just for this) in a bit.

tnajdek:
I've moved the `_initSave` into the `try`/`catch` block to fix this issue.

Any thoughts on the idea of surgically removing any entries in the undo stack that reference a removed object? I'm leaning towards that approach.

dstillman:
> Any thoughts on the idea of surgically removing any entries in the undo stack that reference a removed object?

I dunno. Seems a little dicey? I'd probably err on the side of clearing the stack when in doubt. Particularly if this is mostly about things being permanently removed — I feel like it would even be totally reasonable for that to just always clear the stack.

tnajdek:
>  I feel like it would even be totally reasonable for that to just always clear the stack.

You mean for every permanent deletion? It would clear the stack in some unexpected circumstances, e.g., "Undo Retrieve Metadata" (removes the created item) or the idle automatic trashed items cleanup. Personally I'd rather have some "ghost" entries on the stack that don't do anything than loose the entire undo stack because I've left my machine idle for 5 minutes. Perhaps we should just leave them?

dstillman:
I meant for emptying (or purging specific items from) the trash. I didn't mean that we needed to — just that I don't think we need to go too far out of our way to handle conflicts from those purges.

tnajdek:
> I meant for emptying (or purging specific items from) the trash. I didn't mean that we needed to — just that I don't think we need to go too far out of our way to handle conflicts from those purges.

I've been going back and forth on this for a bit; here is what I settled on in the end:

1. On an explicit "Empty Trash": clear the undo/redo stack. The message already says that this action cannot be undone, so clearing the stack here makes sense. This addresses the most common cause of "ghost" entries on the stack.
2. Ignore other deletions. These may create "ghost" entries, but they are harmless.

Let me know if this works or if you'd like a different approach.

dstillman:
Screenshot 2026-06-11 at 11 08 07 AM


I tweaked the header a bit — added a blue background to distinguish this mode (seemed too dangerous otherwise, even with undo), added the pencil icon that @yexingsha had, and added a Done button, which isn't strictly necessary with the click-off disabling but seems more user-friendly.

dstillman:
(And I made various other tweaks before that, like removing the ellipsis from "Multiple", which looked really messy appearing many times. So making the mode clearer seemed like a good idea.)

tnajdek:
I like the blue background on the header. Batch editing UI was missing something that immediately made it obvious. At some point we had blue "Multiple" values, but this is even better!

dstillman:
@tnajdek: Trying to think through potential failure modes. See the two failing tests in the last commit.

yexingsha:
The light blue looks good to me, just need to align the icon and "Editing…" to the icon and "Info" below.

dstillman:
Alignment fixed:

Screenshot 2026-06-11 at 4 58 41 PM

tnajdek:
> Trying to think through potential failure modes. See the two failing tests in the last commit.

For the third-party writer, we could check if the value is as expected before applying undo and skip it otherwise.
The mid-sync undo is trickier. The only thing I can think of right now is to disable the undo functionality entirely until the sync completes.

I'll look into fixing both cases.

tnajdek:
Turns out that my fix for the third-party plugin edit issue also solves the potential mid-sync undo problem. We now store the expected value and verify it hasn't changed before applying the undo/redo value from the stack. In case of a mismatch, we reject the undo/redo and drop the entire stack. This directly solves the third-party writer case, and also covers the mid-sync issue: since we drop the stack at the end of any sync that includes remote changes, a mid-sync undo where a local value was already updated will do the same.

dstillman:
Added failing tests for two more potential edge cases:

1. undo/redo aren't serialized. A second undo issued mid-transaction (e.g., holding Cmd-Z) runs its `pop()` + `_entryIsStale()` before the first undo commits. The field still holds the old value, so the second entry looks stale and we clear() the whole stack. Fix: serialize undo/redo through a queue.
2. Staleness checked outside the transaction (TOCTOU). The check passes, then we yield at `await executeTransaction`. An external write committing in that window gets clobbered by the now-stale snapshot. Fix: re-check inside the transaction, after the lock.

tnajdek:
Thanks for the review! Fixed both issues and few more I've found as well.

> undo/redo aren't serialized. A second undo issued mid-transaction...

The test for this case had a slightly wrong order, I've tweaked the test that in a separate commit and confirmed it was still failing before doing the fix.

PR #5827, opened 2026-03-07
"Export Note" context menu option on annotation rows

### LINKED ISSUE #5722: Add "Export Note…" context-menu option for selected annotations in items list?
https://forums.zotero.org/discussion/129198/export-annotations

Just to save the extra step of creating a note?

abaevbog:
Display "Export Note" menu option when annotations are selected to allow one to export selected annotations without having to create a note first.

Modifier Zotero.EditorInstance.createNoteFromAnnotations to be able to convert annotations into a note without saving the note item to DB.

Fixes: #5722

### DIFF
--- a/chrome/content/zotero/fileInterface.js
+++ b/chrome/content/zotero/fileInterface.js
@@ -181,6 +181,7 @@ var Zotero_File_Interface = new function () {
 	this.exportCollection = exportCollection;
 	this.exportItemsToClipboard = exportItemsToClipboard;
 	this.exportItems = exportItems;
+	this.exportAnnotations = exportAnnotations;
 	this.bibliographyFromItems = bibliographyFromItems;
 	
 	/**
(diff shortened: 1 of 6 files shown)


dstillman:
It's showing "Export Notes…" for multiple annotations, but it should always be singular in this case.

We probably want to expose it from the parent attachment too? "Export Note from Annotations…", below "Add Note from Annotations"?

Could do the same on the parent item, since we have "Add Note from Annotations" there too. We don't currently have an "Export PDF…" below "Export Item…" there, only in the File menu, but…maybe we should? Seems like we could add this, in any case — probably a relatively common thing you'd want to do, to export a note from multiple top-level items?

abaevbog:
Right, that makes sense. If we display that context menu option on attachments and parent items, it's probably best to make it a new option independent of "Export Note" as I initially did. How does this look? I also added "Export PDF(s)" that you mentioned as another new context menu options in a separate commit.
Screenshot 2026-03-11 at 12 47 27 PM
Screenshot 2026-03-11 at 12 47 19 PM
Screenshot 2026-03-11 at 12 47 08 PM

PR #5832, opened 2026-03-09
Fix menu API icon inheritance

windingwind:
Change menu API l10nArgs from object to string to be consistent with other APIs and function calls

https://groups.google.com/g/zotero-dev/c/gEQLvmGdfXU

### DIFF
--- a/chrome/content/zotero/xpcom/pluginAPI/menuManager.js
+++ b/chrome/content/zotero/xpcom/pluginAPI/menuManager.js
@@ -81,7 +81,7 @@
 	 * @type {object}
 	 * @property {string} menuType - The type of the menu item
 	 * @property {string} [l10nID] - The l10n ID for the menu item
-	 * @property {object} [l10nArgs] - Arguments for the l10n ID
+	 * @property {string} [l10nArgs] - Arguments for the l10n ID. Support for object type is deprecated.
 	 * @property {string} [icon] - The icon for the menu item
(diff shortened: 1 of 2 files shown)


dstillman:
And why is a JSON string better here? Isn't it only a string in Fluent when it's used as an attribute, where there's no other choice? When called via functions, they pass objects:

https://searchfox.org/firefox-main/rev/7b1a35d6c70c08d2966c6c8dfc9aa9c04319f721/dom/webidl/DOMLocalization.webidl#105

Same in [setAttributes](https://projectfluent.org/dom-l10n-documentation/using-l10n-args.html). Aren't those equivalent to this?

windingwind:
> why is a JSON string better here

Several different reasons here:

1. By design pattern, we prevent passing non-primitives in function calls when possible from the beginning, which is important for cross-compartment code execution (which will be everywhere once we introduce the content principle plugin sandbox), mainly for performance reasons.
2. Since the `setL10nArgs()` is designed to be accepting string, to avoid confusion like the discussions we had in that dev list thread, I think it's better to also accept string type for the attribute `l10nArgs` as well.
3. Eventually, the object passed to either `setArgs` or `setAttribute` gets JSON stringified (https://searchfox.org/firefox-main/source/dom/l10n/DOMLocalization.cpp#192), so it doesn't make much difference. Though very minor, for simple strings, directly assembling them is more performant than calling `JSON.stringify`

PR #5834, opened 2026-03-10
feat(notes): add content width prefs for note and note tab

ramonmi:
There are some FRs about this todo, and I think we should leave the choice to the user.

https://github.com/windingwind/zotero-better-notes/issues/1520
https://github.com/windingwind/zotero-better-notes/issues/1523#issuecomment-4018975829
https://github.com/windingwind/zotero-better-notes/discussions/1517
https://github.com/windingwind/zotero-better-notes/issues/1549

### DIFF
--- a/chrome/content/zotero/preferences/preferences_general.xhtml
+++ b/chrome/content/zotero/preferences/preferences_general.xhtml
@@ -224,6 +224,22 @@
 				preference="extensions.zotero.openNoteInNewWindow"
 				native="true"
 			/>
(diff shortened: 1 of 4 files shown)


windingwind (commented):
Since we are doing this, maybe we should also expose the settings for the note window as well.

ramonmi:
Done. Changed the width control from a checkbox to dropdown menus, enabling easier addition of new view modes.

windingwind:
thanks. looks good to me.

PR #5836, opened 2026-03-11
Suggestion: Retry tests up to three times on CI

tnajdek:
The idea is that when tests pass locally but intermittently fail on CI, those failures are noise, not signal. Adding automatic retries for flaky tests would make the CI pipeline more useful.

Any thoughts on this?


dstillman:
Good idea!

dstillman:
(I mean, I think there's a slight chance that this hides some real cross-test condition by just introducing a delay, and the proper fix would be to add an explicit delay, missing await, or whatever at the right place. But we won't retry locally, and it's still more important to not have lots of spurious failures in CI.)

PR #5838, opened 2026-03-11
Citation dialog: fix popup buttons overflow for some locales

abaevbog:
With some locales (e.g. German), on Linux, the buttons do not fit into the fixed width of the item details popup. To avoid buttons overflowing, stretch the popup if buttons require more width than the popup currently has.

Reported in: https://forums.zotero.org/discussion/130208/spacing-is-off-in-citation-detail-pop-up

Before:
Screenshot 2026-03-11 at 3 54 46 PM

After:
Screenshot 2026-03-11 at 3 52 11 PM

### DIFF
--- a/scss/components/_citationDialog.scss
+++ b/scss/components/_citationDialog.scss
@@ -621,7 +621,8 @@
 		}
 		#itemDetails {
 			.popup {
-				width: 364px; // with padding and border, the panel is 400px
+				width: min-content;
+				min-width: 364px; // with padding and border, the panel is 400px
 				.details-header {
 					.icon {
 						margin-inline-end: 5px;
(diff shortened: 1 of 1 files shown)


dstillman:
Can we not do this with just CSS? We shouldn't have to put in custom JS everywhere we need to deal with variable-length text.

abaevbog:
Initially, I had some troubles with just css-based approach because we want the popup to stretch but only by the buttons - we want it to not be stretchable by the item's title and description at the top. But this should work!

Screenshot 2026-03-12 at 11 07 22 AM
Screenshot 2026-03-12 at 11 07 00 AM
Screenshot 2026-03-12 at 11 10 19 AM

PR #5839, opened 2026-03-13
Annotation type, color, author search conditions

### LINKED ISSUE #5837: Add search annotation conditions by color, type, author for advanced search
As a simpler alternative to #5324. This will make annotation search much more straightforward after #5653

abaevbog:
- added annotationType, annotationColor, and annotationAuthor search conditions
- updated logic in search.js to only include annotation on negative operators. Otherwise, 'annotationTitle' 'isNot' 'X' would always match every attachment and top-level item in the library

Fixes: #5837

https://github.com/user-attachments/assets/acbfcecf-ef83-45bb-9c63-3896532cfc7d

### DIFF
--- a/chrome/content/zotero/elements/zoteroSearch.js
+++ b/chrome/content/zotero/elements/zoteroSearch.js
@@ -374,7 +374,7 @@
 			return false;
 		}
 
-		onConditionSelected(conditionName, reload) {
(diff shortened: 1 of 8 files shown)


dstillman:
Ported to new search architecture in https://github.com/zotero/zotero/commit/2717aafedd469f00d007a689ce650cec51681bf2

PR #5847, opened 2026-03-17
Add "Remove from Recently Read"

dstillman:
- Show "Remove from Recently Read" in item context menu when in Recently Read collection (reuses "Remove Item from Collection…" menu slot)
- Non-force delete clears `lastRead` on all child attachments (with prompt), force delete moves to trash (matching collection behavior)
- Re-run search on item modify in Recently Read view so items disappear when lastRead is cleared (also handles sync changes)
- Change `attachmentLastRead` setter to accept null for clearing
- Extract "Recently Read" as shared FTL term

### DIFF
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -2003,11 +2003,22 @@ var ItemTree = class ItemTree extends LibraryTree {
 			else if (collectionTreeRow.isRecentlyRead() && !force) {
 				await Zotero.DB.executeTransaction(async () => {
 					for (let item of selectedItems) {
-						let topLevelItem = item.topLevelItem;
-						let attachments = topLevelItem.isAttachment()
-							? [topLevelItem]
-							: Zotero.Items.get(topLevelItem.getAttachments(false))
-								.filter(a => a.attachmentLastRead);
+						let attachments;
+						// Child attachment -- clear only this one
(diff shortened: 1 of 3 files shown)


AbeJellinek:
- Ellipsis, since it opens a confirmation dialog like Remove from Collection...?
- The dialog seems to be using collection strings - "Are you sure you want to remove the selected item from this collection?" Not crucial but kind of confusing.
- If I right-click on a child attachment/note and select Remove from Recently Read (or press delete), it shows the dialog, but nothing changes unless the selected child is the attachment that was read.

dstillman:
Yeah, I wasn't sure if it was worth adding a new prompt. It's sort of a collection! Can't remember what we do for My Publications…

AbeJellinek:
It sort of is, just surprised me when I saw it.

AbeJellinek:
I guess you could argue that there's some kind of use case for clearing last read on a single child attachment... but probably not much of one.

dstillman:
Oh, no, you're right — I didn't mean to do that. It should only remove the one you selected. I mean, I don't know why you'd want to? But if we're exposing this as an option on the child attachments at all, they should operate on the one you selected?

dstillman:
And the menu option should be disabled for attachments that don't actually have lastRead.

AbeJellinek:
Looks good to me.

PR #5848, opened 2026-03-17
Fix search condition localization crash for non-en-US locales

tnajdek:
A minimum solution would be a try/catch block but I also moved all the `searchConditions` strings to .ftl.

Screenshot 2026-03-18 at 00 05 44

Regression in 34039991f


dstillman:
It's not technically a regression — new .properties/.dtd strings are always broken for non-English locales until we merge strings from Transifex. But that's partly why we don't add new .properties/.dtd strings. I think #2854 just predated that policy. I already moved one of the other strings in https://github.com/zotero/zotero/pull/5847. I may have left this one in because fixing it was kind of annoying, and we try to avoid forcing relocalizations, but hopefully Transifex's Translation Memory will transfer these automatically, so I guess we can move them.

PR #5862, opened 2026-03-25
Per-group file renaming settings + Store `isAdmin`

tnajdek:
This PR is a follow-up to #5774 but makes the dialog read-only for non-admins. The annoying thing is that a sync must be triggered to populate that flag; otherwise, everyone appears as a non-admin.

1. Store `isAdmin` per library
2. Add a migration step to add the `isAdmin` field to the libraries table. Reset all group versions to 0 to force re-fetching so that the field gets populated on the next sync.
3. Tweak the file renaming settings dialog: for non-admins, make all fields read-only. Non-admins can still run file renaming using the rules configured by admins.

We will also need a tweak for this server-side.

### DIFF
--- a/chrome/content/zotero/fileRenamingDialog.js
+++ b/chrome/content/zotero/fileRenamingDialog.js
@@ -81,16 +81,15 @@ var FileRenamingDialog = { // eslint-disable-line no-unused-vars
 		let autoRenameEnabled = this.settingsEl.autoRenameEnabled;
 		let library = Zotero.Libraries.get(this._currentLibraryID);
 		let isAdmin = this.isUserLibrary || library.isAdmin;
-		this.renameFilesBtn.hidden = !autoRenameEnabled || !isAdmin;
-		this.renameFilesBtn.disabled = this._isRenameFilesBtnDisabled();
+		this.renameFilesBtn.hidden = !isAdmin;
+		this.renameFilesBtn.disabled = !autoRenameEnabled || this._isRenameFilesBtnDisabled();
 	},
 
 	_isRenameFilesBtnDisabled: function () {
-		let libraryID = this._currentLibraryID;
-		if (libraryID === Zotero.Libraries.userLibraryID) {
+		if (this.isUserLibrary) {
 			return Zotero.Prefs.get('autoRenameFiles.done');
 		}
-		return !Zotero.Attachments.isAutoRenameFilesEnabledForLibrary(libraryID);
+		return false;
 	},
 
 	_openRenameFilesPreview: function (libraryID) {


dstillman:
Having only "Rename Files…" as a button doesn't work. People need to be able to close the window without renaming files, and clicking on the window controls isn't enough. We need a "Done", even if that then prompts.

> When switching to another library, "Rename Files..." cancels the switch and "Keep Existing Filenames" continues with it

"Preview Changes…", you mean? Seems weird that either would actually cancel the switch — shouldn't we do whatever we need to do and then switch after?

dstillman:
(If you cancel the rename after previewing, that should probably cancel the library switch, though.)

dstillman:
- The library switcher should probably be bigger — something closer to what we have in the Sync → Reset pane, if not bigger. (Not sure why that one is wrapping to the next line.)
- More space between the checkboxes and the text, less space between the text and the textbox.
- Why can't we know in group libraries whether files are in sync? Because they could have been changed on another computer?

tnajdek:
Changes done:
- Add "Done" button and right-align both buttons with gap
- Remove console.log
- Fix extra quotes in xhtml stylesheet references
- Style library picker larger with larged, bold label
- Adjust spacing between sections

> Seems weird that either would actually cancel the switch — shouldn't we do whatever we need to do and then switch after?

Rename files preview is a separate, non-modal window. Previously it was used in a different context but no longer is.
I think we can now merge it into the dialog and improve the flow as suggested.

tnajdek:
Much nicer flow after integrating the preview into the file renaming settings dialog. It also now automatically closes the preview once changes are applied, which is a nicer experience, especially if you're coming from a prompt.

I experimented with skipping the prompt entirely and going straight into the preview, but it didn't feel right (like the app is doing something you didn't want), so I scrapped that approach.

Here is how the new flow looks:

https://github.com/user-attachments/assets/8d2269dc-53a2-4902-96d1-92dc820b8a5d

@dstillman Any thoughts or suggestions?

dstillman:
Not necessarily new, but is it supposed to be prompting me even if I didn't change anything? I never did a batch-update, and if I open the the settings window and either 1) immediately click Done or 2) switch to another library, it prompts me to rename all file, which feels a bit pushy.

I guess "Preview Changes…" is the indicator that there are unrenamed files? I feel like that should just be "Rename Files…" — the ellipsis is what tells you there might be a preview — and if I haven't actually made changes to a library's settings this session, it shouldn't bother me with a prompt. If I do make changes and then try to close or switch away, it could maybe still prompt.

tnajdek:
> Not necessarily new, but is it supposed to be prompting me even if I didn't change anything? I never did a batch-update, and if I open the the settings window and either 1) immediately click Done or 2) switch to another library, it prompts me to rename all file, which feels a bit pushy.

For "My Library" it uses `autoRenameFiles.done` so that must be `false`? For group libraries it will only prompt if you make an actual change.

> I guess "Preview Changes…" is the indicator that there are unrenamed files? 

I've switched from "Rename Files..." to "Preview Changes…", it seemed better for me, but we can revert back.

dstillman:
> For "My Library" it uses `autoRenameFiles.done` so that must be `false`?

Yes. But I think that should just control whether "Rename Files…" is grayed out. We should only prompt if you just made changes.

> I've switched from "Rename Files..." to "Preview Changes…", it seemed better for me, but we can revert back.

Yeah, I think it's clearer for that to be the ultimate action you're trying to accomplish. The preview is just a version of a confirmation dialog. But we could still say "Preview Changes…" in the prompt, since that already asks you if you want to rename files, so it makes sense for it to tell you what the next step is.

I'm not sure about reusing the same window for the preview. It seems pretty bizarre and non-standard for a dialog to change dynamically in this way. Shouldn't it just be a modal sheet within the dialog?

tnajdek:
- Display the file preview in a separate modal dialog
- Only prompt the user to rename files if they actually made changes to the library's settings in the current session
- Changed label "Preview Changes…" → "Rename Files…" in the dialog (left it as "Preview Changes…" in the prompt)
- Tweaked intro label in the Rename Files dialog to explain which library is being affected, since we no longer display the library picker above

@dstillman Any further suggestions, please let me know!

dstillman:
(Rebased to avoid Firefox rebuilds after the origin/main version bump)

dstillman:
I'm hiding "Rename Files…" for non-admins. They may still have write access to the group, but the rest of the pane is disabled, and I think it's reasonable to disallow mass changes for non-admins.

PR #5863, opened 2026-03-26
Citation dialog: skip irrelevant child rows in itemTree

abaevbog:
Pass `filterChildItems` to `itemTree` from `citationDialog` to filter out unwanted child rows before adding them in `toggleOpenState`.

In annotations mode, do not add child notes and attachments without annotations
In add-note mode, only keep child notes that have content

https://github.com/user-attachments/assets/52b93366-9225-4c69-968c-8bba7622e3b8

---

Initially, I tried to overload the `toggleOpenState` in citationDialog to keep all logic there and avoid conflicts with the big `itemTree` refactor. But that lead to issues during column sorting, since `toggleOpenState` is used in very many places. So this is safer and much easier.

@adomasven, is adding this going to cause issues for the itemTree refactor?

### DIFF
--- a/chrome/content/zotero/integration/citationDialog.js
+++ b/chrome/content/zotero/integration/citationDialog.js
@@ -697,6 +697,16 @@ class LibraryLayout extends Layout {
 			},
 			emptyMessage: Zotero.getString('pane.items.loading'),
 			columns: itemColumns,
+			// Skip irrelevant child rows (e.g. no notes in annotations mode)
+			filterChildItems: (item) => {
+				if (DIALOG_STATE.isAddingAnnotations()) {
+					return SearchHandler.isItemWithAnnotations(item);
+				}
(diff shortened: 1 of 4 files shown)


adomasven:
> is adding this going to cause issues for the itemTree refactor?

In the refactor `ItemTreeRows` and their subclasses can specify what children they return, so you wouldn't do it like that there, and implement this somewhat differently probably.

Porting this as a change into the refactor is easy, but we probably won't want to lock into this solution there.

abaevbog:
> In the refactor ItemTreeRows and their subclasses can specify what children they return

Great!

> so you wouldn't do it like that there, and implement this somewhat differently probably.

Yeah, this is just a temp solution if we are not waiting for the itemTree to be merged before next release. When that is merged, we can re-implement this properly.

PR #5865, opened 2026-03-27
Citation dialog: variable min-width for layouts + smooth window resizing

abaevbog:
- Allow different min-width for list vs library layouts. In library mode, we need at least 1000px to have a functional itemTree, especially when sidepane is present. But for list mode, 1000px is very wide, so min-width should be 800px.
- When the dialog closes, store the width/height of the current dialog's layout. Then, when the dialog opens next time in that layout, set window's width/height based on last saved params. That way, one can open a citation dialog in list mode at 800px, then open annotations dialog (always library mode) at 1000px, and then have the citation dialog in list mode open again at its intended 800px.
- If one switches from library to list mode, add a smooth window resizing animation to shrink from current width to 800px. If one switches from list to library mode, window smoothly resizes to 1000px min-width. This is not applied to linux where win resizing behavior is a bit jumpy.

https://github.com/user-attachments/assets/f1af2a7d-6d25-45bf-bbff-59b2a2975c19

### DIFF
--- a/chrome/content/zotero/integration/citationDialog.xhtml
+++ b/chrome/content/zotero/integration/citationDialog.xhtml
@@ -113,7 +113,7 @@
 						<span class="icon icon-css icon-item-type" data-item-type="note"></span>
 					</div>
 					<div id="dialog-type-annotations" class="option keyboard-clickable" value="annotations" role="radio" aria-checked="false" tabindex="-1" data-arrow-nav-enabled="true" data-l10n-id="integration-citationDialog-btn-type-annotations">
-						<span class="icon icon-css icon-attachment-annotations"></span>
+						<span class="icon icon-css"></span>
 					</div>
 				</div>
 				<div class="hbox" style="flex:1"></div>
--- a/chrome/skin/default/zotero/16/dark/annotations.svg
+++ b/chrome/skin/default/zotero/16/dark/annotations.svg
@@ -1,0 +1,12 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+<g clip-path="url(#clip0_603_184)">
+<path opacity="0.6" d="M16 13H13V16H7V15H12V4H1V9H0V3H3V0H16V13ZM4 3H13V12H15V1H4V3Z" fill="white"/>
+<path opacity="0.4" d="M6.5 9.5H0.5L6.5 15.5V9.5Z" fill="#A28AE5"/>
+<path fill-rule="evenodd" clip-rule="evenodd" d="M7 16H6.29297L0 9.70703V9H7V16ZM6 14.293V10H1.70703L6 14.293Z" fill="#A28AE5"/>
+</g>
+<defs>
+<clipPath id="clip0_603_184">
+<rect width="16" height="16" fill="white"/>
+</clipPath>
+</defs>
+</svg>
--- a/chrome/skin/default/zotero/16/light/annotations.svg
+++ b/chrome/skin/default/zotero/16/light/annotations.svg
@@ -1,0 +1,13 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+<g clip-path="url(#clip0_16597_153884)">
+<path d="M15 1V12H12V15H6.70703L1 9.29297V4H4V1H15Z" fill="white"/>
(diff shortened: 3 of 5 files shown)


abaevbog:
@dstillman, this is a followup to our conversation about the min-width for the dialog depending on the layout. Is this the kind of transition that you had in mind?

dstillman:
I was imagining it would shrink from both sides to a smaller, centered popup, but not sure if we're able to do that. I was also assuming we'd do diagonal resizing rather than one axis and then the other. Not sure how either would look in practice, though.

@yexingsha, thoughts on this?

abaevbog:
> I was imagining it would shrink from both sides to a smaller, centered popup, but not sure if we're able to do that.

The most functional would be to shrink the dialog from the left side. Then the user's mouse would often remain over the toggle that switches the layout type. But `window.resizeTo` resizes the window just from the right. So to do it any other way, we do many iterations of resizing the window a tiny bit, and then moving it a tiny bit to preserve the positioning. Moving + resizing the window like that does not flow particularly smoothly, which is why I abandoned that approach. I am afraid we would run into the same issue when trying to resize from "both sides".

https://github.com/user-attachments/assets/e055d210-858e-45d6-9b39-edc1d8ef60be

> I was also assuming we'd do diagonal resizing rather than one axis and then the other.

We can try diagonal resizing to see if that feels better, for sure!

dstillman:
I don't think the button positioning matters — with no results, you'd already be switching from a full library browser window to just a thin input bar. 

But if we can't do this smoothly, maybe it doesn't work.

abaevbog:
That makes sense. This is how resizing + shifting of the window may work to make it seem like we are resizing from both ends. I think it works fairly well! On macOS, it's quite smooth. On windows, I see is a bit of jumpiness sometimes - but that could be my VM.

MacOS:

https://github.com/user-attachments/assets/7a8a6116-2d98-45e9-98a2-b304a3f0d30a

Win:

https://github.com/user-attachments/assets/c0b60056-5bb5-4d8c-bf62-6cea732ca70b

yexingsha:
Resizing from center + diagonally looks pretty good to me, though I think we need to increase the min height of library mode to ~600px so that it doesn't look too weird with the increased min width.

Also, now that I'm looking at this more closely, I see we need to use a different icon for the Add Annotations switch:
annotation

And let's try using a checkmark instead of the right arrow to make the accept button clearer in both modes, to address https://forums.zotero.org/discussion/129622/suggested-change-to-the-new-citation-dialog-box: 
checkmark

abaevbog:
How does this look? I bumped the min-height for the main table from 200 to 400px, which gives us the min height of ~570px. And plugged in the new icons too. I quite like the checkmark instead of an arrow - but there must have been some reason why we did not do a checkmark to begin with ... It may have been because that's what the accept button looked like in quickFormat dialog before the redesign.

One question: we need two icons for the annotations icon here - one for dark mode and one for light mode as they are a bit different. I got the dark mode icon from figma - but it had a white border that I tweaked to look more light-gray-ish. Let me know if I should use a different svg for the dark icon:

Screenshot 2026-03-27 at 2 55 20 PM

yexingsha:
The dark mode icon has the correct opacity, looks good!

Also, I finally realized that part of the reason the accept/cancel buttons look too subtle is because the 20px icons inside are shrunk down to ~17px. Looks like this is caused by `.btn-icon {background-size: 60% !important;}`, which in turn might be related to the settings button using a 16px icon from a previous design instead of the 20px icon:
options

abaevbog:
Good catch! It was meant to be `background-size: auto`. It does increase the sizing of the icons by a tiny bit.

Before:
Screenshot 2026-03-29 at 9 48 47 PM

After:

Screenshot 2026-03-29 at 9 48 16 PM

dstillman:
I'm seeing a weird stutter in the middle:

https://github.com/user-attachments/assets/de05ca5e-e5aa-40b1-88e6-0f2a67e20599

abaevbog:
@dstillman, this should fix it. It likely happened when search runs when window resizing is not done yet. Now we'll run window resizing first and then trigger the search.

Just to be sure, is the library you were testing on somewhat large?

dstillman:
Super smooth now. Can you squash this into a few relevant commits?

dstillman:
(Yes, 10K-item library.)

abaevbog:
Squashed into two commits: the resizing logic and the updated icons. In the process, I ran into another edge case when the window sizing would be jumpy on the initial load of the list mode (due to long loading in hidden trees library mode), so I tweaked the timing of library mode loading as well.

PR #5866, opened 2026-03-28
fix: respect explicit open-in-tab/window prefs for note and attachments

ramonmi:
- Introduce `_getOpenInNewWindow` to distinguish between note and reader preferences
- Add `forceOpenInWindow` logic to locateMenu and zoteroPane
- Ensure explicit menu selections override default preferences and alternate behavior toggles
- Fix visibility logic for context menu options based on current locate mode

fixes:
1. https://forums.zotero.org/discussion/128815/shift-enter-is-not-opening-notes-in-tabs
2. The 'Open in new tab' option incorrectly appears in the tab locate menu when the `openReaderInNewWindow` and/or `openNoteInNewWindow` prefs are enabled. **This logic seems flawed. If the item is currently open in a tab, the menu should always display the Open in new window option.** While PR #5528 resolved the issue where the ['Open Note in New Tab' menu item was non-functional](https://github.com/zotero/zotero/pull/5528#issuecomment-3280611174), incorrect behavior persists when the `openReaderInNewWindow` or `openNoteInNewWindow` prefs are enabled.
3. With `openReaderInNewWindow` enabled and `openNoteInNewWindow` disabled, selecting both a note and a PDF and choosing 'Open in Tab' from the ZoteroPane locate menu incorrectly opens the note in a new window.

Maybe @windingwind  could take a look at this when you have time?

### DIFF
--- a/chrome/content/zotero/locateMenu.js
+++ b/chrome/content/zotero/locateMenu.js
@@ -377,6 +377,11 @@ var Zotero_LocateMenu = new function () {
 	function ViewItem(alternateWindowBehavior) {
 		this._viewItemType = "mixed";
 		this._numItems = 0;
+		let _getOpenInNewWindow = function (viewItemType) {
+			return Zotero.Prefs.get(
+				viewItemType === "note" ? "openNoteInNewWindow" : "openReaderInNewWindow"
+			);
+		};
 		Object.defineProperty(this, "className", {
 			get() {
 				switch (this._viewItemType) {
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/locateMenu.js
```diff
@@ -473,10 +482,15 @@ var Zotero_LocateMenu = new function () {
 				if (usableItem) usableItems.push(usableItem);
 			}
 			
+			let forceOpenInWindow;
+			let openIn = this.l10nArgs.openIn;
```
windingwind: This doesn't feel right to me. Using l10nArgs in actual logic violates its designed purpose.

windingwind (changes requested):

chrome/content/zotero/locateMenu.js
ramonmi: Resolved

windingwind:
thanks, looks good to me

windingwind (approved):

PR #5868, opened 2026-03-29
Fix reader cleanup on tab close and event listener removal

hello-wy:
## Summary
- fix reader cleanup path when tab is closed via tab notifier
- make customEvent handler removable and clean it in uninit
- ensure ReaderTab.uninit removes window-level pointer/load listeners
- fix unregisterEventListener filter logic so handlers are actually removed


mrtcode:
Looks good. Thanks.

PR #5870, opened 2026-04-01
Citation dialog: fix CI test breakage

abaevbog:
- prevent the focus handler from resetting the cached items in search handler
- wait for search to finish after changing dialog type in "should not display empty note child rows" test


abaevbog:
@dstillman, these couple of tweaks fix the CI test failure race condition. How does it look?

PR #5871, opened 2026-04-02
QuickCopy: CSL formatting of citations in notes

abaevbog:
On quickCopy (via keyboard shortcut or drag-drop), if all items are notes (including mock notes that
annotations are wrapped into), go through all citations, manually format them in the currently selected CSL style and return a new unsaved mock note with those citations replacing original default ones.

Added note.export.useCSLCitation pref to enable this behavior. If disabled, default to the original simple format of citations. 
Brought up in https://forums.zotero.org/discussion/130689/quick-copy-inconsistency-drag-with-shift-doesn-t-match-ctrl-shift-a-zotero-8

https://github.com/user-attachments/assets/972e2c73-faf5-43c4-8cf5-79c02562dfca


dstillman:
Does this apply to the reader too?

Why a pref? I don't think there's any reason we'd keep the simple format. Using the Quick Copy citation format was just never implemented.

dstillman:
Oh, but what if you've set Quick Copy to an export format?

I've actually been wondering if we should separate bibliographic vs. export formats for Quick Copy. Add another shortcut key for export, with a pref to determine which one a drag triggers? And then you'd always have a citation format configured for this? Being able to have shortcuts for both is a [common request](https://forums.zotero.org/discussion/131245/export-biblatex-directly-to-clipboard). Arguments against: 1) the vast majority of the time it's probably for BibTeX, in which case you should be using BBT anyway, and 2) you can also do a bib format via the item pane header in Z7+. But not sure how we'd handle this if you can set Quick Copy to an export format.

abaevbog:
> Why a pref? I don't think there's any reason we'd keep the simple format.

Just for backwards compatibility - if that's not something we're worried about here, we can totally remove the pref for the old default format.

> Oh, but what if you've set Quick Copy to an export format?

That's a good question. Export formats like BibTeX or CSV make sense for regular items but they make less sense for notes and annotations. Currently, if you try to copy an annotation and BibTeX is the specified QuickCopy format, we fallback to the default citation format. This is where the simple format still comes in handy - but I can already imagine it may cause some confusion.

 > I've actually been wondering if we should separate bibliographic vs. export formats for Quick Copy. Add another shortcut key for export, with a pref to determine which one a drag triggers? And then you'd always have a citation format configured for this? 

That would make sense! It would solve the issue above by always giving us a citation format to use for annotations and notes. To determine which format (bib vs export) to use on drag for regular items, we would need to have a new pref - would it be necessary to add an extra shortcut key in that case? We could decide what format to use on Cmd-Shift-C based on that pref too. If we added a new shortcut for export format, we would be somewhat loosing the parity between QuickCopy via keyboard and QuickCopy on drag-drop. But if that's not a problem, Cmd-Shift-E may be a candidate for export shortcut.

Worth noting that the bib vs export format pref would still end up being ignored for notes and annotations.

> 1) the vast majority of the time it's probably for BibTeX, in which case you should be using BBT anyway, and 2) you can also do a bib format via the item pane header in Z7+.

Per the linked post, the main issue is not that you can't export items now - you can already do it via context menu, it just takes a few extra clicks. We do have all the toolset already in place, so it's just about making it more accessible. If we can save users a couple of clicks, why not?

abaevbog:
While I am at it, the behavior of site-specific quick copy settings seems strange to me.

We allow to override the selected quick-copy format for specific domains based on `Zotero.QuickCopy.lastActiveURL`. If I add a special case for zotero.org, and have a forums page open in the browser, all quick copy actions for all items in the library end up using the specified APA style. But then if I open another page in the browser, the next next quick copy will use the selected chicago style. To me that feels a bit unpredictable, especially because it applies to all items. At first, I didn't actually realize what was going on. 

May I ask what is the rationale behind this `Zotero.QuickCopy.lastActiveURL` logic, as opposed to checking the domain against the `URL` field from the item to decide what format to use?

Screenshot 2026-05-01 at 11 40 10 AM

abaevbog:
> May I ask what is the rationale behind this Zotero.QuickCopy.lastActiveURL logic, as opposed to checking the domain against the URL field from the item to decide what format to use?

Is it because there can be many items selected?

dstillman:
Yes, I think we'd add Cmd-Shift-E for export. If Quick Copy was already set to an export format, move that to the new export pref, and set bib back to the default. New pref for whether dragging should use bib or export, default to bib.

The site-specific settings are for picking a format for the site you're currently browsing — e.g., Wikipedia Citation Templates while you're on Wikipedia. Nothing to do with the items you're exporting.

dstillman:
I'm not sure what we'd do for site-specific settings if we had two different formats, though. What would that affect? Unless it let you override one or both for a given site? Not sure what the alternative would be.

> Per the linked post, the main issue is not that you can't export items now - you can already do it via context menu, it just takes a few extra clicks. We do have all the toolset already in place, so it's just about making it more accessible. If we can save users a couple of clicks, why not?

I suspect they just didn't realize you could use export for Quick Copy — otherwise they would've asked for a separate shortcut or to switch between them more easily, not talked about the menu-based option.

abaevbog:
> Yes, I think we'd add Cmd-Shift-E for export. 

Sounds good!

> The site-specific settings are for picking a format for the site you're currently browsing — e.g., Wikipedia Citation Templates while you're on Wikipedia.

Ah, yes, sorry, that's clear!

> I'm not sure what we'd do for site-specific settings if we had two different formats, though. What would that affect? Unless it let you override one or both for a given site? 

I assume you use site-specific settings for very specific use cases (e.g. wikipedia example), so I think whatever choice you make there should override all other prefs you have selected? So if you select Wikipedia Citation Template (export format), that's what we should always use regardless of any other configuration when you're on wikipedia.org. 

We could of course allow one to select both citation and export formats there as well if there are situations when we need more flexibility.

dstillman:
> So if you select Wikipedia Citation Template (export format), that's what we should always use regardless of any other configuration when you're on wikipedia.org.

But we're talking about there being two separate format settings. So then what does it override? If you choose Wikipedia Citation Templates on Wikipedia, does it only override your export setting, and leave your bib setting alone? And it overrides the drag setting to force use of export even if you have it set to use bib for drag?

abaevbog:
> If you choose Wikipedia Citation Templates on Wikipedia, does it only override your export setting, and leave your bib setting alone

In this case, Cmd-Shift-E should definitely use Wikipedia Citation Templates disregarding the Export format pref. On drag, I think it would also make sense to use Wikipedia Citation Templates as well, disregarding the Format vs Bibliography format pref. I think explicit site-specific pref warrants it.

The only thing I'm not sure about is Cmd-Shift-C... It would be odd if in this case we were to use Wikipedia Citation Templates because the user would press "Copy Bibliography" and end up with a copied export format. So I would assume we don't do that and have Cmd-Shift-C copy the bibliography?

It does feel like a confusing setup for the user to have a site-specific setting that overrides one of 2 settings conditionally depending on what value is selected. That might be the biggest drawback of a dedicated Cmd-Shift-E shortcut - whatever we do with it feels a bit odd.

dstillman:
But I think that's an argument for having site-specific settings be able to override all three settings — bib, export, and drag — with the possibility of choosing "Default" for any of them. Then there's no real awkwardness, right?

abaevbog:
I thought it might be too much to put all of this config into the dialog for each site - but now I agree that's the only way to avoid such awkwardness. And it's explicit what the behavior is too. So yeah, let's try that!

abaevbog:
Yeah, I think after this refactoring it's clearer how the logic flows. You can specify which format should be used on drag, and Cmd+Shift+C/E allows you to do both easily via keyboard. Site-specific settings largely duplicate the settings in the main prefs panel, so you can override any part of the global quick copy prefs. 

I refactored the quickCopy prefs logic a bit to try to use the same format of stringified json for global prefs (`{ id, contentType, locale, mode}` for bib and `{id, mode}` for export), and for site specific settings. Older site-specific settings (with either export or bib format, not both) remain in DB as they are - but we just handle them when parsing JSON to ensure they fit the current workflow. How does this look?

Screenshot 2026-05-05 at 12 06 54 AM
Screenshot 2026-05-05 at 12 07 00 AM
Screenshot 2026-05-05 at 12 06 20 AM

dstillman:
While we're at it, we should probably make some changes to the citation/bib shortcuts, and also change how Copy Citation/Bibliography work in the reader:

https://forums.zotero.org/discussion/comment/512563/#Comment_512563

I was originally going to say Copy Bibliography should be Cmd-C and Copy Citation should be Cmd-Shift-C as the modified action, but 1) it probably makes more sense for the lighter-weight action to be a shorter shortcut (even if historically it's the modified one), 2) copying an annotation with Cmd-C is arguably closer to copying an citation (since it actually includes citations), 3) Cmd-Shift-C could remain what it's been for 20 years, and 4) Cmd-Shift-C would stay consistent with Cmd-Shift-E for export.

abaevbog:
Right, I remember finding myself a bit confused by all the different ways to copy items in the past as well. I am hoping that this refactoring makes it a bit more straightforward.

Now, the main way to copy an item (regular, note, annotation) is Cmd/Ctrl+C (or Edit > Copy menu). This is the new behavior:

In library tab:
- when a regular item is selected, copy the citation. Old Cmd-Shift-A shortcut for that is removed. 
- when a note or annotation is selected, copy them via quickCopy in note format

In reader tab:
- when nothing is selected, copy the citation of the top-level item (if it exists)
- when an annotation is selected, copy the annotation as before
- when text is selected, copy the text as before

Depending on what is selected, the label of Edit > Copy is updated (e.g. "Copy citation" or "Copy note" or "Copy annotations").

To streamline quick copy logic, I made a single entrypoint `Zotero.QuickCopy.getContentFromItems` that handles all transformation of items and their translation. So now, all components using quick copy (reader, zoteroPane, drag utils) just call it with the right items to receive the text and html.

A few notes:
- there are two migration steps now, one for each distinct update here. 22 -> 23 to separate `export.quickCopy.setting` into two different prefs. 23->24 to remove Cmd+Shift+A and rename `copySelectedItemsToClipboard` and `copySelectedItemsToClipboardAsExport` to `keys.copyAsBibliography` and `keys.copyAsExport` for clarify. If we want to combine it all into a single commit, I'll combine those pref migrations steps into one.
- now, copy of annotations from the reader is handled not in reader's `onSetDataTransferAnnotations` but in `ZoteroPane.copySelectedItemsToClipboard`. Cmd+C is handled by `ZoteroStandalone.handleCopyMenuCommand`, which runs native `goDoCommand('cmd_copy')` when text is selected in the reader, or runs `ZoteroPane.copySelectedItemsToClipboard` in all other cases  to keep that flow simple. It means `zotero/annotations` is not set on the `copy` event data, so pasting a copied annotation into the note editor will not render it's citation. https://github.com/zotero/note-editor/pull/66 from earlier addresses it, and also allows us to properly render pasted annotations copied from itemTree as well.

New Edit > Copy behavior:

https://github.com/user-attachments/assets/1ec799c0-2d01-4078-81b8-b0dc98870019

PR #5876, opened 2026-04-07
Add File menus to note tab

ramonmi:
Show in library and Close

Current:
image


dstillman:
Similar to https://github.com/zotero/zotero/pull/5877#issuecomment-4199793661 — should use a single menu item that calls a ZoteroPane function that does the right thing based on the tab type.

ramonmi:
Resolved by f974700.

dstillman:
Thanks!

dstillman:
https://forums.zotero.org/discussion/131256/missing-file-menu-content-from-a-note-in-a-tab

PR #5877, opened 2026-04-07
Fix broken Find option under Edit menu in reader

ramonmi:


dstillman:
Merging as a quick fix, but instead of duplicating the menuitem, probably better to leave this targeting `cmd_find` and have `cmd_find` call a ZoteroPane.js function that runs the appropriate find based on the tab type.

dstillman:
(Because this is also broken in note tabs)

ramonmi:
> (Because this is also broken in note tabs)

Resolved separately in #5881

PR #5878, opened 2026-04-07
Set Recently Read threshold relative to most recent item

AbeJellinek:
Starting two weeks before the last-read item was read.

And:
- Allow overriding `idColumn` in `idsToTempTable()`, since the `tempTable` search condition (confusingly!) wants `id`, not `itemID` like `idsToTempTable()` produces by default.
- Add `libraryID` option to test support import functions
- Add tests


AbeJellinek:
Tests were passing locally! I'll investigate.

PR #5881, opened 2026-04-08
add Find option to Edit menu in note tab and window

ramonmi:
unify Find option as a ZoteroPane.js function that runs the appropriate find based on the tab type

requires Zotero/note-editor#75


windingwind:
looks good to me

Issue #5322, opened 2025-05-31
Missing border for tag circles on a highlighted item


tnajdek:
Looks like this now:
![Image](https://github.com/user-attachments/assets/3fce1a96-892f-4c2b-a592-0ebeb18706b0)

Should look like this:
![Image](https://github.com/user-attachments/assets/64eb5ba0-2ed8-48b1-ba38-08daa829351c)

tnajdek:
There are actually two regressions, first one is the missing border, second is the transparency of the colors (probably due to FF version upgrade which now supports transparency here?)

<img width="405" alt="Image" src="https://github.com/user-attachments/assets/cc22d34d-44e7-4d8b-87ab-5271ae578d02" />

PR #5883, opened 2026-04-10
fix tag colors order for focused row

ramonmi:
fixes https://forums.zotero.org/discussion/comment/510193/#Comment_510193

Before: image
After: image

### DIFF
--- a/scss/components/_item-tree.scss
+++ b/scss/components/_item-tree.scss
@@ -104,6 +104,16 @@
 					transform: scaleX(-1);
 					flex-direction: row-reverse;
 				}
+				
+				// Fix the stacking order reversing issue when selected
+				.tag-swatch {
+					position: relative;
+					@for $i from 1 through 9 {
+						&:nth-child(#{$i}) {
+							z-index: 10 - $i;
+						}
+					}
+				}
 			}
 
 			.icon-item-type + .tag-swatch,


dstillman:
What's making it wrong when selected?

ramonmi:
> What's making it wrong when selected?

Regarding the previous change: https://github.com/zotero/zotero/issues/5322#issuecomment-2980321662

Just wondering if this is the intended design? @tnajdek

tnajdek:
I don't believe it was. More likely, I wanted to fix the issues that arose with the upgrade of the Gecko engine.
Also, looking at the diff, I don't think that change had anything to do with the order.

The real reason why these are rendered in the incorrect order is the actual icon being used:
For unselected/unfocused items we have circles and crescents; the latter have a cutout for the previous swatch. Even though it renders in front (in DOM order), because it never actually overlaps a circle/crescent before it, it appears as if it were behind.
For selected items we only use circles, so the cutout is not there -- and the order appears inverted.

Compare it with the web library where we have used crescents for both selected and unselected items, which enables the correct order without the z-index hack for selected items.

I can look into migrating that approach here.

ramonmi:
I don't recall which version this issue started in, as I only recently discovered it.

However, after checking Zotero 7.1-beta.39+0acfcd3f9_win-x64, I found that the label order issue of the selected items was already present in that version.

> I can look into migrating that approach here.

Thanks. That approach would be more fitting.

tnajdek:
Fixed in ad4bb17db91137dd89d2f62a584a9c109c483423

PR #5884, opened 2026-04-10
Add a icon for "Relate Items"

ramonmi:
What do you think about reusing the "related" icon? cc: @yexingsha 

image


yexingsha:
Looks good to me!

ramonmi:
Thanks!

dstillman (changes requested):
Let's rename to `related.svg` to match the existing file in the tree. We can't reuse that one without refactoring in the mixin, but no need for a different name.

We also already have a visually identical `link.svg` in `universal`, which is used by `attachments-web-link`. Semantically different, but we could rename to `chain-link.svg` and change both to use that. But for 1KB, probably not worth worrying about. So let's just keep the filenames consistent.

ramonmi:
> So let's just keep the filenames consistent.

Done.

PR #5889, opened 2026-04-22
Make icons directory during build

Mynacol:
I'm not quite sure why your build scripts don't need that, but the builds for [Alpine Linux](https://gitlab.alpinelinux.org/alpine/aports/-/blob/19e1f927f8491b433b9c2de8395daf3f28d8658a/community/zotero/zotero_build-modifications.patch#L41) and [NixOS](https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/zo/zotero/build-fixes.patch#L33) require this patch. It shouldn't hurt anyone, and reduce our patches a bit.

### DIFF
--- a/app/build.sh
+++ b/app/build.sh
@@ -1055,6 +1055,8 @@ if [ $BUILD_LINUX == 1 ]; then
 		mkdir "$APPDIR/integration"
 		cp -RH "$CALLDIR/modules/zotero-libreoffice-integration/install" "$APPDIR/integration/libreoffice"
 		
+		# Make icons dir for firefox builds without an updater (i.e. Linux distributions)
+		mkdir -p "$APPDIR/icons"
 		# Copy icons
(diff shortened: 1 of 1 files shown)


dstillman:
Stock Firefox tarballs have an `icons` folder containing updater.png. If Alpine/NixOS rebuilds don't (because they've already removed the Mozilla updater?), that could be the issue.

In any case, this would break any build using an official tarball — where the folder already exists — due to not using `-p`. Fix the `mkdir` and we can merge.

dstillman:
Also worth adding a comment explaining that this is needed only for third-party builds from Firefox tarballs that have removed the updater and don't have `icons`, so it doesn't get removed by accident in the future.

Mynacol:
Great, also for knowing what's going on. Not sure why I dropped the `-p` for this PR. Now it cannot break anything :)

Mynacol:
@dstillman can we merge this? :)

Mynacol:
Thanks!

PR #5890, opened 2026-04-22
Add support for clearing challenge in browser during translation

AbeJellinek:
Generalizing `Zotero.BrowserDownload` to `Zotero.BrowserRequest`. (I couldn't find any plugins using `BrowserDownload` directly, so I didn't keep the old namespace for compatibility, but we can if it's a concern.)

And replace `cookieContextId` with `userContextId` to match standard Firefox terminology. No reason to use something different in our code.

Needs zotero/translate#45


AbeJellinek:
(Tests will fail without the updated submodule)

chrome/content/zotero/xpcom/prefs.js
```diff
@@ -220,6 +220,21 @@ Zotero.Prefs = new function () {
 					case 21:
 						this.set('firstRunGuidanceShown.readAloud', false);
 						break;
+
+					// downloadPDFViaBrowser.* -> browserRequest.*;
+					// module was generalized to handle more than PDF downloads
+					case 22:
```
AbeJellinek, line 226: We're normally OK leaving somewhat outdated pref keys (like `pdfReader`) alone, but to me `downloadPDFViaBrowser` seems specific enough, and now inaccurate enough, that it's worth changing.

PR #5891, opened 2026-04-24
Add support for SDT

AbeJellinek:
Generating via `document-worker`, passing to reader in `getSDT()`.

And:
- Bundle all `document-worker` resources, not just the worker script
- Update submodules for testing

Needs https://github.com/zotero/reader/pull/204, and will need the `document-worker`... `document-worker` branch to be merged into the default branch.

### DIFF
--- a/chrome/content/zotero/preferences/preferences_general.xhtml
+++ b/chrome/content/zotero/preferences/preferences_general.xhtml
@@ -217,6 +217,23 @@
 			</vbox>
 		</groupbox>
 
+		<groupbox id="zotero-prefpane-read-aloud-groupbox" aria-labelledby="preferences-read-aloud-title">
+			<label><html:h2 id="preferences-read-aloud-title" data-l10n-id="preferences-read-aloud-title"/></label>
+
+			<hbox align="center">
+				<label data-l10n-id="preferences-read-aloud-highlight-granularity" control="read-aloud-highlight-granularity-menulist"/>
+				<menulist id="read-aloud-highlight-granularity-menulist"
+						preference="extensions.zotero.reader.readAloud.highlightGranularity"
+						native="true">
+					<menupopup>
+						<menuitem data-l10n-id="preferences-read-aloud-highlight-granularity-paragraph" value="paragraph"/>
+						<menuitem data-l10n-id="preferences-read-aloud-highlight-granularity-sentence" value="sentence"/>
+						<menuitem data-l10n-id="preferences-read-aloud-highlight-granularity-word" value="word"/>
+					</menupopup>
+				</menulist>
+			</hbox>
+		</groupbox>
+
 		<groupbox id="zotero-prefpane-note-groupbox" aria-labelledby="preferences-note-title">
 			<label><html:h2 id="preferences-note-title" data-l10n-id="preferences-note-title"/></label>
 			<checkbox id="open-note-in-new-window"
--- a/chrome/content/zotero/xpcom/pdfWorker/manager.js
+++ b/chrome/content/zotero/xpcom/pdfWorker/manager.js
@@ -24,6 +24,7 @@
 */
 
 const WORKER_URL = 'chrome://zotero/content/xpcom/pdfWorker/worker.js';
+const ASSETS_URL = 'chrome://zotero/content/xpcom/pdfWorker/';
 const CMAPS_URL = 'resource://zotero/reader/pdf/web/cmaps/';
 const STANDARD_FONTS_URL = 'resource://zotero/reader/pdf/web/standard_fonts/';
 const WASM_URL = 'resource://zotero/reader/pdf/web/wasm/';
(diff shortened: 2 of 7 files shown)


mrtcode:
`document-worker` default branch now has the new worker. I added various changes all around, so let me know if something is not right.

A few observations:

With some PDF files (i.e. 1k pages), it can take minutes to extract structured text, therefore, we might need caching, though it would probably need to take into account schema and processor version…

We should probably exclude more block types in Read Aloud now because we mostly care about heading and paragraph nodes?

dstillman:
> With some PDF files (i.e. 1k pages), it can take minutes to extract structured text

Can we do it incrementally?

mrtcode:
> > With some PDF files (i.e. 1k pages), it can take minutes to extract structured text
> 
> Can we do it incrementally?

Normally, no. But we can try to do something extremely hacky just to temporarily work around the long wait for very large PDFs. I.e. process up to 100-200 pages at once.

@AbeJellinek That means Read Aloud would need to somehow work on partial data? Any ideas how to solve it reasonably?

dstillman:
Why does it need to be the whole file?

What are the times for shorter files?

What else wouldn't work until the SDT was ready?

dstillman:
We already want to fix the problem of the delay before it starts reading large PDFs. Definitely don't want to make that worse.

mrtcode:
> Why does it need to be the whole file?

For normal SDT, we process each page and build a graph of all candidate objects, including citations, reference sections, figures, math, and so on. The outline tree and page label ranges also depend on all processed pages.

That said, I can make a separate document-worker branch specifically for Read Aloud that supports degraded SDT on limited page ranges.

> What are the times for shorter files?

I'll include my previous test results:
Screenshot 2026-05-01 at 17 26 12


> What else wouldn't work until the SDT was ready?
This is a temporary wiring for Read Aloud alone, nothing else depends on SDT. But everything that depends on our single threaded document-worker WebWorker will be slowed down, which includes fulltext extraction, annotation extraction, etc.

This is temporary wiring for Read Aloud only, so nothing else depends on SDT.

However, everything that depends on our single-threaded document-worker Web Worker will be slowed down. That includes full-text extraction, annotation extraction, and similar work.

In the future, once SDT extraction is fully integrated into Zotero and works automatically like full-text extraction does now, the PDF reader will behave like a regular, basic PDF reader when a PDF is opened and SDT is not ready yet. It will have a basic text layer, which can be used for text selection and Read Aloud. We will also indicate this state in the UI, possibly show SDT extraction progress for the PDF, and set worker priority higher.

But we could add that text layer degradation now while SDT is being processed. That may be the simplest option, instead of maintaining another document-worker branch.

dstillman:
Could you still create an annotation before the SDT was processed? Or not, because we wouldn't be able to do text extraction? Or allow some sort of queued annotation creation that finished when the processing was done?

It looks like this mostly won't be a problem, but the times above 100 pages get a little long.

In any case, yes, making Read Aloud usable ASAP even if there's further processing would be good.

mrtcode:
Without SDT, it would be the same situation we had before SDT: we used the basic text layer, and everything, including annotations, worked.

But that is not sufficient. Note that SDT is only about 2x slower than plain pdf.js, which we have always used for the reader and full-text extraction. So to support instant Read Aloud, even with the basic text layer, we need incremental page processing anyway.

The issue is that Read Aloud internals currently expect to receive the full data up front. We need @AbeJellinek to tell us how that could be changed so Read Aloud can request more data from the PDF view on demand.

dstillman:
The headphones show in the wrong place (and are hard to get to appear) in a header on bitcoin.pdf.

Screenshot 2026-05-21 at 11 55 24 AM

vs. Z9:

Screenshot 2026-05-21 at 11 56 49 AM

dstillman:
Not sure if this is from this PR, but a 402 isn't being handled properly. Stuck at spinning, with this in the console:

```
HTTP POST https://api.zotero.org/tts/speak failed with status code 402 2
TypeError: e.xmlhttp.response.text is not a function
```

dstillman:
When reading is active on page N, right-clicking the middle of a paragraph on page N+1 and choosing "Read Aloud from Here" is restarting reading on page N instead of at the new location. Works properly in Z9.

dstillman:
Highlights can get out of sync with content in Reading Mode when resizing the window:
Screenshot 2026-05-21 at 12 25 09 PM


Screenshot 2026-05-21 at 12 25 07 PM

AbeJellinek:
> Highlights can get out of sync with content in Reading Mode when resizing the window

Can you consistently reproduce this with a specific PDF / sequence of actions? I'm not seeing it.

dstillman:
> Can you consistently reproduce this with a specific PDF / sequence of actions? I'm not seeing it.

This was with 10.48550/arXiv.2405.00208

I have to resize in and out a few times to get it to happen (but I noticed it first when I just shrunk the window).

AbeJellinek:
Headphones icon location, 402 handling, and resize issues should all be fixed. Working on in-text citation skipping.

> When reading is active on page N, right-clicking the middle of a paragraph on page N+1 and choosing "Read Aloud from Here" is restarting reading on page N instead of at the new location. Works properly in Z9.

I haven't been able to reproduce this, unfortunately.

AbeJellinek:
In-text citations should now be skipped. The issue was streaming SDT - we didn't get refs (linking in-text citations to the bibliography) until the whole document was processed. Trying the SDT preload approach instead; I think it works pretty well. If we replace fulltext with SDT, we should have SDT pregenerated for most items already, but for now we generate when the reader opens.

PR #5892, opened 2026-04-24
Citation dialog: resizable width of collectionTree

### LINKED ISSUE #5765: Citation dialog: make collectionTree width more flexible?
Currently, the width of collectionTree in citation dialog is static, same way it is static in `selectItems` dialog. If you have a very nested collections structure, collections at the bottom of the hierarchy have so much indent that they get mainly cutoff. Maybe we should have a dragable resizable splitter, similar to one in ZoteroPane? 

Image

Not directly mentioned in https://forums.zotero.org/discussion/comment/506282#Comment_506282 but that's where I noticed it.

abaevbog:
Divider between collectionTree and itemTree is now draggable and allows to change the width of collectionTree to take up to 40% of the window's width.

Fixes: #5765

https://forums.zotero.org/discussion/comment/511070/#Comment_511070

https://github.com/user-attachments/assets/273d4594-e513-4af2-ae99-64617ff946a6

### DIFF
--- a/chrome/content/zotero/integration/citationDialog.js
+++ b/chrome/content/zotero/integration/citationDialog.js
@@ -218,6 +218,10 @@ function cleanupBeforeDialogClosing() {
 	allParams[currentLayout.type] = params;
 	Zotero.Prefs.set("integration.citationDialog.windowParams", JSON.stringify(allParams));
 
+	if (libraryLayout.collectionTreeWidth) {
+		Zotero.Prefs.set("integration.citationDialog.collectionTreeWidth", libraryLayout.collectionTreeWidth);
+	}
+
 	// Only library mode in annotations dialog
 	if (!DIALOG_STATE.isAddingAnnotations()) {
 		Zotero.Prefs.set("integration.citationDialogLastUsedMode", currentLayout.type);
@@ -523,6 +527,7 @@ class LibraryLayout extends Layout {
 		// on mouse scrollwheel in suggested items, scroll the list horizontally
 		_id("library-other-items").addEventListener('wheel', this._scrollHorizontallyOnWheel);
 		this._initSidepane();
+		this._initCollectionsTreeDivider();
 	}
 
 	// Create item node for an item group and store item ids in itemIDs attribute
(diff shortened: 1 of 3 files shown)


chrome/content/zotero/integration/citationDialog.js
```diff
@@ -218,6 +218,10 @@ function cleanupBeforeDialogClosing() {
 	allParams[currentLayout.type] = params;
 	Zotero.Prefs.set("integration.citationDialog.windowParams", JSON.stringify(allParams));
 
+	if (libraryLayout.collectionTreeWidth) {
+		Zotero.Prefs.set("integration.citationDialog.collectionTreeWidth", libraryLayout.collectionTreeWidth);
```
dstillman: Does this really require a new pref? Can we either reuse `windowParams`, which we already have above, or integrate the `zotero-persist` mechanism we use for this sort of thing in the main window?

chrome/content/zotero/integration/citationDialog.js
```diff
@@ -824,6 +828,53 @@ class LibraryLayout extends Layout {
 			_id("annotations-sidebar-filter").dispatchEvent(new Event('input', { bubbles: true }));
 		});
 	}
+
+	// Draggable divider to resize collectionTree
+	_initCollectionsTreeDivider() {
+		let setCollectionTreeWidth = (width) => {
+			let minWidth = 200;
```
dstillman, line 839: Just set `width` to `minWidth` if falsy and avoid repeating the `200` below?

chrome/content/zotero/integration/citationDialog.js
```diff
@@ -218,6 +218,10 @@ function cleanupBeforeDialogClosing() {
 	allParams[currentLayout.type] = params;
 	Zotero.Prefs.set("integration.citationDialog.windowParams", JSON.stringify(allParams));
 
+	if (libraryLayout.collectionTreeWidth) {
```
dstillman: Would this ever not be truthy, if it's set in the init?

chrome/content/zotero/integration/citationDialog.js
abaevbog: A separate pref is not necessary, I moved `collectionTreeWidth` into `windowParams`. That was actually my initial approach but then I thought a separate pref would be somehow cleaner.

chrome/content/zotero/integration/citationDialog.js
abaevbog: Good call - if width is not specified, we'll default it to `minWidth` here

chrome/content/zotero/integration/citationDialog.js
abaevbog: Not really, that was just a redundant sanity check. I removed it.

abaevbog:
Thank you for the notes! These should be addressed now, along with a fix to a small edge case that I just ran into: min-width sometimes not getting properly updated.

PR #5895, opened 2026-04-28
Add Go menu for better keyboard navigation

abaevbog:
Adds a top-level Go menu to jump to common UI elements without the mouse. Each entry refreshes lazily on popup show, and items reflect what's visible for the current tab.

- Tabs Menu, Locate: available in any tab type. Locate expands the context pane first in reader/note tabs.
- Library, Quick Search: library tab only. Display the shortcut configured by the respective pref.
- Item-pane sections (Info, Abstract, Tags, etc.): listed dynamically for library, reader, and note tabs. Selecting one opens its collapsible section and focuses its header. Sections added later (built-in or via plugins) are picked up automatically.
- Item Info submenu: in library/reader tabs with a regular item visible, display an additional menu with item's fields.

---

Should address a frequent complaint that it takes a very long time to navigate to a specific field in Item Box or to a specific section in Item/Context Pane without using the keyboard. To get to an arbitrary ItemBox field now one can press Fn+Control+F2 on macOS to focus the top-level menu bar, get to Go menu with right arrow, and then navigate it with down arrow or keyboard letters to get to an arbitrary field.

https://github.com/user-attachments/assets/eb90f549-efa6-48ed-b47e-dbee6412276c

If you are in itemBox, it can also take a while to get to another part of the itemPane (e.g. tags box) by tabbing. It's faster to press Escape to get to itemTree and then tab back to focus the section header, and get to the right section with arrows. It's not intuitive that you should go back to go forward, so I assume few people do it. Now, you can jump to, say, Tags pane directly from anywhere.

Finally, other components like Locate button or collapsible sections below the note editor are tricky to get to, so using Go menu for those is probably better too.

Screenshot 2026-04-27 at 6 47 05 PM

### DIFF
--- a/chrome/content/zotero/contextPane.js
+++ b/chrome/content/zotero/contextPane.js
@@ -61,6 +61,15 @@ var ZoteroContextPane = new function () {
 		}
 	});
 
+	Object.defineProperty(this, 'itemDetails', {
+		get: () => {
+			if (_contextPaneInner.mode === 'notes') {
+				return this.activeEditor?.querySelector('links-box');
+			}
+			return document.getElementById('zotero-context-pane-item-deck').selectedPanel;
+		}
+	});
+
 	this.focus = () => {
 		return _contextPaneInner.handleFocus();
 	};
--- a/chrome/content/zotero/elements/abstractBox.js
+++ b/chrome/content/zotero/elements/abstractBox.js
@@ -112,6 +112,11 @@
 			}
 		}
 
+		focusSection() {
+			this.open = true;
+			this._abstractField?.focus({ focusVisible: true });
+		}
+
 		render() {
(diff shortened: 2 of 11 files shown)


abaevbog:
@dstillman, this is what I had in mind for the Go menu that we briefly discussed. We can add some static options like Locate button that it's hard to get to, as well as Library and QuickSearch shortcut destinations. And then, we can display an option for each section of item/context pane, since it's the most complex UI component with a lot of tab stops. 

Incidentally, it also helps with an old issue of accessing tags and related section below note editor: #3675.

What do you think? Now only Item Info has special handling that displays the list of item fields to focus. If helpful, we can do the same for tags and related item fields - but those sections have way less tab stops, so I'm not sure how useful that would be.

PR #5896, opened 2026-04-28
Fix plugin FTL registration

windingwind:
Use one registry to avoid competing of language detect between plugins, as resources declared as optional is identical to resource missing and thus fails the check and returns null for the string (https://searchfox.org/mozilla-esr140/source/intl/l10n/rust/l10nregistry-rs/src/registry/asynchronous.rs#140) Register the plugin FTL for all languages Zotero supports with proper fallback logic so that even resources are declared as required, the check doesn't fail when plugin doesn't provide the resource. FIx Zotero.File.getResourceAsync to use NetUtil channel to handle jar: url with `@`.

We have two different long-existing issues with the FTL registration:

1. Plugin X FTL provides LANG_A, registered in window as optional, later initialized plugin Y doesn't provide LANG_A => plugin X's optional FTL get ignored (https://forums.zotero.org/discussion/comment/510987/#Comment_510987)

2. Zotero in LANG_B, plugin doesn't provide LANG_B, plugin FTL registered in window as required => DOM localized together with the missing plugin l10n will all fallback to en-US (https://github.com/jlegewie/beaver-zotero/issues/162)

So to fix 1. we now register all plugin FTL in one registry instead of multiple, so that no matter if they are optional or required the resource check always works (detailed reason see below)

To fix 2., we simply register all resources for all supported languages in Zotero.

| Before fix for 2 | After |
| ---- | ---- |
|  |  |
| The plugin causes Zotero UI to partially fallback to en-US in fr | the Zotero UI uses fr correctly |

Below is context summarized by CC:

----

Root cause: two interacting Gecko L10nRegistry solver behaviors, exposed by Zotero's per-plugin L10nFileSource registration

Load-order-dependent null for optional: true resources. When a plugin registers its FTL via document.l10n.addResourceIds([{ path, optional: true }]), Zotero adds a per-plugin L10nFileSource. During bundle generation, the async solver's tester maps ResourceOption::MissingOptional to true via !is_required_and_missing() ([asynchronous.rs:140-147](https://searchfox.org/mozilla-esr140/source/intl/l10n/rust/l10nregistry-rs/src/registry/asynchronous.rs#140)) — i.e., a source that doesn't have an optional resource looks identical to one that does. The solver locks in the first "available" source, which is the latest-registered source ([registry/mod.rs:140](https://searchfox.org/mozilla-esr140/source/intl/l10n/rust/l10nregistry-rs/src/registry/mod.rs#140) — sources are accessed in reverse-push order). Then bundle_from_order calls fetch_file_sync, sees MissingOptional, and silently skips the resource. Net effect: plugin A's resource gets routed through plugin B's source, plugin B doesn't actually have the file, bundle ends up empty for that resource → formatMessages returns null. Whether plugin A or plugin B "wins" depends entirely on load order.

MozXULElement.insertFTLIfNeeded registers as REQUIRED, killing entire bundles. That path appends a <link rel="localization"> and Gecko's Document::LocalizationLinkAdded calls mDocumentL10n->AddResourceId(href) with ResourceType::Required — there is no optional variant for the link path. If the plugin only ships en-US, then for any other Zotero locale (e.g. de), no source provides the required resource → solver fails the entire de bundle with MissingResource → Gecko falls back to the en-US bundle for all ids in that batch, including unrelated core Zotero strings. This is what issue [jlegewie/beaver-zotero#162](https://github.com/jlegewie/beaver-zotero/issues/162) was about; switching to optional: true worked around it, which exposed bug 1.

### DIFF
--- a/chrome/content/zotero/xpcom/file.js
+++ b/chrome/content/zotero/xpcom/file.js
@@ -352,13 +352,20 @@ Zotero.File = new function () {
 	}
 
 	/**
-	 * Return a promise for the contents of resource.
+	 * Return a promise for the contents of a local resource URL as a UTF-8
+	 * string. Goes through an nsIChannel to handle jar: URLs containing '@'.
 	 *
-	 * @param {String} url - the resource url
+	 * @param {String} url - the resource url (file://, jar:, resource://, chrome://)
 	 * @return {Promise<String>} the resource contents as a string
 	 */
 	this.getResourceAsync = function (url) {
-		return getContentsFromURLAsync(url);
+		let channel = NetUtil.newChannel({
+			uri: url,
+			loadUsingSystemPrincipal: true,
+			securityFlags: Ci.nsILoadInfo.SEC_ALLOW_CROSS_ORIGIN_SEC_CONTEXT_IS_NULL,
+			contentPolicyType: Ci.nsIContentPolicy.TYPE_OTHER,
+		});
+		return this.getContentsAsync(channel, 'UTF-8');
 	}
 	
 	
--- a/chrome/content/zotero/xpcom/plugins.js
+++ b/chrome/content/zotero/xpcom/plugins.js
@@ -40,7 +40,15 @@ Zotero.Plugins = new function () {
 	var scopes = new Map();
 	var observers = new Set();
 	var addonVersions = new Map();
-	var addonL10nSources = new Map();
+
+	// All plugins' FTL content is consolidated into one L10nFileSource, so the
+	// solver never has to choose between competing per-plugin sources.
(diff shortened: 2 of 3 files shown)


dstillman:
Doesn't block this, and was preexisting, but .ftl files can collide if they use the same filename, right?

windingwind:
> Doesn't block this, and was preexisting, but .ftl files can collide if they use the same filename, right?

Yes, I guess we kept this because plugins need to load resources with the original filename, and it would be confusing if we namespaced them. The plugins will anyway have to namespace their resources (and they should already be doing so, otherwise they will get reports from users).

PR #5897, opened 2026-04-30
Encrypt API key and WebDAV password using OS keychain

dstillman:
Wraps the values stored in nsILoginManager with OSKeyStore, which derives its master key from Keychain on macOS, DPAPI on Windows, and libsecret on Linux. A copy of the profile alone is no longer enough to extract these credentials.

Existing plaintext entries are mirrored once per session to a new "(encrypted)" realm but preserved in the original realm so a user can still downgrade to a release that doesn't know about encryption. Active credential changes (sign in, sign out, password change) write to the encrypted realm only and remove the legacy entry. A future version can clear any remaining legacy entries on startup.

Patches MOZ_APP_BASENAME in the bundled runtime so the keychain master key is labeled "Zotero Encrypted Storage" rather than "Firefox Encrypted Storage", with a check_line guard so a future Mozilla change to the OSKeyStore label format fails the build instead of silently rebranding the entry. Also fixes check_line to take an explicit file argument.


dstillman:
I've been running this seemingly without trouble, so let's try it.

PR #5899, opened 2026-05-02
fix: push non-orderable itemDetail sections to the bottom

ramonmi:
Ensure `orderable: false` sections are consistently rendered last, resolving previous display order inconsistencies.

@windingwind, kindly request your review on this change when you have a moment.

### DIFF
--- a/chrome/content/zotero/elements/itemDetails.js
+++ b/chrome/content/zotero/elements/itemDetails.js
@@ -395,7 +395,28 @@
 				if (this._pendingRender) {
 					elem.pendingRender = true;
 				}
-				this._paneParent.append(elem);
+				// If the section to insert is non-orderable, directly append it
+				if (!this.sidenav.isPaneOrderable(paneID)) {
+					this._paneParent.append(elem);
+				}
+				else {
+					// Loop from tail to find the first orderable section and insert after it
+					let children = Array.from(this._paneParent.children);
+					let inserted = false;
+					for (let i = children.length - 1; i >= 0; i--) {
+						let child = children[i];
+						let childPaneID = child.dataset.pane;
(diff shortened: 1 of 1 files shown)


chrome/content/zotero/elements/itemDetails.js
```diff
@@ -395,7 +395,20 @@
 				if (this._pendingRender) {
 					elem.pendingRender = true;
 				}
-				this._paneParent.append(elem);
+				let children = Array.from(this._paneParent.children);
+				let inserted = false;
+				for (let child of children) {
```
windingwind: Optimizations:

1. Since we have the assumption that non-orderable sections are always at the end of the list, we should loop from the tail of the list and find the first orderable section to insert after it, instead of from the head of the list
2. If the section to insert itself is non-orderable, we can directly append.

ramonmi: Thanks. Addressed.

chrome/content/zotero/elements/itemDetails.js
```diff
@@ -395,7 +395,28 @@
...
+				}
+				else {
+					// Loop from tail to find the first orderable section and insert after it
+					let children = Array.from(this._paneParent.children);
+					let inserted = false;
+					for (let i = children.length - 1; i >= 0; i--) {
+						let child = children[i];
+						let childPaneID = child.dataset.pane;
+						if (this.sidenav.isPaneOrderable(childPaneID)) {
+							this._paneParent.insertBefore(elem, child.nextSibling);
```
windingwind: this is confusing, isn't it as equivalent `after` the current `child`?

ramonmi: Yes, exactly. Addressed.

PR #5900, opened 2026-05-02
fix: ensure consistent ordering of rows inserted by plugins

ramonmi:
Ensure deterministic ordering for rows with the same `position`.

@windingwind, kindly request your review on this change when you have a moment.

### DIFF
--- a/chrome/content/zotero/elements/itemBox.js
+++ b/chrome/content/zotero/elements/itemBox.js
@@ -953,51 +953,45 @@
 		restoreCustomRowElements() {
 			if (!this._customRowElemCache) return;
 			for (let position of Object.keys(this._customRowElemCache)) {
-				this.insertCustomRow(this._customRowElemCache[position], position);
+				this._customRowElemCache[position].forEach((rowElem) => {
+					this.insertCustomRow(rowElem, position);
+				});
 				this._customRowElemCache[position] = [];
 			}
 		}
 
 		insertCustomRow(rowElem, position = "end") {
-			// Handle both single elements and arrays
-			let elements = Array.isArray(rowElem) ? rowElem : [rowElem];
-			
-			// Determine insertion order based on position
-			// Only reverse when there are multiple elements
-			if (position !== "end" && elements.length > 1) {
-				// For 'start' and 'afterCreators', reverse to maintain order
-				elements = elements.reverse();
-			}
-			
-			for (let elem of elements) {
-				switch (position) {
-					case "start": {
-						this._infoTable.prepend(elem);
-						break;
+			switch (position) {
+				case "start": {
+					let itemTypeRow = this._infoTable.querySelector(".meta-label[fieldname=itemType]")?.parentElement;
+					if (itemTypeRow) {
+						this._infoTable.insertBefore(rowElem, itemTypeRow);
 					}
-					case "afterCreators": {
-						// The `_firstRowBeforeCreators` is actually the first row after creator rows
-						if (this._firstRowBeforeCreators) {
-							this._infoTable.insertBefore(elem, this._firstRowBeforeCreators);
-							// Update the anchor node for creator rows
-							this._firstRowBeforeCreators = elem;
-						}
-						else {
-							this._infoTable.append(elem);
-						}
-						break;
+					else {
+						this._infoTable.append(rowElem);
 					}
-					case "end":
-					default: {
-						let dateAddedRow = this._infoTable.querySelector(".meta-label[fieldname=dateAdded]")?.parentElement;
-						if (dateAddedRow) {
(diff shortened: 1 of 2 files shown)


chrome/content/zotero/elements/itemBox.js
```diff
@@ -861,7 +861,15 @@
 
 		renderCustomRows(rowIDs) {
 			let { options: targetRows, updateID } = Zotero.ItemPaneManager.customInfoRowData;
-			
+
+			// Ensure deterministic ordering for rows with the same `position`.
+			// Sort `rowID` so insertion (prepend/insertBefore) produces a
+			// stable visual order regardless of registration timing.
+			if (Array.isArray(targetRows) && targetRows.length > 1) {
```
windingwind: Should do the sorting when building `Zotero.ItemPaneManager.customInfoRowData` instead of everything the render gets called.

chrome/content/zotero/elements/itemBox.js
```diff
@@ -881,19 +889,34 @@
...
-			// Add rows that are in the target rows but not in the current rows
+			// Build ordered row lists by position for deterministic insertion/reordering
+			let startRows = [];
+			let afterCreatorsRows = [];
+			let endRows = [];
 			for (let row of targetRows) {
 				let rowElem = this._infoTable.querySelector(`[data-custom-row-id="${CSS.escape(row.rowID)}"]`);
+				let position = row.position || "end";
 				if (rowElem) {
+					rowElem.dataset.position = position;
```
windingwind: Is it necessary here? I assume the row never changes its position

chrome/content/zotero/elements/itemBox.js
```diff
@@ -931,13 +954,37 @@
...
 
 				this.updateCustomRowProperty(rowElem);
 
 				// The row is new, always update the data
 				this.updateCustomRowData(rowElem);
 			}
+
+			// Insert collected rows in stable order.
+			// For `start` and `afterCreators`, prepend in reverse
+			// so the final order matches targetRows order.
```
windingwind: This part, as well as the one below in `restoreCustomRowElements`, can simply be replaced by changing the logic in `insertCustomRow`.

Since the array's order is already secured by the `Zotero.ItemPaneManager.customInfoRowData`, we shouldn't need startRows/afterCreatorRows/endRows to manually build the order here.

windingwind (changes requested):

chrome/content/zotero/elements/itemBox.js
ramonmi: All addressed, thanks.

chrome/content/zotero/elements/itemBox.js
```diff
@@ -953,41 +953,51 @@
...
-				case "afterCreators": {
-					// The `_firstRowBeforeCreators` is actually the first row after creator rows
-					if (this._firstRowBeforeCreators) {
-						this._infoTable.insertBefore(rowElem, this._firstRowBeforeCreators);
-						// Update the anchor node for creator rows
-						this._firstRowBeforeCreators = rowElem;
-					}
-					else {
-						this._infoTable.append(rowElem);
+			// Handle both single elements and arrays
```
windingwind: Any reason for this change here? This is only for internal use, and I would prefer not over-engineering it.

ramonmi: This addresses the issue where, when restoring cached custom rows, rows at non-end positions must be inserted in reverse order to prevent order inversion when switching items.
To avoid modifying the `insertCustomRow` function, I will push a commit to resolve this in the `restoreCustomRowElements` function instead.

windingwind: I assume we just need to make sure the `insertCustomRow` treats the newly inserted element in the same order as how the method is called. It doesn't necessarily need to accept an array to achieve that.

chrome/content/zotero/elements/itemBox.js
```diff
@@ -953,7 +953,11 @@
 		restoreCustomRowElements() {
 			if (!this._customRowElemCache) return;
 			for (let position of Object.keys(this._customRowElemCache)) {
-				this._customRowElemCache[position].forEach((rowElem) => {
+				let rowElems = this._customRowElemCache[position];
+				let elementsToRestore = position === "end"
+					? rowElems
+					: [...rowElems].reverse();
```
windingwind: As commented earlier, the `insertCustomRow` should be able to insert the row in the order they should, instead of relying on an external `reverse` to ensure the order.

ramonmi: That makes sense—thanks for the clarification. I've addressed this.

windingwind:
looks good to me, thanks

Issue #2269, opened 2021-12-10
Look for PDF via DOI URL when saving from PubMed


dstillman:
Or, better yet, provide a way for translators to return a URL where translation might reveal a PDF

https://forums.zotero.org/discussion/comment/395230/#Comment_395230

But until/unless we support EZproxy in-app this would be limited to IP/VPN access, like Find Available PDF

pi-dal:
I dug into a current PubMed save failure and there is a narrower fallback worth adding here.

On PubMed pages such as `https://pubmed.ncbi.nlm.nih.gov/35288754/`, the item metadata includes both:
- DOI: `10.1093/nar/gkac173`
- PMCID: `PMC9262588`

The PubMed translator already stores the PMCID in `Extra`, and the connector passes the saved item through to the desktop resolver flow. But the attachment resolver path only considers DOI-based OA lookup here, so if the DOI-based route does not yield a PDF, Zotero never falls back to the PMC article URL even when PubMed explicitly shows `Free full text at PubMed Central`.

A useful fix for this issue would be:
- treat PMCID as a valid attachment-resolver signal for PubMed-saved items
- add a PMC OA resolver based on `https://pmc.ncbi.nlm.nih.gov/articles/<PMCID>/`
- keep the DOI-based resolver in place, with PMC as an additional fallback rather than a replacement

That keeps the logic in the existing resolver path instead of adding a PubMed-specific branch in the connector.

PR #5906, opened 2026-05-09
fix(attachments): add PMCID fallback for PubMed resolver lookup

pi-dal:
## Summary
- treat PMCID as a valid full-text lookup signal when deciding whether an item has attachment resolvers
- add a PubMed Central resolver based on the saved PMCID before DOI-based OA lookup
- cover the PMCID fallback path in attachment resolver tests and the connector resolver endpoint test

## Problem
PubMed-saved items can include a `PMCID` in `Extra` and show a `Free full text at PubMed Central` path even when DOI-based full-text lookup does not produce a PDF. The connector already passes the saved item through to the desktop resolver flow, but the resolver logic only considered DOI and URL data, so this PMC fallback was never attempted.

## Implementation Notes
The change keeps the logic in the existing desktop attachment resolver path rather than adding a PubMed-specific branch in the connector. `getFileResolvers()` now normalizes `PMCID` values and adds a PMC article resolver at `https://pmc.ncbi.nlm.nih.gov/articles/<PMCID>/`. `canFindFileForItem()` also treats `PMCID` as a valid signal so the connector fallback endpoints can surface the resolver.

## Test Plan
- `test/runtests.sh -b -f -g "should include a PubMed Central resolver from a PMCID in Extra|should include a PubMed Central resolver alongside DOI OA resolvers|should allow finding files for an item with a PMCID in Extra|should respond with .true. if the item has a PMCID fallback" attachments server_connector`

Refs #2269

### DIFF
--- a/chrome/content/zotero/xpcom/attachments.js
+++ b/chrome/content/zotero/xpcom/attachments.js
@@ -1239,9 +1239,10 @@ Zotero.Attachments = new function () {
 	
 	
 	this.canFindFileForItem = function (item) {
+		let pmcid = item.getField('PMCID') || item.getExtraField('PMCID');
 		return item.isRegularItem()
 			&& !item.isFeedItem
-			&& (!!item.getField('DOI') || !!item.getField('url') || !!item.getExtraField('DOI'))
+			&& (!!item.getField('DOI') || !!item.getField('url') || !!item.getExtraField('DOI') || !!pmcid)
 			&& this.FIND_AVAILABLE_FILE_TYPES.every(type => item.numFileAttachmentsWithContentType(type) == 0);
 	};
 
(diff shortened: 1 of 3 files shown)


pi-dal:
cc @dstillman

This change keeps the fix in the desktop attachment resolver path rather than adding a PubMed-specific branch in the connector.

What it does:
- treats `PMCID` as a valid signal for attachment resolver eligibility
- adds a PMC resolver at `https://pmc.ncbi.nlm.nih.gov/articles/<PMCID>/`
- keeps the existing DOI-based OA lookup in place as an additional path
- adds targeted tests for both resolver generation and `/connector/hasAttachmentResolvers`

The motivating case is PubMed saves that already contain a `PMCID` and show a PubMed Central full-text path, but where DOI-based lookup does not produce a PDF.

Would appreciate a review when you have a chance.

pi-dal:
cc @dstillman

Just a gentle follow-up in case this slipped through.

This PR adds a PMCID-based PMC fallback in the existing desktop attachment resolver flow for PubMed saves, and includes targeted coverage for both resolver generation and the connector resolver check.

Would appreciate a review when you have a moment. Thanks.

dstillman:
Thank you! I switched it to use the real PMCID field, which all current Zotero clients have. (The PubMed translator still saves to Extra, but that gets auto-migrated to the real PMCID field. We should update the translator to save directly to `PMCID`, but it's not a blocker for this.)

pi-dal:
Thanks for the review, the improvements, and for getting this merged!

Switching to the real PMCID field is definitely cleaner than relying on an Extra fallback, and it's good to know the migration path already covers existing PubMed imports. I also appreciate the small refinements to the tests and resolver logic—they make the final version feel more consistent with the rest of the codebase.

And thanks for the note about the translator as well. It would be nice to have it write directly to PMCID eventually, but I'm glad the current behavior isn't a blocker for users.

I learned quite a bit from the review process. Really appreciate you taking the time to look through this and help get it across the finish line.

PR #5914, opened 2026-05-14
 Add Abstract section to merge pane with value chooser 

AbeJellinek:
- Extract basic logic from `ItemDetails` into a new `ItemPaneContainerBase`
- Turn `DuplicatesMergePane` into a full-fledged implementation of that class with collapsible Info and Abstract sections (but no support for reordering or pinning), instead of only showing a single non-collapsible Info box
- Add value chooser to `AbstractBox`
- Convert the item pane CE base classes to ES modules (imported with `global: 'current'` so DOM is available) - cleaner than defining globally via `loadSubScript()`


dstillman:
"Disable reordering/pinning interactions entirely" → "Disable reordering/pinning in the duplicates pane", right?

AbeJellinek:
Better: "Hide reordering/pinning menuitems and prevent drag when disallowed". But really that commit should just be rebased into the first one (it was a fix for nonfunctional menuitems still showing - the actions were already prevented). I'll do that.

AbeJellinek:
Or this whole thing can be squashed into one commit.

dstillman:
Keeping the ESM migration separate seems reasonable

AbeJellinek:
Done

PR #5916, opened 2026-05-14
Citation dialog: display preview of the citation

### LINKED ISSUE #5910: Citation dialog: Contextual full-citation preview?
https://forums.zotero.org/discussion/comment/512174/#Comment_512174

@yexingsha, should we see what this would look like? I assume a button in the bottom toolbar that brought up a pane at the bottom? This would show the full multi-item citation as it would be rendered in context in the document (like the old editor did), with all citation customizations (prefix, suffix) applied.

abaevbog:
In Add/Edit Citation mode, display a preview of the citation in the bottom section. The section can be hidden/displayed via the checkbox in the settings popup.

Fixes: #5910

Screenshot 2026-05-14 at 3 01 04 PM
Screenshot 2026-05-14 at 2 59 19 PM

---

@yexingsha, how is this looking to you? I started by placing the checkbox to display/hide preview into the settings popup because the bottom area seemed more balanced this way. But let me know if you think it's worth to move it into a separate button.


yexingsha:
I think the main issue with putting the preview toggle into the settings popup is that it's hard to discover; even after users first find it, it's still easy to forget after a while. The other issue is that some people might want to only turn on preview for complex citations, and burying the toggle means one extra click every time. So I would prefer surfacing it as a separate button.

Regarding the visual imbalance issue, I guess it's okay if we keep the list/library toggle on the right. My rationale was to group similar controls together, but keeping the toggle on the right also makes sense if we think of the right side as dialog-specific controls. We can also change the order of these buttons to make them look more balanced: from left to right, it's settings, preview, and then list/library toggle.

abaevbog:
> I think the main issue with putting the preview toggle into the settings popup is that it's hard to discover; even after users first find it, it's still easy to forget after a while. The other issue is that some people might want to only turn on preview for complex citations, and burying the toggle means one extra click every time.

Ok, yeah, that's a very fair point. How about this option with the button to the right? The only drawback is there's a bit of jumpiness when you switch between list and library layouts because the citation preview toggle is shown/hidden then. But otherwise, I think it looks nice.

Screenshot 2026-05-14 at 4 56 57 PM

yexingsha:
This works, though we can put the buttons on the right in this order from left to right: citation preview toggle - settings - list/library toggle. I think this will look better and also alleviate the jumpiness issue.

abaevbog:
Good call! No jumpiness now:


https://github.com/user-attachments/assets/886bde44-c26a-44b8-9565-d31bc429dc4f

abaevbog:
@dstillman, you mentioned we may want to add a "Looking for the classic dialog?" link somewhere. I think it would be very relevant here, since there is a preview that looks similar to the old editor. I assume if you are coming from the classic citation dialog, you might think that you can edit the citation here.

So what if we show a small link below the rendered citation? That resolves the question of how to dismiss the link - you just hide the citation preview panel.

yexingsha:
Oh, I didn't realize we aren't adding this to list mode. I think people only ask for it in library mode because they see it as a replacement for the classic dialog, but a preview would be helpful in list mode as well.

abaevbog:
> I think people only ask for it in library mode because they see it as a replacement for the classic dialog, but a preview would be helpful in list mode as well.

Yeah, that was my thinking to only add it to the library because it's a direct replacement of the "classic dialog", and quick format (list mode predecessor) never had it. I suppose there is no harm to make it visible in all layouts.

I also thought list layout resizes a bit too often (likely on every search) to pin something like a preview panel to the bottom because it would frequently change it's positioning. But I guess it's ok because if you don't want it, you just hide it, right?

abaevbog:
Yeah it looks just fine in list mode to me. For most styles, the preview panel isn't very large so it doesn't affect the UI of list mode as much as I assumed it would.

Screenshot 2026-05-19 at 2 18 41 PM

dstillman:
I guess one question here would be whether this needs to be toggleable at all or if the preview aspect of the editor was just a useful feature of the classic dialog that we should restore to the main dialog. It's a little more visual noise, but it seems like a reasonable thing to always show. And for the people who want it, the button would be visual noise too.

But then maybe we'd want to tweak the appearance a little. It looks a little bare right now. It would take a little more vertical space, but it might need a "Preview" header? @yexingsha, can you play with that?

dstillman:
Not particularly useful if you just use numeric styles, though even there if you use a prefix/suffix it's still a little useful.

yexingsha:
Adding a header does work, also we can use a serif font to distinguish it a bit more:

preview-library

preview-list

With the addition of this header, I'm leaning more toward keeping this toggleable to save some space, both for the people who use numeric styles and for those who seldom modify their citations.

dstillman:
https://github.com/user-attachments/assets/3812d22f-45d5-4daf-8a61-6ca9e70fa5e9

Can we slide to make space before showing the content?

dstillman:
@yexingsha: Header is maybe a little distracting? Can we get away with non-bold?

dstillman:
What's the argument for that icon? Maybe some more specific to previewing and less similar to the config icon (horizontal lines) would be clearer? A more standard eye or magnifying glass?

abaevbog:
> Can we slide to make space before showing the content?

Good call, will do.

> What's the argument for that icon? Maybe some more specific to previewing and less similar to the config icon (horizontal lines) would be clearer? A more standard eye or magnifying glass?

Yeah, we do have 2 icons with 3 horizontal lines next to each other. I see the idea, with each line representing text, and it's surrounded by the quotes but they do look somewhat similar at a glance.

---

Initially, I imagined this would be mainly a backwards-compatibility feature, to ensure that those who are used to it in the classic dialog would still have access to the preview. It is off by default on first open at the moment. But if one is used to that preview from classic dialog, they can toggle it on. The visible toggle is indeed necessary for discoverability in this case, since otherwise you would not know this option appeared.

If we are now considering it as a generally more useful feature for many people, we could default the preview to being visible on first open. In that case, hiding the toggle in the settings panel could be considered again, since you would know that the preview is available, and you would toggle it off only if it's in the way. 


One more idea for hiding/displaying the toggle I have is to borrow some inspiration from the header of the `collapsible-section`. We could have an always-visible header saying "Preview" (maybe less prominent than now), with a chevron, and clicking that would expand/collapse the preview without adding a new button or hiding the toggle in the settings panel. I don't know if it's better than a visible toggle but just a thought.

yexingsha:
I was worried that an eye icon would be too vague as to what would be shown, and we needed something that's obviously "citation" for people to discover this, but that was also based on the assumption that preview would be off by default. If it's on by default, then an eye looks fine:

view

preview-regular-eye

The non-bold header looks a bit too faint to me, and I don't think making it black is the solution. Making it bold but 12pt would look like this (also trying out a different preview icon):

preview-small-viewfinder

Putting the toggle inside the settings panel is also okay, though one concern is whether people will be able to find it again after they toggle it off. An always-visible header saying "Preview" seems a little wasteful of space. I think I still slightly prefer the toggle being an always visible button.

abaevbog:
The resizing issue should be addressed. Now the window displays the preview after the resize animation is complete.

Making the preview header a bit smaller looks good to me too! Between the two icons, I personally prefer the second one. An eye icon does feel a bit too ambiguous, especially if we ever add any other element that is shown/hidden in the future. But the quotes surrounded by brackets get the message across, I think.

How does this look?

https://github.com/user-attachments/assets/5bb3deaf-964b-42e2-bad2-84932537a19c

yexingsha:
Looks good to me!

abaevbog:
Great, I rebased and squashed into one commit. I also updated the pref to show the preview by default (correct me if we want to not show it instead).

dstillman:
I think the eye makes more sense? That's just pretty universally the icon for "preview". The other one is more of a zoom (e.g., I think YouTube used to use the outer lines for full-screen mode?). And the quotation marks don't really uniquely speak to a rendered preview — that's our icon for a citation, but this whole dialog is about citations.

I'd even say that with the eye icon visible and toggled on, we could maybe avoid the "Preview" header altogether and save some space. The bold header is just a bit noisy.

dstillman:
Trying with just a sans-serif "Preview" for the placeholder and no header.

Screenshot 2026-06-09 at 12 14 28 PM

Screenshot 2026-06-09 at 12 14 38 PM

I can't find the eye icon to swap it in.

abaevbog:
With the "eye" icon:

Screenshot 2026-06-09 at 9 33 11 AM

---

I don't think I realized that an "eye" does seem to be a standard for anything preview-related. A quick google search for "preview icon" mainly shows different variations of an eye. So this works!

PR #5918, opened 2026-05-15
Read Aloud: Support custom Zotero/OpenAI endpoint

AbeJellinek:
For use with OpenAI, OpenAI-compatible APIs like https://github.com/remsky/Kokoro-FastAPI, or a custom local implementation of the Zotero TTS API that bridges it to a different provider.


AbeJellinek:
@dstillman: Could we get this in?

PR #5925, opened 2026-05-19
Citation dialog: sort libraries by cited count in list mode

### LINKED ISSUE #5924: Citation Dialog: wrong sorting of libraries in list mode
https://forums.zotero.org/discussion/131660/libraries-displaying-in-reverse-order-when-adding-citations-in-word-after-update

Currently, libraries in list mode are sorted by the number of their search matches. Previously, in [Quick Format](https://github.com/zotero/zotero/blob/7dad8377df8901d50573d38eddb8ebbeee874a71/chrome/content/zotero/integration/quickFormat.js#L735-L749), we would sort the libraries by the number of cited items from each library. If all else is equal, libraries would be sorted by their ID, so they are the same as in collection tree.

abaevbog:
In list mode, sort libraries by the count of cited items from each library, falling back to sorting by libraryID when cited count is equal. Restores QuickFormat [behavior](https://github.com/zotero/zotero/blob/7dad8377df8901d50573d38eddb8ebbeee874a71/chrome/content/zotero/integration/quickFormat.js#L735-L749).

Fixes: #5924

Note the "Embeddings" group being moved to the top after cited items are loaded because the cited item belongs to that group:

https://github.com/user-attachments/assets/d2d50056-6453-467a-8a17-3f89ae99771d

### DIFF
--- a/chrome/content/zotero/integration/citationDialog/searchHandler.mjs
+++ b/chrome/content/zotero/integration/citationDialog/searchHandler.mjs
@@ -50,6 +50,9 @@ export class CitationDialogSearchHandler {
 		this.selectedItems = null;
 		this.openItems = null;
 		this.citedItems = null;
+		// number of cited items in each library, keyed by libraryID.
+		// Populated once cited items are loaded in refreshCitedItems.
+		this.citedItemCountsByLibrary = {};
 	}
 
 	setSearchValue(str, enforceMinQueryLength) {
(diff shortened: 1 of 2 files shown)


dstillman (commented):
Not sure why we were ever falling back to libraryID. Wouldn't My Library followed by alphabetical make more sense? I imagine libraryID was just a quick hack to get My Library first.

(And the relevant bug for the thread is sorting by number of search results rather than cited items, right? The fallback shouldn't be operative if one is a library they never use.)

abaevbog:
> Not sure why we were ever falling back to libraryID. Wouldn't My Library followed by alphabetical make more sense? I imagine libraryID was just a quick hack to get My Library first.

That makes sense. It means the order is slightly different than it used to be in quick format though. But it is somewhat more intuitive, so it's probably worth it. We could consider adding some kind of indicator of how many cited items there are to make it clearer why exactly groups are sorted the way they are. We were already looking into something similar in https://github.com/zotero/zotero/pull/5636 (e.g. https://github.com/zotero/zotero/pull/5636#issuecomment-3599909637), so maybe it's best to do it there.

> (And the relevant bug for the thread is sorting by number of search results rather than cited items, right? The fallback shouldn't be operative if one is a library they never use.)

Correct, yes. Before, larger groups with more matches would always win. Now, we sort libraries by the number of cited items in the document. When the number of cited items is the same, groups are sorted alphabetically. "My Library" gets a special priority treatment, so it appears above any other group that has the same number of cited items (so when nothing is cited, it's at the top).

I re-pushed to update the commit message with the alphabetic sorting behavior.

PR #5932, opened 2026-05-27
Citation dialog: show placeholder after a new bubble is added

abaevbog:
After a new bubble is added, display a placeholder on the last input indicating that one can type a number to add a locator to it.

https://github.com/user-attachments/assets/7e424eaa-83b3-4c69-86d8-11e35801e8fa

---

If the placeholder does not fully fit on the current line in multi-item citation, it will be truncated to avoid overflowing onto the next line. 

Initially, I tried the opposite, to always fully display the placeholder and let it overflow. But it often left too much space above it, enough for 2 or 3 bubbles, so it seemed too awkward to me. This is an example from that previous iteration that I am trying to avoid:

Screenshot 2026-05-26 at 11 03 33 PM


abaevbog:
@dstillman, is this what you had in mind to indicate that page locators can be added by typing?

dstillman:
It is, but I'm not going to say it's not awkward…

With a little more padding, it might not be terrible in the default case, though it's not great when it gets truncated. It could appear only after the first bubble to avoid that?

I'm not sure there's anything here that won't be somewhat clumsy, but we just really do want to convey this info somehow.

@yexingsha, any thoughts?

abaevbog:
I don't disagree about the awkwardness, for sure. A placeholder getting truncated is not something common - still I think it's less bad than having it overflow to the second line leaving a lot of empty space on the first line (and resizing the window back and forth in list mode as the placeholder appears/goes away).

> It could appear only after the first bubble to avoid that?

Potentially. My concern would be that now placeholder indicates the current state of the locator input handling. If you see the placeholder, typing "123" will add "123" page locator to that bubble. When the placeholder goes away (e.g. on any focus change) - "123" will run the search. If we only show the placeholder after the first bubble, it almost looks like typing "123" will add it as a page only after the very first bubble is added.

> I'm not sure there's anything here that won't be somewhat clumsy, but we just really do want to convey this info somehow.

The shorter the placeholder is, the less awkward it is looking to me. Since it's mainly about highlighting you can type to add a locator, maybe it's all we need there?

Screenshot 2026-05-28 at 9 46 07 AM



---

The first run guidance panel is already supposed to convey all of this information - but I can see that it can still be skipped if all you need is to add an item this time. Next time, if you want to add a locator, that point of reference is fully gone. Maybe something like a ❓icon in the bottom that would bring up the popup again would be helpful?

yexingsha:
I agree that a `?` icon at the bottom for bringing up first-run popups is broadly helpful, and if we simply want to remind people that they can type locators, maybe adding that plus linking to our (new and better) docs in those popups would be enough.

But if it's not enough, and we decide that this warrants a more visible indicator in the UI, even to the potential annoyance of people who will never add locators in this way, then what this indicator should look like really depends on what exactly we want to achieve here. Do we simply want to let people know that sometimes they can type numbers which will immediately become page locators? If that is the case, the shortened placeholder with 4px padding on the left is not too bad.

Or, do we want to also let them know that they can type other locators? If that is the case, I feel like there's no way to convey this info cleanly via the placeholder approach. The placeholder is already appearing quite frequently, so if we go with the longer version, we might as well make it a proper hint. It can go below the input area and change as you type or switch focus, which should be much clearer and much less awkward, though at the cost of increased complexity. To make things even clearer, we could add some affordance to the last added bubble to solve the issue we discussed in https://github.com/zotero/zotero/pull/5529:

locator-hints

abaevbog:
> But if it's not enough, and we decide that this warrants a more visible indicator in the UI, even to the potential annoyance of people who will never add locators in this way, then what this indicator should look like really depends on what exactly we want to achieve here

Without knowing that you can type "123" to add a page locator, one gets the impression it takes longer to do it (via bubble popup) than if you are used to editor in the classic dialog. This is the main issue we want to clarify. If you work with sources that involve other locators, you can type e.g. "Chapter 13" and press Enter as well, which is also faster than opening a popup (and likely faster than adding it in editor of classic dialog too). So we need to communicate a little bit of both? With the "123" turned into page locator maybe being a bit more important, as it's a more common use case.

---

> even to the potential annoyance of people who will never add locators in this way

The hint below the `bubble-input` looks nice - it's long enough to have a proper sentence with maybe a link to the docs. But it's a good point about potential annoyance of people who don't need it. If you never use locators OR if you already know about them, it's just an unnecessary message that you would rather not have cluttering your UI. We also would need to manage resizing the window in list mode to make space for it/shrink when it goes away, and list mode already resizes quite often.

Now that I look at what the "?" suggestion might look like, we would be accumulating a lot of icons in the bottom right corner. With the upcoming toggle to show/hide citation preview, 3 icons + List/Library toggle looks a bit overwhelming to me. (But I also know that I get overwhelmed by more than a few icons easily, so maybe it's not a general issue).

---

Here's another suggestion. What if we add the hint text with the link into the popup with locator selection itself? These are examples of what happens:

https://forums.zotero.org/discussion/comment/512515/#Comment_512515
https://forums.zotero.org/discussion/131508/old-citation-dialogue-please

Folks seem to discover that you click on the bubble to customize the citation. Then, they type the page locator, click "Done", and it feels like too many steps. So what if we have the text with link to the docs about shortcuts for locator inputs there? That way, we have a place for a proper sentence, linking to docs that explain what locator shortcuts are available, without it all having to clutter main UI. This can go in combination with the short placeholder. Maybe something like "e.g. 10-15 for page" would make sense?

abaevbog:
As an experiment, this is what the instructions + link in popup with shorter locator might look like. What do you think?

Screenshot 2026-05-29 at 1 55 08 PM
Screenshot 2026-05-29 at 1 55 03 PM

yexingsha:
I like this overall direction! I think the helper text can go under the locator text field, and it can be longer to contain more immediately useful information. Maybe "After adding an item, you can type page numbers directly into the main input field, or type other locators then press Enter. Learn more in docs"?

I worry a little that the short placeholder will confuse people as to what they are meant to do. Maybe `type "10-15" for page` so it sounds like an action?

abaevbog:
> I think the helper text can go under the locator text field, and it can be longer to contain more immediately useful information. Maybe "After adding an item, you can type page numbers directly into the main input field, or type other locators then press Enter. Learn more in docs"?

That's fair. We might as well write some useful info since we have space. What do you think about the positioning of the text? This is what it looks like now. With 3 lines of text, it visually breaks the locator input from the other ones somehow oddly?

Screenshot 2026-05-29 at 2 41 08 PM


We could move the text to the right, so it aligns with inputs. It takes up even more vertical space - but oddly looks somewhat nicer to me.

Screenshot 2026-05-29 at 2 41 50 PM

yexingsha:
Yes, I was thinking of aligning it with the inputs. And we can reduce the gap between the input field and the helper text by ~4px.

abaevbog:
Yeah, with decreased vertical spacing it looks nicer!

Screenshot 2026-05-29 at 4 38 02 PM

yexingsha:
It seems the horizontal spacing was changed as well, so the helper text is no longer aligned with the input field. Otherwise this looks good!

abaevbog:
Ah, yes! Good catch, thank you. Inputs have 5px start padding, so we need one for the locator description too:

Screenshot 2026-05-29 at 4 43 32 PM

abaevbog:
On a separate note, I also tried to add click-through to the item details popup here, so you can click on the "Accept" button without having to close the popup first. I think it works well, so it is less steps to add locator/suffix/prefix from the popup - don't need to close it first if you use the mouse.

Similarly, we could accept the popup from everywhere via cmd/ctrl+Enter. It used to be disabled in panels just for consistency - but now you can open the popup add a locator/suffix/prefix, and press Cmd+Enter to accept the dialog without closing the popup first.

abaevbog:
Before we link to anything in the docs from the citation dialog, we'll want to ensure those docs are up-to-date and reflect the current UI. Current https://www.zotero.org/support/word_processor_plugin_usage still mentions the classic citation dialog and has red quick format dialog screenshots and behaviors. 

I attempted to rewrite the sections related to word processor integrations to be more current, update screenshots, and etc. in https://github.com/zotero/zotero-docs/pull/8.

dstillman:
I think that's way too verbose to be a permanent fixture of the popup, so trying a shorter version that also gets hidden after you use the shortcut for the first time:

Screenshot 2026-06-11 at 1 25 40 PM

Screenshot 2026-06-11 at 1 26 20 PM

abaevbog:
I think that works! The phrasing is 100% better now

dstillman:
Also fixed a bug where the placeholder kept being shown after you entered "10-15", and tweaked the string a little:

Screenshot 2026-06-11 at 1 53 25 PM

abaevbog:
> Also fixed a bug where the placeholder kept being shown after you entered "10-15", and tweaked the string a little:

I see! In fact, I was meaning to show the placeholder whenever typing a number would add a locator. 

Now, you can type "10-15", it will be added as a page locator, but if you type the next number (e.g. "1"), it will still be appended to the page locator, making it "10-151". That's the behavior we had in quickFormat - I kept it so that you don't have to type the entire page locator within the brief debounce window.

dstillman:
Yeah, but I think that's fine — it's more confusing to still say "type “10-15”" when you've already done it.

abaevbog:
After the placeholder is gone, does it seem like the next typed number will no longer be added as a page locator? Because it will be, and I wonder if that may cause confusion?

We could clear `just-added` bubbles record as well. Then, there will be no placeholder, and the next number will be a search term - maybe a bit more predictable? It does mean you need to type the entire page locator within the debounce window then.

abaevbog:
Actually, maybe I'm just overthinking this. Once you do it once or twice, you'll get the idea

abaevbog:
> After the placeholder is gone, does it seem like the next typed number will no longer be added as a page locator? 

This concern mainly comes from https://forums.zotero.org/discussion/132134/error-of-add-edit-citation-in-word where the source of confusion is this feature of adding page locators from the main input without it being clear why. So I'm just thinking about how we can be transparent about what's happening.

dstillman:
But the confusion in that thread should be helped just by the initial post-bubble placeholder that this PR is adding. My change is just having it not continue to show after you've already entered a page number, which just seems more confusing. I don't think most people are going to stop typing their page number and look at the placeholder and be confused and not try to keep typing the page number.

(I'm also making a separate change to change the placeholder back to the original string when it's no longer accepting locators.)

abaevbog:
Right, yeah, that make sense.

dstillman:
OK, try that.

Initially after the first add (of any number of items):

<p>Screenshot 2026-06-11 at 3 07 33 PM</p>

If you press right-arrow (as in https://forums.zotero.org/discussion/comment/514037/#Comment_514037):

<p>Screenshot 2026-06-11 at 3 07 41 PM</p>

After you've made another addition:

<p>Screenshot 2026-06-11 at 3 07 56 PM</p>

The idea is that, after you've started adding items, you know the drill, and this mostly avoids placeholder truncation. (You could still hit truncation if you add multiple items in the first add, but it's less likely.)

It also no longer disables the locator entry if you just press down-arrow, only left/right-arrow. I assume disabling on down-arrow was intentional (because once you're navigating, you're thinking about the next ref?), but now that the placeholder changes, it just seemed kind of arbitrary, whereas right-arrow is an explicit signal that you wanted to move on and have a number search instead of be a locator.

abaevbog:
That looks good! It feels more natural than I expected to stop showing the placeholder after the first couple of bubbles. It will remove the need for truncation for 99% of all cases too.

Regarding my scenario from earlier, I think it was OK when we just hid the placeholder after typing "10-15" for the page locator. Then, the user could infer what's happening. But now I feel like the default placeholder that appears after the "10-15" is somewhat deceiving? It declares that now you search for items, so I do as I am told and search for a year - but search doesn't happen. I think if I was a user, this would feel like a bug to me because we are actively showing the same placeholder as when the dialog has no bubbles. Am I overthinking this?

https://github.com/user-attachments/assets/abae15ff-7c27-4b6e-bac7-1a6099e8df64

dstillman:
No, that's fair. Now that we're including "Search" in there, I think it's OK to keep it as long as locator input is active.

yexingsha:
If we are keeping the placeholder around, perhaps we should flip the order of the wording, so that it reads:

> Type "10-15" to cite pages or search for items

So that the more important part gets read first, and there's less chance of it getting truncated in the case of multiple items in the first add.

dstillman:
> Type "10-15" to cite pages or search for items

Yeah, I was thinking about that. It's logically a bit better, since the locator is the new option available in that moment, but the problem is that that's grammatically harder to parse — "Search […] or type […]" vs. "Type […] to cite […] or search" (???).

abaevbog:
I think everything else works!

>  I assume disabling on down-arrow was intentional (because once you're navigating, you're thinking about the next ref?)

And yes, this was exactly the idea. Once you go to navigate items in the list, you are likely not thinking about locators already. No real harm in excluding up/down arrows from clearing just-added bubbles to be consistent.

abaevbog:
> Type "10-15" to cite pages or search for items

Yeah, I see the issue. It's more appropriate but sounds like "10-15" would also search the items. Does a comma help? Something like "Type '10-15' to cite pages, or search for the next item"?

yexingsha:
> but the problem is that that's grammatically harder to parse — "Search […] or type […]" vs. "Type […] to cite […] or search" (???)

That's true. Maybe we can add a comma there?

> Type "10-15" to cite pages, or search for items

dstillman:
<p>Screenshot 2026-06-12 at 2 43 35 PM</p>

After pressing right-arrow:

<p>Screenshot 2026-06-12 at 2 43 42 PM</p>

abaevbog:
I think it's better! "search for items" is an independent clause, so grammatically, I believe this is right. Now the more relevant part is the first thing you see, so I think it's worth it.

dstillman:
Merged in 40f949ba94, 0f2d3e103d, cd29a818b0

Issue #5933, opened 2026-05-28
`navigator.language` is en-US when `Services.locale.requestedLocale` is en-CA


AbeJellinek:
Causing the Last Read column, which uses a windowed `Date` object's `toLocaleString()`, to return en-US results:

https://forums.zotero.org/discussion/131886/last-read-column-date-format-doesnt-respect-zotero-application-language

PR #5934, opened 2026-05-28
Fix US-formatted Last Read dates in en-* alias locales

AbeJellinek:
By applying `dateOverrides.js` to all windows and `require.js` globals.

This issue didn't affect Date Modified etc. because they use `Zotero.Date.sqlToDate()`, which returns a Date object from the global scope, where `toLocale[Date]String()`was already being patched. Now we patch it in all scopes.

Addresses #5933, but `navigator.language` is actually still incorrect - correcting that seems to require turning `en-CA`, `en-AU`, and `en-NZ` into real locales instead of aliases - so that issue isn't fully fixed.

I would love to find a cleaner, more comprehensive fix for this. Firefox uses `toLocaleString()` without passing a locale parameter and it apparently works fine for them, so I assume the issue is caused by something we're doing, but I just can't find it. This PR addresses the symptoms for now.

### DIFF
--- a/chrome/content/zotero/dateOverrides.js
+++ b/chrome/content/zotero/dateOverrides.js
@@ -1,5 +1,9 @@
 /* eslint-disable no-extend-native */
 
+// Patches Date.prototype.toLocale*String in whichever global this script is
+// loaded into, so that date formatting follows the app locale when no
+// explicit locale is passed
+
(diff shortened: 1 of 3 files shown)


dstillman:
This applies to all non-en-US/GB English locales, right?

AbeJellinek:
Right, yes. Other languages are fine.

PR #5935, opened 2026-05-29
Improve Add/Edit Citation button behaviour when cursor is in a non-citation Zotero field

adomasven:
When in bibliography, show an appropriate message.
When in any other non-citation Zotero field - just open an empty edit citation dialog.

image


abaevbog:
Looks good to me!

---

In the future we could show a similar warning on all operations when the cursor is in bibliography. If you press "Add Note" or "Add Annotation", we now have the warning to replace the current field. It may be appropriate to show a similar warning saying that to edit bibliography, there is Add/Edit Bibliography dialog.

Also, we may consider to ignore leftover {Citation} placeholders on "Add Note" and "Add Annotation" operations as well, instead of showing the "Replace current field" warning, since those placeholders are just empty.

chrome/locale/en-US/zotero/zotero.properties
```diff
@@ -894,6 +894,7 @@ integration.error.mustInsertCitation 	= You must insert a citation before perfor
 integration.error.mustInsertBibliography = You must insert a bibliography before performing this operation.
 integration.error.cannotInsertHere		= Zotero fields cannot be inserted here.
 integration.error.notInCitation			= You must place the cursor in a Zotero citation to edit it.
+integration.error.inBibliography		= To insert a citation, place the cursor in the body of the text. To edit the bibliography directly, use the Add/Edit Bibliography button.
```
dstillman: `use the Add/Edit Bibliography button` → `use Add/Edit Bibliography`

But if this applies to Google Docs too, I guess we should pass in the command name as a variable, since it's "Add/edit bibliography" there (and not actually localized right now). And when sentence case, we'd also want to put it in typographical quotes so that it becomes `use “Add/edit bibliography”.`.

adomasven: Done

chrome/content/zotero/xpcom/integration.js
```diff
@@ -1513,10 +1513,24 @@ Zotero.Integration.Session.prototype.cite = async function (field, addNote=false
...
 
-		if (field.type != INTEGRATION_TYPE_ITEM) {
-			throw new Zotero.Exception.Alert("integration.error.notInCitation");
+		if (field.type === INTEGRATION_TYPE_ITEM) {
+			citation = new Zotero.Integration.Citation(field, await field.unserialize(), await field.getNoteIndex());
+		}
+		else if (field.type === INTEGRATION_TYPE_BIBLIOGRAPHY) {
+			let commandName = this._app.processorName == 'Google Docs'
+				? '“Add/edit bibliography”'
+				: 'Add/Edit Bibliography';
```
dstillman, line 1522: Oh, because these are just hard-coded everywhere? Is there any chance these will be localized later?

adomasven: Not on the horizon for Word (maybe with Word.js) and LibreOffice, is probably theoretically possible for Google Docs now.

PR #5941, opened 2026-06-05
Restore -default suffix on item tree column-prefs keys

dstillman:
The earlier fix (ef7896ab5f) migrated 9.0.x's `<id>-default` prefs forward to the refactor's new `<id>` key, but that left 9.0.x and 10 beta keying differently in treePrefs.json, so switching between versions always reset 9.0.x to defaults.

Put the `-default` suffix back so both versions read and write the same key, and migrate any unsuffixed `<id>` data written by an earlier beta back to `<id>-default` on first load. The migration runs once per process and can be removed in a future release once beta users have all updated.

https://forums.zotero.org/discussion/131653/zotero-10-beta-5-reinitializes-the-columns-in-main-pane

### DIFF
--- a/chrome/content/zotero/collectionViewItemTree.jsx
+++ b/chrome/content/zotero/collectionViewItemTree.jsx
@@ -124,10 +124,7 @@ class CollectionViewItemTreeRowProvider extends ItemTreeRowProvider {
 
 		// Set ID based on visibilityGroup
 		const visibilityGroup = collectionTreeRow.visibilityGroup || 'default';
-		let treeID = "item-tree-" + this.itemTree.props.id;
-		if (visibilityGroup != 'default') {
-			treeID += "-" + visibilityGroup;
-		}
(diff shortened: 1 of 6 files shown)


dstillman:
@adomasven: Any reason not to do this?

chrome/content/zotero/collectionViewItemTree.jsx
```diff
@@ -124,10 +124,7 @@ class CollectionViewItemTreeRowProvider extends ItemTreeRowProvider {
 
 		// Set ID based on visibilityGroup
 		const visibilityGroup = collectionTreeRow.visibilityGroup || 'default';
-		let treeID = "item-tree-" + this.itemTree.props.id;
-		if (visibilityGroup != 'default') {
-			treeID += "-" + visibilityGroup;
-		}
```
adomasven, line 130: Yeah, no reason why this conditional was added. AI assisted coding fumbles.

adomasven:
Looks good.

PR #5944, opened 2026-06-07
Add update metadata from DOI for existing items

### LINKED ISSUE #1515: Update metadata for existing items
A common request, as in https://forums.zotero.org/discussion/72336/feature-request-update-or-force-recheck-of-metadata

This would probably be a context-menu option, and we'd automatically try to find an appropriate canonical item using various APIs and possibly also a modified version of recognizer-server. (The person in the linked thread is coming from Mendeley, which has both a context-menu option to "Update Details" and an icon next to each identifier, requiring the user to somehow guess which identifier would produce the best results. The identifiers are also often completely wrong, resulting in the item turning into a totally different item.)

Not sure whether we want to show the merge pane or just update the metadata automatically. Certainly we can fill in empty fields, but the user might have made other corrections (including sentence-casing the title) and it would be annoying to overwrite those.

TeoSushi1014:
Closes #1515

## Summary

- add `Zotero.ItemMetadataUpdater` backed by the existing `Zotero.Translate.Search` DOI lookup flow
- update selected regular items in place, filling empty fields by default
- offer an explicit overwrite mode that can also correct the item type
- process multi-item selections sequentially with progress, cancellation, and per-item results
- prefer a longer publisher-page abstract when DOI search metadata is missing or truncated
- leave attachments, notes, tags, collections, and relations unchanged

## Safety

The default action does not overwrite non-empty fields or existing creators. Overwrite is a separate confirmation choice.

## Testing

- `itemMetadataUpdater`: 16/16 tests passed
- DOI item context menu: 5/5 focused tests passed
- Windows/Cygwin source build passed


AbeJellinek:
Thanks, but we already have #5120. This could definitely be extracted into a plugin if you want something available before that PR makes it in.

PR #5949, opened 2026-06-10
Add by Identifier: Stay open with input, don't clear unless submitted

AbeJellinek:
- Always close on click outside or Escape
- When the window is deactivated, only close if there's no input in the box
- Don't clear input when closing unless the user submitted (pressed Enter)

I think this is less frustrating than the current behavior, which feels non-obvious to new users (and might cause them to lose work once or twice before they learn that Add by Identifier can't be relied on to stay open or remember input). And it makes it so lists of identifiers can be built within Zotero, instead of needing to build them in a text editor and paste them in.

https://forums.zotero.org/discussion/comment/513918/#Comment_513918


dstillman:
This is better, but maybe it should also persist the value within a given window even if you do blur it, if you never actually submitted it.

AbeJellinek:
Sure, try that.

PR #5952, opened 2026-06-11
Add SDT support

mrtcode:
This is initial SDT pack support mainly for Read Aloud needs, but soon also for reading mode and possibly other things in the Zotero client.

@AbeJellinek if everything is fine and we merge this, then for the reader side I made a test [branch](https://github.com/zotero/reader/tree/sdt-integration-example) that I was using for testing SDT with Read Aloud, from which you can just copy whatever makes sense into the main SDT + Read Aloud PR you were working.

Main things:
- Introduces `Zotero.SDT`
- Stores SDT as `.zotero-sdt-cache` in the attachment directory
- `document-worker` is now served from `resource://zotero/document-worker/`
- If only the processor version is stale (after processor version bump in document-worker), it serves the old compatible cache immediately and regenerates it in the background
- Bundles the `structured-document-text` submodule and builds `structured-document-text.js` locally with various utilities for interacting with SDT files. It doesn't download a prebuilt zip file like `reader` or `note-editor`, but we can add that later if needed
- Doesn't do any automatic SDT processing in the background if not specifically triggered for a concrete file


structured-document-text
AbeJellinek, line 1: Do we need to add this as a submodule, since it's already a submodule of document-worker? We already have a situation like this for `utilities` and `translate` (which itself includes `utilities`) and I think it's less than ideal there, since it makes it easy for copies of the submodule to get out of sync.

mrtcode: Yes, that makes sense.

AbeJellinek:
This looks good to me! A lot cleaner than what I had come up with on the client side.

On the reader side, I see that you went with SDT as an alternative segment source for per-view-type Read Aloud implementations. I do still like my approach where we have a single unified SDT Read Aloud implementation, since it allows views to know/care much less about Read Aloud (their main task becomes display, instead of segment calculation, sentence splitting, and so on, which lets us remove a bunch of duplicate logic across views). In your reader branch, it looks like the PDF implementation always awaits SDT processing anyway, and only falls back to non-SDT segment generation if it fails - how common is that, and is it worth maintaining view-specific implementations for?

chrome/content/zotero/xpcom/fulltext.js
```diff
@@ -381,12 +381,23 @@ Zotero.Fulltext = Zotero.FullText = new function () {
 		// If file is stored outside of Zotero, create a directory for the item
 		// in the storage directory and save the cache file there
 		if (linkMode == Zotero.Attachments.LINK_MODE_LINKED_FILE) {
-			var parentDirPath = await Zotero.Attachments.createDirectoryForItem(item);
+			var parentDirPath = Zotero.Attachments.getStorageDirectory(item).path;
+			// Create only if missing -- don't use createDirectoryForItem(),
```
AbeJellinek, line 388: IMO we should update this method so it keeps an existing directory (or make callers explicitly opt into the replace-existing behavior). Seems messy to have to avoid using `createDirectoryForItem()` to... create a directory for an item, in every situation in which the directory might already have existing contents that we don't control (which is all of them, unless the item is brand new).

chrome/content/zotero/xpcom/fulltext.js
```diff
@@ -381,12 +381,23 @@ Zotero.Fulltext = Zotero.FullText = new function () {
...
+			// which deletes and recreates the directory and would destroy
+			// other files stored there (e.g., the SDT cache)
+			await Zotero.File.createDirectoryIfMissingAsync(parentDirPath);
 		}
 		else {
 			var parentDirPath = PathUtils.parent(filePath);
 		}
 		var cacheFilePath = OS.Path.join(parentDirPath, this.fulltextCacheFile);
+		if (linkMode == Zotero.Attachments.LINK_MODE_LINKED_FILE) {
+			// createDirectoryForItem() above used to delete any previous cache
```
AbeJellinek: This just feels like a typical case of an agent writing very suboptimal code to avoid touching something two inches outside the scope it inferred from the prompt, and I don't like it!

mrtcode:
> This looks good to me! A lot cleaner than what I had come up with on the client side.
> 
> On the reader side, I see that you went with SDT as an alternative segment source for per-view-type Read Aloud implementations. I do still like my approach where we have a single unified SDT Read Aloud implementation, since it allows views to know/care much less about Read Aloud (their main task becomes display, instead of segment calculation, sentence splitting, and so on, which lets us remove a bunch of duplicate logic across views). In your reader branch, it looks like the PDF implementation always awaits SDT processing anyway, and only falls back to non-SDT segment generation if it fails - how common is that, and is it worth maintaining view-specific implementations for?

I mean, for now feel free to implement the reader side as you think is right. Later, when we add reading mode, the PDF text layer, the new text search layer for pdf-view, and more, we can figure out the best final shape for the SDT integration.

The same goes for the Zotero side. This is a hurried-up implementation mainly for Read Aloud.

chrome/content/zotero/xpcom/fulltext.js
mrtcode: Right, but that's a job for another commit. Also, `fulltext.js` will be rewritten soon to consume SDT.

mrtcode:
OK, cleaned up a little bit and made `document-worker` to produce `structured-document-text.js` utility itself, without any extra build steps in Zotero client needed.

PR #5954, opened 2026-06-15
Support multiple-collection selection

dstillman:
Adds multiple-collection selection to the collection tree.

## Selection

- Select multiple collections, saved searches, or library roots — within a library or across libraries — and see the union of their items. The selection is threaded through the pane as an array (`getCollectionTreeRows()`, `changeCollectionTreeRows()`, etc.); the item pane, tag selector, reports, and export all operate on every selected row.
- Adding items (new items and notes, drag-and-drop, the attachment dialog, import, Add by Identifier) targets every selected collection.
- Only rows that can share an items view may be combined: collections, saved searches, and library roots mix freely; multiple Recently Read rows can be combined across libraries; other special views (Trash, Duplicates, etc.) and rows from different visibility groups can't be shown together, so a mixed selection keeps only the focused row.

## Cross-library behavior

- Items are grouped by library, in collections-list order (independent of the active sort), with a section heading above each library's items — the first flush at the top, later ones offset with a gap above and a hairline below.
- The tag selector shows the union of tags (colored tags only when a single library is in scope, since colors are per-library); deleting a tag spans all selected libraries, while rename/color/split are disabled.
- In-window advanced search runs across all selected collections. Its value autocomplete is now scoped to the searched library/libraries, fixing a long-standing TODO where suggestions were drawn from all libraries. The collection and saved-search conditions are omitted for cross-library selections, since each is scoped to a single library.

## Notes

- The Connector saves to a single coherent target (the focused row), since its own tree view has no multi-selection concept yet.
- Grouping is keyed on the items list spanning multiple libraries, not the selection that produced it, so any future source of multi-library items gets the same treatment.


dstillman:
@windingwind: With multiple collections selected, the `main/library/collection` and `main/library/item` contexts for the plugin menu API still expose a single `collectionTreeRow` (the primary/first selected row), but we probably want to expose full selection as `collectionTreeRows`?

dstillman:
@yexingsha: This now allow dragging files into empty space in the items list to add them to all selected collections/libraries. Maybe worth a more dramatic effect in that case? Some sort of colored background and thicker colored border with some large overlaid text saying "Add to 3 Collections" / "Add to 3 Libraries" or something like that? (Various versions of things like this around, like the gray dashed border + large text when dragging to a Gmail message.) Not totally sure what tense the wording should be ("Adding […]", "Add […]", "Drop to add […]").

This would also make the target clearer vs. dragging on top of an individual item.

dstillman:
Actually, this is a little tricky. Right now empty space and between-items are both add-to-collection (so now add-to-all-collections) targets. The latter is necessary if the items list is full. But that means we can't really have a blocking overlay with text, since it would flicker on and off as you moved up the items list, over and between items. A border could still work, and that would solve the general problem of clarifying add-to-item vs. add-to-collection behavior, but it wouldn't actually let us communicate the more serious implications here of adding to multiple collections/libraries. Another option would be to disable dragging to individual items when you're in multi-selection mode, and then we could add clear large text, but that would introduce a functional inconsistency, which I don't love.

Another possibility would be to somehow make it clearer all the time that you're in a multi-selection — sort of like we do for batch editing — and then rely on that + the border to communicate the target. But the seamless composability is part of the point here, so we don't want to do anything too visually disruptive if you just Cmd-click a second collection.

windingwind:
> With multiple collections selected, the `main/library/collection` and `main/library/item` contexts for the plugin menu API still expose a single `collectionTreeRow` (the primary/first selected row), but we probably want to expose full selection as `collectionTreeRows`

I think so. For compatibility, we can provide a new arg as `collectionTreeRows` (for single selection, as `[collectionTreeRow]`) while keeping the old one for the primary/first.

dstillman:
`collectionTreeRows` added

AbeJellinek:
Command-clicking on an already selected collection tree row "deselects" it visually, but its items remain displayed in the item tree. Probably shouldn't be possible to have no selection.

It's currently possible to drag My Library (or Unfiled, etc.) to its own trash. (Not sure if new.)

dstillman:
> It's currently possible to drag My Library (or Unfiled, etc.) to its own trash. (Not sure if new.)

Not new, apparently!

AbeJellinek:
When you select a saved search, Cmd-click on a collection, then press Enter, the collection name opens for editing. When two collections are selected and you press Enter, nothing happens.

It's also possible to open the Edit Saved Search interface with multiple searches selected by pressing Enter.

AbeJellinek:
> Not new, apparently!

At least it doesn't do anything!

AbeJellinek:
1. Select a saved search and press Enter to start editing it
2. Cmd-click another saved search -> "Do you want to save changes you made to this saved search?" dialog shows
3. Try to cancel -> dialog immediately reopens

AbeJellinek:
Various context menu options and their confirmation dialogs need plural variants.

AbeJellinek:
On every reload, each selected collection/feed creates a separate `tmpSearchResults` table with a separate itemID index, and then it runs individual `SELECT DISTINCT tagID || ':' || type` queries on each one. I don't have a huge test library, but I imagine this could be pretty slow. Not sure how easy it would be to optimize.

AbeJellinek:
When I select across multiple libraries, the library heading in the item tree is right-clickable. When an item is selected, it shows context menu options for that item. When nothing is selected, it shows Change Parent Item and Add to Collection options that I'm not going to try. (The My Library text/icon also looks slightly offset toward the top, and the crescent-ish tapered border on a rounded corner looks kind of odd to me.)

AbeJellinek:
The macOS focus row mechanism (Cmd-up/down arrow, Space to toggle selection) seems to be supported, but the dotted line showing the currently selected row is missing.

AbeJellinek:
When I select multiple rows and then filter the list, very weird things happen:

image

I think the selection indices are being preserved in the filtered list, rather than the actual set of selected rows.

AbeJellinek:
Not totally clear why it's possible to select a library and its child collections/searches - is there a use case for that?

dstillman:
> Not totally clear why it's possible to select a library and its child collections/searches - is there a use case for that?

No reason, other than that they're not technically incompatible. But that can just behave the same as other incompatible combinations (i.e., ignore the Cmd and select the new row).

chrome/content/zotero/xpcom/translation/translate_item.js
```diff
@@ -1086,19 +1086,19 @@ Zotero.Translate.ItemGetter.prototype = {
 	setCollection: function (collection, getChildCollections) {
 		// get items in this collection
 		var items = new Set(collection.getChildItems());
-		
+
```
AbeJellinek: All the changes in this file are just whitespace

chrome/content/zotero/zoteroPane.xhtml
```diff
@@ -1048,9 +1048,14 @@
 					var libraryID = ZoteroPane.tagSelector.libraryID;
 					var library = Zotero.Libraries.get(libraryID);
 					var enabled = false;
-					if (library.editable) {
-						if ((await Zotero.Tags.getAutomaticInLibrary(libraryID)).length) {
-							enabled = true;
+					// 'Delete Automatic Tags in This Library' is per-library, so disable it
+					// when the selection spans multiple libraries (nested ifs to avoid '&amp;&amp;'
```
AbeJellinek: Can we pull this out? Long past being appropriate for an inline handler

dstillman:
Thanks! All addressed, I believe.

For strings, I just disabled heterogeneous deletes, and I simplified new strings to mostly drop nouns (except for "Delete Collection[s]" / "Delete Collection[s] and Items…", which we've considered doing for a long time anyway.

yexingsha:
I do think we need a persistent indicator for multi-collection selection. If drag-and-drop into multi-collection selection warrants a more dramatic effect, then the same logic should apply for adding items via toolbar buttons and menu. Also, I feel like it's easy to forget that you have multiple collections selected after a while, especially when the collection tree is long and only one of the selected collections is shown on screen. If the selected collections are in different libraries, then people might be confused as to why the toolbar buttons and menu actions are disabled.

As to where to put this indicator, one option is in the toolbar next to the add buttons, but it doesn't really work when the window is narrow. Another is on top of the collection tree, which is more logical and plays better with narrow windows, but a bit far from where the actions are triggered and thus more easily overlooked. So I'm leaning toward the item tree itself: we are already showing section headers for cross-library selection, so I think a section header that reads "2 collections selected" makes sense as well. We can make these headers sticky on top of their own section (see Finder grouped list view, but without putting the section header into the column header) so that they are always visible, which also helps with e.g. cross-library search.

With the sticky section headers established, I don't think we need too much of a dramatic effect for drag-and-drop. A hint in the top right corner should be good enough:

single-library

multiple-libraries

This can also work in Recently Read, Saved Search, etc. to explain why drag-and-drop doesn't create new items there.

Some other things to note:

- Section headers styling:
	- Ideally only have hairlines on scroll
	- No rounded corners so the hairlines don't curve up
	- Should always use `var(--material-background)`, and the following items are striped accordingly
	- Increase the white space above it to one row height
	- Gap after icon should be 4px instead of 6
	- The headers say e.g. "2 Saved Searches selected", "My Library", "My Library (Recently Read)", etc. depending on the selection.
- Some old issues with drag-and-drop:
	- Contents of item row shifting to the right when the drop indicator is on top of it
	- When an item row is highlighted as the drop target, the twisty should change color same as the text, and the icons should switch to their dark/light versions.

dstillman:
OK! Try that out. Didn't do the hints yet, but we can do those after landing this.

(Misaligned Title column header text is an unrelated earlier regression.)

**Two collections, same library:**

<p>Screenshot 2026-06-17 at 3 26 56 PM</p>

**Collection and search, same library:**

<p>Screenshot 2026-06-17 at 3 27 13 PM</p>

**Two library roots:**

<p>Screenshot 2026-06-17 at 3 27 21 PM</p>

**Two library roots, scrolled up:**

<p>Screenshot 2026-06-17 at 3 27 40 PM</p>

yexingsha:
The section headers are looking really good! Just a few issues:

- After selecting a few collection, when an item is selected, the top section header gets a dotted border. If I scroll the item tree before selecting an item, the dotted border shows up then immediately goes away.
	Screenshot 2026-06-17 at 17 29 42

- The last selected item scrolls on top of the section header instead of underneath:
	Screenshot 2026-06-17 at 17 30 26

- When scrolling up at the very top, a second section title gets revealed:
	Screenshot 2026-06-17 at 17 31 22

dstillman:
All fixed

dstillman:
Merged

PR #5955, opened 2026-06-15
Fix keyboard navigation through advanced search buttons and deck

abaevbog:
- Make Advanced Search button focusable, include it into the tab sequence
- Tab: quick-search textbox -> Advanced Search button -> item tree
- Shift-Tab: Advanced Search button -> quick-search textbox
- Shift-Tab: item tree -> Advanced Search button (when the deck is closed)
- Shift-Tab: Advanced Search collapse button -> New Item button
- Fix Shift-Tab from the item tree doing nothing when the Advanced Search deck is open; now moves focus to the deck's last focusable element

### DIFF
--- a/chrome/content/zotero/elements/quickSearchTextbox.js
+++ b/chrome/content/zotero/elements/quickSearchTextbox.js
@@ -124,7 +124,12 @@
 			});
 			this.querySelector('.advanced-close-button').addEventListener('command', (event) => {
 				event.stopPropagation();
+				// if activated via the keyboard, move focus to the Advanced Search button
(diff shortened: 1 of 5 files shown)


dstillman:
If you Shift-Tab to the Close button and press Space, focus gets lost. Should probably go back to the filter button?

abaevbog:
Good call, just added.

I also added a tooltip to the collapse/expand and X icons and fixed another edge case of focus not moving on shift-tab from itemTree when advanced search is active but collapsed.

dstillman:
Merged in e6f82dabfd52cb39e6e902f8b7e0f398b7f3c812 and 4d38970ef086c18b40aee0e04ec1ccfbb6193034

Issue #1655, opened 2019-03-04
Support more items-list features in Advanced Search results


dstillman:
E.g., https://github.com/zotero/zotero/issues/322, but also context menu, delete, etc.

01baftb:
Unsure if this is relevant: An ability to filter or search for items with attachments as "linked" or "stored copy".

dstillman:
I was going to try to get to that soon, but that'll be separate. Issue created: https://github.com/zotero/zotero/issues/1657

dstillman:
Superseded by #5658

PR #5962, opened 2026-06-18
Add condition groups, cross-level result levels, and quick-search prefill to Advanced Search

dstillman:
Adds nested condition groups to saved searches and Advanced Search, allowing searches to combine conditions with arbitrary nesting and per-group match modes, such as `title contains "foo" AND (tag is "x" OR tag is "y")`.

A search can also now be given a result level — top-level item, attachment, note, or annotation — and each condition will be mapped to that level, so one search can mix conditions that match at different levels of the item hierarchy (e.g., a top-level item with a given author *and* a red annotation on one of its PDFs). The Advanced Search UI exposes this through a top-level result-level menu, per-group controls for binding descendant conditions to the same item, and warnings for combinations that would never produce results.

Opening Advanced Search from a non-empty quick search now seeds it with editable conditions matching the current mode. The quick-search modes are now represented as ordinary grouped searches rather than being special internal cases. (This also addresses the ancient feature request of wanting to pin quick searches across collections. Now you can just click the filter button and switch collections, and the advanced search persists.)

This removes various legacy features, including "Show top-level items" (result level = top-level item — migrated on save), the awful "Include parent and child items" hack (no result-level equivalent, so it keeps working and round-trips on existing searches but isn't offered on new ones), the unused `required` flag, and various hacks (the tag→annotation-parent map, the special quick-search handling).

## TODO before merge

- [ ] Bump the schema version on all three clients (desktop, iOS, Android)
- [x] Update the dataserver to return `invalidProp` for searches that contain group markers (`groupStart`/`groupEnd`) to clients below the new schema version. iOS/Android never run searches and tolerate unknown conditions, but without a bump they would show a spurious "saved in a newer version" error for grouped searches they don't even display.

### DIFF
--- a/chrome/content/zotero/elements/advancedSearchPane.js
+++ b/chrome/content/zotero/elements/advancedSearchPane.js
@@ -127,6 +127,9 @@
 			}
 			else {
 				this._search = new Zotero.Search();
+				// Default a fresh search to top-level items, so a condition on a child
+				// (e.g. attachment content) maps up to its item without any grouping
+				this._search.addCondition('resultLevel', 'item');
 				this._search.addCondition('title', 'contains', '');
 			}
 			this._searchElem.search = this._search;
--- a/chrome/content/zotero/elements/advancedSearchPane.js
+++ b/chrome/content/zotero/elements/advancedSearchPane.js
@@ -187,15 +187,20 @@
 			await ZoteroPane.setSavedSearchEditorState('closed');
 		}
 
-		async submit() {
+		async submit({ focusResults = true } = {}) {
 			if (this.type === 'saved') {
 				throw new Error('submit() is unsupported for saved search');
 			}
-			
+
 			this._searchElem.updateSearch();
 			this._active = true;
 			await ZoteroPane.itemsView.setFilter('advanced-search', this._search);
-			ZoteroPane.itemsView.focus();
+			// Running the search normally moves focus to the results, but a caller that
+			// wants to keep focus in the search builder (e.g. quick search prefill) can
+			// opt out
+			if (focusResults) {
+				ZoteroPane.itemsView.focus();
+			}
 		}
 		
 		async clear() {
--- a/chrome/content/zotero/elements/quickSearchTextbox.js
+++ b/chrome/content/zotero/elements/quickSearchTextbox.js
@@ -111,7 +111,15 @@
 				advancedButton.addEventListener('command', (event) => {
 					// Don't trigger a quick search via the oncommand handler
 					event.stopPropagation();
-					ZoteroPane.toggleAdvancedSearchState('open');
+					// If there's text in the field, seed the Advanced Search with it,
+					// reproducing the current quick search mode as editable conditions.
+					let mode = Zotero.Prefs.get('search.quicksearch-mode');
+					if (this.value) {
+						ZoteroPane.openAdvancedSearchFromQuickSearch(this.value, mode);
+					}
+					else {
+						ZoteroPane.toggleAdvancedSearchState('open');
+					}
 				});
 				wrapper.appendChild(advancedButton);
 			}
--- a/chrome/content/zotero/elements/zoteroSearch.js
+++ b/chrome/content/zotero/elements/zoteroSearch.js
@@ -308,8 +308,11 @@
 					</menulist>
 					<label class="join-mode-suffix" value="&zotero.search.joinMode.suffix;"/>
 					<spacer flex="1"/>
-					<toolbarbutton class="remove-group zotero-clicky zotero-clicky-minus" tabindex="0" hidden="true" data-l10n-id="advanced-search-remove-group-btn" onclick="this.closest('search-condition-group').onRemoveGroupClicked()"/>
-					<toolbarbutton class="add-condition zotero-clicky zotero-clicky-plus" tabindex="0" data-l10n-id="advanced-search-add-btn" onclick="this.closest('search-condition-group').onAddSiblingClicked()"/>
+					<hbox class="group-actions">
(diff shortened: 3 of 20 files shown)


dstillman:
@yexingsha: Please have a look at the condition-group UI.

A bunch of different ways to do condition groups (did anyone know you could Option-click the + in the Finder search builder to created nested groups?), but I went with a "( )" button that turns that row into a group and lets you add more conditions, which maybe is most compatible with how people will reason about this?

<p>Screenshot 2026-06-18 at 4 12 34 PM</p>

I increased the vertical spacing between conditions to try to convey that the group in the middle is essentially its own condition of the root group, but that may or may not be clear enough, and it does space out non-nested conditions a bit. Finder goes with horizontal lines instead.

I don't love the mess of unaligned buttons on the right side, but not sure what we can do there given the nested-group padding. At the very least, I assume the temporary "( )" button is just a different width from the other two and we can at least have all the buttons at a given level line up.

dstillman:
Switched to horizontal lines, which cleans things up and lets us align all the buttons on the right:

dstillman:
@yexingsha: A bunch of things for you to look at here. In particular, see the hint when (say) adding two top-level Annotation Comment conditions at the top level (hint to match to the same annotation) and the warning when adding disjoint conditions (e.g., find notes with annotation comments). Those currently have blue and red backgrounds, and I was thinking it could be neat to flag the conditions it's referring to with the same color. (Technically the disjoint warning can appear multiple times for different sets of conditions, but probably don't need to worry about that.) But not a huge priority compared to any general tweaking.

AbeJellinek:
When I switch a condition to Attachment File Type, it gets a menulist *and* a textbox:

image

dstillman:
Fixed

yexingsha:
The cross-level condition matching is great, though admittedly it took me a bit to figure out that it can only match parent or descendant, never horizontally under the same top-level item, and "any items" does not do cross-level condition matching at all. Am I understanding this correctly? 

All that is to say, I feel like quite a few people will be confused as to why they could target different levels in some occasions but not others. I suppose the warnings do communicate some of this information, but they are context-specific and scattered. I think we should definitely put a link to documentation somewhere, either in the warnings or as a dedicated `?` button. As to the warnings, maybe putting more emphasis on how things are supposed to work might help? For example, "'annotations' can only take conditions applied to annotations, their parent attachment, or their top-level item"; "'any items' requires conditions to match the same item".

Binding descendant conditions works pretty well, though I'm curious why the hint/auto-grouping only triggers when there's value entered in both input fields. Wouldn't it make more sense for them to trigger on initial condition creation?

Here are some bracket icons. I feel like the square ones look better, even though the round ones work better semantically. And I feel like we need an "Ungroup" button for group rows, but maybe that's much rarer in practice than in testing:
bracket
bracket-remove
square-bracket
square-bracket-remove
curly-bracket
curly-bracket-remove

Other UI tweaks:
- The button in hints contains an nonexistent image, pushing the text off-center.
- The text size in warnings should match the hints.
- When conditions target descendant items in a top-level item search, I think we should expand the top-level items to show the matching child items in gray, so that people can see what exactly is matching.
- The first line ("Find … matching … ") doesn't need the extra left padding, since it's at the highest level of all conditions. Its top and bottom paddings are also no longer necessary.
- I assume the pop-up buttons have smaller top margin than bottom margin to match text baseline? But this causes them to sit higher in condition rows and are not vertically aligned with the text fields and the buttons.
- Shrink the text fields so that they match the height and margins of the pop-up buttons, preventing the rows from shifting when switching condition types. The text fields also have an extra left margin that causes a bigger gap.
- The condition gap can be reduced to 6px to match the row padding.
- Fix for narrow window: overflow "Advanced Search" so that both buttons remain visible, wrap the first line, and add a horizontal scroll for the condition area.
- Fix for Stacked View: there should be a vertical line between Close Advanced Search and Toggle Item Pane, aligned with the vertical line above (between tab list and sync).

Bugs(?):
- Searching for annotations with the Title condition always returns nothing. The same values works fine if I apply the condition Title, Creator, Year.
- "Attachment Last Read" doesn't seem to work.
- Condition menus sometimes cannot be interacted with via cursor. I haven't figured out how to trigger this consistently, but it happens most often with the `more-conditions-menu` when I first open the Advanced Search panel, or when I switch focus back from a different window.
- I feel like there is some issue with the focus on text fields. For example, if I edit the value in a "contains" text field, then switch the condition type, the edited value doesn't stick. The only way for the value to stick is to run the search, and then any new edits won't stick either.

Also, with the cross-level condition matching, we can tweak the conditions a bit:
- Maybe I'm missing something here, but isn't "Child Note" now obsolete?
- Now that annotations can be targeted, I think it makes more sense to move "Annotation Comment" and "Annotation Text" into the top level menu. Same goes for if/when we have "Annotation Type" and "Annotation Color".
- This is a bit of a leap, but I wonder if we should separate some conditions into two tiers, somewhat like "Kind is Image - PNG" in Finder. The first tier would contain a more manageable list: Any Field, Attachment, Annotation, Collection, Date Added, Date Modified, Item Info, Item Type, Note, Tag, Title/Creator/Year. Some of these, like Collection, will not have a secondary tier; others like Item Info will show a second popup button with a menu list of metadata fields, with the most commonly used on top. The default new condition would be Item Info - Title. This makes it clearer which level you are targeting with each condition, and also make our huge list of conditions more readable and more expandable in the future. Though obviously this require more steps, and the fields will have to become misaligned to save space. I guess this will be a downgrade for people who are comfortable navigating with keyboards, and an upgrade for people who are not.

iagogv3:
This seems promising. Thanks! (would it be closing #1655 #1657 #5875 ?)

PR #5964, opened 2026-06-19
Fix stale `itemType` in `_previousData`

tnajdek:
This PR fixes a problem where, on subsequent item type changes, a stale `itemType` is reported by the notifier, and as a result, the filename is not updated by the auto file renaming.


dstillman:
Changed this to be a direct data-layer test, since it's not actually related to attachments

PR #5966, opened 2026-06-20
feat(integration): add narrative "Author (year)" citations

dawidstone5:
## What

Adds a per-citation **narrative mode** to the citation dialog. The author
becomes the subject of the sentence — "Dawid and Stone (2026)" — rather than
sitting inside parentheses as "(Dawid & Stone, 2026)".

This reuses citeproc-js's existing **composite** cluster mode, so the work is
mostly UI wiring plus one engine fix for the textual "and".

## Changes

**Citation dialog UI**
- `citationDialog.xhtml` — "Author in text" checkbox (`#narrativeMode`), scoped
  to `data-dialog-type="citation"`.
- `citationDialog.js` — the toggle sets or clears
  `citation.properties.mode = "composite"`. Hidden and disabled unless items are
  being cited; refreshes the preview and bubbles on change.
- `integration.ftl` — string `integration-citationDialog-settings-narrative`.

**Persistence**
- `integration.js` — `Citation.toJSON()` now keeps `mode` and `infix`, so a
  narrative citation survives the round-trip to the document field.

**citeproc fix**
- `citeproc.js` — for a composite/author-only render under a style with
  `and="symbol"`, citeproc keeps the parenthetical "&" but the narrative form
  needs the textual "and" ("Dawid and Stone (2026)", per APA). The fix
  pre-builds an `and_narrative` join and applies it only on author-only renders,
  leaving the parenthetical output unchanged.

**Tests**
- `integrationTest.js` — narrative output and the textual-"and" behavior.

## Notes

The branch also carries a merge of current `main` and a revert of unrelated
`package-lock.json` peer-flag churn, leaving the net diff logic-only.

### DIFF
--- a/chrome/content/zotero/integration/citationDialog.js
+++ b/chrome/content/zotero/integration/citationDialog.js
@@ -259,6 +259,10 @@ async function setDialogType(type) {
 	_id("bubble-input").sortable = DIALOG_STATE.isCitingItems();
 	_id("keepSorted").disabled = !io.sortable || !DIALOG_STATE.isCitingItems();
 	_id("keepSorted").checked = !_id("keepSorted").disabled && !io.citation.properties.unsorted;
+	// Narrative ("Author (year)") mode is only meaningful when citing items
+	_id("narrativeMode").disabled = !DIALOG_STATE.isCitingItems();
+	_id("narrativeMode").parentElement.hidden = _id("narrativeMode").disabled;
+	_id("narrativeMode").checked = io.citation.properties.mode === "composite";
 	if (DIALOG_STATE.isCitingItems()) {
 		_id("keepSorted").disabled = !io.sortable;
 		_id("keepSorted").parentElement.hidden = !io.sortable;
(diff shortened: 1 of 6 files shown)


dawidstone5:
**Update (resolved in this PR):** the two-author APA conjunction now renders with the textual "and" — `Smith and Jones (2020)` — not `&`. The `citeproc.js` change pre-builds an `and_narrative` join and applies it only on author-only (narrative/composite) renders, so the parenthetical form keeps `&` while the narrative form uses "and", per APA. This no longer depends on CSL 1.0.2 `cs:intext` adoption; citeproc-js's existing `composite` mode is sufficient.

(Earlier comments on this thread described the `&` behavior as an unfixed limitation — that was accurate before commit 2911f3ec and is now superseded.)

adomasven:
We're working on this, but there are additional considerations. See https://github.com/zotero/zotero/issues/1580

Particularly, this still requires some citing UI and word processor plugin upgrades before we can push it to our users in a state that we are happy with.

dawidstone5:
Does that mean it ain't viable?

adomasven:
This is viable and in active development.

dawidstone5:
Alright, thank you. I had created a version that works and tested locally but since you're working on it. I really can't wait to integrate it in my workflow.
Once again, thank you for this amazing tool.

PR #5967, opened 2026-06-22
Add local-cache versioning for Read Aloud audio

dstillman:
Lets the server force the Read Aloud local audio cache to update, instead of a given voice+text rendering being frozen on a client forever. The trigger could be anything -- improved models, pronunciation changes, or recovering from a server outage that cached silent audio.

- Cache keys now include the voice's `cacheVersion`; a server-side bump changes the key, so old entries miss and correct audio is re-fetched
- Stale entries (versions no longer offered by the server) are pruned once per session
- Responses sent with `Cache-Control: no-store` are not cached

Depends on zotero/reader#219, which feeds `cacheVersion` through to the voice config and surfaces `no-store`.

### DIFF
--- a/chrome/content/zotero/xpcom/reader.js
+++ b/chrome/content/zotero/xpcom/reader.js
@@ -1758,10 +1758,7 @@ class ReaderInstance {
 	// cacheVersion is included so that a server-side version bump changes the
 	// key, causing old entries to miss and the correct audio to be re-fetched.
 	_getReadAloudCacheURL(segment, voice) {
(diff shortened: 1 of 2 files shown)


AbeJellinek:
In the zotero/reader PR, I don't think we totally need cache pruning in `index.dev.js` ("once per session" is... pretty frequent in an auto-reloading dev server, and orphaned cache entries are harmless and will eventually be deleted by the browser), but otherwise this looks good.

dstillman:
Removed

PR #5977, opened 2026-06-25
Item/collection tree: Resize using ResizeObserver, not debounce

AbeJellinek:
And expose `ResizeObserver` in `require()`'d scopes.

Fixes the noticeable lag after collapsing Advanced Search before new rows appear at the bottom, as well as other, less common cases like collapsing the tag selector in the collection tree and collapsing the item pane in Stacked mode.

This may theoretically affect performance a bit during window resize, since we no longer wait 200ms to rerender (and instead do it once per frame via `ResizeObserver`), but to my eyes it looks a *lot* less glitchy and doesn't cause any noticeable lag.

This wasn't a major issue before, since window resizing is relatively uncommon and other interface elements rarely changed the item and collection trees' height, but now that Advanced Search does, I think the tradeoff is worth it.

### DIFF
--- a/chrome/content/zotero/components/virtualized-table.jsx
+++ b/chrome/content/zotero/components/virtualized-table.jsx
@@ -1102,9 +1102,8 @@ class VirtualizedTable extends React.Component {
 		this._setXulTooltip();
 
 		this._topDiv.style.setProperty("--first-column-extra-width", `${this.firstColumnExtraWidth}px`);
-		window.addEventListener("resize", () => {
-			this._debouncedRerender();
-		});
+		this._resizeObserver = new ResizeObserver(() => this.rerender());
+		this._resizeObserver.observe(this._jsWindow.targetElement);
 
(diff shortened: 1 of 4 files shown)


adomasven:
Worth testing on an older laptop or something. The 200ms wait was added specifically because rerendering on every frame caused lag on my older work machine, so an average laptop is going to fare poorly for window resizes. A specific alternative for advanced search would be to not downsize the tree when it's open.

AbeJellinek:
There have been a lot of improvements on the Firefox end in that time, so maybe things are better now. I don't have a good test device for this, though.

adomasven:
I tested in Windows VM with power saving and normal profiles and couldn't tell a meaningfull difference when resizing. If I cap the frequency of my CPUs low enough, it is laggy both on main and with this, and with middling capping it's smooth. On native Ubuntu even in extreme powersaving mode it's fast.

Can be merged

PR #5980, opened 2026-06-29
Implement MacOS Spotlight Support

### LINKED ISSUE #5859: MacOS Spotlight Intergration
Ref: https://github.com/windingwind/zotero-spotlight-search

Things to discuss:

1. Currently, the attachments are already indexed by Spotlight. Once we add support for the library search, they will appear twice. How do we handle that? Users can type Zotero > Tab and then search only within Zotero, but direct searching would anyway include the attachments.
2. Multiple libraries - in the plugin, users can decide what to display in the subtitle of the search entries, including library name, in a similar manner as the attachment rename template.
3. Multiple profiles - in the plugin, we replace the results completely once another profile with a different database comes to active. I think that should be fine.

windingwind:
Use a in-app dylib for indexing control
Patch the xul lib for spotlight action handling
Resolve #5859

The XUL custom build: https://1drv.ms/u/c/174388539e0b05be/IQBQ2mF9C-kjTZGJLU6UgwcVAVFEtPc-psJ4o_F8PBW73KI?e=j4C0qy

Or you can apply the patch mozilla-140.patch and build yourself.

image

image

### DIFF
--- a/app/build.sh
+++ b/app/build.sh
@@ -727,7 +727,20 @@ if [ $BUILD_MAC == 1 ]; then
 	mkdir "$CONTENTSDIR/Resources/integration"
 	cp -RH "$CALLDIR/modules/zotero-libreoffice-integration/install" "$CONTENTSDIR/Resources/integration/libreoffice"
 	cp -RH "$CALLDIR/modules/zotero-word-for-mac-integration/install" "$CONTENTSDIR/Resources/integration/word-for-mac"
-	
+
+	# Build the Core Spotlight bridge dylib
+	if command -v clang >/dev/null 2>&1; then
+		clang -dynamiclib -fobjc-arc -O2 -mmacosx-version-min=11.0 \
+			-arch x86_64 -arch arm64 \
+			-framework Foundation -framework AppKit -framework CoreServices \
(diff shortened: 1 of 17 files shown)


windingwind:
Pipeline:

Zotero update indexing
-> User sees them in the Spotlight search
-> Clicks
-> Launch the executable providing the index with `zotero://spotlight`  (because the url can survive the profile picker at cold start)
-> Zotero starts and either selects in library or opens the best attachment.

Some more details:

1. Multi-instance (stable, beta, source, etc.): each release channel has its own spotlight entries and is managed separately.
2. Multi-profile: only one profile gets indexed. In another profile, the user can click to switch to the current profile.
3. Indexing update: on item modified (single item) or library added/removed (batch for the whole library). The full reindexing finishes in seconds for a multi-library large profile.
4. Indexed items: top-level items. Each item has its fields, notes, attachment metadata, and attachment full-text (if enabled) indexed.
5. Cold start, cold start with profile picker, or start when main window closed: all tested.
6. For ignoring the data dir, we open the system settings page and data dir for the user, but not sure how that would be straightforward in UI without some screenshot-like guidance (maybe put a link to the doc?)

Issue #5978, opened 2026-06-26
Advanced search condition-groups/result-level follow-up


dstillman:
_Originally posted by @yexingsha in https://github.com/zotero/zotero/issues/5962#issuecomment-4807253243_

The cross-level condition matching is great, though admittedly it took me a bit to figure out that it can only match parent or descendant, never horizontally under the same top-level item, and "any items" does not do cross-level condition matching at all. Am I understanding this correctly? 

- [ ] All that is to say, I feel like quite a few people will be confused as to why they could target different levels in some occasions but not others. I suppose the warnings do communicate some of this information, but they are context-specific and scattered. I think we should definitely put a link to documentation somewhere, either in the warnings or as a dedicated `?` button. As to the warnings, maybe putting more emphasis on how things are supposed to work might help? For example, "'annotations' can only take conditions applied to annotations, their parent attachment, or their top-level item"; "'any items' requires conditions to match the same item".

- [x] Binding descendant conditions works pretty well, though I'm curious why the hint/auto-grouping only triggers when there's value entered in both input fields. Wouldn't it make more sense for them to trigger on initial condition creation?

- [x] Here are some bracket icons. I feel like the square ones look better, even though the round ones work better semantically. And I feel like we need an "Ungroup" button for group rows, but maybe that's much rarer in practice than in testing:
<img width="16" height="16" alt="bracket" src="https://github.com/user-attachments/assets/7f246f74-0aad-4581-a791-ef72bb318f12" />
<img width="16" height="16" alt="bracket-remove" src="https://github.com/user-attachments/assets/40f89f91-6725-4bd2-90aa-a1d9d9ead691" />
<img width="16" height="16" alt="square-bracket" src="https://github.com/user-attachments/assets/e677c30e-1faf-44b6-ac95-e0b9a68f6e5d" />
<img width="16" height="16" alt="square-bracket-remove" src="https://github.com/user-attachments/assets/84265db0-9e68-4864-bbdd-133cc1577676" />
<img width="16" height="16" alt="curly-bracket" src="https://github.com/user-attachments/assets/26278db7-e392-4312-ba75-d859e733fa77" />
<img width="16" height="16" alt="curly-bracket-remove" src="https://github.com/user-attachments/assets/081a3be8-f19b-47a0-8a7d-0a9da4cfa510" />

Other UI tweaks:
- [x] The button in hints contains an nonexistent image, pushing the text off-center.
- [x] The text size in warnings should match the hints.
[...]

PR #5981, opened 2026-06-30
Group attachment and annotation conditions into Advanced Search submenus

dstillman:
Shows just the shorter names in the submenu:

<p> </p>

The full condition name is still shown when selected (and you can use that via find-as-you-type if you tab to the condition menu):

<p></p>

Addresses https://github.com/zotero/zotero/issues/5978


dstillman:
@yexingsha: Certainly an improvement. Not sure if showing the short condition names is confusing, but it's pretty repetitive to show the prefix for each one.

dstillman:
Merging for now. We can reconsider if we think the prefix-less names aren't clear.

yexingsha:
Feels very clear to me and not at all confusing.

PR #5983, opened 2026-07-01
Fix tree multi-selection collapse/expand behaviour

### LINKED ISSUE #5974: Collection/item tree: Selection shifts when rows above are expanded/collapsed
1. In the collection tree, select the last feed row
2. Use Cmd-Up arrow to move the focus row to a group library header
3. Press the left arrow key to collapse it -> selection drops off the end of the screen

Or:
1. In the collection tree, select the first feed row
2. Use Cmd-Up arrow to move the focus row to a group library header
3. Press the left arrow key to collapse it -> selection moves down by however many rows were removed

Or:
1. In the collection tree, collapse a group library header
2. Select the first feed row
3. Use Cmd-Up arrow to move the focus row to the group library header
4. Press the right arrow key to expand it -> selection moves up by however many rows were added

This behavior (selection not being preserved) seems to happen whenever the focus row has been moved away from the selection, even when the row being expanded/collapsed is not the same as the focus row. For example, the same thing happens after selecting a feed, moving focus to My Publications, and using the mouse to toggle a group.

https://github.com/user-attachments/assets/d6ee62ef-b9ba-46e6-96ba-9701e7a50f32

adomasven:
Closes #5974.

Additional fixes for broken item tree behaviour when multiple items are selected, and changing focus with ctrl/cmd-arrow keys.

Aligned Collection Tree/Virtualized Tree collapse/expand behaviour when mutliple containers are selected, one of them is focused, and arrow key left-right is pressed, to the behaviour in Item Tree - now all of them are collapsed/expanded.

### DIFF
--- a/chrome/content/zotero/collectionTree.jsx
+++ b/chrome/content/zotero/collectionTree.jsx
@@ -1266,9 +1266,7 @@ var CollectionTree = class CollectionTree extends LibraryTree {
 		if (treeRow.isLibrary(true) || treeRow.isCollection() || treeRow.isFeeds()) {
 			count = await this._expandRow(this._rows, index, true);
 		}
-		if (this.selection.focused > index) {
-			this.selection.select(this.selection.focused + count);
-		}
+		this.selection.adjustForRowInsertion(index, count);
 		this.selection.selectEventsSuppressed = false;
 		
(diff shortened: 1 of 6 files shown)


dstillman:
Rebased and pushed some regressions from the original fix

PR #5984, opened 2026-07-02
Local semantic search on item abstracts

abaevbog:
- added environment to run embedding models locally (transformers.js, ONNX Runtime WASM binary, etc.) in a worker
- added local `itemEmbeddings` table to store embeddings locally
- in advanced preferences, one can select two options for semantic search model: english and multilingual. English model (bge-small-en-v1.5) is better for english-only corpus but multilingual (multilingual-e5-small) is necessary to handle abstracts and queries with any other language than english. We can add more models as needed.
- when the model is selected, `Zotero.Embeddings.download` will download the model  and store it locally
- then, `Zotero.Embeddings.Indexing` will start a process to index all items with abstracts. It happens in batches and takes some time. The progress will appear in the advanced preferences. Embeddings are inserted into itemEmbeddings SQL table. For now, the table is local only, no syncing is involved.
- `Zotero.Embeddings.Indexing` also handles re-indexing abstracts when updated
- when embedding model pref is set to "Disabled", the model is deleted and embeddings table is cleared.
- when an embedding model is selected, quick search dropdown has a new "Similarity" mode, which will run semantic search on the current scope of items.
- semantic search happens in-memory. On my test 5K library it's fast, but we could consider sqlite-vec extension if needed.

Open questions:

- Semantic search does not clearly define what counts as "relevant" and what is "not relevant". It will vary depending on the library and query. So we cannot have semantic search filter out items, the way it is done via SQL. Semantic search essentially sorts items by relevance - but sorting is being handled by the itemTree based on column selection. I think the best solution is to show top K semantic matches in the current scope, and allow the user to select what K is depending on circumstances. I added a new "Similarity" mode to quicksearch when semantic search is enabled. It has a dropdown to select how many top relevant items to keep (top 5 - top 100). I think it works functionally, though the UI may need some tweaks. 
- As mentioned, the `itemEmbeddings` is just a local SQL table that does not involve syncing. Should it?
- Currently, the models are downloaded from huggingface. We may want to host them ourselves.

---

Step 1 - enable and wait for indexing:

https://github.com/user-attachments/assets/311caab4-c0e6-4aca-96f1-23d84875db3d

Step 2 - demo:

https://github.com/user-attachments/assets/5d7beffa-6028-4f73-90bc-f74496d7b1d4


dstillman:
Superseded by https://github.com/zotero/zotero/pull/6012

PR #5987, opened 2026-07-02
Fix "redo" command not enabled in the note editor, close #3398

### LINKED ISSUE #3398: Erratic Undo/Redo menu options behavior in note editor
https://forums.zotero.org/discussion/107521/zotero-7-problems-with-undo-redo-in-notes

tnajdek:
While working on global undo/redo I've noticed that "redo" option is never available when editing a note (#3398). The fix for this has two parts:

1. Expose `doUndo`/`doRedo`/`canUndo`/`canRedo` in the note editor
2. Detect if focus is inside a note editor,  query `canUndo`/`canRedo and update edit menu entries 
3. Forward edit menu commands via the newly-exposed API 

Depends on https://github.com/zotero/note-editor/pull/77

### DIFF
--- a/chrome/content/zotero/xpcom/editorInstance.js
+++ b/chrome/content/zotero/xpcom/editorInstance.js
@@ -53,6 +53,7 @@ const DOWNLOADED_IMAGE_TYPE = [
 class EditorInstance {
 	constructor() {
 		this.instanceID = Zotero.Utilities.randomString();
+		this._undoRedoController = null;
 	}
 
 	get itemID() {
@@ -242,6 +243,7 @@ class EditorInstance {
 	}
 
 	async uninit() {
+		this._unregisterUndoRedoController();
 		this._prefObserverIDs.forEach(id => Zotero.Prefs.unregisterObserver(id));
 		if (this._citationDialogWindow) {
 			this._citationDialogWindow.close();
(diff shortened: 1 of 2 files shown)


mrtcode:
@tnajdek I did some changes on top of yours. Could you check if they are worthwhile to keep or not (the Zotero part)?

Changes:
- Previously Edit-menu commands used custom focus interception. Now they use Gecko’s native controller routing.
- Previously menu Undo skipped input-rule history. Now it matches Cmd-Z and restores text such as # .
- Previously Find-field Undo could affect the note. Now each keeps its own history.
- Previously embedded and standalone editors needed separate setup. Now each editor registers itself.
- Previously reinitialization did not manage the controller lifecycle. Now the old controller is removed before replacement.
- Previously read-only or crashed editors could receive Undo. Now Undo safely returns false.

https://github.com/mrtcode/zotero/tree/pr5987-lean-undo-redo
https://github.com/mrtcode/note-editor/tree/pr77-semantic-undo

tnajdek:
@mrtcode I've tested your changes and had a quick look at the code: it's a better integration and fixes some real issues! I've removed some white space leftovers and pushed your commit here.

Once the changes in the note editor are merged, I'm happy to merge this one (probably as a single commit).

mrtcode:
Thanks!

Ok, I merged the note-editor changes.

PR #5989, opened 2026-07-04
Extend the template engine and add template validation

### LINKED ISSUE #5965: Validate file renaming template in prefpane
https://forums.zotero.org/discussion/132296/pdfs-have-stopped-attaching-after-upgrade-to-zotero-9

The fix in https://github.com/zotero/zotero/commit/4a318617d6602d37296d4961b4f206ee6490f913 may or may not have even be right — maybe safer to just let it throw and use the default template. But if possible, we should validate the template and show a warning in the pane so that it doesn't break things to begin with.

tnajdek:
This PR improves the template engine and adds syntactic validation mechanism which is used prior to any file rename. Other template engine users are not updated to use the new validation.

If currently stored template is invalid, the new system will fallback for the default template. Should we add a migration step?

 - Fix a template engine crash when comparing against a bare number when the value isn't numeric
 - Add support for specifying string literals in the template engine
 - Validate `if/else/elseif` order and matching clause closures
 - Validate to ensure every `{{` is properly closed with a matching `}}`
 - When template is invalid, prompt to fix instead of offering batch-renaming tools

Screenshot 2026-07-04 at 15 31 22

Close #5965


dstillman:
A little awkward that it appears immediately on keypress, so the warning shows as you're typing. Probably better to show on blur and rely on the alert if someone clicks straight a button?

The template stuff is getting a bit unwieldy in utilities_internal, including internal helper functions (quotedStringBodySource/matchQuotedString) that don't need to be exposed to the outside. Can we (first, in a separate commit) separate out template stuff to a separate (lazy-loaded?) module, and then add the new stuff there?

scss/elements/_fileRenamingSettings.scss
```diff
@@ -15,15 +15,34 @@ file-renaming-settings {
...
+		padding: 8px 10px;
+		background: #fce8ea;
+		border-radius: 6px;
+
+		@media (prefers-color-scheme: dark) {
+			color: #ff7078;
+			background: #40191c;
+		}
+	}
+
```
dstillman, line 30: Are these existing accent colors? If so, reuse those variables, and don't bother with the WCAG line?

tnajdek: We don't AFAIK. Last time I've checked the only other place that used red message was Advanced Search where WCAG is not met (https://github.com/zotero/zotero/issues/5990)

My suggestion is: make re-usable variables from these, use in both cases, drop the WCAG line here.

dstillman: That works.

tnajdek:
- We now display the preview for the last valid template when the template is invalid, warning when input is blurred
- Separate commit now extracts the templating engine into a standalone, lazy-loaded module
- WCAG-compliant colors extracted in `main` (ccb592ba68317d3267975cfab5e0d3c96d442c0b), I've rebased this PR to use it

dstillman:
If you have a custom template set, try to edit it but make it invalid, click Done, get the prompt, and choose Reset to Default, the dialog closes without offering to update existing files. In a user library, it marks the renaming as pending, so you can reopen the pane and choose "Rename Files…". In a group library, there's no pending flag, you just miss the opportunity to update existing files.

tnajdek:
That's true. We can fix it by having two consecutive prompts (which is annoying) or having a single prompt that requires user to read through carefully because the action is complex: "Reset template to the default and rename files" (but it does take you to the preview first, so it's not as bad).

Preference on which approach to take?

dstillman:
I would just have the two prompts.

tnajdek:
Done, the flow now goes through two prompts.

dstillman:
Merged in efcfaec5d4520a17deef25e28c7fc4e32d7f84f1 and c39f16891fdecd9b35360f1d1cdc7b189c5fcddd

Issue #5676, opened 2025-12-23
Reduce PDF view memory usage


mrtcode:
For the PDF view, one of the biggest memory consumers is large canvases. They tend to become large when zooming in or when maximizing the Zotero window on a high-resolution display with zoom to page width enabled. Retina displays, of course, make this worse.

The first step is to destroy canvas objects when switching to another tab. When switching back, re-rendering should be instant. If it is still distracting, we may need smarter logic to decide when to unload canvases and when not to.

https://forums.zotero.org/discussion/128665/feature-request-reduce-zotero-ram-usage

dstillman:
See also https://github.com/zotero/zotero/issues/3843

pistachionet:
@mrtcode @dstillman see https://github.com/zotero/zotero/pull/5992

dstillman:
Addressed by https://github.com/zotero/reader/commit/a5e868f39bdd68dd1a69855c704db9a021551fee https://github.com/zotero/reader/commit/b1c5c8d6997fd426d890bbc56ef10e417b0f179e https://github.com/zotero/reader/commit/2f805314726706770e0459bff6a9dfc0a07ab692

Issue #3843, opened 2024-03-13
More sophisticated tab unloading


dstillman:
https://github.com/zotero/zotero/pull/3770 adds hidden prefs for tab unloading, but people shouldn't have to customize these limits on each system. At least on non-32-bit builds, it seems like we should keep tabs loaded more aggressively and just unload on memory pressure:

`nsIAvailableMemoryWatcherBase`:

https://searchfox.org/mozilla-central/rev/898d40036b2646daea9b44fd2e2fc44dc050158d/browser/modules/TabUnloader.sys.mjs#162-170

Firefox's tab-discarding logic:

https://searchfox.org/mozilla-central/rev/898d40036b2646daea9b44fd2e2fc44dc050158d/browser/modules/TabUnloader.sys.mjs#210-236

Firefox puts this behind a `browser.tabs.unloadOnLowMemory` pref (default true), though I don't know what happens if you turn that off and then use up all memory:

https://searchfox.org/mozilla-central/rev/898d40036b2646daea9b44fd2e2fc44dc050158d/browser/modules/TabUnloader.sys.mjs#185

Doesn't seem like we should spend much time worrying about 32-bit builds anymore, so I guess there maybe we just keep a hard limit.

dstillman:
https://forums.zotero.org/discussion/116437/how-to-prevent-tabs-reloading#latest

I think we do want to do both of the things that Firefox does: unload tabs after they haven't been viewed in a while (for more than ~3 seconds, to avoid just moving through tabs?) and keep active tabs in memory unless there's memory pressure.

PR #5992, opened 2026-07-05
Suspend hidden reader tabs and unload tabs on memory pressure

pistachionet:
## Why idle memory stays high with PDFs open

Four compounding causes, found by tracing the reader stack:

1. **Each open PDF keeps at least 10 fully rendered page canvases.** `PDFPageViewBuffer` (`web/pdf_viewer.js`, `DEFAULT_CACHE_SIZE = 10`) never shrinks below 10 and `update()` only ever grows it. At page-width zoom on a 2x HiDPI display a page canvas is ~2400×3100 device pixels ≈ 30 MB RGBA, capped at `maxCanvasPixels = 2^25` (128 MB). Zoomed pages add a detail-view canvas.
2. **The reader doubles that.** The annotation overlay renderer (`reader/src/pdf/page.js`) keeps a full-size `_snapshotCanvas` copy of every rendered page canvas — captured unconditionally, even for pages with no annotations — plus extracted per-page data in `_pdfPages`. So ~60 MB per buffered page, ~600 MB per loaded PDF tab, before the per-tab pdf.js worker heap (fonts/images).
3. **Hidden tabs release nothing.** On `tab-selection-change` the reader only ran `_updateLayout()` for the newly selected tab; a deselected reader kept its whole canvas buffer, snapshots, and worker heap indefinitely (#5676). pdf.js's idle `_cleanup` doesn't help: `PDFViewer.cleanup()` only resets non-FINISHED views, so finished canvases survive it.
4. **Tab unloading doesn't react to memory pressure** (#3843): `MAX_LOADED_TABS` = 3–5 and `UNLOAD_UNUSED_AFTER` = 24 h, so 3–5 idle tabs × ~600 MB lands right in the 1–4 GB range reported on the forums.

## What this PR does

- **`xpcom/reader.js`**: reader tabs call the reader's new `setSuspended(true)` 60 seconds after being hidden (grace period so quick tab switching doesn't thrash re-renders) and `setSuspended(false)` on selection. Suspension destroys all page views through the existing buffer-eviction path, freeing page canvases, detail-view canvases, snapshot copies, per-page data, and worker-side page/font resources; on selection, visible pages re-render through the normal path. The timer is cleared on tab teardown.
- **`tabs.js`**: a `memory-pressure` observer unloads all background tabs immediately when the OS reports low memory (`low-memory`/`low-memory-ongoing`, or `heap-minimize` from about:memory), complementing the count/age-based unloading. The observer is removed on window unload.

Net effect: a hidden PDF tab drops from ~600 MB of canvases to near zero, and background tabs are shed the moment the system is actually under pressure.

Addresses #5676 and #3843

### DIFF
--- a/chrome/content/zotero/tabs.js
+++ b/chrome/content/zotero/tabs.js
@@ -344,9 +344,23 @@ var Zotero_Tabs = new function () {
 		}
 	});
 	
+	// Unload background tabs right away when the OS reports memory pressure
+	// ("low-memory"/"low-memory-ongoing", or "heap-minimize" from about:memory),
+	// instead of waiting for the count/age-based unloading
+	this._memoryPressureObserver = {
+		observe: (subject, topic, data) => {
+			Zotero.debug(`Memory pressure (${data}): unloading background tabs`);
+			for (let tab of this._tabs.slice()) {
+				this.unload(tab.id);
+			}
+		}
+	};
+	Services.obs.addObserver(this._memoryPressureObserver, 'memory-pressure');
+
 	window.addEventListener('unload', () => {
 		Zotero.Notifier.unregisterObserver(this._notifierID);
 		Zotero.Prefs.unregisterObserver(this._prefsObserverID);
+		Services.obs.removeObserver(this._memoryPressureObserver, 'memory-pressure');
 	});
 
 	this._unloadInterval = setInterval(() => {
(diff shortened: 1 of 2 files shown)


dstillman:
Thanks. We've implemented several optimizations, including similar suspension of hidden tabs, in the latest beta.

We didn't add tab unloading on memory pressure. That should be much less necessary now, but it's also ~10 lines, so we could add that separately.

PR #5994, opened 2026-07-07
Update staged builds in place for faster dev builds

dstillman:
`dir_build` now saves a manifest of `build/` file hashes and stats after a full build. On subsequent runs, if all changed files are ones that go into `omni.ja` unmodified (`chrome/`, `components/`, `resource/`, and `test/` when tests are staged), it zips just those into the staged `omni.ja` instead of rebuilding — ~15s → ~0.3s on an M1 Mac, for both `build_and_run -r` and `runtests.sh`.

Zotero `.ftl` files are also updated at their `localization/<locale>/` paths, and test files are copied to the staged `tests/` directory.

Anything else falls back to a full rebuild automatically: files transformed by `build.sh` (`defaults/`, `chrome.manifest`, `version`, `translators/`, `styles/`, mozilla `.ftl`s, CSL locales), removed files, changes in `app/` (size/mtime fingerprint; xulrunner covered by the `hash-*` files), or requesting tests/devtools that aren't staged.

Flag changes:

- `dir_build` no longer takes `-q` — it always builds the uncompressed `omni.ja`, since the compressed/optimized version only matters for distribution builds via `build.sh` (unchanged). This also means the incremental path never encounters an optimized jar, which `zip` can't update.
- New `-f` on `dir_build` and `build_and_run` forces a full rebuild.
- `build_and_run` now always rebuilds — `-r` is gone, and a rebuild with no changes only takes ~1s even without the watch process running. New `-n` skips the rebuild and just launches.
- `build_and_run` no longer passes `-purgecaches`: startup caches are invalidated automatically via the BuildID, which is now bumped whenever `omni.ja` changes (including by `add_omni_file`), so relaunches of an unchanged build can use the startup cache.
- `build_and_run` and `runtests.sh` invoke `js-build/build.js` directly instead of via `npm run`, saving ~270ms of npm overhead.

Also fixes `dir_build` to remove broken symlinks left in `build/` by deleted source files, which previously broke the rsync in `prepare_build`.

### DIFF
--- a/app/scripts/add_omni_file
+++ b/app/scripts/add_omni_file
@@ -37,11 +37,14 @@ win_path="$STAGE_DIR/Zotero_win-x64"
 linux_path="$STAGE_DIR/Zotero_linux-x86_64"
 
 added=0
+build_id=$(date +%Y%m%d%H%M%S)
 for path in "$mac_path" "$win_path" "$linux_path"; do
 	if [ -d "$path" ]; then
 		echo "$path/app/omni.ja"
 		echo "Updating $(basename $(dirname $(dirname $path)))"
 		zip "$path/app/omni.ja" $files
+		# Bump BuildID so that startup caches are invalidated
+		perl -pi -e "s/^BuildID=.*/BuildID=$build_id/" "$path/app/application.ini"
 		added=1
 	fi
 done
--- a/app/scripts/build_and_run
+++ b/app/scripts/build_and_run
@@ -10,13 +10,23 @@ if [ -n "${ZOTERO_PROFILE:-}" ]; then
 	profile_args=(-p "$ZOTERO_PROFILE")
 fi
 
-REBUILD=0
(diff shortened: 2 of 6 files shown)


AbeJellinek:
Maybe the build manifest should go somewhere that gets cleaned by `npm run clean` (or it should be specifically added to `clean.js`).

dstillman:
I don't think that's necessary. If you do an `npm run clean` and rebuild and the output is exactly the same, there isn't actually anything new to copy into the omni.ja, so reusing it as is is correct. And if it deletes even a single file, that forces a full rebuild.

AbeJellinek:
Can it ignore `-r` (or print a warning?) instead of erroring, so I don't need to update my run config when switching to branches without this PR?

dstillman:
Done

AbeJellinek:
Works well for me. Big improvement!

PR #5995, opened 2026-07-07
Run tests in parallel CI shards

dstillman:
CI runs currently take ~11–12 minutes, almost all of it in the test run. This splits the tests across 4 parallel jobs, bringing a run down to ~5 minutes.

- `runtests.sh` gets a `-p SHARD/TOTAL` option (e.g., `-p 2/4`). The sorted test file list is split into contiguous alphabetical chunks of roughly equal total file size, with size as a stand-in for run time, so new test files are included automatically and no timing data needs to be maintained. Since chunks are contiguous, files run with the same preceding files as in a full run (except at chunk starts), and a shard can be reproduced locally by passing its first and last files to `-s`/`-e`.
- The CI workflow runs a matrix of 4 shard jobs, each doing the same build + ZIP round-trip as before. Each shard builds itself rather than sharing an artifact from an upstream job — with caching the build is ~17s, less than an artifact download would cost, and shards don't have to wait for a separate build job to finish before starting. A failure in one shard cancels the others. The deployment ZIP upload and utilities tests move to separate jobs, and `pull_request` jobs are skipped for same-repo PRs, which are already covered by the push run (fork PRs are unaffected).
- The test reporter now logs failures that will be retried, which previously weren't shown at all, and `waitForCallback` timeout errors include the callback source.

Sharding exposed several tests that depended on state from earlier test files, fixed here:

- `citationDialogTest`, `recognizeDocumentTest`, and `translatorsTest` were missing `Zotero.Styles.init()`/`Zotero.Translators.init()` calls.
- `recognizeDocumentTest` left a sinon stub wrapped after a failure, which broke retries.
- itemPaneTest attachments-pane tests relied on window size from a previous test or on `pinnedPane` to trigger rendering. (Setting `itemDetails.pinnedPane` can silently fail to stick — possibly stale pref observers from destroyed itemDetails instances clearing the pref. Worth a separate look, since it presumably affects actual pinning; the tests now scroll to the pane instead.)

Shard times are currently ~135/171/152/120s; changing the shard count is just editing the matrix list.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,10 @@ jobs:
   test:
     name: Test (shard ${{ matrix.shard }})
     runs-on: ubuntu-latest
+    # Same-repo PRs are covered by the push run, so run pull_request jobs only for forks
+    if: >
+      github.event_name != 'pull_request'
+      || github.event.pull_request.head.repo.full_name != github.repository
     strategy:
       matrix:
         shard: [1, 2, 3, 4]
@@ -83,6 +87,9 @@ jobs:
   utilities:
     name: Utilities Tests
     runs-on: ubuntu-latest
+    if: >
+      github.event_name != 'pull_request'
+      || github.event.pull_request.head.repo.full_name != github.repository
     steps:
       - uses: actions/checkout@v4
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,59 +4,62 @@ concurrency:
   group: ${{ github.ref }}
   cancel-in-progress: true
 jobs:
-  build:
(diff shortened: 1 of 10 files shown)


tnajdek:
This is really nice! I wish there was a way to parallelize local test runs as well but with real windows stealing focus from each other that might not be feasible.

dstillman:
We could possibly shard local tests by 1) parameterizing the port, 2) forcing the itemPane and zoteroPane suites, and maybe a few elsewhere, into foreground-only runs, and 3) forcing the others to run in the background. But not sure it's worth it. For complete test runs, pushing to a branch and letting it run in 5 minutes isn't the end of the world. (I guess non-Linux tests, and non-CI tests, make it a bit more worthwhile.)

PR #6000, opened 2026-07-13
Server: Warn when HTTP server falls back to a different port

### LINKED ISSUE #5999: Desktop silently switches httpServer.port to a fallback when 23119 is occupied, breaking browser connector
## Summary

On Linux (Ubuntu 24.04), if port `23119` is briefly occupied when Zotero 9 starts — for example by a short-lived local service — Zotero silently binds to a different port (`23124` in my case) and writes that port into `prefs.js` as `extensions.zotero.httpServer.port`. The browser connector hard-codes `http://127.0.0.1:23119/` and has no discovery mechanism, so saving from the connector stops working permanently until the user manually edits `prefs.js`. The next startup, if `23119` is free again, Zotero still reads the persisted fallback port and stays on `23124`, so the link never self-heals.

## Environment

- OS: Ubuntu 24.04 (Noble Numbat)
- Kernel: 6.17.0-35-generic
- Session: X11, `ubuntu:GNOME` desktop
- Architecture: x86_64, glibc 2.39
- Zotero: 9.0.6 (official tarball at `~/Zotero`; no Flatpak / Snap)
- Browser: Microsoft Edge 150.0.4078.65 (`microsoft-edge-stable` from the official .deb repo, amd64)
- Zotero Connector: 5.0.210 installed from the Edge Add-ons store (extension ID `nmhdhpibnnopknkmonacoephklnflpho`)

## Reproduction

1. Briefly occupy port `23119` during Zotero startup. In my case it was a reverse-proxy style local service that bound `23119`, was killed before Zotero finished initializing, and never came back.
2. Start Zotero. Confirm it appears to be running normally (taskbar icon, no error dialog).
3. List the ports it actually owns:
   ```bash
   ss -lntp | grep zotero-bin
   # LISTEN  127.0.0.1:19876  ...  zotero-bin
   # LISTEN  127.0.0.1:23116  ...  zotero-bin
   # LISTEN  127.0.0.1:23124  ...  zotero-bin
   ```
   Note: `23119` is not in the list. By now the short-lived service has exited, so `23119` is free — but Zotero is still bound to `23124`.
4. Inspect the profile prefs:
   ```bash
   grep httpServer ~/.zotero/zotero/<profile>/prefs.js
   # user_pref("extensions.zotero.httpServer.localAPI.enabled", true);
   # user_pref("extensions.zotero.httpServer.port", 23124);
   ```
   The fallback port has been persisted as the new "configured" value.
5. Try to save anything from the browser connector. The save icon stays gray and reports "unable to communicate with Zotero".
6. `curl http://127.0.0.1:23119/connector/ping` returns `connection refused` (Zotero KB's suggested diagnostic URL).
7. `curl http://127.0.0.1:23124/connector/ping` returns `Zotero is running`.

## Root cause

`chrome/content/zotero/xpcom/server/server.js`:

```js
port = port || Zotero.Prefs.get('httpServer.port');
serv = new HttpServer();
serv.registerPrefixHandler('/', this.handleRequest);
serv.start(port);
Zotero.debug(`HTTP server listening on 127.0.0.1:${serv.identity.primaryPort}`);
```

`HttpServer.start()` (Firefox's `netwerk/base/nsIServerSocket`) auto-selects the next free port when the requested port is unavailable. Zotero's wrapper code reads back `serv.identity.primaryPort` and logs it, but does not detect the discrepancy with the requested port. Worse, the actual bound port appears to be written to `prefs.js` (the file's mtime equals the Zotero startup time, and no other process touched it).

The connector side, in contrast, has no fallback of its own: `src/common/zotero.js` in `zotero/zotero-connectors` ships with `"connector.url": 'http://127.0.0.1:23119/'` hard-coded, and no native messaging host is queried to discover the real port.

The two ends can silently disagree, with no signal to the user.

## Suggested fix

Desktop side (this report's main concern):

- **Don't persist the fallback port to `prefs.js`.** Keep the user-configured value (or the default `23119`) as the canonical port. Log the actual bound port to debug output only.
- **Or**, when startup falls back, surface a clear warning in the UI ("Connector communication is unavailable because port 23119 is occupied by another process"). Otherwise users have no way to know why the connector is broken.
- **Or**, at minimum, expose the actual bound port in `about:config` next to the pref, or in the existing "Allow other applications on this computer to communicate with Zotero" preference UI (cf. #4612).

Connector side: a real fix probably needs the connector to learn the port dynamically — for example via native messaging — but that's tracked separately in zotero/zotero-connectors#170 and is out of scope here.

## Related

- zotero/zotero-connectors#170 — "Add hidden pref for connector server port" (closed without implementation). Connector side of the same problem.
- zotero/zotero#4612 — "Show current server port under 'Allow other applications on this computer to communicate with Zotero' when enabled?" (closed without implementation). Same discoverability gap from a different angle.

cislunarspace:
## Summary

When `extensions.zotero.httpServer.port` (default `23119`) is occupied at startup, Firefox's `HttpServer` silently binds to the next free port. `Zotero.Server.init()` only logged this at debug level 5, leaving users with no visible signal that the browser connector had stopped working — symptom is a grayed-out save icon and the standard "unable to communicate with Zotero" message.

This commit makes the fallback detectable:

- When the actual bound port differs from the configured port, emit a level-1 (warning) log that names both ports and tells the user to free the configured port and restart.
- Always populate a new pref `extensions.zotero.httpServer.actualPort` with the bound port, so the value is visible in `about:config` for diagnostic purposes.
- The configured `extensions.zotero.httpServer.port` is **never** written from `init()`, so it remains stable across restarts and the warning only fires when a real conflict exists.

Fixes zotero/zotero#5999.

## Reproduction

1. Briefly occupy port `23119` during Zotero startup, e.g. `nc -lk 23119 &`.
2. Start Zotero.
3. Without this patch: nothing visible to the user. `~/.zotero/zotero/<profile>/prefs.js` is left untouched, but `Zotero.Server.port` is whatever Firefox picked (often `23124`).
4. Browser connector stops working. The user has no clue why.

## Tests

- Added `test/tests/serverTest.js > #init()` block with two assertions:
  - `httpServer.actualPort` pref equals `Zotero.Server.port` after init.
  - `httpServer.port` pref is not mutated by `init()`.

These can't exercise the actual fallback path in a single-process test runner (the test server shares the process), but they pin down the two observable invariants of the new behavior.

## Manual verification

- `npx eslint` reports the same number of pre-existing issues before and after the diff (no new lint regressions).
- `node --check` confirms the JS syntax is valid.
- Full `npm run build` and the test runner were not executed locally because the project's sparse-checkout clone does not have a built Zotero binary to run tests against. Zotero CI will run them.

## Related

- zotero/zotero-connectors#170 — connector-side: add hidden pref for connector server port (closed without implementation). The connector still hard-codes `http://127.0.0.1:23119/`; that fix is out of scope here.
- zotero/zotero#4612 — desktop-side: show current server port under "Allow other applications…" preference UI (closed without implementation). This PR makes the value available via `httpServer.actualPort`, which a follow-up could surface in the UI.
- zotero/zotero#4903 — related cleanup that fixed `Zotero.Server.port` getter semantics.

### DIFF
--- a/chrome/content/zotero/xpcom/server/server.js
+++ b/chrome/content/zotero/xpcom/server/server.js
@@ -62,15 +62,32 @@ Zotero.Server = new function () {
 			Zotero.debug("Already listening on port " + serv.port);
 			return;
 		}
-		
+
 		port = port || Zotero.Prefs.get('httpServer.port');
+		let requestedPort = port;
 		try {
 			serv = new HttpServer();
 			serv.registerPrefixHandler('/', this.handleRequest)
 			serv.start(port);
-			
-			Zotero.debug(`HTTP server listening on 127.0.0.1:${serv.identity.primaryPort}`);
-				
+
+			let actualPort = serv.identity.primaryPort;
+			// Always expose the actually-bound port for diagnostic purposes.
+			// This intentionally uses a separate pref so it never overwrites
+			// the user's configured httpServer.port on shutdown.
+			Zotero.Prefs.set('httpServer.actualPort', actualPort);
+
+			if (actualPort !== requestedPort) {
+				Zotero.debug(
+					`HTTP server: requested port ${requestedPort} was unavailable; `
+					+ `bound to ${actualPort} instead. The browser connector will not be able `
+					+ `to reach Zotero until port ${requestedPort} is freed and Zotero is restarted.`,
+					1
+				);
+			}
+			else {
+				Zotero.debug(`HTTP server listening on 127.0.0.1:${actualPort}`);
+			}
+
 			// Close port on Zotero shutdown (doesn't apply to translation-server)
 			if (Zotero.addShutdownListener) {
 				Zotero.addShutdownListener(this.close.bind(this));
--- a/defaults/preferences/zotero.js
+++ b/defaults/preferences/zotero.js
@@ -160,6 +160,7 @@ pref("extensions.zotero.integration.citationPreviewShown", true);
 // Connector settings
 pref("extensions.zotero.httpServer.enabled", true);
 pref("extensions.zotero.httpServer.port", 23119);	// ascii "ZO"
+pref("extensions.zotero.httpServer.actualPort", 0);	// populated at startup with the actually-bound port
 pref("extensions.zotero.httpServer.localAPI.enabled", false);
 
 // Zeroconf
--- a/test/tests/serverTest.js
+++ b/test/tests/serverTest.js
@@ -395,4 +395,29 @@ describe("Zotero.Server", function () {
 			});
 		});
 	});
+
+	describe('#init()', function () {
+		it("should set httpServer.actualPort pref to the bound port", function () {
+			assert.equal(
+				Zotero.Prefs.get('httpServer.actualPort'),
+				Zotero.Server.port,
+				'httpServer.actualPort pref matches the bound port'
+			);
+		});
+
+		it("should preserve httpServer.port pref across startup", function () {
+			// The init() implementation must never overwrite the user's configured
+			// httpServer.port, even when the bound port differs from it.
+			// (See zotero/zotero#5999.)
+			let configured = Zotero.Prefs.get('httpServer.port');
+			assert.isAbove(configured, 0, 'httpServer.port is set');
+			// No matter what the actualPort ended up being, the configured
+			// value should be unchanged.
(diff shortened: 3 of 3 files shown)


dstillman:
https://github.com/zotero/zotero/issues/5999#issuecomment-4960964503

PR #6012, opened 2026-08-05
Semantic search

dstillman:


abaevbog:
Brief summary the main recent changes here and my rationale:
1. Model calibration moved from tests into the embeddings module `Zotero.Embeddings.Calibration`. It contains the corpus of X passages along with a query to match each. We then generate embeddings for each query paired with each passage (X by X), so there is a score for expected query-passage match and many scores for query-non-matching-passage pairs. The scores are sorted, and we place the min floor to include a minimal portion of unrelated pair. With this, calibration can happen dynamically.
2. Annotations and notes are embedded and searchable. Regular item gets the max relevance score of itself or any of its children, so a very relevant annotation pushes the result higher.
3. Added embedding and semantic search across attachments. Chunking happens in `Zotero.Utilities.Internal.Chunking`, that tries to chunk as close to outline as possible. The attachment row gets the score of its best matching chunk. Text matching the chunk is not stored - just structured text location. It means, SDT needs to re-derive text on demand to fetch the text.
4. Added lexical search as a part of "Best Match" search mode. It runs bm25 on the fulltext index and a newly-added index of metadata of regular items, notes, and annotations in `fulltext`, and combines results. Results from lexical search are then fused with results from semantic search via strength-weighted Reciprocal Rank Fusion. Why lexical is needed:
  - can be used to show relevance without involving embeddings (during indexing or with embeddings disabled)
  - necessary to locate keywords in brief queries that don’t have strong semantic representation. (e.g. surface item mentioning “unleaded gas” for a query that just says “unleaded”, or handle technical terms/abbreviation/etc. that model doesn’t know about).
 
One side effect: keywords on long queries can add a lot of noise. The longer a query is, the more likely it will be better served by just semantic search. So if a query is > 2 searchable terms, only perform lexical search on title + abstract.

5. Up to 3 search previews are rendered as child rows in itemTree below annotations. Text of matched chunks is extracted via SDT when possible. Matches that have some lexical matches will render text around those matches. Purely semantic matches just render first 1-2 sentences. SDT extraction takes some work, so we don’t render all previews in one go. Top 10 results are rendered, the rest is filled lazily when browser is free.
6. New “Search results” section in itempane of the attachment to render all matching chunks. 
7. Run vector search via `sqlite-vec` extension (`Zotero.DB.loadExtension('vec’)`) vs doing it in-memory to speed up the search. 
8. Add optional reranker to rerank top results after scoring to move most relevant items closer to top 10.

The main question: during attachment embedding, we rely on section's outline being extracted, which means running SDT for every attachment. Is that an issue? Should it be a separate process independent of the embeddings pipeline, since SDT is also needed to locate page of a purely lexical search match?

PR #6015, opened 2026-08-07
Citation dialog: suggest page locator of currently opened documents

### LINKED ISSUE #6001: Citation dialog: Suggest page number for open documents
https://forums.zotero.org/discussion/132711/feature-request-suggest-current-page-number-in-citation-popup

Need to think about how to present this. I don't think we'd want to just auto-populate a page number for any open tab, but we should make it easy to select the current page via both keyboard and mouse. Need to do it in some way that doesn't conflict with the placeholder we're now showing ("Type 10–15…").

Not sure what to do about non–page numbers. @mrtcode, do we have any automatic detection for other locator types?

(Also, the current tab shows up in Selected Items rather than Open Documents, which is a bug. I'll fix.)

abaevbog:
Display a suggestion popup with the current page label after an open document item is added as a bubble. Unloaded tabs do not have an opened reader, so the suggestion does not apply to them.

The suggestion popup appears below the bubble-input and can be accepted via click or arrow down + Enter.
When the popup is visible, any other keypress dismisses the popup. The popup does not consume keypress, except for Escape, so one can continue typing.

Fixes: #6001

https://github.com/user-attachments/assets/86452af3-6b45-4636-9b0c-c0553e27b39f

### DIFF
--- a/chrome/content/zotero/integration/citationDialog.js
+++ b/chrome/content/zotero/integration/citationDialog.js
@@ -196,6 +196,10 @@ function onUnload() {
 }
 
 function cleanupBeforeDialogClosing() {
+	if (IOManager._notifierID) {
+		Zotero.Notifier.unregisterObserver(IOManager._notifierID);
+		IOManager._notifierID = null;
+	}
 	if (!currentLayout || !libraryLayout) return;
 	
 	// Save window params for current layout mode so we can restore it on next dialog open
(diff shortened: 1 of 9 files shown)


dstillman:
Seems like this is dependent on the tab actually being loaded? It's not being triggered for unloaded tabs.

abaevbog:
Right, I actually thought that a suggested "current page" for a tab that's unloaded and doesn't have a proper current page would seem off. But to keep it consistent, we can save the last page label to the attachment's state file (similar to page index) and make a suggestion based on that.

dstillman:
Would be neat to update the page number in the tooltip if you bring Zotero up behind the citation dialog and start scrolling to a different page

abaevbog:
Good call. Apparently there already is a notifier event that fires when the page changes, so all we need to do is just update the locator when notifier fires.


https://github.com/user-attachments/assets/c615f0c4-8a9d-4fe2-8f7f-fc8a570157cb

PR #6016, opened 2026-08-10
Fix zero local data object versions

schmidma:
The local `clientVersion` migration initializes existing items, collections, and saved searches to `0`. Their row loaders handle `version` as an integer, but not `clientVersion`, so the generic fallback converts the falsy zero to an empty string. The Local API consequently returns `version: ""` for migrated objects until they are saved.

Handle `clientVersion` alongside `version` in all three loaders, preserving numeric zero. The regression test simulates migrated rows and verifies the value after unloading and reloading each object type.

🤖 **AI disclosure:** This contribution was developed with AI coding assistance. I directed the investigation, reviewed the implementation and tests, and validated the fix against a migrated Zotero database and Zotero’s test suite.

### DIFF
--- a/chrome/content/zotero/xpcom/data/collection.js
+++ b/chrome/content/zotero/xpcom/data/collection.js
@@ -140,6 +140,7 @@ Zotero.Collection.prototype.loadFromRow = function (row) {
 		
 		// Integer or 0
 		case 'version':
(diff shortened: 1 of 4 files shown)


dstillman:
Thanks!

PR #6021, opened 2026-08-14
Reader: Mirror undoable actions to global undo history

AbeJellinek:
This keeps the reader's own undo/redo mechanism intact for the web library and other platforms, but when running inside Zotero, the reader relinquishes control over undo/redo keyboard shortcuts and instead sends its history up to the client, which interleaves it with global history (with appropriate labels) and calls back to the reader if an undo/redo is requested.

In the process, add a couple mechanisms to `Zotero.UndoHistory`:

- Add support for updating Undo/Redo menu items in a non-library window, e.g., a reader window.
- Add support for a new "provider" interface, representing a component that pushes its own independently managed undo history and is responsible for undoing/reapplying its changes when asked. Provider histories are merged into the main global undo/redo history. Right now this is only applied to the reader, but it could be used for the note editor as well.
	- Each reader instance registers itself as a provider on open, and unregisters on close.

Currently, library undo/redo actions don't provide any feedback. We don't select the item that an edit was undone/reapplied to, or switch to the library it's in if the user had switched away. This PR diverges from that behavior for reader undo/redo -  when a reader action is undone from a library tab, the reader regains focus, and vice versa, so it's a little more obvious where things are happening. It felt confusing to use without that feedback. If that isn't desirable, I could (a) remove the switch mechanism, or (b) implement modified item selection for library actions as well. (macOS Finder does select the modified file/folder when you undo, and I personally think it's good behavior!)

### DIFF
--- a/chrome/content/zotero/standalone/standalone.js
+++ b/chrome/content/zotero/standalone/standalone.js
@@ -237,52 +237,11 @@ const ZoteroStandalone = new function () {
 		this.updateQuickCopyOptions();
 		// goUpdateGlobalEditMenuItems(true) is necessary to update Edit menu when contenteditable is focused
 		window.goUpdateGlobalEditMenuItems(true);
-		this._updateUndoRedoLabels();
+		Zotero.UndoHistory.updateMenuItems(document);
 
 		this.onUpdateCustomMenus(event, 'edit');
 	};
 
-	this._updateUndoRedoLabels = function () {
-		let undoItem = document.getElementById('menu_undo');
-		let redoItem = document.getElementById('menu_redo');
-		if (!undoItem || !redoItem) return;
-
-		// When a native text-editing controller handles undo/redo
-		// (e.g. focused input), show generic labels and let it take over
-		let nativeUndo = Zotero.UndoHistory.hasNativeUndo(document);
-		let nativeRedo = Zotero.UndoHistory.hasNativeRedo(document);
-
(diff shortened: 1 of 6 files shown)


tnajdek:
Works great overall, I really like the fact that undo/redo switches to the correct tab.

One minor concern is that`_lastHistoryPointID` doesn't reset when undo stack is cleared. I suspect this causes an issue I've observed:

1. Edit an annotation.
2. Make remote change, sync (or wait for auto-sync)
3. Edit the same annotation again (any number of times)

Undo option shows generic `Undo` rather than `Undo Edit Annotation` and it no longer applies the undo. 

Also, couple of nitpicks:

-> Redo stack isn't cleared on a provider revision bump; scenario:

  1. Edit annotation twice
  2. Switch to the library, rename a collection
  3. Undo
  4. Back to the reader, edit annotation third time
  5. Redo option still available, should be gone

-> `_isTextBoxFocused` also returns `true` for `radio`/`checkbox` inputs.

Finally, I'm going to update #5986 to work with this once it lands, but I'll keep my `pushCustom` API as a sibling to `setProviderSteps` since they address sufficiently different needs, any thoughts on that?

AbeJellinek:
Should all be fixed, thanks!

> I'll keep my `pushCustom` API as a sibling to `setProviderSteps` since they address sufficiently different needs, any thoughts on that?

Agreed. Let me know if you have any issues with the rebase. I think they mostly coexist separately but there's some logic we can combine (`_apply()` and the `libraryID` field on entries, at least).

PR #6026, opened 2026-08-20
Fix item tree indent when dragging a file between rows

### LINKED ISSUE #6024: Item tree: Top-level item indents when dragging attachment over it
https://github.com/user-attachments/assets/72a513d4-2f37-4681-add6-1f44ab229cbd

Drag a file from Finder over top-level item rows. When the drag target shows a line (rather than highlighting a row), one of the top-level item rows becomes indented as if it will become a child of the new attachment.

BetterAndBetterII:
A between-row file drop inserted `span.drop-before` / `span.drop-after` as the row's first child. First-column CSS then treated that line as a cell and indented a top-level item as if it were becoming a child of the incoming attachment.

Append the drop line after the cells, apply first-column padding to `.first-column` as well as `:first-child`, and position the line against the row.

Fixes #6024

### DIFF
--- a/chrome/content/zotero/itemTree.jsx
+++ b/chrome/content/zotero/itemTree.jsx
@@ -2249,16 +2249,8 @@ var ItemTree = class ItemTree extends LibraryTree {
 			div.classList.remove('tight');
 		}
 		
-		if (this._dropRow == index) {
-			let span;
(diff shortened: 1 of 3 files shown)


dstillman:
Thanks. Fixed (a bit more simply) by c30163ed907bd17dd1ed66cc63eaaccad6471e5b

PR #6028, opened 2026-08-24
Match duplicate books by ISBN-10/13 equivalence

jnylin:
## Summary
  Zotero's duplicate detection compares ISBNs as raw cleaned strings, so
  two records for the same book — one catalogued with the older ISBN-10
  and one with the modern ISBN-13 — were never flagged as duplicates,
  even though they're the same identifier.

  - Canonicalize ISBNs to ISBN-13 (`Zotero.Utilities.toISBN13`, already
    used elsewhere in the codebase for this exact purpose) before
    comparing, so ISBN-10 and ISBN-13 forms of the same book match.
  - ~~Also cache ISBNs for `bookSection` items. A section's ISBN
    identifies its *containing* book, not the section itself, so
    sections are intentionally **not** unioned directly by ISBN (that
    would wrongly merge distinct chapters of the same book). Instead,
    the cached value now lets the existing title/creator match correctly
    veto false positives between same-titled sections from different
    books — a case the ISBN veto previously missed entirely for this
    item type.~~

  ## Test plan

  - [x] Added `test/tests/duplicatesTest.js` cases: ISBN-10/ISBN-13
        equivalence match, differing-ISBN books stay unmatched, ~~and the
        bookSection veto (same title/creator, different container ISBN)
        does not merge~~.
  - [x] `test/runtests.sh -f duplicates` — all ~~6 tests pass~~ ~~5 tests pass~~ 4 tests pass.

### DIFF
--- a/test/tests/duplicatesTest.js
+++ b/test/tests/duplicatesTest.js
@@ -110,4 +110,26 @@ describe("Duplicate Items", function () {
 			assert.sameMembers(item3.relatedItems, [item1.key]);
 		});
 	});
+
+	describe("ISBN matching", function () {
+		async function getDuplicateSets() {
+			var duplicates = new Zotero.Duplicates(Zotero.Libraries.userLibraryID);
+			var search = await duplicates.getSearchObject();
+			return search.search();
+		}
+
+		it("should match books with equivalent ISBN-10 and ISBN-13", async function () {
+			var item1 = await createDataObject('item', { itemType: 'book', title: 'Effective Java' });
+			item1.setField('ISBN', '0134685997');
+			await item1.saveTx();
+
+			var item2 = await createDataObject('item', { itemType: 'book', title: 'Effective Java, 3rd Edition' });
+			item2.setField('ISBN', '9780134685991');
+			await item2.saveTx();
+
+			var ids = await getDuplicateSets();
+			assert.include(ids, item1.id);
+			assert.include(ids, item2.id);
+		});
+	});
 });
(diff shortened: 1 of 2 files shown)


dstillman:
Thanks. We'll take the ISBN-10/13 equivalence one-liner and test if you limit this to those changes.

The other part is making a much bigger change that should be submitted in a separate PR for discussion. "Introduction" is a problem, but I don't think it's clear that different container ISBNs should disqualify a match, since there could be various editions of the container. `bookTitle` seems like a better signal? Anyway, let's look at that separately.

jnylin:
Makes complete sense! I’ll strip out the container ISBN logic and narrow this PR down strictly to the ISBN-10/13 equivalence one-liner (plus tests).

Regarding the container logic, I agree that deserves a dedicated discussion—I’ll open a separate PR/issue for that so we can explore using bookTitle or other signals instead.

I’ll update the branch tomorrow!

jnylin:
I’ve updated the branch now to focus strictly on the ISBN-10/13 equivalence logic and tests. Ready for another look!

test/tests/duplicatesTest.js
```diff
@@ -110,4 +110,40 @@ describe("Duplicate Items", function () {
...
+
+			var item2 = await createDataObject('item', { itemType: 'book', title: 'Learning JavaScript' });
+			item2.setField('ISBN', '0596009089');
+			await item2.saveTx();
+
+			var ids = await getDuplicateSets();
+			assert.notInclude(ids, item1.id);
+			assert.notInclude(ids, item2.id);
+		});
+	});
```
dstillman: This test doesn't seem to have anything to do with the changes here and wouldn't fail without the changes.

jnylin: Fair point! A bit of belt-and-suspenders on my part. Removed in c94fe48e55699bb48605f03d1e099b6fddebba2f

PR #6030, opened 2026-08-26
Fix symlink unit tests on NixOS

Mynacol:
The unit tests
- should throw error on broken symlink
- should convert the target of a symlinked database file

are failing in the nix world, because it has no `/bin/ln`. I rewired both invocations to support this use case.

Feel free to adapt, change the commit messages etc.; I'd love to see something like that land.


Mynacol:
Thanks for picking it up!
If you want to double down, there is a /bin/mv invocation somewhere left in the code that could similarly be replaced (I think).

PR #6035, opened 2026-09-01
Support multiple main windows

dstillman:
Other bugs found during this work, with fixes committed separately:

b3b6e7f33db06322e83e81b2771566d2af8c719d Reopen closed reader and note tabs at their original position
d0a32a3b4f1999b1191c9bd319d89385a8f37ba4 Poll for expected tag selector state in flaky tests
0e46a5356b955fa87b5de84298a53cd9a5e5ad07 Fix tag selector observer leaks on uninit
22a1b33a429ffb7042752977ad8ae5d5f937953f Remove library tree event listeners on teardown
53bcd30a5fac0aaef3f54298ab788ea9c22505f2 Don't report a window's observers as leaked while it's still closing
92dfac8c6445f19b2cb0343739fd0cfbd77cc663 Don't make an unloaded tab reopenable with Undo Close Tab

### DIFF
--- a/chrome/content/zotero/modules/treePrefs.mjs
+++ b/chrome/content/zotero/modules/treePrefs.mjs
@@ -1,0 +1,118 @@
+/*
+    ***** BEGIN LICENSE BLOCK *****
+
+    Copyright © 2026 Corporation for Digital Scholarship
+                     Falls Church, Virginia, USA
+                     https://www.zotero.org
+
+    This file is part of Zotero.
+
+    Zotero is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Affero General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    Zotero is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
+    along with Zotero.  If not, see <http://www.gnu.org/licenses/>.
+
+    ***** END LICENSE BLOCK *****
+*/
+
+ChromeUtils.defineESModuleGetters(globalThis, {
+	Zotero: "chrome://zotero/content/zotero.mjs"
+});
+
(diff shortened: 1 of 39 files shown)


chrome/content/zotero/modules/treePrefs.mjs
```diff
@@ -0,0 +1,104 @@
...
+	},
+
+	/**
+	 * @param {String} id - Tree id
+	 * @return {Promise<Object>}
+	 */
+	async get(id) {
+		let settings;
+		await this._withSettings((all) => {
+			settings = all[id];
```
adomasven: Should be
```
settings = structuredClone(all[id] || {});
```
otherwise getters can modify the global settings object that is written above. Same below for `set()`

chrome/content/zotero/modules/treePrefs.mjs
```diff
@@ -0,0 +1,104 @@
...
+export var TreePrefs = {
+
+	/**
+	 * Run a function with the settings for all trees, one caller at a time, writing the file
+	 * back out if the function returns true
+	 *
+	 * @param {Function} fn - Passed the settings object for all trees
+	 * @return {Promise}
+	 */
+	_withSettings(fn) {
```
adomasven: The semantics of fn being able/having to return true/false are a bit confusing/implicit. Could take an argument `options = {write: false}`, or a separate function `_updateSettings(fn)`, which always writes after `fn()` is called, and `_withSettings()` is merely for getting.

chrome/content/zotero/modules/treePrefs.mjs
adomasven, line 1: This is good for enforcing the mutex, but ultimately unclear why it's part of multiple main windows commit, and not a separate refactor.

dstillman: Well, it's only relevant now because there can be multiple trees saving at once, but I've moved it to a separate commit for clarity.

chrome/content/zotero/modules/treePrefs.mjs
dstillman: Moved to a separate function

chrome/content/zotero/modules/treePrefs.mjs
dstillman: Fixed

PR #6049, opened 2026-09-17
Don't scroll item tree on changes made by the user in the view

### LINKED ISSUE #6023: Item tree: Expanding a row above the current selection causes expanded row to move up
https://github.com/user-attachments/assets/af53527d-f844-4d0b-b711-e516e67175dd

Works as expected when expanding a row at/below the selection, but when the expanded row is above, it moves up. If it's already at the top of the view, it moves offscreen.

I hadn't noticed this before the item tree refactor, but I'm not sure exactly when it was introduced.

dstillman:
Fixes #6023


adomasven:
Removed extra conditionals and options that need to be threaded through everywhere and made the default behavior reasonable. Can merge.

Issue #294, opened 2013-04-10
WebDAV support for groups


zwieberl:
Hello,

there has been some discussions on saving pdf-attachments of groups on external servers (and its impossibility).
I'd like to propose a compromise on that:

Scenario:
We are a group which would like to share pdf-collections. So we create group-librarys. BUT we do not want to store them on zotero-Servers (some of the documents are too sensitive). So what to do?

Solution:
We would like to be able to define for each group a different Storage-Location. So the Meta-data will get sync'ed to zotero-servers, but the actual files should remain within our local network.
Now, only the owner of that group has the pdfs. Everybody could sync (=copy) now his storage-folder. But then everyone would have his personal libraries as well, not only the group-docs.
So if one could specify a different storage-location (which is already possible) for a specific group, than people could sync only the group-pdfs locally and the Meta-Data from zotero.

A little quicker and dirtier workaround:
Name the folders in the storage-dir not something like 'WKELSDL' but 'groupname_WKELSDL' if they are in a certain group. This way it is much easier to rsync them somewhere else and leave the personal ones untouched.

Thanks!

lzbk:
>  just that we're unlikely to implement it anytime soon without someone with a vested interest in this functionality paying for the development time.
Is there an estimate of the price for the development time ?

lzbk:
> @dstillman Is there an estimate of the price for the development time ?
Still interested in this question…

que-vector:
My group and I would like to use this feature too.

> Would indeed be great to be able to configure a specific remote Webdav folder for a group collection. Would even be fine if this configuration was done locally by each member of the group − who would of course all have to communicate somehow to make sure to use the same remote folder.

lzbk:
> > @dstillman Is there an estimate of the price for the development time ?
> 
> Still interested in this question…
@dstillman Still, still interested ;) ;)

michaelweinold:
@dstillman, any update on the prioritisation of this feature request?

MustafaElshani:
I think this was first asked back in 2009 in forums, any update or workaround on how to implement it?

kuang-da:
I feel like if the dev team started a kickstart-like campaign for this feature, it would probably be fulfilled in a flash. 🚀

ltalirz:
It might help to get some more upvotes for this feature.

[...]

PR #6053, opened 2026-09-22
Per library webdav storage

RoySRC:


dstillman:
Thanks, but we won't be adding WebDAV support in groups:

https://github.com/zotero/zotero/issues/294#issuecomment-5770451238

PR #6054, opened 2026-09-22
Prevent stale Add-ons removal from deleting replacement

MinhHaDuong:
## Summary

The Add-ons window can retain a pending-uninstall entry for an old add-on wrapper after installing a replacement with the same ID. When the list closes, finalizing that stale entry can remove the replacement.

Patch Firefox's `addon-list.mjs` during Zotero's `fetch_xulrunner` build step. An install clears stale pending entries for that ID. List teardown snapshots pending entries before asynchronous lookup and finalizes only an entry that is still the current add-on wrapper and still pending uninstall. If the list reconnects during lookup, teardown stops.

## Test

- The added Node test passed all seven cases against the patched Firefox 153.1.0 `addon-list.mjs`: replacement after install event, replacement before install event, genuine pending removal, reconnect during lookup, install notification during lookup, reconnect during a multi-add-on teardown, and lookup failure.
- `bash -n app/scripts/fetch_xulrunner` passed.

The complete Zotero Add-ons UI flow has not been run in this checkout.


MinhHaDuong (commented):
APPROVED — Independent internal review of 2fe028cb8880347889f99810c4631aae42c7ba9e. The patch prevents a stale pending-uninstall wrapper from removing a replacement add-on, while preserving finalization for the current pending wrapper. The synchronous snapshot and clear, followed by the connection check after lookup, addresses the reconnect race. The four focused tests cover replacement, delayed install notification, genuine pending removal, and reconnection during lookup. Firefox's XPIDatabase caches each AddonInternal wrapper and getAddonByID returns that wrapper, supporting the identity check. I found no blocking issue in this diff. This verdict covers the patch and focused module tests; it is not an end-to-end Zotero UI run.

MinhHaDuong (commented):
**Independent review — Claude Opus (agent), reviewing head `2fe028cb8880347889f99810c4631aae42c7ba9e`.**

I reviewed this without relying on the earlier review in the thread. To check the build-time patch I extracted `chrome/toolkit/content/mozapps/extensions/components/addon-list.mjs` from Firefox 153.1.0, replayed the five `replace_line`/`remove_line` calls from `modify_omni` verbatim through `app/scripts/utils.sh`, and ran the new test file against both the patched and the unpatched copy.

## Verdict: APPROVED

### What I verified

- **The patch applies to real 153.1.0 source.** All five anchors match, each exactly once (`^  disconnectedCallback\(\) \{`, `^    this\.pendingUninstallAddons\.clear\(\);`, `^    for \(const addon of this\.pendingUninstallAddons\) \{`, `^      if \(isPending\(addon, "uninstall"\)\) \{`, `^  onInstalled\(addon\) \{`). The two near-collisions I looked for are not hit: `for (let addon of this.pendingUninstallAddons)` in `render()` uses `let`, and the other `isPending(addon, "uninstall")` site in `getAddons()` is `} else if (…)`, so the six-space prefix in the pattern excludes it.
- **Ordering is right.** `remove_line` deletes the original `clear()` before the loop replacement reintroduces one; reversing those two would have left the set uncleared.
- **The result parses and is coherent.** `AddonManager` and `isPending` are already in scope in this module, so both insertions resolve.
- **The tests discriminate.** Against the patched file all four pass; against the unpatched 153.1.0 file 1, 2 and 4 fail and 3 passes. So the suite is a genuine positive control on the patch having applied, not just on it having run.
- **Version drift fails closed.** `replace_line`/`remove_line` abort the build when a pattern is absent, which is the guard that actually protects the next Firefox uplift.

### Findings (none blocking)

1. **`return` inside the teardown loop abandons the rest of the snapshot.** On reconnect mid-loop, the early `return` skips every remaining entry, and those entries were already removed from `pendingUninstallAddons` by the `clear()` above. I confirmed this with a two-add-on snapshot where the lookup for the first flips `isConnected`: the second is neither finalized nor still in the set. It is recoverable only because `connectedCallback` → `render()` → `getAddons()` re-adds pending-uninstall add-ons — and that recovery holds only if every section `filterFn` excludes them, which is an assumption about the caller, not about this file. `continue` (re-checking `isConnected` per iteration) would not depend on it.

2. **The correctness of the guard rests on `AddonWrapper` identity, and its failure mode is silent.** `current === addon` discriminates old wrapper from replacement only because the wrapper is cached per underlying add-on record. If that ever stops holding, genuine pending uninstalls silently stop being finalized at teardown (deferred to the next restart) — and no test here can catch it, since the tests hand back the same object by construction. `current && isPending(current, "uninstall")` on the freshly looked-up add-on, calling `current.uninstall()`, gives the same protection for the replacement case without depending on identity at all. Worth considering as a hardening.

3. **`disconnectedCallback` is now async and nothing awaits it, so a rejection is unhandled.** If `AddonManager.getAddonByID` rejects (e.g. during shutdown), the returned promise rejects with no consumer, and every remaining pending uninstall in the snapshot is dropped from an already-cleared set. A `try`/`catch` around the lookup would confine the damage to one add-on.

4. **Test coverage: the suite is real but unreachable.** `app/scripts/tests/` is new and holds only this file; there is no npm script, no CI step, and the test needs an argument pointing at a file extracted from a built `omni.ja`. It will not run unless someone remembers to run it, so it cannot protect the next uplift — the `replace_line` abort is doing that work. Two gaps inside the suite itself: no case with more than one add-on in the snapshot (which is exactly where finding 1 hides — the reconnect test asserts only that the newly queued add-on survives, not what happened to the one it was processing), and no case where `getAddonByID` resolves to `null`/`undefined` or rejects.

5. **Minor, build-script robustness.** `remove_line` is `grep -v`, so it deletes *every* matching line, and `replace_line`'s `s///` applies to every matching line. Both are single-match today; if a future Firefox adds a second occurrence of either anchor, the build changes more than intended and stays silent about it. `check_line`-style assertions, or anchoring the patterns, would keep that loud.

None of these can cause the behaviour the PR fixes — removing the freshly installed copy. The worst case in each is a removal deferred to the next restart, against a pre-patch worst case of destroying a just-installed plugin, so the change is a clear net improvement and I would not hold it for any of the above.

MinhHaDuong (commented):
BLOCKED — Independent internal re-review of c1bf7a4865ad3bd20a80c31d82de7f9a33e7ed4f. The reconnect restoration can reintroduce a stale pending-uninstall wrapper after a replacement installation. During the awaited getAddonByID(old.id), the snapshot has already been removed from pendingUninstallAddons. If onInstalled(replacement) fires then, it finds no old entry to clear. Once the list reconnects, the callback restores old because isPending(old, "uninstall") is still true, even though current is the replacement. A failed lookup can similarly requeue old after the replacement event. This can make the old pending-removal bar reappear; reconcile restored entries with the current add-on or record replacement events across the await, and test that interleaving. The six current focused tests do not cover it.

MinhHaDuong (commented):
APPROVED — Independent internal re-review of 66c1f8a89891b93a25b0338d51281ba1bf3e259a. The live pending set now remains available to onInstalled while getAddonByID is awaited, so a replacement event removes the stale wrapper. The callback checks membership and current wrapper identity before finalizing; reconnect and lookup failure return with pending entries intact. The new interleaving test covers my previous blocker, and the seven focused tests exercise the targeted paths against the transformed Firefox module. I found no remaining blocking issue in this diff. This verdict covers code review and focused module tests, not an end-to-end Zotero UI run.

MinhHaDuong (commented):
**Re-review at head `66c1f8a89891b93a25b0338d51281ba1bf3e259a`** — Claude Opus, re-reading after my earlier pass on `2fe028c`.

**Verdict: APPROVED.**

### What I ran

I rebuilt the patched file rather than reading the diff alone. I fetched `toolkit/mozapps/extensions/content/components/addon-list.mjs` at tag `FIREFOX_153_1_0esr_RELEASE` (matching `GECKO_VERSION_*=153.1.0esr` in `app/config.sh`), applied the five `replace_line` calls verbatim through the real `replace_line` from `app/scripts/utils.sh`, and ran the new test file against the result.

- Patched: **7/7 pass**.
- Positive control, same test file against the **unpatched** 153.1.0esr source: **6 of 7 fail**, and the one that passes is `a genuinely pending removal still finalizes` — exactly the regression guard that should pass on both. So the suite discriminates; it is not green by construction.
- All five patterns are unique in the 153.1.0esr file, which matters because `replace_line` is `perl -pi -e` and rewrites *every* matching line. In particular `      if (isPending(addon, "uninstall")) {` does not match the `} else if (...)` occurrence in `getAddons()` (the six spaces do not immediately precede `if`), and `for (const addon of ...)` does not match the `for (let addon of ...)` in `render()`. `bash -n app/scripts/fetch_xulrunner` passes.

### The two findings from my earlier pass

Both are addressed, and I checked the mechanism rather than the assertion:

1. **Multi-add-on reconnect.** The `[...this.pendingUninstallAddons]` snapshot plus the per-iteration `this.isConnected` check means a reconnect mid-loop stops teardown and leaves every unprocessed entry in the live set. The early `return` also skips the terminal `this.pendingUninstallAddons.clear()`, which is what actually preserves them — worth knowing, since that `clear()` is not visible in the diff.
2. **Failed lookup.** The `catch` retains the entry and reports the error; no `uninstall()` is issued on an unknown state.

### The internal finding — stale snapshot restoration after an install event

Also addressed, and correctly: the loop iterates the snapshot but re-checks `this.pendingUninstallAddons.has(addon)` before acting, so an entry deleted by `onInstalled` during the lookup is visited and then skipped rather than resurrected. Nothing reinstates the snapshot into the set.

One observation on the test that covers it. `disconnectedCallback` calls `this.removeListener()` *before* the loop, and `AddonManagerListenerHandler.callListeners` dispatches synchronously over the live `Set`, so after that line no add-on event can reach the list until it reconnects — and a reconnect trips the `isConnected` guard first. The interleaving in `install notification during lookup removes the stale entry before reconnect` therefore does not seem reachable in production: the test stubs `removeListener`, which is what makes it reachable in the harness. The `has()` re-check is still justified by the snapshot, so this is a note on what the test evidences, not a defect.

### The identity check

`current === addon` is load-bearing and rests on `XPIDatabase`'s wrapper caching: `AddonInternal.wrapper` memoises one `AddonWrapper` per `_wrapper` field, so an unchanged add-on returns the same object and a replacement — a new `AddonInternal` — returns a different one. That is exactly the discrimination the patch needs, but it is a provider implementation detail rather than an `AddonManager` API guarantee. If it ever stopped holding, the failure direction is benign: pending uninstalls stop being finalized at teardown and fall through to the restart path, instead of the wrong copy being removed. Same benign direction for every other new guard here, which is the right bias for this bug.

### Two non-blocking notes, both measured

**`return` rather than `continue` on lookup failure.** A failure on one entry abandons the rest of the loop. I ran two pending uninstalls where the first lookup rejects: the second add-on's genuine pending removal is not finalized, the set keeps both entries, and `clear()` is skipped. Safe (AddonManager applies the pending uninstall at restart), but `continue` would confine the loss to the entry that actually failed, and no test covers a sibling after a failure.

**Concurrent teardowns after remove → re-add → remove.** Because teardown is now async, a second `disconnectedCallback` can start while the first is still awaiting. Probe: `a` pending, teardown starts and awaits; the element reconnects; `c` is queued pending-uninstall; the element is removed again, starting a second teardown that snapshots `[a, c]`; the first lookup then resolves. Result — `a` finalized once (no double `uninstall()`, the `has()`/`delete()` pair is atomic against the second loop), but `c` is never finalized and is dropped from the set by the first teardown's terminal `clear()`. The unpatched synchronous code would have finalized `c`. Very narrow, benign direction, and I would not hold the PR for it; it is worth a line in the comment block so the next reader knows the async teardown admits overlap.

Neither note changes the verdict. The patch fixes a real data-loss path — finalizing a stale wrapper's pending uninstall removes the replacement by ID — and every new branch it adds fails toward keeping the add-on.

*Reviewed by Claude Opus. No files were edited and nothing was merged as part of this review.*

PR #6055, opened 2026-09-22
Include plugin toolbar buttons in keyboard focus route

MinhHaDuong:
## Summary

Plugin toolbar buttons with `tabindex="0"` are focusable, but Zotero's explicit keyboard route skips them. Include eligible appended plugin buttons after the built-in toolbar controls in Tab traversal, before the item tree, and route Shift+Tab back through them. Arrow navigation moves among adjacent plugin buttons without crossing the search control.

The route reads the current buttons for each key event, so installing or removing a plugin needs no registration step. The test covers two plugin buttons, forward and reverse traversal, arrow keys, and hidden, disabled, and display-none buttons.

## Test

- `node --check chrome/content/zotero/zoteroPane.js`
- `node --check test/tests/zoteroPaneTest.js`

Both syntax checks passed. The Zotero UI test has not been run in this checkout.


MinhHaDuong (commented):
APPROVED — I reviewed head 785e3633903c05436f1859ed589bca2b6e43fdd4. The Tab route follows the appended toolbar buttons into the item tree, Shift+Tab returns to the last plugin button, and arrow navigation includes eligible plugin buttons. The regression test covers both tree transitions and skips hidden buttons. I found no blocking issue. The Zotero UI test was not run in my review environment.

MinhHaDuong (commented):
**Independent review — I am Claude Opus**, reviewing head `785e3633903c05436f1859ed589bca2b6e43fdd4`. I read the diff against a local checkout at that commit, plus `zoteroPane.xhtml` and `moveFocus()`, and formed my own conclusions rather than checking the existing discussion.

**Verdict: BLOCKED** (changes requested, not a rejection). Nothing here crashes or is dead code, but two design points about focus order and two gaps in the test are worth settling before this lands.

### What I verified as correct

- **Event routing works.** The `keydown` listener is on `toolbar#zotero-toolbar-item-tree`; plugin buttons appended to `#zotero-items-toolbar` are descendants, so the event bubbles and `actionsMap[event.target.id]` is populated in time. Not dead code.
- **The selector cannot swallow built-ins.** Every `toolbarbutton` under `#zotero-items-toolbar` carries `tabindex="-1"`, so `[tabindex="0"]` selects only added buttons.
- **Off-by-one is right in both directions.** `buttons[index - 1]` at index 0 yields `undefined` (no wrap) and falls through to the `||` fallback, while `pluginButtons().at(-1)` in the item-tree Shift+Tab handler is deliberately the last element. Those are not the same idiom by accident.
- An empty `pluginButtons()` degrades cleanly to the previous routes at all four call sites.

### 1. The opt-in marker contradicts the surrounding markup, and is undocumented

The contract is "set `tabindex="0"` on your toolbarbutton", but every built-in control in this toolbar uses `tabindex="-1"`, and outside the post-upgrade banner link there is no `tabindex="0"` anywhere in `zoteroPane.xhtml`. The pane's whole model is: nothing is a native tab stop, focus is routed manually. Two consequences:

- A plugin author who copies the adjacent built-in markup writes `tabindex="-1"` and is silently excluded — the failure is invisible, with no warning and no fallback.
- `tabindex="0"` makes these buttons real entries in the browser's sequential focus order, inside a pane deliberately built to have none. Any Tab/Shift+Tab that `moveFocus()` does not handle (no `actionsMap` entry → early `return`, no `preventDefault`) now falls through to the native order and can land on a plugin button. I did not enumerate an unmapped Shift+Tab path that reaches it, so I am reporting this as an unverified risk rather than a defect — but it is a behaviour change to the pane's focus model that the PR does not mention.

Concretely: either document the attribute as the registration API (the code comment says "installation needs no registration", which is only true if you already know the attribute is required), or key on something that does not collide with the house convention — e.g. `#zotero-items-toolbar > toolbarbutton:not([tabindex="-1"])`, which catches the common case of a plugin that sets no `tabindex` at all, or a dedicated class/`data-` attribute.

### 2. Arrow navigation now jumps across the quick-search field

`'zotero-tb-note-add'.ArrowNext` becomes `pluginButtons()[0]`. In DOM and visual order the toolbar is: add, lookup, attachment-add, note-add, **search spinner, quick-search textbox, pane toggle**, then plugin buttons. Arrow keys reach none of the three middle controls — the arrow group has always been the contiguous left-hand cluster. After this change, pressing → on "New Note" teleports focus over the search box and the pane toggle to the far-right plugin button, and ← comes back the same way. That is a discontiguous roving group, and it is the kind of thing this file is otherwise careful about. My suggestion: leave the arrow chain ending at `zotero-tb-note-add` and let Tab be the only route into plugin buttons (which the PR already implements), or rule explicitly that arrows should span the whole toolbar — but then the search box and pane toggle should join the chain too.

### 3. The test has a surviving mutant

`ArrowPrevious: () => adjacentPluginButton(event.target, -1) || document.getElementById('zotero-tb-note-add')` — the fallback leg is never asserted. The test does left-arrow only from `buttons[1]` to `buttons[0]`, never from `buttons[0]` back to "New Note". Replace that fallback with `|| null` and the suite stays green. One more `leftArrow` dispatch and an assertion closes it.

Untested for the same reason: the `disabled` and `display: none` legs of the `pluginButtons()` filter (only `hidden` is exercised). Also note the filter checks `button.hidden` but not `button.parentNode.hidden`, which `moveFocus()` does check — harmless while the buttons are direct children of the toolbar, but the two visibility predicates now differ.

### 4. The one Shift+Tab assertion cannot distinguish success from fallback

```js
assert.include(
    ['zotero-tb-toggle-item-pane-stacked', 'zotero-tb-search-advanced-button'],
    doc.activeElement.id
);
```

The pane toggle is "Stacked mode only" per the markup comment, so under the default layout this assertion is almost certainly passing on the *fallback* branch (hidden target → `moveFocus()` re-dispatch → advanced-search button), not on the route the code intends. Accepting both outcomes makes the assertion insensitive to which one occurred. It still catches a total break (focus not moving at all), but not a wrong route. Pinning the layout for that one assertion, or asserting the re-dispatch explicitly, would make it mean something.

Related: the new test calls `this.skip()` with no log line, while the two sibling tests above it log `Zotero.debug("Skipping test -- pane window isn't active")` first. Since this is the PR's only coverage, a silent skip in CI makes the feature look tested when it is not. Worth matching the neighbours.

### Smaller notes

- `paneToggle.dispatchEvent(tab)` is dispatched on an element that is not focused, so the test exercises the `actionsMap` entry but not the "user actually tabbed here" path. Acceptable, given the stacked-mode problem above; just be aware of what it does and does not prove.
- `getComputedStyle()` per button per keydown is negligible at toolbar scale — no objection.

None of the above is hard to fix, and the routing design (read the DOM at event time, no registry) is the right call. I would approve with item 3 closed and a ruling on items 1 and 2.

MinhHaDuong (commented):
APPROVED — I re-reviewed head f55d03891309351bc28ff2c9d9377a6aa3eb8ee0. The arrow route now stays within its toolbar-button group; Tab and Shift+Tab preserve the appended buttons’ DOM order. The updated test checks the exact Shift+Tab fallback and skips hidden, disabled, and display:none plugin buttons. I found no blocking issue. The Zotero UI test was not run in my review environment.

MinhHaDuong (commented):
**Re-review at `f55d03891309351bc28ff2c9d9377a6aa3eb8ee0`** (previous review was at `785e363`, BLOCKED).

Reviewer: Claude Opus, reading the diff and the surrounding `setUpKeyboardNavigation` / `moveFocus` code at this head.

## Verdict: BLOCKED

One blocking finding, test-only and cheap to fix. The two behavioural blockers from the previous round are resolved.

## Resolved since `785e363`

- **Cross-search arrow jump — fixed.** `ArrowNext`/`ArrowPrevious` on a plugin button now resolve through `adjacentPluginButton()`, which indexes with `buttons[index + direction]` (no `.at()` wraparound) and returns `null` at either end. Arrow keys can no longer cross the quick-search control, which matches the built-in convention (`zotero-tb-note-add.ArrowNext: () => null`). The test pins both ends of the group and adds the `zotero-tb-note-add` + ArrowRight control that would catch the regression.
- **Assertions — largely fixed.** Comparing `doc.activeElement` against the element references rather than against ids removes the earlier ambiguity. Forward and reverse traversal, and the `zotero-tb-lookup` / `zotero-tb-attachment-add` Tab routes, are now pinned.
- The route itself reads correctly. `#zotero-items-toolbar > toolbarbutton[tabindex="0"]` cannot capture a built-in: every built-in toolbarbutton in that hbox carries `tabindex="-1"` (`zoteroPane.xhtml` 1246–1353), so there is no loop back into `zotero-tb-add`. Entry from `zotero-tb-toggle-item-pane-stacked` and the `.at(-1)` return path from the item tree are symmetric, and when the pane toggle is itself hidden `moveFocus`'s re-dispatch still reaches the plugin buttons.

## Blocking: the hidden / disabled / display-none cases are asserted but not covered

The PR body states the test "covers ... hidden, disabled, and display-none buttons". It does not. All three assertions pass unchanged if the entire `.filter(...)` in `pluginButtons()` is deleted, because `moveFocus` already has its own hidden/disabled fallback (`zoteroPane.js` ~244–259): it detects `disabled || hidden || parentNode.hidden || display === 'none'` on the target, re-dispatches a synthetic keydown on it, and the re-entered handler forwards to the next plugin button. Traced for each case:

- `buttons[0].hidden = true` — with the filter, `pluginButtons()[0]` is `buttons[1]`; without it, the target is `buttons[0]`, `moveFocus` re-dispatches, `buttons[0]`'s own `Tab` returns `adjacentPluginButton(buttons[0], 1)` = `buttons[1]`. Same `activeElement`.
- `disabled = true` — identical path.
- `style.display = 'none'` — identical path.

Two of the three are additionally non-discriminating for a second reason: focus is already on `buttons[1]` when the dispatch happens (left there by the preceding case), so the assertion holds even if the handler does nothing at all.

I could not construct a traversal in which the filter changes the observed focus outcome; it changes only the number of synthetic re-dispatches. The one behavioural difference I did find runs the wrong way: a plugin button that is focused and then programmatically disabled falls out of `pluginButtons()`, so `pluginButtons().includes(event.target)` is false, no `actionsMap` entry is installed, and `moveFocus` returns without moving focus — a small trap that the unfiltered version does not have, since `moveFocus`'s own fallback would route onward.

So the choice is one of:

1. **Drop the filter** and rely on `moveFocus`'s existing fallback. This is my recommendation: it removes a `getComputedStyle()` style flush that currently runs on the `pluginButtons().includes(event.target)` line for *every* keydown in the item-tree toolbar — including every keystroke typed into quick search — and it removes the disabled-while-focused trap. Then delete the three cases, or keep them as regression coverage of the fallback with the PR body reworded.
2. **Keep the filter** and give it a test that fails without it. It needs a case where the two paths diverge, and a focus reset (`paneToggle.focus()` or focusing another control) before each dispatch so the assertion starts from a state it can actually leave.

Either way the PR body's coverage claim needs to match what the test pins.

## Non-blocking

- **Route order assumes append.** `querySelectorAll` returns DOM order, but the only entry point is Tab from `zotero-tb-toggle-item-pane-stacked`. A plugin that inserts its button *before* `zotero-tb-add` rather than appending gets a route that runs backwards against the visual order: Tab from the pane toggle jumps to the leftmost button. Nothing breaks and nothing becomes unreachable, but it is a WCAG 2.4.3 focus-order mismatch. Worth either anchoring the set to buttons that follow the pane toggle (`compareDocumentPosition`), or stating "appended after the pane toggle" as part of the contract.
- **The opt-in is undocumented outside this PR.** `tabindex="0"` on a direct `toolbarbutton` child of `#zotero-items-toolbar` is now load-bearing plugin API, and the only statement of it is a four-line code comment. The selector is strict in ways a plugin author will not guess: a button wrapped in an `hbox`, or appended to `#zotero-toolbar-item-tree` rather than `#zotero-items-toolbar`, silently gets nothing. This repository has no plugin-API doc to amend, so it is a note for the maintainers rather than a change request on this branch — but the contract should land somewhere a plugin author reads.
- **Arrow from the last plugin button is untested.** `leftArrow` from `buttons[0]` is pinned; `rightArrow` from `buttons[1]` (the other boundary, the one nearest the item tree) is not.
- The test suite has not been run at this head (per the PR body, only `node --check`), so everything above is from reading, not from an executed run. The three findings in the blocking section are traced through the code rather than observed.

No edits or merge actions taken on this branch.

MinhHaDuong (commented):
APPROVED — I re-reviewed head a1d3eea8b25c5c2bc6cdb257b27f8285a198c17a. Keeping disabled and hidden plugin buttons in the route lets the existing moveFocus handler skip them while preserving a route when the current button becomes disabled. The new disabled-button event assertion distinguishes this behavior from the prior filtered implementation. I found no blocking issue. The Zotero UI test was not run in my review environment.

MinhHaDuong (commented):
**Re-review at head `a1d3eea8b25c5c2bc6cdb257b27f8285a198c17a`** (previous review was at `f55d038`).

## Verdict: APPROVED

The blocker from the prior review is resolved.

### Why the previous blocker is gone

At `f55d038`, `pluginButtons()` filtered its own result on `hidden`/`disabled`/`parentNode.hidden`/`getComputedStyle(...).display`. That filter had two costs: it forced a style flush on every key event, and — because `moveFocus`'s re-dispatch trick relies on the route recognising the *hidden* element as a known target — it excluded a button that went disabled while focused, so `pluginButtons().includes(event.target)` was false, no `actionsMap` entry was installed, `focusFunction` came back `undefined`, and focus was trapped on the disabled button.

The new head drops the filter entirely and lets `moveFocus` (zoteroPane.js:244-256) do the skipping, which is the same mechanism the built-in buttons already use. A disabled plugin button is now still a member of `pluginButtons()`, so the fake keydown dispatched on it resolves through the plugin branch and focus moves on. Uniform with the rest of the toolbar, and one less forced reflow per keystroke.

The last block of the new test is a genuine discriminating control for exactly that case:

```js
buttons[0].disabled = true;
doc.getElementById('zotero-tb-add').focus();
buttons[0].dispatchEvent(tab);
assert.equal(doc.activeElement, buttons[1]);
```

At `f55d038` this asserts `zotero-tb-add`; at this head it asserts `buttons[1]`. It fails on the old code and passes on the new — which is what makes it worth having.

### Regression checks I ran against this head

- **Selector cannot sweep built-ins.** Every `toolbarbutton` that is a direct child of `#zotero-items-toolbar` in `zoteroPane.xhtml` carries `tabindex="-1"` (`zotero-tb-add`, `-lookup`, `-attachment-add`, `-note-add`, `-toggle-item-pane-stacked`). `[tabindex="0"]` is a literal attribute match, so the route only picks up opted-in buttons.
- **Event routing works.** `#zotero-items-toolbar` is a direct child of `#zotero-toolbar-item-tree`, which owns the `keydown` listener, so both real events on appended buttons and the `bubbles: true` fake events re-dispatched by `moveFocus` reach the handler.
- **All four boundaries are covered, not just the forward one.** Forward entry `zoteroPane.js:433`, forward exit `:442`, reverse entry `:491`, reverse exit `:444`. No remaining hardcoded hop past the plugin segment.
- **Non-stacked mode still reaches the plugin buttons.** When `zotero-tb-toggle-item-pane-stacked` is hidden, Tab from `zotero-tb-search-advanced-button` targets it, `moveFocus` re-dispatches on the hidden toggle, and its own route returns `pluginButtons()[0]`. The test handles the mirror case explicitly via `toggleHidden`.
- **No unbounded recursion.** `adjacentPluginButton` indexes rather than wrapping, `buttons[-1]` and `buttons[n]` are `undefined`, so a run of hidden plugin buttons drains to the item tree or the toggle and stops.

### Non-blocking notes

1. **Edge arrows swallow the event.** `ArrowNext: () => adjacentPluginButton(event.target, 1) || null` returns `null` at the ends of the run, and `moveFocus` treats `null` as handled: it `preventDefault()`s and stops. Built-in edge buttons instead omit the key, letting default handling run. The PR description says arrows should not cross the search control, so this reads deliberate, but it is a behavioural difference from the neighbouring entries and not obvious from the code. Returning `false` is the documented "not handled, fall back" value if the intent is ever the other one.
2. **The `|| null` is redundant.** `adjacentPluginButton` already returns `null` for an unknown button and `undefined` past the end; `undefined` and `null` take the same branch in `moveFocus`.
3. **DOM order is the contract, and it is unstated.** `querySelectorAll` yields document order, so a plugin that inserts its button *before* `zotero-tb-search` still lands after the pane toggle in Tab order — visual position and focus order would disagree. Worth one line in the comment at `:177` saying the route assumes appended-last, since that is the assumption the whole segment rests on.
4. **Empty or colliding ids.** `actionsMap[event.target.id] = {...}` writes `actionsMap['']` for a button with no id, which still resolves correctly. But a plugin button whose id duplicates a built-in toolbar id would shadow that built-in's route for the event. Duplicate ids are already invalid, so this is a low-severity robustness note rather than a defect.
5. **The test is unexecuted.** The PR body reports `node --check` only. The discriminating assertion above is the strongest thing in this change and no run of it is on record yet; a maintainer run of `zoteroPaneTest.js` is what would convert it from a well-formed control into evidence. Two things I could not verify statically and that a run would settle: whether a bare `createXULElement('toolbarbutton')` with no label, image or `class="zotero-tb-button"` has non-zero layout and accepts focus, and whether the reused `KeyboardEvent` instances behave across this many dispatches.

Nothing in 1-5 blocks the change.

---
Reviewed by Claude Opus at head `a1d3eea8b25c5c2bc6cdb257b27f8285a198c17a`. Static review of the diff plus `zoteroPane.js` and `zoteroPane.xhtml` at that SHA; the UI test suite was not run.

